PR #1839, opened 2025-06-09
Update Spark Client license 

gh-yzou:
The following two files does not contain copy from Iceberg library:
 plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/PolarisInMemoryCatalog.java

Remove them from the license

### DIFF
--- a/plugins/spark/v3.5/spark/LICENSE
+++ b/plugins/spark/v3.5/spark/LICENSE
@@ -204,8 +204,6 @@
 
 This product includes code from Apache Iceberg.
 
-* plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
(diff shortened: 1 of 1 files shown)


gh-yzou:
@flyrain @jbonofre i removed two files that does not contain copy from iceberg in the license, please help take a look

flyrain (approved):

plugins/spark/v3.5/spark/LICENSE
```diff
@@ -204,8 +204,6 @@
 
 This product includes code from Apache Iceberg.
 
-* plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
```
jbonofre, line 207: Then https://github.com/apache/polaris/blob/main/LICENSE should be updated as well.

PR #1840, opened 2025-06-09
Improve the parsing and validation of UserSecretReferenceUrns

poojanilangekar:
This change addresses all the TODOs found the org.polaris.core.secrets package. 
Main changes:

- Create a helper to parse, validate and build the URN strings. 
- Use Regex instead of `String.split()`.
- Add Precondition checks to ensure that the URN is valid and the UserSecretManager matches the expected type. 
- Remove the now unused `GLOBAL_INSTANCE` of the UnsafeInMemorySecretsManager.

Testing 
- Existing `UnsafeInMemorySecretsManagerTest` captures most of the functional changes. 
- Added `UserSecretReferenceUrnHelperTest` to capture the utilities exposed.


poojanilangekar:
CC @dennishuo @eric-maynard @XJDKC

polaris-core/src/main/java/org/apache/polaris/core/secrets/UnsafeInMemorySecretsManager.java
```diff
@@ -73,9 +73,9 @@ public UserSecretReference writeSecret(
 
     String secretUrn;
     for (int secretOrdinal = 0; ; ++secretOrdinal) {
+      String typeSpecificIdentifier = forEntity.getId() + ":" + secretOrdinal;
       secretUrn =
-          String.format(
-              "urn:polaris-secret:unsafe-in-memory:%d:%d", forEntity.getId(), secretOrdinal);
+          UserSecretReferenceUrnHelper.buildUrn(SECRET_MANAGER_TYPE, typeSpecificIdentifier);
```
eric-maynard: Can we not just add `buildUrn` or something like it to `UserSecretsManager`?

polaris-core/src/main/java/org/apache/polaris/core/secrets/UnsafeInMemorySecretsManager.java
```diff
@@ -45,6 +43,8 @@ public class UnsafeInMemorySecretsManager implements UserSecretsManager {
   private static final String CIPHERTEXT_HASH = "ciphertext-hash";
   private static final String ENCRYPTION_KEY = "encryption-key";
 
+  private static final String SECRET_MANAGER_TYPE = "unsafe-in-memory";
```
eric-maynard: Can we make a static constant like `UnsafeInMemorySecretsManager.prefix`? (NID?)

polaris-core/src/main/java/org/apache/polaris/core/secrets/UserSecretReference.java
```diff
@@ -88,8 +88,7 @@ public UserSecretReference(
    */
   @JsonIgnore
   public String getUserSecretManagerTypeFromUrn() {
-    // TODO: Add better/standardized parsing and validation of URN syntax
-    return urn.split(":")[2];
+    return UserSecretReferenceUrnHelper.getSecretManagerType(urn);
```
eric-maynard: It might be nice just to offer something like `UserSecretReference usr = UserSecretReference.fromUrn(...)`

polaris-core/src/main/java/org/apache/polaris/core/secrets/UserSecretReferenceUrnHelper.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /**
+   * Extracts the type-specific identifier from a valid URN.
+   *
+   * @param urn The URN string to parse.
+   * @return The type-specific identifier.
+   */
+  @Nonnull
+  public static String getTypeSpecificIdentifier(@Nonnull String urn) {
```
eric-maynard: Where do we validate the returned value?

polaris-core/src/main/java/org/apache/polaris/core/secrets/UnsafeInMemorySecretsManager.java
poojanilangekar: Done, I've made it public.

polaris-core/src/main/java/org/apache/polaris/core/secrets/UnsafeInMemorySecretsManager.java
poojanilangekar: Done.

Although I might have missed something here, please let me know if you think the buildUrn function meets your intended goal.

polaris-core/src/main/java/org/apache/polaris/core/secrets/UserSecretReference.java
poojanilangekar: At the moment, we don't keep a registry of all the UserSecretReferences created by a given Polaris instance. 

Or are you suggesting something different?

polaris-core/src/main/java/org/apache/polaris/core/secrets/UserSecretReferenceUrnHelper.java
poojanilangekar: Currently, the function is never called (except in tests). I added it here to support any future use cases that may arise. 

From a validation perspective, I think it is the responsibility of the caller because the type-specific format is unknown to this function. Unless you're suggesting that we accept a regex of the expected format.

poojanilangekar (commented):
Hey Eric, thanks for taking time to review the PR. I have implemented some of the changes you suggested and wanted to get some more clarity on the rest.

polaris-core/src/main/java/org/apache/polaris/core/secrets/UserSecretReference.java
eric-maynard: I'm wondering why we need `UserSecretReferenceUrnHelper` rather than just using `UserSecretReference`. Right now it seems like the responsibility of parsing a URN has been shifted to `UserSecretReferenceUrnHelper`, but it's only ever done in the context of a `UserSecretReferece` right? So why not keep the code in `UserSecretReference`?

poojanilangekar: I did this to avoid mixing static and non-static methods in the `UserSecretReference` class. Currently, `UserSecretReference` is only instantiated when we have a secret that needs to stored and retrieved. 

If you think that's okay, I can remove the helper and move its functionality to `UserSecretReference`.

eric-maynard: Yeah, I thinking mixing static and nonstatic methods is fine (and an intended feature of the language, as opposed to e.g. Scala where we must separate statics into an object and nonstatics into a class). If it keeps the related logic together without making that class too huge, let's try it.

poojanilangekar: Done.

polaris-core/src/main/java/org/apache/polaris/core/secrets/UnsafeInMemorySecretsManager.java
eric-maynard: Yeah it looks good, though seems like callers could just directly use `UserSecretReference.buildUrnString`. If there's a situation where the implementation might diverge, having the additional `buildUrn` method makes sense. My original comment was more about the necessity of `UserSecretReferenceUrnHelper`.

poojanilangekar: Makes sense.

Issue #1848, opened 2025-06-10
Pagination: Failed to parse pageSize


XJDKC:
### Describe the bug

Got this error message when testing Catalog Federation:
```
{
  "error": {
    "message": "For input string: \"null\"",
    "type": "NumberFormatException",
    "code": 400
  }
}
```

### Additional context

The tokenSize param in the request in an `Integer` object, but later we convert it to a String. Then we try to parse it to get an integer again. If the pageSize is null, we will fail to parse the string and convert it to Integer.

[CatalogHandlerUtils.java#L164](https://github.com/apache/polaris/blob/5fe9fd9e5b7e7db5604011cf2de4197430882cef/service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java#L164)

snazy:
@XJDKC is this still an issue?

XJDKC:
@snazy It has been resolved in this PR: https://github.com/apache/polaris/pull/1849

flyrain:
Can we close it, @XJDKC?

XJDKC:
> Can we close it, [@XJDKC](https://github.com/XJDKC)?

Done!

PR #1849, opened 2025-06-10
Fix Pagination for Catalog Federation

XJDKC:
Details can be found in this issue: https://github.com/apache/polaris/issues/1848


eric-maynard:
@dimas-b pointed out that we could actually pass the pagination info down to the other catalog, which I agree with. But this change LGTM

eric-maynard (approved):

dimas-b (approved):
LGTM :+1: Thanks, @XJDKC !

PR #1830, opened 2025-06-07
Add license and notice to spark client jar and push jar to maven

gh-yzou:
This PR does following:
1) add default license and notice to spark client jar
2) update the spark client jar name to align with polaris published jar
3) enable publish spark client jar to maven

PR #1303, opened 2025-04-04
Add CreateTable and LoadTable implementation for SparkCatalog

gh-yzou:
Overall architecture for SparkCatalog

Screenshot 2025-04-03 at 10 18 55 AM

This PR provides the implementation for createTable and loadTable, which includes
1) PolarisRestCatalog that makes rest call to generatic table APIs
2)PolarisSparkCatalog that implements the Spark TableCatalog for non-iceberg tables
3)SparkCatalog that initiates both Iceberg SparkCatalog and PolarisSparkCatalog, and implements the following methods
    a) createNamespace
    b) loadNamespaceMetadata
    c) createTable
    d) loadTable
4) InMemoryCatalog classes for unit testing purpose

Follow up tasks will coming:
1) Support for all namespace and view related APIs,
2) Support for other table related APIs
3) end to end spark integration tests

PR #1857, opened 2025-06-11
Reuse shadowJar for spark client bundle jar maven publish

gh-yzou:
We previously added a special check in PublishingHelperPlugin.kt to check specifically for the jar job for polaris-spark project, and publish the artifact output for the ShadowJar Task added. However, we already have a shadowJar infra that takes care of the maven publish. 
In this PR, we switch to reuse the shardowJar Infra, and reverted the change we added before.

### DIFF
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -42,7 +42,7 @@ The result jar is located at plugins/spark/v3.5/build/<scala_version>/libs after
 Once the jar is built, we can manually test it with Spark and a local Polaris service.
 
 The following command starts a Polaris server for local testing, it runs on localhost:8181 with default
-realm `POLARIS` and root credentials `root:secret`:
+realm `POLARIS` and root credentials `root:secret`:./gr
 ```shell
 ./gradlew run
 ```
@@ -67,12 +67,12 @@ bin/spark-shell \
 ```
 
 Assume the path to the built Spark client jar is
-`/polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT.jar`
+`/polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar`
 and the name of the catalog is `polaris`. The cli command will look like following:
 
 ```shell
 bin/spark-shell \
---jars /polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT.jar \
+--jars /polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar \
 --packages org.apache.iceberg:iceberg-aws-bundle:1.9.0,io.delta:delta-spark_2.12:3.3.1 \
 --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,io.delta.sql.DeltaSparkSessionExtension \
 --conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -38,7 +38,7 @@ A shadowJar task is added to build a jar for the Polaris Spark plugin, the jar i
 
 The result jar is located at plugins/spark/v3.5/build/<scala_version>/libs after the build.
 
-The shadowJar task is also executed automatically when you run the `gradlew assemble` or `gradlew build`.
+The shadowJar task is also executed automatically when you run `gradlew assemble` or `gradlew build`.
 
 # Start Spark with Local Polaris Service using built Jar
 Once the jar is built, we can manually test it with Spark and a local Polaris service.
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -29,12 +29,12 @@ Right now, the plugin only provides support for Spark 3.5, Scala version 2.12 an
 and depends on iceberg-spark-runtime 1.9.0.
 
 # Build Plugin Jar
-A task createPolarisSparkJar is added to build a jar for the Polaris Spark plugin, the jar is named as:
+A shadowJar task is added to build a jar for the Polaris Spark plugin, the jar is named as:
 `polaris-spark-<sparkVersion>_<scalaVersion>-<polarisVersion>-bundle.jar`. For example:
 `polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar`.
 
-- `./gradlew :polaris-spark-3.5_2.12:createPolarisSparkJar` -- build jar for Spark 3.5 with Scala version 2.12.
-- `./gradlew :polaris-spark-3.5_2.13:createPolarisSparkJar` -- build jar for Spark 3.5 with Scala version 2.13.
+- `./gradlew :polaris-spark-3.5_2.12:shadowJar` -- build jar for Spark 3.5 with Scala version 2.12.
+- `./gradlew :polaris-spark-3.5_2.13:shadowJar` -- build jar for Spark 3.5 with Scala version 2.13.
 
 The result jar is located at plugins/spark/v3.5/build/<scala_version>/libs after the build.
 
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -38,11 +38,13 @@ A shadowJar task is added to build a jar for the Polaris Spark plugin, the jar i
 
 The result jar is located at plugins/spark/v3.5/build/<scala_version>/libs after the build.
 
+The shadowJar task is also executed automatically when you run the `gradlew assemble` or `gradlew build`.
+
 # Start Spark with Local Polaris Service using built Jar
 Once the jar is built, we can manually test it with Spark and a local Polaris service.
 
 The following command starts a Polaris server for local testing, it runs on localhost:8181 with default
-realm `POLARIS` and root credentials `root:secret`:./gr
+realm `POLARIS` and root credentials `root:secret`:
 ```shell
 ./gradlew run
 ```
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -67,12 +67,12 @@ bin/spark-shell \
 ```
 
 Assume the path to the built Spark client jar is
-`/polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar`
+`/polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT.jar`
 and the name of the catalog is `polaris`. The cli command will look like following:
 
 ```shell
 bin/spark-shell \
---jars /polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar \
+--jars /polaris/plugins/spark/v3.5/spark/build/2.12/libs/polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT.jar \
 --packages org.apache.iceberg:iceberg-aws-bundle:1.9.0,io.delta:delta-spark_2.12:3.3.1 \
 --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,io.delta.sql.DeltaSparkSessionExtension \
 --conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -113,7 +113,6 @@ dependencies {
 }
 
 tasks.register<ShadowJar>("createPolarisSparkJar") {
-  archiveClassifier = null
   isZip64 = true
 
   // include the LICENSE and NOTICE files for the shadow Jar
@@ -141,4 +140,10 @@ tasks.register<ShadowJar>("createPolarisSparkJar") {
 
 tasks.withType(Jar::class).named("sourcesJar") { dependsOn("createPolarisSparkJar") }
 
-tasks.named<Jar>("jar") { archiveClassifier.set("internal") }
+tasks.named<Jar>("jar") {
+  // retain the default jar job, and add a classifier to avoid conflict
+  // with the createPolarisSparkJar. This jar is needed by the task "test",
+  // which can not be switched to depends on createPolarisSparkJar due to
+  // relocation of com.fasterxml.
+  archiveClassifier.set("internal")
+}
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -141,3 +141,5 @@ tasks.named<ShadowJar>("shadowJar") {
   relocate("com.fasterxml", "org.apache.polaris.shaded.com.fasterxml")
   relocate("org.apache.avro", "org.apache.polaris.shaded.org.apache.avro")
 }
+
+tasks.withType(Jar::class).named("sourcesJar") { dependsOn("shadowJar") }
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -141,4 +141,4 @@ tasks.register<ShadowJar>("createPolarisSparkJar") {
 
 tasks.withType(Jar::class).named("sourcesJar") { dependsOn("createPolarisSparkJar") }
 
-tasks.named<Jar>("jar") { archiveClassifier.set("defaultJar") }
+tasks.named<Jar>("jar") { archiveClassifier.set("internal") }
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -19,7 +19,10 @@
 
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 
-plugins { id("polaris-client") }
+plugins {
+  id("polaris-client")
+  id("com.gradleup.shadow")
+}
 
 // get version information
 val sparkMajorVersion = "3.5"
@@ -112,7 +115,8 @@ dependencies {
   }
 }
 
-tasks.register<ShadowJar>("createPolarisSparkJar") {
+tasks.named<ShadowJar>("shadowJar") {
+  archiveClassifier = null
   isZip64 = true
 
   // include the LICENSE and NOTICE files for the shadow Jar
@@ -136,14 +140,4 @@ tasks.register<ShadowJar>("createPolarisSparkJar") {
 
   relocate("com.fasterxml", "org.apache.polaris.shaded.com.fasterxml")
   relocate("org.apache.avro", "org.apache.polaris.shaded.org.apache.avro")
-}
-
-tasks.withType(Jar::class).named("sourcesJar") { dependsOn("createPolarisSparkJar") }
-
-tasks.named<Jar>("jar") {
-  // retain the default jar job, and add a classifier to avoid conflict
-  // with the createPolarisSparkJar. This jar is needed by the task "test",
-  // which can not be switched to depends on createPolarisSparkJar due to
-  // relocation of com.fasterxml.
-  archiveClassifier.set("internal")
 }
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -142,7 +142,10 @@ tasks.named<ShadowJar>("shadowJar") {
   relocate("org.apache.avro", "org.apache.polaris.shaded.org.apache.avro")
 }
 
-tasks.withType(Jar::class).named("sourcesJar") { dependsOn("shadowJar") }
+// ensure the shadowJar job is run for both `assemble` and `build` task
+tasks.named("assemble") { dependsOn("shadowJar") }
+
+tasks.named("build") { dependsOn("shadowJar") }
 
 tasks.named<Jar>("jar") {
   // retain the default jar job, and add a classifier to avoid conflict
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -143,3 +143,11 @@ tasks.named<ShadowJar>("shadowJar") {
 }
 
 tasks.withType(Jar::class).named("sourcesJar") { dependsOn("shadowJar") }
+
+tasks.named<Jar>("jar") {
+  // retain the default jar job, and add a classifier to avoid conflict
+  // with the createPolarisSparkJar. This jar is needed by the task "test",
+  // which can not be switched to depends on createPolarisSparkJar due to
+  // relocation of com.fasterxml.
+  archiveClassifier.set("internal")
+}
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -116,7 +116,7 @@ dependencies {
 }
 
 tasks.named<ShadowJar>("shadowJar") {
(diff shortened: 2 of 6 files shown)


plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -178,3 +178,5 @@ tasks.register<ShadowJar>("createPolarisSparkJar") {
 }
 
 tasks.withType(Jar::class).named("sourcesJar") { dependsOn("createPolarisSparkJar") }
+
+tasks.named<Jar>("jar") { archiveClassifier.set("testJar") }
```
snazy: `testJar`??

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -151,7 +151,7 @@ tasks.register("checkNoDisallowedImports") {
 tasks.named("check") { dependsOn("checkNoDisallowedImports") }
 
 tasks.register<ShadowJar>("createPolarisSparkJar") {
-  archiveClassifier = "bundle"
+  archiveClassifier = null
```
dimas-b: What is it when we do not override it to `null`?

gh-yzou: the original name is something like polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar, we did this because the jar without classifier is taken by the default jar job with name polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT.jar. However, Spark does not support using classifier in the package config, so we make this jar the jar for this project, since this jar is the actual jar needed by spark, i think it actually should be the jar project without any classifier

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: sorry for the confusion, that is a bad name. I am just referring the original default jar job,  updated the classifier to  "defaultJar" to be more clear.

plugins/spark/v3.5/spark/build.gradle.kts
dimas-b: Yes, I understand the intent :) My question is about the need to set `archiveClassifier` to `null`... Do we have to use `null` here?

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -178,3 +178,5 @@ tasks.register<ShadowJar>("createPolarisSparkJar") {
 }
 
 tasks.withType(Jar::class).named("sourcesJar") { dependsOn("createPolarisSparkJar") }
+
+tasks.named<Jar>("jar") { archiveClassifier.set("defaultJar") }
```
dimas-b: Why not remove the plain `jar` artifact from this module completely?

gh-yzou: I tried that before, however, it seems the task `test` depends on the jar job in the default configuration. i tried to switch the `test` task to depends on the createPolarisSparkJar, but because that jar job did a relocation of module com.fasterxml, one of our test fails the deserialization test, because it is now looking for the shaded one, not the original one. 
So far I haven't found a good solution yet, so I kept the original jar. wondering if you got some better solutions for this problem?

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: Oh, sorry, we don't have to, the default is null. i was putting it there to be clear, and I can remove it if preferred, but I think it might be better to be more explicit in the code.

plugins/spark/v3.5/spark/build.gradle.kts
dimas-b: Thanks for the detailed analysis, @gh-yzou ! Unfortunately, I do not have a better solution off the top of my head.

dimas-b: How about using the `internal` classifier for this jar? I suppose it is not meant for reuse.

plugins/spark/v3.5/spark/build.gradle.kts
dimas-b: From my POV removing the assignment is preferable since the value is the same as default.

dimas-b: I'd prefer to have a comment about adding a classifier to the `jar` task instead.

site/content/in-dev/unreleased/polaris-spark-client.md
```diff
@@ -128,3 +128,21 @@ The Polaris Spark client has the following functionality limitations:
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
 5) For other non-Iceberg tables like csv, it is not supported.
+
+## Iceberg Spark Runtime Client compatibility with Polaris Spark Client
```
dimas-b, line 132: side question: what is the Iceberg artifact called "runtime" but the corresponding Polaris artifact is a "client"?

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: yes, it is not intended for reuse. The name "internal" make sense to me, upated

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: sg! i removed the specification of the classifier, and added a comment at the place where i added the classifier for the jar task

dimas-b (approved):
LGTM :+1: Thanks, @gh-yzou !

gh-yzou:
@dimas-b i think you asked a question somewhere, but it doesn't show up in the PR for some reason. For the artifact, i don't think we have "client" in the artifact name, the iceberg one is called iceberg-spark-runtime-xxxx.jar, and our polaris one is called polaris-spark-xxx.jar. For iceberg, i guess the reason is that iceberg-spark is already taken by another projects, but i don't think we need to be exactly the same as iceberg. 
Some of the doc description might introduce some confusion, i went one more pass to make sure the description are  more consistent.

dimas-b:
Re: `polaris-spark-xxx.jar.` it is not really related to this PR :)

I value short jar names, but at the same time it might be worth clarifying whether this jar applies to the whole of Polaris integration with Spark or just to Generic Tables.

In other words, do we foresee making any other Polaris jars to be put on the Spark class path?

If no, the current name is fine from my POV, if yes, let's discuss that naming convention on the `dev` ML (since it's not about this build change really).

dimas-b:
Let's hold the merge until tomorrow in case @snazy has any suggestions for gradle script changes.

gh-yzou:
The current jar is actually for the whole polaris spark integration, it works for both IRC endpoints and generic table endpoints. Eventually can be extended to talk to other Polaris specific endpoints, which i think is easier for users. So far, i don't expect that is the only Polaris jar need to be added for spark. If in the future we want to split that jar into multiple jars, we can definitely bring that up to discussion.

snazy (commented):
I wondered how the shadow-jar actually gets published, and realized that [this change](https://github.com/apache/polaris/pull/1830/files#diff-d380e458e986c6183b1e23fcea1169811acfc68b2662ff9c88bb26b821238607R137-R140) (created and merged on Saturday) added a project specific change to a generic build convention plugin.

I'd prefer to remove the above change and rely on the publishing-support of the shadow plugin itself and the already present shadow-jar-support right above the change added above.

Referring to the `ShadowJar` task implementationwithout applying the shadow-plugin (`createPolarisSparkJar` task) should also be avoided. This relies on a side-effect.

Also noticed that the `checkNoDisallowedImports` task is not cachable nor does it support up-to-date checking, leading to longer CI times.

adutra:
> Also noticed that the `checkNoDisallowedImports` task is not cachable nor does it support up-to-date checking, leading to longer CI times.

Should be fixed by https://github.com/apache/polaris/pull/1880

gh-yzou:
@snazy Thanks for the pointer! i didn't realize that we have that plugin, I updated the jar to use the ShadowPlugin and just uses task shadowJar, and reverted the change we added in the publishHelperPlugin. 
For the checkNoDisallowedImports, as @adutra mentioned there is a follow up to improve that in a separate PR.

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: @dimas-b after switch to use the shadowJarPlugin, i need to specify the classifier here, otherwise, it seems configuring to generate a jar with classifier "all", but I was also able to get rid of the other jar change

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -179,3 +179,5 @@ tasks.named<ShadowJar>("shadowJar") {
   relocate("com.fasterxml", "org.apache.polaris.shaded.com.fasterxml")
   relocate("org.apache.avro", "org.apache.polaris.shaded.org.apache.avro")
 }
+
+tasks.withType(Jar::class).named("sourcesJar") { dependsOn("shadowJar") }
```
snazy: Why this depends-on?

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: actually, sorry, it seems still needed, my previous gradlew build seems coming from cache. Added it back!

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: This is to make sure when we run both assemble and build, the shadow Jar job is also run. I can add a depends on for assemble and build independently, but seems adding it for sourceJar can make sure both works. I can add a comment there to be more clear

gh-yzou: Actually, i updated to have separate dependency to be more clear, and added comments

plugins/spark/v3.5/spark/build.gradle.kts
dimas-b: sgtm

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -173,8 +176,19 @@ tasks.register<ShadowJar>("createPolarisSparkJar") {
...
 
-tasks.withType(Jar::class).named("sourcesJar") { dependsOn("createPolarisSparkJar") }
+// ensure the shadowJar job is run for both `assemble` and `build` task
+tasks.named("assemble") { dependsOn("shadowJar") }
+
+tasks.named("build") { dependsOn("shadowJar") }
+
+tasks.named<Jar>("jar") {
+  // retain the default jar job, and add a classifier to avoid conflict
+  // with the createPolarisSparkJar. This jar is needed by the task "test",
```
adutra:
```suggestion
  // with the shadowJar task. This jar is needed by the task "test",
```

site/content/in-dev/unreleased/polaris-spark-client.md
```diff
@@ -128,3 +128,21 @@ The Polaris Spark client has the following functionality limitations:
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
 5) For other non-Iceberg tables like csv, it is not supported.
+
+## Iceberg Spark Client compatibility with Polaris Spark Client
+The Polaris Spark client today is not designed to be used with Iceberg Spark client together. In other words,
+there is no guarantee provided if both `org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:<iceberg_version>` and 
+`org.apache.polaris:polaris-spark-3.5_2.12:<polaris_version>` are provided for the Spark `package` configuration.
+
+In order to help the usage of IcebergExtension use the same client, Polaris Spark client also ships a version of
```
adutra: This needs some rephrasing: "In order to help the usage of IcebergExtension use the same client..."

site/content/in-dev/unreleased/polaris-spark-client.md
```diff
@@ -128,3 +128,21 @@ The Polaris Spark client has the following functionality limitations:
...
+Iceberg Spark client with it, the version information is described in the following table:
+
+| Spark Client Version | Iceberg Spark Client Version |
+|----------------------|------------------------------|
+| 1.0.0                | 1.9.0                        |
+
+## Trouble Shooting
+1. When starting spark, sometimes it complains it failed to download a package, for example,
```
adutra:
```suggestion
## Troubleshooting
1. When starting Spark, sometimes it complains it failed to download a package, for example, 
```

site/content/in-dev/unreleased/polaris-spark-client.md
```diff
@@ -128,3 +128,21 @@ The Polaris Spark client has the following functionality limitations:
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
 5) For other non-Iceberg tables like csv, it is not supported.
+
+## Iceberg Spark Client compatibility with Polaris Spark Client
+The Polaris Spark client today is not designed to be used with Iceberg Spark client together. In other words,
```
adutra: Is this design decision definitive?

I remember I asked @gh-yzou a while ago:

https://github.com/apache/polaris/pull/1303#discussion_r2028560451

Including the entire Iceberg Spark runtime jar in this artifact has a few drawbacks such as: increased jar size (~60Mb), increased complexity for LICENSE and NOTICE files.

Just compiling against the Spark runtime jar without including it would imo solve these problems.

gh-yzou: @adutra Yes, there are certain drawbacks. However, as we mentioned before, there are couple of challenges:
1) some classes will be missing due to reflection (construct based on class name) used in the Iceberg class. 
2) compatibility with other iceberg version needs to be tested, instead of letting user try to figure out what is the version they need to use to make sure thing are working. It would be easier to just include it to start with, so that user don't have to worry about that too much to begin with.

Again, this is just the first release, future improvement can always be made once we add improvement and compatibility testing to provide better guarantee to the user.

site/content/in-dev/unreleased/polaris-spark-client.md
gh-yzou: updated

site/content/in-dev/unreleased/polaris-spark-client.md
gh-yzou: re-phased, hopefully it could make more sense

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: updated

site/content/in-dev/unreleased/polaris-spark-client.md
adutra: @gh-yzou can you clarify item 1 above? I.e. "some classes will be missing due to reflection (construct based on class name) used in the Iceberg class".

For item 2, I'd say this is already the case for anybody. I could mix the wrong Spark runtime jar with an incompatible AWS SDK package and things would break at runtime.

Also, not sure if asking users to _replace_ the Spark runtime jar with the Polaris/Spark runtime jar isn't actually worse than asking them to _add_ an extra jar while keeping the Spark runtime jar they are already using.

This would obviously require us to publish and maintain a detailed matrix compatibility table. But we'll have to do it anyways imo.

How de we consider this issue wrt to the upcoming 1.0 Polaris release? Changing this after the 1.0 release will be impossible, I guess.

gh-yzou: For 1, for example, when we create the catalog, the catalog is created as using
```
catalog = ctor.newInstance();
```
And the ctor is decided by the String in the configuration. For this kind of use case, those class will be missing during package if we just pack the dependent classes.I believe it is the same for auth manager class.


```
How de we consider this issue wrt to the upcoming 1.0 Polaris release? Changing this after the 1.0 release will be impossible, I guess.
```
That should be possible, every project should be allowed to introduce behavior change with established process, for example, iceberg can move on to v2, v3 specs, which is dramatic change for users. 
Especially the whole generic table project is marked as beta for now,  and the spark client is mainly introduced for generic table usage.

RussellSpitzer: I think we really like the idea that we can ship a single jar that a user can use to enable this functionality. A lot of folks do not have access to maven and the ability to share a single jar and begin working is a great help. This comes up a lot.

It also makes sense to me to provide the non-bundled resource just in case users do what to do just-in-time substitutions and possibly shade it other resources. I can say from my own previous experience the single Jar versions are much more popular with Spark plugins.

We could essentially have two projects here (as in the Iceberg repo), the first being an unshaded version of the build which doesn't include any dependencies by default. End users would then download all the dependencies using --spark.packages or what not. 

The second would be a shadow jar project (polaris-spark-catalog-plugin-bundle) or something like that. Which just uses the first project as a dependency and shades everything in it.

---

Now both of these things I don't think are necessarily related or blocking 1.0. If users start using the bundle and like it, then we should keep shipping it. If they also need the non-bundled version we can add that later. Even if we restructure the build it doesn't really matter as long as we keep the artifact names the same.

gh-yzou: @adutra do you think it would be ok that if we deliver both the bundle and non-bundled one to satisfy different use cases? i can work on adding both support

adutra: Just for reference, Dremio Auth Manager took the opposite approach: https://github.com/dremio/iceberg-auth-manager/blob/9cc9459ed5cd0ad30a9c08143c98d8b0c620e3a2/oauth2/runtime/build.gradle.kts#L62-L71

> I think we really like the idea that we can ship a single jar that a user can use to enable this functionality.

Imho for users that can fetch from Maven Central, fetching one or two artifacts really doesn't make any difference. And for the adepts of manual jar downloads, indeed they'd need to download one more jar; but how many jars are they downloading already :-)

Anyways, If you all want to keep the uber-jar approach around, who am I to disagree.

> Now both of these things I don't think are necessarily related or blocking 1.0.

Well, yes and no: as you pointed out, Spark/Ivy does not recognize classifiers, so if we go with the big-jar as the default classifier, this means that the "slim jar" artifact will have to have a different artifact ID.

> Do you think it would be ok that if we deliver both the bundle and non-bundled

That's definitely better, yes. Maybe could we prepare for that and rename this artifact's ID to `polaris-spark-catalog-plugin-bundle` as suggested above?

dimas-b: Publishing both bundled and non-bundled jars SGTM. However, I believe the non-bundled jar will work well for automated downloads only if it declares all runtime dependencies. I hope it's already the case, just wanted to double check :)

gh-yzou: @dimas-b and @adutra i have an update here. 
```
I believe the non-bundled jar will work well for automated downloads only if it declares all runtime dependencies. I
```
Yes, that is actually the case, i think I made some mistake in my previous testing. The release today actually already works for both ways:
1) --packages org.apache.polaris:polaris-spark-3.5_2.12:0.11.0-beta-incubating-SNAPSHOT
2) or use --jars polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar

I have double confirmed this in couple of ways. So I think we can keep the current clarssifier to produce the bundle jar, i can update this PR as an improvement to remove my previous maven hack to reuse the shadow jar. However, I do not think that should be a blocker for 1.0. cc @flyrain

dimas-b: Given the latest test, I agree it's not a 1.0 blocker (but I did not personally test that :sweat_smile: )

flyrain: I've tested it locally with Spark 3.5.6 against apache maven repo. The package option works well for me. I don't think it's a blocker after this test. Here is the command I used for test, which picked up the Polaris nightly build:
```
bin/spark-shell \
--packages org.apache.polaris:polaris-spark-3.5_2.12:1.1.0-incubating-SNAPSHOT,io.delta:delta-spark_2.12:3.3.1 \
--repositories https://repository.apache.org/content/groups/snapshots/ \
--conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,io.delta.sql.DeltaSparkSessionExtension \
--conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
--conf spark.sql.catalog.polaris.warehouse=manual_spark \
--conf spark.sql.catalog.polaris.header.X-Iceberg-Access-Delegation=vended-credentials \
--conf spark.sql.catalog.polaris=org.apache.polaris.spark.SparkCatalog \
--conf spark.sql.catalog.polaris.uri=http://localhost:8181/api/catalog \
--conf spark.sql.catalog.polaris.credential='root:secret' \
--conf spark.sql.catalog.polaris.scope='PRINCIPAL_ROLE:ALL' \
--conf spark.sql.catalog.polaris.token-refresh-enabled=true
```

gh-yzou: @adutra since we have tested and clarified that we already support both approach today without this change. i updated the PR title as an improvement to reuse the shadowJar for maven publish.
The published jar name is called polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar, which i think is a pretty clear name. If we prefer name like polaris-spark-bundle, i will need to create a separate project to just pack the uber jar, and I can do it in a follow up PR.

plugins/spark/README.md
```diff
@@ -29,20 +29,20 @@ Right now, the plugin only provides support for Spark 3.5, Scala version 2.12 an
...
+- `./gradlew :polaris-spark-3.5_2.13:shadowJar` -- build jar for Spark 3.5 with Scala version 2.13.
 
 The result jar is located at plugins/spark/v3.5/build/<scala_version>/libs after the build.
 
 # Start Spark with Local Polaris Service using built Jar
 Once the jar is built, we can manually test it with Spark and a local Polaris service.
 
 The following command starts a Polaris server for local testing, it runs on localhost:8181 with default
-realm `POLARIS` and root credentials `root:secret`:
+realm `POLARIS` and root credentials `root:secret`:./gr
```
flyrain: typo?
```suggestion
realm `POLARIS` and root credentials `root:secret`:
```

site/content/in-dev/unreleased/polaris-spark-client.md
flyrain: I think we got a consensus here. I will remove the `1.0-blocker` label if there is no objection.

dimas-b: I think we're good for 1.0 without this PR. I've removed the blocker label.

plugins/spark/README.md
```diff
@@ -29,20 +29,20 @@ Right now, the plugin only provides support for Spark 3.5, Scala version 2.12 an
...
 # Build Plugin Jar
-A task createPolarisSparkJar is added to build a jar for the Polaris Spark plugin, the jar is named as:
+A shadowJar task is added to build a jar for the Polaris Spark plugin, the jar is named as:
 `polaris-spark-<sparkVersion>_<scalaVersion>-<polarisVersion>-bundle.jar`. For example:
 `polaris-spark-3.5_2.12-0.11.0-beta-incubating-SNAPSHOT-bundle.jar`.
 
-- `./gradlew :polaris-spark-3.5_2.12:createPolarisSparkJar` -- build jar for Spark 3.5 with Scala version 2.12.
-- `./gradlew :polaris-spark-3.5_2.13:createPolarisSparkJar` -- build jar for Spark 3.5 with Scala version 2.13.
+- `./gradlew :polaris-spark-3.5_2.12:shadowJar` -- build jar for Spark 3.5 with Scala version 2.12.
+- `./gradlew :polaris-spark-3.5_2.13:shadowJar` -- build jar for Spark 3.5 with Scala version 2.13.
```
dimas-b, line 37: nit: `:assemble` is higher level (less internal details visible to end users) and does not have much overhead compared to `shadowJar`.

site/content/in-dev/unreleased/polaris-spark-client.md
adutra: The current jar name is fine imho, thank you @gh-yzou! I appreciate your patience 😄

plugins/spark/README.md
gh-yzou: Yes a typo

plugins/spark/README.md
gh-yzou: I think it is still good to let user know that there is a specific task to just produce the jar in case they don't want to run assemble or build. I added one sentence below to mention that this task is also executed when run `gradlew assemble` or `gradlew build`. so user can choose whatever way they want.

dimas-b (commented):
Changes LGTM, but I believe the PR description is a bit off WRT actual changes now :thinking: WDYT?

gh-yzou:
@dimas-b sorry, i updated the title, but forgot to update the description, also updated the description

flyrain (approved):

PR #1356, opened 2025-04-11
Remove PolarisConfiguration.loadConfig 

eric-maynard:
`PolarisConfiguration.loadConfig` relies on `CallContext.getCurrentContext` which may not be apparent to callers and often will not work. If the CallContext ThreadLocal hasn't been appropriately set in the calling thread, this method will return invalid results. This PR proposed to remove the problematic method as part of a series of changes to reduce or remove the usage of ThreadLocal throughout Polaris.

PR #1858, opened 2025-06-11
Remove PolarisConfiguration.loadConfig (v2)

eric-maynard:
`PolarisConfiguration.loadConfig` relies on `CallContext.getCurrentContext` which may not be apparent to callers and often will not work. If the CallContext ThreadLocal hasn't been appropriately set in the calling thread, this method will return invalid results.

This PR proposed to remove the problematic method as part of a series of changes to reduce or remove the usage of ThreadLocal throughout Polaris. This is a rewrite of #1356 now that we have a more CDI-oriented `PolarisConfigurationStore` and an integration between `CallContext` and `PolarisCallContext`.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
@@ -248,9 +248,7 @@ public class CatalogEntity extends PolarisEntity {
     }
 
     public Builder setStorageConfigurationInfo(
-        CallContext callContext,
-        StorageConfigInfo storageConfigModel,
-        String defaultBaseLocation) {
+        CallContext callContext, StorageConfigInfo storageConfigModel, String defaultBaseLocation) {
       if (storageConfigModel != null) {
         PolarisStorageConfigurationInfo config;
         Set<String> allowedLocations = new HashSet<>(storageConfigModel.getAllowedLocations());
@@ -309,14 +307,14 @@ public class CatalogEntity extends PolarisEntity {
 
     /** Validate the number of allowed locations not exceeding the max value. */
     private void validateMaxAllowedLocations(
-        CallContext callContext,
-        Collection<String> allowedLocations) {
-      int maxAllowedLocations = callContext
-          .getPolarisCallContext()
-          .getConfigurationStore()
-          .getConfiguration(
-              callContext.getRealmContext(),
-              BehaviorChangeConfiguration.STORAGE_CONFIGURATION_MAX_LOCATIONS);
+        CallContext callContext, Collection<String> allowedLocations) {
+      int maxAllowedLocations =
+          callContext
+              .getPolarisCallContext()
+              .getConfigurationStore()
+              .getConfiguration(
+                  callContext.getRealmContext(),
+                  BehaviorChangeConfiguration.STORAGE_CONFIGURATION_MAX_LOCATIONS);
       if (maxAllowedLocations != -1 && allowedLocations.size() > maxAllowedLocations) {
         throw new IllegalArgumentException(
             String.format(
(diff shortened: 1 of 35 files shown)


polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -304,19 +301,6 @@ public Builder setStorageConfigurationInfo(
       return this;
     }
 
-    /** Validate the number of allowed locations not exceeding the max value. */
-    private void validateMaxAllowedLocations(Collection<String> allowedLocations) {
```
adutra, line 308: Don't we need this validation anymore? Or is it being done elsewhere?

polaris-core/src/test/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheTest.java
```diff
@@ -72,12 +72,16 @@ public StorageCredentialCacheTest() {
         new TreeMapTransactionalPersistenceImpl(store, Mockito.mock(), RANDOM_SECRETS);
     callCtx = new PolarisCallContext(() -> "testRealm", metaStore, diagServices);
     metaStoreManager = Mockito.mock(PolarisMetaStoreManager.class);
-    storageCredentialCache = new StorageCredentialCache();
+    storageCredentialCache = getStorageCredentialCache();
+  }
+
+  private StorageCredentialCache getStorageCredentialCache() {
```
adutra: nit: maybe `newStorageCredentialCache`?

polaris-core/src/test/java/org/apache/polaris/core/storage/BaseStorageIntegrationTest.java
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.polaris.core.storage;
+
+import org.apache.polaris.core.PolarisCallContext;
+import org.apache.polaris.core.PolarisDiagnostics;
+import org.apache.polaris.core.context.CallContext;
+import org.apache.polaris.core.persistence.BasePersistence;
+import org.mockito.Mockito;
+
+public abstract class BaseStorageIntegrationTest {
+  protected CallContext getCallContext() {
```
adutra: nit: maybe `newCallContext` or `createCallContext`?

polaris-core/src/test/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheTest.java
dimas-b: I agree "new" is more intuitive in this context.

quarkus/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
```diff
@@ -211,8 +211,9 @@ public Map<String, String> getConfigOverrides() {
   @Inject PolarisDiagnostics diagServices;
   @Inject PolarisEventListener polarisEventListener;
 
+  protected String realmName;
```
dimas-b: `protected` fields are hard to refactor. Non-final shared fields are also hard to reason about (in general). Would you mind adding a `protected` getter for a `private` field instead?

dimas-b (commented):
Thanks for reducing the use of thread locals, @eric-maynard ! :+1: 

This PR LGTM overall, just some minor comments.

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
eric-maynard: This is a temporary `BehaviorChangeConfiguration` (the default is no validation) so I thought this might be a good time to remove it. but if you think we'd better separate out that change, I can try to re-add the check. Unfortunately, there's not a great narrow waist for adding a `PolarisCallContext` to an entity.

adutra: Ah right, it's a `BehaviorChangeConfiguration`. OK to remove this validation then.

eric-maynard: I undid the change for now and piped a CallContext in, we can always remove it separately. Thanks for flagging this

dimas-b (approved):

PR #1862, opened 2025-06-11
Initial integration for hudi tables within Polaris 

rahil-c:
## Motivation

Issue: https://github.com/apache/polaris/issues/1896

The Polaris Spark client currently supports Iceberg and Delta table. This PR aims to add support for Apache Hudi tables as Generic Tables.

## Current behavior

Currently, the Polaris Spark client routes Iceberg table requests to Iceberg REST endpoints and Delta table requests to
Generic Table REST endpoints. This PR aims to allow Hudi to follow a similar support as for what was done for Delta in Polaris.

## Desired Behavior

Enable basic Hudi table operations through the Polaris Spark catalog by:
- Adding Hudi table detection and routing logic

## Changes Included

- Core Implementation: Added HudiHelper utility class and enhanced PolarisCatalogUtils with Hudi-specific table loading
logic
- Catalog Integration: Modified SparkCatalog to detect and route Hudi table operations appropriately
- Testing: Added unit tests for testingHudi integration
- Documentation: Updated README with Hudi development support

### Special note

Will follow up in another pr for integration and regression testing as they will need to consume the latest hudi point release artifact, once some changes in hudi land.

### DIFF
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -124,3 +124,8 @@ Following describes the current functionality limitations of the Polaris Spark c
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
 5) For other non-Iceberg tables like csv, it is not supported today.
+
+### Hudi Support
+Currently support for Hudi tables within the Polaris catalog is still under development. 
+The Hudi community has made a change to integrate with Polaris, and is planning on doing a minor release.
+Once the Hudi minor release is out, we will update the documentation on how users can use the Polaris Spark client with Hudi tables.
--- a/plugins/spark/README.md
+++ b/plugins/spark/README.md
@@ -127,5 +127,6 @@ Following describes the current functionality limitations of the Polaris Spark c
 
 ### Hudi Support
 Currently support for Hudi tables within the Polaris catalog is still under development. 
-The Hudi community has made a change to integrate with Polaris, and is planning on doing a minor release.
-Once the Hudi minor release is out, we will update the documentation on how users can use the Polaris Spark client with Hudi tables.
+The Hudi community has made a change to integrate with Polaris, and is planning on doing a Hudi 1.0.3 minor release.
+hudi-spark3.5-bundle_2.12:1.0.3 is required for hudi table support to work end to end, which is still under releasing.
+Once the Hudi 1.0.3 release is out, we will update the documentation on how users can use the Polaris Spark client with Hudi tables
--- a/plugins/spark/v3.5/integration/build.gradle.kts
+++ b/plugins/spark/v3.5/integration/build.gradle.kts
@@ -60,6 +60,7 @@ dependencies {
     exclude("org.apache.logging.log4j", "log4j-core")
     exclude("org.slf4j", "jul-to-slf4j")
   }
+
   // enforce the usage of log4j 2.24.3. This is for the log4j-api compatibility
   // of spark-sql dependency
   testRuntimeOnly("org.apache.logging.log4j:log4j-core:2.25.1")
--- a/plugins/spark/v3.5/integration/build.gradle.kts
+++ b/plugins/spark/v3.5/integration/build.gradle.kts
@@ -60,7 +60,6 @@ dependencies {
     exclude("org.apache.logging.log4j", "log4j-core")
     exclude("org.slf4j", "jul-to-slf4j")
   }
-
   // enforce the usage of log4j 2.24.3. This is for the log4j-api compatibility
   // of spark-sql dependency
   testRuntimeOnly("org.apache.logging.log4j:log4j-core:2.25.1")
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
@@ -71,7 +71,11 @@ public class PolarisSparkCatalog implements TableCatalog {
     try {
       GenericTable genericTable =
           this.polarisCatalog.loadGenericTable(Spark3Util.identifierToTableIdentifier(identifier));
-      return PolarisCatalogUtils.loadSparkTable(genericTable, identifier);
+      if (PolarisCatalogUtils.useHudi(genericTable.getFormat())) {
+        return PolarisCatalogUtils.loadV1SparkHudiTable(genericTable, identifier, name());
+      } else {
+        return PolarisCatalogUtils.loadV2SparkTable(genericTable);
+      }
     } catch (org.apache.iceberg.exceptions.NoSuchTableException e) {
       throw new NoSuchTableException(identifier);
     }
@@ -110,7 +114,11 @@ public class PolarisSparkCatalog implements TableCatalog {
               baseLocation,
               null,
               properties);
-      return PolarisCatalogUtils.loadSparkTable(genericTable, identifier);
+      if (PolarisCatalogUtils.useHudi(format)) {
+        return PolarisCatalogUtils.loadV1SparkHudiTable(genericTable, identifier, name());
+      } else {
+        return PolarisCatalogUtils.loadV2SparkTable(genericTable);
+      }
     } catch (AlreadyExistsException e) {
       throw new TableAlreadyExistsException(identifier);
     }
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
@@ -71,7 +71,7 @@ public class PolarisSparkCatalog implements TableCatalog {
     try {
       GenericTable genericTable =
           this.polarisCatalog.loadGenericTable(Spark3Util.identifierToTableIdentifier(identifier));
-      return PolarisCatalogUtils.loadSparkTable(genericTable);
+      return PolarisCatalogUtils.loadSparkTable(genericTable, identifier);
     } catch (org.apache.iceberg.exceptions.NoSuchTableException e) {
       throw new NoSuchTableException(identifier);
     }
@@ -110,7 +110,7 @@ public class PolarisSparkCatalog implements TableCatalog {
               baseLocation,
               null,
               properties);
-      return PolarisCatalogUtils.loadSparkTable(genericTable);
+      return PolarisCatalogUtils.loadSparkTable(genericTable, identifier);
     } catch (AlreadyExistsException e) {
       throw new TableAlreadyExistsException(identifier);
     }
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
@@ -72,7 +72,7 @@ public class PolarisSparkCatalog implements TableCatalog {
       GenericTable genericTable =
           this.polarisCatalog.loadGenericTable(Spark3Util.identifierToTableIdentifier(identifier));
       if (PolarisCatalogUtils.useHudi(genericTable.getFormat())) {
-        return PolarisCatalogUtils.loadV1SparkHudiTable(genericTable, identifier, name());
+        return PolarisCatalogUtils.loadV1SparkTable(genericTable, identifier, name());
       } else {
         return PolarisCatalogUtils.loadV2SparkTable(genericTable);
       }
@@ -115,7 +115,7 @@ public class PolarisSparkCatalog implements TableCatalog {
               null,
               properties);
       if (PolarisCatalogUtils.useHudi(format)) {
-        return PolarisCatalogUtils.loadV1SparkHudiTable(genericTable, identifier, name());
+        return PolarisCatalogUtils.loadV1SparkTable(genericTable, identifier, name());
       } else {
         return PolarisCatalogUtils.loadV2SparkTable(genericTable);
       }
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
@@ -30,6 +30,7 @@ import org.apache.iceberg.rest.auth.OAuth2Util;
 import org.apache.iceberg.spark.SupportsReplaceView;
 import org.apache.iceberg.util.PropertyUtil;
 import org.apache.polaris.spark.utils.DeltaHelper;
+import org.apache.polaris.spark.utils.HudiHelper;
 import org.apache.polaris.spark.utils.PolarisCatalogUtils;
 import org.apache.spark.sql.catalyst.analysis.NamespaceAlreadyExistsException;
 import org.apache.spark.sql.catalyst.analysis.NoSuchNamespaceException;
@@ -72,6 +73,7 @@ public class SparkCatalog
   @VisibleForTesting protected org.apache.iceberg.spark.SparkCatalog icebergsSparkCatalog = null;
   @VisibleForTesting protected PolarisSparkCatalog polarisSparkCatalog = null;
   @VisibleForTesting protected DeltaHelper deltaHelper = null;
+  @VisibleForTesting protected HudiHelper hudiHelper = null;
 
   @Override
   public String name() {
@@ -133,6 +135,7 @@ public class SparkCatalog
     this.catalogName = name;
     initRESTCatalog(name, options);
     this.deltaHelper = new DeltaHelper(options);
+    this.hudiHelper = new HudiHelper(options);
   }
 
   @Override
@@ -156,15 +159,19 @@ public class SparkCatalog
         throw new UnsupportedOperationException(
             "Create table without location key is not supported by Polaris. Please provide location or path on table creation.");
       }
-
       if (PolarisCatalogUtils.useDelta(provider)) {
         // For delta table, we load the delta catalog to help dealing with the
         // delta log creation.
         TableCatalog deltaCatalog = deltaHelper.loadDeltaCatalog(this.polarisSparkCatalog);
         return deltaCatalog.createTable(ident, schema, transforms, properties);
-      } else {
-        return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
       }
+      if (PolarisCatalogUtils.useHudi(provider)) {
+        // For creating the hudi table, we load HoodieCatalog
+        // to create the .hoodie folder in cloud storage
+        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
+        return hudiCatalog.createTable(ident, schema, transforms, properties);
+      }
+      return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
     }
   }
 
@@ -182,7 +189,11 @@ public class SparkCatalog
         //     using ALTER TABLE ...SET LOCATION, and ALTER TABLE ... SET FILEFORMAT.
         TableCatalog deltaCatalog = deltaHelper.loadDeltaCatalog(this.polarisSparkCatalog);
         return deltaCatalog.alterTable(ident, changes);
+      } else if (PolarisCatalogUtils.useHudi(provider)) {
+        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
+        return hudiCatalog.alterTable(ident, changes);
       }
+
       return this.polarisSparkCatalog.alterTable(ident);
     }
   }
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
@@ -192,9 +192,9 @@ public class SparkCatalog
       } else if (PolarisCatalogUtils.useHudi(provider)) {
         TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
         return hudiCatalog.alterTable(ident, changes);
+      } else {
+        return this.polarisSparkCatalog.alterTable(ident);
       }
-
-      return this.polarisSparkCatalog.alterTable(ident);
     }
   }
 
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
@@ -164,14 +164,14 @@ public class SparkCatalog
         // delta log creation.
         TableCatalog deltaCatalog = deltaHelper.loadDeltaCatalog(this.polarisSparkCatalog);
         return deltaCatalog.createTable(ident, schema, transforms, properties);
-      }
-      if (PolarisCatalogUtils.useHudi(provider)) {
+      } else if (PolarisCatalogUtils.useHudi(provider)) {
         // For creating the hudi table, we load HoodieCatalog
         // to create the .hoodie folder in cloud storage
         TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
         return hudiCatalog.createTable(ident, schema, transforms, properties);
+      } else {
+        return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
       }
-      return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
     }
   }
 
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiHelper.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiHelper.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.spark.utils;
+
+import org.apache.iceberg.common.DynConstructors;
+import org.apache.polaris.spark.PolarisSparkCatalog;
+import org.apache.spark.sql.connector.catalog.DelegatingCatalogExtension;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+
+public class HudiHelper {
+  public static final String HUDI_CATALOG_IMPL_KEY = "hudi-catalog-impl";
+  private static final String DEFAULT_HUDI_CATALOG_CLASS =
+      "org.apache.spark.sql.hudi.catalog.HoodieCatalog";
+
+  private TableCatalog hudiCatalog = null;
+  private String hudiCatalogImpl = DEFAULT_HUDI_CATALOG_CLASS;
+
+  public HudiHelper(CaseInsensitiveStringMap options) {
+    if (options.get(HUDI_CATALOG_IMPL_KEY) != null) {
+      this.hudiCatalogImpl = options.get(HUDI_CATALOG_IMPL_KEY);
+    }
+  }
+
+  public TableCatalog loadHudiCatalog(PolarisSparkCatalog polarisSparkCatalog) {
+    if (this.hudiCatalog != null) {
+      return this.hudiCatalog;
+    }
+
+    DynConstructors.Ctor<TableCatalog> ctor;
+    try {
+      ctor = DynConstructors.builder(TableCatalog.class).impl(hudiCatalogImpl).buildChecked();
+    } catch (NoSuchMethodException e) {
+      throw new IllegalArgumentException(
+          String.format("Cannot initialize Hudi Catalog %s: %s", hudiCatalogImpl, e.getMessage()),
+          e);
+    }
+
+    try {
+      this.hudiCatalog = ctor.newInstance();
+    } catch (ClassCastException e) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Cannot initialize Hudi Catalog, %s does not implement Table Catalog.",
+              hudiCatalogImpl),
+          e);
+    }
+
+    // set the polaris spark catalog as the delegate catalog of hudi catalog
+    // will be used in HoodieCatalog's loadTable
+    ((DelegatingCatalogExtension) this.hudiCatalog).setDelegateCatalog(polarisSparkCatalog);
+    return this.hudiCatalog;
+  }
+}
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
@@ -29,14 +29,23 @@ import org.apache.iceberg.rest.auth.OAuth2Util;
 import org.apache.iceberg.spark.SparkCatalog;
 import org.apache.polaris.spark.rest.GenericTable;
 import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.catalyst.TableIdentifier;
+import org.apache.spark.sql.catalyst.catalog.CatalogTable;
+import org.apache.spark.sql.catalyst.catalog.CatalogTableType;
+import org.apache.spark.sql.connector.catalog.Identifier;
 import org.apache.spark.sql.connector.catalog.Table;
 import org.apache.spark.sql.connector.catalog.TableCatalog;
 import org.apache.spark.sql.connector.catalog.TableProvider;
 import org.apache.spark.sql.execution.datasources.DataSource;
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Utils;
 import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.Option;
 
 public class PolarisCatalogUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(PolarisCatalogUtils.class);
+
   public static final String TABLE_PROVIDER_KEY = "provider";
   public static final String TABLE_PATH_KEY = "path";
 
@@ -48,6 +57,10 @@ public class PolarisCatalogUtils {
   /** Check whether the table provider is delta. */
   public static boolean useDelta(String provider) {
     return "delta".equalsIgnoreCase(provider);
+  }
+
+  public static boolean useHudi(String provider) {
+    return "hudi".equalsIgnoreCase(provider);
   }
 
   /**
@@ -64,9 +77,14 @@ public class PolarisCatalogUtils {
    * Load spark table using DataSourceV2.
    *
    * @return V2Table if DataSourceV2 is available for the table format. For delta table, it returns
-   *     DeltaTableV2.
+   *     DeltaTableV2. For hudi tables we will return a Spark V1Table.
    */
-  public static Table loadSparkTable(GenericTable genericTable) {
+  public static Table loadSparkTable(GenericTable genericTable, Identifier identifier) {
+    if (genericTable.getFormat().equalsIgnoreCase("hudi")) {
+      // hudi does not implement Spark V2 table provider interface
+      // therefore will need to return a V1Table
+      return loadV1SparkHudiTable(genericTable, identifier);
+    }
     SparkSession sparkSession = SparkSession.active();
     TableProvider provider =
         DataSource.lookupDataSourceV2(genericTable.getFormat(), sparkSession.sessionState().conf())
@@ -89,6 +107,85 @@ public class PolarisCatalogUtils {
     }
     return DataSourceV2Utils.getTableFromProvider(
         provider, new CaseInsensitiveStringMap(tableProperties), scala.Option.empty());
+  }
+
+  /**
+   * Extract catalog name from Spark session configuration. Looks for configuration like:
+   * spark.sql.catalog.<CATALOG_NAME>=org.apache.polaris.spark.SparkCatalog
+   */
+  private static String getCatalogName() {
+    SparkSession spark = SparkSession.active();
+    String catalogPrefix = "spark.sql.catalog.";
+    String polarisSparkCatalog = "org.apache.polaris.spark.SparkCatalog";
+
+    scala.collection.Iterator<scala.Tuple2<String, String>> configIterator =
+        spark.conf().getAll().iterator();
+    while (configIterator.hasNext()) {
+      scala.Tuple2<String, String> config = configIterator.next();
+      String key = config._1();
+      String value = config._2();
+
+      if (key.startsWith(catalogPrefix) && polarisSparkCatalog.equals(value)) {
+        return key.substring(catalogPrefix.length());
+      }
+    }
+
+    throw new IllegalStateException(
+        "Could not obtain Polaris catalog identifier."
+            + "Expected following configuration to be set in session: spark.sql.catalog.<CATALOG_NAME>=org.apache.polaris.spark.SparkCatalog");
+  }
+
+  public static Table loadV1SparkHudiTable(GenericTable genericTable, Identifier identifier) {
+    Map<String, String> tableProperties = Maps.newHashMap();
+    tableProperties.putAll(genericTable.getProperties());
+    tableProperties.put(
+        TABLE_PATH_KEY, genericTable.getProperties().get(TableCatalog.PROP_LOCATION));
+
+    // Need full identifier in order to construct CatalogTable correctly for Hudi
+    String namespacePath = String.join(".", identifier.namespace());
+    TableIdentifier tableIdentifier =
+        new TableIdentifier(
+            identifier.name(), Option.apply(namespacePath), Option.apply(getCatalogName()));
+
+    scala.collection.immutable.Map<String, String> scalaOptions =
+        (scala.collection.immutable.Map<String, String>)
+            scala.collection.immutable.Map$.MODULE$.apply(
+                scala.collection.JavaConverters.mapAsScalaMap(tableProperties).toSeq());
+
+    org.apache.spark.sql.catalyst.catalog.CatalogStorageFormat storage =
+        DataSource.buildStorageFormatFromOptions(scalaOptions);
+
+    // Currently Polaris generic table does not contain any schema information, partition columns,
+    // stats, etc
+    // for now we will just use fill the parameters we have from catalog, and let underlying client
+    // resolve the rest within its catalog implementation
+    org.apache.spark.sql.types.StructType emptySchema = new org.apache.spark.sql.types.StructType();
+    scala.collection.immutable.Seq<String> emptyStringSeq =
+        scala.collection.JavaConverters.asScalaBuffer(new java.util.ArrayList<String>()).toList();
+    CatalogTable catalogTable =
+        new CatalogTable(
+            tableIdentifier,
+            CatalogTableType.EXTERNAL(),
+            storage,
+            emptySchema,
+            Option.apply(genericTable.getProperties().get("provider")),
+            emptyStringSeq,
+            scala.Option.empty(),
+            genericTable.getProperties().get("owner"),
+            System.currentTimeMillis(),
+            -1L,
+            "",
+            scalaOptions,
+            scala.Option.empty(),
+            scala.Option.empty(),
+            scala.Option.empty(),
+            emptyStringSeq,
+            false,
+            true,
+            scala.collection.immutable.Map$.MODULE$.empty(),
+            scala.Option.empty());
+
+    return new org.apache.spark.sql.connector.catalog.V1Table(catalogTable);
   }
 
   /**
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
@@ -74,21 +74,11 @@ public class PolarisCatalogUtils {
   }
 
   /**
-   * Load spark table using DataSourceV2.
-   *
-   * @return V2Table if DataSourceV2 is available for the table format. For delta table, it returns
-   *     DeltaTableV2. For hudi tables we will return a Spark V1Table.
+   * Normalize table properties for loading Spark tables by ensuring the TABLE_PATH_KEY is properly
+   * set. DataSourceV2 requires the path property on table loading.
    */
-  public static Table loadSparkTable(GenericTable genericTable, Identifier identifier) {
-    if (genericTable.getFormat().equalsIgnoreCase("hudi")) {
-      // hudi does not implement Spark V2 table provider interface
-      // therefore will need to return a V1Table
-      return loadV1SparkHudiTable(genericTable, identifier);
-    }
-    SparkSession sparkSession = SparkSession.active();
-    TableProvider provider =
-        DataSource.lookupDataSourceV2(genericTable.getFormat(), sparkSession.sessionState().conf())
-            .get();
+  private static Map<String, String> normalizeTablePropertiesForLoadSparkTable(
+      GenericTable genericTable) {
     Map<String, String> properties = genericTable.getProperties();
     boolean hasLocationClause = properties.get(TableCatalog.PROP_LOCATION) != null;
     boolean hasPathClause = properties.get(TABLE_PATH_KEY) != null;
@@ -105,47 +95,35 @@ public class PolarisCatalogUtils {
         tableProperties.put(TABLE_PATH_KEY, properties.get(TableCatalog.PROP_LOCATION));
       }
     }
+    return tableProperties;
+  }
+
+  /**
+   * Load spark table using DataSourceV2.
+   *
+   * @return V2Table if DataSourceV2 is available for the table format. For delta table, it returns
+   *     DeltaTableV2. For hudi tables we will return a Spark V1Table.
+   */
+  public static Table loadV2SparkTable(GenericTable genericTable) {
+    SparkSession sparkSession = SparkSession.active();
+    TableProvider provider =
+        DataSource.lookupDataSourceV2(genericTable.getFormat(), sparkSession.sessionState().conf())
+            .get();
+    Map<String, String> tableProperties = normalizeTablePropertiesForLoadSparkTable(genericTable);
     return DataSourceV2Utils.getTableFromProvider(
         provider, new CaseInsensitiveStringMap(tableProperties), scala.Option.empty());
   }
 
-  /**
-   * Extract catalog name from Spark session configuration. Looks for configuration like:
-   * spark.sql.catalog.<CATALOG_NAME>=org.apache.polaris.spark.SparkCatalog
-   */
-  private static String getCatalogName() {
-    SparkSession spark = SparkSession.active();
-    String catalogPrefix = "spark.sql.catalog.";
-    String polarisSparkCatalog = "org.apache.polaris.spark.SparkCatalog";
-
-    scala.collection.Iterator<scala.Tuple2<String, String>> configIterator =
-        spark.conf().getAll().iterator();
-    while (configIterator.hasNext()) {
-      scala.Tuple2<String, String> config = configIterator.next();
-      String key = config._1();
-      String value = config._2();
-
-      if (key.startsWith(catalogPrefix) && polarisSparkCatalog.equals(value)) {
-        return key.substring(catalogPrefix.length());
-      }
-    }
-
-    throw new IllegalStateException(
-        "Could not obtain Polaris catalog identifier."
-            + "Expected following configuration to be set in session: spark.sql.catalog.<CATALOG_NAME>=org.apache.polaris.spark.SparkCatalog");
-  }
-
-  public static Table loadV1SparkHudiTable(GenericTable genericTable, Identifier identifier) {
-    Map<String, String> tableProperties = Maps.newHashMap();
-    tableProperties.putAll(genericTable.getProperties());
-    tableProperties.put(
-        TABLE_PATH_KEY, genericTable.getProperties().get(TableCatalog.PROP_LOCATION));
+  /** Return a Spark V1Table for Hudi tables. */
+  public static Table loadV1SparkHudiTable(
+      GenericTable genericTable, Identifier identifier, String catalogName) {
+    Map<String, String> tableProperties = normalizeTablePropertiesForLoadSparkTable(genericTable);
 
     // Need full identifier in order to construct CatalogTable correctly for Hudi
     String namespacePath = String.join(".", identifier.namespace());
     TableIdentifier tableIdentifier =
         new TableIdentifier(
-            identifier.name(), Option.apply(namespacePath), Option.apply(getCatalogName()));
+            identifier.name(), Option.apply(namespacePath), Option.apply(catalogName));
 
     scala.collection.immutable.Map<String, String> scalaOptions =
         (scala.collection.immutable.Map<String, String>)
@@ -168,7 +146,7 @@ public class PolarisCatalogUtils {
             CatalogTableType.EXTERNAL(),
             storage,
             emptySchema,
-            Option.apply(genericTable.getProperties().get("provider")),
+            Option.apply(genericTable.getFormat()),
             emptyStringSeq,
             scala.Option.empty(),
             genericTable.getProperties().get("owner"),
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
@@ -102,7 +102,7 @@ public class PolarisCatalogUtils {
    * Load spark table using DataSourceV2.
    *
    * @return V2Table if DataSourceV2 is available for the table format. For delta table, it returns
-   *     DeltaTableV2. For hudi tables we will return a Spark V1Table.
+   *     DeltaTableV2.
    */
   public static Table loadV2SparkTable(GenericTable genericTable) {
     SparkSession sparkSession = SparkSession.active();
@@ -114,12 +114,15 @@ public class PolarisCatalogUtils {
         provider, new CaseInsensitiveStringMap(tableProperties), scala.Option.empty());
   }
 
-  /** Return a Spark V1Table for Hudi tables. */
-  public static Table loadV1SparkHudiTable(
+  /**
+   * Return a Spark V1Table for formats that do not use DataSourceV2. Currently, this is being used
+   * for Hudi tables
+   */
+  public static Table loadV1SparkTable(
       GenericTable genericTable, Identifier identifier, String catalogName) {
     Map<String, String> tableProperties = normalizeTablePropertiesForLoadSparkTable(genericTable);
 
-    // Need full identifier in order to construct CatalogTable correctly for Hudi
+    // Need full identifier in order to construct CatalogTable
     String namespacePath = String.join(".", identifier.namespace());
     TableIdentifier tableIdentifier =
         new TableIdentifier(
--- a/plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
+++ b/plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
@@ -657,7 +657,7 @@ public class SparkCatalogTest {
           mockedStaticUtils
               .when(
                   () ->
-                      PolarisCatalogUtils.loadV1SparkHudiTable(
+                      PolarisCatalogUtils.loadV1SparkTable(
                           Mockito.any(), Mockito.any(), Mockito.any()))
               .thenReturn(table);
         } else {
--- a/plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
+++ b/plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
@@ -35,6 +35,7 @@ import org.apache.iceberg.spark.actions.DeleteReachableFilesSparkAction;
 import org.apache.iceberg.spark.actions.SparkActions;
 import org.apache.iceberg.spark.source.SparkTable;
 import org.apache.polaris.spark.utils.DeltaHelper;
+import org.apache.polaris.spark.utils.HudiHelper;
 import org.apache.polaris.spark.utils.PolarisCatalogUtils;
 import org.apache.spark.SparkContext;
 import org.apache.spark.sql.SparkSession;
@@ -58,6 +59,7 @@ import org.apache.spark.sql.internal.SQLConf;
 import org.apache.spark.sql.internal.SessionState;
 import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -67,6 +69,9 @@ import org.mockito.Mockito;
 import scala.Option;
 
 public class SparkCatalogTest {
+  private static MockedStatic<SparkSession> mockedStaticSparkSession;
+  private static SparkSession mockedSession;
+
   private static class InMemoryIcebergSparkCatalog extends org.apache.iceberg.spark.SparkCatalog {
     private PolarisInMemoryCatalog inMemoryCatalog = null;
 
@@ -104,6 +109,7 @@ public class SparkCatalogTest {
       this.polarisSparkCatalog.initialize(name, options);
 
       this.deltaHelper = new DeltaHelper(options);
+      this.hudiHelper = new HudiHelper(options);
     }
   }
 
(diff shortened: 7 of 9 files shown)


dimas-b:
Thanks for you contribution, @rahil-c ! Would you mind opening a discussion for this feature on dev@polaris.apache.org?

rahil-c:
Thanks @dimas-b will do so! Have raised a email on dev list here: https://lists.apache.org/thread/66d39oqkc412kk262gy80bm723r9xmpm

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -156,15 +159,23 @@ public Table createTable(
...
+        // First make a call via polaris's spark catalog
+        // to ensure an entity is created within the catalog and is authorized
+        polarisSparkCatalog.createTable(ident, schema, transforms, properties);
+
+        // Then for actually creating the hudi table, we load HoodieCatalog
+        // to create the .hoodie folder in cloud storage
+        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
+        return hudiCatalog.createTable(ident, schema, transforms, properties);
+      }
+      return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
```
eric-maynard: Let's keep the `if { } else if {} else { }` structure

rahil-c:
cc @flyrain @gh-yzou @singhpk234

plugins/spark/v3.5/integration/build.gradle.kts
```diff
@@ -60,12 +60,51 @@ dependencies {
     exclude("org.apache.logging.log4j", "log4j-core")
     exclude("org.slf4j", "jul-to-slf4j")
   }
+
+  // Add spark-hive for Hudi integration - provides HiveExternalCatalog that Hudi needs
+  testImplementation("org.apache.spark:spark-hive_${scalaVersion}:${spark35Version}") {
+    // exclude log4j dependencies to match spark-sql exclusions
```
gh-yzou: can we remove the spark_sql dependency above?

plugins/spark/v3.5/integration/build.gradle.kts
```diff
@@ -60,12 +60,51 @@ dependencies {
...
+    exclude("org.apache.logging.log4j", "log4j-slf4j-impl")
+  }
   // enforce the usage of log4j 2.24.3. This is for the log4j-api compatibility
   // of spark-sql dependency
   testRuntimeOnly("org.apache.logging.log4j:log4j-core:2.24.3")
   testRuntimeOnly("org.apache.logging.log4j:log4j-slf4j2-impl:2.24.3")
 
   testImplementation("io.delta:delta-spark_${scalaVersion}:3.3.1")
+  testImplementation("org.apache.hudi:hudi-spark3.5-bundle_${scalaVersion}:0.15.0") {
+    // exclude log4j dependencies to match spark-sql exclusions and prevent version conflicts
```
gh-yzou: does the bundle already contain all spark dependency needed? if that is the case, we shouldn't need the spark_hive dependency anymore, right?

plugins/spark/v3.5/integration/src/intTest/resources/logback.xml
```diff
@@ -32,6 +32,9 @@ out the configuration if you would like ot see all spark debug log during the ru
     </encoder>
   </appender>
 
+  <!-- Hudi-specific loggers for test -->
+  <logger name="org.apache.hudi" level="INFO"/>
```
gh-yzou: does hudi output a log of logs? too much logging was causing problems to the test efficiency, so we only turned on the error log here. How long does the integration test take now?

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -46,6 +46,47 @@ dependencies {
   // TODO: extract a polaris-rest module as a thin layer for
   //  client to depends on.
   implementation(project(":polaris-core")) { isTransitive = false }
+  implementation(project(":polaris-api-iceberg-service")) {
```
gh-yzou: why does hudi need those dependency ?

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -46,6 +46,47 @@ dependencies {
...
+    exclude("software.amazon.awssdk", "*")
+    exclude("com.google.cloud", "*")
+    exclude("io.airlift", "*")
+    exclude("io.smallrye", "*")
+    exclude("io.smallrye.common", "*")
+    exclude("io.swagger", "*")
+    exclude("org.apache.commons", "*")
+  }
+
+  implementation("org.apache.iceberg:iceberg-core:${icebergVersion}")
```
gh-yzou: we do not intend to depends on the iceberg core, it causes spark compatible issues

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -46,6 +46,47 @@ dependencies {
...
+  compileOnly("org.apache.hudi:hudi-spark3.5-bundle_${scalaVersion}:0.15.0")
```
gh-yzou: i don't think we need to depends on hudi bundle here, similar as how we handles delta

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -156,15 +160,23 @@ public Table createTable(
...
-        return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
       }
+      if (PolarisCatalogUtils.useHudi(provider)) {
+        // First make a call via polaris's spark catalog
+        // to ensure an entity is created within the catalog and is authorized
+        polarisSparkCatalog.createTable(ident, schema, transforms, properties);
+
+        // Then for actually creating the hudi table, we load HoodieCatalog
+        // to create the .hoodie folder in cloud storage
+        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
```
gh-yzou, line 168: it might be better to first make sure the hudiCatalog.createTable can be done successfully first, and then create the catalog with the remote service

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -270,18 +286,24 @@ public Map<String, String> loadNamespaceMetadata(String[] namespace)
   public void createNamespace(String[] namespace, Map<String, String> metadata)
       throws NamespaceAlreadyExistsException {
     this.icebergsSparkCatalog.createNamespace(namespace, metadata);
+    HudiCatalogUtils.createNamespace(namespace, metadata);
```
gh-yzou: What is the difference between delta and hudi? I assume the hudi catalog will be used as spark session catalog and directly called for namespace operations

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * <p>Hudi table loading requires namespace validation through the session catalog, but only the
+ * Polaris catalog contains the actual namespace metadata. This class provides methods to
+ * synchronize namespace operations to maintain consistency between catalogs.
+ */
+public class HudiCatalogUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(HudiCatalogUtils.class);
+
+  /**
+   * Synchronizes namespace creation to session catalog when Hudi extension is enabled. This ensures
```
gh-yzou: @rahil-c this part of change is actually out of my expectation, I might need some more time to understand the motivation for this part.  Can you update the readme for the project with instruction about how to use hudi? I want to checkout the code and try it out. Thanks!

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -87,6 +119,32 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+    CatalogTable catalogTable = null;
+    try {
+      catalogTable = sparkSession.sessionState().catalog().getTableMetadata(tableIdentifier);
+    } catch (NoSuchDatabaseException e) {
+      throw new RuntimeException(
+          "No database found for the given tableIdentifier:" + tableIdentifier, e);
+    } catch (NoSuchTableException e) {
+      LOG.debug("No table currently exists, as an initial create table");
+    }
+    return new HoodieInternalV2Table(
```
gh-yzou: does hudi catalog provides any load table functionality, which also reads the hudi logs? if yes, we can load the hudi catalog here and call the functions, it would be better if we could avoid any table format specific dependency in the client

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkHudiIT.java
```diff
@@ -0,0 +1,174 @@
...
+
+  @Test
+  public void testBasicTableOperations() {
+    // create a regular hudi table
+    String huditb1 = "huditb1";
+    sql(
+        "CREATE TABLE %s (id INT, name STRING) USING HUDI LOCATION '%s'",
+        huditb1, getTableLocation(huditb1));
+    sql("INSERT INTO %s VALUES (1, 'anna'), (2, 'bob')", huditb1);
+    List<Object[]> results = sql("SELECT id,name FROM %s WHERE id > 1 ORDER BY id DESC", huditb1);
```
eric-maynard: nit: Some of these queries could be cleaned up

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkHudiIT.java
```diff
@@ -0,0 +1,174 @@
...
+
+    // CTAS fails
+    assertThatThrownBy(
+            () ->
+                sql(
+                    "CREATE TABLE %s USING HUDI LOCATION '%s' AS SELECT 1 AS id",
+                    huditb, getTableLocation(huditb)))
+        .isInstanceOf(IllegalArgumentException.class);
+  }
+}
```
eric-maynard: Is there a test here that actually checks the tables are written as Hudi?

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkHudiIT.java
```diff
@@ -0,0 +1,174 @@
...
+        .isInstanceOf(UnsupportedOperationException.class);
+
+    sql("DROP TABLE %s", huditb);
+  }
+
+  @Test
+  public void testUnsupportedTableCreateOperations() {
+    String huditb = getTableNameWithRandomSuffix();
+    // create hudi table with no location
+    assertThatThrownBy(() -> sql("CREATE TABLE %s (id INT, name STRING) USING HUDI", huditb))
```
eric-maynard: Why should this fail?

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -46,6 +46,47 @@ dependencies {
...
+    exclude("com.azure", "*")
+    exclude("software.amazon.awssdk", "*")
+    exclude("com.google.cloud", "*")
+    exclude("io.airlift", "*")
+    exclude("io.smallrye", "*")
+    exclude("io.smallrye.common", "*")
+    exclude("io.swagger", "*")
+    exclude("org.apache.commons", "*")
+  }
+  implementation(project(":polaris-api-catalog-service")) {
```
eric-maynard: ditto, why is this new dep needed for the implementation?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -182,7 +194,11 @@ public Table alterTable(Identifier ident, TableChange... changes) throws NoSuchT
         //     using ALTER TABLE ...SET LOCATION, and ALTER TABLE ... SET FILEFORMAT.
         TableCatalog deltaCatalog = deltaHelper.loadDeltaCatalog(this.polarisSparkCatalog);
         return deltaCatalog.alterTable(ident, changes);
+      } else if (PolarisCatalogUtils.useHudi(provider)) {
+        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
```
eric-maynard, line 191: The same comment above applies here, right? Let's restructure to make that more clear.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -270,18 +286,24 @@ public Map<String, String> loadNamespaceMetadata(String[] namespace)
...
     this.icebergsSparkCatalog.alterNamespace(namespace, changes);
+    HudiCatalogUtils.alterNamespace(namespace, changes);
   }
 
   @Override
   public boolean dropNamespace(String[] namespace, boolean cascade)
       throws NoSuchNamespaceException {
-    return this.icebergsSparkCatalog.dropNamespace(namespace, cascade);
+    boolean result = this.icebergsSparkCatalog.dropNamespace(namespace, cascade);
+    if (result) {
```
eric-maynard: Why would `icebergsSparkCatalog.dropNamespace` returning `true` mean this is a Hudi namespace?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
eric-maynard: Why are we _always_ calling this?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
```diff
@@ -0,0 +1,220 @@
...
+   * session catalog metadata stays consistent with Polaris catalog for comprehensive Hudi
```
eric-maynard: What does `comprehensive` actually mean?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    // Sync namespace with filtered metadata to session catalog only when Hudi is enabled
+    // This is needed because Hudi table loading uses the spark session catalog
+    // to validate namespace existence and access metadata properties.
+    // Reserved properties (owner, location, comment) are automatically filtered out.
+    try {
+      SparkSession spark = SparkSession.active();
+      String ns = String.join(".", namespace);
+
+      // Build CREATE NAMESPACE SQL with metadata properties (filtered for reserved properties)
+      String createSql = String.format("CREATE NAMESPACE IF NOT EXISTS spark_catalog.%s", ns);
```
eric-maynard: This doesn't look correct; why do we need to keep the spark catalog in sync with the actual Polaris catalog?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
```diff
@@ -0,0 +1,220 @@
...
+   *
+   * @param namespace The namespace that failed to sync
+   * @param operation The operation being performed (for logging)
+   * @param e The exception that occurred
+   * @param throwOnError Whether to throw RuntimeException for critical errors (vs log warning)
+   * @throws RuntimeException For critical errors when throwOnError is true
+   */
+  private static void handleNamespaceSyncError(
+      String[] namespace, String operation, Exception e, boolean throwOnError) {
+    String errorMsg = e.getMessage();
```
eric-maynard: Necessity aside, this error handling looks quite brittle. You should probably be traversing the cause chain and looking at exception types rather than just doing string-matching on the message.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -64,9 +92,13 @@ public static boolean isTableWithSparkManagedLocation(Map<String, String> proper
    * Load spark table using DataSourceV2.
    *
    * @return V2Table if DataSourceV2 is available for the table format. For delta table, it returns
-   *     DeltaTableV2.
+   *     DeltaTableV2. For hudi it should return HoodieInternalV2Table.
    */
-  public static Table loadSparkTable(GenericTable genericTable) {
+  public static Table loadSparkTable(GenericTable genericTable, Identifier identifier) {
+    if (genericTable.getFormat().equalsIgnoreCase("hudi")) {
+      // hudi does not implement table provider interface, so will need to catch it
```
eric-maynard: Catch it? 

It's not clear to me why this is different from Delta?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
```diff
@@ -68,7 +68,7 @@ public Table loadTable(Identifier identifier) throws NoSuchTableException {
     try {
       GenericTable genericTable =
           this.polarisCatalog.loadGenericTable(Spark3Util.identifierToTableIdentifier(identifier));
-      return PolarisCatalogUtils.loadSparkTable(genericTable);
+      return PolarisCatalogUtils.loadSparkTable(genericTable, identifier);
```
eric-maynard: It's not clear to me why the interface had to change, is this related to the namespace thing?

plugins/spark/v3.5/integration/build.gradle.kts
rahil-c: We will need this dependency as a test dependency, as when hudi insert is invoked it looks for this `HiveExternalCatalog`, when removing this dependency as suspected this fails for class not found exception.

```
org/apache/spark/sql/hive/HiveExternalCatalog$
java.lang.NoClassDefFoundError: org/apache/spark/sql/hive/HiveExternalCatalog$
	at org.apache.spark.sql.hudi.ProvidesHoodieConfig.buildHiveSyncConfig(ProvidesHoodieConfig.scala:496)
	at org.apache.spark.sql.hudi.ProvidesHoodieConfig.buildHiveSyncConfig$(ProvidesHoodieConfig.scala:467)
	at org.apache.spark.sql.hudi.analysis.HoodieSpark35DataSourceV2ToV1Fallback.buildHiveSyncConfig(HoodieSpark35Analysis.scala:39)
	at org.apache.spark.sql.hudi.ProvidesHoodieConfig.buildHoodieConfig(ProvidesHoodieConfig.scala:63)
	at org.apache.spark.sql.hudi.ProvidesHoodieConfig.buildHoodieConfig$(ProvidesHoodieConfig.scala:55)
	at org.apache.spark.sql.hudi.analysis.HoodieSpark35DataSourceV2ToV1Fallback.buildHoodieConfig(HoodieSpark35Analysis.scala:39)
	at org.apache.spark.sql.hudi.analysis.HoodieSpark35DataSourceV2ToV1Fallback.org$apache$spark$sql$hudi$analysis$HoodieSpark35DataSourceV2ToV1Fallback$$convertToV1(HoodieSpark35Analysis.scala:62)
	at org.apache.spark.sql.hudi.analysis.HoodieSpark35DataSourceV2ToV1Fallback.apply(HoodieSpark35Analysis.scala:50)
	at org.apache.spark.sql.hudi.analysis.HoodieSpark35DataSourceV2ToV1Fallback.apply(HoodieSpark35Analysis.scala:39)
	at org.apache.spark.sql.catalyst.rules.RuleExecutor.$anonfun$execute$2(RuleExecutor.scala:222)
```

Have also explained further here in related comment: https://github.com/apache/polaris/pull/1862#discussion_r2180478143

RussellSpitzer: Don't we only need it as a test-runtime?

plugins/spark/v3.5/integration/build.gradle.kts
RussellSpitzer: I think this and the above dependency can both be test runtime? Or are we using Delta and Hudi specific code in the test code itself?

rahil-c: Similar as comment above: https://github.com/apache/polaris/pull/1862#discussion_r2180424419

When running the following command
` jar tf hudi-spark3.5-bundle_2.12-0.15.0.jar | grep -i "org/apache/spark/sql/hive"`

I do not see the `HiveExternalCatalog` provided by the `hudi-spark-bundle`.
```
org/apache/spark/sql/hive/
org/apache/spark/sql/hive/HiveClientUtils.class
org/apache/spark/sql/hive/HiveClientUtils$.class
```

Based on my understanding of the `hudi-spark-bundle` it aims to provide the core hudi dependencies needed for spark-hudi integration to work but expects certain spark dependencies to be on the engine class path.
For example when checking the oss spark engine `jars` folder by running these commands, spark will provide these:

```
cd ~
wget https://archive.apache.org/dist/spark/spark-3.5.5/spark-3.5.5-bin-hadoop3.tgz
mkdir spark-3.5
tar xzvf spark-3.5.5-bin-hadoop3.tgz  -C spark-3.5 --strip-components=1
cd spark-3.5
```

```
 rahil@mac  ~/spark-3.5/jars  ls -l | grep hive
-rw-r--r--@ 1 rahil  staff    183633 Feb 23 12:45 hive-beeline-2.3.9.jar
-rw-r--r--@ 1 rahil  staff     44704 Feb 23 12:45 hive-cli-2.3.9.jar
-rw-r--r--@ 1 rahil  staff    436169 Feb 23 12:45 hive-common-2.3.9.jar
-rw-r--r--@ 1 rahil  staff  10840949 Feb 23 12:45 hive-exec-2.3.9-core.jar
-rw-r--r--@ 1 rahil  staff    116364 Feb 23 12:45 hive-jdbc-2.3.9.jar
-rw-r--r--@ 1 rahil  staff    326585 Feb 23 12:45 hive-llap-common-2.3.9.jar
-rw-r--r--@ 1 rahil  staff   8195966 Feb 23 12:45 hive-metastore-2.3.9.jar
-rw-r--r--@ 1 rahil  staff    916630 Feb 23 12:45 hive-serde-2.3.9.jar
-rw-r--r--@ 1 rahil  staff   1679366 Feb 23 12:45 hive-service-rpc-3.1.3.jar
-rw-r--r--@ 1 rahil  staff     53902 Feb 23 12:45 hive-shims-0.23-2.3.9.jar
-rw-r--r--@ 1 rahil  staff      8786 Feb 23 12:45 hive-shims-2.3.9.jar
-rw-r--r--@ 1 rahil  staff    120293 Feb 23 12:45 hive-shims-common-2.3.9.jar
-rw-r--r--@ 1 rahil  staff     12923 Feb 23 12:45 hive-shims-scheduler-2.3.9.jar
-rw-r--r--@ 1 rahil  staff    258346 Feb 23 12:45 hive-storage-api-2.8.1.jar
-rw-r--r--@ 1 rahil  staff    572320 Feb 23 12:45 spark-hive-thriftserver_2.12-3.5.5.jar
-rw-r--r--@ 1 rahil  staff    725252 Feb 23 12:45 spark-hive_2.12-3.5.5.jar
 rahil@mac  ~/spark-3.5/jars 
```

Which allows the hudi to not hit the class not found exception(atleast when testing via my local spark). Therefore I believe we will need to explicitly provide this in test env in order for spark hudi integration test to work.

plugins/spark/v3.5/integration/src/intTest/resources/logback.xml
rahil-c: This is a good point I can remove this `INFO` then, in order to avoid the extra logging. I am curious though if we should keep the test root logger at `WARN` instead of its current `ERROR`?

plugins/spark/v3.5/spark/build.gradle.kts
rahil-c: Thanks for catching this, you are right these polaris/iceberg dependencies are not needed will remove these.

plugins/spark/v3.5/spark/build.gradle.kts
rahil-c: Will remove this as not needed.

plugins/spark/v3.5/spark/build.gradle.kts
rahil-c: Currently I am directly using this one hudi class 
`import org.apache.spark.sql.hudi.catalog.HoodieInternalV2Table;` within the `PolarisCatalogUtils` 
https://github.com/apache/polaris/pull/1862/files#diff-f351cfc050ac63c907c7f35f8052eb88ee6241c39f95a8b81df3d895349bafa0R139

So we will need to keep a provided/compileOnly dependency, in order to ensure that compilation succeeds, as the user will anyway have to provide the hudi spark bundle directly like this https://hudi.apache.org/docs/0.15.0/quick-start-guide

If we are still concerned of having the bundle as provided/compileOnly dependency, I can try to narrow to just the exact hudi spark dep that contains this class which is hudi-spark3.2plus-common based on what I see in intellij and try declaring this instead.

Screenshot 2025-07-02 at 10 06 46 AM

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
rahil-c: Thanks Yun for the suggestion,

My question is shouldn't the engine first consult the polaris catalog service to see if its even authorized to perform a `createTable`? https://github.com/apache/polaris/blob/main/service/common/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java#L74

If this returns back to the client an exception, then I am not sure if it makes sense for client to perform logic of creating the table format metadata in cloud storage.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
rahil-c: This is only being called when the spark session has also added `HoodieSparkSessionExtensions` https://github.com/apache/polaris/pull/1862/files#diff-b9c22ea4a50bbb1b3e3db1f420b3656557c4875019321c24be2f1ee718e67b25R54 

https://github.com/apache/polaris/pull/1862/files
I can restucture the code to make this more clear.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
rahil-c: Yes can provide this

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkHudiIT.java
rahil-c: Inside the SparkCatalogTest we check to see if the table that implements spark's`Table` interface returned is a Hudi table https://github.com/apache/polaris/pull/1862/files#diff-0417fe873afac9a5be6c99e1089b52bb0cb5f004cdae1df5dcfc0c9b0faa9fdfR749

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
rahil-c: Currently PolarisCatalogUtils itself is not leveraging any interface, its just a util method https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java#L69 thats used to provide back an impl of Spark's `Table` interface. So there are no interfaces changes just an additional param passed in this method.

 I needed to do this for hudi as it will need a full identifier to the table in order to create construct Spark's V1 Catalog Table.

plugins/spark/v3.5/spark/build.gradle.kts
rahil-c: Have removed these

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
rahil-c: This `true` returned means the namespace entry was successfully dropped from Polaris catalog. If so, then  for hudi since it is using spark session catalog, we are dropping the namespace as well.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: Currently the `PolarisCatalogUtils`.`loadSparkTable` is using a `DataSourceV2Utils` Util in order to load the table using sparks table provider as seen [here](https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java#L86).

In the case for Delta this will go thru Delta's datasource impl which is implementing this V2`TableProviderInterface`, see DeltaDataSource https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaDataSource.scala#L57

The `PolarisCatalogUtils`.`loadSparkTable` currently assumes that other formats also implement this same TableProvider interface. However if they do not then this will fail with an exception.

Hudi in its spark integration does not implement that interface, see the entry point class for hudi datasource entry point `DefaultSource` https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/hudi/DefaultSource.scala#L55, meaning we will have to provide another way to load the hudi table, which is why I have added this condition and method called `loadHudiSparkTable`

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: `PolarisCatalogUtils.loadSparkTable` seems to be to return something that is a child of Spark `Table`, as it invoked by the `PolarisSparkCatalog#loadTable`
https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java#L67

When checking the hudi code base the only class I see that extends this `Table` is the `HoodieInternalV2Table` https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/catalog/HoodieInternalV2Table.scala#L44

And when checking the hudi code base i did not see any util class for providing the `HoodieInternalV2Table` hence why I had to construct it in this manner.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
rahil-c: This is fair point, will try to improve on this.

plugins/spark/v3.5/integration/build.gradle.kts
rahil-c: Thanks @RussellSpitzer for the suggestion, I can also try setting this as test runtime

plugins/spark/v3.5/integration/build.gradle.kts
rahil-c: @RussellSpitzer For delta it seems we have declared the test dep with testImplementation https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/integration/build.gradle.kts#L68
So i had done the same for hudi.

plugins/spark/v3.5/integration/build.gradle.kts
gh-yzou: sorry, what i mean is does spark-hive contains spark-sql? if yes, can we remove 
```
testImplementation("org.apache.spark:spark-sql_${scalaVersion}:${spark35Version}") {
    // exclude log4j dependencies. Explicit dependencies for the log4j libraries are
    // enforced below to ensure the version compatibility
    exclude("org.apache.logging.log4j", "log4j-slf4j2-impl")
    exclude("org.apache.logging.log4j", "log4j-1.2-api")
    exclude("org.apache.logging.log4j", "log4j-core")
    exclude("org.slf4j", "jul-to-slf4j")
  }
```

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkHudiIT.java
Davis-Zhang-Onehouse: +1, this should not fail as some default location will be used instead

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkHudiIT.java
```diff
@@ -0,0 +1,174 @@
...
+        .isInstanceOf(UnsupportedOperationException.class);
+
+    // CTAS fails
+    assertThatThrownBy(
+            () ->
+                sql(
+                    "CREATE TABLE %s USING HUDI LOCATION '%s' AS SELECT 1 AS id",
```
Davis-Zhang-Onehouse: CTAS as long as we write it properly, it should not fail.

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -46,6 +46,8 @@ dependencies {
   // TODO: extract a polaris-rest module as a thin layer for
   //  client to depends on.
   implementation(project(":polaris-core")) { isTransitive = false }
+  compileOnly("org.apache.hudi:hudi-spark3.2plus-common:0.15.0")
```
Davis-Zhang-Onehouse: maybe Ethan know this part better, starting hudi 1.x we deprecate some old spark version support, not sure about the status of spark 3.2

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
Davis-Zhang-Onehouse: Rahil it make sense. Just for your info, for alter table command, create hoodie table command, it seems like it is always we do actual hudi storage layer change before make catalog change. I would suggest we keep the pattern consistent.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
Davis-Zhang-Onehouse: +1

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
eric-maynard: > So there are no interfaces changes just an additional param passed in this method.

In this case, the method itself is the functional interface being changed. Java doesn't support static methods in interfaces, so there's obviously not an actual java interface being changed. 

It's still not clear to me why we can't load a table directly from object storage.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -263,25 +279,39 @@ public String[][] listNamespaces(String[] namespace) throws NoSuchNamespaceExcep
...
+      HudiCatalogUtils.createNamespace(namespace, metadata);
+    }
   }
 
   @Override
   public void alterNamespace(String[] namespace, NamespaceChange... changes)
       throws NoSuchNamespaceException {
     this.icebergsSparkCatalog.alterNamespace(namespace, changes);
+    if (PolarisCatalogUtils.isHudiExtensionEnabled()) {
+      HudiCatalogUtils.alterNamespace(namespace, changes);
```
eric-maynard: Looks like a bunch of comments got cleared, but I still see all these namespace-related changes which should not be necessary.

eric-maynard (changes requested):

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
eric-maynard: Why was this comment resolved?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
eric-maynard: 1. But you're not catching anything here
2. Sounds like a limitation on the Hudi side; maybe it makes sense for us to only support `V2TableProviderInterface` then?

plugins/spark/v3.5/spark/build.gradle.kts
RussellSpitzer: I'm in agreement here, I think we don't want to have any datasource specific code as a compile time dependency here (although runtime is fine).

My main reasons are
1. Avoiding having a dependency that would be un-needed by users without hudi
2. Licensing headaches

Could we check via a String and Reflection? I think "provided" is maybe an ok workaround but that makes this a but strongly tied in version to the Hudi client library? I was hoping we could keep the Catalog plugin basically version independent and relying pretty much only on Datasource Apis.

plugins/spark/v3.5/integration/src/intTest/resources/logback.xml
gh-yzou: We can if the WARN level log output doesn't significantly slow the test, but based on my previous testing with Spark, even with WARN, there seems quite a lot of Spark logs, that was part of the reason why we set the level to Error. If the test fails, people can always tune the log level to check more logs if needed, so i wouldn't worry too much about the restricted log level.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
gh-yzou: I take a look at how HoodieCatalog.scala works https://github.com/apache/hudi/blob/3369b09fd8fca1db1f8e655f79ecb7a97c57367b/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/catalog/HoodieCatalog.scala#L123, it actually works very similar as DeltaCatalog, https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L230. Before, the approach I have tried is to let PolarisSparkCatalog returns V1Table object, and in SparkCatalog we just call DeltaCatalog.loadTable, and it will return DeltaV2Table, in this way, we will not introduce any format specific dependency to the client side. I didn't end up doing that because Delta actually provided the DataSourceV2 support, and the loadTableUtils automatically loads the DeltaV2Table.

Since Hudi doesn't support DataSourceV2 yet, i think you can go with my original approach, let PolarisSparkCatalog returns V1Table on load, and let the HudiCatalog.loadTable takes care of the final returned table format

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
gh-yzou: @rahil-c I tried out the current client without your change for hudi usage, although it fails on loadTable, it doesn't really complain about namespaces, so I strongly believe we don't need any namespace specific changes here. I will check out the current change to verify also.

gh-yzou: After checking out your code with testing, I believe the fact you need those namespace operations are because you are calling the following code when load hudi table
```
catalogTable = sparkSession.sessionState().catalog().getTableMetadata(tableIdentifier);
```
This triggers the spark internal session catalog, which actually checks whether db exists. However,  i don't think we need to do that for loading hudi tables. I think we should just create a V1Table, and let the HudiCatalog take care of the final table format. To construct V1Table, you can either do that manually with filling all possible field, or another approach is do it in a similar way as Unity Catalog here https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L283

gh-yzou: For the CatalogTableCreation, maybe you can follow up on how hudi create the catalog table here https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/catalog/HoodieCatalog.scala#L293.

gh-yzou: As for createTable, i think the real problem comes from here https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/command/CreateHoodieTableCommand.scala#L198
Similar as delta, when we are using rest catalog as delegation, i think we should call into the catalogPlugin for table creation, instead of spark session catalog. https://github.com/delta-io/delta/blob/2d89954008b6c53e49744f09435136c5c63b9f2c/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L218

Delta today triggers a special check for unity catalog here https://github.com/delta-io/delta/blob/2d89954008b6c53e49744f09435136c5c63b9f2c/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L77, One way I am thinking is that we can introduce a special flag for Polaris SparkCatalog to represent that a third party catalog plugin  is used, and then do similar thing as DeltaCatalog

gh-yzou:
@rahil-c sorry, i made my comment yesterday, but forgot to push it. I did a push, and added some more comments, please let me know if you have more questions about this! 
As we have discussed, there are two main concerns for this PR:
1) the hudi dependency introduced for spark client, which is caused by the usage of HoodieInternalV2Table. This can be resolved by loading V1Table, and then let HudiCatalog loadTable to handle the final table result https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/catalog/HoodieCatalog.scala#L123
2) the extra namespace creation for HudiCatalog. Polaris Spark Client reuses the whole Iceberg namespace, ideally we do not want to maintain extra namespace creation just for specific table format. The needs of extra namespace creation is because HudiCatalog only works with SparkSession Catalog and HiveCatalog today https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/command/CreateHoodieTableCommand.scala#L198, however, since Polaris is rest catalog, this will not work anymore. We want to see if we can push forward on hudi community to improve the catalog implementation regarding to the third party catalog plugin. Similar as Delta did a special case for unity catalog here https://github.com/delta-io/delta/blob/2d89954008b6c53e49744f09435136c5c63b9f2c/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L218

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
rahil-c: Will follow @gh-yzou suggestion

settings.gradle.kts
```diff
@@ -113,6 +113,7 @@ dependencyResolutionManagement {
   repositoriesMode = RepositoriesMode.FAIL_ON_PROJECT_REPOS
   repositories {
     mavenCentral()
+    mavenLocal()  // Local builds still take precedence for SNAPSHOT versions
```
rahil-c: @gh-yzou this is just done for the purpose of my local testing, will remove once approach looks good to you.

rahil-c:
> @rahil-c sorry, i made my comment yesterday, but forgot to push it. I did a push, and added some more comments, please let me know if you have more questions about this! As we have discussed, there are two main concerns for this PR:
> 
> 1. the hudi dependency introduced for spark client, which is caused by the usage of HoodieInternalV2Table. This can be resolved by loading V1Table, and then let HudiCatalog loadTable to handle the final table result https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/catalog/HoodieCatalog.scala#L123
> 2. the extra namespace creation for HudiCatalog. Polaris Spark Client reuses the whole Iceberg namespace, ideally we do not want to maintain extra namespace creation just for specific table format. The needs of extra namespace creation is because HudiCatalog only works with SparkSession Catalog and HiveCatalog today https://github.com/apache/hudi/blob/master/hudi-spark-datasource/hudi-spark-common/src/main/scala/org/apache/spark/sql/hudi/command/CreateHoodieTableCommand.scala#L198, however, since Polaris is rest catalog, this will not work anymore. We want to see if we can push forward on hudi community to improve the catalog implementation regarding to the third party catalog plugin. Similar as Delta did a special case for unity catalog here https://github.com/delta-io/delta/blob/2d89954008b6c53e49744f09435136c5c63b9f2c/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L218

Thanks @gh-yzou, I have followed the recommendations above and updated the pr. Let me know if the approach looks good to you. If so then I can try to break this down into smaller prs.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
rahil-c: @gh-yzou @eric-maynard  Have followed the recommendation above to remove namespace-related changes, this should be resolved now.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiCatalogUtils.java
rahil-c: Resolving this comment as this code is no longer present.

plugins/spark/v3.5/spark/build.gradle.kts
rahil-c: I have removed this hudi class in order to avoid bringing in a hudi dependency for now based on @gh-yzou recommendation.

plugins/spark/v3.5/integration/build.gradle.kts
```diff
@@ -60,6 +60,7 @@ dependencies {
     exclude("org.apache.logging.log4j", "log4j-core")
     exclude("org.slf4j", "jul-to-slf4j")
   }
+
```
gh-yzou: let's revert this line, seems not necessary

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -91,6 +109,85 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+
+    throw new IllegalStateException(
+        "Could not obtain Polaris catalog identifier."
+            + "Expected following configuration to be set in session: spark.sql.catalog.<CATALOG_NAME>=org.apache.polaris.spark.SparkCatalog");
+  }
+
+  public static Table loadV1SparkHudiTable(GenericTable genericTable, Identifier identifier) {
+    Map<String, String> tableProperties = Maps.newHashMap();
+    tableProperties.putAll(genericTable.getProperties());
+    tableProperties.put(
```
gh-yzou: this part is the same as the loadSparkTable where we are updating the table property for spark load table. how about extract a common function called noralizeTablePropertyForLoadSparkTable, and put all property handling in the function

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -91,6 +109,85 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+    org.apache.spark.sql.types.StructType emptySchema = new org.apache.spark.sql.types.StructType();
+    scala.collection.immutable.Seq<String> emptyStringSeq =
+        scala.collection.JavaConverters.asScalaBuffer(new java.util.ArrayList<String>()).toList();
+    CatalogTable catalogTable =
+        new CatalogTable(
+            tableIdentifier,
+            CatalogTableType.EXTERNAL(),
+            storage,
+            emptySchema,
+            Option.apply(genericTable.getProperties().get("provider")),
```
gh-yzou: let's use genericTable.format for the format instead of the provider as source of truth come from catalog service.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -156,15 +159,19 @@ public Table createTable(
...
-      } else {
-        return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
       }
+      if (PolarisCatalogUtils.useHudi(provider)) {
+        // For creating the hudi table, we load HoodieCatalog
+        // to create the .hoodie folder in cloud storage
+        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
+        return hudiCatalog.createTable(ident, schema, transforms, properties);
+      }
+      return this.polarisSparkCatalog.createTable(ident, schema, transforms, properties);
```
gh-yzou: @rahil-c I believe with the change you are doing on hudi community, you don't need the extra polarisSparkCatalog.createTable anymore. 

Let's restructure this like what eric suggested, we can do 
```
if (PolarisCatalogUtils.useDelta(provider)) {
} else if (PolarisCatalogUtils.useHudi(provider)) {
} else {
}
```

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -91,6 +109,85 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+    Map<String, String> tableProperties = Maps.newHashMap();
+    tableProperties.putAll(genericTable.getProperties());
+    tableProperties.put(
+        TABLE_PATH_KEY, genericTable.getProperties().get(TableCatalog.PROP_LOCATION));
+
+    // Need full identifier in order to construct CatalogTable correctly for Hudi
+    String namespacePath = String.join(".", identifier.namespace());
+    TableIdentifier tableIdentifier =
+        new TableIdentifier(
+            identifier.name(), Option.apply(namespacePath), Option.apply(getCatalogName()));
```
gh-yzou: when I was handling delta with V1Spark table, i don't think we actually need this. If we do not put the catalog name here, what happens?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -91,6 +109,85 @@ public static Table loadSparkTable(GenericTable genericTable) {
         provider, new CaseInsensitiveStringMap(tableProperties), scala.Option.empty());
   }
 
+  /**
+   * Extract catalog name from Spark session configuration. Looks for configuration like:
+   * spark.sql.catalog.<CATALOG_NAME>=org.apache.polaris.spark.SparkCatalog
+   */
+  private static String getCatalogName() {
```
gh-yzou: do you need the catalog name at client side, or server side? from your description, it seems you only need the client side, then you should be able to direct get the catalog name from the PolarisSparkCatalog.catalogName, we probably don't need this parsing function

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: I can look into doing this.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: @gh-yzou 
I believe the constructor of Spark's `catalogTable` is expecting this `provider`
Screenshot 2025-07-25 at 4 21 36 PM

If you feel though that we should be using instead `genericTable.format` (which likely should have the same value as provider) then I can try switching to it. Thoughts?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
rahil-c: Will restructure this

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: Hudi is expecting the Spark `V1Table` which will returned by Polaris to contain this catalog Identifier. This is because Hudi is currently using Spark DSV1 it will invoke the following Spark `Rule`, `FindDataSourceTable` which expects this catalog identifier `table.identifier.catalog.get`. If the catalog does not return this catalog identifier, then during `.get` it will hit a NPE on spark side and error out.

Therefore we will need this in order for the integration to work.

Class is `DataSourceStrategy.scala` in spark code base
Screenshot 2025-07-25 at 4 57 12 PM

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: Will try this suggestion

plugins/spark/README.md
```diff
@@ -124,3 +124,8 @@ Following describes the current functionality limitations of the Polaris Spark c
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
 5) For other non-Iceberg tables like csv, it is not supported today.
+
+### Hudi Support
+Currently support for Hudi tables within the Polaris catalog is still under development. 
+The Hudi community has made a change to integrate with Polaris, and is planning on doing a minor release.
```
gh-yzou: -> hudi-spark-xxx is required for hudi table support to work end to end, which is still under releasing.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -87,10 +95,77 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+    TableProvider provider =
+        DataSource.lookupDataSourceV2(genericTable.getFormat(), sparkSession.sessionState().conf())
+            .get();
+    Map<String, String> tableProperties = normalizeTablePropertiesForLoadSparkTable(genericTable);
     return DataSourceV2Utils.getTableFromProvider(
         provider, new CaseInsensitiveStringMap(tableProperties), scala.Option.empty());
   }
 
+  /** Return a Spark V1Table for Hudi tables. */
+  public static Table loadV1SparkHudiTable(
```
gh-yzou: actually this function doesn't seem very specific for huid, maybe we can just call it loadV1SaprkTable, and in the comment mention that it is currently only used by hudi.

rahil-c: will do so

plugins/spark/README.md
rahil-c: will add this line

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
rahil-c: This should be resolved now based on recent discussions.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: This should be resolved now based on recent discussions.

rahil-c:
@flyrain @gh-yzou @eric-maynard https://github.com/apache/polaris/pull/1862#discussion_r2190630881 has been resolved now based on recent changes.

Was wondering if we can land this as all comments should be addressed.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
rahil-c: This has been addressed please see recent discussion here https://github.com/apache/polaris/pull/1862#discussion_r2191328980

flyrain (dismissed):
LGTM. Thanks @rahil-c !

gh-yzou (dismissed):

gh-yzou:
@eric-maynard I think @rahil-c addressed all the comments. I am going to dismiss the requested change for this PR, so that we can move forward. Once you are back, maybe we can follow up with a post review.

eric-maynard:
I’ll take a look soon — it looks like the PR was just updated yesterday to address the comments

eric-maynard:
I’ll take a look soon — it looks like the PR was just updated yesterday to address the comments

plugins/spark/README.md
```diff
@@ -124,3 +124,9 @@ Following describes the current functionality limitations of the Polaris Spark c
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
 5) For other non-Iceberg tables like csv, it is not supported today.
+
+### Hudi Support
+Currently support for Hudi tables within the Polaris catalog is still under development.
```
eric-maynard: I'm confused as to why we would need this. If the integration needs changes on the Hudi side to work, why would we merge anything into Polaris now?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiHelper.java
```diff
@@ -0,0 +1,71 @@
...
+          String.format("Cannot initialize Hudi Catalog %s: %s", hudiCatalogImpl, e.getMessage()),
+          e);
+    }
+
+    try {
+      this.hudiCatalog = ctor.newInstance();
+    } catch (ClassCastException e) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Cannot initialize Hudi Catalog, %s does not implement Table Catalog.",
```
eric-maynard: Is this supposed to say TableCatalog?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -29,14 +29,23 @@
...
 import org.apache.spark.sql.connector.catalog.TableProvider;
 import org.apache.spark.sql.execution.datasources.DataSource;
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Utils;
 import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.Option;
 
 public class PolarisCatalogUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(PolarisCatalogUtils.class);
```
eric-maynard: We use LOGGER elsewhere

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -87,10 +95,80 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+        (scala.collection.immutable.Map<String, String>)
+            scala.collection.immutable.Map$.MODULE$.apply(
+                scala.collection.JavaConverters.mapAsScalaMap(tableProperties).toSeq());
+
+    org.apache.spark.sql.catalyst.catalog.CatalogStorageFormat storage =
+        DataSource.buildStorageFormatFromOptions(scalaOptions);
+
+    // Currently Polaris generic table does not contain any schema information, partition columns,
+    // stats, etc
+    // for now we will just use fill the parameters we have from catalog, and let underlying client
```
eric-maynard: `use fill`?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
```diff
@@ -87,10 +95,80 @@ public static Table loadSparkTable(GenericTable genericTable) {
...
+    CatalogTable catalogTable =
+        new CatalogTable(
+            tableIdentifier,
+            CatalogTableType.EXTERNAL(),
+            storage,
+            emptySchema,
+            Option.apply(genericTable.getFormat()),
+            emptyStringSeq,
+            scala.Option.empty(),
+            genericTable.getProperties().get("owner"),
```
eric-maynard: This property currently isn't defined anywhere. Are you somehow setting it on writes?

If so, this should be a constant somewhere. If not, you should remove this.

plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
```diff
@@ -418,7 +449,6 @@ void testCreateAndLoadGenericTable(String format) throws Exception {
             () -> catalog.createTable(identifier, defaultSchema, new Transform[0], newProperties))
         .isInstanceOf(TableAlreadyExistsException.class);
 
-    // drop the iceberg table
```
eric-maynard, line 421: Spurious change?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
eric-maynard: Oh, it looks like 2 classes (both in the client) use LOG, hm. I wouldn't fix that here, but maybe just stick with LOGGER

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiHelper.java
rahil-c: Let me fix this to say `TableCatalog`, i believe had gotten it from the setup of `DeltaHelper`  https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/DeltaHelper.java#L66

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: Will fix

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: Yes we have usages of `LOG` in the following code paths.

https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java#L69

https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/DeltaHelper.java#L32

Personally Im not sure the naming between `LOG` and `LOGGER` makes a difference from either a functional or aesthetics perspective but I can make this change so we can move forward.

plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
rahil-c: Yea will include this line back.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: On my side, I am not explicitly setting this property on hudi side changes, or in the polaris changes. This seems to be coming from Spark engine itself setting this value in the properties map.

For example this property gets propagated during the Polaris `SparkCatalog#createTable` which overrides Spark's `TableCatalog` interface. If you try testing with `Delta` in the create table and examine the `properties` map. 

Screenshot 2025-07-29 at 12 13 55 PM
 
You can see the `owner` for the table is already set, before we even make a createGenericTable request
Screenshot 2025-07-29 at 12 14 01 PM
The `createGenericTableRequest` will then take those properties, and ensure they get persisted in the `GenericTable` object on the catalog side. 

If the ask is to just have this  `"owner"` be a constant variable called `public static final String OWNER = "owner";` I can do that.

plugins/spark/README.md
rahil-c: First off, we have already landed the initial changes on hudi side https://github.com/apache/hudi/pull/13558 in order to integrate Hudi with Polaris, based on discussion between members of the Polaris community and Hudi community.

We have discussed between both communities that Polaris would need the latest Hudi release artifact, and for that we will need to do a hudi point release which I have already started the thread here https://lists.apache.org/thread/4ztwgclljojg7r08mzm2dkynwfrvjlqb.

As doing a release for any open source project can be timely(with alot of back and forth), we aligned that before even doing starting the hudi point release we should first ensure this initial polaris side change is landed, as this change is not even dependent on any hudi release artifact and would allow confidence to even start the point hudi release.

This was already aligned between both communities hence why this discussion thread was started yesterday:https://lists.apache.org/thread/k524b5xq7l75tzz6sdzth15wjxdgp3gf for our hudi code freeze process, as we had obtained approval of a PMC and a committer for this PR.

cc @flyrain @gh-yzou @singhpk234

flyrain: I'm fine with working in parallel with the Hudi community. This PR servers as the first step of the integration, and we file follow-up PRs once Hudi 1.0.3 is out. With that, I guess this doc section itself is not quite necessary. It's normal that a feature is split to multiple PRs.

gh-yzou: @eric-maynard, @rahil-c is trying to parallelize the development work based on his POC, the hudi development is already in and under the release, meanwhile, @rahil-c is trying to start the necessary development wrok in Polaris. The Polaris Hudi support is still under development (which is what the readme change intended to say), and once the hudi plugin is released, @rahil-c will follow up with integration and regression tests, and claim hudi is supported with clear documentation.
If the readme description is confusing, we can remove it or make it more clear if that works?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiHelper.java
gh-yzou: @rahil-c maybe fix both

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
gh-yzou: @rahil-c will you be able to update the other two to LOGGER if we are changing this one?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/HudiHelper.java
rahil-c: Ack will do so

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
rahil-c: Sure

plugins/spark/README.md
vinothchandar: Hi all - FWIW.. we have active work around this on the next Hudi patch release, based on the feedback from the polaris community. So we are aligned on the goals of making it work.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PolarisCatalogUtils.java
eric-maynard: The ask is that, ideally, we can re-use the existing constant (which based on your comment looks to be coming from [here](https://github.com/apache/spark/blob/dc8fba647ac1042fd83a8ecbeadb45687c784b5b/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/TableCatalog.java#L89)). Barring that, yes, please make a new constant.

plugins/spark/README.md
eric-maynard: >  based on discussion between members of the Polaris community and Hudi community.

>  we aligned that before even doing starting the hudi point release we should first ensure this initial polaris side change is landed,

> This was already aligned between both communities

Did this happen on a mailing list? Can you point me to that?

eric-maynard: It's less that the README is confusing (as Yufei said, it's normal to merge things that are in-development within a project) and more that it's confusing why we would merge something into Polaris because we expect an upstream dependency to eventually change. Couldn't we just forego the README note and hold the Polaris change until the dependency is ready?

yihua (approved):
LGTM based on the alignment between Hudi and Polaris community.  The integration will work on the patch release of Hudi containing this change https://github.com/apache/hudi/pull/13558.

rahil-c:
Can we open this pr back up not sure why it was closed? @singhpk234 @flyrain @gh-yzou @eric-maynard

flyrain:
Hi @rahil-c. reopened.  Please update as needed.

plugins/spark/README.md
```diff
@@ -123,4 +123,4 @@ Following describes the current functionality limitations of the Polaris Spark c
 2) Create a Delta table without explicit location is not supported.
 3) Rename a Delta table is not supported.
 4) ALTER TABLE ... SET LOCATION is not supported for DELTA table.
-5) For other non-Iceberg tables like csv, it is not supported today.
+5) For other non-Iceberg tables like csv, it is not supported today.
```
sfc-gh-ygu: Not a blocker: This change is not needed, but I think we will need a doc change to claim that Hudi table is supported. I'm OK to add it in a followup PR.

rahil-c:
@flyrain @gh-yzou Now that Hudi 1.1.0 is out (see hudi release docs https://hudi.apache.org/releases/release-1.1.0) I think we can land this PR. Based on the original discussion we wanted to have a follow up PR for the IT test plan, which will follow a similar flow as the `SparkDeltaIT`: https://github.com/apache/polaris/blob/main/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkDeltaIT.java

Once we land this PR, will raise the Hudi IT [PR](https://github.com/rahil-c/polaris/pull/10/files#diff-ec6cb0157edd392da9149d2627b6a928846e236050232678c525f0dc744c6963R37 (adding hudi 1.1.0 as a test dependency similar to Delta) and the test which verifies that DDL, DML, DQL queries with hudi as a generic table.

plugins/spark/README.md
flyrain: Given the Hudi 1.1.0 has been released, I think this PR is ready to merge.

plugins/spark/README.md
gh-yzou: there seems no change, maybe revert this part

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
```diff
@@ -71,7 +71,11 @@ public Table loadTable(Identifier identifier) throws NoSuchTableException {
     try {
       GenericTable genericTable =
           this.polarisCatalog.loadGenericTable(Spark3Util.identifierToTableIdentifier(identifier));
-      return PolarisCatalogUtils.loadSparkTable(genericTable);
+      if (PolarisCatalogUtils.useHudi(genericTable.getFormat())) {
```
gh-yzou, line 75: can you add a comment here about why we are returning V1Table for Hudi, but V2 table for rest? For example, "hudi doesn't support datasource v2 yet, and therefore returns V1Table"

PR #1876, opened 2025-06-11
Avoid using org.testcontainers.shaded.**

dimas-b:

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -40,6 +40,7 @@ antlr4-runtime = { module = "org.antlr:antlr4-runtime", version.strictly = "4.9.
 assertj-core = { module = "org.assertj:assertj-core", version = "3.27.3" }
 auth0-jwt = { module = "com.auth0:java-jwt", version = "4.5.0" }
 awssdk-bom = { module = "software.amazon.awssdk:bom", version = "2.31.59" }
+awaitility = { module = "org.awaitility:awaitility", version = "4.3.0" }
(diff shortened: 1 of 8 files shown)


snazy (approved):

gradle/libs.versions.toml
```diff
@@ -40,6 +40,7 @@ antlr4-runtime = { module = "org.antlr:antlr4-runtime", version.strictly = "4.9.
 assertj-core = { module = "org.assertj:assertj-core", version = "3.27.3" }
 auth0-jwt = { module = "com.auth0:java-jwt", version = "4.5.0" }
 awssdk-bom = { module = "software.amazon.awssdk:bom", version = "2.31.59" }
+awaitility = { module = "org.awaitility:awaitility", version = "4.3.0" }
```
singhpk234, line 43: do we need to add this in licence ?

singhpk234 (approved):
LGTM ! one minor doubt on the LICENSE

gradle/libs.versions.toml
dimas-b: I do not think so: we use it only in test code; Awaitility is ALv2 and does not have a NOTICE... but happy to include in a follow-up PR (merging for now to unblock related work)... @jbonofre WDYT?

singhpk234: sg!

snazy: Generally speaking: only stuff that's distributed or "copied" into the source tree needs to be mentioned in license/notice.

Issue #1861, opened 2025-06-11
Two links in top README point to wrong places


flyrain:
### Describe the bug

Here are the code snippet:
```
[Polaris management API doc](https://polaris.apache.org/index.html#tag/polaris-management-service_other)
and [Apache Iceberg REST API doc](https://polaris.apache.org/index.html#tag/Configuration-API).
```

PR #1879, opened 2025-06-12
Fix two wrong links in README.md

flyrain:
Fix https://github.com/apache/polaris/issues/1861

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ enabling seamless multi-engine interoperability across a wide range of platforms
 Apache Spark™, StarRocks, and Trino.
 
 Documentation is available at https://polaris.apache.org, including
-[Polaris management API doc](https://polaris.apache.org/index.html#tag/polaris-management-service_other)
-and [Apache Iceberg REST API doc](https://polaris.apache.org/index.html#tag/Configuration-API).
+[Polaris management API doc](https://editor-next.swagger.io/?url=https://raw.githubusercontent.com/apache/polaris/refs/heads/main/spec/polaris-management-service.yml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,8 +25,8 @@ enabling seamless multi-engine interoperability across a wide range of platforms
 Apache Spark™, StarRocks, and Trino.
 
 Documentation is available at https://polaris.apache.org, including
-[Polaris management API doc](https://polaris.apache.org/index.html#tag/polaris-management-service_other)
-and [Apache Iceberg REST API doc](https://polaris.apache.org/index.html#tag/Configuration-API).
+[Polaris management API doc](https://editor-next.swagger.io/?url=https://raw.githubusercontent.com/apache/polaris/refs/heads/main/spec/polaris-management-service.yml)
+and [Polaris Catalog API doc](https://editor-next.swagger.io/?url=https://raw.githubusercontent.com/apache/polaris/refs/heads/main/spec/generated/bundled-polaris-catalog-service.yaml).
```
dimas-b, line 29: While the new links work, the overall statement becomes somewhat confusing. The first part states that documentation is available on the `polaris.apache.org` site, but the two following links point to `swagger.io`. There is not technical problem there, but from the UX perspective I think it is preferable to avoid taking the reader to a web site that differs from what the reader is lead to believe to be the destination.

flyrain: Agreed. Changed the wording a bit.

dimas-b: thx :+1:

dimas-b (approved):

PR #1880, opened 2025-06-12
feat(build): Add Checkstyle plugin and an IllegalImport rule

adutra:
Now ready for review.

I'm implementing my own suggestion of using a Checkstyle rule to check for forbidden packages.

### DIFF
--- a/build-logic/src/main/kotlin/polaris-java.gradle.kts
+++ b/build-logic/src/main/kotlin/polaris-java.gradle.kts
@@ -38,14 +38,13 @@ plugins {
 
 apply<PublishingHelperPlugin>()
 
-val checkstyleVersion =
-  versionCatalogs
-    .named("libs")
-    .findVersion("checkstyle")
-    .orElseThrow { GradleException("checkstyle version not found in libs.versions.toml") }
-    .requiredVersion
-
 checkstyle {
+  val checkstyleVersion =
+    versionCatalogs
+      .named("libs")
+      .findVersion("checkstyle")
+      .orElseThrow { GradleException("checkstyle version not found in libs.versions.toml") }
+      .requiredVersion
   toolVersion = checkstyleVersion
   configFile = rootProject.file("codestyle/checkstyle.xml")
   isIgnoreFailures = false
@@ -54,9 +53,7 @@ checkstyle {
 }
 
 // Ensure Checkstyle runs after jandex to avoid task dependency issues
-tasks
-  .matching { it.name.startsWith("checkstyle") }
-  .configureEach { mustRunAfter(tasks.matching { it.name == "jandex" }) }
+tasks.withType<Checkstyle>().configureEach { tasks.findByName("jandex")?.let { mustRunAfter(it) } }
 
 tasks.withType(JavaCompile::class.java).configureEach {
   options.compilerArgs.addAll(listOf("-Xlint:unchecked", "-Xlint:deprecation"))
--- a/build-logic/src/main/kotlin/polaris-java.gradle.kts
+++ b/build-logic/src/main/kotlin/polaris-java.gradle.kts
@@ -30,6 +30,7 @@ plugins {
   `java-library`
   `java-test-fixtures`
   `jvm-test-suite`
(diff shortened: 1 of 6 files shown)


quarkus/test-commons/build.gradle.kts
```diff
@@ -24,9 +24,11 @@ plugins {
 }
 
 configurations.all {
-  exclude(group = "org.antlr", module = "antlr4-runtime")
-  exclude(group = "org.scala-lang", module = "scala-library")
-  exclude(group = "org.scala-lang", module = "scala-reflect")
+  if (name != "checkstyle") {
```
adutra, line 27: Checkstyle depends on `antlr4-runtime`.

build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -30,18 +30,34 @@ plugins {
...
+  toolVersion = checkstyleVersion
+  configFile = rootProject.file("codestyle/checkstyle.xml")
+  isIgnoreFailures = false
+  maxErrors = 0
+  maxWarnings = 0
 }
 
+// Ensure Checkstyle runs after jandex to avoid task dependency issues
+tasks
+  .matching { it.name.startsWith("checkstyle") }
```
snazy: This is a trap and prevents "configure on demand", `.matching` gives you the fully configured tasks.
I'd use
```kotlin
tasks.withType<Checkstyle>().configureEach { ...
```

build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -30,18 +30,34 @@ plugins {
...
   id("jacoco-report-aggregation")
   id("net.ltgt.errorprone")
 }
 
 apply<PublishingHelperPlugin>()
 
-if (project.extra.has("duplicated-project-sources")) {
-  // skip the style check for duplicated projects
-  tasks.withType<Checkstyle>().configureEach { enabled = false }
+val checkstyleVersion =
```
snazy: Nit: move this into the `checkstyle {` block.

codestyle/checkstyle.xml
```diff
@@ -0,0 +1,52 @@
...
+      default="checkstyle-suppressions.xml" />
+    <property name="optional" value="true"/>
+  </module>
+
+  <module name="TreeWalker">
+    <!-- Checks for imports                            -->
+    <!-- See http://checkstyle.org/config_imports.html -->
+    <module name="IllegalImport">
+      <property name="illegalPkgs"
+        value="org.apache.iceberg.shaded, org.apache.iceberg.relocated, org.testcontainers.shaded"/>
```
dimas-b: nit: I wonder if we could have `**.shaded.**`, `**.relocated.**`?

dimas-b: it looks like `regexp` is allowed :thinking:

snazy:
Not against this PR, but wouldn't it be better to not even have those dependencies that bring relocated code or not have dependencies on uber jars at all, as long as there are alternatives?

adutra:
> Not against this PR, but wouldn't it be better to not even have those dependencies that bring relocated code or not have dependencies on uber jars at all, as long as there are alternatives?

Well, we certainly could try to ban the iceberg shaded dependencies, but afair the Spark Client module is fairly complex in terms of dependency management.

And as for Testcontainers, I think that if your tests require testcontainers, you get the testcontainers shaded dependencies (e.g. Awaitility) "for free" 🤷‍♂️ – and there is no way to exclude them since they are in the same jar.

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -112,44 +112,6 @@ dependencies {
   }
 }
 
-// TODO: replace the check using gradlew checkstyle plugin
-tasks.register("checkNoDisallowedImports") {
```
gh-yzou, line 116: Thanks for help getting rid of that

gh-yzou (approved):

PR #485, opened 2024-11-26
WIP Release infra stuff

snazy:
Introduce a couple of shell scripts to automate the release process. Some docs are included under `releases/README.md`.

Generally, releases at the ASF follow the following workflow:
1. Draft a release
2. Start a VOTE on the dev mailing list
3. If the VOTE fails, the release has failed - "go to step 1"
4. If the VOTE passes, publish the release

The above process is, without release branches, reflected in the scripts:
1. `releases/bin/draft-release.sh --major <major-version-number> --minor <minor-version-number> --commit <Git-commit-ID>`
2. if the vote passes: `releases/bin/publish-release.sh --major <major-version-number> --minor <minor-version-number>`
3. if the vote fails, just run `draft-release.sh` again

The change includes scripts to handle version branches, however, using those is not required for the two release scripts above.

It's important to know that the scripts handle changes to the `version.txt` file and that a specific syntax for Git tags is expected, which is required to automatically use/generate RC and final versions, including the artifact publishing via Sonatype.

PR #1883, opened 2025-06-12
Bump version in version.txt [1.0]

eric-maynard:
With the 1.0 release branch being cut, we should bump this to reflect the current state of main

### DIFF
--- a/version.txt
+++ b/version.txt
@@ -1,1 +1,1 @@
-0.11.0-beta-incubating-SNAPSHOT
+1.1.0-incubating-SNAPSHOT


snazy (changes requested):
Let's not use the 1.0.x branch. It's not the branch that's documented in the release guide on the web site.

eric-maynard:
@snazy I understood that the branch was going to be renamed, are you suggesting that we instead cut a new branch from the tip of main?

version.txt
```diff
@@ -1 +1 @@
-0.11.0-beta-incubating-SNAPSHOT
+1.1.0-incubating-SNAPSHOT
```
jbonofre: I guess it's `1.0.0-incubating` on the `1.0.x` release branch.

The release should be named `release/1.0.x` according to https://polaris.apache.org/community/release-guide/

version.txt
```diff
@@ -1 +1 @@
-0.11.0-beta-incubating-SNAPSHOT
+1.0.0-incubating-SNAPSHOT
```
jbonofre: Sorry, I wasn't clear in my previous comment.

On release branch, it's not a SNAPSHOT, it's a release, so `1.0.0-incubating`.

snazy:
> @snazy I understood that the branch was going to be renamed, are you suggesting that we instead cut a new branch from the tip of main?

Yep (probably quicker than tweaking an existing PR).

version.txt
snazy: Well, nope. On a _tag_ it's a release version - but not in a branch.

snazy: (Otherwise you get a non-reproducible build)

eric-maynard: Based on [the other branches](https://github.com/apache/polaris/blob/release/0.9.x/version.txt) it looks like it should have no SNAPSHOT but let me know if this does need to change back

eric-maynard:
@snazy I'll leave the question of rename vs. new cut to the release manager; my (incorrect?) assumption was that once we've picked a commit to release from we'd prefer not to change that lightly. If we end up closing the target branch here we can cherry pick this onto the new branch 👍

snazy:
> @snazy I'll leave the question of rename vs. new cut to the release manager; my (incorrect?) assumption was that once we've picked a commit to release from we'd prefer not to change that lightly. If we end up closing the target branch here we can cherry pick this onto the new branch 👍

Yea - cherry-picking on the new branch is fine for me. I'd be very careful with removing the `-SNAPSHOT` suffix in a Git _branch_ - there can be unforeseen consequences.

BTW, if you're interested: Some technical background on semi-automatic releases [is here](https://github.com/apache/polaris/pull/485/files#diff-c06e8f0d65e645e2d79dec3ae62605594fb8f5949f24b9d23344ff4f2b0a287c). There are quite some annoying things for releases that can go wrong - most annoying is "concurrent merges/commits" that do happen in real life.

eric-maynard:
Just changed the target to the new branch `release/1.0.x` -- thanks guys!

dimas-b:
I do not think the scope for 1.0 is completely set per dev email thread: https://lists.apache.org/thread/g1m3nj9g7jnrvkd9ogq6h3v4cmll5lwo

Let's hold version updates until we have consensus there.

version.txt
```diff
@@ -1 +1 @@
-0.11.0-beta-incubating-SNAPSHOT
+1.0.0-incubating
```
dimas-b, line 1: I suppose _after_ the 1.0.0 release is under way, this version ought to be something like `1.0.1-incubating`.

eric-maynard:
@dimas-b I have only been loosely following that thread, but it looks like we may be all set now?

dimas-b:
The question of re-branching from `main` has been settled, indeed, so I'm ok with merging this PR. No re-branching per: https://lists.apache.org/thread/m4lfwo9lkwjj8gjhqm8no5nhlnd2cdhq

Re: scope, let's use that dev ML thread for any follow-up items.

dimas-b (approved):

Issue #1785, opened 2025-06-03
LoadGenericTable table properties not being set correctly when using Polaris Spark Client


rahil-c:
## Issue Summary
Currently when trying out the spark polaris plugin for creating generic tables with DELTA format running into the following issue below. 

Followed the docs here: https://polaris.apache.org/in-dev/unreleased/polaris-spark-client/ however noticed that there is an issue with the artifact missing `org.apache.polaris:polaris-iceberg-1.8.1-spark-runtime-3.5_2.12:1.0.0`

So had to build the latest polaris spark client jar. Also note that I am using S3 as my storage type and creating a Delta Table, so included some additional jars such as (org.apache.hadoop:hadoop-aws:3.3.4,com.amazonaws:aws-java-sdk-bundle:1.12.671). See repro steps below.

### To Reproduce

```
spark-sql \
  --jars /Users/rahil/workplace/polaris/plugins/spark/v3.5/build/2.12/libs/polaris-iceberg-1.8.1-spark-runtime-3.5_2.12-0.10.0-beta-incubating-SNAPSHOT.jar \
  --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.9.0,org.apache.iceberg:iceberg-aws-bundle:1.9.0,io.delta:delta-spark_2.12:3.3.1,org.apache.hadoop:hadoop-aws:3.3.4,com.amazonaws:aws-java-sdk-bundle:1.12.671 \
  --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,io.delta.sql.DeltaSparkSessionExtension \
  --conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
  --conf spark.sql.catalog.quickstart_catalog.warehouse=quickstart_catalog \
  --conf spark.sql.catalog.quickstart_catalog.header.X-Iceberg-Access-Delegation=vended-credentials \
  --conf spark.sql.catalog.quickstart_catalog=org.apache.polaris.spark.SparkCatalog \
  --conf spark.sql.catalog.quickstart_catalog.uri=http://localhost:8181/api/catalog \
  --conf spark.sql.catalog.quickstart_catalog.credential="${USER_CLIENT_ID}:${USER_CLIENT_SECRET}" \
  --conf spark.sql.catalog.quickstart_catalog.scope='PRINCIPAL_ROLE:ALL' \
[...]

PR #1889, opened 2025-06-12
Add webpage for Generic Table support

gh-yzou:
add a webpage for generic table support


site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+- **doc** (optional): Comment or description for the table
+
+## Generic Table API Vs. Iceberg Table API
+
+Generic Table provides a different set of APIs to operate on the Generic Table entities while Iceberg APIs operates on
+the Iceberg Table entities.
+
+| Operations   | **Generic Table API**                                       | **Iceberg Table API**                                       |
+|--------------|-------------------------------------------------------------|-------------------------------------------------------------|
+| Create Table | create an Iceberg table                                     | create a generic table                                      |
```
dimas-b: Are titles of columns 2 and 3 swapped WRT row contents by any chance?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+| Load Table   | load an Iceberg table, load a generic table throws an error | load a generic table, load an Iceberg table throws an error |
```
dimas-b: Does this mean that attempting to load an Iceberg table via the Generic Tables API will throw an error?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+| Drop Table   | drop an Iceberg table, drop a generic table throws an error | drop a generic table, drop an Iceberg table throws an error |
+| List Table   | list all Iceberg tables                                     | list all generic tables                                     |
+
+Note that generic table shares the same namespace with Iceberg tables, the table name has to be unique under the same namespace.
+
+## Working with Generic Table
+
+There are two ways to work with Polaris Generic Tables today:
+1) Directly communicate with Polaris through REST API calls using curl. Details will be described in the later section.
```
dimas-b: `using curl` seems to be an example rather than a requirement here, right? I suppose any tool capable to making HTTP requests will work just as well.

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+2) Use the Spark Client provided if you are working with Spark. Please refer to [Polaris Spark Client]({{% ref "polaris-spark-client" %}}) for detailed instructions.
+
+### Create a Generic Table
+
+To create a generic table, you need to provide the corresponding fields as described in [What is a Generic Table](#what-is-a-generic-table).
+Following is the REST API for creating a generic Table:
+
+```json
+POST /polaris/v1/{prefix}/namespaces/{namespace}/generic-tables
```
dimas-b: nit: since this block of text is in JSON format the `POST` line should probably be outside (otherwise it violates the format)

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+```
+And the response looks like the following:
+```json
+{
+  "table": {
+    "name": "delta_table",
+    "format": "delta",
+    "base-location": null,
+    "doc": null,
+    "properties": null
```
dimas-b: This example is rather confusing IMHO. It provides neither the location nor properties... What does it represent?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+  - If no location is provided, clients or users are responsible for managing the location.
+- **properties** (optional): Properties for the generic table passed on creation
+- **doc** (optional): Comment or description for the table
+
+## Generic Table API Vs. Iceberg Table API
+
+Generic Table provides a different set of APIs to operate on the Generic Table entities while Iceberg APIs operates on
+the Iceberg Table entities.
+
+| Operations   | **Generic Table API**                                       | **Iceberg Table API**                                       |
```
dimas-b: What is the Operation for making changes to the data in a Generic Table?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+To drop the table `delat_table`, use the following:
+```json
+DELETE /polaris/v1/delta_catalog/namespaces//generic-tables/{generic-table}
+```
+
+### API Reference
+
+For the complete and up-to-date API specification, see the [generic-tables-api.yaml](https://github.com/apache/polaris/blob/main/spec/polaris-catalog-apis/generic-tables-api.yaml).
```
dimas-b: nit: maybe use a `swagger.io` reference as in #1879 ?

dimas-b: Ideally this should point to the same YAML version as the version of the doc (e.g. 1.0.0 vs. `main`).... not sure how to do it, though :thinking:

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+A generic table in Polaris is a structured entity that defines the necessary information. Each 
+generic table entity contains the following properties:
+
+- **name** (required): A unique identifier for the table within a namespace
+- **format** (required): The format for the generic table, i.e. "delta", "csv"
+- **base-location** (optional): Table base location in URI format. For example: s3://<my-bucket>/path/to/table
+  - The table base location is a location that includes all files for the table
+  - A table with multiple disjoint locations (i.e. containing files that are outside the configured base location) is not compliant with the current generic table support in Polaris.
+  - If no location is provided, clients or users are responsible for managing the location.
+- **properties** (optional): Properties for the generic table passed on creation
```
dimas-b: Does the Spark plugin use any specific property names? If yes, it would be good to add a section for them:
* as an illustration,
* to avoid name clashes with other use cases.

dimas-b (commented):
Thanks for writing this doc, @gh-yzou ! I think it is very valuable for end users (even though the feature itself is still "beta").

site/content/in-dev/unreleased/generic-table.md
eric-maynard: The client does that, not the server

site/content/in-dev/unreleased/generic-table.md
eric-maynard: Currently yes, if it was created via the IRC API

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+## Generic Table API Vs. Iceberg Table API
+
+Generic Table provides a different set of APIs to operate on the Generic Table entities while Iceberg APIs operates on
+the Iceberg Table entities.
+
+| Operations   | **Generic Table API**                                       | **Iceberg Table API**                                       |
+|--------------|-------------------------------------------------------------|-------------------------------------------------------------|
+| Create Table | create an Iceberg table                                     | create a generic table                                      |
+| Load Table   | load an Iceberg table, load a generic table throws an error | load a generic table, load an Iceberg table throws an error |
+| Drop Table   | drop an Iceberg table, drop a generic table throws an error | drop a generic table, drop an Iceberg table throws an error |
```
eric-maynard: I think the  note here should be like "drop a generic table, **dropping** an Iceberg table throws an error"

site/content/in-dev/unreleased/generic-table.md
eric-maynard: This looks like a minimal generic table, with no optional fields set? But it's not a great example, since the table created in the above section had optional fields populated.

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+For the complete and up-to-date API specification, see the [generic-tables-api.yaml](https://github.com/apache/polaris/blob/main/spec/polaris-catalog-apis/generic-tables-api.yaml).
+
+
+## Limitations
+
+The support for cross engine sharing of Generic Table is very limited:
+1) Limited spec information. Currently, there is no spec for information like Schema, Partition etc. 
+2) No commit coordination or update capability provided at the catalog service level.
+
+Today, all those responsibility will be at the client side. For example, the Delta log serialization, deserialization and update are all done at Client side.
```
eric-maynard:
```suggestion
Today, all of these responsibilities are on the client side. For example, Delta log serialization, deserialization and updates are all done on the Client side.
```

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+### API Reference
+
+For the complete and up-to-date API specification, see the [generic-tables-api.yaml](https://github.com/apache/polaris/blob/main/spec/polaris-catalog-apis/generic-tables-api.yaml).
+
+
+## Limitations
+
+The support for cross engine sharing of Generic Table is very limited:
```
eric-maynard:
```suggestion
The support for cross engine sharing of Generic Tables is very limited:
```

site/content/in-dev/unreleased/generic-table.md
dimas-b: Do you mean that the server does not "know" what such a table change? If so, it certainly deserves a dedicated paragraph :sweat_smile:  As for me, I tend to view the Generic Tables API as something similar to the Iceberg REST Catalog API, which does control commits and by extension conflict resolution on the server side.

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+title: Generic Table (Beta)
+type: docs
+weight: 435
+---
+
+The Generic Table in Apache Polaris provides basic management support for non-Iceberg tables.
```
snazy: What is a "generic table"?
What does "basic" and "management" mean here?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+
+The Generic Table in Apache Polaris provides basic management support for non-Iceberg tables. 
+
+With the Generic Table API, you can:
+- Create generic tables under a namespace
+- Load a generic table 
+- Drop a generic table
+- List all generic tables under a namespace
+
+**NOTE** The current generic table is in beta release, there can still be incomplete features and bugs. Please use it
```
snazy: IMO "beta" is not restricted to "features" and "bugs". "Beta" IMO means that is it not stable and can change fundamentally, breaking changes are likely to happen.

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+with caution and report any issue if encountered.
+
+## What is a Generic Table?
+
+A generic table in Polaris is a structured entity that defines the necessary information. Each
```
snazy: Is it "generic" or "structured"?

"necessary" for what?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+## What is a Generic Table?
+
+A generic table in Polaris is a structured entity that defines the necessary information. Each 
+generic table entity contains the following properties:
+
+- **name** (required): A unique identifier for the table within a namespace
+- **format** (required): The format for the generic table, i.e. "delta", "csv"
+- **base-location** (optional): Table base location in URI format. For example: s3://<my-bucket>/path/to/table
+  - The table base location is a location that includes all files for the table
+  - A table with multiple disjoint locations (i.e. containing files that are outside the configured base location) is not compliant with the current generic table support in Polaris.
```
snazy, line 41: This prevents leveraging "object store friendly paths", no?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,168 @@
...
+### API Reference
+
+For the complete and up-to-date API specification, see the [generic-tables-api.yaml](https://github.com/apache/polaris/blob/main/spec/polaris-catalog-apis/generic-tables-api.yaml).
+
+
+## Limitations
+
+The support for cross engine sharing of Generic Table is very limited:
+1) Limited spec information. Currently, there is no spec for information like Schema, Partition etc. 
+2) No commit coordination or update capability provided at the catalog service level.
```
snazy, line 164: To me this is a very serious issue. No way to coordinate changes means there will be consistency issues.

snazy (commented):
This is a good start.

I've phrased some of my comments as questions as a reader of the doc.

Generally, the phrasing needs to be more precise and terms need to be explained before being used/referred to.

What I'm missing is the reason for "generic", because the only implemented use case is very specific. I.e. the text is missing an explanation how other formats would/could be represented. The reference of "structured" implies an explicit exclusion of "semi/unstructured" data.

The doc should IMO also describe the various interactions, edge cases and failure scenarios from an client integration's  view.

I was not aware of `No commit coordination or update capability provided at the catalog service level.`. This is IMHO a very serious issue, because it means that there is absolutely no guarantee that the state is consistent. Older changes can overwrite newer changes (ordering of request executions). This means (data) consistency issues.

adutra:
> What I'm missing is the reason for "generic", because the only implemented use case is very specific

There is some prior discussion around this naming choice in the dev ML:

https://lists.apache.org/thread/9jcx656ybkn132qw94g5wh8n5nmkg1d9

snazy:
> There is some prior discussion around this naming choice in the dev ML:

yup - but we cannot expect readers to know the whole dev-ML history

site/content/in-dev/unreleased/generic-table.md
RussellSpitzer: Not all formats even have a way to do transactional commits. The basic premises here is to behave like the Spark Catalog with HMS (or Unity) which have these same guarantees for any sources. 

For example registering a Cassandra would work but there is nothing in the Polaris world that would (or could) manage commits for a C* source.  

Another example would be Delta Lake, which only can optionally (in 4.0) use a Catalog based commit coordinator and usually does not even consider the catalog when making commits. 

Polaris is only guaranteeing a consistent view of the metadata about the entity, not any guarantees about the underlying data.

site/content/in-dev/unreleased/generic-table.md
RussellSpitzer: This is much closer to the Spark "HMS" catalog integration. The catalog itself is unaware of the anything about the underlying table except for some loosely defined metadata about it. It's up to the engine (and plugins in that engine) to determine exactly how loading data or committing data actually occur based on that metadata. 

You could imagine examples of use cases being something like a CSV based table, or a JDBC Table. When these are stored in the HMS by Spark, the HMS doesn't know how to actually interact with the metadata.

dimas-b: I do not challenge this mode of operation from a technical POV. I mean that it was a surprise for me and might be a surprise to other people coming from the Iceberg REST Catalog side. I'd appreciate if this aspect were discussed in more details in this doc.

site/content/in-dev/unreleased/generic-table.md
gh-yzou: I just meant the operations listed below, i removed those unclear wording, and just point to the list of operations

site/content/in-dev/unreleased/generic-table.md
gh-yzou: sorry for the confusion, I removed the wording of features and bugs. This is just a beta feature.

site/content/in-dev/unreleased/generic-table.md
dimas-b: This example shows that there is a Delta table named `delta_table`, but what can the reader/user do with this information other than avoid creating other tables with the same name?

site/content/in-dev/unreleased/generic-table.md
gh-yzou: It should be just an entity with defined fields. i removed those unnecessary wording

site/content/in-dev/unreleased/generic-table.md
dimas-b: Please consider explaining this in more details. As for me I had to read this table a few times to try and understand the intended behaviour :sweat_smile:

site/content/in-dev/unreleased/generic-table.md
gh-yzou: Do you mean volume usage? I think how volume is going to be supported in Polaris is currently not discussed yet. Since this is a beta feature, if we decided to support the use cases with mutilple locations, we can evolve quickly to support this.

site/content/in-dev/unreleased/generic-table.md
gh-yzou: Yes, sorry for the mistake.

site/content/in-dev/unreleased/generic-table.md
gh-yzou: updated. i also uses `.` in between to avoid potential confusion to user

site/content/in-dev/unreleased/generic-table.md
gh-yzou: Good point, i updated the wording to `using tools such as curl`

site/content/in-dev/unreleased/generic-table.md
gh-yzou: updated

site/content/in-dev/unreleased/generic-table.md
gh-yzou: updated

site/content/in-dev/unreleased/generic-table.md
gh-yzou: updated the link to the swagger.io, but seems we only have the catalog bundle yaml, so updated the text to Catalog API Spec also.

site/content/in-dev/unreleased/generic-table.md
gh-yzou: Thanks for pointing that out, move it out to shell format

site/content/in-dev/unreleased/generic-table.md
gh-yzou: yes

site/content/in-dev/unreleased/generic-table.md
gh-yzou: Do you mean does spark plugin actually do transfer of specific property it receives from Spark and convert to a reserved key name ? Today, we do not do this. All properties a defined by Spark, and Spark have the right to update any of the property keys, which i don't think it is a good idea to doc it here for Generic tables

site/content/in-dev/unreleased/generic-table.md
gh-yzou: sorry, I was just trying to add a very simple example for illustration, i updated the example to have a more concrete example. 
However, today, the kind of guarantee that the catalog provides is limited today, a lot of the responsibility are just on client, includes the contact of the properties passed. I will add more description in the limitation section

site/content/in-dev/unreleased/generic-table.md
gh-yzou: I will add more description in the limitation section.

site/content/in-dev/unreleased/generic-table.md
gh-yzou: added more description, hopefully it is more clear now

site/content/in-dev/unreleased/generic-table.md
dimas-b: I'm afraid, I'm a bit lost here :sweat_smile: 

I mean: do we know what specific properties are currently set/retrieved from this `properties` list by Polaris code on the client or server side?

Is the `properties` property used by any code now (apologies that I did not review the Spark Client end-to-end)?

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,178 @@
...
+  "format": "<table_format>",
+  "base-location": "<table_base_location>",
+  "doc": "<comment or description for table>",
+  "properties": {
+    "<property-key>": "<property-value>"
+  }
+}
+```
+
+Here is an example to create a generic table with name `delta_table` and format as `delta` under a namespace `delta_ns`
```
dimas-b, line 88: Since there's no "update" API, what happens if there's a mistake in the initial create request? WIll the client have to delete and re-create the table?

I do not mean to cause API changes at this point, just trying to clarity things for potential non-Polaris readers.

site/content/in-dev/unreleased/generic-table.md
dimas-b: thx for the fix :+1:

site/content/in-dev/unreleased/generic-table.md
dimas-b: thx :+1:

site/content/in-dev/unreleased/generic-table.md
dimas-b: ping: the format issue is still there :)

site/content/in-dev/unreleased/generic-table.md
dimas-b: thx :+1:

site/content/in-dev/unreleased/generic-table.md
dimas-b: Thx for the update, @gh-yzou !

dimas-b:
Thanks again for making this doc page, @gh-yzou ! I think I'm done with comments from my side. I'd be fine with merging this PR. Not approving only to allow other reviewers to have another round of comments.

site/content/in-dev/unreleased/generic-table.md
snazy: > Not all formats even have a way to do transactional commits.

Delta has, no?

> Polaris is only guaranteeing a consistent view of the metadata about the entity, not any guarantees about the underlying data.

How is a _consistent_ view on metadata **ensured**?

snazy: Mean, [this blog post](https://www.snowflake.com/en/engineering-blog/apache-polaris-supports-iceberg-delta-lake/) mentions `A data catalog serves as the central registry for a table’s metadata. It manages transactions and table state, as well as access controls and read/write interoperability.`

RussellSpitzer: Because the metadata only exists in polaris. Only this set of properties.

> > Not all formats even have a way to do transactional commits.
> 
> Delta has, no?
> 

Delta does without using the catalog, and has an optional "commit coordinator" which uses another api which is not provided here. So like users of HMS for a delta table they would need to use a third party commit coordinator if they wanted to use the optional "commit coordinator"

> > Polaris is only guaranteeing a consistent view of the metadata about the entity, not any guarantees about the underlying data.
> 
> How is a _consistent_ view on metadata **ensured**?

I think you may want to check out the original design docs here. The "metadata" we are talking about here is referring to what the user puts in their Create statement that talks to Polaris. That is the only thing Polaris knows about the table and is the part that will not change. Again, this is similar to how the HMS works with Spark or how Iceberg originally worked with the HMS Catalog implementaiton. The Catalog is essentially just holding a bag of properties we we will maintain. Changes to these properties (not yet allowed) would be atomic but they are essentially disconnected from the underlying format.

snazy: So if basically only the table name is there, what's a user's benefit for it?

RussellSpitzer: A user is tying a table-identifier to a bucket of properties which an engine can use to resolve and read a table.

For example, Let's say I want to make a Delta Lake table in Spark with Unity, or HMS, or whatnot. The Delta catalog plugin will essentially bundle up a set of properties that say Table X is at location Y and potentially has other properties that should be used when accessing this table (Split sizes, compute behaviors etc ...) . That is packed up and put in a Catalog be in HMS or Unity or whatnot. A user when using SQL can reference the Identifier and the Catalog Plugin within Spark will pull down the set properties and use them to initialize a Spark Format Specific plugin to use those properties. In Delta's case it will load up the Delta Lake plugin and use the properties in the Catalog as the Datasource Reader properties. The datasource will then load a Spark Table object and return it to the user (or use it in a query). In the case of Delta the only properties actually relating to the table's state are the "Location" and "format", everything else is essentially re-loaded from the delta log found at the Location.

If instead you had a Cassandra entity, it would have properties format="cassandra" and then a bunch of properties which would dictate connection parameters for the C* cluster. Actual information about the "state" of the table other than it's connection format can only be read by loading the appropriate plugin and using the properties to connect to the cluster. 

This is model used by Spark for interacting with all Spark sources that don't have an independent V2 Catalog implementation (and actually works for those with a V2 Catalog Implementation as well as long as they also have a V2 source register implementation). 

So the user benefit of Polaris providing this is essentially allowing Polaris to be used as a generic Spark catalog and gaining support for all DSv1 and most DSv2 sources without having an additional store for a Spark cluster.

site/content/in-dev/unreleased/generic-table.md
eric-maynard: Currently, yes. I think adding update is a good idea but the intent here should be to document the existing behavior.

IIRC the motivation to not have update in v0 was due to a potential lack of clarity around what responsibilities the catalog takes on for updates (i.e. it's not the same as an Iceberg update where the catalog writes metadata).

site/content/in-dev/unreleased/generic-table.md
eric-maynard: > So if basically only the table name is there, what's a user's benefit for it?

I'm confused -- are we cross-examining the feature here or documenting it?

The benefit is as is documented here; you can use Delta and other non-Iceberg tables in Spark using the Spark connector. The doc walks you through how that works. If that benefit is unclear in the doc, let's fix that.

site/content/in-dev/unreleased/generic-table.md
eric-maynard: I think the answer to your question is no, it's not used. Actually per #1785 we don't propagate the properties from client to server, which seems incorrect to me. But it does mean there's not some special property to call out here.

gh-yzou: so today, the polaris doesn't set/retrieve properties at server side. At client side today, we do retrive the "provider" and "location" property at spark client side and then translate it to the format and location. However, this is more of a contract between spark and spark client, which may not be suitable to mention in the generic table webpage. I can mention that our Polaris spark client today looks into the table property and translate the provider and location into our catalog format and location, but i don't think we want to make it like the property has to contain fields like "provider" or "location" from spark.

dimas-b: Suppose a Generic Table named `A` exists and has some properties usable from Spark via our new Spark Client. Now, if another client wants to query table `A` via the Generic Tables API, how should that (new) client interpret existing table properties?

The new client may not be aware of the Spark Client, but the properties are observable via API. I believe this aspect of the Generic Tables API ought to be clarified in this doc.

site/content/in-dev/unreleased/generic-table.md
gh-yzou:
```
So if basically only the table name is there, what's a user's benefit for it?
```
As @RussellSpitzer mentioned, the milestone polaris accomplishes today is enable polaris as a centralized catalog service for Spark Catalog. Furthermore, for delta, the state is inside the delta log, as far as the client is able to load the delta log, "base-location" is the only information needed to enable access to the delta table. I can try to make it more clear in the doc.

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+### API Reference
+
+For the complete and up-to-date API specification, see the [Catalog API Spec](https://editor-next.swagger.io/?url=https://raw.githubusercontent.com/apache/polaris/refs/heads/main/spec/generated/bundled-polaris-catalog-service.yaml).
+
+
+## Limitations
+
+The Generic Table support today is very limited:
```
flyrain: Nit: can we reword a bit like this? 
```suggestion
Current limitations of Generic Table support:
```

flyrain (approved):

site/content/in-dev/unreleased/generic-table.md
eric-maynard: Isn't it the same for Iceberg tables? There may be some property on a table that my Spark job knows how to interpret, but yours doesn't. The REST catalog itself does not take any actions based on generic table properties currently.

dimas-b: True, but the REST catalog spec is not owned / defined by Polaris, while this one is.

I do not insist on a complete enumeration of possibilities. I think it should be sufficient to make a broad statement, for example: at this time, the contents of the `properties` map are not strictly defined and their interpretation is delegated to client / engine implementations, including interoperability concerns.

In fact, this is what happens with Iceberg tables too, IMHO, but I believe it is valuable to be explicit in specs.

gh-yzou: sounds good! i added the following description
```
  - Currently, there is no reserved property key defined.
  - The definition and interpretation is delegated to client or engine implementations.
```

site/content/in-dev/unreleased/generic-table.md
gh-yzou: oh, sorry. i moved it as plain text with ``, and I also added an actual example with curl call.

site/content/in-dev/unreleased/generic-table.md
gh-yzou: sorry for the late reply! as we don't have update capability today, rename will require a re-creation, I added some    description at section Generic Table API Vs. Iceberg Table API

site/content/in-dev/unreleased/generic-table.md
gh-yzou: updated

gh-yzou:
@snazy When we come with the name generic table, the intension is to evolute the support across different table formats, I added some more description at the top to help make the naming more clear. The whole feature is currently marked as beta, which is indicating things are still evolving.

site/content/in-dev/unreleased/generic-table.md
dimas-b: thx - sgtm

site/content/in-dev/unreleased/generic-table.md
```diff
@@ -0,0 +1,169 @@
...
+# under the License.
+#
+title: Generic Table (Beta)
+type: docs
+weight: 435
+---
+
+The Generic Table in Apache Polaris is designed to provide support for non-Iceberg tables across different table formats includes delta, csv etc. It currently provides the following capabilities:
+- Create a generic table under a namespace
+- Load a generic table
```
dimas-b, line 27: nit: the term "load", given other conversations under this PR, is still a bit confusing, IMHO, because it resonates with Iceberg's `loadTable`, which provides tables's metadata... However, this call is more like "get properties". All in all, I think it's ok since subsequent doc sections provide more clarity.

site/content/in-dev/unreleased/generic-table.md
snazy: What I really do not like is implying promises (`A data catalog serves as the central registry for a table’s metadata. It manages transactions and table state, as well as access controls and read/write interoperability.`) which just do not hold true: there is no way to manage transactions and table state, control access, etc.

WRT to what's in Polaris - it's incomplete (I suspect there's no doubt there).

site/content/in-dev/unreleased/generic-table.md
eric-maynard: The API itself is actually named loadGenericTable, so I don't think it's exactly misleading. It does load the generic table's metadata, which is whatever metadata was registered in the catalog during createGenericTable. This is very similar to the behavior of, say, the HMS's `getTable`. Actually cracking open the metadata.json and returning its contents in the IRC is the exception, not the rule.

site/content/in-dev/unreleased/generic-table.md
RussellSpitzer: I don't actually think it's incomplete in that context? I wouldn't imagine we would ever support those things for this endpoint and I don't think it would be a surprise to any user of Spark who uses this?

site/content/in-dev/unreleased/generic-table.md
RussellSpitzer: This is a confusing one but to match behavior of other systems with similar functionality I think load or get is probably correct

dimas-b:
This PR probably needs rebasing to catch up with CI changes (even though it's only a doc change)

site/content/in-dev/unreleased/generic-table.md
gh-yzou: yeah, i was mainly try to matching the term of other systems to use "load or get", and I don't think load has to be specifically to the iceberg metadata.

gh-yzou:
@dimas-b Thanks! i just rebased

site/content/in-dev/unreleased/generic-table.md
dimas-b: I agree with @snazy 's point that the blog post (link above) is not really aligned with the Generic Tables API as far as Catalog as a "central registry" for table's metadata is concerned. The Generic Table API actually diminishes the role of the catalog as metadata registry by delegating most of the metadata loading to the client (even `location` is optional in the API).

That said, as far as this PR is concerned, I believe it should be sufficient to describe actual behaviour of Polaris in this respect. There is certainly room for improvements in terms of clarity and precision in this doc, but I think the current state of this PR is probably acceptable for 1.0.

flyrain (approved):

PR #1890, opened 2025-06-13
docs: Add `Polaris Evolution` page

dimas-b:

### DIFF
--- a/site/content/in-dev/unreleased/evolution.md
+++ b/site/content/in-dev/unreleased/evolution.md
@@ -102,7 +102,7 @@ in the previous release is not considered a major change.
 
 * Supporting a new revision of the Iceberg REST Catalog spec in a backward-compatible way
 is not considered a major change. Specifically, supporting new REST API prefixes (e.g. `v2`)
-is not a mojor change because it does not affect older clients.
+is not a major change because it does not affect older clients.
 
 * Changing the implementation of an Iceberg REST Catalog feature / endpoint in a non-backward
 compatible way (e.g. removing or renaming a request parameter) is a major change.


site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,61 @@
...
+
+In case a major change is required to the Management API that cannot be implemented in a
+backward-compatible way, a new version of the API will be introduced. Polaris servers will support
+old API version for some transition period to allow clients to migrate. Note that API versions
+are not meant to be 1:1 with Polaris releases.
+
+The Generic Tables API is an experimental API at this time. Refer to its documentation page
+for more details about compatibility expectations.
+
+## Polaris as a Library
```
adutra: This doc is a great start! For the "Polaris as a Library" section, I am wondering:

* Should we distinguish guarantees offered by different modules? E.g.  should `polaris-core` or any of the API modules offer stronger backwards compatibility guarantees than persistence or service modules?
* A statement like `Maintainers try to keep binary compatibility on the "best effort" basis` effectively contradicts Semver guarantees; it could be good to explicitly define our relationship with Semver here, wdyt? As most people implicitly infer semver when the versioning scheme is not explicitly stated.
* Should we also mention Java version guarantees? Currently `polaris-core`, the API modules, the Spark plugin modules, and some tooling modules all require Java 11, whereas the service modules, the Quarkus modules, the persistence modules and a few tooling modules require Java 21. It's not clear if these requirements are allowed to evolve, and how.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,61 @@
...
+
+This page discusses what can be expected from Apache Polaris as the project evolves.
+
+## Polaris as a Catalog
+
+Polaris is primarily intended to be used as a Catalog of Tables and Views. As such, 
+it implements the Iceberg REST API, Polaris Management API and the Generic Tables API.
+
+Revisions of the Iceberg REST API are controlled by the [Apache Iceberg](https://iceberg.apache.org/)
+community. Polaris attempts to accurately implement the most recent version of this specification,
```
snazy:
```suggestion
community. Polaris attempts to accurately implement this specification,
```
I think, "most recent" is a too strong commitment. There might be legit reasons to not implement the "latest greatest". WDYT?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,61 @@
...
+which is available at the time of Polaris releases.
+
+The Polaris Management API is maintained as a versioned REST API. New releases of Polaris may
+include changes to the current version of the Management API. When that happens those changes
+are intended to be compatible with prior versions of Polaris clients.
```
snazy: > `compatible with prior versions`

Worth mentioning deprecation-for-removal?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,61 @@
...
+
+In case a major change is required to the Management API that cannot be implemented in a
+backward-compatible way, a new version of the API will be introduced. Polaris servers will support
```
snazy: > `a new version of the API will be introduced`

It's also a very strong commitment. It _may_ be a new version or a deprecation or addition.

"Version" can be ambiguous here IMO. It might be a new version of the spec or a new base path (e.h. `.../api/v42/...`) or both. Mean, it's fine to be vague here, but then it should be clear that it's vague. WDYT?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,61 @@
...
+old API version for some transition period to allow clients to migrate. Note that API versions
+are not meant to be 1:1 with Polaris releases.
+
+The Generic Tables API is an experimental API at this time. Refer to its documentation page
+for more details about compatibility expectations.
```
snazy: Not sure we should refer to specific APIs from a general doc.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Polaris as a Library
+
+Polaris produces several jars. These jars or custom builds of Polaris code may be used in
+downstream projects according to the terms of the license included into Polaris distributions.
+
+However, Polaris' focus as a project is not on ensuring ideal binary compatibility at the java
```
snazy: Nit: "ABI" is the term here. But "binary" is much stronger than "source".

site/content/in-dev/unreleased/evolution.md
snazy:
```suggestion
## Polaris is not a Library
```

The Polaris project is primarily intended for consumption "as is" by end users, while also allowing people to integrate it into their services. But there are no clear boundaries between API/SPI and implementation. Those could be introduced by having a clear separation of these via specific modules - but that's not the case yet.

Binary compatibility (in the sense of "ABI", application binary interface) is a very strong commitment, much stronger than source compatibility (think: rebuild).

From experience, we already had quite some changes throughout the (short) history of the project that count as ABI breaking changes, and quite a few exist as PRs (in progress/review) for "1.x".

My take on this is that there is strictly speaking neither a binary nor a source level compatibility guarantee. We should however work towards (more or less) strict API/SPI and implementation boundaries and clearly state the intent, scope and compatibility guarantees for each of the APIs and SPIs are.

Java visibility modifiers are not a way to "declare" APIs/SPIs and their guarantees. `public`, `protected` et al just define the technical visibility, often a broad(er) visibility is required by pure technical reasons - but those are not (necessarily) "public stable APIs". `@VisibleForTesting` means that the technical visibility for something has been changed for testing purposes - but it says absolutely nothing about the "assumed production visibility" ; it's also just an annotation, nothing that's enforced, similar to other annotations.

dimas-b: I agree with this comment, but I'd like to keep the original section title :) Once way or another Polaris jars / code will be used downstream so in this section we're just setting expectations for these workflows.

dimas-b: Will clarify later today.

From my POV it is too early to marks some modules as "API" with stronger guarantees than other module. I think we should strive for that, but in the current state of the project, I think, all code is equally subject to major changes.

Attaching semver semantics to code at this time will probably require a version sequence like 1.0, 2.0, 3.0 for a few releases at least -- this is something I am personally fine with, but I wonder what other people think.

site/content/in-dev/unreleased/evolution.md
eric-maynard: This is a good time to resolve this ambiguity maybe. I have often wondered under what conditions we'll cut a new version of the spec.

site/content/in-dev/unreleased/evolution.md
eric-maynard: Maybe we could have some kind of feature overview or roadmap that gives information like this, and then link it here? That could be a followup

site/content/in-dev/unreleased/evolution.md
eric-maynard: `Polaris as a Library` reads almost like `Polaris is a Library` to me... Maybe `Building Dependencies on Polaris` or `Using Polaris as a Dependency` or some variation?

site/content/in-dev/unreleased/evolution.md
snazy: Classic answer is probably: it depends. :shrug:

site/content/in-dev/unreleased/evolution.md
flyrain: That's a valid concern. Can we remove the second sentence completely? We may not document it if we don't have a formal strategy to update IRC in Polaris.

dimas-b: updated (slightly different text)

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Polaris as a Catalog
+
+Polaris is primarily intended to be used as a Catalog of Tables and Views. As such, 
+it implements the Iceberg REST API, Polaris Management API and the Generic Tables API.
+
+Revisions of the Iceberg REST API are controlled by the [Apache Iceberg](https://iceberg.apache.org/)
+community. Polaris attempts to accurately implement this specification. However, there
+is no guarantee that Polaris releases always implement the latest version of the Iceberg
+REST Catalog API.
```
flyrain:
```suggestion
community. 
```
Is it beneficial to mention an ad-hoc approach here? Does this "Polaris attempts to accurately implement this specification." imply Polaris isn't 100% compatible with IRC? I'd suggest to just remove it.

site/content/in-dev/unreleased/evolution.md
dimas-b: updated / rephrased and expanded a bit

site/content/in-dev/unreleased/evolution.md
dimas-b: removed this paragraph

site/content/in-dev/unreleased/evolution.md
dimas-b: changed title to `Using Polaris as a Build-Time Dependency` ... WDYT?

site/content/in-dev/unreleased/evolution.md
dimas-b: What is ad-hoc here? Accurately implementing a particular version of a spec does not imply implementing the highest version of the spec. I believe it is valuable to inform users that Polaris as a project does not guarantee implementing the latest version of the Iceberg REST Catalog API.

site/content/in-dev/unreleased/evolution.md
dimas-b: updated and added semver

site/content/in-dev/unreleased/evolution.md
flyrain: > there is no guarantee that Polaris releases always implement the latest version of the Iceberg
REST Catalog API.

This means users don't know when Polaris will follow the latest version of the Iceberg REST Catalog API, which is true. The Polaris community decide when to update. But why do we mention it? What value do users get from this message?

> Polaris attempts to accurately implement this specification.

This means Polaris may not be 100% compatible with Iceberg REST spec. I think it's untrue, which may confuse users.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,93 @@
...
+Polaris servers will support deprecated API endpoints / parameters / versions / etc. 
+for some transition period to allow clients to migrate.
+
+### Managing Polaris Database
+
+Polaris stores its data in a database, which is sometimes referred to as "Meta Store" or
+"Persistence" in other docs.
+
+Several Persistence implementations may be supported by Polaris in each release.
+For example: "EclipseLink" (deprecated) and "JDBC" (current).
```
singhpk234: [optional] 
```suggestion
Each Polaris release may support multiple Persistence implementations—for example, "EclipseLink" (now deprecated) and "JDBC" (currently in use).
```

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,93 @@
...
+
+Each type of Persistence evolves individually. Within each Persistence type, Polaris
+attempts to support rolling upgrades (both version X and X + 1 servers running at the
+same time).
+
+However, users should not expect that a transition from "EclipseLink" to "JDBC" can be
+done in a rolling upgrade manner. Polaris provides tools for migrating between different
```
singhpk234: a link to polaris-tools repo would be really helpful

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,93 @@
...
+"Persistence" in other docs.
+
+Several Persistence implementations may be supported by Polaris in each release.
+For example: "EclipseLink" (deprecated) and "JDBC" (current).
+
+Each type of Persistence evolves individually. Within each Persistence type, Polaris
+attempts to support rolling upgrades (both version X and X + 1 servers running at the
+same time).
+
+However, users should not expect that a transition from "EclipseLink" to "JDBC" can be
```
singhpk234: how about this, 
```suggestion
However, migrating between different Persistence types—such as from "EclipseLink" to "JDBC"—is not supported as a rolling upgrade
```
just wanted on re-emphasize on persistent types

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,93 @@
...
+
+Polaris produces several jars. These jars or custom builds of Polaris code may be used in
+downstream projects according to the terms of the license included into Polaris distributions.
+
+Polaris strives to follow [Semantic Versioning](https://semver.org/) conventions both with
+respect to java code and REST APIs.
+
+However, Polaris' focus as a project is not on ensuring ideal binary compatibility at the java 
+class level between releases. Maintainers try to keep binary compatibility on the "best effort"
+basis. The main goal being the implementation of Polaris capabilities for access via REST APIs.
```
snazy: Let's remove the "binary compat" sentence entirely.

site/content/in-dev/unreleased/evolution.md
dimas-b: > What value do users get from this message?

Avoiding false expectation that a Polaris release on Jun 16, 2025 (hypothetical example) supports Iceberg REST Catalog API as it is defined in Iceberg v. `1.9.1`.

dimas-b: I'll rephrase the "attempts" part to carry a stronger commitment.

site/content/in-dev/unreleased/evolution.md
dimas-b: removed

site/content/in-dev/unreleased/evolution.md
dimas-b: updated

site/content/in-dev/unreleased/evolution.md
dimas-b: Thx! I agree that active voice is preferable, but is often hard for me personally :sweat_smile:

site/content/in-dev/unreleased/evolution.md
dimas-b: added

site/content/in-dev/unreleased/evolution.md
dimas-b: updated

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+catalogs and those tools may be used to migrate between different Persistence types
+as well. Service interruption (downtime) should be expected in those cases.
+
+## Using Polaris as a Build-Time Dependency
+
+Polaris produces several jars. These jars or custom builds of Polaris code may be used in
+downstream projects according to the terms of the license included into Polaris distributions.
+
+Polaris strives to follow [Semantic Versioning](https://semver.org/) conventions both with
+respect to java code and REST APIs.
```
eric-maynard: super nit: `Java`

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+
+This means that major version increments should be expected often (in the SemVer sense).
+
+The minimal version of the JRE required by Polaris code (compilation target) may be updated in
+any release. Different Polaris jars may have different minimal JRE version requirements.
+
+This approach is not meant to discourage the use of Polaris code in downstream projects, but
+to allow more flexibility in evolving the codebase to support new Catalog-level features
```
eric-maynard: super nit: `catalog-level`

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+type: docs
+weight: 1000
+---
+
+This page discusses what can be expected from Apache Polaris as the project evolves.
+
+## Using Polaris as a Catalog
+
+Polaris is primarily intended to be used as a Catalog of Tables and Views. As such, 
+it implements the Iceberg REST API, Polaris Management API and the Generic Tables API.
```
eric-maynard: It might be nice to link here.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+Polaris servers will support deprecated API endpoints / parameters / versions / etc. 
+for some transition period to allow clients to migrate.
+
+### Managing Polaris Database
+
+Polaris stores its data in a database, which is sometimes referred to as "Meta Store" or
+"Persistence" in other docs.
+
+Each Polaris release may support multiple Persistence implementations,
+for example, "EclipseLink" (deprecated) and "JDBC" (current).
```
eric-maynard, line 61: again it might be nice to link out

eric-maynard (approved):
Nice, this doc is a great idea

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+title: Polaris Evolution
+type: docs
+weight: 1000
```
eric-maynard, line 22: ~~IIUC this puts the doc at the way top, above the admin tool and potentially above getting started. Is that intended?~~ I had it backwards, sorry.

Also, it might be good to link to this doc in another doc like `Production Configuration`

dimas-b: Cross-linking makes sense. Let's this one after this PR is settled :)

site/content/in-dev/unreleased/evolution.md
dimas-b: I thought about linking to https://polaris.apache.org/in-dev/unreleased/polaris-management-service/, but  that page auto-redirects to swagger, which I personally do not like... but I do not mind adding this link if you prefer. WDYT?

site/content/in-dev/unreleased/evolution.md
dimas-b: linked to [../metastores](https://polaris.apache.org/in-dev/unreleased/metastores/)

site/content/in-dev/unreleased/evolution.md
dimas-b: fixed

site/content/in-dev/unreleased/evolution.md
dimas-b: fixed

site/content/in-dev/unreleased/evolution.md
dimas-b: Actually, I added a section to the "prod config" page that links here.

site/content/in-dev/unreleased/evolution.md
eric-maynard: Hm, good point. Maybe it's okay to route to Swagger for now since we're specifically mentioning the API here?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+catalogs and those tools may be used to migrate between different Persistence types
+as well. Service interruption (downtime) should be expected in those cases.
+
+## Using Polaris as a Build-Time Dependency
+
+Polaris produces several jars. These jars or custom builds of Polaris code may be used in
+downstream projects according to the terms of the license included into Polaris distributions.
+
+Polaris strives to follow [Semantic Versioning](https://semver.org/) conventions both with
+respect to Java code and REST APIs.
```
flyrain: I think the REST APIs have its own versioning, which doesn't follow semantic versioning. As you put in above, the REST APIs versioning only has `v1`, `v2`, there is no minor versions like `v1.1.0` for the REST API.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+[tools](https://github.com/apache/polaris-tools/) for migrating between different
+catalogs and those tools may be used to migrate between different Persistence types
+as well. Service interruption (downtime) should be expected in those cases.
+
+## Using Polaris as a Build-Time Dependency
+
+Polaris produces several jars. These jars or custom builds of Polaris code may be used in
+downstream projects according to the terms of the license included into Polaris distributions.
+
+Polaris strives to follow [Semantic Versioning](https://semver.org/) conventions both with
```
flyrain: > For this system to work, you first need to declare a public API. This may consist of documentation or be enforced by the code itself. Regardless, it is important that this API be clear and precise. Once you identify your public API, you communicate changes to it with specific increments to your version number.

The semver requires a project maintainers to declare public APIs. We will still need to identify which are public APIs. Are all java public classes/methods are considered as public APIs in Polaris? Or just applies to a few modules. I believe we still still be clear about that.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+respect to Java code and REST APIs.
+
+This means that major version increments should be expected often (in the SemVer sense).
```
flyrain: I don't think following semver means Polaris will update major versions often. At this moment, we don't know if the major version will be updated often. I'd suggest to not say a thing that we don't know when will happen.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+
+The minimal version of the JRE required by Polaris code (compilation target) may be updated in
+any release. Different Polaris jars may have different minimal JRE version requirements.
+
+This approach is not meant to discourage the use of Polaris code in downstream projects, but
+to allow more flexibility in evolving the codebase to support new catalog-level features
```
flyrain, line 85: I'm a bit confused by `This approach` here. Semver doesn't discourage the downstream dependency on code/jars, it's actually widely used by a lot of projects to allow dependencies, e.g., Jackson, Apahce Iceberg, etc. What's the approach we meant here? Do we only consider REST APIs as public APIs? Appreciate if you can elaborate a bit.

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,88 @@
...
+
+This page discusses what can be expected from Apache Polaris as the project evolves.
+
+## Using Polaris as a Catalog
+
+Polaris is primarily intended to be used as a Catalog of Tables and Views. As such, 
+it implements the Iceberg REST API, Polaris Management API and the Generic Tables API.
+
+Revisions of the Iceberg REST API are controlled by the [Apache Iceberg](https://iceberg.apache.org/)
+community. Polaris accurately implements this specification. However, there
```
eric-maynard: Sorry, I know that we've gone back and forth on this wording a lot. Can we try `Polaris implements this specification, but there is no guarantee ...` to make this flow a little better?

I feel like `Polaris accurately...` doth protest too much. I actually thought that `attempts to` was fine, since after all we don't actually implement every API right now. We just try to :)

flyrain:
Can we also mention the policy schema as well? For example, here is the data compaction policy schema: https://polaris.apache.org/schemas/policies/system/data-compaction/2025-02-03.json. It's not a blocker though, we can add them as a followup.

site/content/in-dev/unreleased/evolution.md
eric-maynard: I think this means the approach of having frequent breaking changes isn't explicitly intended to discourage the use of Polaris, but to allow ...

I agree with this sentiment, but I think the comments above about a need to clearly define which APIs are actually considered public under SemVer are fair.

site/content/in-dev/unreleased/evolution.md
eric-maynard: I feel like the spirit of the message is good, which is basically that we don't want to promise long-term stability of APIs. This implies, if those APIs are public under SemVer, that we can't promise there won't be major releases happening pretty often. But again this connects to the thread about what is exactly covered under our SemVer guarantees.

flyrain: I'm OK if we would agree to update the major version frequently. Putting it here a bit premature considering we don't have the process yet. Maybe we could add it back once we have it?

dimas-b: We can certainly refine the definition of the "API" and what is "public" as we produce more releases. I tend to think we need to set a baseline for 1.0 anyway... will try to clarity shortly.

dimas-b:
> Can we also mention the policy schema as well?

I'm not sure I have enough context to describe compatibility / evolution for policies. Please add that as a follow-up.

site/content/in-dev/unreleased/evolution.md
dimas-b: rephrased back to "attempts to" and added a note about optional APIs... Is it better? :sweat_smile:

site/content/in-dev/unreleased/evolution.md
dimas-b: expanded on SemVer. PTAL.

site/content/in-dev/unreleased/evolution.md
dimas-b: updated

site/content/in-dev/unreleased/evolution.md
dimas-b: Good catch :+1:  Updated to describe "API surface".

site/content/in-dev/unreleased/evolution.md
dimas-b: link added :+1:

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,121 @@
...
+* Iceberg REST Catalog API and Generic Tables API (refer to this [link](../polaris-catalog-service/)
+for their combined Open API definition).
+  * Note: Polaris implementing an optional Iceberg REST Catalog feature that was unimplemented
+  in the previous release is not considered a major change.
+  * Supporting a new revision of the Iceberg REST Catalog spec in a backward-compatible way
+  is not considered a major change.
+  * Changing the implementation of an Iceberg REST Catalog feature / endpoint in a non-backward
+  compatible way is a major change.
+* [Polaris Management API](../polaris-management-service/)
+* [Polaris Policies](http://localhost:1313/in-dev/unreleased/policy/)
```
flyrain: Can we make it either a relative path or a path pointing to the website? And this needs both REST APIs and policy schemas.

flyrain: Again, as I said here, https://github.com/apache/polaris/pull/1890#discussion_r2150806575. I don't think the REST APIs follow Semver. Can we remove them here?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,121 @@
...
+This approach is not meant to discourage the use of Polaris code in downstream projects, but
+to allow more flexibility in evolving the codebase to support new catalog-level features
+and improve code efficiency. Maintainers of downstream projects are encouraged to join Polaris 
+mailing lists to monitor project changes, suggest improvements, and engage with the Polaris
+community in case of specific compatibility concerns.   
+
+## Semantic Versioning
+
+Polaris strives to follow [Semantic Versioning](https://semver.org/) conventions both with
+respect to Java code and REST APIs.
```
flyrain: Same comment here that REST APIs doesn't follow semver.

site/content/in-dev/unreleased/evolution.md
eric-maynard: We have a generalized problem with linking across docs... Linking like this used to work when these were just mds, but now I think the best practice is Hugo-style links to web pages?

dimas-b: oops, it's just a copy-paste mistake :sweat_smile: fixed.

dimas-b: what is hugo-style links?

dimas-b: I'm not sure I follow your suggestion, @flyrain . This is not about identifying a particular API spec revision (v1, v2, etc.) but about changes in Polaris server behaviour.

IMHO, adding support for a new API spec version (e.g. v2) is not a major change from the SemVer perspective, because it does not break backward-compatibility. In that regard, Polaris may support several API spec versions in the same release of Polaris.

On the other hand, if Polaris had a bug in implementing the Catalog REST API (for example) such that fixing it introduced a backward incompatible change in behaviour, why would we not do a major Polaris version bump? I think it would make sense. WDYT?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -27,19 +27,19 @@ This page discusses what can be expected from Apache Polaris as the project evol
 ## Using Polaris as a Catalog
 
 Polaris is primarily intended to be used as a Catalog of Tables and Views. As such, 
-it implements the [Iceberg REST API, Generic Tables API](../polaris-catalog-service/),
-and [Polaris Management API](../polaris-management-service/).
+it implements the Iceberg REST Catalog API and its own REST APIs
```
adutra:
```suggestion
it implements the Iceberg REST Catalog API and its own REST APIs.
```

site/content/in-dev/unreleased/evolution.md
```diff
@@ -27,19 +27,19 @@ This page discusses what can be expected from Apache Polaris as the project evol
...
 optional REST Catalog features may or may not be supported immediately. In general,
 there is no guarantee that Polaris releases always implement the latest version of
 the Iceberg REST Catalog API.
 
-Any API under Polaris control (e.g. the Management API) is maintained as a versioned REST
-API. New releases of Polaris may include changes to the current version of the API. When
-that happens those changes are intended to be compatible with prior versions of Polaris 
-clients. Certain endpoints and parameters may be deprecated.
+Any API under Polaris control that is not in an "experimental" or "beta" state 
+(e.g. the Management API) is maintained as a versioned REST  API. New releases of Polaris
```
adutra:
```suggestion
(e.g. the Management API) is maintained as a versioned REST API. New releases of Polaris
```

site/content/in-dev/unreleased/evolution.md
dimas-b: thx - fixed

site/content/in-dev/unreleased/evolution.md
dimas-b: thx - fixed

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,115 @@
...
+there is no guarantee that Polaris releases always implement the latest version of
+the Iceberg REST Catalog API.
+
+Any API under Polaris control that is not in an "experimental" or "beta" state 
+(e.g. the Management API) is maintained as a versioned REST API. New releases of Polaris
+may include changes to the current version of the API. When that happens those changes
+are intended to be compatible with prior versions of Polaris clients. Certain endpoints
+and parameters may be deprecated.
+
+In case a major change is required to the Management API that cannot be implemented in a
```
eric-maynard:
```suggestion
In case a major change is required to an API that cannot be implemented in a
```

This isn't specific to management, right?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,115 @@
...
+This approach is not meant to discourage the use of Polaris code in downstream projects, but
+to allow more flexibility in evolving the codebase to support new catalog-level features
+and improve code efficiency. Maintainers of downstream projects are encouraged to join Polaris 
+mailing lists to monitor project changes, suggest improvements, and engage with the Polaris
+community in case of specific compatibility concerns.   
+
+## Semantic Versioning
+
+Polaris strives to follow [Semantic Versioning](https://semver.org/) conventions both with
+respect REST APIs (beta and experimental APIs excepted), [Polaris Policies](../policy/)
```
eric-maynard:
```suggestion
respect to REST APIs (beta and experimental APIs excepted), [Polaris Policies](../policy/)
```

Is this true? Is it still semver if you only have a major version?

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,115 @@
...
+Supporting a new revision of the Iceberg REST Catalog spec in a backward-compatible way
+is not considered a major change. Specifically, supporting new REST API prefixes (e.g. `v2`)
+is not a mojor change because it does not affect older clients.
+
+Changing the implementation of an Iceberg REST Catalog feature / endpoint in a non-backward
+compatible way (e.g. removing or renaming a request parameter) is a major change.
+
+Dropping support for a configuration property with the `polaris.` name prefix is a major change.
+
+Dropping support for any previously defined [Policy](../policy/) type or property is a major change.
```
eric-maynard: Based on the format, it might be good to make this a `<ul>`

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,115 @@
...
+## Using Polaris as a Build-Time Dependency
+
+Polaris produces several jars. These jars or custom builds of Polaris code may be used in
+downstream projects according to the terms of the license included into Polaris distributions.
+
+The minimal version of the JRE required by Polaris code (compilation target) may be updated in
+any release. Different Polaris jars may have different minimal JRE version requirements.
+
+Changes in Java class should be expected at any time regardless of the module name or
+whether the class / method is `public` or `private`.
```
flyrain: Nit:
```suggestion
whether the class / method is `public` or not.
```

flyrain (dismissed):

site/content/in-dev/unreleased/evolution.md
dimas-b: good point - updated

site/content/in-dev/unreleased/evolution.md
dimas-b: thx - fixed

site/content/in-dev/unreleased/evolution.md
dimas-b: thx - fixed

site/content/in-dev/unreleased/evolution.md
dimas-b: fixed

site/content/in-dev/unreleased/evolution.md
```diff
@@ -0,0 +1,115 @@
...
+The following are some examples of Polaris approach to SemVer in REST APIs / configuration.
+These examples are for illustration purposes and should not be considered to be
+exhaustive.
+
+* Polaris implementing an optional Iceberg REST Catalog feature that was unimplemented
+in the previous release is not considered a major change.
+
+* Supporting a new revision of the Iceberg REST Catalog spec in a backward-compatible way
+is not considered a major change. Specifically, supporting new REST API prefixes (e.g. `v2`)
+is not a mojor change because it does not affect older clients.
```
eric-maynard:
```suggestion
is not a major change because it does not affect older clients.
```

site/content/in-dev/unreleased/evolution.md
eric-maynard: Reopening just to confirm the answer is "yes" here, I don't know the details of semver well enough to say myself. We essentially just don't have minor or patch versions?

dimas-b: My take on that is that API _revisions_ are orthogonal to Polaris release versions.

We can have any number of different API revisions (v1, v2, v3) in releases X and Y. The key is what changes between these releases with respect to Polaris behaviour observable via those APIs. If the there are no observable changes, there's no need to bump the major version (from the SemVer perspective). If there are observable changes that are not backward-compatible, we have to bump the major version number.

In that regard, Polaris can introduce API v2 in a 1.0.1 release, as long as we do not drop API v1. We can choose to bump the release version to 2.0.0 in that case, but it's not mandated by SemVer.

snazy (approved):

PR #1893, opened 2025-06-13
dummy commit

fivetran-rahulprakash:


adutra:
Hi @fivetran-rahulprakash thanks for your interest in Polaris. This PR doesn't seem to be in a reviewable state yet. Maybe you should keep it in your own Polaris fork for now and only open a PR here when it's ready for review, wdyt?

PR #1672, opened 2025-05-23
fix: Remove info log about deprecated internal method from PolarisConfiguration

dimas-b:
Remove the INFO log about calls to this method in Polaris, because users cannot do anything about these messages.

Phasing out old property names requires coordination with users (e.g. release notes), so it is not a matter of merely avoiding calls to that method in Polaris code.

Fixes #1666

PR #1894, opened 2025-06-13
fix(build): Fix deprecation warnings in FeatureConfiguration

adutra:


polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -83,6 +83,7 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  @SuppressWarnings("deprecation")
```
eric-maynard, line 86: This does get rid of the warnings, but aren't the warnings the main feature of the `@Deprecated` annotation?

ref: https://github.com/apache/polaris/pull/1672#discussion_r2107737093

snazy: The usage of the old config key is deprecated - using that should trigger a user-facing deprecation warning. The deprecation is triggering at the wrong site then.

adutra: Thanks for the link I didn't know there was a previous conversation on this topic.

That said, I rather disagree: the annotation is valuable mainly for consumers of the API, and especially when it's well documented in terms of when it was deprecated, when it will be removed and how to replace it. 

Here, we own both the source of the deprecation warnings (the `catalogConfigUnsafe` method) and the call sites where the deprecation warnings are triggered (in `FeatureConfiguration` class): IOW, keeping the warnings around is not useful for Polaris devs.

And there is no risk of "forgetting" to update `FeatureConfiguration` later on: as soon as we remove the deprecated method, the `FeatureConfiguration` class wouldn't compile anymore.

dimas-b: My take on this: we should rather remove deprecation from `catalogConfigUnsafe()` and file issues to remove support for those legacy properties. The method itself has a valid use case in Polaris code, which is specifically to support migration from old config names to new name, so the method itself is not conceptually "deprecated".

snazy: A `io.smallrye.config.RelocateConfigSourceInterceptor` implementation can be used to nag users to update the configs.

eric-maynard: There's already a log to remind users to remove the configs, so we are covered from that perspective. This was more intended to nag maintainers to remove the deprecated method at some point in time. Otherwise, why deprecate methods? Will we always add this suppression when a method is deprecated?

snazy:
Looks like the PR needs a rebase

snazy:
+1

adutra:
@eric-maynard is that OK to merge this?

snazy:
I think this is ready to be merged.

eric-maynard:
@adutra my question wasn't really answered -- is the strategy to always suppress all deprecation warnings?

snazy:
@eric-maynard the strategy is to _address_ deprecation coming from dependencies. If something needs to be done in the Polaris code base in the future, it's better to open an issue so it doesn't get lost.

eric-maynard:
That still doesn’t answer my question. Will we suppress all deprecation warnings?

dimas-b:
@eric-maynard @snazy @adutra : I'd like to reopen my suggestion of removing deprecation from `PolarisConfiguration.catalogConfigUnsafe()`

This method is not really intended to be called by "users" of Polaris. For internal calls (or calls from downstream projects) the purpose is to provide an upgrade path for users to gradually adjust to the new config names. So, in my mind the `catalogConfigUnsafe()` _method_ is not deprecated. What is deprecated in the old property name (which is its argument). The use of these deprecated properties already leads to a WARN log message (targeted at end users).

All in all, I think it is preferable to remove deprecation from `catalogConfigUnsafe()` and file GH issues for unsupporting those properties in 2.0. WDYT?

eric-maynard:
That makes more sense to me. If we don't want to mark it as deprecated, that is a reasonable argument, but I don't really grok why we would mark it as deprecated but then disable the warning.

My argument for leaving it as deprecated is mostly to make it clear that there should be no new use of this method. And, when the old configs are yanked, the method can accordingly be yanked as well. In that sense it is marked for deprecation/removal.

dimas-b:
Ok, I opened #1970 for my proposal. If accepted/merged, I'll follow up with an issue for dropping old config names in 2.0

PR #1897, opened 2025-06-13
feat(ci): Split Java Gradle CI in many jobs to reduce execution time

adutra:
This is a simple attempt to reduce the CI execution time for pull requests. Instead of executing all tests in the same job, it parallelizes the tests in 4 jobs:

* Code style & publishing checks
* Unit tests
* Quarkus tests 
* Integration tests

This could certainly be improved further, but I observed a total execution time around 10 minutes instead of 20 minutes before (at the cost of more compute power, of course).

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -34,6 +34,16 @@ on:
 
 jobs:
 
+  # TODO remove this job in a future PR
+  build:
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Always succeeds
+        run: echo "Success!"
+
   unit-tests:
     name: Unit Tests
     runs-on: ubuntu-latest
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -63,8 +63,8 @@ jobs:
       - name: Run unit tests
         run: |
           ./gradlew check sourceTarball distTar distZip publishToMavenLocal \
-            -x :polaris-quarkus-service:test \
-            -x :polaris-quarkus-admin:test \
+            -x :polaris-runtime-service:test \
+            -x :polaris-admin:test \
             -x intTest --continue
       - name: Archive test results
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
@@ -93,8 +93,8 @@ jobs:
       - name: Run Quarkus tests
         run: |
           ./gradlew \
-            :polaris-quarkus-service:test \
-            :polaris-quarkus-admin:test \
+            :polaris-runtime-service:test \
+            :polaris-admin:test \
             --continue
       - name: Archive test results
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -33,65 +33,111 @@ on:
     branches: [ "main" ]
 
 jobs:
-  build:
 
+  style-checks:
+    name: Code Style & Publishing Checks
     runs-on: ubuntu-latest
     permissions:
       contents: read
-
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-      - name: Set up JDK 21
-        uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4
-        with:
-          java-version: '21'
-          distribution: 'temurin'
-
-      # Configure Gradle for optimal use in GiHub Actions, including caching of downloaded dependencies.
-      # See: https://github.com/gradle/actions/blob/main/setup-gradle/README.md
-      - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@ac638b010cf58a27ee6c972d7336334ccaf61c96 # v4
-        with:
-          # The setup-gradle action fails, if the wrapper is not using the right version or is not present.
-          # Our `gradlew` validates the integrity of the `gradle-wrapper.jar`, so it's safe to disable this.
-          validate-wrappers: false
-
-      - name: Code style checks and tests
-        run: ./gradlew --continue check
-
-      - name: Check Maven publication
-        run: ./gradlew publishToMavenLocal sourceTarball
-
-      - name: Archive test results
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
-        if: always()
-        with:
-          name: upload-test-artifacts
-          path: |
-            **/build/test-results/**
-
-      - name: Stop Gradle daemons
-        run: ./gradlew --stop
-
-      # Ensure that the build works properly when building against the "latest greatest" Java version
       - name: Set up JDK 23
         uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4
         with:
           java-version: '23'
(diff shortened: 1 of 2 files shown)


.github/workflows/gradle.yml
```diff
@@ -33,65 +33,111 @@ on:
...
+  style-checks:
+    name: Code Style & Publishing Checks
     runs-on: ubuntu-latest
     permissions:
       contents: read
-
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-      - name: Set up JDK 21
+      - name: Set up JDK 23
```
gh-yzou: i think the ci before runs whole check, build, and tests on JDK 21 because Polaris is currently guarantee the compatibility with java 21, and then we are only doing selected check with java 23. 
If we just update the JDK version to 23, that means we will loose the test coverage for 21, right? if the intension of this PR is to just split the ci into multiple jobs, can we keep the java 21 setup?

adutra: Oh sorry that wasn't my intention. I can set the "Style check" job to 21, would that be enough? If not, I'm afraid we'd need a matrix job.

Also, tbh we should upgrade 23 to 24 but I don't want to introduce any unwanted changes in this PR.

gh-yzou: The original CI have the following coverage for 21 
```
run: ./gradlew --continue check
```
which I believe runs all tests, and on 23 it only does compilation and integration tests. However, with the current change, it seems we are doing full test coverage with java 23, but only style check with 21. Can we still retain the full test coverage with 21, limited test with 23 like before ? so that this is a pure refactoring task

If we also want to introduce full coverage with other jdk version like 23, we can do in follow up PR to introduce a test matrix

adutra: Well, that will limit the amount of parallelism to just 2 jobs, but if that's what you want, OK.

adutra: I went back to 3 jobs, 2 with 21 and 1 with 23, because 2 jobs only was too slow.

.github/workflows/gradle.yml
```diff
@@ -33,65 +33,89 @@ on:
...
 
-      # Ensure that the build works properly when building against the "latest greatest" Java version
+  integration-tests:
+    name: Integration Tests
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - name: Set up JDK 23
```
snazy: 23 is EOL - and 24 won't work. Probably better to go back to 21?

.github/workflows/gradle.yml
```diff
@@ -33,65 +33,89 @@ on:
     branches: [ "main" ]
 
 jobs:
-  build:
 
+  unit-tests:
```
snazy, line 47: Need to update `.asf.yaml` as well

adutra: I think we're going to have a problem to merge this PR because the `build` job is still marked as required, but it doesn't exist anymore.

adutra: I kept the old `build` job around to make it possible to merge this PR. We'd need to remove that job in a separate PR.

.github/workflows/gradle.yml
adutra: All jobs on 21 now, as per @snazy suggestion.

gh-yzou: sorry, for the confusion. i don't really mean still run ./gradlew --continue check. what i mean is when we are breaking it into jobs, make sure the tasks runs on 21 still provides the fully test coverage, include unit test, initTest, style check, publishToMaven etc. For 23 it can remain just have one job with coverage for integration test.

However, it seems we are now removing the coverage on 23 now because 23 is going to be end of like, which i think should be fine. 

If we want to further breakdown the job runs with 21 to more jobs, i am also fine with that.

gh-yzou (dismissed):

.github/workflows/gradle.yml
```diff
@@ -33,65 +33,99 @@ on:
...
-      - name: Run selected tests
-        run: ./gradlew :polaris-quarkus-service:intTest
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@ac638b010cf58a27ee6c972d7336334ccaf61c96 # v4
+        with:
+          validate-wrappers: false
+      - name: Run Quarkus tests
+        run: |
+          ./gradlew \
+            :polaris-quarkus-service:test \
```
dimas-b: Why do we isolate quarkus `test`, but not `intTest`?

dimas-b: I'd assume `intTest` takes the longest time in general, so it might be preferable to break it out if the goal is to reduce CI time :thinking:

dimas-b (dismissed):
LGTM in general. Please consider my comments optional.

snazy:
Just figured out that we should have another change to either this PR or a followup around Gradle caches.
Splitting up the Java CI into multiple parts "borks" the cached Gradle artifacts (cache stored from the last run on "main" wins). This results in unnecessary build effort and test runs. I'd be okay to live with that for a bit, but we should solve it.
NB: we have a mechanism for that in Nessie CI.

.github/workflows/gradle.yml
adutra: > Why do we isolate quarkus test, but not intTest?

Because these two jobs are roughly taking 8 minutes each now, so that sounded like a balanced choice. In fact, the `intTest` job is not necessarily the longest the job currently 🤷‍♂️

adutra:
> Splitting up the Java CI into multiple parts "borks" the cached Gradle artifacts (cache stored from the last run on "main" wins). This results in unnecessary build effort and test runs.

@snazy the caches seem to be working afaict; what makes you say they are "borked"?

snazy:
> @snazy the caches seem to be working afaict; what makes you say they are "borked"?

"Last Gradle cache store" wins. The next job will read the state of that "last one". So you either have test, intTest or ... - but not a merged state of all. This is what Nessie CI does - merging the cache state of all jobs.

adutra:
> > @snazy the caches seem to be working afaict; what makes you say they are "borked"?
> 
> "Last Gradle cache store" wins. The next job will read the state of that "last one". So you either have test, intTest or ... - but not a merged state of all. This is what Nessie CI does - merging the cache state of all jobs.

OK, I will tackle that in a follow-up since this one was approved already.

.github/workflows/gradle.yml
adutra: Also: the job is named "Quarkus tests" for simplicitly but it actually runs all the tests in `polaris-runtime-service` and `polaris-admin`.

We _could_ isolate only the tests annotated with `@QuarkusTest` or `QuarkusMainTest` but that's not a trivial thing to do with Gradle, so I went for simplicity. We can revisit that later if necessary.

adutra:
@snazy or @dimas-b do you mind re-approving please? I will look into cache improvements next.

snazy (approved):

eric-maynard:
Hey @adutra, did this possibly break the tests? I notice that the new tests are stuck in a pending state and PRs can't be merged.

Screenshot 2025-06-18 at 11 29 51 AM

adutra:
> Hey @adutra, did this possibly break the tests? I notice that the new tests are stuck in a pending state and PRs can't be merged.
> 
> 
> 
> Screenshot 2025-06-18 at 11 29 51 AM
> 
> 

Yes, this was expected unfortunately. If you rebase the PR it should get unstuck.

PR #1805, opened 2025-06-03
[Splitting] Initial SigV4 Auth Support for Catalog Federation

XJDKC:
## Introduction

This PR adds SigV4 authentication support for Catalog Federation in Polaris, enabling integrations with services like AWS Glue.

It introduces a complete and extensible credential management flow based on the [Apache Polaris Creds Management Proposal](https://docs.google.com/document/d/1MAW87DtyHWPPNIEkUCRVUKBGjhh5bPn0GbtV7fifm30/edit?tab=t.0#heading=h.v1rcy31tlwav). The key design goals are to:
* Manage Polaris service identities across tenants
* Inject service identity info into catalog entities during creation
* Retrieve credentials securely at runtime to connect with external services

The implementation supports both self-managed (single-tenant) and SaaS (multi-tenant) deployments, while ensuring a clean separation between user-supplied authentication parameters and Polaris-managed identities/credentials.

### How to review this PR?
This PR touches many test files because a new field was added to `PolarisCallContext`.

Pls focus on the core logic. You can safely ignore most of the test churn, please concentrate on the new components and flow listed below.

For the class name, I'm open to choose other names to make it clearer.

## End-to-End flow
The full flow of accessing remote catalog now looks like:
1. User sends CreateCatalogRequest
    * Includes `ConnectionConfigInfoModel` with SigV4 auth
    * `serviceIdentity` field is left null
2. EntityMutationEngine applies registered mutators
    * `CatalogEntityConnectionConfigMutator` injects `ServiceIdentityInfoDpo` into the entity using `ServiceIdentityRegistry`
3. Polaris persists the updated `ConnectionConfigInfoDpo`
    * Now includes both iamArn (user-facing) and a hidden `identityReference` for credential lookup
4.  Accessing Remote Catalog Service
    * `PolarisCredentialManager` uses `ServiceIdentityRegistry` to resolve a `ResolvedServiceIdentity`
    * Assumes the user-provided role using the resolved identity to obtain AWS temp credentials
    * Initialize RESTCatalog client with the temporary credentials

## Key Components Introduced
* **`ServiceIdentityInfoModel`**: The model we use to surface polaris service identity to users.
* **`ServiceIdentityInfoDpo`**: Persisted identity info injected into the catalog entity (includes hidden identityReference), the dpo won't contain the credential.
* **`ResolvedServiceIdentity`**: Encapsulates resolved metadata and credentials needed to interact with cloud services.
[...]

PR #1899, opened 2025-06-16
SigV4 Auth Support for Catalog Federation - Part 1: Entity Transformation System

XJDKC:
## Milestones

This is Part 1 of the [[Splitting] Initial SigV4 Auth Support for Catalog Federation](https://github.com/apache/polaris/pull/1805). Upcoming parts will build on this system:
* #1899
* #2190
* SigV4 Auth Support for Catalog Federation - Part 3: Service Identity Info Injection

## Introduction

This PR introduces a general-purpose `Entity Transformation System` to support future features like injecting service identity metadata into catalog entities for federated access (e.g. AWS Glue using SigV4). It provides a clean and extensible hook mechanism for enriching entities during specific lifecycle phases.

This system is intended to serve as the foundation for service identity injection and other dynamic transformations in subsequent parts of the SigV4 support work.

## Key Components
* `EntityTransformer`: A functional interface representing a single transformation to be applied to a PolarisBaseEntity.
* `EntityTransformationEngine`: A runtime engine that applies a sequence of `EntityTransformer`s to an entity based on the configured `TransformationPoint`.
* `TransformationPoint`: Enum representing defined hook points in the entity lifecycle (e.g. `CATALOG_PRE_PERSIST`).
* `@AppliesTo` annotation: A repeatable qualifier that marks a transformer as applicable to a given TransformationPoint.
* `EntityTransformationConfiguration`: Optional config-driven registration of transformers by identifier (preserving ordering).

## Example Configuration:
```properties
polaris.entity-mutation.mutators=no-op,catalog-connection-config
```

## Flowchart
Catalog Federation - Creds Management

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/transformation/EntityTransformationEngine.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/transformation/EntityTransformationEngine.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.core.entity.transformation;
+
+import org.apache.polaris.core.entity.PolarisBaseEntity;
+
+/**
+ * Engine responsible for applying a sequence of {@link EntityTransformer} transformations to a
+ * {@link PolarisBaseEntity}.
+ *
+ * <p>This abstraction allows Polaris to customize or enrich entities during runtime or persistence,
+ * based on configured or contextual logic (e.g., injecting service identity info, computing derived
+ * fields).
+ */
+public interface EntityTransformationEngine {
+  /**
+   * Applies all registered entity transformers to the provided entity, in order.
+   *
+   * @param transformationPoint The point in the entity lifecycle where transformers should be
+   *     applied.
+   * @param entity The original Polaris entity to mutate.
+   * @return A new transformed copy of the entity of {@link PolarisBaseEntity} after all
+   *     transformers are applied.
+   */
+  PolarisBaseEntity applyTransformers(
+      TransformationPoint transformationPoint, PolarisBaseEntity entity);
+}
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
@@ -47,6 +47,8 @@ import org.apache.polaris.core.entity.PolarisGrantRecord;
 import org.apache.polaris.core.entity.PolarisPrincipalSecrets;
 import org.apache.polaris.core.entity.PolarisPrivilege;
 import org.apache.polaris.core.entity.PolarisTaskConstants;
+import org.apache.polaris.core.entity.transformation.EntityTransformationEngine;
+import org.apache.polaris.core.entity.transformation.TransformationPoint;
 import org.apache.polaris.core.persistence.BaseMetaStoreManager;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.PolarisObjectMapperUtil;
@@ -476,6 +478,14 @@ public class TransactionalMetaStoreManagerImpl extends BaseMetaStoreManager {
     }
 
     ms.persistStorageIntegrationIfNeededInCurrentTxn(callCtx, catalog, integration);
+
+    // CATALOG_PRE_PERSIST transformation point
+    EntityTransformationEngine entityTransformationEngine = callCtx.getEntityTransformationEngine();
+    if (entityTransformationEngine != null) {
+      catalog =
+          entityTransformationEngine.applyTransformers(
+              TransformationPoint.CATALOG_PRE_PERSIST, catalog);
+    }
 
     // now create and persist new catalog entity
     this.persistNewEntity(callCtx, ms, catalog);
--- a/quarkus/defaults/src/main/resources/application.properties
+++ b/quarkus/defaults/src/main/resources/application.properties
@@ -188,6 +188,9 @@ polaris.oidc.principal-roles-mapper.type=default
 # polaris.storage.gcp.token=token
 # polaris.storage.gcp.lifespan=PT1H
 
+# Polaris Entity Transformation Config
+polaris.entity-transformation.transformers=no-op
+
 quarkus.arc.ignored-split-packages=\
   org.apache.polaris.service.catalog.api,\
   org.apache.polaris.service.catalog.api.impl,\
--- a/quarkus/defaults/src/main/resources/application.properties
+++ b/quarkus/defaults/src/main/resources/application.properties
@@ -201,6 +201,7 @@ quarkus.arc.ignored-split-packages=\
   org.apache.polaris.service.quarkus.auth.external.mapping,\
   org.apache.polaris.service.quarkus.auth.external.tenant,\
   org.apache.polaris.service.quarkus.auth.internal,\
+  org.apache.polaris.service.quarkus.entity.transformation,\
   org.apache.polaris.service.quarkus.events,\
   org.apache.polaris.service.quarkus.task,\
   org.apache.polaris.service.quarkus.secrets,\
--- a/persistence/eclipselink/src/main/java/org/apache/polaris/extension/persistence/impl/eclipselink/EclipseLinkPolarisMetaStoreManagerFactory.java
+++ b/persistence/eclipselink/src/main/java/org/apache/polaris/extension/persistence/impl/eclipselink/EclipseLinkPolarisMetaStoreManagerFactory.java
@@ -27,6 +27,7 @@ import java.nio.file.Path;
 import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.config.PolarisConfigurationStore;
 import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.core.entity.transformation.EntityTransformationEngine;
 import org.apache.polaris.core.persistence.LocalPolarisMetaStoreManagerFactory;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.bootstrap.RootCredentialsSet;
@@ -45,15 +46,18 @@ public class EclipseLinkPolarisMetaStoreManagerFactory
 
   @Inject EclipseLinkConfiguration eclipseLinkConfiguration;
   @Inject PolarisStorageIntegrationProvider storageIntegrationProvider;
+  @Inject EntityTransformationEngine entityTransformationEngine;
 
   protected EclipseLinkPolarisMetaStoreManagerFactory() {
-    this(null, null);
+    this(null, null, null);
   }
 
   @Inject
   protected EclipseLinkPolarisMetaStoreManagerFactory(
-      PolarisDiagnostics diagnostics, PolarisConfigurationStore configurationStore) {
-    super(diagnostics, configurationStore);
+      PolarisDiagnostics diagnostics,
+      PolarisConfigurationStore configurationStore,
+      EntityTransformationEngine entityTransformationEngine) {
+    super(diagnostics, configurationStore, entityTransformationEngine);
   }
 
   @Override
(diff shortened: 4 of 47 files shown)


snazy:
@XJDKC I'm a bit concerned about updating the representation of entities, because those represent the _persisted_ state. Maybe an opportunity to abstract such concerns from persistence?

XJDKC:
> @XJDKC I'm a bit concerned about updating the representation of entities, because those represent the _persisted_ state. Maybe an opportunity to abstract such concerns from persistence?

Hey @snazy, it’s definitely possible with this framework, we could define a new `TransformationPoint` like `CATALOG_PRE_LOAD` and dynamically apply a transformer to inject the service identity info when loading the Catalog Entity from the metastore.

I think different vendors may have different views on this. For example, it's quite reasonable to persist the service identity info, since it's very unlikely to change. We already persist service identity in the metastore for storage config, so it might make sense to follow a similar pattern here. Even if we choose not to persist the service identity directly on the Catalog Entity, we still need to persist the mapping between the entity and its assigned identity somewhere (e.g. In a `ServiceIdentityRegistry` I'm planning to add)

Changing the service identity for a Catalog Entity is risky, e.g., users might configure their IAM role’s trust policy to allowlist the specific service identity initially assigned to their federated Polaris catalog. If Polaris later switches to a different identity when accessing the Glue Catalog, that integration could break unexpectedly.

So perhaps the right approach is to allow the service vendor to decide whether the service identity should be persisted or injected dynamically. WDYT?

eric-maynard:
For better or worse, we currently use entities both for the persisted state of an object and for an in-memory representation of the same. I'm still a fan of separating this out (something like the DAO proposal from months back) but that a big refactor that I'm not sure I'd want to block SigV4.

@snazy perhaps we can view this as a step in the right direction as it encourages us to treat entities as immutable and to use discrete "transformations" to generate new entities instead of trying to modify them in place or to construct lots of new entities in an adhoc way throughout the code base. Eventually, entities should be truly immutable and this may become the canonical way to transform an entity into a new one.

I am, however, a little concerned about how heavyweight this process could become... if I just want to update an entity's version today, I can call a method like `setGrantRecordsVersion`. After this change, it looks like the expectation is that I have a transformation engine, and then I construct and execute a transformation to accomplish this change?

XJDKC:
> For better or worse, we currently use entities both for the persisted state of an object and for an in-memory representation of the same. I'm still a fan of separating this out (something like the DAO proposal from months back) but that a big refactor that I'm not sure I'd want to block SigV4.
> 
> @snazy perhaps we can view this as a step in the right direction as it encourages us to treat entities as immutable and to use discrete "transformations" to generate new entities instead of trying to modify them in place or to construct lots of new entities in an adhoc way throughout the code base. Eventually, entities should be truly immutable and this may become the canonical way to transform an entity into a new one.
> 
> I am, however, a little concerned about how heavyweight this process could become... if I just want to update an entity's version today, I can call a method like `setGrantRecordsVersion`. After this change, it looks like the expectation is that I have a transformation engine, and then I construct and execute a transformation to accomplish this change?

Yes, the Entity Transformation System I'm adding won't update the entity in place, it creates a copy and applies the modifications there.

What I want to point out is that, this system isn't meant to replace the current set methods used in the persistence layer. Its purpose is to provide hooks for applying transformations at specific points in the entity lifecycle.

It's admittedly a bit heavyweight, since each transformer creates a new copy of the entity. But without a builder-style abstraction (like what we have for generated model classes), we can't apply the transformation efficiently.

Given that we may eventually refactor this area to introduce a proper DAO layer, and that we're currently only using the transformation system for a single SigV4-related transformer, it might make sense to proceed with the current approach and revisit the refactor later. WDYT?

quarkus/defaults/src/main/resources/application.properties
```diff
@@ -188,6 +188,9 @@ polaris.oidc.principal-roles-mapper.type=default
 # polaris.storage.gcp.token=token
 # polaris.storage.gcp.lifespan=PT1H
 
+# Polaris Entity Transformation Config
```
XJDKC, line 191: Also link comment from Dennis here: https://github.com/apache/polaris/pull/1805#discussion_r2136212374

> On the one hand, it's useful to be able to have this kind of generalized transformer/mutator hook, but if the hook is also strongly coupled to particular features, we don't want to require separate configuration of the info injection here vs the core feature (e.g. sigv4) somewhere else.
>
>Especially if we end up having feature configurations for the core feature, then it's important that the feature itself auto-configures itself to be functional. In a way then the classes that define the sigv4 functionality should be what injects the mutator/transformer.

polaris-core/src/main/java/org/apache/polaris/core/entity/transformation/EntityTransformationEngine.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Engine responsible for applying a sequence of {@link EntityTransformer} transformations to a
+ * {@link PolarisBaseEntity}.
+ *
+ * <p>This abstraction allows Polaris to customize or enrich entities during runtime or persistence,
+ * based on configured or contextual logic (e.g., injecting service identity info, computing derived
+ * fields).
+ */
+public interface EntityTransformationEngine {
```
XJDKC, line 32: > It's unfortunate that we haven't had a chance yet to make PolarisBaseEntity immutable, but maybe our naming for this set of classes should be something that more strongly implies producing a "new transformed copy of the entity" rather than make it tempting for anyone to try mutating them in-place.
> 
> What does everyone think of naming these instead:
> 
> EntityTransformationEngine/EntityTransformer/TransformationPoint
EntityDecorationEngine/EntityDecorator/DecorationPoint
?

polaris-core/src/main/java/org/apache/polaris/core/entity/transformation/EntityTransformationEngine.java
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Applies all registered entity transformers to the provided entity, in order.
+   *
+   * @param transformationPoint The point in the entity lifecycle where transformers should be
+   *     applied.
+   * @param entity The original Polaris entity to mutate.
```
XJDKC, line 38: > It's too bad we don't have a way to express C++-style const correctness as we move towards immutable PolarisBaseEntity, but maybe we can at least mention here that the intention is that implementations should not mutate the input entity, but instead only create the new transformed entity.

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -477,6 +479,14 @@ private void revokeGrantRecord(
 
     ms.persistStorageIntegrationIfNeededInCurrentTxn(callCtx, catalog, integration);
 
+    // CATALOG_PRE_PERSIST transformation point
+    EntityTransformationEngine entityTransformationEngine = callCtx.getEntityTransformationEngine();
```
XJDKC, line 483: > For now, unfortunately we didn't quite manage to refactor all of the shared logic nryerrn AtomicOperationMetaStoreManager and TransactionalMetaStoreManagerImpl into a base class yet; the best we have is the prepareToPersistNewEntity hook defined in BaseMetaStoreManager.
> 
> This means either this logic needs to be added in both the AtomicOperationMetaStoreManager and TransactionalMetaStoreManagerImpl, or needs to be abstracted into BaseMetaStoreManager somehow while making sure both of those impls invoke the shared logic. persistNewEntity isn't specific to catalog-creation so you'd need an enum on entityType there if you just put this logic in there.

quarkus/defaults/src/main/resources/application.properties
```diff
@@ -188,6 +188,9 @@ polaris.oidc.principal-roles-mapper.type=default
...
+polaris.entity-transformation.transformers=no-op
```
XJDKC, line 192: > On the one hand, it's useful to be able to have this kind of generalized transformer/mutator hook, but if the hook is also strongly coupled to particular features, we don't want to require separate configuration of the info injection here vs the core feature (e.g. sigv4) somewhere else.
> 
> Especially if we end up having feature configurations for the core feature, then it's important that the feature itself auto-configures itself to be functional. In a way then the classes that define the sigv4 functionality should be what injects the mutator/transformer.

XJDKC (commented):
Link some comments from @dennishuo in this PR as well:

polaris-core/src/main/java/org/apache/polaris/core/entity/transformation/EntityTransformationEngine.java
XJDKC: The name used to be `EntityMutationEngine`, I have renamed it to `EntityTransformationEngine`

quarkus/defaults/src/main/resources/application.properties
XJDKC: Rulin: That's a good point, perhaps we can consider making some transformers tied to the feature parameters rather than the config itself if they are part of the core functionality of the feature.

snazy:
> Given that we may eventually refactor this area to introduce a proper DAO layer, and that we're currently only using the transformation system for a single SigV4-related transformer, it might make sense to proceed with the current approach and revisit the refactor later. WDYT?

Honestly, I think the approach is too heavyweight and too broad to target a very specific problem. It will be very hard for people to realize (later, by accident) that some entity instance does not reflect the "real" state. That will cause confusion and likely hard to debug issues.

I propose to separate the relevant parts into APIs/implementations that deal with the problem/use-case at hand, and definitely outside of the metastore - storage credential handling doesn't belong there IMO.

XJDKC:
> > Given that we may eventually refactor this area to introduce a proper DAO layer, and that we're currently only using the transformation system for a single SigV4-related transformer, it might make sense to proceed with the current approach and revisit the refactor later. WDYT?
> 
> Honestly, I think the approach is too heavyweight and too broad to target a very specific problem. It will be very hard for people to realize (later, by accident) that some entity instance does not reflect the "real" state. That will cause confusion and likely hard to debug issues.
> 
> I propose to separate the relevant parts into APIs/implementations that deal with the problem/use-case at hand, and definitely outside of the metastore - storage credential handling doesn't belong there IMO.

Agree with the concern about embedding this logic in the persistence layer, moving it outside does make sense in general as the final entity is different from the entity we pass to the metastore. That said, I want to point out a few things specific to the storage config use case.

For storage config, the service identity isn't assigned by Polaris itself, it's provided by users via the `CreateCatalogRequest`. That's why I have this proposal to fix this issue. Also, the storage integration is currently created within the persistence layer:
[TransactionalMetaStoreManagerImpl.java#L972-L978](https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java#L972-L978).

That's why I was leaning toward assigning the service identity in the persistence layer. **For certain vendors, we may need to store and retrieve the mapping between realm and service identity info from the database. Doing this lookup in the persistence layer allows us to keep everything within the same transaction when persisting the catalog entity**. Moving this from the persistence layer may end up causing inconsistency.

Adding some hook points is also very useful in general, e.g. it provides the ability to do some patches if there are some bugs and we end up persisting some wrong data in database. 

We can also put the service identity injection logic here: 
[PolarisAdminService.java#L730-L755](https://github.com/apache/polaris/blob/9b5325b3c7f97d5a34057468719eca679b32ac41/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java#L730-L755)

```
{
  "connectionConfigInfoDpo": {
    "authenticationParametersDpo": { /* ... */ },
    "serviceIdentityDpo": {
      "type": "AWS_IAM",
      "serviceArn": null, // we can choose to not persist this info in metastore, but keep this field to make option 2 and 3 possible
      "serviceInfoAuthLocatorUrn": "urn:polaris-service-identities:realm-id:catalog-id:connection"
    }
  },
}
```


WDYT? @eric-maynard @snazy @dennishuo @dimas-b

XJDKC:
Update from slack:
We're currently leaning toward removing the Entity Transformation System and instead injecting the service identity directly. At the moment, we're considering three possible injection points:
* **Option 1**: Persist only a reference (i.e. URN) and resolve the full identity via a lookup interface at service layer
* **Option 2**: Persist only a reference (i.e. URN) and resolve the full identity via a lookup interface at persistence layer
* **Option 3**: Persist both the reference and user-facing identity fields (userArn) in the persistence layer for convenience or performance.

```
CatalogEntity ServiceIdentityProvider::assignOnCatalogLoad(CatalogEntity entity);
CatalogEntity ServiceIdentityProvider::assignOnCatalogPersist(CatalogEntity entity);
CatalogEntity ServiceIdentityProvider::assignOnServiceLayer(CatalogEntity entity);
```

Also we are still discussing the shape of the `ConnectionConfigInfoDpo`
```
{
  "connectionConfigInfoDpo": {
    "authenticationParametersDpo": { /* ... */ },
    "serviceIdentityDpo": {
      "type": "AWS_IAM",
      "serviceArn": null, // we can choose to not persist this info in metastore, but keep this field to make option 2 and 3 possible
      "serviceInfoAuthLocatorUrn": "urn:polaris-service-identities:realm-id:catalog-id:connection"
    }
  },
}
```

snazy:
> For storage config, the service identity isn't assigned by Polaris itself, it's provided by users via the CreateCatalogRequest. That's why I have this proposal to fix this issue. Also, the storage integration is currently created within the persistence layer: [TransactionalMetaStoreManagerImpl.java#L972-L978](https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java#L972-L978).
> That's why I was leaning toward assigning the service identity in the persistence layer.

This tight coupling of the persistence concerns with object storage concerns frightens me for quite a while.

I propose to remove this unnecessary tight coupling first, which then, according to your statement, eliminates the whole need for this "transformation system". You can then have/configure dedicated services/indirection to do what you need to do.

XJDKC:
Close the PR since it's not necessary

PR #1844, opened 2025-06-10
Add Polaris Events to Persistence

adnanhemani:
As per my [proposal on the Dev ML](https://lists.apache.org/thread/nfp40hmwryybznxtgffvy4c5l5kk19ho), this is the implementation of persisting Polaris Events to a persistence. The overall goal for this PR is to store Events in the Polaris persistence layer - which will later be used to power the Iceberg Events API.

High-level overview of the changes:

* Created a new persistence table and model for Events
* Implemented a new persistence method to flush events to said table
* Implemented an in-memory and file-based buffer implementation for flushing events to the persistence
* Created configurations for enabling all of these features
* Modified PolarisEventListener and PolarisEvents to optimally achieve this goal
* Created new sample events for CreateTable (explicitly part of the Iceberg Events API spec) and CreateCatalog (which is not explicitly part of the Iceberg Events API spec and will be served as a CustomOperation).

Here are the following upcoming items that will be updated in this PR soon:
* Testing for the new code paths added

Upcoming items that will be tackled in subsequent PRs (and were excluded to ensure that this PR contains only the end-to-end MVP):
* Support for multiple datasources to store the events
* Allowing multiple event listeners to be set
* Additional event types and related instrumentation

PR #1904, opened 2025-06-16
Add Events for PolarisServiceImpl APIs

adnanhemani:
Adding instrumentation for events for all events defined in `PolarisServiceImpl.java`. Not all events will be useful and/or necessary - but allow users to plug in with custom event listeners for any of these APIs. The default event listener is a no-op.

The main file to review here is `PolarisServiceImpl`. All new events are simply just a constructor and getter methods.


service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -142,12 +211,24 @@ private PolarisAdminService newAdminService(
   @Override
   public Response createCatalog(
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
+    String requestId = PolarisEvent.createRequestId();
```
adutra: A few remarks:

Injecting before/after event logic in every method looks like we are mixing concerns. I think the decorator pattern could make things look a lot nicer.

About the execution flow: how do we handle errors?

E.g. if the listener throws on a before event call, should the execution halt or continue?

Also: if the listener throws on an after event call, should the execution be aborted and the persisted changes rolled back, or should the business changes be kept? IOW, are we OK with weaker delivery guarantees for events (like "at most once")?

service/common/src/main/java/org/apache/polaris/service/events/AfterAddGrantToCatalogRoleEvent.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+package org.apache.polaris.service.events;
+
+import org.apache.polaris.core.admin.model.AddGrantRequest;
+import org.apache.polaris.core.auth.AuthenticatedPolarisPrincipal;
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+/** Event fired after a grant is added to a catalog role in Polaris. */
+public class AfterAddGrantToCatalogRoleEvent implements PolarisEvent {
```
adutra: Couldn't all these event classes be records? We could even group them in one single file instead of having hundreds of similar files around.

adnanhemani: Same response as: https://github.com/apache/polaris/pull/1844#discussion_r2152337757

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adnanhemani: Re: execution flow

Personally, I'm okay if we do not give a successful execution back to the user if the event listeners throw an error. For the use case in which customers actually add perhaps additional authorization or validations to a call through an event listener, this is exactly what the customer would expect. This is how event listeners are currently instrumented today as well - and I don't see a solid reasoning to change this paradigm right now. (Event Listeners and their configuration are still the responsibility of the Polaris admin, as this is basically like a custom plugin; additionally, the sample events-to-persistence implementations in #1844 take this into account and purposefully will not throw errors during the execution of the event listener in-line code).

----

Re: decorator

I believe the pattern you are referring to includes defining a @Decorator class which would also implement `PolarisCatalogsApiService`/`PolarisPrincipalsApiService`/`PolarisPrincipalRolesApiService` and then use the `PolarisServiceImpl` class as a delegator? If this is not correct, please do help provide some sample/hints that would make it clearer what you mean.

But going back - in the case of the decorator+delegator, sure this may make this particular file cleaner - but the same amount of code will exist as we will just be moving the event listener hooks to a different file and then calling `PolarisServiceImpl` from there. If you think that slight amount of cleanliness outweighs the redirection, I'm happy to take this as an action item. Wanted to confirm what you mean before going ahead with an implementation on this.

snazy:
I think this deserves a dev-ML discussions and consensus on the technical approach.
Alex and Eric already raised concerns about both the general and the specific technical approach

adnanhemani:
@snazy - to be fair, Eric did not raise any concerns on this PR or approach (have verified offline with him as well). And I don't really see why this warrants a ML-thread. It seems that issues with this PR are implementation-specific, and I don't see a good reason to use ML threads for that (that is what PRs comments should be for).

But I will start a ML thread as per your suggestion - I hope there will be fruitful, outcome-driven conversation that will lead to a quick consensus.

eric-maynard:
I did have one or two comments on the related PR #1844, but that only to clarify the approach. I am OK with the direction of this PR and the other one; we need events for these operations. If we can reduce the amount of code needed to add the new events or otherwise break down or streamline the process that would be great but I have no blocking concerns at this time.

adnanhemani:
Updated this PR as per the email thread [here](https://lists.apache.org/thread/s5y24967tyknghnsj7ktwkcxblt1xckc). The main change in this commit is to convert all events back to records.

The email thread shows that there is no hard opinion on the topics raised above - and so I've not made any major changes for the comments above regarding error handling or decorator/delegator design patterns.

adnanhemani:
Update:
* I've removed all event IDs as part of the event (these aren't technically a business logic requirement of an event)

eric-maynard (approved):
LGTM! This isn't all the new events we'll need to add but it's a good start.

adnanhemani:
@eric-maynard - any way to retrigger the tests? Seems like it failed on a processing step unrelated to the changes.

snazy:
This PR received a lot of comments, should wait for @adutra's review as well.

service/common/src/main/java/org/apache/polaris/service/events/AfterCatalogGetEvent.java
```diff
@@ -0,0 +1,25 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.service.events;
+
+import org.apache.polaris.core.admin.model.Catalog;
+
+/** Emitted when Polaris intends to get a catalog. */
```
snazy: "intends to" are `AfterCatalogGet`  not the same?

service/common/src/main/java/org/apache/polaris/service/events/AfterCredentialsRotateEvent.java
```diff
@@ -0,0 +1,30 @@
...
+package org.apache.polaris.service.events;
+
+import org.apache.polaris.core.admin.model.PrincipalWithCredentials;
+
+/**
+ * Event fired after credentials are rotated for a principal in Polaris.
+ *
+ * @param rotatedPrincipal the updated principal with new credentials
+ */
+public record AfterCredentialsRotateEvent(PrincipalWithCredentials rotatedPrincipal)
```
snazy: This seems to expose the new secrets to any event consumer. Strong -1 on exposing secrets.

snazy (changes requested):
This change exposes secrets to event listeners, which is quite concerning.

I've posted a bunch of comments for discussion about the approach on the dev-ML.

service/common/src/main/java/org/apache/polaris/service/events/AfterCredentialsRotateEvent.java
eric-maynard: > Strong -1 on exposing secrets.

What does "exposing" mean? Are you saying no interface methods should have secrets as arguments?

service/common/src/main/java/org/apache/polaris/service/events/AfterCatalogGetEvent.java
adnanhemani: Sorry, typo on my end. Changed in next revision.

service/common/src/main/java/org/apache/polaris/service/events/AfterCredentialsRotateEvent.java
adnanhemani: Changed in next revision.

eric-maynard (approved):
Still LGTM, and the secrets are gone from the events now which is good

eric-maynard:
@snazy are there still changes you'd like to see here?

eric-maynard (approved):

snazy:
Yea, I think we should simplify and enhance the API, as proposed on dev@

sfc-gh-ahemani:
@snazy - I've replied on the referenced email thread >2 weeks ago but there is not any response to any of the comments on the thread since then. If you have a hard opinion on this, can you please reply to the email thread to keep this work moving forward?

If there is not a hard opinion, I will continue pushing to get this PR merged as-is.

adnanhemani:
@snazy - I've replied on the referenced email thread >2 weeks ago but there is not any response to any of the comments on the thread since then. If you have a hard opinion on this, can you please reply to the email thread to keep this work moving forward?

If there is not a hard opinion, I will continue pushing to get this PR merged as-is.

snazy:
My impression was that you're working on the proposed ideas.

My (hard) opinion on this is that we have to allow event listeners to
a) associate after with before events
b) opt in to get success/failure results
and to reduce the amount of code that is necessary in call sites and listener implementations and events themselves.

singhpk234 (commented):
can we have all the records in one class ?

eric-maynard:
The current events aren't all in one file, so while I think grouping them that way is fine we should be consistent.

IIUC association* of different events is possible (and intended) on the EventListener side.

adnanhemani:
> The current events aren't all in one file, so while I think grouping them that way is fine we should be consistent.

Agreed, I can do this in the next round of revisions. But am waiting for the followup meeting regarding events to put further work in on this PR.

> IIUC associated of different events is possible (and intended) on the EventListener side.

This is correct as well. @snazy - does that resolve your concerns?

snazy:
Thanks for removing the secrets.

But I still believe that it can be simplified, and have the currently missing the failure notifications. I've proposed a way to achieve this using Java pseudo code on the dev-ML. Adding it here as well below.

```java
public interface EventListener {
  /**
   * "Dummy event foo" about to happen, with parameters x + y, no additional information collected
   * for "after the fact".
   */
  InFlightVoid beforeEventFoo(String x, int y);

  /** About to commit to an Iceberg table. */
  InFlight<IcebergTableCommitted> beforeIcebergTableCommit(UpdateTableRequest updateTableRequest);
}

public interface InFlight<SUCCESS_PAYLOAD> {
  /** Called "after the fact". */
  void success(SUCCESS_PAYLOAD payload);

  /** Potentially called "after the fact" in case of a failure. */
  void failure(Throwable throwable);

  /** Potentially called "after the fact" in case of a failure. */
  void failure(String reason);
}

/** Convenience for events that do not have a success-payload. */
public interface InFlightVoid extends InFlight<Void> {
  default void success() {
    success(null);
  }
}

/** Success-payload for table-committed. */
public record IcebergTableCommitted(TableMetadata originalState, TableMetadata newState) {}
```

Example for a simple event "foo", w/o a success-payload:
```java
public interface Foo {
  void doFoo();
}

class FooImpl implements Foo {
  EventListener eventListener;

  @Override
  public void doFoo() {
    InFlightVoid event = eventListener.beforeEventFoo("foo", 42);
    try {
      // do foo stuff
      event.success();
    } catch (Exception x) {
      event.failure(x);
    }
  }
}
```

Example for table committed, with update-request, before and after states. Up to the event listener implementation which parts are consumed.
```java
public interface Commit {
  void doCommit(UpdateTableRequest updateTableRequest);
}

class CommitImpl implements Commit {
  EventListener eventListener;

  @Override
  public void doCommit(UpdateTableRequest updateTableRequest) {
    InFlight<IcebergTableCommitted> event =
        eventListener.beforeIcebergTableCommit(updateTableRequest);
    try {
      // do table commit
      TableMetadata previousTableMetadata = ...;
      if (somePreconditionFailed) {
        event.failure("Precondition xyz failed");
        return;
      }
      TableMetadata updatedTableMetadata = ...;
      event.success(new IcebergTableCommitted(previousTableMetadata, updatedTableMetadata));
    } catch (Exception x) {
      event.failure(x);
    }
  }
}
```

adnanhemani:
Further discussion on stateful event handling is on the Dev ML.

adnanhemani:
Closed in favor of #2482.

Issue #1869, opened 2025-06-11
Make the POLARIS_BOOTSTRAP_CREDENTIALS consistent across docker image, docker composer and gradle run


flyrain:
### Describe the bug

We got different setting for the local default. It causes issues like the docker run doesn't work with `./regtests/run_spark_sql.sh`. We should make it consistent across the project.
```
POLARIS_BOOTSTRAP_CREDENTIALS: default-realm,root,s3cr3t
```
```
 "-Dpolaris.bootstrap.credentials=POLARIS,root,secret",
```

sclee01:
Hi @flyrain , this looks like a good issue to start with. Can I take this on?

adutra:
@sclee01 it's assigned to you.

sclee01:
Hi @adutra, I have opened a PR to address this issue here: #1905.  
Please take a look when you get a chance.

MonkeyCanCode:
Closed as https://github.com/apache/polaris/pull/1905 is merged.

PR #1905, opened 2025-06-17
Unify bootstrap credentials format across docker-compose and Gradle

sclee01:
This PR addresses an inconsistency in bootstrap credential settings across various parts of the project.
It is originated from the following issue: [#1869](https://github.com/apache/polaris/issues/1869#issuecomment-2976500860)

### docker-compose.yml used:
`POLARIS_BOOTSTRAP_CREDENTIALS: default-realm,root,s3cr3t`

### whereas other configurations (e.g., gradle.properties, README examples) used:
`-Dpolaris.bootstrap.credentials=POLARIS,root,secret`

This inconsistency caused errors when running commands like `./regtests/run_spark_sql.sh.`

## Summary of Changes

Unified all credential formats to use `POLARIS,root,s3cr3t`

Modified:

- gradle.properties
- docker-compose.yml
- README.md
- run_spark.sql.sh

This change ensures consistent local development setup across the project.

### DIFF
--- a/client/python/integration_tests/conftest.py
+++ b/client/python/integration_tests/conftest.py
@@ -133,7 +133,7 @@ def root_token(
         client_id=client_id,
         client_secret=client_secret,
         grant_type="client_credentials",
-        _headers={"realm": "default-realm"},
+        _headers={"realm": "POLARIS"},
     )
     return token
 
--- a/plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
+++ b/plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
@@ -33,7 +33,7 @@
     "                            client_id=client_id,\n",
     "                          client_secret=client_secret,\n",
     "                          grant_type='client_credentials',\n",
-    "                          _headers={'realm': 'default-realm'})\n"
+    "                          _headers={'realm': 'POLARIS'})\n"
    ]
   },
   {
--- a/plugins/spark/v3.5/regtests/docker-compose.yml
+++ b/plugins/spark/v3.5/regtests/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - "8182"
     environment:
       AWS_REGION: us-west-2
-      POLARIS_BOOTSTRAP_CREDENTIALS: POLARIS,root,secret
+      POLARIS_BOOTSTRAP_CREDENTIALS: POLARIS,root,s3cr3t
       quarkus.log.file.enable: "false"
       quarkus.otel.sdk.disabled: "true"
       polaris.features."ALLOW_INSECURE_STORAGE_TYPES": "true"
--- a/README.md
+++ b/README.md
@@ -67,11 +67,11 @@ Apache Polaris is built using Gradle with Java 21+ and Docker 27+.
 - `./gradlew build` - To build and run tests. Make sure Docker is running, as the integration tests depend on it.
 - `./gradlew assemble` - To skip tests.
 - `./gradlew check` - To run all checks, including unit tests and integration tests.
-- `./gradlew run` - To run the Polaris server locally; the server is reachable at localhost:8181. This is also suitable for running regression tests, or for connecting with Spark. Set your own credentials by specifying system property `./gradlew run -Dpolaris.bootstrap.credentials=POLARIS,root,secret` where:
+- `./gradlew run` - To run the Polaris server locally; the server is reachable at localhost:8181. This is also suitable for running regression tests, or for connecting with Spark. Set your own credentials by specifying system property `./gradlew run -Dpolaris.bootstrap.credentials=POLARIS,root,s3cr3t` where:
   - `POLARIS` is the realm
   - `root` is the CLIENT_ID
   - `secret` is the CLIENT_SECRET
-  - If credentials are not set, it will use preset credentials `POLARIS,root,secret`
+  - If credentials are not set, it will use preset credentials `POLARIS,root,s3cr3t`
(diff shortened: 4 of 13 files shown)


plugins/spark/v3.5/regtests/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - "8182"
     environment:
       AWS_REGION: us-west-2
-      POLARIS_BOOTSTRAP_CREDENTIALS: POLARIS,root,secret
+      POLARIS_BOOTSTRAP_CREDENTIALS: default-realm,root,s3cr3t
```
adutra: @sclee01 I wonder if `POLARIS` isn't a better default, because that is the default realm defined in Polaris configuration:

https://github.com/apache/polaris/blob/23f6480054c389d9ea9414182b8e04f6f7de5586/quarkus/defaults/src/main/resources/application.properties#L107

sclee01: @adutra Updated to POLARIS for consistency — take a look when you can.

dimas-b (commented):
What about `getting-started/jdbc/docker-compose-bootstrap-db.yml`?

sclee01:
Thanks for your notice. I will check it again.

adutra:
> What about `getting-started/jdbc/docker-compose-bootstrap-db.yml`?

It seems all the Jupyter notebooks need to be changed (search for `SparkPolaris.ipynb`).

plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
```diff
@@ -21,7 +21,7 @@
     "from polaris.catalog.api_client import ApiClient as CatalogApiClient\n",
     "from polaris.catalog.api_client import Configuration as CatalogApiClientConfiguration\n",
     "\n",
-    "polaris_credential = 'root:s3cr3t' # pragma: allowlist secret\n",
+    "polaris_credential = 'root:secret' # pragma: allowlist secret\n",
```
adutra: I think you also need to modify line 36 below.

sclee01: Done! Line 36 and others are now fixed as well

client/python/integration_tests/conftest.py
```diff
@@ -133,7 +133,7 @@ def root_token(
         client_id=client_id,
         client_secret=client_secret,
         grant_type="client_credentials",
-        _headers={"realm": "default-realm"},
```
HonahX, line 136: Thanks for the unification, could you please also update the `polaris_root_credential` in this file to match the credentials in `docker-compose.yml`?

getting-started/jdbc/README.md
```diff
@@ -42,7 +42,7 @@ This example requires `jq` to be installed on your machine.
     export QUARKUS_DATASOURCE_PASSWORD=postgres
     export ASSETS_PATH=$(pwd)/getting-started/assets/
     export CLIENT_ID=root
-    export CLIENT_SECRET=s3cr3t
+    export CLIENT_SECRET=secret
```
dimas-b: nit: I'm fine with this change, however the old `s3cr3t` string was quite unique and easy to search for in text ;)

sclee01:
I’ve rechecked for consistency.

getting-started/jdbc/README.md
flyrain: I agreed with @dimas-b for the uniqueness, but the string `secret` was used in regression tests. To change it to something else will have a bigger impact on downstream projects.

flyrain:
There are strings of `s3cr3t` in `BootstrapCommandTestBase`, but I think it's fine. We don't have to unify them.

flyrain (dismissed):
Thanks a lot for doing this, @sclee01 !

sclee01:
Thanks for your comment, @snazy . You're absolutely right that `default-realm` is more generic and descriptive.

I originally used `default-realm,root,s3cr3t`, but updated it to `POLARIS,root,secret` following reviewer feedback — particularly the point that "POLARIS" is set as the default realm in Polaris configuration.

I am okay with keeping it this way if it makes sense, but also open to switching back if there's a strong preference for `default-realm`.  Appreciate your thoughts.

snazy:
> Thanks for your comment, @snazy . You're absolutely right that `default-realm` is more generic and descriptive.

Ah, bummer. Right, it's mentioned in the default configuration that way. Guess we have to live with `POLARIS` then.

But mind changing `secret` back to `s3cr3t` - I suspect that's better to figure out (code search).

sclee01:
Got it. I'll change `secret` back to `s3cr3t` so it's easier to search for.  Thanks for pointing that out!.

sclee01:
All feedback has been applied.

dimas-b (commented):
Changes LGTM overall, but please check CI errors :sweat_smile:

sclee01:
Fixed CI failure caused by credential format mismatch. PTAL

eric-maynard:
I'm having trouble figuring out the state of the PR based on what appear to be two conflicting comments:
* https://github.com/apache/polaris/pull/1905#discussion_r2153433994
* https://github.com/apache/polaris/pull/1905#issuecomment-2983856145

I don't have a preference between the two, but are we aligned on using `s3cr3t`?

sclee01:
We’ve aligned on using `POLARIS,root,s3cr3t` for consistency and searchability. All files now reflect this.

dimas-b:
> I don't have a preference between the two, but are we aligned on using s3cr3t?

This PR moves all default / example credentials to `s3cr3t` as far as I can tell. I'm fine with this approach.

flyrain (approved):

PR #1908, opened 2025-06-18
Update spark client to use the shaded iceberg-core in iceberg-spark-runtime to avoid spark compatibilities issue

gh-yzou:
We run into an issue when testing the spark client using --packages (the jar path works correctly), where the iceberg requires avro 1.12.0, but the one provide by spark is 1.11.4. Even though the dependency is downloaded, the one used is still the one provided by spark. There for we see error like following when do select * from iceberg_tb:
```
java.lang.NoSuchMethodError: 'org.apache.avro.LogicalTypes$TimestampNanos org.apache.avro.LogicalTypes.timestampNanos()'
  at org.apache.iceberg.avro.TypeToSchema.<clinit>(TypeToSchema.java:50)
  at org.apache.iceberg.avro.AvroSchemaUtil.convert(AvroSchemaUtil.java:64)
  at org.apache.iceberg.avro.AvroSchemaUtil.convert(AvroSchemaUtil.java:59)
  at org.apache.iceberg.GenericManifestFile.<clinit>(GenericManifestFile.java:42)
  at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:75)
  at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:53)
  at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
  at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
  at org.apache.iceberg.common.DynConstructors$Ctor.newInstanceChecked(DynConstructors.java:51)
  at org.apache.iceberg.common.DynConstructors$Ctor.newInstance(DynConstructors.java:64)
  at org.apache.iceberg.avro.InternalReaders$PlannedStructLikeReader.reuseOrCreate(InternalReaders.java:67)
  at org.apache.iceberg.avro.InternalReaders$PlannedStructLikeReader.reuseOrCreate(InternalReaders.java:42)
  at org.apache.iceberg.avro.ValueReaders$PlannedStructReader.read(ValueReaders.java:963)
  at org.apache.iceberg.avro.InternalReader.read(InternalReader.java:107)
  at org.apache.iceberg.avro.NameMappingDatumReader.read(NameMappingDatumReader.java:57)
  at org.apache.avro.file.DataFileStream.next(DataFileStream.java:263)
  at org.apache.avro.file.DataFileStream.next(DataFileStream.java:248)

```

A quick mitigation is to remove the default one used by spark. However, to avoid future conflict with other library, we switch to use the shaded library shipped in iceberg-spark-runtime. 

The major change is that instead using com.fasterxml.jackson.annotation, we uses org.apache.iceberg.shaded.com.fasterxml.jackson.annotation for all request and response to ensure they can be serialized and deserialized correctly by Iceberg RESTClient. 
Since the classes now looks different than the one used at server side now, we actually let client uses its own request and response classes. Those classes are added manually at this moment, will follow up to autogenerate them based on spec directly https://github.com/apache/polaris/issues/1909

Since we are using all libraries shipped along with the iceberg-spark-runtime, it simplifies the build for client significantly


codestyle/checkstyle.xml
```diff
@@ -39,13 +39,5 @@
     <property name="optional" value="true"/>
   </module>
 
-  <module name="TreeWalker">
-    <!-- Checks for imports                            -->
-    <!-- See http://checkstyle.org/config_imports.html -->
```
gh-yzou, line 44: @adutra Since now we actually start switching to use the shaded library, this check doesn't make sense anymore, i removed it for now.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.spark.Spark3Util;
-import org.apache.polaris.service.types.GenericTable;
+import org.apache.polaris.spark.rest.GenericTable;
```
dimas-b: This change is fine from my POV, but it does not seem to match the PR title and description. Please mention this refactoring there.

plugins/spark/v3.5/integration/build.gradle.kts
```diff
@@ -45,9 +45,16 @@ dependencies {
...
 
-  testImplementation(project(":polaris-api-management-model"))
+  testImplementation(
+    "org.apache.iceberg:iceberg-spark-runtime-${sparkMajorVersion}_${scalaVersion}:${icebergVersion}"
+  )
   testImplementation(project(":polaris-spark-${sparkMajorVersion}_${scalaVersion}"))
 
+  testImplementation(project(":polaris-api-management-model")) {
+    // exclude the iceberg
+    exclude("org.apache.iceberg", "iceberg-core")
```
dimas-b: Having a dependency from `polaris-api-management-model` to `iceberg-core` is a bit concerning, IMHO, from the overall API design POV. Why should Polaris' own Management API depend on Iceberg core? This is something for follow-up discussions, of course.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/rest/GenericTable.java
```diff
@@ -0,0 +1,190 @@
...
+
+import jakarta.validation.constraints.NotNull;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.iceberg.shaded.com.fasterxml.jackson.annotation.JsonCreator;
+import org.apache.iceberg.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+
+// TODO: auto generate the class based on spec
+@jakarta.annotation.Generated(
```
dimas-b: +1 to the TODO, but as of now this class is checked in to the source repo, not generated at build time :thinking:

codestyle/checkstyle.xml
```diff
@@ -39,13 +39,5 @@
...
-    <module name="IllegalImport">
-      <property name="illegalPkgs" value=".*\.shaded\..*, .*\.relocated\..*"/>
```
dimas-b: -1 to this specific change. I think an exception (cf. line 38) should be added to Spark-related modules as opposed to dropping this check for all modules.

gh-yzou: sure. I removed it because that check was originally just introduced by spark client. however it might be useful for other module also.

Since only one config file can be used, i introduces another style check file without this particular rule, and use that for the spark client check

plugins/spark/v3.5/integration/build.gradle.kts
gh-yzou: sorry, that is actually a false alarm, i double checked the dependency graph, this project doesn't rely on iceberg-core, removed the exclude

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/rest/GenericTable.java
gh-yzou: yes, this class requires a different code generation script because the code looks different compare with the one generated by server side in terms of the import. where the server side uses 
```
import com.fasterxml.jackson.annotation.JsonCreator
```
but we will need to use 
```
import org.apache.iceberg.shaded.com.fasterxml.jackson.annotation.JsonCreator;
```
at client side to enable reuse of iceberg rest client correctly, because the RESTClient shipped in the iceberg-spark-runtime has the jackson library shaded. i will have to add new generation script and build file to make sure things can be generated properly,  and I have an issue to track the improvement https://github.com/apache/polaris/issues/1909

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/PolarisSparkCatalog.java
gh-yzou: because the classes used by client and server now are different in terms of the import (server uses the regular import, but client uses the shaded one coming from iceberg). So the implementation of the class at client side are now different compare with server side, and the client side is now suppose to use the one provided at client side. This change is to use the class provided by the client side, which will be auto generated in the coming task.

Move on, the only contract client have with server will be the API spec, which i think is the correct thing to do.

gh-yzou: I added a comment here, and also updated the description to contain more details

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/PolarisManagementClient.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.iceberg.rest.responses.OAuthTokenResponse;
+import org.apache.polaris.service.it.env.ClientCredentials;
+import org.apache.polaris.service.it.env.ManagementApi;
+import org.apache.polaris.service.it.env.PolarisApiEndpoints;
+
+/**
+ * That class provides rest client that is can be used to talk to Polaris Management service and
+ * auth token endpoint. This class is currently used by Spark Client tests for commands that can not
+ * be issued through spark command, such as createCatalog etc.
+ */
```
flyrain: Nit:
```suggestion
/**
 * This class provides a REST client for the Polaris Management service endpoints and its auth-token endpoint, which used in
 * Spark client tests to run commands that Spark SQL can’t issue directly (e.g., createCatalog).
 */
```

plugins/spark/v3.5/integration/build.gradle.kts
```diff
@@ -45,9 +45,13 @@ dependencies {
 
   implementation(project(":polaris-runtime-service"))
 
-  testImplementation(project(":polaris-api-management-model"))
+  testImplementation(
+    "org.apache.iceberg:iceberg-spark-runtime-${sparkMajorVersion}_${scalaVersion}:${icebergVersion}"
+  )
   testImplementation(project(":polaris-spark-${sparkMajorVersion}_${scalaVersion}"))
 
+  testImplementation(project(":polaris-api-management-model"))
```
flyrain, line 53: +1 this provides more flexibility on the client side by minimizing dependencies, as it only depends on the Polaris REST spec now.

flyrain (dismissed):
Thanks a lot @gh-yzou for the fix! LGTM!

flyrain:
A minor suggestion, would you mind sharing the error msg/stack trace of the issue the PR trying to fix? This will provide more context people who track this issue.

dimas-b:
I raised some concerns about this PR on the dev ML: https://lists.apache.org/thread/0z30f3cfvm41hxlbxgp4fqdpv7mfgnv8

Let's resolve that discussion before merging.

codestyle/checkstyle_no_illegalimport.xml
dimas-b, line 1: I'm not sure this approach is ideal for making exceptions to Checkstyle rules. 

@adutra : WDYT?

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -78,15 +57,9 @@ dependencies {
     exclude("org.apache.commons", "*")
   }
 
-  implementation("org.apache.iceberg:iceberg-core:${icebergVersion}")
```
flyrain, line 78: cc @dimas-b, it doesn't depend on iceberg-core anymore, not leaking of transitive dependencies.

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/PolarisManagementClient.java
gh-yzou: updated

codestyle/checkstyle_no_illegalimport.xml
gh-yzou: switched to use suppress rule now

gh-yzou:
@flyrain i added the stacktrace in the PR description.

codestyle/checkstyle.xml
```diff
@@ -26,6 +26,10 @@
   <property name="severity" value="warning"/>
   <property name="fileExtensions" value="java, properties, xml"/>
 
+  <module name="SuppressionFilter">
+    <property name="file" value="${checkstyle.suppression.file}"/>
```
dimas-b: Is this not the same as lines 41-42?

codestyle/checkstyle_suppressions.xml
```diff
@@ -0,0 +1,32 @@
...
+  "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+
+<suppressions>
+  <!--
+   Suppress IllegalImport in all files under plugins/spark/v3.5/spark.
+   The spark client relies on the shaded libraries from iceberg-spark-runtime,
+   and therefore uses imports like org.apache.iceberg.shaded.* in the code,
+   which is intended.
+  -->
+  <suppress checks="IllegalImport" files="plugins/spark/v3.5/spark/.*"/>
```
dimas-b: Is it possible to keep this file under the Spark module as opposed to the global `codestyle` dir?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/rest/GenericTable.java
dimas-b: I mean: do we need the `@Generated` annotation here? The file appears to be modified after generation and manually committed to git :thinking:

codestyle/checkstyle.xml
gh-yzou: it is not the same, actually line 41-42 is not doing anything, the config says look for config `org.checkstyle.google.suppressionfilter.config`, which i don't think we configure it anywhere in our project, if not configured, use file checkstyle-suppressions.xml, which we do not have it anywhere in the project also, the check style job doesn't fail, because it sets optional to true. 
Now since we do have the suppressions file, i made it non-optional, and removed the other configuration. One thing i did notice here is that if i just use the relative path regarding to root project path, gradlew build seems not able to found the path, the recommended way seems using the config, and then get the absolute path, which is the way i am going with now

codestyle/checkstyle_suppressions.xml
gh-yzou: Actually, i think it make sense to have it at the top level, in the future, if there are other suppress rule we want to add, they can all be added to this file, and we will have centralized file to mange all checkstyle and suppress rule, which i think can make things much easier to mange.

build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -47,6 +47,11 @@ checkstyle {
       .requiredVersion
   toolVersion = checkstyleVersion
   configFile = rootProject.file("codestyle/checkstyle.xml")
+  configProperties =
+    mapOf(
+      "checkstyle.suppression.file" to
+        rootProject.file("codestyle/checkstyle_suppressions.xml").absolutePath
```
snazy: Please do not add a project specific change to all projects?

codestyle/checkstyle_suppressions.xml
snazy: Agree with @dimas-b , let's not add module specifics here globally - that _does_ affect all modules

plugins/spark/v3.5/integration/build.gradle.kts
```diff
@@ -64,13 +68,7 @@ dependencies {
...
 
   testImplementation(platform(libs.jackson.bom))
-  testImplementation("com.fasterxml.jackson.core:jackson-annotations")
-  testImplementation("com.fasterxml.jackson.core:jackson-core")
-  testImplementation("com.fasterxml.jackson.core:jackson-databind")
-
-  testImplementation(
-    "org.apache.iceberg:iceberg-spark-runtime-${sparkMajorVersion}_${scalaVersion}:${icebergVersion}"
-  )
+  testImplementation("com.fasterxml.jackson.jakarta.rs:jackson-jakarta-rs-json-provider")
```
snazy, line 71: Why have two Jackson in tests - the relocated one and this?

codestyle/checkstyle_suppressions.xml
gh-yzou: Moved the suppress config to spark client side.

codestyle/checkstyle.xml
gh-yzou: i moved the suppress file to the polaris-spark project, and reuses the original configuration

build-logic/src/main/kotlin/polaris-java.gradle.kts
gh-yzou: moved to project specific

plugins/spark/v3.5/integration/build.gradle.kts
gh-yzou: That one is used by a polaris client test utility that is used to talk to the polaris management API for test setup, such as createCatalog. This test utility has nothing to do with Iceberg, and is not suppose to rely on the iceberg spark client library.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/rest/GenericTable.java
gh-yzou: i see. yes, we do not need the @Generated annotation here, removed.

flyrain (approved):
Thanks for keeping working on it, @gh-yzou !

codestyle/checkstyle_suppressions.xml
dimas-b: thx!

dimas-b (commented):
I think this that compiling against shaded jackson classes is sub-optimal and can be avoided, but I will not hold this PR because of that. Other changes LGTM.

Issue #1530, opened 2025-05-06
Support for non-AWS S3 compatible storage


dimas-b:
### Is your feature request related to a problem? Please describe.

There are multiple implementations of the S3 API offered by non-AWS systems.

This issues tracks adding support for that in Polaris.

Likely multiple PRs will be required.

lastranget:
When this support is fully provided, will it be required that custom s3 providers support STS for Polaris to be compatible? I ask because I was able to point a vended-credentials gravitino iceberg rest catalog at our s3 provider (which is lacking sts support), but I haven't had success with Polaris yet (built from latest main). I don't understand how this was possible, as #1913 implies that STS is required part of the vended-credentials process to me, so if my assumption is incorrect, I'd appreciate someone correcting me. I had a bug report related to this, but I closed it #2207

dimas-b:
Work done under this issues is still based on STS.

That said, I support implementing non-STS based flows for vending credentials (requires a proposal + `dev` list discussion) and Request Signing (well-specified, but will probably need to be implemented under a separate issue).

metadaddy:
Another vote for a non-STS flow - there are several S3-compatible object stores that do not provide an STS. I work at [Backblaze](https://www.backblaze.com/cloud-storage), one such provider, and I'm happy to help in any way I can.

dimas-b:
I started a `dev` discussion for this: https://lists.apache.org/thread/wtgbhhmd1pw94lbg56ds9p4z275ncfsq

@metadaddy @lastranget : Please consider replying to that email thread with your thought.

dimas-b:
#2213 is the last code change I'm planning to contribute to this issue.

When it is merged, I'll follow up with a docs update and that should hopefully be enough to get Polaris working with MinIO. 

As for other S3-compatible systems let's discuss on the `dev` email list or feel free to open more specific GH issues.

sathesuraj:
@dimas-b - Version 1.1.0 has not been released yet. Could you please share when we can expect this feature to be available?

dimas-b:
Voting thread: https://lists.apache.org/thread/kggv6h13q2dq2cvsc9j62kd7ykf1cz7w

PR #389, opened 2024-10-21
[FEATURE REQUEST #32] On-Premises S3 / S3 Compatible...

lefebsy:
# Description (edited) :

This is a S3 proposition of Polaris core storage implementation, copy of the aws + new parameters : endpoint, path style...

It is tested OK with :
- local MinIO
- Backblaze B2 (thanks to @metadaddy)
- Ceph.io seems also ok  thanks to @mmgaggle & @Gerrit-K tests

This should works with many S3 compatible solutions like Dell ECS, NetApp StorageGRID, etc...

- By default it is trying to respect the same behavior about credentials than AWS (IAM/STS). The same dynamic policy is applied, limiting the scope to the data queried.

- Otherwise if STS is not available 'skipCredentialSubscopingIndirection' = true will disabling Polaris "SubScoping" of the credentials
 
Let me know your opinion about this design proposal.
Thank you

```
curl -X POST -H "Authorization: Bearer ${SPARK_BEARER_TOKEN}" \
     -H 'Accept: application/json' -H 'Content-Type: application/json' \
     http://${POLARIS_HOST}:8181/api/management/v1/catalogs -d \
      '{
          "name": "my-s3compatible-catalog",
          "id": 100,
          "type": "INTERNAL",
          "readOnly": false,
          "properties": {
            "default-base-location": "${S3_LOCATION}"
          },
          "storageConfigInfo": {
            "storageType": "S3_COMPATIBLE",
            "allowedLocations": ["${S3_LOCATION}/"],
            "s3.endpoint": "https://localhost:9000"
          }
        }'
```
```
            # optional:
            "s3.pathStyleAccess": true / false
            "s3.region": "rack-1 or us-east-1"
            "s3.roleArn": "arn:xxx:xxx:xxx:xxxx"
            "s3.credentials.catalog.accessKeyId": "CATALOG_1_ACCESS_KEY_ENV_VARIABLE_NAME"
            "s3.credentials.catalog.secretAccessKey": "CATALOG_1_SECRET_KEY_ENV_VARIABLE_NAME"
```
```            
            # optional in case STS/IAM is not available :
            "skipCredentialSubscopingIndirection": true
                # optional for skipCredentialSubscopingIndirection - served by catalog to client like spark or trino
                "s3.credentials.client.accessKeyId": "CLIENT_OF_CATALOG_1_ACCESS_KEY_ENV_VARIABLE_NAME"
[...]

Issue #32, opened 2024-07-31
[FEATURE REQUEST] On-Premise S3 & Remote Signing


c-thiel:
**Is your feature request related to a problem? Please describe.**
Currently Polaris only works for AWS S3. It would be great to get support for on-prem deployments as well!

**Describe the solution you'd like**
Add an additional storage profile similar to the [AWS one](https://github.com/polaris-catalog/polaris/blob/bb544239f0ca3bcd076cf2e93116b2747e289e10/spec/polaris-management-service.yml#L880) which allows custom Endpoint configuration. Test with MinIO or Ceph. Grant access via remote signing.

**Describe alternatives you've considered**
I don't think there is any?

**Additional context**
Remote signing spec: https://github.com/apache/iceberg/blob/main/aws/src/main/resources/s3-signer-open-api.yaml

mjf-89:
It would be great to know if such feature is already on the roadmap or not. I would be personally interested in contributing here because currently I'm working for a company where we have everything on-premise. We are using Iceberg with a legacy Hive Standalone Metastore and we are looking for a REST alternative. Polaris is really promising but the lack of support for on premise S3 provider will hinder its adoption.

chris922:
When I watched the demo on Snowflake site this was the first thing I noticed - where to configure the S3 endpoint etc.

I can also support here, maybe in development but also testing it with some S3 alternatives. I've got access to Dell ECS, NetApp StorageGRID, MinIO

guitcastro:
The main point is that only few S3 compatible services have support for Security Token Service (STS). Minio does have support for it.

mjf-89:
@guitcastro It might be possible to implement remote signing for on prem S3 implementations other than minio. But that would mean implement also the remote signing open API spec and that is probably outside the scope of polaris?

guitcastro:
@mjf-89 I don't know more than you. I am not maintainer, my comment is just based on how the S3 auth are implemented.

dimas-b:
@mjf-89 : could you provide more details (perhaps a list) about the `remote signing open API spec` that you mentioned above?

mjf-89:
@dimas-b sure, in the openapi spec of the rest catalog you can see that there are currently two supported delegated access mechanisms, vended credentials and remote signing:

https://github.com/apache/iceberg/blob/e9364faabcc67eef6c61af2ecdf7bcf9a3fef602/open-api/rest-catalog-open-api.yaml#L1488

[...]

PR #1913, opened 2025-06-19
Add STS clients pool

dimas-b:
No functional change.

Introduce a dedicated interface for StsClient suppliers and implement it using a pool of cached clients.

All client are "thin" and share the same `SdkHttpClient`. The latter is closed when the server shuts down.

This is a step towards supporting non-AWS S3 storage (#1530). For this reason the STS endpoint is present in new interfaces, but is not used yet.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/StsClientSupplier.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/StsClientSupplier.java
@@ -16,31 +16,30 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.polaris.service.quarkus.storage;
 
-import io.smallrye.config.ConfigMapping;
-import io.smallrye.config.WithName;
-import java.time.Duration;
+package org.apache.polaris.core.storage.aws;
+
+import jakarta.annotation.Nullable;
+import java.net.URI;
 import java.util.Optional;
-import org.apache.polaris.service.storage.StorageConfiguration;
-import org.apache.polaris.service.storage.aws.S3AccessConfig;
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
+import software.amazon.awssdk.services.sts.StsClient;
 
-@ConfigMapping(prefix = "polaris.storage")
-public interface QuarkusStorageConfiguration extends StorageConfiguration, S3AccessConfig {
+public interface StsClientSupplier {
 
-  @WithName("aws.access-key")
-  @Override
-  Optional<String> awsAccessKey();
+  StsClient stsClient(StsDestination destination);
 
-  @WithName("aws.secret-key")
-  @Override
-  Optional<String> awsSecretKey();
+  @PolarisImmutable
+  interface StsDestination {
+    @Value.Parameter(order = 1)
+    Optional<URI> endpoint();
 
-  @WithName("gcp.token")
-  @Override
-  Optional<String> gcpAccessToken();
+    @Value.Parameter(order = 2)
+    Optional<String> region();
 
-  @WithName("gcp.lifespan")
-  @Override
-  Optional<Duration> gcpAccessTokenLifespan();
+    static StsDestination of(@Nullable URI endpoint, @Nullable String region) {
+      return ImmutableStsDestination.of(Optional.ofNullable(endpoint), Optional.ofNullable(region));
+    }
+  }
 }
--- a/runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
@@ -18,6 +18,7 @@
  */
 package org.apache.polaris.service.quarkus.config;
 
+import io.micrometer.core.instrument.MeterRegistry;
 import io.smallrye.common.annotation.Identifier;
 import io.smallrye.context.SmallRyeManagedExecutor;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -75,12 +76,16 @@ import org.apache.polaris.service.quarkus.ratelimiter.QuarkusTokenBucketConfigur
 import org.apache.polaris.service.quarkus.secrets.QuarkusSecretsManagerConfiguration;
 import org.apache.polaris.service.ratelimiter.RateLimiter;
 import org.apache.polaris.service.ratelimiter.TokenBucketFactory;
+import org.apache.polaris.service.storage.aws.S3AccessConfig;
+import org.apache.polaris.service.storage.aws.StsClientsPool;
 import org.apache.polaris.service.task.TaskHandlerConfiguration;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.context.ManagedExecutor;
 import org.eclipse.microprofile.context.ThreadContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.http.SdkHttpClient;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
 
 public class QuarkusProducers {
   private static final Logger LOGGER = LoggerFactory.getLogger(QuarkusProducers.class);
@@ -168,6 +173,34 @@ public class QuarkusProducers {
       QuarkusSecretsManagerConfiguration config,
       @Any Instance<UserSecretsManagerFactory> userSecretsManagerFactories) {
     return userSecretsManagerFactories.select(Identifier.Literal.of(config.type())).get();
+  }
+
+  @Produces
+  @Singleton
+  @Identifier("http-client-s3")
+  public SdkHttpClient sdkHttpClient(S3AccessConfig config) {
+    ApacheHttpClient.Builder httpClient = ApacheHttpClient.builder();
+    config.maxHttpConnections().ifPresent(httpClient::maxConnections);
+    config.readTimeout().ifPresent(httpClient::socketTimeout);
+    config.connectTimeout().ifPresent(httpClient::connectionTimeout);
+    config.connectionAcquisitionTimeout().ifPresent(httpClient::connectionAcquisitionTimeout);
+    config.connectionMaxIdleTime().ifPresent(httpClient::connectionMaxIdleTime);
+    config.connectionTimeToLive().ifPresent(httpClient::connectionTimeToLive);
+    config.expectContinueEnabled().ifPresent(httpClient::expectContinueEnabled);
+    return httpClient.build();
+  }
+
+  public void closeSdkHttpClient(@Disposes @Identifier("http-client-s3") SdkHttpClient client) {
+    client.close();
+  }
+
+  @Produces
+  @ApplicationScoped
+  public StsClientsPool stsClientsPool(
+      @Identifier("http-client-s3") SdkHttpClient httpClient,
+      S3AccessConfig config,
+      MeterRegistry meterRegistry) {
+    return new StsClientsPool(config, httpClient, meterRegistry);
   }
 
   /**
--- a/runtime/service/src/main/java/org/apache/polaris/service/storage/aws/S3AccessConfig.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/storage/aws/S3AccessConfig.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.service.storage.aws;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.OptionalInt;
+
+public interface S3AccessConfig {
+  /** Default value for {@link #sessionCacheMaxSize()}. */
+  int DEFAULT_MAX_SESSION_CREDENTIAL_CACHE_ENTRIES = 1000;
+
+  /** Default value for {@link #clientsCacheMaxSize()}. */
+  int DEFAULT_MAX_STS_CLIENT_CACHE_ENTRIES = 50;
+
+  /** Default value for {@link #sessionGracePeriod()}. */
+  Duration DEFAULT_SESSION_REFRESH_GRACE_PERIOD = Duration.ofMinutes(5);
+
+  /**
+   * The time period to subtract from the S3 session credentials (assumed role credentials) expiry
+   * time to define the time when those credentials become eligible for refreshing.
+   */
+  Optional<Duration> sessionGracePeriod();
+
+  default Duration effectiveSessionGracePeriod() {
+    return sessionGracePeriod().orElse(DEFAULT_SESSION_REFRESH_GRACE_PERIOD);
+  }
+
+  /**
+   * Maximum number of entries to keep in the session credentials cache (assumed role credentials).
+   */
+  OptionalInt sessionCacheMaxSize();
+
+  default int effectiveSessionCacheMaxSize() {
+    return sessionCacheMaxSize().orElse(DEFAULT_MAX_SESSION_CREDENTIAL_CACHE_ENTRIES);
+  }
+
+  /** Maximum number of entries to keep in the STS clients cache. */
+  OptionalInt clientsCacheMaxSize();
+
+  default int effectiveClientsCacheMaxSize() {
+    return clientsCacheMaxSize().orElse(DEFAULT_MAX_STS_CLIENT_CACHE_ENTRIES);
+  }
+
+  /** Override the default maximum number of pooled connections. */
+  OptionalInt maxHttpConnections();
+
+  /** Override the default connection read timeout. */
+  Optional<Duration> readTimeout();
+
+  /** Override the default TCP connect timeout. */
+  Optional<Duration> connectTimeout();
+
+  /**
+   * Override default connection acquisition timeout. This is the time a request will wait for a
+   * connection from the pool.
+   */
+  Optional<Duration> connectionAcquisitionTimeout();
+
+  /** Override default max idle time of a pooled connection. */
+  Optional<Duration> connectionMaxIdleTime();
+
+  /** Override default time-time of a pooled connection. */
+  Optional<Duration> connectionTimeToLive();
+
+  /** Override default behavior whether to expect an HTTP/100-Continue. */
+  Optional<Boolean> expectContinueEnabled();
+}
--- a/runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.service.storage.aws;
+
+import static java.util.Collections.singletonList;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.stats.StatsCounter;
+import com.google.common.annotations.VisibleForTesting;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.cache.CaffeineStatsCounter;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import org.apache.polaris.core.storage.aws.StsClientSupplier;
+import software.amazon.awssdk.endpoints.Endpoint;
+import software.amazon.awssdk.http.SdkHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.StsClientBuilder;
+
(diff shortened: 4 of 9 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/StsClientSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.polaris.core.storage.aws;
+
+import jakarta.annotation.Nullable;
+import java.net.URI;
+import java.util.Optional;
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
+import software.amazon.awssdk.services.sts.StsClient;
+
+public interface StsClientSupplier {
```
eric-maynard: Perhaps this is rather a `Factory`? At first, I thought `StsClientSupplier` would be a `Supplier<StsClient>`. Even if it's not literally this type it's also not logically this type because it needs an input `StsDestination` to get you a `StsClient`.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -170,6 +175,34 @@ public UserSecretsManagerFactory userSecretsManagerFactory(
...
+    config.readTimeout().ifPresent(httpClient::socketTimeout);
+    config.connectTimeout().ifPresent(httpClient::connectionTimeout);
+    config.connectionAcquisitionTimeout().ifPresent(httpClient::connectionAcquisitionTimeout);
+    config.connectionMaxIdleTime().ifPresent(httpClient::connectionMaxIdleTime);
+    config.connectionTimeToLive().ifPresent(httpClient::connectionTimeToLive);
+    config.expectContinueEnabled().ifPresent(httpClient::expectContinueEnabled);
+    return httpClient.build();
+  }
+
+  public void closeSdkHttpClient(@Disposes @Identifier("http-client-s3") SdkHttpClient client) {
```
eric-maynard: This feels out of place in `QuarkusProducers`. Maybe we could create a wrapper over `SdkHttpClient` and handle that like our other `Closeable`s or use `AutoCloseable`

eric-maynard:
I took a quick look -- pooling the STS clients seems like a reasonable thing to do and this approach to doing so looks pretty correct to me. However it's not totally clear to me why this is needed to implement #32. Looking at #389, this part of the code wasn't touched at all.

Could we instead start more top-down with S3-compatible storage support and figure out if these changes are actually necessary?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/S3AccessConfig.java
```diff
@@ -0,0 +1,85 @@
...
+
+package org.apache.polaris.service.storage.aws;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.OptionalInt;
+
+public interface S3AccessConfig {
+  /** Default value for {@link #sessionCacheMaxSize()}. */
+  int DEFAULT_MAX_SESSION_CREDENTIAL_CACHE_ENTRIES = 1000;
```
eric-maynard: nit: the convention used elsewhere seems to be `..._DEFAULT` not `DEFAULT_...`

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/S3AccessConfig.java
```diff
@@ -0,0 +1,85 @@
...
+ * under the License.
+ */
+
+package org.apache.polaris.service.storage.aws;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.OptionalInt;
+
+public interface S3AccessConfig {
```
eric-maynard, line 33: Would be nice to have a Javadoc on the new types

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.polaris.core.storage.aws.StsClientSupplier;
+import software.amazon.awssdk.endpoints.Endpoint;
+import software.amazon.awssdk.http.SdkHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.StsClientBuilder;
+
+/** Maintains a pool of STS clients. */
+public class StsClientsPool implements StsClientSupplier {
+  // CODE_COPIED_TO_POLARIS from Project Nessie 0.104.2
```
eric-maynard, line 44: Why are we copying such generic code from another project?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
```diff
@@ -0,0 +1,101 @@
...
+          completedFuture(Endpoint.builder().url(parameters.endpoint().get()).build());
+      builder.endpointProvider(params -> endpointFuture);
+    }
+
+    parameters.region().ifPresent(r -> builder.region(Region.of(r)));
+
+    return builder.build();
+  }
+
+  static StatsCounter statsCounter(Optional<MeterRegistry> meterRegistry, int maxSize) {
```
eric-maynard, line 91: Is this really related to S3-compatible storage support?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @Override
+  public StsClient stsClient(StsDestination destination) {
+    return clients.get(destination, clientBuilder);
+  }
+
+  private static StsClient defaultStsClient(StsDestination parameters, SdkHttpClient sdkClient) {
+    StsClientBuilder builder = StsClient.builder();
+    builder.httpClient(sdkClient);
```
eric-maynard, line 79: This is the only usage of `SdkHttpClient` outside of `QuarkusProducers` itself, I wonder if we can simplify things by just injecting the `StsClientBuilder`

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -170,6 +175,34 @@ public UserSecretsManagerFactory userSecretsManagerFactory(
     return userSecretsManagerFactories.select(Identifier.Literal.of(config.type())).get();
   }
 
+  @Produces
+  @Singleton
+  @Identifier("http-client-s3")
+  public SdkHttpClient sdkHttpClient(S3AccessConfig config) {
```
singhpk234, line 181: if we are having shared connection pool, which will be used by multiple clients which i assume have different routes , wouldnt we want to restrict per route max connection ? to avoid starvation ?

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -170,6 +175,34 @@ public UserSecretsManagerFactory userSecretsManagerFactory(
...
+    ApacheHttpClient.Builder httpClient = ApacheHttpClient.builder();
```
singhpk234, line 182: why only ApacheHttpClient  and not UrlConnectionHttpClient >

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
```diff
@@ -0,0 +1,101 @@
...
+    if (parameters.endpoint().isPresent()) {
+      CompletableFuture<Endpoint> endpointFuture =
+          completedFuture(Endpoint.builder().url(parameters.endpoint().get()).build());
```
singhpk234, line 82: if we already know the url why do we need to wrap it in the future ?

singhpk234 (commented):
why do we need the pool

singhpk234 (commented):
my comment doesn't shows up linking here 
1. https://github.com/apache/polaris/pull/1913#discussion_r2157998980
2. https://github.com/apache/polaris/pull/1913#discussion_r2158001989
3. https://github.com/apache/polaris/pull/1913#discussion_r2158007526

dimas-b:
> Could we instead start more top-down with S3-compatible storage support and figure out if these changes are actually necessary?

I can do that, but the diff is going to be larger ;)

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/StsClientSupplier.java
dimas-b: Does `Provider` sound ok? A "factory" feels like it makes something new every time, while this interface is intended to surface a pool of clients.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: What do you mean by "out of place"? `QuarkusProducers` creates a singleton client, `QuarkusProducers` also closes that same client. Who would call `.close()` on the wrapper?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: Because it is convenient as opposed to writing it from scratch.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: Stats are not directly relates to supporting S3-compatible storage. However, the STS clients pool is. Having a pool in the system, it is good to expose metrics about it for observability.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: `StsClientsPool` is injected where it is needed. `SdkHttpClient` is injected into `StsClientsPool`. Are you suggesting to hide the `SdkHttpClient` from CDI? If we did that, it would not be possible to close the singleton `SdkHttpClient` using CDI callbacks.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
singhpk234: if we are using AWS ask it has metrics inbuilt https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/metrics-list.html is the current mechanism more comprehensive than that ?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: If the suggestion is to make `SdkHttpClient` an inner (owned) fields in `StsClientsPool` and close the pool, IMHO, that introduces unnecessary coupling, because the pool only needs the interface (namely `SdkHttpClient`), but `QuarkusProducers` chooses a specific implementation of that interface.

dimas-b:
@singhpk234 : Sorry, I do not see your comments (even via above links). Could you re-summit them?

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
singhpk234, line 181: if HTTP connection pool is shared, shouldn't we restrict max connection per route ? as if lets say my pool size is 50 and for route 1 hit with 50 concurrent request so pool now only has 50 connection of route 1 and if now route 2 comes it starves ? I am not sure how to do it here HTTPClient use to PoolingConnectionManager

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
singhpk234, line 182: why only ApacheHttpClient and not URLConnection client ?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: these stats are about the cache of STS clients we keep in Polaris code.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
singhpk234, line 82: why do we need a future if we already know the URL ? is it still an expensive operations ?

singhpk234 (commented):
Sure, i resubmitted !

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
singhpk234: I see, we just plotting a gauge for max clients, wouldn't it be better to have the stats for the http connection pool ? for ex lets say i have 2 entires in the cache and one entry hogs my 49 connection and other only 1 i would want to know who hogs 49 connection ?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/S3AccessConfig.java
dimas-b: added

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/StsClientSupplier.java
dimas-b: renamed to `StsClientProvider`

dimas-b:
Marking this PR as "draft" to add end-to-end support for MinIO. Please feel free to comment anyway.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: Is the AWS SDK not able to deal with this situation on its side?

Did you observe this kind of starvation in practice?

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: I can make it configurable in runtime if you prefer. Let's do that as a follow-up PR, though.

dimas-b: `ApacheHttpClient` is more capable in general, AFAIK.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: It's what the AWS SDK builder wants as a parameter :shrug:

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
singhpk234: Iceberg supports both, hence coming from there, Its good to have alternative specially considering how they behave. for example The Apache HTTP connection will always read the [entire stream](https://github.com/apache/httpcomponents-core/blob/2135596b1319910d56efb148566ecbbec0884638/httpcore/src/main/java/org/apache/http/impl/io/ContentLengthInputStream.java#L108) on close, which results in much more data being read than needed. The URL connection client does not behave this way.  for more details : https://github.com/apache/iceberg/pull/7262

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
singhpk234: This is practically possible, i think since we don't specify it default kicks in from http client, TBH i haven't hit in practice as in AWS EMR spark jobs generally have one endpoint to connect to, but with a long running service like polaris, i think the possibility being increased, WDYT ?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: Stats for the HTTP client would be nice to have for sure, but let's defer it to another PR. I do not have that code handy ATM and I would not want to bloat the scope of this PR even more :)

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: Certainly. If do not mind supporting both, but let's do that in a follow-up PR to avoid bloating the scope of this PR (the main idea in this PR is not the ApacheHttpClient client, but supporting multiple endpoints).

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: @snazy WDYT?

dimas-b:
`CatalogSerializationTest` currently fails because unset `endpoint` values serialize as `null`. This is technically a breaking change for older clients. I'll see how it can be fixed later (under this PR). Please review the main change for now.

dimas-b:
Prerequisite PR: #1955

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -170,6 +175,34 @@ public UserSecretsManagerFactory userSecretsManagerFactory(
     return userSecretsManagerFactories.select(Identifier.Literal.of(config.type())).get();
   }
 
+  @Produces
+  @Singleton
+  @Identifier("http-client-s3")
```
snazy: Maybe
```suggestion
  @Identifier("sdk-http-client")
```

snazy (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: Updated to `aws-sdk-http-client`

snazy (approved):

dimas-b:
@eric-maynard PTAL :)

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
flyrain: I'm curious what will happen if anyone change this class later? Should we keep `CODE_COPIED_TO_POLARIS ` for ever?

To reduce the maintenance burden, I'd suggest to rewrite to avoid copying from another project.

flyrain:
Thanks for doing this. I think it's the right direction. We could also make CLI to support new fields, which could be done as a followup.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: OSS is meant to be reused. Rewriting to avoid attribution is against the grain of OSS concepts.

dimas-b:
This is not the final PR for non-AWS S3 :) We should certainly add CLI support, but probably in a follow-up PR.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: Re: forever: I'd guess it's very likely as I do not personally see a reason for rewriting this code in any substantial way in the foreseeable future.

snazy: @eric-maynard @flyrain We have a _ton_ of code taken from other projects _already_ in the code base. Most of that was already part of the initial contribution to Apache. A lot of that was also "generic" and literally copies of rather trivial code. So I seriously do not understand what you're trying to say here. The goal should really _not_ be to reinvent "all the wheels".

snazy (approved):

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
eric-maynard: What’s an example of similarly generic code that’s been copied?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
eric-maynard: Yes, that is the suggestion. They are always coupled together, so encapsulating the client inside the pool simplifies usage of the pool.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
```diff
@@ -0,0 +1,101 @@
...
+public class StsClientsPool implements StsClientProvider {
+  // CODE_COPIED_TO_POLARIS from Project Nessie 0.104.2
+
+  private static final String CACHE_NAME = "sts-clients";
+
+  private final Cache<StsDestination, StsClient> clients;
+  private final Function<StsDestination, StsClient> clientBuilder;
+
+  public StsClientsPool(
+      S3AccessConfig effectiveSts, SdkHttpClient sdkHttpClient, MeterRegistry meterRegistry) {
```
flyrain: Can we pass `effectiveSts` as an integer instead of passing the whole S3AccessConfig object?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: Currently the pool and the HTTP client are not coupled. The pool needs a client. Yes, the client can be reused by other parts of Polaris (if it becomes necessary in the future).

The pool is easily injectable in the current state of the PR. Use sites need not be aware of how the client is constructed.

I'd like to keep the current approach of having distinct CDI beans for the HTTP client and the STS client pool.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: certainly - updated.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
flyrain: It's always a good practice to reduce the dependency for code. In this case, we have to add additional information in the LICENSE. If rewriting isn't a big task which I believe so, we mighty avoid extra maintenance burden. That's also just a suggestion, I'm NOT gonna block this PR for that.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: Also, as @singhpk234 commented in another thread, we're going to support multiple HTTP client implementations (later), selectable in runtime (by config).

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java
dimas-b: License checks for `CODE_COPIED_TO_POLARIS` marks are automated - there's minimal cognitive effort involved in maintaining these license references (plus they are fixed and do not need to change over time).

snazy: There is, as mentioned, a lot of other projects' code in Polaris (OpenAPI generator files, the web site, code copied from Iceberg, build scripts, etc etc etc).

There is _no_ ongoing maintenance burden, but there is a huge plus taking code that's tested and known to work in production for years. New developed code is _way more_ work.

snazy (approved):

PR #1919, opened 2025-06-20
[Docs] Rollback compaction on Conflict documentation

singhpk234:
### About the change 

Adds the docs  for rolling back commit conflict, the commit message has been thorough https://github.com/apache/polaris/commit/793cf086f1f784c36296b1ea75298a08ce608ef2

please review with turn off whitespace, i am not sure why, but my IJ reformatted it


flyrain:
Thanks @singhpk234 for adding the config! Not a blocker: I think it's also worth to add a description how the feature works! We can do that in a followup PR.

PR #1920, opened 2025-06-20
Allow for Customization of Polaris Authorizer

travis-bowen:
Want to be able to allow for Polaris Authorizer to be customizable at the request level, while still maintaining existing behavior of a single application level implementation.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
@@ -38,7 +38,7 @@ import org.apache.polaris.core.PolarisDefaultDiagServiceImpl;
 import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.auth.AuthenticatedPolarisPrincipal;
 import org.apache.polaris.core.auth.PolarisAuthorizer;
-import org.apache.polaris.core.auth.PolarisAuthorizerImpl;
+import org.apache.polaris.core.auth.PolarisAuthorizerFactory;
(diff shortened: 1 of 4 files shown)


dimas-b (commented):
Thanks for your interest and contribution, @travis-bowen !

Your code changes make sense to me, however, if your intention is to have an alternative `PolarisAuthorizer` in a downstream build (not in this repo), I believe these changes are not necessary. The downstream build could have an overriding producer method like this:
```
  @Produces
  @Alternative
  @Priority(10)
  @Singleton
  public PolarisAuthorizer polarisAuthorizer(...) {...}
```
to cause a different implementation to be used. You can have any config and other beans injected into the producer method depending on its needs. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -101,9 +101,16 @@ public StorageCredentialCache storageCredentialCache(
   }
 
   @Produces
-  @ApplicationScoped
-  public PolarisAuthorizer polarisAuthorizer(PolarisConfigurationStore configurationStore) {
-    return new PolarisAuthorizerImpl(configurationStore);
+  @RequestScoped
```
travis-bowen, line 104: If this is contentious I can probably change back to ApplicationScoped and utilize the call context within my own implementation to do different things per request.

travis-bowen:
Thanks for the input @dimas-b ! I'm pretty new to the Quarkus stuff so let me try that out!

travis-bowen:
Custom producer worked, so closing this out!

PR #1921, opened 2025-06-21
Revert "Reuse shadowJar for spark client bundle jar maven publish (#1…

gh-yzou:
…857)"

This reverts commit 1f7f127536a088911bf940addd1d05c07ff99a68.

The shadowJar plugin actually stops publish the original jar, which is not what spark client intend to publish for the --package usage. 

Revert it for now, will follow up with a better way to reuse the shadow jar plugin, likely with a separate bundle project


gh-yzou:
@flyrain yes, once i got those the two PRs handled, i can take care of them together for 1.0

PR #1922, opened 2025-06-23
fix(build): Gradle caching effectively not working

snazy:
Using a `custom()` spotless formatter check effectively disables caching, see `com.diffplug.gradle.spotless.FormatExtension#custom(java.lang.String, com.diffplug.spotless.FormatterFunc)` using `globalState`, which is a `NeverUpToDateBetweenRuns`. This change refactors this to be cachable.


adutra:
We already have an errorprone rule for wildcard imports:

https://github.com/apache/polaris/blob/0faf948c482ca1177da8e3016bfaa28b85f0f3ea/codestyle/errorprone-rules.properties#L298

So I'm wondering: do we need the Spotless formatter?

snazy:
The spotless rule was always there. But yea - in that case we can probably remove the spotless one.

adutra (approved):

PR #1506, opened 2025-05-01
Spec: Add SigV4 Auth Support for Catalog Federation

XJDKC:
### Prior Catalog Federation work:
* https://github.com/apache/polaris/pull/1026
* https://github.com/apache/polaris/pull/1305

### Reference

Below are all the reference mentioned in this PR:

* AWS Docs:
  * AWS SigV4 protocol: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html
  * Configure IAM for Glue: https://docs.aws.amazon.com/glue/latest/dg/configure-iam-for-glue.html
  * [AWS STS Regions and endpoints](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_region-endpoints.html)
* Iceberg SDK 
  * [AuthProperties.java#30](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/auth/AuthProperties.java#L30)
  * [RESTSigV4Signer.java#L91](https://github.com/apache/iceberg/blob/apache-iceberg-1.9.0/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4Signer.java#L91)
  * [AwsProperties.java#L417-L435](https://github.com/apache/iceberg/blob/apache-iceberg-1.9.0/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java#L417-L435)
  * [RESTSigV4AuthSession.java#L71](https://github.com/apache/iceberg/blob/07ad9a635c6e24959280f53b76ef3036e33d13ee/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L71)
* Polaris Code
  * param to skip assuming IAM role: [SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION](https://github.com/apache/polaris/blob/4db7998381a61e9cab82cdc4fded6867b0bca464/service/common/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java#L92)
  * roleSessionName of the StorageConfig: https://github.com/apache/polaris/blob/ca35bbaf10aff807053bfc032f8e71a3f007cc79/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java#L68

## Description
This PR adds the SigV4 Auth Support so that Polaris can federate to Glue IRC / S3Tables / catalog server hosted behind the AWS API Gateway.

[...]

PR #1925, opened 2025-06-24
Support IMPLICIT authentication type for federated catalogs

poojanilangekar:
Previously, the ConnectionConfigInfo required explicit AuthenticationParameters for every federated catalog. However, certain catalogs types that Polaris federates to (either now or in the future) allow `IMPLICIT` authentication, wherein the authentication parameters are picked from the environment or configuration files. This change enables federating to such catalogs without passing dummy secrets. 

The `IMPLICIT` option is guarded by the `SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES`. Hence users may create federated catalogs with `IMPLICIT` authentication only when the administrator explicitly enables this feature.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -256,13 +256,6 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
           .defaultValue(false)
           .buildFeatureConfiguration();
 
-  public static final FeatureConfiguration<Boolean> ENABLE_CATALOG_FEDERATION_NO_AUTH =
-      PolarisConfiguration.<Boolean>builder()
-          .key("ENABLE_CATALOG_FEDERATION_NO_AUTH")
-          .description("If true, allows federation to remote catalogs without authentication.")
-          .defaultValue(false)
-          .buildFeatureConfiguration();
-
   public static final FeatureConfiguration<Boolean> ENABLE_POLICY_STORE =
       PolarisConfiguration.<Boolean>builder()
           .key("ENABLE_POLICY_STORE")
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -256,6 +256,13 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> ENABLE_CATALOG_FEDERATION_NO_AUTH =
+      PolarisConfiguration.<Boolean>builder()
+          .key("ENABLE_CATALOG_FEDERATION_NO_AUTH")
+          .description("If true, allows federation to remote catalogs without authentication.")
+          .defaultValue(false)
+          .buildFeatureConfiguration();
+
   public static final FeatureConfiguration<Boolean> ENABLE_POLICY_STORE =
       PolarisConfiguration.<Boolean>builder()
           .key("ENABLE_POLICY_STORE")
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
@@ -39,6 +39,7 @@ import org.apache.polaris.core.secrets.UserSecretReference;
 @JsonSubTypes({
   @JsonSubTypes.Type(value = OAuthClientCredentialsParametersDpo.class, name = "1"),
   @JsonSubTypes.Type(value = BearerAuthenticationParametersDpo.class, name = "2"),
+  @JsonSubTypes.Type(value = NoAuthAuthenticationParametersDpo.class, name = "3"),
 })
 public abstract class AuthenticationParametersDpo implements IcebergCatalogPropertiesProvider {
 
@@ -80,6 +81,9 @@ public abstract class AuthenticationParametersDpo implements IcebergCatalogPrope
         config =
             new BearerAuthenticationParametersDpo(
                 secretReferences.get(INLINE_BEARER_TOKEN_REFERENCE_KEY));
+        break;
+      case NO_AUTH:
+        config = new NoAuthAuthenticationParametersDpo();
         break;
       default:
         throw new IllegalStateException(
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
@@ -39,7 +39,7 @@ import org.apache.polaris.core.secrets.UserSecretReference;
 @JsonSubTypes({
   @JsonSubTypes.Type(value = OAuthClientCredentialsParametersDpo.class, name = "1"),
   @JsonSubTypes.Type(value = BearerAuthenticationParametersDpo.class, name = "2"),
-  @JsonSubTypes.Type(value = NoneAuthenticationParametersDpo.class, name = "3"),
+  @JsonSubTypes.Type(value = ImplicitAuthenticationParametersDpo.class, name = "3"),
 })
 public abstract class AuthenticationParametersDpo implements IcebergCatalogPropertiesProvider {
 
@@ -82,8 +82,8 @@ public abstract class AuthenticationParametersDpo implements IcebergCatalogPrope
             new BearerAuthenticationParametersDpo(
                 secretReferences.get(INLINE_BEARER_TOKEN_REFERENCE_KEY));
         break;
-      case NONE:
-        config = new NoneAuthenticationParametersDpo();
+      case IMPLICIT:
+        config = new ImplicitAuthenticationParametersDpo();
         break;
       default:
         throw new IllegalStateException(
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
@@ -39,7 +39,7 @@ import org.apache.polaris.core.secrets.UserSecretReference;
 @JsonSubTypes({
   @JsonSubTypes.Type(value = OAuthClientCredentialsParametersDpo.class, name = "1"),
   @JsonSubTypes.Type(value = BearerAuthenticationParametersDpo.class, name = "2"),
-  @JsonSubTypes.Type(value = NoAuthAuthenticationParametersDpo.class, name = "3"),
+  @JsonSubTypes.Type(value = NoneAuthenticationParametersDpo.class, name = "3"),
 })
 public abstract class AuthenticationParametersDpo implements IcebergCatalogPropertiesProvider {
 
@@ -82,8 +82,8 @@ public abstract class AuthenticationParametersDpo implements IcebergCatalogPrope
             new BearerAuthenticationParametersDpo(
                 secretReferences.get(INLINE_BEARER_TOKEN_REFERENCE_KEY));
         break;
-      case NO_AUTH:
-        config = new NoAuthAuthenticationParametersDpo();
+      case NONE:
+        config = new NoneAuthenticationParametersDpo();
         break;
       default:
         throw new IllegalStateException(
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
@@ -33,6 +33,7 @@ public enum AuthenticationType {
   NULL_TYPE(0),
   OAUTH(1),
   BEARER(2),
+  NO_AUTH(3),
   ;
 
   private static final AuthenticationType[] REVERSE_MAPPING_ARRAY;
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
@@ -33,7 +33,7 @@ public enum AuthenticationType {
   NULL_TYPE(0),
   OAUTH(1),
   BEARER(2),
-  NONE(3),
+  IMPLICIT(3),
   ;
 
   private static final AuthenticationType[] REVERSE_MAPPING_ARRAY;
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
@@ -33,7 +33,7 @@ public enum AuthenticationType {
   NULL_TYPE(0),
   OAUTH(1),
   BEARER(2),
-  NO_AUTH(3),
+  NONE(3),
   ;
 
   private static final AuthenticationType[] REVERSE_MAPPING_ARRAY;
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisAdminServiceAuthzTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisAdminServiceAuthzTest.java
@@ -132,7 +132,8 @@ public class PolarisAdminServiceAuthzTest extends PolarisAuthzTestBase {
             adminService.grantPrivilegeOnRootContainerToPrincipalRole(
                 PRINCIPAL_ROLE2, PolarisPrivilege.CATALOG_DROP))
         .isTrue();
-    final CatalogEntity newCatalog = new CatalogEntity.Builder().setName("new_catalog").build();
+    final CatalogEntity newCatalog =
+        new CatalogEntity.Builder().setName("new_catalog").setCatalogType("INTERNAL").build();
     final CreateCatalogRequest createRequest = new CreateCatalogRequest(newCatalog.asCatalog());
 
     doTestSufficientPrivileges(
@@ -151,7 +152,8 @@ public class PolarisAdminServiceAuthzTest extends PolarisAuthzTestBase {
 
   @Test
   public void testCreateCatalogInsufficientPrivileges() {
-    final CatalogEntity newCatalog = new CatalogEntity.Builder().setName("new_catalog").build();
+    final CatalogEntity newCatalog =
+        new CatalogEntity.Builder().setName("new_catalog").setCatalogType("INTERNAL").build();
     final CreateCatalogRequest createRequest = new CreateCatalogRequest(newCatalog.asCatalog());
 
     doTestInsufficientPrivileges(
@@ -287,7 +289,8 @@ public class PolarisAdminServiceAuthzTest extends PolarisAuthzTestBase {
             adminService.grantPrivilegeOnRootContainerToPrincipalRole(
                 PRINCIPAL_ROLE2, PolarisPrivilege.CATALOG_CREATE))
         .isTrue();
-    final CatalogEntity newCatalog = new CatalogEntity.Builder().setName("new_catalog").build();
+    final CatalogEntity newCatalog =
+        new CatalogEntity.Builder().setName("new_catalog").setCatalogType("INTERNAL").build();
     final CreateCatalogRequest createRequest = new CreateCatalogRequest(newCatalog.asCatalog());
     adminService.createCatalog(createRequest);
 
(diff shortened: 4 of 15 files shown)


poojanilangekar:
CC @eric-maynard  @dennishuo

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
```diff
@@ -33,6 +33,7 @@ public enum AuthenticationType {
   NULL_TYPE(0),
   OAUTH(1),
   BEARER(2),
+  NO_AUTH(3),
```
dimas-b: How is it different from `NULL_TYPE`?

spec/polaris-management-service.yml
```diff
@@ -911,6 +911,7 @@ components:
             - OAUTH
             - BEARER
             - SIGV4
+            - NO_AUTH
```
snazy: Beside that this is a breaking change, any user facing API change requires a discussion on the dev mailing list.

dimas-b: +1 to dev list discussion

eric-maynard: This is not a breaking change btw

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
eric-maynard: With some connection types like HIVE, you may either have or not have an authentication mechanism in place. It's reasonable that admins might want to limit which authentication types can be used for their service (e.g. disallow auth-less federation), but they can't disallow having a `null` authentication type because that is also used by "classic" EXTERNAL catalogs that don't federate at all.

In other words, if you have a non-null connection, you should have a non-null authentication type configured. We should still leave room however for users to be explicit about certain federation modes like HIVE that may not have authentication in use.

spec/polaris-management-service.yml
snazy: It is an `enum` in older clients - and Java enum `valueOf` throws and does not gracefully handle unknown values. Hence it is a breaking change.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -247,6 +247,13 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> ENABLE_CATALOG_FEDERATION_NO_AUTH =
```
eric-maynard: Instead, I think we need a list like `SUPPORTED_CATALOG_CONNECTION_TYPES`

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
```diff
@@ -39,6 +39,7 @@
 @JsonSubTypes({
   @JsonSubTypes.Type(value = OAuthClientCredentialsParametersDpo.class, name = "1"),
   @JsonSubTypes.Type(value = BearerAuthenticationParametersDpo.class, name = "2"),
+  @JsonSubTypes.Type(value = NoAuthAuthenticationParametersDpo.class, name = "3"),
```
eric-maynard: `NONE` might be a better name than `NO_AUTH`?

spec/polaris-management-service.yml
eric-maynard: In the semver sense, I'm not sure that needs to qualify as a breaking change -- at least it hasn't been called out as such when we've made similar changes in the past. The REST API is not breaking, even if some clients implemented in a fragile way might. Our Python client seems to fail quite gracefully. See #1506 and the associated doc [here](https://docs.google.com/document/d/1MAW87DtyHWPPNIEkUCRVUKBGjhh5bPn0GbtV7fifm30/edit?tab=t.0#heading=h.v1rcy31tlwav) for a similar change made recently.

Separately, if you think it's a good idea to harden our Java client against this kind of enum change, I think that would be a great contribution.

snazy: > Separately, if you think it's a good idea to harden our Java client against this kind of enum change, I think that would be a great contribution.

I guess that it would be a hard fight to remove the code generator, mind proposing that change to get that implemented?

eric-maynard: In fact, is the Java client already set up to treat this as NULL_TYPE?

```java
if (authTypeCode < 0 || authTypeCode >= REVERSE_MAPPING_ARRAY.length) {
  return NULL_TYPE;
}
```

snazy: And the other way around?

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
dimas-b: I agree that we need to distinguish defined auth from undefined auth (i.e. `null`). However, a `null` value is not the same as a `NULL_TYPE` value. This is probably a concern about the existing codebase rather than this particular PR :thinking: but I did not find any usages of `NULL_TYPE` in the current codebase.

So, I wonder whether introducing the `NO_AUTH` enum value is really necessary (as opposed to renaming `NULL_TYPE`). WDYT?

poojanilangekar:
Thanks everyone for looking into this PR, I have started a devlist discussion as suggested.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: +1 to `SUPPORTED_CATALOG_CONNECTION_TYPES`

dimas-b: however, if we go with `SUPPORTED_CATALOG_CONNECTION_TYPES`, I'd appreciate if it were introduced for existing auth types in a separate PR (per-requisite to this one) and this PR would then just add new options for the value of that property.

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
dimas-b: +1 to `NONE`

spec/polaris-management-service.yml
eric-maynard: Sorry, what do you mean by the other way around? Does this relate to the doc linked above?

edit: Oh, are you referring to what happens if a client tries to provide this new enum value to an older version of the service? How is that different from a client calling an API that doesn't exist on the service? I don't think that typically qualifies as a breaking change under semver

Regardless of whether or not it is a breaking change however, we do need to discuss the spec change on the ML just like the linked PR

dimas-b: @snazy : the API-level enum is auto-generated from this YAML spec: `AuthenticationTypeEnum`. 
Decoding code: https://github.com/apache/polaris/blob/24683d86b46c63a198b2f7b42725037f32d603b8/polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java#L66
Encoding: https://github.com/apache/polaris/blob/cc7ffb6ae36c15884d2483572146184a145be176/polaris-core/src/main/java/org/apache/polaris/core/connection/OAuthClientCredentialsParametersDpo.java#L120

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
poojanilangekar: From my understanding AuthenticationType (and ConnectionType) should keep the NULL_TYPE to indicate an unsupported/missing value. Specifically to handle the case pointed out in the discussion that follows. If a newer version of AuthenticationType is used to create a federated catalog and the type is persisted in the ConnectionConfigInfoDPO. Later, an older client tries to read this DPO, it will return NULL_TYPE; which is an appropriate response considering it read an unknown/missing type from the DPO.

dimas-b: I question the usefulness of `NULL_TYPE` to represent an "unsupported" value. ATM, I do not see any existing usage of `NULL_TYPE` in Polaris code.

As for REST API clients, they do not use this enum at all, as far as I understand. This enum appears to used by Polaris server code only internally.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -700,16 +700,9 @@ private Map<String, UserSecretReference> extractSecretReferences(
...
-    Catalog catalog = catalogRequest.getCatalog();
-    if (catalog instanceof ExternalCatalog externalCatalog) {
-      if (externalCatalog.getConnectionConfigInfo() != null) {
-        // TODO: Make this more targeted once we have connection configs that don't involve
-        // processing of inline secrets.
-        return true;
-      }
-    }
-    return false;
+  private boolean requiresSecretReferenceExtraction(ConnectionConfigInfo connectionConfigInfo) {
```
harishch1998: nit: Can add @NotNull here..

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationType.java
poojanilangekar: While `NULL_TYPE` is not explicitly used in the codebase, it is implied (and therefore used) in the default case of switch statements. I can go over a more explicit case for `ConnectionType` here:

1. Assume that sometime in the future, Polaris adds support for `HIVE` connections. 
2. The user then creates an external catalog called `hive-catalog` of type `HIVE`. At which point, Polaris will store the corresponding ConnectionConfigInfoDpo into the underlying persistence layer. While creating the persistence object, the constructor similar to `IcebergRestConnectionConfigInfoDpo`, called `HiveConnectionConfigInfoDpo` will create this object with enums defined internally by the Polaris server. It'd make a call of the form `ConnectionType.HIVE.getCode()`. 
3. Further, let's assume that one of the two happens (1) we rollback the hive change the newer versions of Polaris don't support Hive connections, or (2) the Polaris instance is rebooted with an older version. In both cases, the Polaris instance won't recognize the `HIVE` connection type. 
4. Lastly, let's say the user wants to perform some operation on the `hive-catalog`. The `IcebergCatalogHandler::initializeCatalog()` would now read the persisted object and parse it's information, specifically, let's look at this snippet:
```   ConnectionType connectionType =
          ConnectionType.fromCode(connectionConfigInfoDpo.getConnectionTypeCode());     
        switch (connectionType) {
        case ICEBERG_REST:
          SessionCatalog.SessionContext context = SessionCatalog.SessionContext.createEmpty();
          federatedCatalog =
              new RESTCatalog(
                  context,
                  (config) ->
                      HTTPClient.builder(config)
                          .uri(config.get(org.apache.iceberg.CatalogProperties.URI))
                          .build());
          federatedCatalog.initialize(
              ((IcebergRestConnectionConfigInfoDpo) connectionConfigInfoDpo).getRemoteCatalogName(),
              connectionConfigInfoDpo.asIcebergCatalogProperties(getUserSecretsManager()));
          break;
        case HADOOP:
          federatedCatalog = new HadoopCatalog();
          federatedCatalog.initialize(
              ((HadoopConnectionConfigInfoDpo) connectionConfigInfoDpo).getWarehouse(),
              connectionConfigInfoDpo.asIcebergCatalogProperties(getUserSecretsManager()));
          break;
        default:
          throw new UnsupportedOperationException(
              "Connection type not supported: " + connectionType);
      }

```
Since `HIVE` is not a valid connection type in this Polaris instance, the `ConnectionType::fromCode()` function will return `NULL_TYPE` and hence activate the default case. 

This was my understanding of why we need a `NULL_TYPE`. Essentially, since we always want to support not just backward compatibility but also forward compatibility wherein the data store might contain some objects/types that  the current Polaris instance does not recognize. The `NULL_TYPE` provide a nice catch-all type to allow appropriate error handling.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java

polaris-core/src/main/java/org/apache/polaris/core/connection/AuthenticationParametersDpo.java
poojanilangekar: Done.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
poojanilangekar: Done.

poojanilangekar:
Hi everyone, 

@dennishuo suggested that instead of calling the this `AuthenticationType::NONE`, we can call it `AuthenticationType::ENVIRONMENT` because really the motivation behind adding `NONE` was to enable federation without explicitly passing `OAUTH`/`BEARER` tokens. Essentially, this authentication type tells Polaris that when federating to this external catalog don't expect any tokens, instead use whatever information you find in the environment/configuration files. I prefer this over `NONE` because it better represents what is happening under the hood. 


@eric-maynard @dimas-b What do you think?

dimas-b:
I like the idea, but I think term `ENVIRONMENT` is not well-defined. What is implied here looks like it's `PROVIDER`... Meaning whoever makes a catalog with this auth type (user) relies on the Polaris administrator (service provider) to perform some per-arranged kind of auth flow... Did get this right?

poojanilangekar:
>  relies on the Polaris administrator (service provider) to perform some per-arranged kind of auth flow

This part isn't accurate. The Polaris service does not perform any pre-arranged authentication. It relies on the underlying connection library to carry out the authentication flow based on the environment or configuration file. Really what we want to convey here is that:
1.  Polaris does not store or manage the authentication parameters.
2.  It does not apply any authentication flow explicitly. 

To elaborate on 2: If the config file provides no authentication parameters, then it will use an unauthenticated mechanism, on the other hand, let's say the config file contains auth method Kerberos and provides a path to a key-tab file, then the connection library will connect to the kerberized hadoop instance. In both cases Polaris is oblivious to what happens under the hood.

dimas-b:
`environment` is specific to how the service provider runs Polaris. Any credentials and auth flows based on that are what I termed "pre-arranged" because the Polaris user does not control them.

From the user's perspective the term `ENVIRONMENT` is not well defiled because it is not clear whether it is the user's env. or some other env. and the user does not have visibility into how the service provider runs Polaris.

poojanilangekar:
I agree that the user does not control the `ENVIRONMENT` on the polaris instance. Would you be okay with the term `DEFAULT_CONFIG`? I am also open to suggestions. We could also do `PROVIDER` but I don't think it's immediately apparent in this context. However, if you think that's the correct term, I can augment the API description and use it.

dimas-b:
@poojanilangekar : could you post the enum name options on the dev ML thread for visibility? This is an API change, so extra care is necessary.

poojanilangekar:
Hi everyone, just a gentle reminder: can you please respond to the mailing list with your naming preference?

poojanilangekar:
I've updated the PR by renaming the AuthenticationType to IMPLICIT.

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisAdminServiceAuthzTest.java
```diff
@@ -151,7 +152,8 @@ public void testCreateCatalogSufficientPrivileges() {
 
   @Test
   public void testCreateCatalogInsufficientPrivileges() {
-    final CatalogEntity newCatalog = new CatalogEntity.Builder().setName("new_catalog").build();
+    final CatalogEntity newCatalog =
+        new CatalogEntity.Builder().setName("new_catalog").setCatalogType("INTERNAL").build();
```
dimas-b: Why is this change required?

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -699,16 +701,10 @@ private Map<String, UserSecretReference> extractSecretReferences(
...
-        // TODO: Make this more targeted once we have connection configs that don't involve
-        // processing of inline secrets.
-        return true;
-      }
-    }
-    return false;
+  private boolean requiresSecretReferenceExtraction(
+      @NotNull ConnectionConfigInfo connectionConfigInfo) {
+    return connectionConfigInfo.getAuthenticationParameters().getAuthenticationType()
+        != AuthenticationParameters.AuthenticationTypeEnum.IMPLICIT;
```
dimas-b, line 707: This will return `true` for `NULL_TYPE`... Is that intended?

poojanilangekar: The type here refers to `AuthenticationTypeEnum` (generated based on the OpenAPI definition) and not `AuthenticationType` (declared in `...core.connection.AuthenticationType`). So `AuthenticationTypeEnum` can never be `NULL_TYPE`, it will be one of the types defined in polaris-management-service.yml (Line 910 - 914).

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisAdminServiceAuthzTest.java
poojanilangekar: Polaris assumes any catalog that is not defined as "INTERNAL" is by default "EXTERNAL", it ends up activating the code-path in PolarisAdminService.java:733 

Granted that since we allow legacy `EXTERNAL` catalogs without ConnectionConfigInfos, the test would still pass without the change. But I think that wasn't the intention of the test. 
 
For this test and other tests in `catalog/`, I think it's ideal to not activate that code path in the first place.  If you want me to I can revert this and other changes. Please let me know what you think.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: Ah, sorry, I misinterpreted :facepalm:

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisAdminServiceAuthzTest.java
dimas-b: I'm not so much worried about this test code change, but about defaulting to `EXTERNAL`. Existing test code appears to assume that the default is/was `INTERNAL` (which makes sense to me).

@dennishuo @eric-maynard : WDYT?

poojanilangekar: I think the fix is simple, I can create a separate PR to fix it. 

https://github.com/apache/polaris/blob/9cb71f8125d79e9bb16117c9110211394173b821/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java#L109

The builder checks if the `catalogType` is `INTERNAL`, if not, it assumes that the type is `EXTERNAL`, we can flip the condition to check if the type is set to `EXTEERNAL`.  I also think that the change only affects tests, because in other cases, it's generated from polaris-management-service.yml which defaults to `INTERNAL`.

dimas-b: Opening a PR for that would be great. Thx!

Let's review the impact of that change on that specific PR (I hope it's ok).

poojanilangekar: Done.

poojanilangekar: Since I opened #1998, I reverted this and other similar changes. The build still passes but I think we should try to get 1998 merged before this change.

dimas-b: Thx! Let's merge #1998 first and rebase.

poojanilangekar: Done.

spec/polaris-management-service.yml
```diff
@@ -917,6 +917,7 @@ components:
             - OAUTH
             - BEARER
             - SIGV4
+            - IMPLICIT
```
dimas-b, line 920: Please add a brief note to the `New Features` section in CHANGELOG.md about this.

dimas-b (commented):
LGTM overall with one more minor comment.

spec/polaris-management-service.yml
poojanilangekar: Done.

dimas-b (approved):

PR #1928, opened 2025-06-24
feat(ci): Improve Gradle cache in CI

adutra:

### DIFF
--- a/.github/actions/ci-incr-build-cache-prepare/action.yml
+++ b/.github/actions/ci-incr-build-cache-prepare/action.yml
@@ -1,0 +1,82 @@
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# CODE_COPIED_TO_POLARIS Copied from Project Nessie v. 0.104.2
+
+name: 'Incremental Gradle build cache prepare'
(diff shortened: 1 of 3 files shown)


.github/actions/ci-incr-build-cache-prepare/action.yml
```diff
@@ -0,0 +1,82 @@
...
+description: 'Prepare to save incremental Gradle build cache'
+inputs:
+  cache-read-only:
+    description: 'Gradle cache read only'
+    default: 'true'
+  java-version:
+    description: 'Java version'
+    default: '21'
+  job-id:
+    description: 'Job name to prefer'
```
snazy: Nit: the description should actually be
```suggestion
    description: 'Job ID to prefer'
```

.github/workflows/regtest.yml
```diff
@@ -53,9 +61,40 @@ jobs:
               :polaris-server:quarkusAppPartsBuild --rerun \
               -Dquarkus.container-image.build=true
 
+      - name: Save partial Gradle build cache
```
snazy: We should only store the Gradle caches from `gradle.yaml`, not from other workflows.

It's doesn't make much sense either, because the cache contents are only read, but never modified.

.github/workflows/spark_client_regtests.yml
```diff
@@ -56,10 +64,41 @@ jobs:
...
+
       # NOTE: the regression test runs with spark 3.5.5 and scala 2.12 in Java 17. We also have integration
       # tests runs with the existing gradle.yml, which only runs on Java 21. Since spark Java compatibility
       # for 3.5 is 8, 11, and 17, we should run spark client with those compatible java versions.
       # TODO: add separate spark client CI and run with Java 8, 11 and 17.
       - name: Regression Test
         run: |
           docker compose -f plugins/spark/v3.5/regtests/docker-compose.yml up --build --exit-code-from regtest
+
+  store-gradle-cache:
```
snazy: Same here, no need to store

snazy: Looks like we're good if this one is not stored

snazy (approved):
Thanks!
The changes LGTM!

PR #1931, opened 2025-06-24
Add SUPPORTED_FEDERATION_AUTHENTICATION_TYPES feature configuration

poojanilangekar:
Based on the discussion in #1925, this PR adds a SUPPORTED_FEDERATION_AUTHENTICATION_TYPES feature configuration.

I prefer the name SUPPORTED_FEDERATION_AUTHENTICATION_TYPES over SUPPORTED_CATALOG_AUTHENTICATION_TYPES because catalog in this context can be ambiguous. However, if others feel strongly, I can change it to SUPPORTED_CATALOG_AUTHENTICATION_TYPES to make it consistent with SUPPORTED_CATALOG_CONNECTION_TYPES. 

For now, we only support OAUTH and BEARER. Eventually, we can enable SIGV4 and NONE.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -20,6 +20,7 @@ package org.apache.polaris.core.config;
 
 import java.util.List;
 import java.util.Optional;
+import org.apache.polaris.core.admin.model.AuthenticationParameters;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
 import org.apache.polaris.core.connection.ConnectionType;
 import org.apache.polaris.core.context.CallContext;
@@ -259,6 +260,16 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
           .key("SUPPORTED_CATALOG_CONNECTION_TYPES")
           .description("The list of supported catalog connection types for federation")
           .defaultValue(List.of(ConnectionType.ICEBERG_REST.name()))
+          .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<List<String>> SUPPORTED_FEDERATION_AUTHENTICATION_TYPES =
+      PolarisConfiguration.<List<String>>builder()
+          .key("SUPPORTED_FEDERATION_AUTHENTICATION_TYPES")
+          .description("The list of supported authentication types for catalog federation")
+          .defaultValue(
+              List.of(
+                  AuthenticationParameters.AuthenticationTypeEnum.OAUTH.name(),
+                  AuthenticationParameters.AuthenticationTypeEnum.BEARER.name()))
           .buildFeatureConfiguration();
 
   public static final FeatureConfiguration<Integer> ICEBERG_COMMIT_MAX_RETRIES =
--- a/runtime/defaults/src/main/resources/application.properties
+++ b/runtime/defaults/src/main/resources/application.properties
@@ -112,6 +112,7 @@ polaris.features."ENFORCE_PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_CHECKING"=false
 polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE"]
 # polaris.features."ENABLE_CATALOG_FEDERATION"=true
 polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
+polaris.features."SUPPORTED_FEDERATION_AUTHENTICATION_TYPES"=["OAUTH", "BEARER"]
 
 # realm overrides
 # polaris.features.realm-overrides."my-realm"."SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION"=true
--- a/service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
+++ b/service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
@@ -29,11 +29,13 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.NotAuthorizedException;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.admin.model.AddGrantRequest;
+import org.apache.polaris.core.admin.model.AuthenticationParameters;
 import org.apache.polaris.core.admin.model.Catalog;
 import org.apache.polaris.core.admin.model.CatalogGrant;
 import org.apache.polaris.core.admin.model.CatalogRole;
 import org.apache.polaris.core.admin.model.CatalogRoles;
 import org.apache.polaris.core.admin.model.Catalogs;
+import org.apache.polaris.core.admin.model.ConnectionConfigInfo;
 import org.apache.polaris.core.admin.model.CreateCatalogRequest;
 import org.apache.polaris.core.admin.model.CreateCatalogRoleRequest;
 import org.apache.polaris.core.admin.model.CreatePrincipalRequest;
@@ -145,7 +147,7 @@ public class PolarisServiceImpl
     PolarisAdminService adminService = newAdminService(realmContext, securityContext);
     Catalog catalog = request.getCatalog();
     validateStorageConfig(catalog.getStorageConfigInfo());
-    validateConnectionConfigInfo(catalog);
+    validateExternalCatalog(catalog);
     Catalog newCatalog = new CatalogEntity(adminService.createCatalog(request)).asCatalog();
     LOGGER.info("Created new catalog {}", newCatalog);
     return Response.status(Response.Status.CREATED).build();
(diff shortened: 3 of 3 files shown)


poojanilangekar:
CC @eric-maynard @dimas-b

runtime/defaults/src/main/resources/application.properties
```diff
@@ -112,6 +112,7 @@ polaris.features."ENFORCE_PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_CHECKING"=false
 polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE"]
 # polaris.features."ENABLE_CATALOG_FEDERATION"=true
 polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
+polaris.features."SUPPORTED_FEDERATION_AUTHENTICATION_TYPES"=["OAUTH", "BEARER"]
```
dimas-b: why re-add the same value as the code-level default here?

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -169,30 +171,51 @@ private void validateStorageConfig(StorageConfigInfo storageConfigInfo) {
...
+    String authenticationType = authenticationParameters.getAuthenticationType().name();
+    List<String> supportedAuthenticationTypes =
+        callContext
+            .getPolarisCallContext()
+            .getConfigurationStore()
+            .getConfiguration(
+                callContext.getRealmContext(),
+                FeatureConfiguration.SUPPORTED_FEDERATION_AUTHENTICATION_TYPES);
+    if (!supportedAuthenticationTypes.contains(authenticationType)) {
+      throw new IllegalStateException("Unsupported authentication type: " + authenticationType);
```
dimas-b, line 218: Could you add a unit test for this, please?

eric-maynard:
Maybe `SUPPORTED_CATALOG_AUTHENTICATION_TYPES`? So far we haven't introduced the term "FEDERATION" anywhere, including in the analogous config `SUPPORTED_CATALOG_CONNECTION_TYPES`

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
poojanilangekar: Done. Since there were no existiing tests for the PolarisServiceImpl, I added some setup code to mock the service/context needed by the PolarisServiceImpl; this is not necessary for the test but used by the service. Let me know if you have suggestions to avoid it.

runtime/defaults/src/main/resources/application.properties
poojanilangekar: I actually followed the SUPPORTED_CATALOG_CONNECTION_TYPES here. Nonetheless, I've removed it.

eric-maynard: In turn, I followed `SUPPORTED_CATALOG_STORAGE_TYPES` for `SUPPORTED_CATALOG_CONNECTION_TYPES` :)

I think the idea is to be uber-explicit about having a "safe" default everywhere by excluding the unsafe options. However for `SUPPORTED_FEDERATION_AUTHENTICATION_TYPES` there is currently no unsafe option. There may be in the future though (NONE?) so keeping it wouldn't be the worst thing in light of that.

poojanilangekar:
@eric-maynard  I think the name `SUPPORTED_CATALOG_AUTHENTICATION_TYPES` is ambiguous. Polaris itself is a catalog, so if we have a configuration called `SUPPORTED_CATALOG_AUTHENTICATION_TYPES`, a user might think of it as the authentication types supported by this catalog (Polaris) instance. The feature configuration is related to federation and I don't see anything wrong in making the name explicitly related to the feature. 

I think the same goes for `SUPPORTED_CATALOG_CONNECTION_TYPES`. 

If you are opposed to adding `FEDERATION` in the name, how about we name (and rename) to `SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES` and `SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES` respectively?

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -169,30 +171,51 @@ private void validateStorageConfig(StorageConfigInfo storageConfigInfo) {
...
+  private void validateAuthenticationParameters(AuthenticationParameters authenticationParameters) {
+
+    String authenticationType = authenticationParameters.getAuthenticationType().name();
+    List<String> supportedAuthenticationTypes =
+        callContext
+            .getPolarisCallContext()
+            .getConfigurationStore()
+            .getConfiguration(
+                callContext.getRealmContext(),
+                FeatureConfiguration.SUPPORTED_FEDERATION_AUTHENTICATION_TYPES);
```
harishch1998: Do we need to handle uppercasing for this, similar to `supportedConnectionTypes` above? It might be redundant since those enums are uppercase anyway, so it's optional and upto you to make the change :p

eric-maynard:
@poojanilangekar 

> If you are opposed to adding FEDERATION in the name, how about we name (and rename) to SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES and SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES respectively?

Oh, that's a great idea. If you are amenable, let's do `..._EXTERNAL_...`

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
poojanilangekar: Done.

runtime/defaults/src/main/resources/application.properties
poojanilangekar: Makes sense, I've added it back in.

dimas-b: ok

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -254,12 +255,24 @@ public static void enforceFeatureEnabledOrThrow(
...
-  public static final FeatureConfiguration<List<String>> SUPPORTED_CATALOG_CONNECTION_TYPES =
-      PolarisConfiguration.<List<String>>builder()
-          .key("SUPPORTED_CATALOG_CONNECTION_TYPES")
-          .description("The list of supported catalog connection types for federation")
-          .defaultValue(List.of(ConnectionType.ICEBERG_REST.name()))
-          .buildFeatureConfiguration();
+  public static final FeatureConfiguration<List<String>>
+      SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES =
+          PolarisConfiguration.<List<String>>builder()
+              .key("SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES")
```
dimas-b: The change of config key is technically a breaking change. I'm personally ok with this, but it has sem. ver. implications per out current policy: https://polaris.apache.org/in-dev/unreleased/evolution/

Ideally we should support the old name for a while, but I do not think we have a mechanism for that ATM.

That said, since this PR is mostly about adding `SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES`, it might be best to do the renaming in a separate PR (plus add backward compatibility).

@eric-maynard WDYT?

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
dimas-b: @poojanilangekar : I'm not sure I understand your message completely :sweat_smile: Current test code LGTM, but if you think that some code can be removed without compromising tests' validity, I'd support that.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
poojanilangekar: Done, I've removed it. I'll send out another PR and we can discuss options there.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
poojanilangekar: Sorry, I just wanted to check if you're okay with the test code (or if you knew anyway to simplify it). Since you're okay with it, I don't plan to make any changes.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
eric-maynard: I'm okay with it too, but agreed that playing it safe is fine for now. We can always add support for multiple keys and then change this

eric-maynard (approved):

service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
dimas-b: sgtm :+1:

dimas-b (approved):

service/common/src/test/java/org/apache/polaris/service/admin/PolarisServiceImplTest.java
```diff
@@ -0,0 +1,239 @@
...
+  @Mock private ReservedProperties reservedProperties;
+  @Mock private PolarisCallContext polarisCallContext;
+  @Mock private PolarisConfigurationStore configurationStore;
+  @Mock private RealmContext realmContext;
+
+  private PolarisServiceImpl polarisService;
+
+  @BeforeEach
+  void setUp() {
+    entityManagerFactory = Mockito.mock(RealmEntityManagerFactory.class);
```
dimas-b, line 65: Do we need `@Mock` annotations above since we mock explicitly here?

poojanilangekar: Good catch, I have removed it.

PR #1686, opened 2025-05-27
Optimize the location overlap check with an index

eric-maynard:
The location overlap check for "sibling" tables (those which share a parent) has been a performance bottleneck since its introduction, but we haven't historically had a good way around this other than just disabling the check. 

<hr>

### Current Behavior

The current logic is that when we create a table, we list all sibling tables and check each and every one to ensure there is no location overlap. This results in O(N^2) checks when adding N tables to a namespace, quickly becoming untenable.

With the `CreateTreeDataset` [benchmark](https://github.com/eric-maynard/polaris-tools/blob/main/benchmarks/src/gatling/scala/org/apache/polaris/benchmarks/simulations/CreateTreeDataset.scala) I tested creating 5000 sibling tables using the current code:

Screenshot 2025-05-27 at 4 26 56 PM

It is apparent that latency increases over time. Runs took between 90 and 200+ seconds, and Polaris instances with a small memory allocation were prone to crashing due to OOMs:

Screenshot 2025-05-27 at 4 33 57 PM

### Proposed change

This PR adds a new persistence API, `hasOverlappingSiblings`, which if implemented can be used to directly check for the presence of siblings at the metastore layer.

This API is implemented for the JDBC metastore in a new schema version, and some changes are made to account for an evolving schema version now and in the future.

This implementation breaks a location down into components and queries for a sibling at each of those locations, so a new table at location `s3://bucket/root/n1/nA/t1/` will require checking for an entity with location `s3://bucket/`, `s3://bucket/root/`, `s3://bucket/root/n1/`, `s3://bucket/root/n1/nA/`, and finally `s3://bucket/root/n1/nA/t1/%`. All of this can be done in a single query which makes a single pass over the data. 

The query is optimized by the introduction of a new index over a new _location_ column.

With the changes enabled, I tested creating 5000 sibling tables:

Screenshot 2025-05-27 at 4 32 12 PM

Latency is stable over time, and runs consistently completed in less than 30 seconds. I did not observe any OOMs when testing with the feature enabled.

PR #1932, opened 2025-06-24
Fixed #214 - Added support for `s3a` scheme

### LINKED ISSUE #214: [FEATURE REQUEST] Add Support for S3A prefix

I have set the allowed location of the created catalog to S3 storage type using `s3://` prefix. When I run `remove_orphan_files` procedure in Spark, it results in an error message: `No FileSystem for scheme "s3"`.  To solve this problem, I attempted to create the catalog with the `s3a://` prefix, but I received a 400 Bad Request error with the message: `Location prefix not allowed`.
Here's my spark configuration
```
spark = SparkSession.builder \
            .config("spark.jars.packages","org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.5.2,org.apache.iceberg:iceberg-aws-bundle:1.5.2,org.apache.hadoop:hadoop-aws:3.4.0,org.apache.hadoop:hadoop-common:3.4.0") \
            .config("spark.sql.extensions", "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions") \
            .config('spark.sql.catalog.polaris.header.X-Iceberg-Access-Delegation', 'true') \
            .config("spark.sql.catalog.polaris.uri", POLARIS_URI) \
            .config("spark.sql.catalog.polaris.type", "rest") \
            .config("spark.sql.catalog.polaris", "org.apache.iceberg.spark.SparkCatalog") \
            .config("spark.sql.catalog.polaris.warehouse", POLARIS_CATALOG_NAME) \
            .config("spark.sql.catalog.polaris.io-impl", "org.apache.iceberg.aws.s3.S3FileIO") \
            .config("spark.hadoop.fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem") \
            .config('spark.sql.catalog.polaris.credential', POLARIS_CREDENTIALS) \
            .config('spark.sql.catalog.polaris.scope', POLARIS_SCOPE) \
            .config('spark.sql.catalog.polaris.token-refresh-enabled', 'true') \
            .getOrCreate()
```

Probably add the s3a:// prefix as an alternative for the S3 storage type.

_No response_

### Additional context

_No response_

pavibhai:
**Summary**
* Added support for s3a scheme when referring to paths on AWS S3
* scheme is ignored when comparing s3 and s3a paths for e.g. in overlap checks
* Tests have been added
* Some tests have been modified to run parameterized both for `s3` and `s3a`

**Motivation**
In the absence of this change, for tables involving `S3AFileSystem` the default scheme of `s3a` was used in the Iceberg metadata files. When you try to register this table it fails as `s3a` is not a valid scheme in Apache Polaris

As a workaround we had to at a minimum update the snapshot file to change the table location from `s3a://` to `s3://` for the table to be registered with Polaris

**Desired Behavior**
We should be able to treat both `s3://` and `s3a://` as valid AWS S3 paths

### DIFF
--- a/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
+++ b/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
@@ -21,31 +21,47 @@ package org.apache.polaris.spark.quarkus.it;
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 import org.apache.spark.sql.SparkSession;
 
-@QuarkusIntegrationTest
-public class SparkCatalogIcebergIT extends SparkCatalogBaseIT {
-  /** Initialize the spark catalog to use the iceberg spark catalog. */
-  @Override
-  protected SparkSession.Builder withCatalog(SparkSession.Builder builder, String catalogName) {
-    return builder
-        .config(
-            "spark.sql.extensions",
-            "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions")
-        .config(
-            String.format("spark.sql.catalog.%s", catalogName),
-            "org.apache.iceberg.spark.SparkCatalog")
-        .config("spark.sql.warehouse.dir", warehouseDir.toString())
-        .config(String.format("spark.sql.catalog.%s.type", catalogName), "rest")
-        .config(
-            String.format("spark.sql.catalog.%s.uri", catalogName),
-            endpoints.catalogApiEndpoint().toString())
-        .config(String.format("spark.sql.catalog.%s.warehouse", catalogName), catalogName)
-        .config(String.format("spark.sql.catalog.%s.scope", catalogName), "PRINCIPAL_ROLE:ALL")
-        .config(
-            String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
-        .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
-        .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
-        .config(
-            String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
-        .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
+public class SparkCatalogIcebergIT {
+
+  abstract static class BaseTest extends SparkCatalogBaseIT {
+    /** Initialize the spark catalog to use the iceberg spark catalog. */
+    @Override
+    protected SparkSession.Builder withCatalog(SparkSession.Builder builder, String catalogName) {
+      return builder
+          .config(
+              "spark.sql.extensions",
+              "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions")
+          .config(
+              String.format("spark.sql.catalog.%s", catalogName),
+              "org.apache.iceberg.spark.SparkCatalog")
+          .config("spark.sql.warehouse.dir", warehouseDir.toString())
+          .config(String.format("spark.sql.catalog.%s.type", catalogName), "rest")
+          .config(
+              String.format("spark.sql.catalog.%s.uri", catalogName),
+              endpoints.catalogApiEndpoint().toString())
+          .config(String.format("spark.sql.catalog.%s.warehouse", catalogName), catalogName)
+          .config(String.format("spark.sql.catalog.%s.scope", catalogName), "PRINCIPAL_ROLE:ALL")
+          .config(
+              String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
+          .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
+          .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
+          .config(
+              String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
+          .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
+    }
+  }
+
+  @QuarkusIntegrationTest
+  static class S3ATest extends BaseTest {
+    public S3ATest() {
+      s3Scheme = "s3a";
+    }
+  }
+
+  @QuarkusIntegrationTest
+  static class S3Test extends BaseTest {
+    public S3Test() {
+      s3Scheme = "s3";
+    }
   }
 }
--- a/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
+++ b/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
@@ -21,47 +21,31 @@ package org.apache.polaris.spark.quarkus.it;
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 import org.apache.spark.sql.SparkSession;
 
-public class SparkCatalogIcebergIT {
-
-  abstract static class BaseTest extends SparkCatalogBaseIT {
-    /** Initialize the spark catalog to use the iceberg spark catalog. */
-    @Override
-    protected SparkSession.Builder withCatalog(SparkSession.Builder builder, String catalogName) {
-      return builder
-          .config(
-              "spark.sql.extensions",
-              "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions")
-          .config(
-              String.format("spark.sql.catalog.%s", catalogName),
-              "org.apache.iceberg.spark.SparkCatalog")
-          .config("spark.sql.warehouse.dir", warehouseDir.toString())
-          .config(String.format("spark.sql.catalog.%s.type", catalogName), "rest")
-          .config(
-              String.format("spark.sql.catalog.%s.uri", catalogName),
-              endpoints.catalogApiEndpoint().toString())
-          .config(String.format("spark.sql.catalog.%s.warehouse", catalogName), catalogName)
-          .config(String.format("spark.sql.catalog.%s.scope", catalogName), "PRINCIPAL_ROLE:ALL")
-          .config(
-              String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
-          .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
-          .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
-          .config(
-              String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
-          .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
-    }
-  }
-
-  @QuarkusIntegrationTest
-  static class S3ATest extends BaseTest {
-    public S3ATest() {
-      s3Scheme = "s3a";
-    }
-  }
-
-  @QuarkusIntegrationTest
-  static class S3Test extends BaseTest {
-    public S3Test() {
-      s3Scheme = "s3";
-    }
+@QuarkusIntegrationTest
+public class SparkCatalogIcebergIT extends SparkCatalogBaseIT {
+  /** Initialize the spark catalog to use the iceberg spark catalog. */
+  @Override
+  protected SparkSession.Builder withCatalog(SparkSession.Builder builder, String catalogName) {
+    return builder
+        .config(
+            "spark.sql.extensions",
+            "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions")
+        .config(
+            String.format("spark.sql.catalog.%s", catalogName),
+            "org.apache.iceberg.spark.SparkCatalog")
+        .config("spark.sql.warehouse.dir", warehouseDir.toString())
+        .config(String.format("spark.sql.catalog.%s.type", catalogName), "rest")
+        .config(
+            String.format("spark.sql.catalog.%s.uri", catalogName),
+            endpoints.catalogApiEndpoint().toString())
+        .config(String.format("spark.sql.catalog.%s.warehouse", catalogName), catalogName)
+        .config(String.format("spark.sql.catalog.%s.scope", catalogName), "PRINCIPAL_ROLE:ALL")
+        .config(
+            String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
+        .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
+        .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
+        .config(
+            String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
+        .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
   }
 }
--- a/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogPolarisIT.java
+++ b/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogPolarisIT.java
@@ -20,5 +20,18 @@ package org.apache.polaris.spark.quarkus.it;
 
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 
-@QuarkusIntegrationTest
-public class SparkCatalogPolarisIT extends SparkCatalogBaseIT {}
+public class SparkCatalogPolarisIT {
+  @QuarkusIntegrationTest
+  static class S3ATest extends SparkCatalogBaseIT {
+    public S3ATest() {
+      s3Scheme = "s3a";
+    }
+  }
+
+  @QuarkusIntegrationTest
+  static class S3Test extends SparkCatalogBaseIT {
+    public S3Test() {
+      s3Scheme = "s3";
+    }
+  }
+}
--- a/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogPolarisIT.java
+++ b/plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogPolarisIT.java
@@ -20,18 +20,5 @@ package org.apache.polaris.spark.quarkus.it;
 
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 
-public class SparkCatalogPolarisIT {
-  @QuarkusIntegrationTest
-  static class S3ATest extends SparkCatalogBaseIT {
-    public S3ATest() {
-      s3Scheme = "s3a";
-    }
-  }
-
-  @QuarkusIntegrationTest
-  static class S3Test extends SparkCatalogBaseIT {
-    public S3Test() {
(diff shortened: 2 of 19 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
```diff
@@ -0,0 +1,60 @@
...
+
+import jakarta.annotation.Nonnull;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.polaris.core.storage.StorageLocation;
+
+public class S3Location extends StorageLocation {
+  private static final Pattern URI_PATTERN = Pattern.compile("^(s3a?)://(.+)$");
+  private final String scheme;
+  private final String object_key_w_bucket;
```
eric-maynard: I think a pass for style would be useful

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageLocation.java
```diff
@@ -32,6 +33,8 @@ public static StorageLocation of(String location) {
     // TODO implement StorageLocation for all supported file systems and add isValidLocation
     if (AzureLocation.isAzureLocation(location)) {
       return new AzureLocation(location);
+    } else if (S3Location.isS3Location(location)) {
```
eric-maynard, line 44: This looks like exactly the right approach to me

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
pavibhai: Yeah, missed that one, will change the variable name.

pavibhai: Fixed

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkIT.java
```diff
@@ -30,175 +30,190 @@
...
-    sql("DROP TABLE %s", icebergTable);
-    sql("DROP TABLE %s", deltaTable);
-    sql("DROP VIEW %s", viewName);
-    sql("DROP NAMESPACE %s", namespace);
-
-    // clean up delta directory
-    File dirToDelete = new File(deltaDir);
-    FileUtils.deleteQuietly(dirToDelete);
+  @QuarkusIntegrationTest
+  static class S3Test extends BaseTest {
```
eric-maynard: Are you planning to also create an S3ATest here? I don't think that's necessary, but we should at least have a couple ITs using s3a

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkIT.java
```diff
@@ -30,175 +30,190 @@
...
-    tables = sql("SHOW TABLES");
-    assertThat(tables.size()).isEqualTo(1);
-    assertThat(tables).contains(new Object[] {namespace, newIcebergTable, false});
-
-    // clean up the resources
-    sql("DROP VIEW %s", renamedView);
-    sql("DROP TABLE %s", newIcebergTable);
-    sql("DROP NAMESPACE %s", namespace);
+  @QuarkusIntegrationTest
+  static class S3ATest extends BaseTest {
```
eric-maynard: Ah, yeah, found it. Let's not duplicate all these tests, that's quite expensive. Can we just do one smoke test reading/writing a table?

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogPolarisIT.java
```diff
@@ -20,5 +20,18 @@
 
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 
-@QuarkusIntegrationTest
-public class SparkCatalogPolarisIT extends SparkCatalogBaseIT {}
+public class SparkCatalogPolarisIT {
+  @QuarkusIntegrationTest
+  static class S3ATest extends SparkCatalogBaseIT {
```
eric-maynard: Same comment -- these tests are slow, let's not duplicate all of them

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
```diff
@@ -21,31 +21,47 @@
...
+          .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
+          .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
+          .config(
+              String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
+          .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
+    }
+  }
+
+  @QuarkusIntegrationTest
+  static class S3ATest extends BaseTest {
```
eric-maynard: ditto

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
eric-maynard, line 1: Can we write some UTs for this class, similar to AzureLocationTest?

polaris-core/src/test/java/org/apache/polaris/service/storage/StorageLocationTest.java
```diff
@@ -77,4 +81,12 @@ public void testSpecialCharacters() {
             })
         .hasCauseInstanceOf(URISyntaxException.class);
   }
+
+  @ParameterizedTest
+  @CsvSource({"s3,s3a", "s3a,s3"})
+  public void testPrefixValidationIgnoresScheme(String parentScheme, String childScheme) {
```
eric-maynard: This would be a good fit for S3LocationTest

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisOverlappingCatalogTest.java
```diff
@@ -112,6 +128,20 @@ public void testBasicOverlappingCatalogs(boolean initiallyExternal, boolean late
         .hasMessageContaining("One or more of its locations overlaps with an existing catalog");
   }
 
+  @ParameterizedTest
+  @CsvSource({"s3,s3a", "s3a,s3"})
+  public void testBasicOverlappingCatalogWSchemeChange(String rootScheme, String overlapScheme) {
```
eric-maynard, line 133: Great idea to test this

eric-maynard (commented):
Looks good overall, I left some comments on tests. You may need to make some adjustments to make this compatible with #1686 but I otherwise feel this is pretty ready to merge

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+    Matcher matcher = URI_PATTERN.matcher(location);
+    return matcher.matches();
+  }
+
+  @Override
+  public boolean isChildOf(StorageLocation potentialParent) {
+    // Given that S3 and S3A are to be treated similarly, the parent check ignores the prefix
+    if (potentialParent instanceof S3Location) {
+      return this.objectKeyWBucket.startsWith(((S3Location) potentialParent).objectKeyWBucket);
```
flyrain: The previous code path will go through ensureTrailingSlash(). Should we also add it here for both parent and child `objectKeyWBucket`s?

flyrain: I'm a bit surprised that there is no test failure here. I think we will need to add more tests for tailing slash.

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
```diff
@@ -79,7 +80,7 @@ public void testGetSubscopedCreds() {
                   .returns(roleARN, AssumeRoleRequest::roleArn);
               return ASSUME_ROLE_RESPONSE;
             });
-    String warehouseDir = "s3://bucket/path/to/warehouse";
+    String warehouseDir = scheme + "://bucket/path/to/warehouse";
```
flyrain, line 86: Can we test the mixed use cases? For example, we create a catalog with allowed locations like `s3://bucket1/abc/`, while some of tables paths are with `s3a`. Does credential vending work in that case? Also i'm not sure how much this simulation can catch. It'd be nice to have a test against the real s3 storage for the mixed use cases.

flyrain (commented):
Thanks a lot for working on it @pavibhai !  The approach looks good to me. And thanks for adding a lot of tests. Left some comments.

pavibhai:
Thanks @eric-maynard and @flyrain for your comments, I will try to address them today.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
pavibhai: > The previous code path will go through ensureTrailingSlash(). Should we also add it here for both parent and child `objectKeyWBucket`s?

I was debating that. But makes sense even if we choose to remove for S3 we can have that as a separate conversation.

I will add that back.

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
pavibhai: > Does credential vending work in that case? Also i'm not sure how much this simulation can catch. It'd be nice to have a test against the real s3 storage for the mixed use cases.

@flyrain Yes from my integration test it works, but your question is about a test in the project itself. When building the policy content, we don't use the scheme anywhere, it is just bucket and object key.
Do we have any means for performing integration tests with AWS services already in the project? Otherwise this is something we should discuss.

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkIT.java
pavibhai: Sure that makes sense. Let me add one test that does the basic and cross check

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
pavibhai: sure

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
flyrain: The regression tests under dir `regtests` are designed for that. It only run against local file system now. S3, Azure, and GCP tests are skipped. @jbonofre and me are working on an AWS account to enable that as a github CI. Here is the command to run them. You will see s3 ones are skipped
```
./gradlew run
env POLARIS_HOST=localhost ./regtests/run.sh
```

I'm OK to merge this PR if we can have a manual test against real s3. The regtests could be added later.

pavibhai: thanks @flyrain will check that out

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkIT.java
pavibhai: I have reverted this and added a new test `PolarisS3InterOperabilityTest.java`

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogPolarisIT.java
pavibhai: I have reverted this and added a new test `PolarisS3InterOperabilityTest.java`

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
pavibhai: I have reverted this and added a new test `PolarisS3InterOperabilityTest.java`

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
pavibhai: Added this

polaris-core/src/test/java/org/apache/polaris/service/storage/StorageLocationTest.java
pavibhai: Moved this to S3LocationTest

pavibhai:
> Looks good overall, I left some comments on tests. You may need to make some adjustments to make this compatible with #1686 but I otherwise feel this is pretty ready to merge

I have rebased onto the current version. I can submit another PR to handle the `withoutScheme` method better in S3Location as that information is already available with the regex parsing.

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
pavibhai: @flyrain I have adjusted the test in `AwsCredentialsStorageIntegrationTest:testGetSubscopedCreds` to validate that the policy that we generate does not have any references to S3A.

This should be sufficient proof on this until we get to the test in `regtests`

pavibhai: > Can we test the mixed use cases? For example, we create a catalog with allowed locations like `s3://bucket1/abc/`, while some of tables paths are with `s3a`. Does credential vending work in that case? Also i'm not sure how much this simulation can catch. It'd be nice to have a test against the real s3 storage for the mixed use cases.

These tests have been moved to `PolarisS3InterOperabilityTest`, please check

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
pavibhai: This has been reverted back.

pavibhai:
@eric-maynard and @flyrain thanks for your comments, here are the key highlights of the change
* Rebased onto main (force push)
* Added a test `PolarisS3InterOperabilityTest` to include all of the S3 and S3A interoperability tests and removed the changes onto the IT tests
* Added a test `S3LocationTest` that includes the unit tests on S3Location
* Fixed a test to show case that even with S3A the policy content generated has no reference to S3A so should work similar to S3 for vending credentials

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisS3InterOperabilityTest.java
eric-maynard, line 1: `Interoperability` is one word

polaris-core/src/test/java/org/apache/polaris/core/storage/aws/S3LocationTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.apache.polaris.core.storage.aws;
+
+import org.apache.polaris.core.storage.StorageLocation;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class S3LocationTest {
```
eric-maynard, line 28: We should also add tests for `withoutScheme` just to be safe

eric-maynard (commented):
The tests look way better now, thanks for the update!

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisS3InterOperabilityTest.java
pavibhai: fixed

polaris-core/src/test/java/org/apache/polaris/core/storage/aws/S3LocationTest.java
pavibhai: Added

eric-maynard (approved):

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
```diff
@@ -0,0 +1,69 @@
...
+
+import jakarta.annotation.Nonnull;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.polaris.core.storage.StorageLocation;
+
+public class S3Location extends StorageLocation {
+  private static final Pattern URI_PATTERN = Pattern.compile("^(s3a?)://(.+)$");
+  private final String scheme;
+  private final String objectKeyWBucket;
```
eric-maynard: Is this always the same as `withoutScheme`? If so, let's remove it and use that. If not, let's improve the name please

runtime/service/src/test/java/org/apache/polaris/service/quarkus/entity/CatalogEntityTest.java
```diff
@@ -171,15 +179,29 @@ public void testValidAllowedLocationPrefix() {
...
+  @Test
+  public void testValidAllowedLocationPrefix() {
+    String basedLocation = "abfs://container@storageaccount.blob.windows.net/path";
     AzureStorageConfigInfo azureStorageConfigModel =
         AzureStorageConfigInfo.builder()
             .setAllowedLocations(List.of(basedLocation))
             .setStorageType(StorageConfigInfo.StorageTypeEnum.AZURE)
             .setTenantId("tenantId")
             .build();
+    CatalogProperties prop = new CatalogProperties(basedLocation);
```
eric-maynard, line 193: nit: `properties`; `prop` sounds like a single property

eric-maynard (commented):
The tests look great now; just one comment on `S3Location` and this LGTM

runtime/service/src/test/java/org/apache/polaris/service/quarkus/entity/CatalogEntityTest.java
pavibhai: done

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/S3Location.java
pavibhai: done

~~Subsequently we should change the method `withoutScheme` to use this instead~~

Actually let me include that also in this itself. This should be further broken down into bucket and key, but we can follow that up with another PR where we avoid any further parsing of the path e.g. during policy generation for credential vending

eric-maynard (approved):

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
flyrain: LGTM, thanks for adding these tests.

flyrain (approved):

flyrain:
Thanks @pavibhai for working on it. Thanks @eric-maynard for the review!

PR #1933, opened 2025-06-24
Exclude unused dependency for polaris spark client dependency

gh-yzou:
Today, when use the polaris spark client with Spark using --package, it also tries to download the following unused package
```
	found org.apache.polaris#polaris-api-management-model;1.1.0-incubating-SNAPSHOT in local-m2-cache
	found com.fasterxml.jackson.core#jackson-annotations;2.19.0 in central
	found com.fasterxml.jackson.core#jackson-core;2.19.0 in central
	found com.fasterxml.jackson.core#jackson-databind;2.19.0 in central
	found com.github.ben-manes.caffeine#caffeine;3.2.1 in central

```
Exclude those dependency to avoid downloading unnecessary package when using the Polaris Spark Client with spark using --packages. 
After exclusion, only the following dependency are pulled 
```
found org.apache.polaris#polaris-spark-3.5_2.12;1.1.0-incubating-SNAPSHOT in local-m2-cache
found org.apache.polaris#polaris-core;1.1.0-incubating-SNAPSHOT in local-m2-cache
found org.slf4j#slf4j-api;2.0.17 in central
found org.apache.iceberg#iceberg-spark-runtime-3.5_2.12;1.9.0 in central
```

### DIFF
--- a/plugins/spark/v3.5/spark/build.gradle.kts
+++ b/plugins/spark/v3.5/spark/build.gradle.kts
@@ -55,6 +55,15 @@ dependencies {
     exclude("io.smallrye.common", "*")
     exclude("io.swagger", "*")
     exclude("org.apache.commons", "*")
+    exclude("com.fasterxml.jackson", "*")
+    exclude("com.fasterxml.jackson.core", "*")
+    exclude("jakarta.inject", "*")
+    exclude("jakarta.validation", "*")
+    exclude("jakarta.ws.rs", "")
+    exclude("com.github.ben-manes.caffeine", "*")
+    exclude("commons-codec", "*")
+    exclude("com.google.guava", "*")
+    exclude("org.apache.polaris", "polaris-api-management-model")
   }
 
   implementation(


plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -55,6 +55,15 @@ dependencies {
...
     exclude("org.apache.commons", "*")
+    exclude("com.fasterxml.jackson", "*")
+    exclude("com.fasterxml.jackson.core", "*")
+    exclude("jakarta.inject", "*")
+    exclude("jakarta.validation", "*")
+    exclude("jakarta.ws.rs", "")
+    exclude("com.github.ben-manes.caffeine", "*")
+    exclude("commons-codec", "*")
+    exclude("com.google.guava", "*")
+    exclude("org.apache.polaris", "polaris-api-management-model")
```
flyrain: Another option here is to set `isTransitive =false`, so that all transitive dependency are excluded.

flyrain (approved):
+1. Thanks for the improvement.

gh-yzou:
@flyrain i updated the pr to use your suggestion simply exclude all transitive dependency

plugins/spark/v3.5/spark/build.gradle.kts
gh-yzou: Yes, that actually works, updated

flyrain (approved):

dimas-b (approved):

plugins/spark/v3.5/spark/build.gradle.kts
```diff
@@ -45,17 +45,7 @@ val scalaLibraryVersion =
...
-    exclude("com.azure", "*")
-    exclude("software.amazon.awssdk", "*")
-    exclude("com.google.cloud", "*")
-    exclude("io.airlift", "*")
-    exclude("io.smallrye", "*")
-    exclude("io.smallrye.common", "*")
-    exclude("io.swagger", "*")
-    exclude("org.apache.commons", "*")
-  }
+  implementation(project(":polaris-core")) { isTransitive = false }
```
dimas-b, line 48: As far as I understand, the only thing the Spark Client uses from core is `PolarisResourcePaths`  + `PolarisEndpoints`.

Depending on `polaris-core` without using any "core" functionality looks awkward to me. If the intention is to reuse those two classes, I believe it would be preferable to move them to an API module. WDYT (for followup)?

gh-yzou: Yes, there is a TODO on top to put the class that can be shared between client and server into a different module, and the module can potentially be used by other clients that would like to talk to polaris

PR #1937, opened 2025-06-25
Add relational-jdbc to helm

jparkzz:
## Motivation for the Change
  Polaris needs to support relational-jdbc as the default persistence type for simpler database configuration and better cloud-native deployment experience.

## Description of the Status Quo (Current Behavior)
  Currently, the Helm chart only supports eclipse-link persistence type as the default, which requires complex JPA configuration with persistence.xml files.

## Desired Behavior
  - Add relational-jdbc persistence type support to Helm chart
  - Use relational-jdbc as the default persistence type
  - Inject JDBC configuration (username, password, jdbc_url) through Kubernetes Secrets as environment variables
  - Maintain backward compatibility with eclipse-link

## Additional Details
  - Updated persistence-values.yaml for CI testing
  - Updated test coverage for relational-jdbc configuration
  - JDBC credentials are injected via QUARKUS_DATASOURCE_* environment variables from Secret
  - Secret keys: username, password, jdbc_url

### DIFF
--- a/helm/polaris/templates/configmap.yaml
+++ b/helm/polaris/templates/configmap.yaml
@@ -53,6 +53,11 @@ data:
     {{- $_ = set $map "polaris.persistence.eclipselink.persistence-unit" .Values.persistence.eclipseLink.persistenceUnit -}}
     {{- $_ = set $map "polaris.persistence.eclipselink.configuration-file" (printf "%s/persistence.xml" .Values.image.configDir ) -}}
     {{- end -}}
+    {{- if eq .Values.persistence.type "relational-jdbc" -}}
+    {{- if .Values.persistence.relationalJdbc.dbKind -}}
+    {{- $_ = set $map "quarkus.datasource.db-kind" .Values.persistence.relationalJdbc.dbKind -}}
+    {{- end -}}
+    {{- end -}}
 
     {{- /* File IO */ -}}
     {{- $_ = set $map "polaris.file-io.type" .Values.fileIo.type -}}
--- a/helm/polaris/templates/deployment.yaml
+++ b/helm/polaris/templates/deployment.yaml
@@ -72,11 +72,20 @@ spec:
           {{- end }}
           image: "{{ tpl .Values.image.repository . }}:{{ tpl .Values.image.tag . | default .Chart.Version }}"
           imagePullPolicy: {{ tpl .Values.image.pullPolicy . }}
-          {{ if or .Values.storage.secret.name .Values.extraEnv -}}
+          {{ if or .Values.storage.secret.name .Values.persistence.relationalJdbc.secret.name .Values.extraEnv -}}
           env:
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "awsAccessKeyId" "polaris.storage.aws.access-key") | indent 12 -}}
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "awsSecretAccessKey" "polaris.storage.aws.secret-key") | indent 12 -}}
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "gcpToken" "polaris.storage.gcp.token") | indent 12 -}}
+            {{- if and ( eq .Values.persistence.type "relational-jdbc" ) .Values.persistence.relationalJdbc.secret.name }}
+            {{- range $key := list "username" "password" "jdbc_url" }}
+            - name: QUARKUS_DATASOURCE_{{ $key | upper }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $.Values.persistence.relationalJdbc.secret.name }}
+                  key: {{ $key }}
+            {{- end }}
+            {{- end }}
             {{- if .Values.extraEnv -}}
             {{- tpl (toYaml .Values.extraEnv) . | nindent 12 -}}
             {{- end -}}
--- a/helm/polaris/templates/deployment.yaml
+++ b/helm/polaris/templates/deployment.yaml
@@ -78,8 +78,8 @@ spec:
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "awsSecretAccessKey" "polaris.storage.aws.secret-key") | indent 12 -}}
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "gcpToken" "polaris.storage.gcp.token") | indent 12 -}}
             {{- if and ( eq .Values.persistence.type "relational-jdbc" ) .Values.persistence.relationalJdbc.secret.name }}
-            {{- range $key, $envVar := dict "username" "username" "password" "password" "jdbcUrl" "jdbc.url" }}
-            - name: quarkus.datasource.{{ $envVar }}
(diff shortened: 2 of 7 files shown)


helm/polaris/templates/configmap.yaml
```diff
@@ -53,6 +53,11 @@ data:
     {{- $_ = set $map "polaris.persistence.eclipselink.persistence-unit" .Values.persistence.eclipseLink.persistenceUnit -}}
     {{- $_ = set $map "polaris.persistence.eclipselink.configuration-file" (printf "%s/persistence.xml" .Values.image.configDir ) -}}
     {{- end -}}
+    {{- if eq .Values.persistence.type "relational-jdbc" -}}
+    {{- if .Values.persistence.relationalJdbc.dbKind -}}
+    {{- $_ = set $map "quarkus.datasource.db-kind" .Values.persistence.relationalJdbc.dbKind -}}
```
dimas-b, line 58: `quarkus.datasource.db-kind` is a build-time setting. It is probably not useful to set it in helm.

https://quarkus.io/guides/datasource#quarkus-datasource_quarkus-datasource-db-kind

dimas-b: Cf. #1945

dimas-b (commented):
@jparkzz : Thanks for you contribution! It is very timely :) 

The LGTM, but I'll leave final reviews to people more experienced in Helm :)

helm/polaris/values.yaml
```diff
@@ -522,7 +522,16 @@ features:
 # -- Polaris persistence configuration.
 persistence:
   # -- The type of persistence to use. Two built-in types are supported: in-memory and eclipse-link.
```
```suggestion
  # -- The type of persistence to use. Two built-in types are supported: in-memory and relational-jdbc.
  # The eclipse-link type is also supported but is deprecated.
```

helm/polaris/templates/deployment.yaml
```diff
@@ -72,11 +72,20 @@ spec:
...
           imagePullPolicy: {{ tpl .Values.image.pullPolicy . }}
-          {{ if or .Values.storage.secret.name .Values.extraEnv -}}
+          {{ if or .Values.storage.secret.name .Values.persistence.relationalJdbc.secret.name .Values.extraEnv -}}
           env:
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "awsAccessKeyId" "polaris.storage.aws.access-key") | indent 12 -}}
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "awsSecretAccessKey" "polaris.storage.aws.secret-key") | indent 12 -}}
             {{- include "polaris.secretToEnv" (list .Values.storage.secret "gcpToken" "polaris.storage.gcp.token") | indent 12 -}}
+            {{- if and ( eq .Values.persistence.type "relational-jdbc" ) .Values.persistence.relationalJdbc.secret.name }}
+            {{- range $key := list "username" "password" "jdbc_url" }}
+            - name: QUARKUS_DATASOURCE_{{ $key | upper }}
```
adutra: Let's use `jdbcUrl` as the secret key, for consistency. Also, we should use lower-case Quarkus options if possible.

```suggestion
            {{- range $key, $envVar := dict "username" "username" "password" "password" "jdbcUrl" "jdbc.url" }}
            - name: quarkus.datasource.{{ $envVar }}
```

j1wonpark: I'm happy that this change meets your needs, thank you!

adutra (approved):

j1wonpark:
@adutra In Quarkus, when injecting configuration options as environment variables, they need to be in uppercase with underscores. For secret keys, camelCase can be used as well. The test was failing due to this issue, so I made the necessary corrections.

dimas-b:
> In Quarkus, when injecting configuration options as environment variables, they need to be in uppercase with underscores. [...]

This is not exactly correct. AFAIK, Quarkus understands env. variables with mixed case and special chars in the name as long as they match canonical configuration property names (e.g. `quarkus.datasource.jdbc.url`).

Uppercasing is usually necessary in shells (like `bash`) but k8s is more lenient and allows names like `quarkus.datasource.jdbc.url`, etc.

j1wonpark:
> > In Quarkus, when injecting configuration options as environment variables, they need to be in uppercase with underscores. [...]
> 
> This is not exactly correct. AFAIK, Quarkus understands env. variables with mixed case and special chars in the name as long as they match canonical configuration property names (e.g. `quarkus.datasource.jdbc.url`).
> 
> Uppercasing is usually necessary in shells (like `bash`) but k8s is more lenient and allows names like `quarkus.datasource.jdbc.url`, etc.

You were right. I verified that even if I use the original property name in lowercase with dot as an environment variable in quarkus, the property is loaded just fine. Thanks for the advice!

adutra (approved):

singhpk234 (approved):

singhpk234:
Thank you for the contribution @jparkzz ! Welcome to the Polaris community !

Thanks a ton for reviews @adutra @dimas-b !

helm/polaris/templates/configmap.yaml
dimas-b: @jparkzz : I still believe this configuration setting is not necessary and could be misleading because the value is not really changeable at the helm level. Would you mind following you with another PR to remove it?

dimas-b:
@singhpk234 : this comments was not really addressed : https://github.com/apache/polaris/pull/1937#discussion_r2168076991 it would have been nice to allow some more review time for before merging :)

helm/polaris/templates/configmap.yaml
j1wonpark: Would you prefer to configure dbKind within the secret? With current chart, even if we set it in the secret, I believe the value would still be explicitly visible in the helm chart when it's injected as an environment variable

dimas-b: No, my point is that `db-kind` can only be configured at build time (link above). It is not possible to overwrite that in Helm.

ATM only PostgreSQL is supported.

singhpk234:
@dimas-b noted ! Apologies, I missed your comment there, I saw @adutra's approval and I was also +1 so i merged. 
I agree with you here that db-kind should configured at build time. Please let me know how can i help, happy to put out a pr for this !

adutra:
> @dimas-b noted ! Apologies, I missed your comment there, I saw @adutra's approval and I was also +1 so i merged. 
> 
> I agree with you here that db-kind should configured at build time. Please let me know how can i help, happy to put out a pr for this !

Apologies, I approved too quickly 😅 Indeed db-kind cannot be modified at runtime so we should remove it from the Helm chart.

jbonofre:
@adutra @dimas-b about the db-kind that should be removed from Helm chart, do we have a PR about that ? I'm starting the 1.0.1 release preparation and I would like to be sure to cherry-pick all necessary commit. Thanks !

adutra:
@jbonofre it's done already: #1987

jbonofre:
@adutra got it, thanks !

PR #1838, opened 2025-06-09
Persistence implementation for pagination in some requests 

dimas-b:
Following up on #1555

* Refactor pagination code to delineate API-level page tokens and
  internal "pointers to data"

* Requests deal with the "previous" token, user-provided page size
  (optional) and the previous request's page size.

* Concentrate the logic of combining page size requests
  and previous tokens in PageTokenUtil
  
* PageToken subclasses are no longer necessary. EntityIdPaging handles
  pagination over ordered result sets with static helper methods.

PR #1555, opened 2025-05-09
Persistence implementations for list pagination

eric-maynard:
In #1528, we introduced the interface changes necessary to paginate requests to listTables, listViews, and listNamespaces. This PR adds the persistence-level logic for pagination and a new PageToken type `EntityIdPageToken` used to paginate requests based on entity ID.

PR #1189, opened 2025-03-17
NoSQL database agnostic persistence

snazy:
Mostly complete proposal for the database agnostic persistence layer to support NoSQL databases in Apache Polaris.

This PR is intended to illustrate the “whole spiel” and contains all necessities. Smaller, easier reviewable PRs will be opened soon for some of the areas covered in this PR to initially enable MongoDB persistence.

The overall approach borrows the architectural approach from Nessie to isolate database specifics from higher level concerns.

The most important objective for the implementation is correctness, especially in scenarios with high concurrent load. Explicit tests to verify the correctness are included, for the CI “use case” and for manual/special runs against a clustered database setup (which are just “too much” for the Github hosted runners).

The current integration point is `MetaStoreManagerFactory`/`PolarisMetaStoreManager` implemented in the “bridge” Gradle project.

See `components/persistence/README.md` in this PR for more technical information.

To run Polaris w/ the “dummy” in-memory “database”:
```bash
./gradlew quarkusBuild
java -jar -Djava.security.manager=allow \
  -Dpolaris.persistence.type=persistence \
  -jar quarkus/server/build/quarkus-app/quarkus-run.jar
```

To run Polaris against MongoDB:
```bash
docker run --publish 27017:27017 --rm -ti mongo:8.0.5
./gradlew quarkusBuild
java -jar -Djava.security.manager=allow \
  -Dpolaris.persistence.type=persistence \
  -Dpolaris.backend.name=MongoDb \
  -Dquarkus.mongodb.connection-string=mongodb://localhost:27017 \
  -jar quarkus/server/build/quarkus-app/quarkus-run.jar
```

PR #1938, opened 2025-06-25
Extensible pagination token implementation

snazy:
Based on #1838, following up on #1555

* Allows multiple implementations of `Token` referencing the "next page", encapsulated in `PageToken`. No changes to `polaris-core` needed to add custom `Token` implementations.
* Extensible to (later) support (cryptographic) signatures to prevent tampered page-token
* Refactor pagination code to delineate API-level page tokens and internal "pointers to data"
* Requests deal with the "previous" token, user-provided page size (optional) and the previous request's page size.
* Concentrate the logic of combining page size requests and previous tokens in `PageTokenUtil`
* `PageToken` subclasses are no longer necessary.
* Serialzation of `PageToken` uses Jackson serialization (smile format)

Since no (metastore level) implementation handling pagination existed before, no backwards compatibility is needed.


snazy:
This approach also works for NoSQL #1189 (last commit in [this branch](https://github.com/snazy/polaris/tree/persistence-nosql-paging))

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/EntitiesResult.java
```diff
@@ -48,41 +43,19 @@ public static EntitiesResult fromPage(Page<PolarisBaseEntity> page) {
...
-  public EntitiesResult(
-      @Nonnull List<PolarisBaseEntity> entities, @Nonnull Optional<PageToken> pageTokenOpt) {
+  public EntitiesResult(@Nonnull Page<PolarisBaseEntity> entities) {
     super(ReturnStatus.SUCCESS);
     this.entities = entities;
-    this.pageTokenOpt = pageTokenOpt;
-  }
-
-  @JsonCreator
-  private EntitiesResult(
```
eric-maynard, line 71: If we can avoid it, I think it's better not to change this. This will be a breaking change for any persistence implementation depending on wire-compatibility of the result.

eric-maynard: cc @dennishuo for visibility as well

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/ListEntitiesResult.java
```diff
@@ -46,43 +41,26 @@ public static ListEntitiesResult fromPage(Page<EntityNameLookupRecord> page) {
...
-  public ListEntitiesResult(
-      @Nonnull List<EntityNameLookupRecord> entities, @Nonnull Optional<PageToken> pageTokenOpt) {
+  public ListEntitiesResult(Page<EntityNameLookupRecord> entities) {
     super(ReturnStatus.SUCCESS);
     this.entities = entities;
-    this.pageTokenOpt = pageTokenOpt;
   }
 
-  @JsonCreator
-  private ListEntitiesResult(
```
eric-maynard, line 71: ditto the comment on wire-compatibility with previous versions

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.polaris.immutables.PolarisImmutable;
+
+/** Pagination {@linkplain Token token} backed by {@link PolarisBaseEntity#getId() entity ID}. */
+@PolarisImmutable
+@JsonSerialize(as = ImmutableEntityIdToken.class)
+@JsonDeserialize(as = ImmutableEntityIdToken.class)
+public interface EntityIdToken extends Token {
+  String ID = "e";
+
+  @JsonProperty("i")
```
eric-maynard, line 36: Why would we use the property name `i`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
```diff
@@ -0,0 +1,66 @@
...
+  long entityId();
+
+  @Override
+  default String getT() {
```
eric-maynard, line 40: `getT`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.annotation.Nullable;
+import org.apache.polaris.core.entity.PolarisBaseEntity;
+import org.apache.polaris.immutables.PolarisImmutable;
+
+/** Pagination {@linkplain Token token} backed by {@link PolarisBaseEntity#getId() entity ID}. */
+@PolarisImmutable
+@JsonSerialize(as = ImmutableEntityIdToken.class)
+@JsonDeserialize(as = ImmutableEntityIdToken.class)
+public interface EntityIdToken extends Token {
+  String ID = "e";
```
eric-maynard, line 34: Is this meant to be short for `EntityId`? Why not use an enum?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
```diff
@@ -18,25 +18,99 @@
...
+      Iterator<T> it = iterator(items.spliterator());
+      int limit = request.pageSize().orElse(Integer.MAX_VALUE);
+      while (it.hasNext() && data.size() < limit) {
+        last = it.next();
+        data.add(mapper.apply(last));
+      }
+
+      // Signal "no more data" if the number of items is less than the requested page size or if
+      // there is no more data.
+      if (data.size() < limit || !it.hasNext()) {
```
eric-maynard, line 88: It should be expected that the iterator has no more items if the limit has been pushed down into whatever provided the `items`, no?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
```diff
@@ -18,25 +18,99 @@
...
+    return items;
+  }
+
+  /**
+   * Returns a page token in encoded form suitable for returning to API clients. The string returned
+   * from this method is expected to be parsed by {@link PageToken#build(String, Integer)} when
+   * servicing the request for the next page of related data.
+   */
+  public @Nullable String encodedResponseToken() {
+    return PageTokenUtil.encodePageToken(request, nextToken);
```
eric-maynard, line 106: Same comments from the previous PR -- encoding/decoding a given PageToken clearly seems to be within the purview of a given PageToken implementation

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
-public abstract class PageToken {
-
-  /** Build a new PageToken that reads everything */
-  public static PageToken readEverything() {
-    return build(null, null);
-  }
+/** A wrapper for pagination information passed in as part of a request. */
+@PolarisImmutable
+@JsonSerialize(as = ImmutablePageToken.class)
+@JsonDeserialize(as = ImmutablePageToken.class)
```
eric-maynard, line 32: Would rather avoid these if possible

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
-  public static PageToken fromString(String token) {
-    return build(token, null);
-  }
-
-  /** Build a new PageToken from a limit */
-  public static PageToken fromLimit(Integer pageSize) {
-    return build(null, pageSize);
-  }
+  /** The requested page size (optional). */
+  @JsonProperty("p")
```
eric-maynard, line 38: ditto the comment on `i` -- this is not a good property name.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
-
   /**
-   * Builds a new page token to reflect new data that's been read. If the amount of data read is
-   * less than the pageSize, this will return a {@link DonePageToken}
+   * Paging token value, if present. Serialized paging tokens always have a value, but "synthetic"
+   * paging tokens like {@link #readEverything()} or {@link #fromLimit(int)} do not have a token
+   * value.
    */
-  protected abstract PageToken updated(List<?> newData);
+  @JsonProperty("v")
```
eric-maynard, line 51: ditto

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
+  Optional<Token> value();
```
eric-maynard, line 52: What is `value`? This seems a bit abstract.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
-      } else {
-        return new ReadEverythingPageToken();
-      }
-    } else {
-      // TODO implement, split out by the token's prefix
-      throw new IllegalArgumentException("Unrecognized page token: " + token);
-    }
+  /** Convenience for {@code pageSize().isPresent()}. */
+  default boolean paginationRequested() {
+    return pageSize().isPresent();
```
eric-maynard, line 43: It's possible per the spec to request pagination by providing a page token without page size. This logic is not correct.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
+  default <T extends Token> Optional<T> valueAs(Class<T> type) {
+    return value()
+        .flatMap(
+            t ->
+                type.isAssignableFrom(t.getClass()) ? Optional.of(type.cast(t)) : Optional.empty());
+  }
+
+  /** Represents a non-paginated request. */
+  static PageToken readEverything() {
+    return PageTokenUtil.READ_EVERYTHING;
```
eric-maynard, line 70: Is `READ_EVERYTHING` a `Util`? Seems like it would belong in `PageToken`

eric-maynard:
Thanks for taking a crack at this. I appreciate the extra extensibility compared to #1838, but considering this diff is twice the size of #1555, I'm not sure there's enough functionality here to justify all the changes. I took a quick pass for now, but will try to circle back and review the rest of the PR in more detail later.

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapTransactionalPersistenceImpl.java
```diff
@@ -361,13 +362,23 @@ public List<EntityNameLookupRecord> lookupEntityActiveBatchInCurrentTxn(
...
+            .map(
+                entityIdToken -> {
+                  var nextId = entityIdToken.entityId();
+                  return (Predicate<PolarisBaseEntity>) e -> e.getId() > nextId;
+                })
+            .orElse(e -> true);
+
+    data = data.sorted(Comparator.comparingLong(PolarisEntityCore::getId)).filter(tokenFilter);
+
+    data = data.filter(entityFilter);
```
eric-maynard, line 385: Why break this into two calls?

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/EntitiesResult.java
dimas-b: It is a `private` constructor... @eric-maynard @dennishuo : could you give some more details about possible compatibility issues? What's the affected workflow?

polaris-core/src/test/resources/META-INF/services/org.apache.polaris.core.persistence.pagination.Token$TokenType
dimas-b, line 1: `Token$TokenType` looks awkward as a file name (`$` and double `Token`)... WDYT about making `TokenType` a top-level class?

dimas-b (commented):
Changes LGTM :+1:

@eric-maynard : Do you have any other comments?

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/EntitiesResult.java
dimas-b: I do not see any roundtrips through JSON for `EntitiesResult` in Apache Polaris code. If this is a use case downstream, it would be nice to have tests in this repo to assert the expected behaviour.

That said, I think this specific change is not critical to pagination. `EntitiesResult` is not involved handling paginated data. So, I suppose this change can be rolled back.

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/ListEntitiesResult.java
dimas-b: @eric-maynard : It's good that you're able to identify this change as a regression, but what if you're not available ? :) WDYT about encoding this expectation in a unit test?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
dimas-b: I believe short names are meant to reduce the side of the token on the wire.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
dimas-b: Good point! It also looks like we do not push limits down to the database.

I propose to remove `!it.hasNext()` for the sake of simplicity, accept possible empty last pages, and work on exact page boundaries later.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
dimas-b: The logic looks correct to me. Propagating page size from the previous token in handled in `PageTokenUtil.decodePageRequest()`

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapTransactionalPersistenceImpl.java
dimas-b: (I think it came from my PR) No particular reason. Would you prefer a chained call?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -18,82 +18,75 @@
...
- * which is a serialized PageToken.
- *
- * <p>By providing that in the next query's `pageToken`, the client can resume listing where they
- * left off. If the client provides a `pageToken` or `pageSize` but `next-page-token` is null in the
- * response, that means there is no more data to read.
- */
-public abstract class PageToken {
-
-  /** Build a new PageToken that reads everything */
-  public static PageToken readEverything() {
```
dimas-b, line 36: Could we keep this utility method here?

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
snazy: The reason is explained as doc in `Token`

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
snazy: Yes, as explained in `Token`

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
snazy: That contradicts the intent of being extensible.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
snazy: There is only one `PageToken`.

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/EntitiesResult.java
snazy: I do not see a "wire" either.
BTW: The same change wasn't considered an issue [in 1838](https://github.com/apache/polaris/pull/1838/files#diff-3fc6d3601347843472388857f7b6237399973e08711fc3d04b6b7589b1cedc4eL70).

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
snazy: There is a test for this - and it passes

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
snazy: It is still there?

polaris-core/src/test/resources/META-INF/services/org.apache.polaris.core.persistence.pagination.Token$TokenType
snazy: It's the class name.
I'd really prefer keeping these very related things (`Token` + `TokenType`) together - you have to implement both.

snazy:
This change enables pagination also for #1189, whereas #1555 does not.

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
dimas-b: oops, missed it due to odd diff layout :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
dimas-b: I wonder whether the test passes because limits are not pushed down to Persistence queries anywhere :thinking:

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
dimas-b: My take: `value` is the implementation-specific token data (e.g. the entity ID in sorted SQL result sets).

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
dimas-b: `PageToken` provides a convenience getting for "read everything". `PageTokenUtil` groups several implementations.

snazy:
Thanks all!

eric-maynard:
This seems to have caused a regression:
```
2025-07-16 10:01:26,449 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] [,POLARIS] [,,,] (executor-thread-1) HTTP Request to /api/management/v1/catalogs failed, error id: afec5ba7-078c-4917-8d4d-2feaf219fe13-1: java.util.ServiceConfigurationError: com.fasterxml.jackson.databind.Module: Provider com.fasterxml.jackson.module.scala.DefaultScalaModule could not be instantiated
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:586)
	at java.base/java.util.ServiceLoader$ProviderImpl.newInstance(ServiceLoader.java:813)
	at java.base/java.util.ServiceLoader$ProviderImpl.get(ServiceLoader.java:729)
	at java.base/java.util.ServiceLoader$3.next(ServiceLoader.java:1403)
	at com.fasterxml.jackson.databind.ObjectMapper.findModules(ObjectMapper.java:1161)
	at com.fasterxml.jackson.databind.ObjectMapper.findModules(ObjectMapper.java:1145)
	at com.fasterxml.jackson.databind.ObjectMapper.findAndRegisterModules(ObjectMapper.java:1195)
	at org.apache.polaris.core.persistence.pagination.PageTokenUtil.<clinit>(PageTokenUtil.java:46)
	at org.apache.polaris.core.persistence.pagination.PageToken.readEverything(PageToken.java:70)
	at org.apache.polaris.service.admin.PolarisAdminService.listCatalogsUnsafe(PolarisAdminService.java:964)

```

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageTokenUtil.java
```diff
@@ -0,0 +1,267 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.ServiceLoader;
+
+final class PageTokenUtil {
+
+  private static final ObjectMapper SMILE_MAPPER = new SmileMapper().findAndRegisterModules();
```
eric-maynard, line 46: I think the issue I mentioned is likely happening when there are multiple Scala dependencies on the classpath?

If we are calling `findAndRegisterModules`, I wonder if we should be explicitly adding a dependency for Scala, e.g.:
```
implementation("org.scala-lang:scala-library:2.12.15")
```

snazy: I don't think so. That code is for Polaris server, which doesn't have anything from Scala.

eric-maynard: Well, it might, depending on the extensions you have. From what I can tell this actually functions without `findAndRegisterModules` -- why do we need it?

PR #1939, opened 2025-06-25
Fix the LICENSE and NOTICE with the latest dependency updates

jbonofre:


adutra (approved):

flyrain (approved):
Thanks @jbonofre ! Looking forward to the licenses without versions per offline discussion. So that we don't have to update them constantly.

PR #1941, opened 2025-06-25
Simplify install dependency doc

MonkeyCanCode:
Minor change to simply the install dependency doc:
1. change from two lines to one line with `git clone`
2. instead of asking users to check official jq doc (which will tell them the same thing about brew) and we are using brew for `docker` and `jenv` already in this doc. Thus, I think it makes more senses to provide brew command directly.

### DIFF
--- a/site/content/in-dev/unreleased/getting-started/install-dependencies.md
+++ b/site/content/in-dev/unreleased/getting-started/install-dependencies.md
@@ -114,7 +114,7 @@ Ensure that `java --version` and `javac` both return non-zero responses.
 
 ## jq
 
-Most Polaris Quickstart scripts require [jq]((https://jqlang.org/download/)). You can install jq using [homebrew](https://brew.sh/) on MacOS:
+Most Polaris Quickstart scripts require [jq]((https://jqlang.org/download/)). You can install jq using [homebrew](https://brew.sh/):
 ```shell
 brew install jq
 ```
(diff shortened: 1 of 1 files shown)


site/content/in-dev/unreleased/getting-started/install-dependencies.md
```diff
@@ -115,4 +114,7 @@ Ensure that `java --version` and `javac` both return non-zero responses.
 
 ## jq
 
-Most Polaris Quickstart scripts require `jq`. Follow the instructions from the [jq](https://jqlang.org/download/) website to download this tool.
\ No newline at end of file
+Most Polaris Quickstart scripts require [jq]((https://jqlang.org/download/)). You can install jq using [homebrew](https://brew.sh/) on MacOS:
```
eric-maynard: You can use brew on *nix, right?

MonkeyCanCode: Yes. I use it on my Mac as well WSL (ref: https://docs.brew.sh/Homebrew-on-Linux).

eric-maynard: So do we need this change? It makes it sound like it only works on MacOS... though, it's true that it most easily works on MacOS

eric-maynard: Maybe we should try something like:

```
Most Polaris Quickstart scripts require installing [jq]((https://jqlang.org/download/)). For example, with [homebrew](https://brew.sh/):
. . .
```

MonkeyCanCode: I see what you mean. Yeah, I will following the wording above. Let me update it to be consistent.

MonkeyCanCode: @eric-maynard please take a look. It should now be consistent with jenv and the one I refs to above is actually belong to docker on MacOS section which is not related to this.

eric-maynard (approved):

PR #1942, opened 2025-06-25
Add options to the bootstrap command to specify a schema file

eric-maynard:
Instead of always using the hardcoded `schema-v1.sql` file, it would be nice if users could specify a file to bootstrap from. This is especially relevant after #1686 which proposes to add a new "version" of the schema.


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -108,18 +115,19 @@ private void initializeForRealm(
...
+  private DatasourceOperations getDatasourceOperations(Optional<SchemaOptions> schemaOptions) {
     DatasourceOperations databaseOperations;
     try {
       databaseOperations = new DatasourceOperations(dataSource.get(), relationalJdbcConfiguration);
     } catch (SQLException sqlException) {
       throw new RuntimeException(sqlException);
     }
-    if (isBootstrap) {
+    // If this is set, we are bootstrapping
+    if (schemaOptions.isPresent()) {
```
dimas-b: Since we're touching this code, I'd like to avoid having bootstrap logic on the main runtime call path (in server). Do you think we could refactor that to execute DDL only in the admin tool?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -43,7 +45,21 @@ public static DatabaseType fromDisplayName(String displayName) {
...
-  public String getInitScriptResource() {
-    return String.format("%s/schema-v1.sql", this.getDisplayName());
+  public String getInitScriptResource(@Nonnull SchemaOptions schemaOptions) {
+    final String schemaFile;
+    if (schemaOptions.schemaFile() != null) {
+      schemaFile = schemaOptions.schemaFile();
+    } else {
+      String schemaSuffix;
+      switch (schemaOptions.schemaVersion()) {
+        case SchemaOptions.LATEST -> schemaSuffix = "schema-v1.sql";
```
dimas-b: suggestion: make `schemaVersion()` optional and default to latest.

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
```diff
@@ -71,6 +78,23 @@ static class FileInputOptions {
...
+          paramLabel = "<schema version>",
+          description = "The version of the schema to load in [1, 2, LATEST].",
+          defaultValue = SchemaOptions.LATEST)
+      String schemaVersion;
+
+      @CommandLine.Option(
+          names = {"--schema-file"},
+          paramLabel = "<schema file>",
+          description =
+              "A schema file to bootstrap from. If unset, the bundled files will be used.",
```
dimas-b, line 93: "file" may be misleading here. The script is loaded from resources built into the admin tool jar, IIRC.

dimas-b (commented):
The general idea LGTM. I've got some comments, though :)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
eric-maynard: In SchemaOptions it is:
```
  @Value.Default
  default String schemaVersion() {
    return LATEST;
  }
```

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
eric-maynard: I would actually like to modify that behavior; let me make some changes to support both resources on the classpath and arbitrary files to see what you think

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
dimas-b: I mean default in bootstrap code to whatever version happens to be latest when no explicit version is requested, but avoid having the `LATEST` constant.

dimas-b: alternatively: use actual current schema version in the `LATEST` constant.

dimas-b: wait... that does not carry over to NoSQL :thinking:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
eric-maynard: I tried my hand at this in 6fee80f14e4904bc4f43d1849db713d60cd00a2c, let me know if that looks like what you had in mind

eric-maynard: Looking, I only pulled the script execution up one level into `bootstrapRealms` rather than into the admin tool itself -- though it only runs as a result of that admin tool code path. If we want to pull the logic out of even `bootstrapRealms` I think we may need to do some refactoring to `MetaStoreManagerFactory` which I would prefer to avoid doing in this PR.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
dimas-b: how about `Optional<String> schemaVersion()`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/SchemaOptions.java
```diff
@@ -0,0 +1,37 @@
...
+public interface SchemaOptions {
+  public static final String LATEST = "LATEST";
+
+  @Value.Default
+  default String schemaVersion() {
+    return LATEST;
+  }
+
+  @Nullable
+  String schemaFile();
```
dimas-b: `schemaVersion` and `schemaFile`, if set, are probably mutually exclusive. WDYT about adding a check method for that?

eric-maynard: There's a check in picocli, but let me add one to this object as well

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
eric-maynard: I just changed to null by default, does this look better?

dimas-b: Null works too :)

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
dimas-b: sgtm

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
dimas-b: Do we really need the `LATEST` constant now?

eric-maynard: I was thinking it's nice to have so that users can be explicit about taking the LATEST schema -- the alternative is just that you get LATEST by _not_ setting the version, right?

Besides the ability to be explicit about wanting the LATEST, I think the constant is potentially useful in the contrived scenario where a persistence wants to default to a non-LATEST schema

dimas-b: Fair enough... but in this case I'd prefer to go back to the non-null `schemaVersion()` with `LATEST` as default in the Immutable class... Sorry about back-and-forth :facepalm:

eric-maynard: Well, in that case we actually lose this:
> the contrived scenario where a persistence wants to default to a non-LATEST schema

Since there is no way for the value to be unset, it's either LATEST or some number. But I don't feel too strongly about this, as I think that scenario is pretty unlikely. I'm happy to either leave the default or remove LATEST; I'll try getting rid of LATEST for now

dimas-b: right :facepalm:

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/AtomicMetastoreManagerWithJdbcBasePersistenceImplTest.java
```diff
@@ -49,8 +50,11 @@ protected PolarisTestMetaStoreManager createPolarisTestMetaStoreManager() {
     try {
       datasourceOperations =
           new DatasourceOperations(createH2DataSource(), new H2JdbcConfiguration());
-      datasourceOperations.executeScript(
-          String.format("%s/schema-v2.sql", DatabaseType.H2.getDisplayName()));
+      ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
+      InputStream scriptStream =
+          classLoader.getResourceAsStream(
```
dimas-b, line 55: nit: this stream normally needs to be closed, AFAIK.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -80,16 +81,13 @@ DatabaseType getDatabaseType() {
...
+  public void executeScript(InputStream scriptInputStream) throws SQLException {
     runWithinTransaction(
         connection -> {
           try (Statement statement = connection.createStatement()) {
             BufferedReader reader =
                 new BufferedReader(
-                    new InputStreamReader(
-                        Objects.requireNonNull(classLoader.getResourceAsStream(scriptFilePath)),
-                        UTF_8));
+                    new InputStreamReader(Objects.requireNonNull(scriptInputStream), UTF_8));
```
dimas-b: who is closing `scriptInputStream`?

eric-maynard: Who was closing it before? When the JVM exits, I suppose the OS will reclaim the resources eventually... and this is a short-lived bootstrap invocation, not the long-running service. But for now, maybe we can just add a `close` in this method and shift the responsibility to close here.

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/AtomicMetastoreManagerWithJdbcBasePersistenceImplTest.java
eric-maynard: ditto the above -- I think we just weren't closing it before, but added an explicit `close` into `executeScript` to fix this

adutra: Why not a try-with-resources block?

eric-maynard: Per the javadoc, I think it's better to have this method take on the responsibility for closing the stream after it's used rather than scatter try-with-resources amongst the callers. The loan pattern would be better, but we're already getting pretty far away from the intent of this change.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
dimas-b: I did not look into the previous revision of this code in details :sweat_smile:  but in general, I believe theses streams need to be closed for good measure (before JVM exit)... WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -75,46 +76,51 @@ DatabaseType getDatabaseType() {
   }
 
   /**
-   * Execute SQL script.
+   * Execute SQL script and close the associated input stream
```
dimas-b, line 79: WDYT about adding `@WillClose` to the stream parameter?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -75,46 +76,51 @@ DatabaseType getDatabaseType() {
...
             }
-            return true;
-          } catch (IOException e) {
-            throw new RuntimeException(e);
-          }
-        });
+          });
+    } finally {
+      try {
+        scriptInputStream.close();
```
dimas-b, line 119: I'm pretty sure the try on line 88 closes this stream too... Could you double check?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -43,7 +48,25 @@ public static DatabaseType fromDisplayName(String displayName) {
     };
   }
 
-  public String getInitScriptResource() {
-    return String.format("%s/schema-v2.sql", this.getDisplayName());
+  public InputStream getInitScriptResource(@Nonnull SchemaOptions schemaOptions) {
```
dimas-b: Since the returned stream allocates I/O resources and needs to be closed, WDYT about renaming the method to `openInitScriptResouse()`?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
eric-maynard: That sounds like a good idea... is this [the one from jsr305](https://www.javadoc.io/static/com.google.code.findbugs/jsr305/3.0.2/javax/annotation/WillClose.html)? It doesn't seem that we have that dependency currently, and we're not using it anywhere else in the project.

eric-maynard: Oops sorry mixed up the threads.

They don't really _need_ to be closed, but it's definitely a good idea to close them like the code now does.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
eric-maynard: That sounds like a good idea... is this [the one from jsr305](https://www.javadoc.io/static/com.google.code.findbugs/jsr305/3.0.2/javax/annotation/WillClose.html)? It doesn't seem that we have that dependency currently, and we're not using it anywhere else in the project.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
eric-maynard: Currently no, the only resource is the Statement:
```
try (Statement statement = connection.createStatement())
```

But if we added the BufferedReader there too, it might close the stream. Do you want to do that instead?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
eric-maynard: It returns an InputStream, so I think that should hopefully be obvious?

dimas-b: Obvious... not sure. I'd read it as returning the same `InputStream` on every call (pre-computed data).

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
dimas-b: Yeah, if possible... I think try-with-resources is more intuitive and we get better exception propagation OOTB.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
dimas-b: yes, that jar :+1: TBH, I was surprised we did not have any existing use cases :sweat_smile:

eric-maynard: I noticed this is in the `banned-dependencies.txt`?
```
# Contains old javax.* annotations that we do not want
com.google.code.findbugs:jsr305
```

dimas-b: TIL... well, I'm fine with _not_ using it in this case.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
dimas-b: Do we still need this call to `.close()` now that we have the `BufferedReader` inside the try-with-resources?

dimas-b (approved):
LGTM with one minor comment about possibly unnecessary `.close()` call (old comment thread)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
eric-maynard: Me too! I learned this when I went to add the dep 😅

If there is another similar annotation we start using in the future that would be nice as this does sound pretty helpful.

eric-maynard:
Sorry, looks like the old auto-merge went in with your comment still outstanding @dimas-b! That was set a while ago before this InputStream business. `close` is generally safe to double-call, but I agree that the second call shouldn't be necessary. If you want, I can quickly followup and try to remove that.

dimas-b:
No worries, but if you do not mind, let's remove the unnecessary close call. I'm not so much worried about calling it twice, but the surrounding try/catch code makes the method a bit more complex than necessary :)

eric-maynard:
Sounds good @dimas-b, please check PR#1982 and sorry again about the auto merge!

PR #1944, opened 2025-06-25
Update Helm chart versions

flyrain:


helm/polaris/Chart.yaml
```diff
@@ -21,8 +21,8 @@ apiVersion: v2
 name: polaris
 description: A Helm chart for Apache Polaris (incubating)
 type: application
-version: 0.1.0
-appVersion: 1.0.0-incubating-SNAPSHOT
+version: 1.0.0
```
adutra, line 24: Do we want the Helm chart versioning to always follow the versioning of the underlying app?

flyrain, line 24: Helm uses `version` to track the chart package itself (for repo indexes, upgrades, roll-backs, dependency resolution). It has nothing to do with the container image. It must be valid SemVer. It was recommend to update every time we update the helm chart itself.

helm/polaris/Chart.yaml
```diff
@@ -21,8 +21,8 @@ apiVersion: v2
...
+appVersion: 1.0.0-incubating
```
flyrain, line 25: This is the docker image tag. The 1.0.0-incubating docker image will be published after the vote passed per release guide suggsted.

helm/polaris/Chart.yaml
flyrain: We don't have to, but we need to update it every time we update the helm-chart itself, see details here, https://github.com/apache/polaris/pull/1944#discussion_r2167445639. I'm OK with other version number, any suggestions?

adutra: Yes, I know that. But why jump from 0.1.0 to 1.0.0 then?

helm/polaris/Chart.yaml
adutra: `appVersion` has nothing to do with the docker image tag. The docker image tag is set in `values.yaml` at `image.tag`.

adutra: We must keep in mind that a Helm chart published under `appVersion` X should stay compatible with a wide range of Docker images, roughly going from X-N to X, where N is yet to be defined. At some point, it could be useful to keep a formal compatibility matrix somewhere.

helm/polaris/Chart.yaml
flyrain: No particular reason, can you suggest a version?

adutra: I'm OK with 1.0.0, and I am OK releasing the Helm chart every time we release Polaris, updating `version` and `appVersion`.

I am just mentioning this because we never finished the discussion started here:

https://lists.apache.org/thread/rrsjygxdltnbtldyo3zl9x0v0h1t614h

And so I don't know if there is a consensus here.

helm/polaris/Chart.yaml
flyrain: Thanks for the correction. In that case, should we update the tag from `latest` to `1.0.0-incubating` in file `values.yaml`?

flyrain: > We must keep in mind that a Helm chart published under appVersion X should stay compatible with a wide range of Docker images, roughly going from X-N to X, where N is yet to be defined. At some point, it could be useful to keep a formal compatibility matrix somewhere.

How do we archive the the range compatibility? I guess X is `1.0.0-incubating` for the Polaris release `1.0.0-incubating`. Is there any default value for N?

adutra: There are pros and cons:

* `latest` is very convenient, but at the same time, not very precise.
* Specifying a fixed version in `image.tag`, on the contrary, has the merit of making it clear that the Helm chart works best for that specific tag.

I am personally in favor of using `image.tag = latest`, because if people also install the latest version of the helm chart, and provided that we publish images and charts at every release, then they should automatically get a latest chart version that works well with the latest docker image version.

helm/polaris/Chart.yaml
flyrain: Thanks for sharing the thread, that's helpful. I didn't see any objection there. I'm also supportive of `release helm charts separately from the main source / binary bundle`.

flyrain: > And so I don't know if there is a consensus here.

two +1s and no objection for a long time, which is usually considered as a consensus. It'd be nice to conclude it though.

helm/polaris/Chart.yaml
flyrain: LGTM overall. The only concern is that if users want to revert to a historical version of helm chart, do they expect to still get the latest docker image?

helm/polaris/Chart.yaml
flyrain: oh, I see different approaches proposed. 
1. Release helm chart separately from the main source / binary bundle from Dmitri
2. Release Helm charts as part of a Polaris release from JB.

We will need more discussion on the topic.

flyrain:
I will convert this to draft as there are open discussions on helm chart release and tag.

helm/polaris/Chart.yaml
flyrain: Here is an example, which assumes Helm chart version matched with Polaris src/bin/docker version:
1. When an user revert to the Helm Chart v1.0.0, it still points the `latest` docker image, which could be 1.1.0 already. This introduces the incompatible risks. On the other hand, if we specified a version(e.g., 1.0.0) instead of `latest`, the Helm reversion will always be same.
2.  if people install the latest version of the helm chart, it still points to the latest docker image as we release them together. So that use case is covered as well.

With that, I think specifying a version is preferrable than `latest`. WDYT?

adutra: That's a valid point! `latest` is not very friendly to historical chart versions. I am OK with using the release version instead.

helm/polaris/Chart.yaml
adutra: Thanks for reviving the discussion 🙏

flyrain: Sure, would you mind chiming in a bit. I hope it can conclude soon. I'm OK if we didn't figure out helm chart release for 1.0.0, but it'd be nice to have it.

dimas-b: Should we move this discussion to the dev ML?

adutra: Yufei revived that old thread, and I just added my +1 to his suggestions, FYI.

flyrain: Yeah, we did, here are new discussions: https://lists.apache.org/thread/d1vf7xpn6nkzp8gbh417m8qb58tkpcqz

sfc-gh-ygu:
Thanks @adutra and @dimas-b for the review. Will file a similar PR to main soon.

Issue #1943, opened 2025-06-25
Quick start failed on polaris-bootstrap


MonkeyCanCode:
### Describe the bug

While trying out the latest getting start example with jdbc from main branch, polaris-bootstrap service will fail with the docker compose command. 

### To Reproduce

1. Build from source
```
➜  ~ cd polaris
➜  polaris git:(main) ./gradlew \
  :polaris-server:assemble \
  :polaris-server:quarkusAppPartsBuild \
  :polaris-admin:assemble --rerun \
  -Dquarkus.container-image.tag=postgres-latest \
  -Dquarkus.container-image.build=true
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   167  100   167    0     0   1318      0 --:--:-- --:--:-- --:--:--  1325
100    64  100    64    0     0    237      0 --:--:-- --:--:-- --:--:--   237
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 43764  100 43764    0     0   199k      0 --:--:-- --:--:-- --:--:--  199k
Downloading https://services.gradle.org/distributions/gradle-8.14.2-bin.zip
.............10%.............20%.............30%.............40%.............50%.............60%.............70%.............80%.............90%..............100%

Welcome to Gradle 8.14.2!
...
BUILD SUCCESSFUL in 2m 49s
105 actionable tasks: 105 executed
```

2. run docker compose command
```
➜  polaris git:(main) export ASSETS_PATH=$(pwd)/getting-started/assets/
export QUARKUS_DATASOURCE_JDBC_URL=jdbc:postgresql://postgres:5432/POLARIS
export QUARKUS_DATASOURCE_USERNAME=postgres
export QUARKUS_DATASOURCE_PASSWORD=postgres
export CLIENT_ID=root
export CLIENT_SECRET=s3cr3t
docker compose -p polaris -f getting-started/assets/postgres/docker-compose-postgres.yml \
  -f getting-started/jdbc/docker-compose-bootstrap-db.yml \
  -f getting-started/jdbc/docker-compose.yml up -d
WARN[0000] The "STORAGE_LOCATION" variable is not set. Defaulting to a blank string.
WARN[0000] The "AWS_ROLE_ARN" variable is not set. Defaulting to a blank string.
[...]

PR #1945, opened 2025-06-25
Fix admin tool for quick start

MonkeyCanCode:
When attempting to use the `polaris-admin-tool.jar` to bootstrap a realm, the application fails with a `jakarta.enterprise.inject.UnsatisfiedResolutionException` because it cannot find a `javax.sql.DataSource` bean. Detail in https://github.com/apache/polaris/issues/1943

This issue occurs because `quarkus.datasource.db-kind`is a build-time property in Quarkus. Its value must be defined during the application's build process to enable the datasource extension and generate the necessary CDI bean producer (ref: https://quarkus.io/guides/all-config#quarkus-datasource_quarkus-datasource-db-kind).

I think we only support postgres for now, thus, I set `quarkus.datasource.db-kind=postgresql`. This can be problematic if we later want to support more data sources other than postgres. There are couple options we have for this such as use multiple named datasources in the config during build time. But this may be out of scope of this PR. I am open for more discussion on this, but for the time being, it may be better to unblock people who are trying to use the quick start doc.

Sample output for the bootstrap container after the fix:
```
➜  polaris git:(1943) docker logs polaris-polaris-bootstrap-1
Realm 'POLARIS' successfully bootstrapped.
Bootstrap completed successfully.
```


getting-started/jdbc/docker-compose-bootstrap-db.yml
```diff
@@ -22,7 +22,7 @@ services:
     image: apache/polaris-admin-tool:postgres-latest
     environment:
       - POLARIS_PERSISTENCE_TYPE=relational-jdbc
-      - QUARKUS_DATASOURCE_DB_KIND=pgsql
+      - QUARKUS_DATASOURCE_DB_KIND=postgresql
```
dimas-b: Since the property is built-time, we can probably remove it from here for clarity, WDYT?

runtime/admin/src/main/resources/application.properties
```diff
@@ -37,6 +37,7 @@ quarkus.container-image.additional-tags=latest
 
 polaris.persistence.type=relational-jdbc
 
+quarkus.datasource.db-kind=postgresql
```
flyrain: Can we move up a bit if this is a build time setting?  Anything below line 35 should be runtime properties.

flyrain: Hi @dimas-b , does it affect normal use cases beside Quick Start examples?

dimas-b: IDK... I'm not even sure why Quick Start is broken... I never hit this problem in my env, but I used bootstrap a lot with Postgres :shrug:

dimas-b: Good point on property location: the built-time section is at the top of this file.

dimas-b (approved):

PR #1949, opened 2025-06-26
Fix NPE in listCatalogs

andrew4699:
listCatalogs is non-atomic. It first atomically lists all entities and then iterates through each one and does an individual loadEntity call. This causes an NPE when calling `CatalogEntity::new`.

I don't think it's ever useful for listCatalogsUnsafe to return null since the caller isn't expecting a certain length of elements, so I just filtered it there.

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/ManagementServiceTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/ManagementServiceTest.java
@@ -43,10 +43,16 @@ import org.apache.polaris.core.auth.AuthenticatedPolarisPrincipal;
 import org.apache.polaris.core.auth.PolarisAuthorizerImpl;
 import org.apache.polaris.core.context.CallContext;
 import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.core.entity.PolarisBaseEntity;
+import org.apache.polaris.core.entity.PolarisEntity;
+import org.apache.polaris.core.entity.PolarisEntityConstants;
+import org.apache.polaris.core.entity.PolarisEntitySubType;
+import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PrincipalEntity;
 import org.apache.polaris.core.entity.PrincipalRoleEntity;
 import org.apache.polaris.core.persistence.MetaStoreManagerFactory;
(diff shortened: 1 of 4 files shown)


runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/ManagementServiceTest.java
```diff
@@ -276,4 +282,41 @@ public void testCannotAssignFederatedEntities() {
...
+            new PolarisBaseEntity(
+                PolarisEntityConstants.getNullId(),
+                metaStoreManager.generateNewEntityId(callContext).getId(),
+                PolarisEntityType.CATALOG,
+                PolarisEntitySubType.NULL_SUBTYPE,
+                PolarisEntityConstants.getRootEntityId(),
+                "my-catalog-2"),
+            List.of());
+
+    metaStoreManager.fakeEntityNotFoundIds.add(catalog1.getCatalog().getId());
```
dimas-b: Would you mind making a setter/adder method for this as opposed to accessing a public field?

runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/TestPolarisMetaStoreManager.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.polaris.core.entity.PolarisEntityType;
+import org.apache.polaris.core.persistence.dao.entity.BaseResult;
+import org.apache.polaris.core.persistence.dao.entity.EntityResult;
+import org.apache.polaris.core.persistence.transactional.TransactionalMetaStoreManagerImpl;
+
+/**
+ * Intended to be a delegate to TransactionalMetaStoreManagerImpl with the ability to inject faults.
+ * Currently, you can force loadEntity() to return ENTITY_NOT_FOUND for a set of entity IDs.
+ */
+public class TestPolarisMetaStoreManager extends TransactionalMetaStoreManagerImpl {
```
dimas-b: Why not make it an inner class in `ManagementServiceTest`?

dimas-b (commented):
Thanks for the fix @andrew4699 ! `PolarisAdminService` changes LGTM, but I have some minor comments in test code :)

andrew4699:
Updated!

dimas-b (approved):

PR #1951, opened 2025-06-26
Fix doc for sample log and default password

MonkeyCanCode:
Minor updates for the quick start doc:
1. current doc has wrong default password
2. update sample output to reflect with the latest code
3. remove trailing space

### DIFF
--- a/site/content/in-dev/unreleased/getting-started/quickstart.md
+++ b/site/content/in-dev/unreleased/getting-started/quickstart.md
@@ -42,7 +42,7 @@ cd ~/polaris
 
 ## Running Polaris with Docker
 
-To start using Polaris in Docker and launch Polaris, which is packaged with a Postgres instance, Apache Spark, and Trino. 
+To start using Polaris in Docker and launch Polaris, which is packaged with a Postgres instance, Apache Spark, and Trino.
 
 ```shell
 export ASSETS_PATH=$(pwd)/getting-started/assets/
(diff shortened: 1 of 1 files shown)


site/content/in-dev/unreleased/getting-started/quickstart.md
```diff
@@ -82,17 +82,17 @@ Run the following command to start Polaris:
...
 ```
 
 At this point, Polaris is running.
 
-When using a Gradle-launched Polaris instance in this tutorial, we'll launch an instance of Polaris that stores entities only in-memory. This means that any entities that you define will be destroyed when Polaris is shut down. 
+When using a Gradle-launched Polaris instance in this tutorial, we'll launch an instance of Polaris that stores entities only in-memory. This means that any entities that you define will be destroyed when Polaris is shut down.
 For more information on how to configure Polaris for production usage, see the [docs]({{% relref "../configuring-polaris-for-production" %}}).
 
-When Polaris is run using the `./gradlew run` command, the root principal credentials are `root` and `secret` for the `CLIENT_ID` and `CLIENT_SECRET`, respectively.
+When Polaris is run using the `./gradlew run` command, the root principal credentials are `root` and `s3cr3t` for the `CLIENT_ID` and `CLIENT_SECRET`, respectively.
```
eric-maynard, line 95: Oh, great catch

PR #1957, opened 2025-06-26
Update Helm chart version

flyrain:
Update Helm Chart versions per the consensus in the dev ML, https://lists.apache.org/thread/d1vf7xpn6nkzp8gbh417m8qb58tkpcqz.

### DIFF
--- a/helm/polaris/Chart.yaml
+++ b/helm/polaris/Chart.yaml
@@ -21,8 +21,8 @@ apiVersion: v2
 name: polaris
 description: A Helm chart for Apache Polaris (incubating)
 type: application
-version: 0.1.0
-appVersion: 1.0.0-incubating-SNAPSHOT
+version: 1.1.0-incubating-SNAPSHOT
+appVersion: 1.1.0-incubating-SNAPSHOT
 home: https://polaris.apache.org/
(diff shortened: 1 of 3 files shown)


helm/polaris/Chart.yaml
```diff
@@ -36,3 +36,4 @@ maintainers:
   - name: collado-mike
   - name: gerrit-k
   - name: snazy
+  - name: flyrain
```
flyrain, line 39: Not sure this maintainer list is needed. We didn't do that in somewhere else. Is it a Helm Chart thing?

dimas-b (approved):

helm/polaris/Chart.yaml
MonkeyCanCode: Yes, but it is an optional field (ref: https://helm.sh/docs/topics/charts/):
```
maintainers: # (optional)
  - name: The maintainers name (required for each maintainer)
    email: The maintainers email (optional for each maintainer)
    url: A URL for the maintainer (optional for each maintainer)
```

flyrain: Thanks for sharing. In that case, I'd propose to remove the list in a followup PR, as it adds maintenance burden, which makes it inaccurate inevitably. WDYT?

MonkeyCanCode: Yes. I think that is probably better in this case.

PR #1952, opened 2025-06-26
Add CHANGELOG

dimas-b:
* Add rudimentary CHANGELOG.md

* Add the Jetbrains Changelog Gradle plugin to help managing CHANGELOG.md

Dev list discussion: https://lists.apache.org/thread/qznf8toht1r7ml35lt4p8nwlk9op638v

PR #1959, opened 2025-06-26
Rename SUPPORTED_CATALOG_CONNECTION_TYPES 

poojanilangekar:
To ensure backward compatibility, process to rename `SUPPORTED_CATALOG_CONNECTION_TYPES` will consist of two steps. 

1. Add `SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES`. 
2. Remove the legacy `SUPPORTED_CATALOG_CONNECTION_TYPES`.

As discussed in PR #1931 , for a while Polaris will initially support both `SUPPORTED_CATALOG_CONNECTION_TYPES` and `SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES`. Later, we will remove the `SUPPORTED_CATALOG_CONNECTION_TYPES` to make it consistent with `AUTHENTICATION_TYPES`. 

To support the rename, this PR introduces a new PolarisConfiguration parameter called `legacyKey`.


poojanilangekar:
CC @dimas-b @eric-maynard

dimas-b:
Let's hold this until #1952 merges so that we could properly track the property rename in CHANGELOG.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -266,10 +266,18 @@ public static void enforceFeatureEnabledOrThrow(
   public static final FeatureConfiguration<List<String>> SUPPORTED_CATALOG_CONNECTION_TYPES =
       PolarisConfiguration.<List<String>>builder()
           .key("SUPPORTED_CATALOG_CONNECTION_TYPES")
-          .description("The list of supported catalog connection types for federation")
+          .description("(Deprecated) The list of supported catalog connection types for federation")
```
dimas-b: I believe it would be more maintainable to have one `FeatureConfiguration` object which supported both the old and the new key names (cf. `catalogConfigUnsafe()`).

eric-maynard: I see a few options:
1. Introduce a framework to support multiple config names, with some of them marked for deprecation, similar to `catalogConfigUnsafe`
2. Just leave the name alone
3. Just make the breaking change and switch the config name
4. Change the variable name but not the config name
5. Add a second config with a new name, change the logic to take the union of the two, and mark the old one deprecated

I think all of these are fine but (1) is definitely the most work. Looks like this is (5) but this random `(Deprecated)` comment is doing a lot of heavy lifting. We don't get logging like we would get with (1) and the code is less clean.

poojanilangekar: Done, I've implemented 1.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -263,12 +264,14 @@ public static void enforceFeatureEnabledOrThrow(
...
-      PolarisConfiguration.<List<String>>builder()
-          .key("SUPPORTED_CATALOG_CONNECTION_TYPES")
-          .description("The list of supported catalog connection types for federation")
-          .defaultValue(List.of(ConnectionType.ICEBERG_REST.name()))
-          .buildFeatureConfiguration();
+  public static final FeatureConfiguration<List<String>>
+      SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES =
+          PolarisConfiguration.<List<String>>builder()
+              .key("SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES")
+              .deprecatedKey("SUPPORTED_CATALOG_CONNECTION_TYPES")
```
dimas-b: I'd prefer a name like `.legacyKey()` or `.compatibilityKey()` for the sake of clarity. "Deprecated" key as a builder parameter does not convey a reason for this key to exist.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -112,6 +112,7 @@ polaris.features."ENFORCE_PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_CHECKING"=false
 polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE"]
 # polaris.features."ENABLE_CATALOG_FEDERATION"=true
 polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
+polaris.features."ENABLE_EXTERNAL_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
```
dimas-b: Is this change relevant to current PR's scope (and title)?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
poojanilangekar: Done.

runtime/defaults/src/main/resources/application.properties
poojanilangekar: Sorry about that. It was added by auto complete. I have removed it.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
eric-maynard: I think if we're really doing this we need to take multiple keys, so like
`PolarisConfiguration.<>builder().keys(...)`. Maybe that should take some `ConfigurationKey`, or maybe we should have some other mechanism for marking which are deprecated.

dimas-b: Multiple keys SGTM.... but from the builder API POV, we could just accept multiple calls to `.key()`,  `.deprecatedKey()` (or `.legacyKey()`), right?

dimas-b (approved):
Current changes in this PR LGTM :+1: I'd be ok with merging now and enhancing later, but I'm also fine with implementing @eric-maynard 's suggestions in this PR.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
poojanilangekar: I can make the changes. Just so we are on the same page, does the following approach make sense to both of you?:

1. Support a single `key` per config, because ideally at any given time, we want at most one active key per configuration. 
2. Support multiple `legacyKeys` per config. So we can make multiple calls to `legacyKey()` while building a config.

eric-maynard: Whether we use one `key` call that takes N strings or we allow `N` chained calls to `key` / `deprecatedKey`, the point is we should support N different keys rather than 2.

poojanilangekar: Done.

poojanilangekar:
Hi @eric-maynard and @dimas-b can you please take a look at the change to support multiple `legacyKeys`?

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 ### Changes
 
 ### Deprecations
+- Deprecates the `SUPPORTED_CATALOG_CONNECTION_TYPES` and replaces it with `SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES`.
```
dimas-b:
```suggestion
- Deprecates the `SUPPORTED_CATALOG_CONNECTION_TYPES` configuration property and replaces it with `SUPPORTED_EXTERNAL_CATALOG_CONNECTION_TYPES`. 
```

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -41,6 +43,7 @@ public abstract class PolarisConfiguration<T> {
 
   public final String key;
   public final String description;
+  public final Set<String> legacyKeys;
```
dimas-b, line 46: Is this value used for property config value lookup? Sorry, if I missed it in the diff :thinking:

poojanilangekar: Yes, it is used when registering a new configuration. 

Previously (before this change), we would go over existingConfigurations check if it matched any existingConfiguration's key. Now, we check if the new configuration (both key and legacyKeys) match either an existing key (line 61) or existing legacyKeys (line 65 - 76)

dimas-b: Correct... but we should also use it to find config values in case the `key` is not found, right?

poojanilangekar: The way I have implemented it, any configuration always has a `key`(I think that's a valid requirement). Is there a reason we would we keep a configuration with just `legacyKeys`?

poojanilangekar: Actually, I think I get what you mean. I am guessing you meant the aspect of reading in the properties. If so I will send a fix shortly.

dimas-b: Thx! To clarify: old config entries should keep working until we drop support for them (e.g. in 2.0). Let's add a WARN log message when we have to fall back to the "legacy" property name so that users are notified in runtime and are more aware of the need to migrating to the new `key`.

poojanilangekar: Done.

polaris-core/src/test/java/org/apache/polaris/service/storage/PolarisConfigurationStoreTest.java
```diff
@@ -186,4 +185,104 @@ public void testEntityOverrides() {
...
+            return null;
+          }
+        };
+
+    String result = store.getConfiguration(testRealmContext, config);
+    Assertions.assertEquals("default-value", result);
+  }
+
+  @Test
+  public void testLegacyKeyWarningIsLogged() {
```
dimas-b: How does this test ensure the warning is logged?

dimas-b: I'd be ok without this test too :)

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
dimas-b, line 46: Sorry, I did not notice this before, but I think having `public` fiends in general is not a good idea. Since you're toughing this code in current PR, would you mind following up with another PR to convert them to `private` fields?

dimas-b (approved):
LGTM, thanks, @poojanilangekar !

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
poojanilangekar: Sure, I will send out a follow up PR.

dimas-b:
@eric-maynard : since you had previous comments, PTAL.

polaris-core/src/test/java/org/apache/polaris/service/storage/PolarisConfigurationStoreTest.java
poojanilangekar: This was my mistake, I should have removed it, I was trying to get the log warning but couldn't capture it. I've removed it now. (The functionality is still tested in testLegacyKeyLookup)

dimas-b: sgtm :+1:

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -86,18 +102,32 @@ public static List<PolarisConfiguration<?>> getAllConfigurations() {
...
   }
 
+  public boolean hasLegacyKeys() {
+    return !legacyKeys.isEmpty();
+  }
+
+  public String legacyKeys() {
+    if (legacyKeys.isEmpty()) {
+      throw new IllegalStateException(
+          "Attempted to read legacy keys from a configuration that doesn't have any.");
```
eric-maynard: This doesn't look right to me; the check is there for `catalogConfig` because it's not correct to call the variant of `getConfiguration` which takes a catalog when a config doesn't have a `catalogConfig`. However we should always be checking legacyKeys every time we load any config.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -86,18 +102,32 @@ public static List<PolarisConfiguration<?>> getAllConfigurations() {
...
     this.catalogConfigImpl = catalogConfig;
     this.catalogConfigUnsafeImpl = catalogConfigUnsafe;
     this.typ = (Class<T>) defaultValue.getClass();
   }
 
+  public boolean hasLegacyKeys() {
+    return !legacyKeys.isEmpty();
+  }
+
+  public String legacyKeys() {
```
eric-maynard: I would not expect this to return a `String` either.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
```diff
@@ -99,7 +99,25 @@ public interface PolarisConfigurationStore {
    */
   default <T> @Nonnull T getConfiguration(
       @Nonnull RealmContext realmContext, PolarisConfiguration<T> config) {
-    T result = getConfiguration(realmContext, config.key, config.defaultValue);
+    T result = getConfiguration(realmContext, config.key);
+
+    if (result == null && config.hasLegacyKeys()) {
+      for (String legacyKey : config.legacyKeys) {
```
eric-maynard: If `legacyKeys` just returned a set (or there was a `keys` that returned a set) then the if statement here goes away.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
```diff
@@ -99,7 +99,25 @@ public interface PolarisConfigurationStore {
...
+        result = getConfiguration(realmContext, legacyKey);
+        if (result != null) {
+          LOGGER.warn(
+              "Legacy configuration key '{}' is in use. Please use '{}' instead.",
+              legacyKey,
+              config.key);
+          break;
```
eric-maynard: Not sure we want a break here
1. If multiple legacy keys really are in use, we should want to log for each of them
2. Barring the above, you could just early return rather than `break`

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
poojanilangekar: I've removed it.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
poojanilangekar: Done.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
poojanilangekar: Done.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
poojanilangekar: I don't think we can use an early return directly because we need to try and cast the value to its defined type. We could duplicate the code here but I think this is cleaner.

eric-maynard: The cast is just `tryCast(config, result)` which is not a crazy amount of logic to replicate. But I wouldn't advocate for the early return either, because of reason (1) above

poojanilangekar: Regarding (1), I don't think that is accurate, even with the default key, there is an [order of precedence](https://quarkus.io/guides/config-reference#configuration-sources) in which properties are applied.  When multiple keys are found, Quarkus picks the first property that is encounters. So I'm unsure why we'd want to apply multiple legacy keys. Unless you're suggesting simply logging multiple legacy keys while only accepting the fist legacy key. Please let me know.

eric-maynard: > 1. If multiple legacy keys really are in use, we should want to log for each of them

Yes, I'm referring to the logging.

poojanilangekar: Done.

polaris-core/src/main/java/org/apache/polaris/core/config/BehaviorChangeConfiguration.java
```diff
@@ -34,11 +35,12 @@ public class BehaviorChangeConfiguration<T> extends PolarisConfiguration<T> {
 
   protected BehaviorChangeConfiguration(
       String key,
```
eric-maynard, line 37: Why can you have multiple `legacyKeys` but not multiple `keys`? It seems like maybe both should be `Set`s that support multiple calls.

poojanilangekar: Here is my understanding: A PolarisConfiguration or a configuration for any quarkus app has a `name` (`key` in PoalrisConfiguration) that is used to identify it.  The key/name is the identity of the configuration. The point of adding support for legacyKeys was to enable identity changes wherein we can for certain time support deprecated keys. I don't see why we should have **multiple identifying keys** mapping to a **single entity**. If you ever need multiple identities, then the legacy keys do support the need but eventually each entity must converge to a single identity. 

Please let me know if I am missing something here.

eric-maynard: It just seems odd that we would support multiple _invalid_ keys but not multiple _valid_ keys. In my original comment [here](https://github.com/apache/polaris/pull/1959#discussion_r2173075151) I wrote:
> Introduce a framework to support multiple config names, with some of them marked for deprecation, similar to catalogConfigUnsafe

rather than `... with all but one marked for deprecation ...` because I had this in mind.

I do agree that the situation where you'd have multiple valid configs seems contrived, but so does the situation where you'd have multiple invalid configs or really even one invalid config. So if we're overhauling the whole key system, I think it makes sense to go all the way.

poojanilangekar: I don't think we support multiple invalid keys. We track multiple `legacy` keys. 

From a semantic perspective, a `PrimaryKey` in a RDBMS, a `key`in a key-value store or a Java Map (or any other programming language) maps to a single row, value or entity. In all these examples, two keys never map to the same entity. 

What we have with the `legacyKeys` is a way to track historic versions of the key-value pair. In any DBMS or KV store that supports multi-versioning, it's pretty normal to have multiple entries (multiple versions) for a single key. When the primaryKey/key changes, the DBMS/KV store adds a tombstone for the old record and creates a new record with the new key.  While changing keys in PolarisConfig, we are essentially creating  a new key-value pair in itself but its nice to track the older key, hence the support for legacy keys. Essentially one could rewrite the entire PolarisConfig class with a Map where the `key` is the active key and the value is everything else. In which case, we'd track the `legacyKeys` as a part of the value.

eric-maynard: The key is not a primary key though. There's no RDBMS or KV store storing these; they describe entries in a config file. It seems reasonable to expect that multiple different entries could be used to configure the same feature.

eric-maynard: Consider, too, that the current code needs to deal with a situation where the key has value `true` but the legacyKey has value `false`. Or two legacy keys have different values. Clearly, the key(s) is not unique and the code needs to handle multiple keys for a given feature.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
```diff
@@ -99,7 +99,27 @@ public interface PolarisConfigurationStore {
    */
   default <T> @Nonnull T getConfiguration(
       @Nonnull RealmContext realmContext, PolarisConfiguration<T> config) {
-    T result = getConfiguration(realmContext, config.key, config.defaultValue);
+    T result = getConfiguration(realmContext, config.key);
+
+    if (result == null && config.hasLegacyKeys()) {
```
eric-maynard, line 104: So the key overrides the legacy keys?

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
```diff
@@ -99,7 +99,27 @@ public interface PolarisConfigurationStore {
...
+      for (String legacyKey : config.legacyKeys()) {
```
eric-maynard, line 105: This will lead to nondeterministic behavior if two legacy keys for a feature have different values

polaris-core/src/test/java/org/apache/polaris/service/storage/PolarisConfigurationStoreTest.java
```diff
@@ -186,4 +185,78 @@ public void testEntityOverrides() {
...
+            return (T) values.get(configName);
+          }
+        };
+
+    String result = store.getConfiguration(testRealmContext, config);
+    Assertions.assertEquals("legacy-value-1", result);
+  }
+
+  @Test
+  public void testMainKeyTakesPrecedenceOverLegacyKeys() {
```
eric-maynard, line 215: If this happens, we should probably actually error

polaris-core/src/test/java/org/apache/polaris/service/storage/PolarisConfigurationStoreTest.java
```diff
@@ -186,4 +185,78 @@ public void testEntityOverrides() {
     Assertions.assertEquals(
         "entity-legacy4", store.getConfiguration(testRealmContext, entity, cfg.apply(4)));
   }
+
+  @Test
```
eric-maynard, line 189: It would be nice to add a test for conflicting legacy keys

polaris-core/src/main/java/org/apache/polaris/core/config/BehaviorChangeConfiguration.java
poojanilangekar: I disagree with this:

> It seems reasonable to expect that multiple different entries could be used to configure the same feature.

You can have two configurations that store the same logical value and then let the specific module decide upon how to reconcile any conflicting values. But the config service should not register two configurations for the same feature. 

Regarding the case where the `key` is `true` and `legacyKey` is false. I don't see how it is any different from a default value being defined say FeatureConfiguration.java, a value being defined in application.properties and then a JVM system property. In this case the system property takes precedence. Similarly the `key` takes precedence over `legacyKey`.

eric-maynard: > I disagree with this:

> > It seems reasonable to expect that multiple different entries could be used to configure the same feature.

Your current PR has this functionality... except only partially:

| Keys  | Legacy Keys | Multiple Entries | OK? | 
| ------------- | ------------- | --- | --- |
| 0 | 0 | No | Yes |
| 0 | 1  | No | Yes |
| 0 | 2+  | Yes | Yes |
| 1 | 0  | No | Yes |
| 1 | 1  | Yes | Yes |
| 1 | 2+ | Yes | Yes |
| 2 | 0  | Yes | No |
| 2 | 1  | Yes | No |
| 2 | 2+ | Yes | No |

What is the rationale behind this asymmetry? If, as stated before, it is because the key is thought of as a primary key then this is just not correct. You can write a valid config that violates the "primary key constraint" by duplicating an entry.

eric-maynard (changes requested):
This PR is now doing more than just a rename, and we should figure out what model we actually want to do for configs before going ahead with the change.

polaris-core/src/main/java/org/apache/polaris/core/config/BehaviorChangeConfiguration.java
poojanilangekar: The case with 0 keys is not possible. We can't define a PolarisConfig without a key. That was always the case. Unless you'd prefer if we annotate the field with `@NonNull` in the constructor. 

> You can write a valid config that violates the "primary key constraint" by duplicating an entry.

The `registerConfiguration` function will actually disallow duplicate entries and throw an `IllegalArgumentException`. That check is a part of the existing code, this PR is not change it. This PR goes further and checks that any existing `key` or `legacyKey` does not overlap with a new configuration that is being registered (again both `key` and `legacyKeys`).    


I think we're looking at two entirely different aspects of a configuration. 
1. Declaring the configuration in `FeatureConfiguration` or `BehaviorChangeConfiguration`. 
2. Assigning values to a configuration which is done in application.properties / JVM arguments. 

The two are orthogonal to each other. We should impose constraints on how PolarisConfigs are declared. We can't however impose constraints on how the user defines these configs. If the user does have multiple definitions of the same config, then we apply the one with the highest priority and discard the rest. If indeed we should disallow multiple definitions, then any config that has a default value defined should essentially disallow any overwrites in application.properties. AFAIK, Quarkus doesn't do that and there isn't a way for the app to determine such a condition.

poojanilangekar:
Hey @eric-maynard, I agree that this change has gotten larger than what I had initially planned. I am planning to close this PR because I think it's orthogonal to the goal of the PR. 

If @dimas-b and you are amenable to making the breaking change and doing the rename, I can update this PR. 

Otherwise, we can leave the config `SUPPORTED_CATALOG_CONNECTION_TYPES` as is, so I can spend more time on other tasks.

poojanilangekar:
I'd prefer doing something similar to #2000. (The PR adds a deprecated comment on the existing property and creates a new one.)

dimas-b:
I'd prefer using the old config name (for now) to making a breaking change.

Appreciate your effort on this PR @poojanilangekar !

dimas-b:
This PR deals with Polaris-specific config, so I'm not sure #2000 is applicable to this rename "as is". The tricky part is that Polaris config can be realm-specific and we cannot provide redirects in those cases (realm names are unknown at compile time).

dimas-b:
It is possible to have custom code to deal with Smallrye configs at a low level and provide property redirects, but I think it'll be much more complex in the end.

PR #1962, opened 2025-06-27
Remove the maintainer list in Helm Chart

flyrain:
The reason to remove it
1. The maintainer list is optional. 
2. It adds code maintenance burden, which make it inaccurate inevitably.
3. We don't do that in elsewhere of the project. 

More discussions are here, https://github.com/apache/polaris/pull/1957#discussion_r2170162461.

### DIFF
--- a/helm/polaris/Chart.yaml
+++ b/helm/polaris/Chart.yaml
@@ -30,7 +30,7 @@ sources:
 keywords:
   - polaris
   - iceberg
-maintainers: # Maintainers section is kept only to satisfy the Helm Chart Spec and downstream linters.
+maintainers:  # Maintainers section is kept only to satisfy the Helm Chart Spec and downstream linters.
   - name: MonkeyCanCode
   - name: adutra
   - name: collado-mike
--- a/helm/polaris/Chart.yaml
+++ b/helm/polaris/Chart.yaml
@@ -30,3 +30,10 @@ sources:
 keywords:
   - polaris
   - iceberg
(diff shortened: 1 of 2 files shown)


MonkeyCanCode:
@flyrain needed to update CI with `--validate-maintainers=false` such as following:
```
run: ct lint --target-branch ${{ github.event.repository.default_branch }} --validate-maintainers=false --debug --charts ./helm/polaris
```

flyrain:
The linting failed. 
```
Error: failed linting charts: failed processing charts
Validating maintainers...

------------------------------------------------------------------------------------------------------------------------
 ✖︎ polaris => (version: "1.1.0-incubating-SNAPSHOT", path: "./helm/polaris") > chart doesn't have maintainers
------------------------------------------------------------------------------------------------------------------------
failed linting charts: failed processing charts
Error: Process completed with exit code 1.
```

flyrain:
Hi @MonkeyCanCode , maybe we should keep that,  my concern is that the downstream projects may break accidentally if we remove it. Their tooling may not have `--validate-maintainers=false`. WDYT? Let me also document it a bit, so that no extra maintenance work is encouraged.

MonkeyCanCode:
> Hi @MonkeyCanCode , maybe we should keep that, my concern is that the downstream projects may break accidentally if we remove it. Their tooling may not have `--validate-maintainers=false`. WDYT? Let me also document it a bit, so that no extra maintenance work is encouraged.

Yeah, both will work as it is really an optional filed in helm. CT by default will expect maintainer to be available and can be turn off with the above option.

MonkeyCanCode (approved):

PR #1970, opened 2025-06-27
Undeprecate `catalogConfigUnsafe()`

dimas-b:
As discussed in #1894, `catalogConfigUnsafe()` has a use case in the main codebase - that is to support gradual migration to new property names.

Using legacy names in runtime already leads to a user-face WARN log message.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -84,7 +84,6 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
           .defaultValue(false)
           .buildFeatureConfiguration();
 
-  @SuppressWarnings("deprecation")
   public static final FeatureConfiguration<Boolean> ALLOW_TABLE_LOCATION_OVERLAP =
       PolarisConfiguration.<Boolean>builder()
           .key("ALLOW_TABLE_LOCATION_OVERLAP")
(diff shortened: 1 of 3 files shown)


eric-maynard:
> Support for those legacy property names is to be dropped in 2.0.

If that's true, wouldn't we also drop the method `catalogConfigUnsafe`? And, if so, isn't it fair to say the method is marked as deprecated for removal in 2.0?

dimas-b:
I removed `Support for those legacy property names is to be dropped in 2.0.` from the description for now. We can decide what to unsupport and when on a case-by-case basis.

dimas-b:
> [...] isn't it fair to say the method is marked as deprecated for removal in 2.0?

I'd expect that we may want to rename other properties as the project matures... `catalogConfigUnsafe()` might still be useful after 2.0.

eric-maynard:
catalogConfigUnsafe is specific to properties that don’t start with the
polaris prefix and are therefore unregistered. I believe there’s an assert
for this and the Javadoc details the same. So there should be no new usage
as long as we don’t introduce any more unsafe configs. If we migrate other
config names, we would so through a different mechanism that preserves the
registration of both names.

dimas-b:
I see your point now @eric-maynard . In this case, I believe changes in #1894 are actually correct (no need to span build logs with warnings about known calls to `catalogConfigUnsafe()`) and sufficient for now. We should not add any new calls to `catalogConfigUnsafe()` and probably remove it in 2.0. I'll file an issue for that.

dimas-b:
Cf. #1974

PR #1981, opened 2025-06-30
Add Delegation Service Module and REST Framework

williamhyun:
This PR aims to enable a separate, delegated mode of task execution starting with the REST API framework.

Discussion thread: 
- https://lists.apache.org/thread/ph10th4ocjczpf5gz17mqys4fkp5qrzw

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/DelegationServiceConfiguration.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/DelegationServiceConfiguration.java
@@ -69,7 +69,8 @@ public class DelegationServiceConfiguration {
    */
   public void validate() {
     if (enabled && (baseUrl == null || baseUrl.trim().isEmpty())) {
-      throw new IllegalArgumentException("Delegation service baseUrl must be provided when enabled");
+      throw new IllegalArgumentException(
+          "Delegation service baseUrl must be provided when enabled");
     }
 
     if (timeoutSeconds <= 0) {
--- a/runtime/service/src/main/java/org/apache/polaris/service/delegation/DelegationClient.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/delegation/DelegationClient.java
@@ -49,8 +49,8 @@ public interface DelegationClient {
   /**
    * Delegates table data file cleanup to the delegation service.
    *
-   * Note: This only handles data file deletion from storage. Metadata drop from the catalog
-   * is still handled by Polaris after successful delegation.
+   * <p>Note: This only handles data file deletion from storage. Metadata drop from the catalog is
+   * still handled by Polaris after successful delegation.
    *
    * @param catalogName the name of the catalog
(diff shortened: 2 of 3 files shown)


williamhyun:
cc: @eric-maynard, @flyrain

eric-maynard (approved):

dimas-b:
Please allow some more review time. From my POV the dev list discussion does not have a conclusion yet :thinking:

delegation-service/src/main/java/org/apache/polaris/delegation/api/DelegationApi.java
```diff
@@ -0,0 +1,99 @@
...
+ * allows the main Polaris catalog to offload resource-intensive operations to maintain low-latency
+ * performance.
+ *
+ * <p><strong>Note:</strong> This is the initial API framework implementation. The actual task
+ * execution logic will be implemented in future development phases.
+ */
+@Path("/api/v1/tasks/execute")
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+public class DelegationApi {
```
jbonofre: Can we add `@experimental` here ?

eric-maynard: This isn't a Polaris API, and within the context of the delegation service (itself experimental), this is not really an experimental API.

eric-maynard:
Hey @dimas-b, is there a particular thread on the ML you're looking to close? The most recent message from you just says `I'll certainly have a deeper look later` but that was a couple of weeks ago. I don't see a message from @snazy, who -1'd.

delegation-service/src/main/java/org/apache/polaris/delegation/api/DelegationApi.java
jbonofre: My proposal was more to "flag" it experimental (from code perspective). If the "whole" service is experimental, I'm fine with that.

Maybe a couple of tests to illustrate how it works would be welcome.

eric-maynard: From reading the proposal, I think what will actually be experimental will be the `RemoteTaskExecutor` within Polaris, but that's not implemented in this PR.

williamhyun: Thank you @jbonofre for your concern! As @eric-maynard has mentioned, the task executor for delegating to the remote service can be flagged as experimental. 

In regards to illustrating how it would all work, an end-to-end test is in preparation.

snazy:
The question is which approach two use. There are _two_ proposals on the dev-ML, with no technical pro/con discussion and therefore no decision. As mentioned in other discussions, things do take time. The whole topic is too sensitive and needs to be handled carefully, considering all aspects.

Having one proposal not receiving comments and weeks after another, different counter-proposal plus a concrete implementation is not great.

PR #1982, opened 2025-06-30
Remove unnecessary `InputStream.close` call

eric-maynard:
#1942 changed the way that the bootstrap init script is handled, but it added an extra `InputStream.close` call that shouldn't be needed after the BufferedReader [here](https://github.com/apache/polaris/pull/1942/files#diff-de43b240b5b5e07aba7e89f5515a417cefd908845b85432f3fcc0819911f3e2eR89) is closed. This PR removes that extra call.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -82,37 +82,38 @@ public class DatasourceOperations {
    * @throws SQLException : Exception while executing the script.
    */
   public void executeScript(InputStream scriptInputStream) throws SQLException {
-    runWithinTransaction(
-        connection -> {
-          try (Statement statement = connection.createStatement();
-              BufferedReader reader =
-                  new BufferedReader(
-                      new InputStreamReader(Objects.requireNonNull(scriptInputStream), UTF_8))) {
-            StringBuilder sqlBuffer = new StringBuilder();
-            String line;
-            while ((line = reader.readLine()) != null) {
-              line = line.trim();
-              if (!line.isEmpty() && !line.startsWith("--")) { // Ignore empty lines and comments
-                sqlBuffer.append(line).append("\n");
-                if (line.endsWith(";")) { // Execute statement when semicolon is found
-                  String sql = sqlBuffer.toString().trim();
-                  try {
-                    // since SQL is directly read from the file, there is close to 0 possibility
-                    // of this being injected plus this run via an Admin tool, if attacker can
-                    // fiddle with this that means lot of other things are already compromised.
-                    statement.execute(sql);
-                  } catch (SQLException e) {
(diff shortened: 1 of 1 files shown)


snazy:
The current code is correct. This change leaves the risk of not closing the input stream.

eric-maynard:
What makes you say that, @snazy? Have you found a JDK that behaves that way? 

Oracle documents `BufferedReader#close` as `[Closing] the stream and releases any system resources associated with it.` [here](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#close--) and `InputStreamReader#close` the same way [here](https://docs.oracle.com/javase/8/docs/api/java/io/InputStreamReader.html#close--). I believe this comes from the javadoc of `Reader` itself.

Looking at OpenJDK, the `BufferedReader` will [close](https://github.com/openjdk/panama-foreign/blob/37adee9a04a11f86c67e848a93bc142316622e41/src/java.base/share/classes/java/io/BufferedReader.java#L529) the `InputStreamReader`, and then `InputStreamReader` will [close](https://github.com/openjdk/panama-foreign/blob/37adee9a04a11f86c67e848a93bc142316622e41/src/java.base/share/classes/java/io/InputStreamReader.java#L197) the stream.

snazy:
@eric-maynard you see that a) the `BufferedReader` is created somewhere inside some function calls and b) that the given stream should be closed nontheless, right?

eric-maynard:
> a) the BufferedReader is created somewhere inside some function calls 

What function calls? It's created inside a try-with-resources which is visible in the diff [here](https://github.com/apache/polaris/pull/1982/files#diff-de43b240b5b5e07aba7e89f5515a417cefd908845b85432f3fcc0819911f3e2eR88).

> that the given stream should be closed nontheless, right?

That is what the try-with-resources is for

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -82,45 +82,37 @@ DatabaseType getDatabaseType() {
...
-                    } catch (SQLException e) {
-                      throw new RuntimeException(e);
-                    }
-                    sqlBuffer.setLength(0); // Clear the buffer for the next statement
+    runWithinTransaction(
+        connection -> {
+          try (Statement statement = connection.createStatement();
+              BufferedReader reader =
+                  new BufferedReader(
+                      new InputStreamReader(Objects.requireNonNull(scriptInputStream), UTF_8))) {
```
dimas-b: Regardless of the `.close()` call, I just realised that `runWithinTransaction()` has re-tries... So if this code is called again, the reading of the `scriptInputStream` will lead to invalid data. WDYT?

eric-maynard: Yeah, to fix that I think we need to go back to having two try-with-resources. Still, the extra close call can be removed. Just pushed a fix.

PR #1983, opened 2025-07-01
Fix doc to remove outdated note about fine-grained access control support

cgpoh:
Minor update for the access control doc:

1. Remove the misleading section on privileges can only be granted at catalog level. I've tested the fine-grained access controls and confirmed that privileges can be applied to an individual table in the catalog.

### DIFF
--- a/site/content/in-dev/unreleased/access-control.md
+++ b/site/content/in-dev/unreleased/access-control.md
@@ -103,10 +103,6 @@ This section describes the privileges that are available in the Polaris access c
 roles are granted to principal roles, and principal roles are granted to service principals to specify the operations that service principals can
 perform on objects in Polaris.
 
-> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


eric-maynard (approved):

site/content/in-dev/unreleased/access-control.md
eric-maynard, line 1: There are a few other issues in this doc, such as the fact that a Policy is now a valid securable object, and the note that `If you update the privileges bestowed to a service principal, the updates won't take effect for up to one hour.`. I also think the diagram at the bottom is confusing at best.

However, this change by itself looks good enough to merge as it's just not accurate.

cgpoh: Thanks @eric-maynard for your review. Maybe a separate PR to fix the other issues? Actually, the diagram together with the example help me when trying to understand the RBAC :)

PR #1985, opened 2025-07-01
Add a Spark session builder for the tests

flyrain:


integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+ *     .withTestDefaults()
+ *     .addCatalog("catalog1", CatalogType.ICEBERG, endpoints, token)
+ *     .addCatalog("catalog2", CatalogType.ICEBERG, endpoints, token)
+ *     .createSession();
+ * </pre>
+ */
+public class SparkSessionBuilder {
+
+  public enum CatalogType {
+    ICEBERG("org.apache.iceberg.spark.SparkCatalog"),
```
gh-yzou: i think we can simply have catalogImplClass, which is a string, so the test can choose whatever SparkCatalog they want to sue

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  private SparkSession.Builder applyCatalogConfigurations(SparkSession.Builder builder) {
+    for (CatalogConfig catalog : catalogs) {
+      builder = applySingleCatalogConfig(builder, catalog);
+    }
+    return builder;
+  }
+
+  private SparkSession.Builder applySingleCatalogConfig(
```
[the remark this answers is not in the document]
flyrain: the `s3.region` isn't needed in the tests.

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+      this.implementationClass = implementationClass;
+    }
+
+    public String getImplementationClass() {
+      return implementationClass;
+    }
+  }
+
+  public enum ExtensionType {
+    ICEBERG_ONLY("org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions"),
```
gh-yzou: similarly, let's just take a list of extension class strings, so later when we do hudi test, and possible combination, the test can just pass whatever extension string they want

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+    this.builder = builder;
+  }
+
+  /**
+   * Create a SparkSessionBuilder with common test defaults: local master and disabled UI.
+   *
+   * @return new builder instance with test defaults
+   */
+  public static SparkSessionBuilder withTestDefaults() {
+    return new SparkSessionBuilder(SparkSession.builder()).withLocalMaster().withDisabledUI();
```
gh-yzou: we are always using testDefaults, right? i don't see a case that we need to use none localMaster without disableUI, let's just remove this configurability

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+    return this;
+  }
+
+  public SparkSessionBuilder withDeltaCatalogConfig() {
+    this.includeDeltaCatalogConfig = true;
+    return this;
+  }
+
+  public SparkSessionBuilder withS3MockContainer() {
+    return withS3FileSystem("foo", "bar");
```
gh-yzou: i don't think we need this, just withS3FileSystem seems good enough to me

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+  public SparkSessionBuilder withLocalMaster(int cores) {
+    return master(String.format("local[%d]", cores));
+  }
+
+  public SparkSessionBuilder withLocalMaster() {
+    return withLocalMaster(1);
+  }
+
+  public SparkSessionBuilder withDisabledUI() {
+    return withConfig("spark.ui.showConsoleProgress", "false")
```
gh-yzou: Let's remove withDisabledUI and withLocalMaster, seems not useful for testing, which could be quite confusing when people are using it also

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+    return this;
+  }
+
+  public SparkSessionBuilder withExtensions(ExtensionType extensionType) {
+    this.extensionType = extensionType;
+    return this;
+  }
+
+  public SparkSessionBuilder withDeltaCatalogConfig() {
+    this.includeDeltaCatalogConfig = true;
```
gh-yzou: let's remove the delta config, so we don't have to add hudiConfig or other format config later which could make this utility unncessary complicated

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+            .config(String.format("spark.sql.catalog.%s.type", catalog.catalogName), "rest")
+            .config(
+                String.format("spark.sql.catalog.%s.warehouse", catalog.catalogName),
+                catalog.catalogName)
+            .config(
+                String.format("spark.sql.catalog.%s.scope", catalog.catalogName),
+                "PRINCIPAL_ROLE:ALL");
+
+    // Add endpoint configuration
+    if (catalog.endpoints != null) {
```
gh-yzou: without endpoint configuration, the test probably won't work, let's just add a precondition check that requires endpoints to be not null

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+ * A fluent builder for configuring SparkSession instances with Polaris catalogs.
+ *
+ * <p>Example usage:
+ *
+ * <pre>
+ * SparkSession session = SparkSessionBuilder
+ *     .withTestDefaults()
+ *     .addCatalog("catalog1", CatalogType.ICEBERG, endpoints, token)
+ *     .addCatalog("catalog2", CatalogType.ICEBERG, endpoints, token)
+ *     .createSession();
```
gh-yzou, line 44: might be also useful to have an example about the actual sparkSession build result with in the comment also 
like 
```
.config(
            String.format("spark.sql.catalog.%s", catalogName),
            "org.apache.iceberg.spark.SparkCatalog")
        .config("spark.sql.warehouse.dir", warehouseDir.toString())
        .config(String.format("spark.sql.catalog.%s.type", catalogName), "rest")
        .config(
            String.format("spark.sql.catalog.%s.uri", catalogName),
            endpoints.catalogApiEndpoint().toString())
        .config(String.format("spark.sql.catalog.%s.warehouse", catalogName), catalogName)
        .config(String.format("spark.sql.catalog.%s.scope", catalogName), "PRINCIPAL_ROLE:ALL")
        .config(
            String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
        .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
        .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
        .config(
            String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
        .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
```

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+  public SparkSessionBuilder withDeltaCatalogConfig() {
+    this.includeDeltaCatalogConfig = true;
+    return this;
+  }
+
+  public SparkSessionBuilder withS3MockContainer() {
+    return withS3FileSystem("foo", "bar");
+  }
+
+  public SparkSessionBuilder withS3FileSystem(String accessKey, String secretKey) {
```
gh-yzou: i think we can also take the s3 region as an input to set the following configuration
```
.config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
```

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
```diff
@@ -19,33 +19,19 @@
...
-            String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
-        .config(String.format("spark.sql.catalog.%s.token", catalogName), sparkToken)
-        .config(String.format("spark.sql.catalog.%s.s3.access-key-id", catalogName), "fakekey")
-        .config(
-            String.format("spark.sql.catalog.%s.s3.secret-access-key", catalogName), "fakesecret")
-        .config(String.format("spark.sql.catalog.%s.s3.region", catalogName), "us-west-2");
+  protected SparkSession buildSparkSession() {
+    return SparkSessionBuilder.withTestDefaults()
+        .withS3MockContainer()
+        .withExtensions(SparkSessionBuilder.ExtensionType.ICEBERG_ONLY)
```
gh-yzou: the realmId set seems missing         
```
 String.format("spark.sql.catalog.%s.header.realm", catalogName), endpoints.realmId())
```

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
gh-yzou: actually, instead of withDeltaCatalogConfig, we can probably take a sparkSessionCatalogImplClass in general for setting up of spark_catalog

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+    // Add endpoint configuration
+    if (catalog.endpoints != null) {
+      builder =
+          builder
+              .config(
+                  String.format("spark.sql.catalog.%s.uri", catalog.catalogName),
+                  catalog.endpoints.catalogApiEndpoint().toString())
+              .config(
+                  String.format("spark.sql.catalog.%s.header.realm", catalog.catalogName),
+                  catalog.endpoints.realmId());
```
flyrain: realmId is set here

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkCatalogIcebergIT.java
flyrain: https://github.com/apache/polaris/pull/1985/files#r2178823978

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
flyrain: `spark.sql.catalog.%s.s3.region` is needed only if we config the fileIO like this 
`spark.sql.catalog.catalog-name.io-impl=xxx`, which isn't applicable here.

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
flyrain: Can you elaborate it?

gh-yzou: sorry, i just mean in the comment also add description about how the final effect will look like to help illustrate the usage

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  /** Configuration for a single catalog. */
+  private static class CatalogConfig {
+    final String catalogName;
+    final String catalogType;
+    final PolarisApiEndpoints endpoints;
+    final String token;
+    final List<ConfigPair> catalogSpecificConfigs;
```
gh-yzou: can we have an example here about what are the catalogSpecificConfigs here? is that used anywhere, i didn't see we use i anywhere, maybe let's just remove this

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+
+    CatalogConfig(
+        String catalogName,
+        String catalogType,
```
gh-yzou: i see we input  "org.apache.iceberg.spark.SparkCatalog" as the catalogType, maybe call it catalogImplClass, which might be more illustrative

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+        new CatalogConfig(catalogName, catalogType, endpoints, token, new ArrayList<>()));
+    return this;
+  }
+
+  public SparkSessionBuilder withConfig(String key, String value) {
+    this.additionalConfigs.add(new ConfigPair(key, value));
+    return this;
+  }
+
+  public SparkSession createSession() {
```
gh-yzou: maybe just call this function getOrCreate() to be more precise, since we call getOrCreate for SparkSession

flyrain: yeah, I thought about it extends from `SparkSession.Builder`, but it seems a bit over-engineering to me.

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
flyrain: removed the field

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
flyrain: Add more descriptions as the Java class doc.

integration-tests/src/main/java/org/apache/polaris/service/it/ext/SparkSessionBuilder.java
gh-yzou: Oh, sorry, i don't mean to extend the SparkSession.build function, I was suggesting to rename the function name from createSession() to getOrCreate()

flyrain: Renamed

gh-yzou (approved):

PR #1987, opened 2025-07-02
fix: Remove runtime-immutable db kind configuration from Helm chart

jparkzz:
Through documentation review and testing, I confirmed that the **db-kind** is determined at build time and cannot be modified at runtime. 

To prevent user confusion, this PR removes the db kind configuration option from the Helm chart since it cannot be set during runtime deployment.

This change improves the user experience by eliminating a non-functional configuration parameter that could mislead users into thinking they can change the database kind after deployment.

### DIFF
--- a/helm/polaris/ci/persistence-values.yaml
+++ b/helm/polaris/ci/persistence-values.yaml
@@ -35,6 +35,5 @@ logging:
 persistence:
   type: relational-jdbc
   relationalJdbc:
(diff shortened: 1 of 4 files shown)


j1wonpark:
Thanks @dimas-b

adutra (approved):

singhpk234 (approved):

PR #1765, opened 2025-05-31
[DRAFT] [DO NOT REVIEW] Introduce CachedSupplier for BasePersistence objects

adnanhemani:
I came across an interesting bug yesterday that we need to fix to ensure that tasks can use the BasePersistence object, as they run outside of user call contexts.

What I was trying to do:
1. Create and run a Task which dumps some information to the persistence. In order to do this, I was using the following line of code to get a BasePersistence object: `metaStoreManagerFactory.getOrCreateSessionSupplier(CallContext.getCurrentContext().getRealmContext()).get();`
2. Get the following error message when executing the last `.get()` call: 
```
jakarta.enterprise.context.ContextNotActiveException: RequestScoped context was not active when trying to obtain a bean instance for a client proxy...
```

When digging deeper into why this is happening, I realized that due to the Supplier's lazy-loading at https://github.com/apache/polaris/blob/main/extension/persistence/relational-jdbc/src/main/java/org/apache/polaris/extension/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java#L100-L105, the `.get()` was actually using a RequestScoped realmContext bean given by the previously-ran `TokenBroker` initialization (which is a `RequestScoped` object here: https://github.com/apache/polaris/blob/main/quarkus/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java#L290-L299. Given this is a relatively-new addition, this may be why we haven't seen this bug previously.

As Tasks run asynchronously, likely after the original request was already completed, this error actually makes sense - we should not be able to use a request scoped bean inside of a Task execution. But upon further looking, we do not actually need `realmContext` for anything other than resolving the `realmIdentifier` once during the BasePersistence object initialization - as a result, we can cache the BasePersistence object using a supplier that caches the original result instead of constantly making new objects. This will also solve our issue, as the original request scoped RealmContext bean will not be used again during the Task's call to get a BasePersistence object.

I've added a test case that shows the difference between the OOTB supplier and my ideal way to solve this problem using a CachedSupplier. If there is significant concern that we cannot cache the BasePersistence object, we can materialize the RealmContext object prior to the supplier so that at a minimum the RequestScoped RealmContext object is not being used - but I'm not sure if there's an easy way to test this, given that the MetastoreFactories are Quarkus `ApplicationScoped` objects.

Please note, this is an issue in both EclipseLink and JDBC, as they have almost identical code paths here.

Many thanks to @singhpk234 for being my debugging rubber ducky :)

PR #1988, opened 2025-07-02
Materialize Realm ID for Session Supplier in JDBC

adnanhemani:
As a continuation of the conversation in #1765, I've reduced the scope of the PR to be solved only for JDBC as a quick workaround. Technically, the `MetastoreManagerFactory` does not imply anything about the implementation of the interface, so I take this decision to materialize the RealmIds as an implementation detail of the JDBC integration.

A quick reproduction of the original issue and the proof that this solves that issue can be found [here](https://github.com/apache/polaris/pull/1765#issuecomment-2942620712).

---

#### Background

It was discovered that the Session Supplier maps used in the `MetaStoreManagerFactory` implementations were passing in `RealmContext` objects to the supplier directly and then using the `RealmContext` objects to create `BasePersistence` implementation objects within the supplier. This supplier is cached on a per-realm basis in most `MetaStoreManagerFactory` implementations. `RealmContext` objects are request-scoped beans.

As a result, if any work is being done outside the scope of the request, such as during a Task, any calls to `getOrCreateSessionSupplier` for creating a `BasePersistence` implementation will fail as the `RealmContext` object is no longer available.

This PR will ensure for the `JdbcMetaStoreManagerFactory` that the Realm ID is materialized from the `RealmContext` and used inside the supplier so that the potentially deactivated `RealmContext` object does not need to be used in creating the BasePersistence object. Given that we are caching on a per-realm basis, this should not introduce any unforeseen behavior for the `JdbcMetaStoreManagerFactory` as the Realm ID must match exactly for the same supplier to be returned from the Session Supplier map.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -99,6 +99,8 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
       DatasourceOperations datasourceOperations,
       RealmContext realmContext,
       RootCredentialsSet rootCredentialsSet) {
+    // Materialize realmId so that background tasks that don't have an active
+    // RealmContext (request-scoped bean) can still create a JdbcBasePersistenceImpl
     String realmId = realmContext.getRealmIdentifier();
     sessionSupplierMap.put(
         realmId,


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -100,17 +99,18 @@ private void initializeForRealm(
...
       RootCredentialsSet rootCredentialsSet) {
+    String realmId = realmContext.getRealmIdentifier();
     sessionSupplierMap.put(
-        realmContext.getRealmIdentifier(),
+        realmId,
         () ->
             new JdbcBasePersistenceImpl(
                 datasourceOperations,
-                secretsGenerator(realmContext, rootCredentialsSet),
+                secretsGenerator(realmId, rootCredentialsSet),
```
singhpk234, line 110: Would you mind adding a comment for this ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -100,17 +99,18 @@ private void initializeForRealm(
       DatasourceOperations datasourceOperations,
       RealmContext realmContext,
       RootCredentialsSet rootCredentialsSet) {
+    String realmId = realmContext.getRealmIdentifier();
     sessionSupplierMap.put(
-        realmContext.getRealmIdentifier(),
```
dimas-b, line 104: Sorry, I do not really see any major runtime difference before/after this change. Could you add comments or clarify what the change is supposed to do in the PR description? It has a link to another PR, but the linked page has a rather long discussion thread... It would be nice to describe specifically what this PR fixes, please.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
adnanhemani: Added!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
adnanhemani: I've added more details to the PR description and added a comment as well!

dimas-b:
Thanks for the updated description, @adnanhemani ! Nice find!

dimas-b (approved):

dimas-b:
Just FYI: https://lists.apache.org/thread/0cyrzft2oon28otxlmmhvd5671rd3r3d

dimas-b:
@singhpk234 : Please feel free to merge after you review the updated PR.

Issue #1390, opened 2025-04-17
Concurrent Namespace Update Fails with RuntimeException and 500 Error


BenjaminSSL:
### Describe the bug

When performing concurrent parallel HTTP requests involving **UPDATE** operations on namespace properties, an **RuntimeException** is encountered.

The error message in the Polaris logs indicates a **Concurrent modification of namespace** issue:
```
ERROR [org.apa.pol.ser.exc.IcebergExceptionMapper] [,POLARIS] [,,,] (executor-thread-2) Unhandled exception returning INTERNAL_SERVER_ERROR: java.lang.RuntimeException: Concurrent modification of namespace: f62452d6-39fc-40a8-87a5-c08f6ee65c80
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog.setProperties(IcebergCatalog.java:710)
	at org.apache.iceberg.rest.CatalogHandlers.updateNamespaceProperties(CatalogHandlers.java:193)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalogHandlerWrapper.updateNamespaceProperties(IcebergCatalogHandlerWrapper.java:538)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalogAdapter.lambda$updateProperties$5(IcebergCatalogAdapter.java:268)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalogAdapter.withCatalog(IcebergCatalogAdapter.java:161)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalogAdapter.updateProperties(IcebergCatalogAdapter.java:264)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalogAdapter_ClientProxy.updateProperties(Unknown Source)
	at org.apache.polaris.service.catalog.api.IcebergRestCatalogApi.updateProperties(IcebergRestCatalogApi.java:911)
	at org.apache.polaris.service.catalog.api.IcebergRestCatalogApi_Subclass.updateProperties$$superforward(Unknown Source)
	at org.apache.polaris.service.catalog.api.IcebergRestCatalogApi_Subclass$$function$$23.apply(Unknown Source)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:62)
[...]

PR #1989, opened 2025-07-02
Respond with 409 in case of concurrent Namespace update failures instead of 500

fabio-rizzo-01:
I followed the same pattern used in methods of the same class, like commit(...). This will return a 409 back to the caller instead of a 500 error

Solves (https://github.com/apache/polaris/issues/1390)


fabio-rizzo-01:
> Thank you for the change @fabio-rizzo-01, can you please add UTs for the same ?

Test added

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
```diff
@@ -486,6 +488,31 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
...
+            Mockito.mock(TaskExecutor.class),
+            fileIOFactory,
+            polarisEventListener);
+    catalog.createNamespace(NS);
+    doReturn(new EntityResult(ENTITY_NOT_FOUND, ""))
+        .when(spyMetaStore)
+        .updateEntityPropertiesIfNotChanged(any(), any(), any());
+    assertThrows(
+        CommitFailedException.class, () -> catalog.setProperties(NS, ImmutableMap.of("a", "b")));
+    assertThrows(CommitFailedException.class, () -> catalog.removeProperties(NS, Set.of("a", "b")));
```
singhpk234: CommitFailedException would throw 409 from IcebergException mapper, but is there a way to just test the status code without relying on exception mapper on this case ?

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -756,7 +756,7 @@ public boolean removeProperties(Namespace namespace, Set<String> properties)
             .map(PolarisEntity::new)
             .orElse(null);
     if (returnedEntity == null) {
-      throw new RuntimeException("Concurrent modification of namespace: " + namespace);
+      throw new CommitFailedException("Concurrent modification of namespace: %s", namespace);
```
singhpk234: I am fine with overloading the CommitFailedException as the iceberg exception for that doesn't say specifically anything about the entities for which commit failed.

singhpk234 (commented):
Mostly LGTM ! Have one minor suggestion

cc @dimas-b as we discussed this before !

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
fabio-rizzo-01: That code is in the Exception mapper,  I can add that specific case in the test:  

https://github.com/apache/polaris/blob/eb6b6ad50f33d0ac73c92753063438dcf428c9db/service/common/src/test/java/org/apache/polaris/service/exception/IcebergExceptionMapperTest.java#L73

like this
`  Arguments.of(new CommitFailedException("Error persisting entity"), 409)),`

but I think that at the catalog level we have to check for the exception.

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dimas-b: IMHO, `CommitFailedException` is too generic to be used for reporting conflicts.

I'd prefer to have a Polaris-specific exception class specifically for reporting conflicts and mapping them to 409 at the REST API layer.

fabio-rizzo-01: Ok, so should I add a new exception in polaris core and add the mapping in PolarisExceptionMapper? Any suggestion for the exception name is welcome.

dimas-b: I think having a polaris-specific exception for this case will simplify maintenance and allow fine-tuning response... so please so.

As for the name, I'd propose `org.apache.polaris.core.exceptions.CommitConflictException` extending `PolarisException`.

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -756,7 +756,7 @@ public boolean removeProperties(Namespace namespace, Set<String> properties)
             .map(PolarisEntity::new)
             .orElse(null);
     if (returnedEntity == null) {
-      throw new RuntimeException("Concurrent modification of namespace: " + namespace);
```
dimas-b, line 750: I think returning 409 in this case is preferable to 500.

However, as far as #1390 is concerned, I believe Polaris can handle this case better. The Iceberg REST API defines property changes in terms or "removals" and "updates". Nothing is specified about checks for the previous state of Namespace properties. In that regard a read-then-write-if-not-modified pattern in the current Polaris code is prone to errors even if the property changes are non-conflicting in themselves.

I believe a better fix would be to add retries in case the entity is modified by another process in this middle of this operation.

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dimas-b: Update: please see my other comment.

Sorry, I was not really up-to-date on current code when I commented earlier.

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dimas-b: Correction: I do _not_ think a 409 response is justified in this case from the client's perspective. Per #1390 clients' requests are not conflicting with each other.

fabio-rizzo-01: I can add the retry logic but there could still be a case when everything fails, so what do we want to return in that case?

dimas-b: I'd use a configurable retry timeout and the 408 status code when the timeout expires.

dimas-b: Actually 408 has a different purpose... Maybe 504 (Gateway Timeout) fits this use case.

@fabio-rizzo-01 : Could you start a `dev` email discussion on this for visibility?

fabio-rizzo-01: @dimas-b  email sent

service/common/src/main/java/org/apache/polaris/service/exception/PolarisExceptionMapper.java
```diff
@@ -45,23 +46,19 @@ public class PolarisExceptionMapper implements ExceptionMapper<PolarisException>
...
-    } else if (exception instanceof PolicyMappingAlreadyExistsException) {
-      return Response.Status.CONFLICT;
-    } else if (exception instanceof PolicyInUseException) {
-      return Response.Status.BAD_REQUEST;
-    } else {
-      return Response.Status.INTERNAL_SERVER_ERROR;
-    }
+    return switch (exception) {
+      case AlreadyExistsException alreadyExistsException -> Response.Status.CONFLICT;
+      case CommitConflictException commitConflictException -> Response.Status.REQUEST_TIMEOUT;
```
dimas-b: I believe 408 does not match this use case. Per PRC [9110](https://httpwg.org/specs/rfc9110.html#status.408): `The 408 (Request Timeout) status code indicates that the server did not receive a complete request message within the time that it was prepared to wait.`

fabio-rizzo-01: I think with the custom exception we can return 500 at this point, what do you think?

dimas-b: With the current state of the PR (no retries on persistence-level conflicts) I believe the 409 (Conflict) response code is the most relevant.

However, as discussed on the [dev email thread](https://lists.apache.org/thread/gp4f84ld08z1khrd09klzdzflv7l7fgh), I do not think 409 is the best possible implementation of the Iceberg REST API. I believe Polaris should retry on the server side and respond with 429 (Too Many Requests) if retries do not help.

That said, I understand that implementing retries is a much bigger effort, I'd be ok with 409 for now (as it is incrementally better than 500).

fabio-rizzo-01: ok so for now I'll use 409 and we can have an enhancement for the retries, ok?

dimas-b: sgtm :)

dimas-b (dismissed):
Thanks, @fabio-rizzo-01 ! I'll leave this PR over for a couple of days for other people to review.

eric-maynard (dismissed):

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
adutra: Sorry for chiming in late, and also apologies if I'm missing some context here. 

I am not convinced that we should introduce `CommitConflictException`. I see many usages of `CommitFailedException` in similar situations, for example:

* Conflict when updating a `Catalog`: https://github.com/apache/polaris/blob/de351deb5648d4785639b0150d0c9df2aaab5b98/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java#L927-L937
* Conflict when updating a `Principal`: https://github.com/apache/polaris/blob/de351deb5648d4785639b0150d0c9df2aaab5b98/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java#L1064-L1074

And so on for `PrincipalRole` and `CatalogRole`.

Besides, `CommitFailedException` is already being mapped to 409:

https://github.com/apache/polaris/blob/a303a1d70df3733f59914bcff62d3fe1a0fc3edc/service/common/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java#L177

In short, I don't mind introducing this new type, but unless I'm missing something obvious, I see little value in it.

dimas-b: My point against `CommitFailedException` is that it comes from Iceberg API jar and as such is not related to the Iceberg REST API directly. Moreover, Polaris needs to have flexibility in producing REST responses, which is not provided for by `CommitFailedException` in general.

I'd personally prefer to gradually migrate all REST error handling to Polaris exceptions in cases where the error is triggered by Polaris code.

I think it's fine to have a mix of exceptions during the transition period.

adutra: Ok then... let's not forget to do the migration later, it would be nice to have a consistent handling of this kind of conflicts in Polaris.

dimas-b: #2168

fabio-rizzo-01:
good to be merged? I'll create a new task for the retry logic @dimas-b

Issue #1929, opened 2025-06-24
Passing ClientId as parameter while creating principal doesn't work


fivetran-arunsuri:
### Describe the bug

Hey Team,
I was trying to verify some of Polaris management API's. I found out that even though the documentation says we can pass clientId as param while creating the principal, it doesn't seems to be working fine

### To Reproduce

1. Start the server and bootstrap with root credentials
2. create the token
3. Hit the http://localhost:8181/api/management/v1/principals API for generating the principal with boday as below
{
    "name": "principal-1",
    "clientId": "arman"
}
 it generates the principal as 
 {
    "principal": {
        "name": "principal-3",
        "clientId": "9cc-blah-blah",
        "properties": {},
        "createTimestamp": 1750756252079,
        "lastUpdateTimestamp": 1750756252079,
        "entityVersion": 1
    },
    "credentials": {
        "clientId": "9cc-bla-bha",
        "clientSecret": "blah-blah"
    }
}

### Actual Behavior

 it generates the principal as 
 {
    "principal": {
        "name": "principal-3",
        "clientId": "9cc-blah-blah",
        "properties": {},
        "createTimestamp": 1750756252079,
        "lastUpdateTimestamp": 1750756252079,
        "entityVersion": 1
    },
    "credentials": {
        "clientId": "9cc-bla-bha",
        "clientSecret": "blah-blah"
    }
}

### Expected Behavior

It should ideally preserve the clientId we are passing from the body and only generate the clientSecret

### Additional context

Also, Why are we not allowing users to set their own clientId and clientSecret while creating the principal?
I am looking for a solution like register principal with older credentials preserved? Can we do that?

eric-maynard:
The spec notes clientId is output-only here:
```yaml
        clientId:
          type: string
          description: The output-only OAuth clientId associated with this principal if applicable
```

>  I found out that even though the documentation says we can pass clientId as param while creating the principal

Where is this in the docs? Sounds like we may need to fix that -- thanks for finding this.

fivetran-arunsuri:
@eric-maynard The swagger Request body for the given API looks like this, could be auto generated but we need to correct that right?

<img width="836" alt="Image" src="https://github.com/user-attachments/assets/2efd9d03-eee2-43af-97c4-0459e0ceff8d" />

Also regarding my second question, I would like to know the feasibility of the following
Why are we not allowing users to set their own clientId and clientSecret while creating the principal?
[...]

PR #1992, opened 2025-07-02
Improve createPrincipal example in API docs

eric-maynard:
In #1929 it was pointed out that the example in the Polaris docs suggests that users can provide a client ID during principal creation:

image

This PR attempts to fix this by adding an explicit example to the spec.


eric-maynard:
In #1929 it was pointed out that the example in the Polaris docs suggests that users can provide a client ID during principal creation:

. . .


This PR attempts to fix this by adding an explicit example to the spec.

PR #1995, opened 2025-07-02
Update Makefile for python client with auto setup

MonkeyCanCode:
The current `Makefile` and `README.md` is asking users to setup needed python environment (either virtual or the one installed on the OS) then installed the needed dependencies. I think it makes more senses to automate it and use a virtual env instead to avoid change an end-users' OS python (as if they didn't setup virtual env, the package will needed to be installed on the OS one). 

This PR does the following:
1. updated the existed `Makefile` with venv setup
2. update doc to remove ref with manual dependency setup steps
3. update pre-commit-hook to run against `cli/*` as well. The current one only ran against `integration_tests/*`
4. `mypy` doesn't appear to be happy with `cli/*`. Add a files regex for now to skip this check (so existed on with `integration_tests/*` will still get run
5. format the code with `make lint` (2 files modified)
6. add helper for Makefile
7. set constant version for poetry

Here is the sample output:
```
➜ make lint
Installing Poetry and project dependencies into .venv...
# Ensure pip is up-to-date within the venv
.venv/bin/pip install --upgrade pip
Requirement already satisfied: pip in ./.venv/lib/python3.13/site-packages (25.1.1)
# Install poetry if not already present
# Install needed dependencies using poetry
.venv/bin/poetry install --all-extras
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: polaris (1.0.0)
Poetry and dependencies installed.
.venv/bin/poetry run pre-commit run --files integration_tests/* cli/*
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
debug statements (python)................................................Passed
ruff check...............................................................Passed
ruff format..............................................................Passed
mypy.....................................................................Passed
```

Here is the sample helper from Makefile:
```
➜  make

Usage:
  make <target>
  help                            Display this help.
  version                         Print version information.
  regenerate-client               Regenerate the client code
  test-client                     Run client tests
  test-integration                Run integration tests
  lint                            Run linting checks
  clean                           Cleanup
```

### DIFF
--- a/client/python/README.md
+++ b/client/python/README.md
@@ -24,7 +24,7 @@ The Apache Polaris Python package provides a client for interacting with the Apa
 
 ### Prerequisites
 - Python 3.9 or later
-- poetry >= 2.0
+- poetry >= 2.1
 
 ### Installation
 First we need to generate the OpenAPI client code from the OpenAPI specification.
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -42,7 +42,7 @@ homepage = "https://polaris.apache.org/"
 repository = "https://github.com/apache/polaris/"
 
 [tool.poetry]
-requires-poetry = ">=2.1"
+requires-poetry = "==2.1.3"
 packages = [{ include = "polaris" }]
 
 [tool.poetry.group.test.dependencies]


client/python/Makefile
```diff
@@ -30,6 +58,22 @@ test-integration:
 	@echo "Polaris is healthy. Starting integration tests..."
 	poetry run pytest integration_tests/ ${PYTEST_ARGS}
 
+# Target to run linting checks
+lint: setup-env
```
HonahX: Thanks for adding the setup-env to make the env more isolated and automated. Could you please also update the workflow file to remove the poetry installation step?
https://github.com/apache/polaris/blob/d962c64260ef7f2a09bd9ab802ce8a01df90ee02/.github/workflows/python-client.yml#L61-L63

This could help make the local test env and CI env consistent. We may need another `test` target to run all the auto-generated unit test.

client/python/Makefile
```diff
@@ -15,10 +15,38 @@
...
+setup-env: $(VENV_DIR) install-poetry-deps
+
+# Target to install poetry and project dependencies
+install-poetry-deps: # This target doesn't *create* $(VENV_DIR), but *uses* it.
+	@echo "Installing Poetry and project dependencies into $(VENV_DIR)..."
+	# Ensure pip is up-to-date within the venv
+	$(VENV_DIR)/bin/pip install --upgrade pip
+	# Install poetry if not already present
+	@if [ ! -f "$(VENV_DIR)/bin/poetry" ]; then \
+		$(VENV_DIR)/bin/pip install poetry; \
```
HonahX: How about pinning the poetry version to make it consistent with
https://github.com/apache/polaris/blob/d962c64260ef7f2a09bd9ab802ce8a01df90ee02/regtests/requirements.txt#L20
It does look weird that client source code refer to a dep version in regtest not vice versa, but I can improve this in follow-ups

MonkeyCanCode:
> @MonkeyCanCode Thanks for improving the Makefile! Overall LGTM
> 
> There are some discussion around doing the code check/format in a different way: #1986. Since the current one is an improvement of the existing solution, I suggest we seek to merge this first and change later based on the result of #1986

Exactly, it is a lot harder to do without this piece first (thinking about using native shell and markfile route instead of a third party plugin which may not be supported after 1-2 years).

client/python/Makefile
MonkeyCanCode: I think that is a good idea. Let me fix this later tonight.

client/python/Makefile
MonkeyCanCode: Good call. Let me update this one later tonight.

MonkeyCanCode:
> > @MonkeyCanCode Thanks for improving the Makefile! Overall LGTM
> > There are some discussion around doing the code check/format in a different way: #1986. Since the current one is an improvement of the existing solution, I suggest we seek to merge this first and change later based on the result of #1986
> 
> Exactly, it is a lot harder to do without this piece first (thinking about using native shell and markfile route instead of a third party plugin which may not be supported after 1-2 years).

@HonahX please take another look when you get a chance. i added a helper for Makefile as well (update the PR summary to reflect the same):
```
➜  make

Usage:
  make <target>
  help                            Display this help.
  version                         Print version information.
  regenerate-client               Regenerate the client code
  test-client                     Run client tests
  test-integration                Run integration tests
  lint                            Run linting checks
  clean                           Cleanup
  ```

File: client/python/Makefile [hunk 3]
```diff
@@ -15,10 +15,65 @@
...
+# Configures the shell for recipes to use bash, enabling bash commands and ensuring
+# that recipes exit on any command failure (including within pipes).
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+
+# Version information
+VERSION ?= $(shell cat pyproject.toml | grep version | sed 's/version *= *"\(.*\)"/\1/')
+BUILD_DATE := $(shell date -u +"%Y-%m-%dT%H:%M:%S%:z")
+GIT_COMMIT := $(shell git rev-parse HEAD)
+POETRY_VERSION := $(shell cat pyproject.toml | grep requires-poetry | sed 's/requires-poetry *= *"\(.*\)"/\1/')
```
HonahX, line 29: The poetry version in pyproject.toml seems not having an upper bound
```
requires-poetry = ">=2.1"
```
I think in the installation script we might want to pin the exact version to avoid CI failures due to breaking changes in automatic version update.

HonahX (approved):
LGTM! Just have one comment about the poetry version in installation script

Thanks for improving this!

client/python/Makefile
MonkeyCanCode: So I did thought about this, there are couple reasons on why I did the above:

There are 3 places where we are specify poetry (README.md within this direction...which we said >2.0 will work), regretest (which we defined a specific version, in this case 2.1.3), and  pyproject.toml where we said anything >= 2.1 should work. I had updated all 3 pieces to be the same version. Please take another look when you get a chance.

HonahX (approved):

PR #1997, opened 2025-07-03
Add regression test coverage for Spark Client with package conf

gh-yzou:
Spark can use the Spark Client with --jars and --packages use case, the regression test today only covers the --jars use case. 
This PR adds regression test with the --packages use case.

### DIFF
--- a/plugins/spark/v3.5/regtests/setup.sh
+++ b/plugins/spark/v3.5/regtests/setup.sh
@@ -39,6 +39,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 SPARK_VERSION=3.5.5
 SCALA_VERSION=2.12
 POLARIS_CLIENT_JAR=""
+POLARIS_VERSION=""
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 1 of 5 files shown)


plugins/spark/v3.5/regtests/setup.sh
```diff
@@ -129,11 +135,27 @@ else
...
+else
+  cat << EOF >> ${SPARK_CONF}
 # POLARIS Spark client test conf
 spark.jars $POLARIS_CLIENT_JAR
 spark.jars.packages io.delta:delta-spark_${SCALA_VERSION}:3.2.1
+EOF
+fi
+
+# spark.jars $POLARIS_CLIENT_JAR
+# spark.jars.packages io.delta:delta-spark_${SCALA_VERSION}:3.2.1
```
flyrain: Nit: what are the comments for?

flyrain (approved):
Thanks @gh-yzou for increasing the test coverage!

plugins/spark/v3.5/regtests/setup.sh
gh-yzou: oh, that is just a missed cleanup, removed. Thanks!

flyrain (approved):

PR #1998, opened 2025-07-03
Update CatalogEntity::Builder to set default CatalogType as INTERNAL

poojanilangekar:
Encountered the issue while adding additional validations to `ExternalCatalog`. The `CatalogEntity::Builder` checks if the `Catalog::type` is set to `INTERNAL`, if not it defaults to `EXTERNAL`. However this is the opposite of the behavior defined in polaris-management-service.yml where the default is set to `INTERNAL`. This change only affects tests because in other cases the catalog entity is generated from the REST request.  

Testing:
Updated CatalogEntityTest to ensure that the default is set to `INTERNAL`.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
@@ -106,17 +106,8 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
         CatalogProperties.builder(propertiesMap.get(DEFAULT_BASE_LOCATION_KEY))
             .putAll(propertiesMap)
             .build();
-    return catalogType == Catalog.TypeEnum.INTERNAL
-        ? PolarisCatalog.builder()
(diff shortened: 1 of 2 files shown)


poojanilangekar:
CC: @dimas-b

eric-maynard (approved):

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -106,25 +106,25 @@ public Catalog asCatalog() {
         CatalogProperties.builder(propertiesMap.get(DEFAULT_BASE_LOCATION_KEY))
             .putAll(propertiesMap)
             .build();
-    return catalogType == Catalog.TypeEnum.INTERNAL
-        ? PolarisCatalog.builder()
-            .setType(Catalog.TypeEnum.INTERNAL)
+    return catalogType == Catalog.TypeEnum.EXTERNAL
```
dimas-b, line 109: nit: would it not be easier to fix line 103 instead?

dimas-b (approved):

PR #1999, opened 2025-07-04
Ignore regenerate.sh on README.md

MonkeyCanCode:
Currently if an end-user try to rerun `regenerate.sh` (as part of polaris repair or OpenAPI code regen, it will keep adding new headers to the existed README.nd. As this file is not part of the template, update the list to exclude this file.

### DIFF
--- a/client/templates/regenerate.sh
+++ b/client/templates/regenerate.sh
@@ -18,7 +18,7 @@
 # under the License.
 #
 
-set -e 
+set -e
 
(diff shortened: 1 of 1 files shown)


eric-maynard:
Which README? The main one for the repo?

MonkeyCanCode:
> Which README? The main one for the repo?

This is a relative path. So it is the one under python dir. 

You can run repair for 1-2 times then do a git diff, you will see this file got multiple headers.

eric-maynard (approved):

eric-maynard:
Oh, yeah, strange that this never got multiple headers before. Good catch!

PR #2000, opened 2025-07-04
feat(auth): Ability to override active roles provider per realm

adutra:
Most of the authentication configuration options are overridable per realm. That wasn't the case of `polaris.active-roles-provider.type` until now. This PR simply enables per-realm overrides for this property.


dimas-b (dismissed):

snazy (approved):

runtime/defaults/src/main/resources/application.properties
```diff
@@ -145,6 +145,8 @@ polaris.rate-limiter.token-bucket.requests-per-second=9999
 polaris.rate-limiter.token-bucket.window=PT10S
 
 polaris.active-roles-provider.type=default
+# Per-realm overrides:
+# polaris.realm1.active-roles-provider.type=custom
```
flyrain: How does `polaris.realm1.active-roles-provider.type` populate into a `QuarkusAuthenticationConfiguration` bean given that the prefix is `polaris.authentication`?
```
@ConfigMapping(prefix = "polaris.authentication")
```

Can we add a test for this?

runtime/defaults/src/main/resources/application.properties
```diff
@@ -145,6 +145,8 @@ polaris.rate-limiter.token-bucket.requests-per-second=9999
 polaris.rate-limiter.token-bucket.window=PT10S
 
 polaris.active-roles-provider.type=default
```
flyrain, line 148: Can we also add an item in this page? https://polaris.apache.org/in-dev/unreleased/configuration/.

runtime/defaults/src/main/resources/application.properties
dimas-b: The mapping is handled by Quarkus per `@WithUnnamedKey(DEFAULT_REALM_KEY)` in `QuarkusAuthenticationConfiguration`

flyrain: Trying to understand, so `prefix = "polaris.authentication"` doesn't matter in this case, and `QuarkusAuthenticationConfiguration` will pick up the configuration based on the method name `activeRolesProvider()` in `QuarkusAuthenticationRealmConfiguration`? How does a key like `polaris.realm1.active-roles-provider.type` map to a method named `activeRolesProvider()`?

dimas-b: The config prefix matters. It is associated with the `QuarkusAuthenticationConfiguration` config class.

A config entry like `polaris.active-roles-provider.type=ABC` will map to the `DEFAULT_REALM_KEY` in the [realms()](https://github.com/apache/polaris/blob/ab228afa4d975faabb7aaf1e8abb0804f5b9d353/runtime/service/src/main/java/org/apache/polaris/service/quarkus/auth/QuarkusAuthenticationConfiguration.java#L36) map.

A config entry like `polaris.realm1.active-roles-provider.type=DEF` will map to the `realm1` key in the same map.

Look up is handled by [AuthenticationConfiguration](https://github.com/apache/polaris/blob/e1c0a1ccdad4e4e432c026cdc3ae30dcb8112da1/service/common/src/main/java/org/apache/polaris/service/auth/AuthenticationConfiguration.java#L34)

runtime/defaults/src/main/resources/application.properties
dimas-b: Yeah, good point, but I'd say it might be best to do it as a separate PR so as not to bloat the scope of this (very narrow) change.

runtime/defaults/src/main/resources/application.properties
flyrain: Is it OK to not follow the prefix `polaris.authentication` in this setting `polaris.active-roles-provider.type` ?

dimas-b: Sorry , I made a typo in my examples and missed the prefix differences.

flyrain: I think the Quarkus config has the name convention to convert something like `active-roles-provider.type` to  `activeRolesProvider()`, which is fine. I'm still confused by the prefix.

dimas-b: I guess the right new property names should be: `polaris.authentication.active-roles-provider.type=ABC` and `polaris.authentication.realm1.active-roles-provider.type=DEF`

flyrain: That makes sense to me. We will have to follow the deprecating process for the old one. And the example isn't  correct in that case. Can we add a simple test for that to avoid any regression?

adutra: My apologies, I forgot to change the default `application.properties` file. And I also agree to follow the deprecation process, I will add that.

Regarding a test: that's mostly Quarkus configuration stuff, so there is little to test. But I will try to come up with something.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -341,9 +342,12 @@ public QuarkusAuthenticationRealmConfiguration realmAuthConfig(
 
   @Produces
   public ActiveRolesProvider activeRolesProvider(
-      @ConfigProperty(name = "polaris.active-roles-provider.type") String activeRolesProviderType,
+      @ConfigProperty(name = "polaris.active-roles-provider.type")
+          Optional<String> deprecatedActiveRolesProviderType,
+      QuarkusAuthenticationRealmConfiguration config,
       @Any Instance<ActiveRolesProvider> activeRolesProviders) {
-    return activeRolesProviders.select(Identifier.Literal.of(activeRolesProviderType)).get();
+    String type = deprecatedActiveRolesProviderType.orElse(config.activeRolesProvider().type());
```
dimas-b: nit: it might be possible to do this fallback in `application.properties`, something like (not tested):

```
polaris.active-roles-provider.type=default
polaris.authentication.active-roles-provider.type=${polaris.active-roles-provider.type}
```

This way feels a bit smoother to me, but it's not critical.

runtime/defaults/src/main/resources/application.properties
adutra: Also, we are preparing a full documentation page on authentication, stay tuned :-)

runtime/defaults/src/main/resources/application.properties
adutra: Added some tests FYI, although the configuration test is imho a bit dummy because it tests mostly SmallRye config mappings.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
adutra: Good idea, it seems to work.

flyrain (approved):
Thanks @adutra !

runtime/defaults/src/main/resources/application.properties
dimas-b: Nice catch on property name, @flyrain ! :sweat_smile:

PR #2001, opened 2025-07-05
Fix helm doc

MonkeyCanCode:
This PR fixed couple issues in the helm doc
1. Earlier maintainers was removed from `README.md` instead of `README.md.gotmpl`. Thus, if anyone reran the helm-doc generation, it will get add back (ref: https://github.com/apache/polaris/pull/1962)
2. Remove reference for eclipse link as the code no longer support those (both doc and helm chart) and add steps needed when working with relational JDBC (e.g. earlier we will need to dump persistence.xml to disk then use the admin tool to bootstrap...the updated steps will set those in env then run the command with right binary path as the previous path is no longer valid)
3. Update run.sh to remove invalid ref with eclipse link and update it with the proper build command
4. Remove `--debug` option from helm (it will print out manifest but doesn't seem likes a production graded option to include in the doc compared to any other charts we can find online)


MonkeyCanCode:
@flyrain see above for the broken doc again

helm/polaris/README.md.gotmpl
```diff
@@ -89,22 +71,7 @@ eval $(minikube -p minikube docker-env)
...
-
-#### Common setup
-
-Create the target namespace:
-
-```bash
-kubectl create namespace polaris
-```
-
-Create all the required resources in the `polaris` namespace. This usually includes a Postgres
```
adutra, line 104: Why remove this paragraph? I think it's still useful. We could instead just remove the reference to persistence.xml and replace it with a reference to the jdbc-persistence and the storage secrets.

helm/polaris/README.md.gotmpl
```diff
@@ -113,67 +80,59 @@ Below are two sample deployment models for installing the chart: one with a non-
...
+  polaris helm/polaris --create-namespace polaris
 ```
 
 #### Persistent backend
 
 > [!WARNING]
 > The Postgres deployment set up in the fixtures directory is intended for testing purposes only and is not suitable for production use. For production deployments, use a managed Postgres service or a properly configured and secured Postgres instance.
 
-Install the chart with a persistent backend. From Polaris repo root:
+Install the dependencies in fixtures directory. From Polaris repo root:
```
adutra: Hmm I don't like much the idea of telling users to install the fixtures in the `helm/polaris/ci/fixtures/` directory. These fixtures are for tests only and they are mostly bogus.

helm/polaris/README.md.gotmpl
```diff
@@ -198,27 +157,18 @@ The following tools are required to run the tests:
...
-
-The `helm/polaris/ci` contains a number of values files that will be used to install the chart with
-different configurations.
+The `helm/polaris/ci` contains a number of values files that will be used to install the chart with different configurations.
 
 ### Running the unit tests
 
-Helm unit tests do not require a Kubernetes cluster. To run the unit tests, execute Helm Unit from
-the Polaris repo root:
+Helm unit tests do not require a Kubernetes cluster. To run the unit tests, execute Helm Unit from the Polaris repo root:
```
adutra: nit: Unnecessary change?

helm/polaris/README.md.gotmpl
```diff
@@ -232,13 +182,11 @@ ct lint --charts helm/polaris
 
 ### Running the integration tests
 
-Integration tests require a Kubernetes cluster. See installation instructions above for setting up
-a local cluster.
+Integration tests require a Kubernetes cluster. See installation instructions above for setting upa local cluster.
```
adutra: nit: unnecessary change? Also, there is a typo:

```suggestion
Integration tests require a Kubernetes cluster. See installation instructions above for setting up a local cluster.
```

helm/polaris/templates/configmap.yaml
```diff
@@ -6,9 +6,9 @@
   to you under the Apache License, Version 2.0 (the
   "License"); you may not use this file except in compliance
   with the License.  You may obtain a copy of the License at
- 
+
```
adutra: Not a big deal, but you should strive not to introduce such changes since it's not related to the PR.

run.sh
```diff
@@ -53,9 +48,11 @@ sh ./kind-registry.sh
 # Build and deploy the server image
 echo "Building polaris image..."
 ./gradlew \
-  :polaris-server:build \
+  :polaris-server:assemble \
   :polaris-server:quarkusAppPartsBuild --rerun \
-  $ECLIPSE_LINK_DEPS \
+  :polaris-admin:assemble \
+  :polaris-admin:quarkusAppPartsBuild --rerun \
+  -Dquarkus.container-image.tag=postgres-latest \
```
adutra, line 55: I think we don't need the special tag anymore. The default image is built with the Postgres JDBC driver afaict.

adutra (commented):
Thanks @MonkeyCanCode for tackling this, I think it's good to see EclipseLink removed from the Helm chart. I'm also OK with the other changes. I just had some concerns about the CI fixtures.

helm/polaris/README.md.gotmpl
MonkeyCanCode: So instead of running one more command, I put `--create-namespace polaris` in the helm command instead. But if this route is more preferred, I can revert it back.

helm/polaris/README.md.gotmpl
MonkeyCanCode: I think that is the current behavior of the doc and we have WARNING banner right after `Persistent backend` already for this. Any preferred changes you have?

helm/polaris/README.md.gotmpl
MonkeyCanCode: Yes. I tried to keep them become one line as oppose to multi-line (we defined max characters per lines?) then format properly for those?

helm/polaris/README.md.gotmpl
MonkeyCanCode: Fixed. Same reason as above. Change comment to one line instead of multi-lines (we no max characters per line defined.)

helm/polaris/templates/configmap.yaml
MonkeyCanCode: Yes. Those were introduced with editor I am using which auto remove empty space for the line. Let me revert it back.

run.sh
MonkeyCanCode: Without `-Dquarkus.container-image.tag=postgres-latest`, it will introduce breaking changes in other getting start examples as they are all refers to `postgres-latest`. If this not preferred, we can remove this and update docs that refs to this. Without using this, we will have `latest` and `<version>` tags.

flyrain (approved):
Thanks a lot for doing this, @MonkeyCanCode !

helm/polaris/README.md.gotmpl
adutra: The thing is, we don't have a rule for that (at least, not that I know of). So going back and forth from single-line to multi-line without changing the actual words seems a rather bad practice, because it adds visual clutter for reviewers.

MonkeyCanCode: That is fair. Let me revert those changes.

helm/polaris/README.md.gotmpl
adutra: Well, all the files in `ci/fixtures/` are completely useless for a production deployment. I'd prefer to just tell users to install whatever resources they need (secrets, service accounts, database, etc.) before installing the Helm chart. But telling them to install `ci/fixtures/` is imho a nonsense.

MonkeyCanCode: K. Let me change the wording and needed resources for persistent backend if they would like to use it. Then move this section to the bottom as test resources.

helm/polaris/README.md.gotmpl
adutra: OK but this paragraph is imho important because it tells users that they _must_ install the prerequisites now before proceeding, otherwise, the Helm chart installation may fail.

That's also why I think that `--create-namespace` is generally useful only in small deployments. As soon as lots of dependencies must be installed (service accounts, secrets, the database, etc), it becomes impractical to create the namespace during the Helm chart installation.

MonkeyCanCode:
@adutra thanks for the review. I had addressed the changes based on feedback. Please take another look when you get a chance.

flyrain:
Not a blocker: Hi @MonkeyCanCode and @adutra , can we move the helm/polaris/README.md to the site so that people can visit it from Polaris website?

MonkeyCanCode:
> Not a blocker: Hi @MonkeyCanCode and @adutra , can we move the helm/polaris/README.md to the site so that people can visit it from Polaris website?  

Sure thing. I can take care of it later tonight.

MonkeyCanCode:
> > Not a blocker: Hi @MonkeyCanCode and @adutra , can we move the helm/polaris/README.md to the site so that people can visit it from Polaris website?
> 
> Sure thing. I can take care of it later tonight.

Here is the sample PR for this: https://github.com/apache/polaris/pull/2006

helm/polaris/README.md.gotmpl
```diff
@@ -96,15 +81,14 @@ The below instructions assume a local Kubernetes cluster is running and Helm is
...
 
 Create the target namespace:
-
 ```bash
 kubectl create namespace polaris
 ```
 
 Create all the required resources in the `polaris` namespace. This usually includes a Postgres
-database and a Kubernetes Secret for the `persistence.xml` file. The Polaris chart does not create
+database and a Kubernetes Secret for Polaris service certification files. The Polaris chart does not create
```
adutra:
```suggestion
database, Kubernetes secrets, and service accounts. The Polaris chart does not create
```

helm/polaris/README.md.gotmpl
```diff
@@ -96,15 +81,14 @@ The below instructions assume a local Kubernetes cluster is running and Helm is
...
 these resources automatically, as they are not required for all Polaris deployments. The chart will
-fail if these resources are not created beforehand.
+fail if these resources are not created beforehand. You can find a reference for these resources in the `Prerequisites` section within `Development & Testing`.
```
adutra:
```suggestion
fail if these resources are not created beforehand. You can find some examples in the `helm/polaris/ci/fixtures` directory, but beware that these are primarily intended for tests.
```

PR #2002, opened 2025-07-05
Helm key grouping and test cases

MonkeyCanCode:
This PR contains the following changes:
1. Group public/private key under `rsaKeyPair` and group symmetric key under `symmetricKey` (it seems to be confusing when we provided all 3 without grouping)
2. Add test case for symmetric key (and updated existed test case with rsa key but provide values for both options...the helper function will correctly populated the right config)
3. Minor change for the comments in values.yaml
4. Re-generate README.md with maintainer section removed (https://github.com/apache/polaris/pull/1962)

### DIFF
--- a/helm/polaris/templates/_helpers.tpl
+++ b/helm/polaris/templates/_helpers.tpl
@@ -182,19 +182,9 @@ Prints the config volume definition for deployments and jobs.
           items:
           {{- if eq .Values.authentication.tokenBroker.type "rsa-key-pair" }}
             {{- /* Backward compatibility for publicKey: new takes precedence */ -}}
-            {{- $publicKey := "" }}
-            {{- if .Values.authentication.tokenBroker.secret.rsaKeyPair.publicKey }}
-              {{- $publicKey = .Values.authentication.tokenBroker.secret.rsaKeyPair.publicKey }}
-            {{- else if .Values.authentication.tokenBroker.secret.publicKey }}
-              {{- $publicKey = .Values.authentication.tokenBroker.secret.publicKey }}
-            {{- end }}
+            {{- $publicKey := coalesce .Values.authentication.tokenBroker.secret.rsaKeyPair.publicKey .Values.authentication.tokenBroker.secret.publicKey }}
             {{- /* Backward compatibility for privateKey: new takes precedence */ -}}
-            {{- $privateKey := "" }}
-            {{- if .Values.authentication.tokenBroker.secret.rsaKeyPair.privateKey }}
-              {{- $privateKey = .Values.authentication.tokenBroker.secret.rsaKeyPair.privateKey }}
-            {{- else if .Values.authentication.tokenBroker.secret.privateKey }}
-              {{- $privateKey = .Values.authentication.tokenBroker.secret.privateKey -}}
-            {{- end }}
+            {{- $privateKey := coalesce .Values.authentication.tokenBroker.secret.rsaKeyPair.privateKey .Values.authentication.tokenBroker.secret.privateKey }}
             - key: {{ tpl $publicKey . }}
               path: public.pem
             - key: {{ tpl $privateKey . }}
@@ -202,12 +192,7 @@ Prints the config volume definition for deployments and jobs.
           {{- end }}
           {{- if eq .Values.authentication.tokenBroker.type "symmetric-key" }}
             {{- /* Backward compatibility for symmetricKey: new takes precedence */ -}}
-            {{- $secretKey := "" }}
-            {{- if .Values.authentication.tokenBroker.secret.symmetricKey.secretKey }}
-              {{- $secretKey = .Values.authentication.tokenBroker.secret.symmetricKey.secretKey }}
-            {{- else if .Values.authentication.tokenBroker.secret.secretKey }}
-              {{- $secretKey = .Values.authentication.tokenBroker.secret.secretKey }}
-            {{- end }}
+            {{- $secretKey := coalesce .Values.authentication.tokenBroker.secret.symmetricKey.secretKey .Values.authentication.tokenBroker.secret.secretKey }}
             - key: {{ tpl $secretKey . }}
               path: symmetric.key
           {{- end }}


helm/polaris/values.yaml
```diff
@@ -583,18 +583,27 @@ authentication:
...
+    # -- Maximum token generation duration (e.g., PT1H for 1 hour).
     maxTokenGeneration: PT1H
     # -- The secret name to pull the public and private keys, or the symmetric key secret from.
     secret:
       # -- The name of the secret to pull the keys from. If not provided, a key pair will be generated.
       # This is not recommended for production.
       name: ~
-      # -- The public key file to use for RSA key pair token broker. Only required when using rsa-key-pair.
+
       publicKey: public.pem
```
adutra, line 583:
```suggestion
```

helm/polaris/values.yaml
```diff
@@ -583,18 +583,27 @@ authentication:
...
-      # -- The private key file to use for RSA key pair token broker. Only required when using rsa-key-pair.
       privateKey: private.pem
```
adutra, line 586:
```suggestion
```

helm/polaris/values.yaml
```diff
@@ -583,18 +583,27 @@ authentication:
...
-      # -- The symmetric key file to use for symmetric key token broker. Only required when using symmetric-key.
       secretKey: secret
```
```suggestion
```

adutra (commented):
Thanks @MonkeyCanCode for this change. 

I'm generally OK with it, but this is a breaking  change in the Helm chart, and since it will be released after 1.0.0, I wonder if we shouldn't proceed with some caution here.

I see a few options:

* Allow the breaking change and mention it in the changelog file.
* Keep the old fields around and deprecate them for removal.

Wdyt?

MonkeyCanCode:
> Thanks @MonkeyCanCode for this change.
> 
> I'm generally OK with it, but this is a breaking change in the Helm chart, and since it will be released after 1.0.0, I wonder if we shouldn't proceed with some caution here.
> 
> I see a few options:
> 
> * Allow the breaking change and mention it in the changelog file.
> * Keep the old fields around and deprecate them for removal.
> 
> Wdyt?

I think that is fair (thought as we tried to get to 1.0.0 release, thus introduced this breaking change before that happened). But think that is reasonable to keep it backward compatible. Let me update this helm chart to do that.

MonkeyCanCode:
> > Thanks @MonkeyCanCode for this change.
> > I'm generally OK with it, but this is a breaking change in the Helm chart, and since it will be released after 1.0.0, I wonder if we shouldn't proceed with some caution here.
> > I see a few options:
> > 
> > * Allow the breaking change and mention it in the changelog file.
> > * Keep the old fields around and deprecate them for removal.
> > 
> > Wdyt?
> 
> I think that is fair (thought as we tried to get to 1.0.0 release, thus introduced this breaking change before that happened). But think that is reasonable to keep it backward compatible. Let me update this helm chart to do that.

@adutra Do you kind take another look when you get a chance. Thanks.

helm/polaris/templates/_helpers.tpl
```diff
@@ -181,13 +181,34 @@ Prints the config volume definition for deployments and jobs.
...
           items:
           {{- if eq .Values.authentication.tokenBroker.type "rsa-key-pair" }}
-            - key: {{ tpl .Values.authentication.tokenBroker.secret.publicKey . }}
+            {{- /* Backward compatibility for publicKey: new takes precedence */ -}}
+            {{- $publicKey := "" }}
+            {{- if .Values.authentication.tokenBroker.secret.rsaKeyPair.publicKey }}
+              {{- $publicKey = .Values.authentication.tokenBroker.secret.rsaKeyPair.publicKey }}
+            {{- else if .Values.authentication.tokenBroker.secret.publicKey }}
+              {{- $publicKey = .Values.authentication.tokenBroker.secret.publicKey }}
+            {{- end }}
```
adutra:
```suggestion
            {{- $publicKey := coalesce .Values.authentication.tokenBroker.secret.rsaKeyPair.publicKey .Values.authentication.tokenBroker.secret.publicKey }}
```

helm/polaris/templates/_helpers.tpl
```diff
@@ -181,13 +181,34 @@ Prints the config volume definition for deployments and jobs.
...
+            {{- /* Backward compatibility for privateKey: new takes precedence */ -}}
+            {{- $privateKey := "" }}
+            {{- if .Values.authentication.tokenBroker.secret.rsaKeyPair.privateKey }}
+              {{- $privateKey = .Values.authentication.tokenBroker.secret.rsaKeyPair.privateKey }}
+            {{- else if .Values.authentication.tokenBroker.secret.privateKey }}
+              {{- $privateKey = .Values.authentication.tokenBroker.secret.privateKey -}}
+            {{- end }}
```
adutra:
```suggestion
            {{- $privateKey := coalesce .Values.authentication.tokenBroker.secret.rsaKeyPair.privateKey .Values.authentication.tokenBroker.secret.privateKey }}
```

helm/polaris/templates/_helpers.tpl
```diff
@@ -181,13 +181,34 @@ Prints the config volume definition for deployments and jobs.
...
           {{- end }}
           {{- if eq .Values.authentication.tokenBroker.type "symmetric-key" }}
-            - key: {{ tpl .Values.authentication.tokenBroker.secret.secretKey . }}
+            {{- /* Backward compatibility for symmetricKey: new takes precedence */ -}}
+            {{- $secretKey := "" }}
+            {{- if .Values.authentication.tokenBroker.secret.symmetricKey.secretKey }}
+              {{- $secretKey = .Values.authentication.tokenBroker.secret.symmetricKey.secretKey }}
+            {{- else if .Values.authentication.tokenBroker.secret.secretKey }}
+              {{- $secretKey = .Values.authentication.tokenBroker.secret.secretKey }}
+            {{- end }}
```
adutra:
```suggestion
            {{- $secretKey := coalesce .Values.authentication.tokenBroker.secret.symmetricKey.secretKey .Values.authentication.tokenBroker.secret.secretKey }}
```

adutra (approved):

PR #1585, opened 2025-05-14
Support retrying non-finished async tasks on startup and periodically

danielhumanmod:
Fix #774 
# Context

Polaris uses async tasks to perform operations such as table and manifest file cleanup. These tasks are executed asynchronously in a separate thread within the same JVM, and retries are handled inline within the task execution. However, this mechanism does not guarantee eventual execution in the following cases:
- The task fails repeatedly and hits the maximum retry limit.
- The service crashes or shuts down before retrying.

# Implementation Plan
Stage 1: Potential improvement - https://github.com/apache/polaris/pull/1523
Introduce per-task transactional leasing in the metastore layer via loadTasks(...)

Stage 2 (Current PR):
Persist failed tasks and introduce a retry mechanism triggered during Polaris startup and via periodic background checks, changes included:
1. **Metastore Layer:** 
   - Exposes a new API `getMetaStoreManagerMap`
   - Ensures `LAST_ATTEMPT_START_TIME` set for each task entity creation, which is important for time-out filtering when `loadTasks()` from metastore, so that prevent multiple executors from picking the same task
2. **`TaskRecoveryManager`:** New class responsible for task recovery logic, including:
   - Constructing execution`PolarisCallContext`
   - Loading tasks from metastore
   - Triggering task execution
3. **`QuarkusTaskExecutorImpl`:** Hook into application lifecycle to initiate task recovery.
4. **Task Retry Strategy:** Failed tasks remain persisted in the metastore and are retried by the recovery manager.
5. **Tests:** Adjusted existing tests and added new coverage for recovery behavior.

# Recommended Review Order
1. Metastore Layer related code
2. `TaskRecoveryManager`
3. `QuarkusTaskExecutorImpl` and `TaskExecutorImpl`
4. Task cleanup handlers
5. Tests

PR #2003, opened 2025-07-06
Support retrying non-finished async tasks on startup and periodically

### LINKED ISSUE #774: Task handling is incomplete
### Describe the bug

Polaris uses some asynchronously executed tasks to run operations for table and manifest file cleanup. Those tasks are potentially executed in a separate thread in the same JVM. There is however no guarantee that those tasks will eventually run for multiple reasons:

* Tasks (e.g. via `org.apache.polaris.service.catalog.BasePolarisCatalog#dropTable`) are triggered _after_ the fact.
* Although tasks are persisted, there is no mechanism to pick up tasks that did not start or did not finish ("long lasting" failures, JVM terminates).

Overall this means that for example a "drop table with purge" returns a successful result to the user, the actual purge may never ever happen.

### To Reproduce

_No response_

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Additional context

_No response_

### System information

_No response_

danielhumanmod:
Fix #774 
# Context

Polaris uses async tasks to perform operations such as table and manifest file cleanup. These tasks are executed asynchronously in a separate thread within the same JVM, and retries are handled inline within the task execution. However, this mechanism does not guarantee eventual execution in the following cases:
- The task fails repeatedly and hits the maximum retry limit.
- The service crashes or shuts down before retrying.

# Implementation
Persist failed tasks and introduce a retry mechanism triggered during Polaris startup and via periodic background checks, changes included:
1. **Metastore Layer:** 
   - Exposes a new API `getMetaStoreManagerMap`
   - Ensures `LAST_ATTEMPT_START_TIME` set for each task entity creation, which is important for time-out filtering when `loadTasks()` from metastore, so that prevent multiple executors from picking the same task
2. **`TaskRecoveryManager`:** New class responsible for task recovery logic, including:
   - Constructing execution`PolarisCallContext`
   - Loading tasks from metastore
   - Triggering task execution
3. **`QuarkusTaskExecutorImpl`:** Hook into application lifecycle to initiate task recovery.
4. **Task Retry Strategy:** Failed tasks remain persisted in the metastore and are retried by the recovery manager.
5. **Tests:** Adjusted existing tests and added new coverage for recovery behavior.

# Recommended Review Order
1. Metastore Layer related code
2. `TaskRecoveryManager`
3. `QuarkusTaskExecutorImpl` and `TaskExecutorImpl`
4. Task cleanup handlers
5. Tests

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -25,6 +25,7 @@ import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
 import java.sql.SQLException;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.function.Supplier;
 import javax.sql.DataSource;
@@ -241,6 +242,11 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
     }
 
     return entityCacheMap.get(realmContext.getRealmIdentifier());
+  }
+
+  @Override
+  public Iterator<Map.Entry<String, PolarisMetaStoreManager>> getMetaStoreManagerMap() {
+    return metaStoreManagerMap.entrySet().iterator();
   }
 
   /**
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/task/TableCleanupTaskHandlerTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/task/TableCleanupTaskHandlerTest.java
@@ -26,7 +26,9 @@ import io.quarkus.test.junit.QuarkusTest;
 import jakarta.annotation.Nonnull;
 import jakarta.inject.Inject;
 import java.io.IOException;
-import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -68,6 +70,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 import org.slf4j.LoggerFactory;
+import org.threeten.extra.MutableClock;
 
 @QuarkusTest
 class TableCleanupTaskHandlerTest {
@@ -78,6 +81,7 @@ class TableCleanupTaskHandlerTest {
   private CallContext callContext;
 
   private final RealmContext realmContext = () -> "realmName";
+  private final MutableClock timeSource = MutableClock.of(Instant.now(), ZoneOffset.UTC);
 
   private TaskFileIOSupplier buildTaskFileIOSupplier(FileIO fileIO) {
     return new TaskFileIOSupplier(
@@ -106,7 +110,7 @@ class TableCleanupTaskHandlerTest {
             metaStoreManagerFactory.getOrCreateSessionSupplier(realmContext).get(),
             diagServices,
             configurationStore,
-            Clock.systemDefaultZone());
+            timeSource);
   }
 
   @Test
@@ -146,6 +150,8 @@ class TableCleanupTaskHandlerTest {
     Assertions.assertThatPredicate(handler::canHandleTask).accepts(task);
 
     handler.handleTask(task, callContext);
+
+    timeSource.add(Duration.ofMinutes(10));
 
     assertThat(
             metaStoreManagerFactory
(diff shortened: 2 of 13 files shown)


runtime/service/src/test/java/org/apache/polaris/service/quarkus/task/TableCleanupTaskHandlerTest.java
```diff
@@ -147,6 +151,8 @@ public void testTableCleanup() throws IOException {
 
     handler.handleTask(task, callContext);
 
+    timeSource.add(Duration.ofMinutes(10));
```
danielhumanmod, line 154: @adnanhemani continue the [previous comment](https://github.com/apache/polaris/pull/1585/files#r2094263997) here:
>  Can you explain this further - I'm not sure why the tests need this 10m jump? Is it so that tasks are "recovered" by the Quarkus Scheduled method?

`metaStoreManager.loadTasks` fetches available tasks from the metastore — meaning tasks that are either not leased by any executor or whose lease has already timed out (after 5 minutes).

In this test, the new tasks are created and not executed by its parent task, so to ensure these tasks are fetched, we need to simulate a time gap longer than 5 minutes.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -243,6 +244,11 @@ public synchronized EntityCache getOrCreateEntityCache(RealmContext realmContext
     return entityCacheMap.get(realmContext.getRealmIdentifier());
   }
 
+  @Override
+  public Iterator<Map.Entry<String, PolarisMetaStoreManager>> getMetaStoreManagerMap() {
```
danielhumanmod, line 248: Make it to Iter<Map.Entry> as https://github.com/apache/polaris/pull/1585/files#r2121942046 suggests

service/common/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
```diff
@@ -188,6 +193,9 @@ private Stream<TaskEntity> getManifestTaskStream(
                   .withData(
                       new ManifestFileCleanupTaskHandler.ManifestCleanupTask(
                           tableEntity.getTableIdentifier(), TaskUtils.encodeManifestFile(mf)))
+                  .withLastAttemptExecutorId(executorId)
+                  .withAttemptCount(1)
```
danielhumanmod, line 197: @adnanhemani continue [previous discussion](https://github.com/apache/polaris/pull/1585/files#r2122664565) here: 

> How can we assume this?

This new task (`ManifestFileCleanupTask`) is created by the current task (`TableCleanupTask`) and will be executed immediately at the end of it.
Since it’s its first execution, we set the attemptCount to 1 here.

service/common/src/main/java/org/apache/polaris/service/task/TaskRecoveryManager.java
```diff
@@ -0,0 +1,93 @@
...
+          metaStoreManagerFactory.getOrCreateSessionSupplier(realmContext).get();
+      PolarisCallContext polarisCallContext =
+          new PolarisCallContext(
+              realmContext,
+              metastore,
+              new PolarisDefaultDiagServiceImpl(),
+              configurationStore,
+              clock);
+      EntitiesResult entitiesResult =
+          metaStoreManager.loadTasks(polarisCallContext, executorId, PageToken.fromLimit(20));
```
danielhumanmod, line 79: @adnanhemani continue [previous comment](https://github.com/apache/polaris/pull/1585/files#r2122779304) here: 

> I'm not sure I'm understanding the logic here: we are asking for 20 tasks here - but what if there are more than 20 tasks that need recovery?

Good catch, we should have a loop here to fetch all tasks

service/common/src/main/java/org/apache/polaris/service/task/TaskRecoveryManager.java
```diff
@@ -0,0 +1,93 @@
...
+          metaStoreManagerFactory.getOrCreateSessionSupplier(realmContext).get();
+      PolarisCallContext polarisCallContext =
+          new PolarisCallContext(
+              realmContext,
+              metastore,
+              new PolarisDefaultDiagServiceImpl(),
+              configurationStore,
+              clock);
+      EntitiesResult entitiesResult =
+          metaStoreManager.loadTasks(polarisCallContext, executorId, PageToken.readEverything());
```
danielhumanmod, line 79: @adnanhemani continue [previous comment](https://github.com/apache/polaris/pull/1585/files#r2122779304) here: 

> I'm not sure I'm understanding the logic here: we are asking for 20 tasks here - but what if there are more than 20 tasks that need recovery?

Good catch, we should update to read all pending tasks

PR #2004, opened 2025-07-07
OpenAPI-generate: Omit generation timestamp

snazy:
The jaxrs-resteasy OpenAPI generator adds the generation timestamp to the generated sources by default. This behavior leads to different code for every generation, leading to unnecessary rebuilds (and re-tests), because the generated `.class` files are different.


dimas-b (approved):

dimas-b:
This looks like a reasonable and valuable build improvement to me... merging. If there are objections later we can certainly discuss / make follow-up PRs.

PR #2005, opened 2025-07-07
Fix bunch of OpenAPI generation issues

snazy:
The current way how OpenAPI Java code is generated suffers from a bunch of issues:
* Changes to any of the source spec files requires a Gradle `clean`, otherwise old generated Java source will remain - i.e. "no longer" existing sources are not removed. This is addressed by adding an additional action to `GenerateTask`.
* The output of `GenerateTask` was explicitly not cached, this is removed, so the output is cached.
* Add explicit inputs to `GenerateTask` to the whole templates and spec folders.

### DIFF
--- a/api/iceberg-service/build.gradle.kts
+++ b/api/iceberg-service/build.gradle.kts
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
+
 plugins {
   alias(libs.plugins.openapi.generator)
   id("polaris-client")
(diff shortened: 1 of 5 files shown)


api/iceberg-service/build.gradle.kts
```diff
@@ -109,16 +122,16 @@ openApiGenerate {
...
+  inputs.dir(templatesDir)
+  inputs.dir(specsDir)
+  actions.addFirst { delete { delete(generatedDir) } }
 }
 
 tasks.named("javadoc") { dependsOn("jandex") }
-
-tasks.named("processResources") { dependsOn("openApiGenerate") }
-
-tasks.named("openApiGenerate") { outputs.cacheIf { false } }
```
eric-maynard, line 124: In the past, this was needed to make CI pass after making any spec changes. Are you able to confirm that this isn't the case now?

snazy: That's mentioned in the first bullet point of the PR description.

eric-maynard: In fact it's not -- or at least it was not obvious to me that these two things were one and the same. Thanks for confirming!

PR #2007, opened 2025-07-07
Make PolarisConfiguration member variables private

poojanilangekar:
As pointed out in #1959, this change converts all public member variable to private variables.


poojanilangekar:
CC @dimas-b

dimas-b (approved):
Thanks @poojanilangekar ! LGTM :+1:

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -124,6 +124,18 @@ T cast(Object value) {
...
+    return key;
+  }
+
+  public String description() {
+    return description;
+  }
+
+  public T defaultValue() {
+    return defaultValue;
+  }
```
eric-maynard: Is there actually a use case for method/nonvalue implementations of these? I notice the methods are not `final`.

If not, it might not be worth breaking the interface and adding the extra overhead for callers. The values were already `final` so there's no risk of someone changing a config in an unexpected way.

poojanilangekar: Adding @dimas-b's comment here, (and let him address your question):

> Sorry, I did not notice this before, but I think having public fiends in general is not a good idea. Since you're toughing this code in current PR, would you mind following up with another PR to convert them to private fields?

I agree that it's probably a bad idea to have public member variables, but don't feel strongly either way. 
The other part that I'd be concerned about is the inconsistent public/private handling of fields. Since the fields are all `final`, why shouldn't `catalogConfigImpl`, `catalogConfigUnsafeImpl`, and `typ` be `public`?

eric-maynard: > I think having public fiends in general is not a good idea.

I'm not sure if I find that a compelling enough reason to make this change. It's generally considered a good idea because it affords you more flexibility to change the behavior of e.g. `key()` without breaking an API, but
1. In this case, we're proposing to actually break the API by removing public field `key`
2. Do we actually want to support that kind of flexibility? If so let's make this change, but originally my intent was to enforce that the child types have only a "plain" implementation of e.g. `key`. 

> Since the fields are all final, why shouldn't catalogConfigImpl, catalogConfigUnsafeImpl, and typ be public?

Those fields don't have a use outside this class.

poojanilangekar: > Those fields don't have a use outside this class.

That's inaccurate, there are several places in the code where `catalogConfig()` is invoked. Which is simply trying to access the `catalogConfigImpl` if present.

dimas-b: Any java symbols inside `polaris-core` do not have backward-compatibility guaranteed per https://polaris.apache.org/in-dev/unreleased/evolution/

dimas-b: Any java symbols inside `polaris-core` do not have backward-compatibility guaranteed per https://polaris.apache.org/in-dev/unreleased/evolution/#using-polaris-as-a-build-time-dependency

dimas-b: Exposing direct field access outside of a class makes refactoring harder, debugging harder, and reasoning about code behaviour harder... IMHO.

eric-maynard: There is no use for `catalogConfigImpl` -- in fact, that's my point. We always want the check in `catalogConfig` and we never want to directly access `catalogConfigImpl`, which is why `catalogConfig` is a method. If we have a similar situation for the other fields like `key`, then I think accessor methods are a good idea. If not, I'm not sure we need to break the API at this time.

eric-maynard: > Any java symbols inside polaris-core do not have backward-compatibility guaranteed per https://polaris.apache.org/in-dev/unreleased/evolution/

This works against the argument to refactor, then -- the idea behind an accessor method which just returns a private field is to add a layer of abstraction to make changes _later_ without breaking an API, but if your argument is that breaking the API is okay then I don't see an impetus to add an accessor method at this time.

> Exposing direct field access outside of a class makes refactoring harder, debugging harder, and reasoning about code behaviour harder... IMHO.

"Honestly"... I disagree. With private fields like this, there's a chance that BehaviorChangeConfiguration and FeatureConfiguration can diverge from one another and from PolarisConfiguration in unexpected ways. If nothing else the fields should be protected. I don't see how a method call that returns a field can possibly make debugging easier than just having the field be accessed.

snazy: `public` fields are generally considered bad practice, carefully designed interfaces are generally preferred. This is how things are composable and provide much better backwards compatibility. There are a lot of prominent examples in Java's core libraries. (Spoiler: Would anyone expect `java.util.Map.Entry.key` being a field?)

In the past, a lot of changes went into Polaris that count as "breaking changes" according to the above arguments. Some received strong objections while others were just approved.

dimas-b: > I don't see how a method call that returns a field can possibly make debugging easier than just having the field be accessed.

Break points on a field do not work well in my experience (e.g. they do not catch reflected access, IIRC).

dimas-b: > With private fields like this, there's a chance that BehaviorChangeConfiguration and FeatureConfiguration can diverge from one another and from PolarisConfiguration in unexpected ways. If nothing else the fields should be protected.

I'm fine with `protected` fields if they are also `final`.

In this case, I do not think it is necessary. If we're concerned with preserving the behaviour of field accessors, we could make the accessor methods `public` and `final`.

That said, divergence among those classes is possible in any case, I think... This is why we have code reviews :)

eric-maynard: Yeah I was thinking the same in my original comment, if the accessors are `public final` then we at least do not have to worry about divergence in the immediate concern and then the only real issue is that we're changing the API for seemingly no value.

> Break points on a field do not work well in my experience (e.g. they do not catch reflected access, IIRC).

Watches are generally used for fields, not breakpoints. 

Reflection and proxying can cause issues for breakpoints both fields and methods, which is one reason to avoid them. Unfortunately, you can observe this when working with method breakpoints in many proxied objects in Polaris today.

dimas-b: @eric-maynard : would you be ok with making these fields `private` and their accessor methods `public final`? Just to confirm :)

eric-maynard: yep!

dimas-b: @poojanilangekar : would you mind applying this suggestion?

poojanilangekar: Done.

dimas-b (dismissed):

dimas-b:
@poojanilangekar : it looks like some direct field access still exists (compilation errors) :thinking: some parallel code changes perhaps...

poojanilangekar:
This is strange, the build passed on my machine. I will try and fix it.

poojanilangekar:
Turns out I just needed to rebase.

dimas-b (approved):

dimas-b:
If no other feedback, I'll merge tomorrow.

PR #2011, opened 2025-07-08
Restructure the download page

flyrain:
Before:
Screenshot 2025-07-08 at 4 03 55 PM
After:
Screenshot 2025-07-08 at 5 32 33 PM


site/hugo.yaml
```diff
@@ -93,9 +93,6 @@ imaging:
 
 menu:
   main:
-    - name: GitHub
-      url: https://github.com/apache/polaris
-
```
flyrain, line 98: This change makes "github" menu a bit nicer, there is no extra space now.

site/content/downloads/current-release.md
```diff
@@ -0,0 +1,33 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+linkTitle: "Current Release"
+weight: 100
+---
+
+### Apache Polaris 0.9.0
```
dimas-b: With this require manually moving "previous" entries to the "past releases" page on every release?

I understand we have to edit this page on every release, but adding entries on top is generally easier that copy-paste across pages :thinking:

dimas-b: How about a page per version? "Current" could be a link to the latest version page... WDYT?

flyrain: ~~yeah, it requires a copy-over.~~ I think the current release page is pretty light at this moment, but overtime, there are more things to add, docker image, helm chart instruction, etc. Or any general comments and instructions could be in that page as the index page of the releases. In that case, putting all release notes there wouldn't be easy to discover. 

This is just my first version. I'm nowhere an UI expert. Feel free to change that.

flyrain: A page per version is also a solution. Spark did that but only for release notes, https://spark.apache.org/releases/spark-release-4-0-0.html.

flyrain: Sorry for confusing. It actually doesn't require a copy-over action. Every time we got a new release, the past releases page will have all of them including the latest one. Only the current release page needs to change.

dimas-b: Well,  the screen capture in the description does have release nodes, which is actually why I thought a page per release would be easier to manage :)

dimas-b: I'm not a Hugo expert myself, but if we can get links / redirects to work then user should still be able to bookmark the generic "downloads" URL and get the latest release in the browser.

flyrain: I didn't quite get it. How about this? merging this, first then you can file a PR to include your ideas on top of it. This way, we can iterate faster.

dimas-b: My comments are non-blocking. Let's see what other reviews think.

site/content/downloads/current-release.md
```diff
@@ -0,0 +1,33 @@
...
+
+Apache Polaris 0.9.0 release is the first Polaris release. Only the source distribution is available for this release.
+
+#### Download Links
+
+- [Source (tar.gz)](https://downloads.apache.org/incubator/polaris/0.9.0-incubating/apache-polaris-0.9.0-incubating.tar.gz)
+- [PGP Signature (.asc)](https://downloads.apache.org/incubator/polaris/0.9.0-incubating/apache-polaris-0.9.0-incubating.tar.gz.asc)
+- [SHA512 Checksum](https://downloads.apache.org/incubator/polaris/0.9.0-incubating/apache-polaris-0.9.0-incubating.tar.gz.sha512)
+- [KEYS File](https://downloads.apache.org/incubator/polaris/KEYS)
```
gh-yzou: should this be 1.0? it seems we have both previous and current release that are 0.9? or we are going to change this in a different PR?

flyrain: 1.0 isn't ready yet -- I will add it once it is approved. The past releases page will include all releases including the current release.

site/content/downloads/current-release.md
flyrain: Consolidated them into a single page.
Screenshot 2025-07-08 at 5 32 33 PM

gh-yzou: @flyrain just unified the current release and past release page, so we just need to update one page on every release, which seems making things simplier

gh-yzou (approved):

gh-yzou:
@flyrain you might want to update the description since it doesn't match what this PR is actually doing now

flyrain:
> @flyrain you might want to update the description since it doesn't match what this PR is actually doing now

any suggestion for the description?

gh-yzou:
I see you updated the description with just one release page now, that should be good enough

PR #2012, opened 2025-07-08
Add `pathStyleAccess` to AwsStorageConfigInfo

dimas-b:
This change allows configuring the "path-style" access mode in S3 clients (both in Polaris Servers and Iceberg REST Catalog API clients).

This change is applicable both to AWS storage and to non-AWS S3-compatible storage (#1530).

The Management API change is backward-compatible with clients lenient to unknown JSON properties.

Includes a fix to AWS config object round-trip via generated API classes.

Dev ML discussion: https://lists.apache.org/thread/oxy2yd4mbq06zr4z57lfptgy95c3lyhb

### DIFF
--- a/spec/polaris-management-service.yml
+++ b/spec/polaris-management-service.yml
@@ -1060,7 +1060,6 @@ components:
               type: boolean
               description: >- 
                 Whether S3 requests to files in this catalog should use 'path-style addressing for buckets'.
-                Default: false.
               example: true
               default: false
           required:


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -71,21 +76,23 @@ public AwsStorageConfigurationInfo(
...
-      @JsonProperty(value = "stsEndpoint") @Nullable String stsEndpoint) {
+      @JsonProperty(value = "stsEndpoint") @Nullable String stsEndpoint,
+      @JsonProperty(value = "pathStyleAccess") @Nullable Boolean pathStyleAccess) {
     super(storageType, allowedLocations);
     this.roleARN = roleARN;
     this.externalId = externalId;
     this.region = region;
     this.endpoint = endpoint;
     this.stsEndpoint = stsEndpoint;
+    this.pathStyleAccess = Optional.ofNullable(pathStyleAccess).orElse(false);
```
eric-maynard: I'm a huge `Optional` booster, but this is not any shorter than a ternary expression. More importantly though, I think it would make sense to have a `PATH_STYLE_ACCESS_DEFAULT` or something like that.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -71,21 +76,23 @@ public AwsStorageConfigurationInfo(
...
   }
 
   public AwsStorageConfigurationInfo(
       @Nonnull StorageType storageType,
       @Nonnull List<String> allowedLocations,
       @Nonnull String roleARN,
       @Nullable String region) {
-    this(storageType, allowedLocations, roleARN, null, region, null, null);
+    this(storageType, allowedLocations, roleARN, null, region, null, null, false);
```
eric-maynard: We could use the default here, or just leave it null to leverage the default used in `orElse`. Better not to proliferate the constants around like this.

spec/polaris-management-service.yml
```diff
@@ -1056,6 +1056,12 @@ components:
               type: string
               description: endpoint for STS requests (optional). If not set, defaults to 'endpoint'.
               example: "https://sts.example.com:1234"
+            pathStyleAccess:
+              type: boolean
+              description: >- 
+                Whether S3 requests to files in this catalog should use 'path-style addressing for buckets'.
+                Default: false.
```
eric-maynard: I believe OpenAPI actually supports default values, maybe we could just use that?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -149,6 +156,12 @@ public URI getEndpointUri() {
     return endpoint == null ? null : URI.create(endpoint);
   }
 
+  /** Returns a flag indicating whether path-style bucket access should be forced in S3 clients. */
+  @JsonIgnore
+  public boolean forcePathStyleAccess() {
```
snazy: Nit:
```suggestion
  public boolean getPathStyleAccess() {
```
... all other functions are `get`ters

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -94,7 +101,7 @@ public AwsStorageConfigurationInfo(
       @Nonnull String roleARN,
       @Nullable String externalId,
       @Nullable String region) {
-    this(storageType, allowedLocations, roleARN, externalId, region, null, null);
+    this(storageType, allowedLocations, roleARN, externalId, region, null, null, false);
```
singhpk234: when we have `PATH_STYLE_ACCESS_DEFAULT` defined we can wire in that or just null here so that actual default get applied with Optional handling L87

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
dimas-b: changed to use `PATH_STYLE_ACCESS_DEFAULT`

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
dimas-b: changed to ternary + `PATH_STYLE_ACCESS_DEFAULT`

spec/polaris-management-service.yml
dimas-b: Added `default: false`, changed example to `true`.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
singhpk234: you beat me to publish the review @dimas-b :) !

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
eric-maynard: Now that we have a default at the API level, is this actually nullable?

spec/polaris-management-service.yml
dimas-b: Actually I had to roll this back. Using Open API defaults breaks strict backward compatibility because is causes the attribute with a `default` value to be always returned to clients (even when not in use) - CI caught this.

dimas-b:
I had to remove one commit (discussed in the API yaml comment thread). PTAL.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
dimas-b: I had to remove the API default.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
dimas-b: renamed

spec/polaris-management-service.yml
eric-maynard: What exactly was the failure in CI? The client would fail to parse a response because it didn't recognize the property?

dimas-b: `CatalogSerializationTest` detected a change in the JSON text (when the new properties are not set)

eric-maynard: Does that actually represent a breaking change?

As best I can tell this would not break clients, and even if it did the problem would not be exclusive to `default`. If you actually read a catalog with the value set you would see the same behavior.

dimas-b: Older clients may not be prepared to handle the new `pathStyleAccess` attribute. This is a risk. Strictly speaking, exposing new JSON attributes in an old type is a breaking change even if the value is default.

dimas-b: New catalogs that have this value configured could be incompatible with older clients. However, this is under the user / operator control. If the catalog data is unchanged, this change is backward-compatible.

eric-maynard: > However, this is under the user / operator control. 

Is it? If I'm reading your PR correctly once I create a new StorageConfigInfo, it will have a value of `false` written into it for `pathStyleAccess`... and that will be returned to all clients. There's no way for me to control that besides just not making any new catalogs. At least, that's what "Default: false" in the description from the spec seems to imply.

If you consider that a breaking change, then this PR is already a breaking change and in that case I think we should just use a default in the API.

If you do not consider that a breaking change, then it's also not a breaking change to use a default in the API.

snazy (dismissed):
LGTM

spec/polaris-management-service.yml
dimas-b: #1955 removes properties with `null` values from REST API responses. As long as `pathStyleAccess` is `null` in `AwsStorageConfigInfo` clients will not see it.

dimas-b: @eric-maynard : your previous comment made me realise that the new `AwsStorageConfigurationInfo` properties are not properly converted back to `AwsStorageConfigInfo` yet. I'll tackle that as a separate PR, if you do not mind.

I'll convert `pathStyleAccess` to object `Boolean` in `AwsStorageConfigurationInfo` to ensure unset values are distinct from defaults (and thus do not affect older catalogs).

eric-maynard: Oh, I think that's okay to defer.

But...
> As long as pathStyleAccess is null in AwsStorageConfigInfo clients will not see it.

That's only true if you don't create a new catalog. So take the following pseudocode:
```
String catalog = ...createNewCatalog().toString
assert(catalog == "{...}")
```

We don't have a test like this in `CatalogSerializationTest`, but if we had one this PR would break it. Personally I don't think this really constitutes a breaking change, but I do think the proper way to handle a default in the API is with a `default` in the spec rather than a blurb in the description.

dimas-b: I agree that we do not have a specific test for new property values in REST responses. However, `CatalogSerializationTest` makes sure defaults in new properties in `AwsStorageConfigInfo` do not show up in JSON.

As I mentioned above, we're missing code for new properties to be propagated from `AwsStorageConfigurationInfo` to `AwsStorageConfigInfo`, so I cannot even add a test for that without extra changes, which is what I'm proposing to do in a follow-up PR.

dimas-b: actually, never mind, I'll expand this PR a bit :)

dimas-b: updated. PTAL.

eric-maynard: Thanks, the changes look better, but my concern about the default remains

snazy: TBH, I do not understand the whole discussion about "is this a breaking change".

As Polaris is primarily an Iceberg REST catalog, we _have to_ let JSON parsing _ignore_ unknown properties, because that is a requirement from Iceberg REST. There is also exactly _one_ JSON parser (`ObjectMapper`) for all endpoints. The situation is unfortunate but a consequence of generating the OpenAPI code. Jackson itself allows much more fine grained control over how types are (de)serialized and allows back/forwards compatibility targeting functionality like `@JsonView` - but that's not usable w/ the code generation approach.

So fact is that unknown properties in JSON objects in any Polaris REST payload have to be ignored. As that fact leaks to all Polaris APIs, it also leaks to clients. This means, that a new property is not a breaking change. If clients fail on unknown properties, it's a bug in those client.

dimas-b:
@eric-maynard : I believe I addressed your backward-compatibility concern in the latest push. Could you review again?

spec/polaris-management-service.yml
dimas-b: @eric-maynard : the new test shows AWS config data roundtrips via API with and without (defaults) values for the new properties.

If the new properties are not set, clients will not see them in JSON outputs of the REST API endpoints. So, this change does not break existing installations and old clients as long as Polaris users do not use the new properties. If users do reconfigure catalogs to use the new properties, it's the user's responsibility to also update clients to be compatible with the catalog JSON configuration (e.g. use new CLI).

eric-maynard: @dimas-b it sounds like you don't agree with @snazy and I that adding the new default would not constitute a breaking change? Serialization tests are fine, but only as a guard against accidental regression. Just because a serialization changes does not make something a breaking change.

dimas-b: My point is two-fold:
1) we add a new property - this can potentially break old clients if they are sensitive to the presence of unknown properties (unlikely but possible)
2) the way this PR is implemented does not expose new properties in JSON responses if they are not explicitly configured - this ensures backward compatibility with in existing servers. (I also mentioned that in the dev email).

If a user decided to configure the new properties, the user has to assume responsibility for possible client-side breakage (which, again, in unlikely).

Does this sound acceptable to you? (note: Robert did not -1 this PR :) )

snazy: > Robert did not -1 this PR

correct, i'm +1

eric-maynard: No, that does not sound acceptable. If the new field has a default, I think that we should have a default in the API rather than a comment saying there's a default. The argument to _not_ do that (both your 1 & 2) seems to be that having a default somehow makes this a breaking change, which @snazy argues above is not the case. I agree with that.

dimas-b: If we add a `default` declaration in the Open API YAML, then the Open API generators will produce API-level classes in a way that will _always_ put the `pathStyleAccess` attribute into JSON responses.

My changes in this PR specifically avoid that situation to ensure "no behaviour" change in existing servers.

I personally think that this kind of backward-compatibility guarantee is preferable to adding a `default` declaration to the Open API YAML.

@eric-maynard : what benefit do you see in having `default` in Open API YAML?

eric-maynard: > I personally think that this kind of backward-compatibility guarantee is preferable to adding a default declaration to the Open API YAML.

This is the crux of the debate here, I guess. I don't see why that's preferable.

> what benefit do you see in having default in Open API YAML?

Besides consistency with the rest of the spec(s), it seems intuitive that we should use the `default` feature to make a `default`. This could, for example, help generated clients populate the correct value.

dimas-b: @eric-maynard : just to clarify: do you prefer adding `default` in the Open API spec for `pathStyleAccess` even though this has some risk of breaking existing clients (as explained above)?

eric-maynard: >  do you prefer adding default in the Open API spec 

Yes, if the field has a default I think that should be governed by the spec.

> this has some risk of breaking existing clients (as explained above)?

The _marginal_ risk increase of adding a default seems small. Besides the fact that the appearance of a new field isn't a breaking change as @snazy and I argue above, the new field also appears if you create new a catalog using the PR as-is

snazy: Sure, it's not a breaking change. But why yield the default value of a field at all? The default for a boolean is false, also the meaning of "not present" is also false.

dimas-b: > the new field also appears if you create new a catalog using the PR as-is

Not quite. If you do not set the new fields in a new catalog, they will not appear in REST responses.

This is the key point from my POV. We delegate behaviour changes to the user / admin of the catalog as opposed to forcing all clients to deal with the default value of the new attribute even if their catalogs do not have to set it.

eric-maynard: > But why yield the default value of a field at all? The default for a boolean is false, also the meaning of "not present" is also false.

This depends on whether you expect the client or the server to interpret "not present". If the client thinks the server is handling it (i.e. never sending "not present") and the server thinks the client is handling it, you have a problem. To resolve this ambiguity, OpenAPI supports adding a default to the spec itself.

> We delegate behaviour changes to the user / admin of the catalog as opposed

The service _admin_  has little control over whether users create a new catalog with this config set. They do however have control over whether to upgrade/downgrade the service.

> forcing all clients to deal with the default value of the new attribute even if their catalogs do not have to set it.

As @snazy said, `If clients fail on unknown properties, it's a bug in those client.`

dimas-b: > If clients fail on unknown properties, it's a bug in those client.

While I agree with this statement, I do not think Polaris should aggravate this situation, where we can avoid it (i.e. no using `default`).

In other words, I'm proposing to take a more lenient approach. Not providing a value for `pathStyleAccess` when creating a catalog does not alter (old) Polaris catalog behaviour and is not even visible to clients (old or new). The Polaris Server behaviour in this case is equivalent to having the default value for `pathStyleAccess`, yet we get the benefit of broader client compatibility.

eric-maynard: Unfortunately I don't think that's the right decision for the spec. With an optional boolean you create a sort of ternary type with values `[true, false, unset]` and there is ambiguity about whether `unset` refers to "really not set" like in the case of a catalog that predates the change or or whether it refers to the default. Imagine what would happen if, for example, we changed the default down the road.

Further, I don't see why `does not alter (old) Polaris catalog behaviour` is necessarily a merit. We've said that it's not a breaking change. Even if it is, we don't need to avoid breaking changes at all costs. Let's not compromise the spec to avoid a non-breaking change.

dimas-b: very well - I'll put `default` in the Open API spec for `pathStyleAccess` presently

dimas-b: updated

eric-maynard (dismissed):

spec/polaris-management-service.yml
```diff
@@ -1056,6 +1056,13 @@ components:
...
```
eric-maynard: nit: With the default in the spec, I don't think we need the default in the doc. Other similar variables don't have it.

dimas-b: removed :+1:

PR #2015, opened 2025-07-09
add PolarisRealmConfig

XN137:
WIP


polaris-core/src/main/java/org/apache/polaris/core/config/PolarisRealmConfig.java
```diff
@@ -0,0 +1,66 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.config;
+
+import jakarta.annotation.Nullable;
+import org.apache.polaris.core.entity.CatalogEntity;
+
+/** Realm-specific configuration used to retrieve runtime parameters. */
+public interface PolarisRealmConfig {
```
snazy: Why `Polaris*`? It's already in the package name.

XN137: we already have these classes:
```
./runtime/service/src/main/java/org/apache/polaris/service/quarkus/auth/QuarkusAuthenticationRealmConfiguration.java
./runtime/service/src/main/java/org/apache/polaris/service/quarkus/context/QuarkusRealmContextConfiguration.java
./service/common/src/main/java/org/apache/polaris/service/auth/AuthenticationRealmConfiguration.java
./service/common/src/main/java/org/apache/polaris/service/context/RealmContextConfiguration.java
```
since we are internally reading from a `PolarisConfigurationStore` i thought keeping the same prefix makes sense and keeps some distinction.
i am open to change this if we can agree on a more suitable name (i dont mind either way)

snazy: Yea - but this is Polaris, no need to repeat something that's already in the package name.

XN137: ok dropped the prefix in another commit

eric-maynard:
If we're making this change for ergonomic reasons, I don't think the new invocation is really ideal: `x.getRealmConfig().getConfig(y)`. Is the return value of `getRealmConfig` the "config" or the return value of `getConfig(y)`? They both seem to claim they're fetching a config.

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
```diff
@@ -0,0 +1,66 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.config;
+
+import jakarta.annotation.Nullable;
+import org.apache.polaris.core.entity.CatalogEntity;
+
+/** Realm-specific configuration used to retrieve runtime parameters. */
```
eric-maynard, line 24: Is it really a `Realm-specific configuration`? This looks more like a configuration **store**, like `PolarisConfigurationStore`.

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.apache.polaris.core.config;
+
+import jakarta.annotation.Nullable;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.core.entity.CatalogEntity;
+
+public class RealmConfigImpl implements RealmConfig {
+
+  private final PolarisConfigurationStore configurationStore;
```
eric-maynard, line 27: If a new `PolarisConfigurationStore` implementation could achieve the same thing, we should try that approach first

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
XN137: depends on one's definition i guess, currently the interface allows retrieval of configuration values. in my book that makes it a "configuration".

do you want me to add "store" to the javadoc or to also rename the class? if the latter, any concrete name suggestions?

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
XN137: the idea of having the interface is to make the life of the callers easier as they most often (/always?) operate in the context of a single realm / `CallContext`.

whether configuration values are stored by a single store for all/multiple realms or not is an implementation detail that callers should not care about.
so afaict `PolarisConfigurationStore` can keep its current implementation unless in the future we notice any advantages of "splitting it up".

XN137:
> I don't think the new invocation is really ideal: x.getRealmConfig().getConfig(y). Is the return value of getRealmConfig the "config" or the return value of getConfig(y)? They both seem to claim they're fetching a config.

`PolarisConfigurationStore.getConfiguration` is also returning configuration values and not a configuration and certainly not a `PolarisConfiguration` (which is a parameter not a return value). :sweat_smile: 

but you are right, i also wasnt very convinced of the `getConfig` name, would you prefer `getConfigValue` or `getValue` or maybe just `get`? would any of those address your concern?

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
eric-maynard: I think calling it a RealmConfigurationStore would be a better name, but that also only makes me think more that this needs to be reconciled with PolarisConfigurationStore

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
eric-maynard: > whether configuration values are stored by a single store for all/multiple realms or not is an implementation detail that callers should not care about.

Isn't that a reason that you'd want to just inject a `PolarisConfigurationStore` rather than handling something called a `RealmConfig`?

dimas-b: `PolarisConfigurationStore` is essentially multi-realm (note its explicit realm method parameters). This class acts as a request-scoped facade to automatically contextualize request-specific config lookup with realm IDs and pass those calls to `PolarisConfigurationStore`.

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
dimas-b: I would not mind the proposed renaming, but current name LGTM too.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -82,11 +78,9 @@ public StorageCredentialCache(
   /** How long credentials should remain in the cache. */
   private long maxCacheDurationMs() {
     var cacheDurationSeconds =
-        configurationStore.getConfiguration(
-            realmContext, FeatureConfiguration.STORAGE_CREDENTIAL_CACHE_DURATION_SECONDS);
+        realmConfig.getConfig(FeatureConfiguration.STORAGE_CREDENTIAL_CACHE_DURATION_SECONDS);
```
dimas-b, line 79: I'm not sure this is 100% correct. `RealmConfig` is a request-scoped bean here, but `maxCacheDurationMs()` may be called by Caffeine for a key in one ream when the cache is operating in a request context from another realm. This could cause configuration confusion across realms, I think.

dimas-b: I believe it is preferable to invoke `maxCacheDurationMs()` proactively when a key is inserted and then reuse the value in Caffeine callbacks (line 68). This duration configuration will remain true to the realm that owns the key.

dimas-b: note: the problem I describe above existed before this PR... but since we're touching this code, I think we ought to fix it :)

dimas-b (commented):
Thanks, @XN137 ! The PR LGTM overall. I think it's a nice improvement in request/application-scoped delineation in the Polaris codebase.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: After reading the related but not changed code, I realised that `StorageCredentialCache` is currently created and kept in-memory on a one instance per realm basis. So the above statement is a non-issue until we make `StorageCredentialCache` application-scoped. Sorry about the confusion.

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
XN137: i'll create a separate PR to address these concerns but imo its not related to this PR

dimas-b: sgmt. Thanks, @XN137 !

dimas-b:
@eric-maynard : are you ok with merging this PR?

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
eric-maynard: I'm all for refactoring `PolarisConfigurationStore` to have both realm-agnostic and realm-gnostic (?) methods. But I do not think a new layer like this over the config store is necessary.

dimas-b: The new layer helps to properly delineate application-scoped data (`PolarisConfigurationStore` and/or it's config source) from request-scoped data (`RealmConfig`). Effectively, request-scoped code does not need to deal with explicit realm ID parameters, which, IMHO, is beneficial to code clarity and maintainability.

eric-maynard: Not having to deal with realms is definitely a great improvement, agreed. But my concern remains that this new class becomes the de facto “Polaris Configuration Store” and this redundancy does not improve clarity. Furthermore, the proposed structure could be interpreted as implying that these configs are set per-realm (while PolarisConfigurationStore confits are not) when in fact this is not the case.

XN137:
rebased after 19f44d8e892d80920980d1d03555535b37147923 due to conflicts:
```
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogViewTest.java
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/PolarisGenericTableCatalogTest.java
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/PolicyCatalogTest.java
```

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
XN137: > Not having to deal with realms is definitely a great improvement, agreed.

if you agree with that, don't you also agree that having "realm-scoped wrappers" around "multi-realm implementation details" simplifies code that operates strictly within a single realm?

> But my concern remains that this new class becomes the de facto “Polaris Configuration Store” and this redundancy does not improve clarity. 

it is just an interface for accessing realm-scoped configuration values. `RealmConfigImpl` is forwarding requests to the actual `PolarisConfigurationStore` under the hood.
it does not replace or overtake the "storage aspect" of configuration values and is thus not redundant.
would it help address your concern if we rename `RealmConfigImpl` to `StoreForwardingRealmConfig` ?
or to `RealmScopedConfigurationStore` (even though it doesnt store anything by itself) ?

> Furthermore, the proposed structure could be interpreted as implying that these configs are set per-realm (while PolarisConfigurationStore confits are not) when in fact this is not the case.

all public methods of `PolarisConfigurationStore` take a `@Nonnull RealmContext` parameter as there can be realm-specific overrides which take precedence.
so if we not only look at the name of the class but at its public api as a whole, it does indeed only allow retrieving configuration values per realm currently, as you can not retrieve any configuration value without providing a `RealmContext`.

if you have concrete ideas how class-naming could improve the "clarity" and "interpretation" of the current suggestion please let me know.

XN137:
rebased due to conflicts with 15f23cae8e678f8a13c20c5012919565288525a0
```
service/common/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java
```

XN137:
rebased (and squashed the rename) due to conflicts after 6ddd1487877ce13f2640ea73437b24bbd7c3eb09

polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
eric-maynard: Looks like @snazy merged this, but in response to the above:

> don't you also agree that having "realm-scoped wrappers" around "multi-realm implementation details" simplifies code that operates strictly within a single realm?

No, not really -- all callers will now just use `RealmConfig`, so `PolarisConfigurationStore` becomes unnecessary. Yes, it is used to implement `RealmConfig`, but we need to consider which is a better interface to prospective callers who want to load a config. Is it `RealmConfig` and its methods or `PolarisConfigurationStore`? Probably the latter, which could have been extended into some `RealmConfigurationStore` which includes methods to load configs without you having to pass in a realm.

> as there can be realm-specific overrides which take precedence.

Is there a case today where we pass in a realm other than the one which would get injected? That is the only situation where I see someone wanting to directly interface with `PolarisConfigurationStore` now.

dimas-b: I agree that `PolarisConfigurationStore` becomes unnecessary. However, from my POV the main advantage of this PR is detaching request-scoped config activities from explicit realm IDs, so I'd say using `RealmConfig` is preferable.

PR #2021, opened 2025-07-10
Make StorageCredentialCache safe for mutli-realm usage

XN137:
Injecting the request-scoped `RealmContext` into the application-scoped `StorageCredentialCache` makes things unnecessarily complicated.
Similarly `StorageCredentialCacheKey` having a `@Nullable callContext` makes it more difficult to reason about.


runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
```diff
@@ -465,7 +465,7 @@ public Supplier<TransactionalPersistence> getOrCreateSessionSupplier(
 
       @Override
       public StorageCredentialCache getOrCreateStorageCredentialCache(RealmContext realmContext) {
```
snazy: If `RealmContext` is no longer needed, let's remove or at least deprecate-for-removal this function and add one that takes no parameters.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -44,18 +42,13 @@ public class StorageCredentialCacheKey {
 
   private final Set<String> allowedWriteLocations;
 
-  /**
-   * The callContext is passed to be used to fetch subscoped creds, but is not used to hash/equals
-   * as part of the cache key.
-   */
-  private @Nullable PolarisCallContext callContext;
```
snazy, line 51: I suspect the removal of this field is the key of this PR?

It totally makes sense to not keep per-request state around.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheEntry.java
```diff
@@ -32,12 +32,19 @@ public class StorageCredentialCacheEntry {
   public final EnumMap<StorageAccessProperty, String> credsMap;
 
   private final ScopedCredentialsResult scopedCredentialsResult;
```
snazy, line 34: Unrelated: this field is unused

snazy (commented):
This change looks good to me, and it doesn't change the semantics.

While reviewing, I noticed that the cache key can become quite large, depending on the paths and storage config. Not sure whether there's a better way.

The thing that's not fully clear to me is how long a cached storage-credentials-entry is usable. I see that the cache-entry expiration is calculated in `org.apache.polaris.core.storage.cache.StorageCredentialCacheEntry#getExpirationTime`, but that returns the expiration time of the cached credentials. I wonder how long a _client_ can actually use such credentials?

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
XN137: i have a followup PR-2022 for dealing with this method, generally speaking since `StorageCredentialCache` is application-scoped and multi-realm then this method is no longer needed at all afaict.

its only getting called from a single place `RealmEntityManagerFactory` so idk if deprecation would achieve much but i can add it.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
XN137: yes but also removal of the `realmContext` field is important as well, as it is request scope but was being used inside the expiry logic of the cache.

dimas-b (approved):
LGTM :+1: Thanks, @XN137 !

XN137:
rebased after conflict with 19f44d8e892d80920980d1d03555535b37147923
```
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/PolarisGenericTableCatalogTest.java
runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/PolicyCatalogTest.java
```

PR #2023, opened 2025-07-10
Remove duplicate MetaStoreManagerFactory mocks

XN137:
also rename the field for clarity and consistency

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
@@ -52,10 +52,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.apache.commons.lang3.NotImplementedException;
 import org.apache.iceberg.BaseTable;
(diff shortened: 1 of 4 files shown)


runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
```diff
@@ -450,42 +446,6 @@ protected boolean supportsNotifications() {
...
-
-      @Override
-      public InMemoryEntityCache getOrCreateEntityCache(RealmContext realmContext) {
-        return new InMemoryEntityCache(realmContext, configurationStore, metaStoreManager);
-      }
-
-      @Override
-      public Map<String, PrincipalSecretsResult> bootstrapRealms(
-          Iterable<String> realms, RootCredentialsSet rootCredentialsSet) {
-        throw new NotImplementedException("Bootstrapping realms is not supported");
```
XN137, line 479: was throwing this exception the reason why some tests where using `createMockMetaStoreManagerFactory()` instead of the Mock in the field?

or just left-over code without a particular reason?
sorry in case i am missing something obvious

snazy (approved):

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogTest.java
dimas-b: I do not see any test relying on this exception, so removing this piece of code LGTM.

PR #2026, opened 2025-07-11
Use the 0.9.0 doc from the releases branch

flyrain:


site/hugo.yaml
```diff
@@ -98,19 +98,18 @@ menu:
...
     - name: "1.0.0"
       url: "/releases/1.0.0/"
       parent: "releases"
-    - name: "In Development"
-      url: "/in-dev/unreleased/"
+      weight: 20
+    - name: "0.9.0"
+      url: "/releases/0.9.0/"
       parent: "releases"
+      weight: 30
```
flyrain, line 112: Reorder them so that the 'in-dev' version is always at the top, followed by the latest release, and then earlier versions in descending order.

singhpk234 (approved):

PR #2027, opened 2025-07-11
Use Makefile to simplify setup and commands

MonkeyCanCode:
This PR brings in a root-level Makefile to simplify and centralize how we set up our development environment and run common build tasks for Polaris.

**The Problem I Noticed:**
Right now, getting set up or performing routine builds often means jumping between different documentation pages to find various tool installations (like `helm`, `kubectl`, `jq`, `jenv`, `git`, `docker`) and then copying lengthy `gradlew` commands. While it makes sense to have granular instructions for deep dives or specific contributions, having a single, straightforward entry point for common tasks could really smooth out onboarding for new folks and make our development flow much more efficient.

**My Proposal (and a Working Example):**
I've put together this PR as a first pass, a working example, to kick off a discussion on whether this centralized Makefile approach is something we want to adopt. It shows how we can:

* **Streamline Dependency Setup:** Get all common project dependencies installed with just one `make` command. (We can definitely refine this further, perhaps breaking it down into component-specific sets if we build this out.)
* **Simplify Builds:** Replace those long `gradlew` calls with concise `make` commands to build the server, admin, or both.
* **Automate Helm Tasks:** Easily generate Helm documentation, run unit tests, and perform linting all with simple `make` targets.
* **Add Pre-commit Automation:** Introduce a `pre-commit` target to automate routine tasks like Helm documentation generation and applying Spotless formatting.
    * _Heads up:_ Python codebase linting/checks aren't included yet. I hit some hard-coded path issues there, but that's something I can tackle if we decide to pursue this direction.
* **Minikube Cluster Management:** Streamlining cluster setup, teardown, and interaction specifically for Helm-related development.

**Beyond this PR: Other Potential Use Cases:**
If we move forward with this approach, the Makefile could also become a central point for:
* **Unified Test Execution:** Running all project test cases (unit, integration, etc.) with a single `make` command.
* **Service State Management & Deployment:** Handling local service lifecycle (e.g., `make start`, `make stop`) and orchestrating Helm deployments.

**What I'm Looking For:**
I'd really appreciate your thoughts on this centralized Makefile idea. Does this feel like the right path for how we want to manage our developer experience? Any feedback on the overall concept or the specific examples here would be super helpful as we decide whether to build this out further.

**Mailing discussion:** 
https://lists.apache.org/thread/vjnptg3prdps9slkt7rnvv7ldcwld88s

**Sample interface and commands output:**
Main interface:
```sh
➜  polaris git:(makefile) make

Usage:
  make <target>

General
[...]

PR #2028, opened 2025-07-11
Remove kind

MonkeyCanCode:
This PR remove reference for `kind` and we will be only supporting `minikube` going forward. Details in mailing thread: https://lists.apache.org/thread/cmf1rcpsd7nv7jx79wvbl5pymhltrhg5


run.sh
```diff
@@ -1,67 +0,0 @@
...
-      ;;
-    *)
-      usage
-      ;;
-  esac
-  shift
-done
-
-# Deploy the registry
-echo "Building Kind Registry..."
```
adutra, line 45: Instead of deleting this file, shouldn't we just adapt it to minikube? I think this script can be useful for people trying out Polaris for the first time, wdyt?

run.sh
```diff
@@ -1,67 +0,0 @@
...
-  -Dquarkus.container-image.registry=localhost:5001
-
-echo "Pushing polaris image..."
-docker push localhost:5001/apache/polaris
-
-echo "Loading polaris image to kind..."
-kind load docker-image localhost:5001/apache/polaris:latest
-
-echo "Applying kubernetes manifests..."
-kubectl delete -f k8/deployment.yaml --ignore-not-found
```
adutra, line 66: Also, here we could instead install the Helm chart with a simple deployment using in-memory persistence, wdyt?

run.sh
adutra: Although, we already have the getting-started examples, so I'm fine deleting this script if everybody is OK with that.

MonkeyCanCode: > Instead of deleting this file, shouldn't we just adapt it to minikube? I think this script can be useful for people trying out Polaris for the first time, wdyt?

I am thinking of moving this to makefile instead of multiple individuals bash script. There is a sample PR created for using makefile to manage various things (minikube is one of the planned things to add if we want to proceed with that route)

adutra: > I am thinking of moving this to makefile instead of multiple individuals bash script.

Nice! I'm OK with the general direction.

adutra (approved):

run.sh
MonkeyCanCode: For the reference, here is the sample PR: https://github.com/apache/polaris/pull/2027

PR #2029, opened 2025-07-11
Use PolarisImmutable for StorageCredentialCacheKey

XN137:
note that `entityId` was not part of the `equals`/`hashCode` and thus did not need to be a field

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
@@ -23,21 +23,28 @@ import java.util.Set;
 import org.apache.polaris.core.entity.PolarisEntity;
 import org.apache.polaris.core.entity.PolarisEntityConstants;
 import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
 
 @PolarisImmutable
 public interface StorageCredentialCacheKey {
 
+  @Value.Parameter(order = 1)
   String realmId();
 
+  @Value.Parameter(order = 2)
   long catalogId();
 
+  @Value.Parameter(order = 3)
   @Nullable
   String storageConfigSerializedStr();
 
+  @Value.Parameter(order = 4)
   boolean allowedListAction();
 
+  @Value.Parameter(order = 5)
   Set<String> allowedReadLocations();
 
+  @Value.Parameter(order = 6)
   Set<String> allowedWriteLocations();
 
   static StorageCredentialCacheKey of(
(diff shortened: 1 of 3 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -18,118 +18,45 @@
...
-        + ", entityId="
-        + entityId
-        + ", allowedListAction="
-        + allowedListAction
-        + ", allowedReadLocations="
-        + allowedReadLocations
-        + ", allowedWriteLocations="
-        + allowedWriteLocations
-        + '}';
+    return ImmutableStorageCredentialCacheKey.builder()
```
snazy: Why not delegate to the existing `ImmutableStorageCredentialCacheKey.of(...)` ?

XN137: it didnt exist until i added `@Value.Parameter`, but done now

snazy: Yea - this, rather the whole change, is overall be a good improvement not just for the code, but also the "free" improvements like lazy/memoized hash values.

snazy (approved):

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -18,118 +18,51 @@
...
 
-  /** The serialized string of the storage config. */
-  private final String storageConfigSerializedStr;
+  @Value.Parameter(order = 2)
+  long catalogId();
 
-  /**
-   * The entity id is passed to be used to fetch subscoped creds, but is not used to do hash/equals
-   * as part of the cache key.
-   */
```
dimas-b, line 37: I know that this PR does not change this behaviour, but conceptually if the entity ID is a substantial parameter to computing the cached value, I believe it should be included in the key.

dimas-b (approved):
This PR LGTM "as is" (pure refactoring), but I think we need to follow up on including the the entity ID into the cache key.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
snazy: Not sure whether I already mentioned it, but I'm not sure the `StorageCredentialCache` works as expected, because when credentials are retrieved from the cache, we only know that those are valid at the time of retrieval, but not _how long_ those will be valid (for a client), it could be 0.1ms or 30days.

snazy: (But that's not a problem of this PR)

snazy: (Created #2046)

PR #2031, opened 2025-07-11
Site: Remove non-OSS query engines from front page

danicafine:
I noticed that a non-OSS query engine is included on the front page. It doesn't feel right to include commercial offerings front and center.

As a solution, I suggest that we remove the mention, focus on OSS engines here, and, perhaps, add a different page in the future to include vendors.

### DIFF
--- a/site/content/_index.adoc
+++ b/site/content/_index.adoc
@@ -23,7 +23,7 @@ cascade:
   no_list: true
 ---
 {{< blocks/cover title="Welcome to the Apache Polaris™ (incubating) web site!" image_anchor="center" color="primary" >}}
-Apache Polaris is an open-source, fully-featured catalog for Apache Iceberg™. It implements Iceberg's REST API, enabling seamless multi-engine interoperability across a wide range of platforms, including Apache Doris™, Apache Flink®, Apache Spark™, Dremio®, StarRocks, and Trino.
+Apache Polaris is an open-source, fully-featured catalog for Apache Iceberg™. It implements Iceberg's REST API, enabling seamless multi-engine interoperability across a wide range of platforms, including Apache Doris™, Apache Flink®, Apache Spark™, StarRocks, and Trino.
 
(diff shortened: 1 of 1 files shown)


site/content/_index.adoc
```diff
@@ -23,7 +23,7 @@ cascade:
   no_list: true
 ---
 {{< blocks/cover title="Welcome to the Apache Polaris™ (incubating) web site!" image_anchor="center" color="primary" >}}
-Apache Polaris is an open-source, fully-featured catalog for Apache Iceberg™. It implements Iceberg's REST API, enabling seamless multi-engine interoperability across a wide range of platforms, including Apache Doris™, Apache Flink®, Apache Spark™, Dremio®, StarRocks, and Trino.
```
dimas-b, line 26: Dremio does have a non-commercial OSS offering, with a considerable user base, AFAIK: https://github.com/dremio/dremio-oss

snazy:
@danicafine Can you clarify which of these are not OSS engines? [StarRocks](https://github.com/StarRocks/starrocks) is AL2, [Trino](https://github.com/trinodb/trino/) is AL2, [Dremio](https://github.com/dremio/dremio-oss) is also AL2, the Apache projects naturally, too. Mean, we could link to the GitHub repos, where the license is prominently visible, WDYT?

jbonofre:
Dremio query engine is OSS (Apache license). If the criteria here is only ASF projects, we should remove trino and starrocks then.

jbonofre:
Also, generally speaking, what's the problem mentioning non oss projects/products ? A lot of ASF projects do that.

danicafine:
Totally fine to mention non-OSS projects elsewhere on the site. But I think it's generally a better practice to focus on OSS, especially on the front page.

I'm updating the PR to make it explicit that we're referring to Dremio's OSS offering.

snazy:
@danicafine Within Apache we have a lot of project who mention other OSS and commercial offerings they work with, also very "front and center".

jbonofre (approved):
@danicafine Thanks ! It looks good to me :) Much appreciated !

PR #2036, opened 2025-07-11
Fix gralde command for helm image and remove simple-values.yaml

MonkeyCanCode:
As per https://github.com/apache/polaris/pull/1957, we are now using `1.1.0-incubating-SNAPSHOT` for image tag as oppose to `latest` (the version image is out of box with current workflow is no overwrite provided). However, for all gradle commands we have that are building images, we are implicitly overwriting the image tag name to `postgres-latest` which then break the workflow in the doc (users who follow the doc will get image not found exception).

Another issue fixed by this PR is to remove the `ci/simple-values.yaml` as it is not necessary (the overwrite values file only set image pull policy and set storage secret for AWS...which is causing issue as that is not necessary for a simple standard to some extends). If we do want to keep the `ci/simple-values.yaml`, we will need to ask user to implcitily create secret `polaris-storage` for this workflow to work.

One follow up which is not a blocker for this PR: we should remove image overwrite from all commands? Is there a specific reason on why we would want to keep it as it is standard Quarkus setting.


dimas-b (dismissed):

flyrain:
LGTM! Thanks @MonkeyCanCode.

MonkeyCanCode:
> LGTM! Thanks @MonkeyCanCode.

Thanks for the review. Btw, should we remove all the images overwrite specified in other docs?

dimas-b:
> should we remove all the images overwrite specified in other docs?

As for me, I'm not sure I understand the question, sorry :sweat_smile:

MonkeyCanCode:
> > should we remove all the images overwrite specified in other docs?
> 
> As for me, I'm not sure I understand the question, sorry :sweat_smile: 

So we have "  -Dquarkus.container-image.tag=postgres-latest" specified all over the place (including the current doc as well which got moved as part of this PR). I am wondering if we need to keep it in other docs as it is a standard Quarkus setting.

dimas-b:
I suppose by default Quarkus will use the project version for the docker tag, right? If so, I think it would make sense to remove `-Dquarkus.container-image.tag=postgres-latest` overrides.

MonkeyCanCode:
> I suppose by default Quarkus will use the project version for the docker tag, right? If so, I think it would make sense to remove `-Dquarkus.container-image.tag=postgres-latest` overrides.

That is correct. By default, it will create one with the project version and latest. Sounds good, I will raise a PR later if no concerns from others.

MonkeyCanCode:
@flyrain need another approval here (fixed conflict)

PR #2038, opened 2025-07-12
Use `closer.lua` download script to download Apache Spark distribution

dongjoon-hyun:
We had better use `closer.lua` download script instead of a direct link in order to reduce the flakiness of `archive.apache.org`.

In addition, this complies with Apache INFRA guidelines, too.

- https://infra.apache.org/release-download-pages.html#download-scripts
> Apache project download pages must use a closer.lua script. You'll find below a standard mechanism to let you easily create scripts that comply with the ASF distribution policy.


dongjoon-hyun:
Could you review this PR, @flyrain ?

dongjoon-hyun:
Thank you, @flyrain !

PR #2039, opened 2025-07-12
Add muti-platform Docker images staging in the release guide

jbonofre:


site/content/release-guide.md
```diff
@@ -210,6 +210,15 @@ Next, you have to close the staging repository:
 4. At the top, select "Close" and follow the instructions
 5. In the comment field, use "Apache Polaris x.y.z RCi"
 
+### Build and staging Docker images
+
+You can now publish Docker images on DockerHub:
+
+```
+./gradlew :polaris-server:assemble :polaris-server:quarkusAppPartsBuild --rerun -Dquarkus.container-image.build=true -Dquarkus.container-image.push=true -Dquarkus.docker.buildx.platform="linux/amd64,linux/arm64" -Dquarkus.container-image.tag=x.y.z-rci
```
snazy, line 218: What are the machine's requirements for "buildx"? Docker? Podman?

site/content/release-guide.md
```diff
@@ -239,8 +248,10 @@ The release tarball, signature, and checksums are here:
 
 Helm charts are available on:
 * https://dist.apache.org/repos/dist/dev/incubator/polaris/helm-chart
-NB: the docker images (polaris-server and polaris-admin) will be
-published on DockerHub once release vote passes.
+
+Docker images:
```
snazy, line 252: Since those are multiplatform images, probably worth to add a link to Dockerhub

site/content/release-guide.md
jbonofre: Docker (we use `quarkus-container-image-docker` dependency by default).

site/content/release-guide.md
jbonofre: Good point, I will.

site/content/release-guide.md
snazy: Hm, not sure Docker's a great implied requirement due to the licensing implications that it brings for most devs.

jbonofre: Podman also supports multi-platform images "natively" (without `buildx`). If we want to use podman, we should update the dependency here: https://github.com/apache/polaris/blob/main/runtime/server/build.gradle.kts#L55 (to use podman instead of docker).

adutra: FYI, last time I tried multi-arch builds with podman, the resulting manifests were corrupted. But, maybe it's fixed/improved now.

snazy: I suspect, it's not an issue (license wise) to use Docker buildx in ASF GH workflows. Just for potentially for devs locally.

MonkeyCanCode: Yes. podman can do multi-arch build without buildx now a day (I used it at work as well). Due to the above concern with licensing around docker, does it makes sense to deprecate docker and promo podman instead? Also, for buildx on docker with Mac/Windows, additional steps are needed on the build machines other than the additional parameters specify above.

One more catch with podman is it will add container registry info (by default `docker.io`) to the image name. When building with docker, we will get `apache/polaris:xxxx`. However, if we build with podman, it will then become `docker.io/apache/polaris:xxxx`. In case if we want to promo podman as oppose to docker, we shoild container add additional parameter to set registry info as part of the image build (so both docke/podman will produce same images name) then update our docs accordingly.

jbonofre: I propose to move forward with Docker for now, we can switch to podman in another PR. Thoughts ?

site/content/release-guide.md
jbonofre: I added the link to DockerHub.

jbonofre:
@MonkeyCanCode @snazy @adutra I propose to move forward and merge this PR (to update the release guide). I would like to create another PR for release guide based on this one. Wdyt ?

adutra:
> @MonkeyCanCode @snazy @adutra I propose to move forward and merge this PR (to update the release guide). I would like to create another PR for release guide based on this one. Wdyt ?

LGTM

jbonofre:
Thanks guys ! Much appreciated !

PR #2041, opened 2025-07-13
Fix invalid redirect from public page

MonkeyCanCode:
The "Get Started" page from the public site is now referring to a invalid page:
image
which will cause 404 as following:
image

This is due to the redirect page is https://polaris.apache.org/in-dev/unreleased/getting-started/install-dependencies/ (used to be valid). But with recent page the new url is https://polaris.apache.org/in-dev/getting-started/install-dependencies/ (removed of `unreleased`)


MonkeyCanCode:
@flyrain

eric-maynard (approved):
Nice, confirmed that the new link is valid

MonkeyCanCode:
> Nice, confirmed that the new link is valid

Thanks for the review. Merged to get the page redirect restored.

PR #2043, opened 2025-07-14
Fix CI

snazy:
The `store-gradle-cache` job in the `gradle.yaml` GitHub workflow is missing a "checkout", this change adds it to fix CI.

PR #2044, opened 2025-07-14
Fix CI (no 2)

snazy:
The newly added `store-gradle-cache` CI job has run some Gradle task to trigger Gradle's automatic cache cleanup. In the source project Nessie we used a simple task `showVersion` to do this. As having this task in Polaris might be useful, adding this task as there's no other suitable task (cheap and not generating much output) seems legit.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -238,3 +238,11 @@ changelog {
   )
   version.set(provider { project.version.toString() })
 }
+
+tasks.register("showVersion") {
+  actions.add {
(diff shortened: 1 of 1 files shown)


adutra (approved):

snazy:
Next one after #2043, example failure https://github.com/apache/polaris/actions/runs/16264995485/job/45919531658

Sadly there's no CI for CI, so these kinds of issues are rather "normal".

PR #1593, opened 2025-05-15
Add Docker images SNAPSHOT on Docker HUB (nightly)

jbonofre:
As discussed on the mailing list, this PR extends nightly build to push SNAPSHOT images on Docker HUB.

PR #2045, opened 2025-07-14
Remove postgres_latest as image tag

MonkeyCanCode:
Based from both https://github.com/apache/polaris/pull/2036 and https://github.com/apache/polaris/pull/1957, we will be removing `postgres_latest` as image tag and use latest as image tag instead.


getting-started/eclipselink/docker-compose-bootstrap-db.yml
```diff
@@ -20,7 +20,7 @@
 services:
   polaris-bootstrap:
     # IMPORTANT: the image MUST contain the Postgres JDBC driver and EclipseLink dependencies, see README for instructions
-    image: apache/polaris-admin-tool:postgres-latest
+    image: apache/polaris-admin-tool:1.1.0-incubating-SNAPSHOT
```
adutra: Shouldn't we use `latest` here? Using a specific version here means that we'd need to update it during or immediately after the release process.

MonkeyCanCode: Yes, I was thinking about the same thing. The current build will do both version and latest. It makes sense to use the latest instead for documentation. I will update it later today.

snazy:
Wait a sec ...

snazy:
Nope, all good. I see you removed the `-SNAPSHOT` reference.

For posterity: as an Apache project we _have to_ direct users to _official releases_, never to snapshots, nightlies and similar.
Since you're referring to the Docker repo for official releases, that's fine.

MonkeyCanCode:
> Nope, all good. I see you removed the `-SNAPSHOT` reference.
> 
> For posterity: as an Apache project we _have to_ direct users to _official releases_, never to snapshots, nightlies and similar.
> Since you're referring to the Docker repo for official releases, that's fine.

Oh, that is good to know. In that case, we will need to change the value files in the helm right? That is referring to the snapshot one @adutra

adutra:
> Oh, that is good to know. In that case, we will need to change the value files in the helm right? That is referring to the snapshot one @adutra

I guess you are right... we should use `tag: latest` instead of `tag: "1.1.0-incubating-SNAPSHOT"`.

I don't know why #1957 included a `-SNAPSHOT` tag in the chart... oddly enough, the tag exists, but it doesn't align with our recent discussions about nightly images in #1593. @flyrain do you remember the rationale? I think we need to revisit this, imho the safest choice is to go with `latest`.

adutra:
OK I finally found the discussion I had with @flyrain about image tags:

https://github.com/apache/polaris/pull/1944#discussion_r2167500715

Yufei was advocating for specific tags in case users want to downgrade the helm chart.

I confess I'm now thinking that it's too much hassle keeping specific Docker image versions in the Helm chart compared to the (limited) benefits. Besides, if we move the Helm chart out of this repo, as it was proposed, then it probably doesn't make sense at all to use specific versions in `image.tag`.

helm/polaris/values.yaml
```diff
@@ -27,7 +27,7 @@ image:
   # -- The image pull policy.
   pullPolicy: IfNotPresent
   # -- The image tag.
-  tag: "1.1.0-incubating-SNAPSHOT"
+  tag: "latest"
```
adutra, line 30: I'm fine with this change, but let's wait for @flyrain to chime in.

adutra:
@flyrain please have a look whenever you have time.

helm/polaris/values.yaml
flyrain: +1 for the change, since we are moving helm chart out of the main repo.

PR #2050, opened 2025-07-15
Unify RealmEntityManagerFactory usage in tests

XN137:
since all ctor params are created in `IcebergCatalogTest.before` we can do the same for `RealmEntityManagerFactory`

`PolarisAuthzTestBase.entityManager` was already getting derived from `realmEntityManagerFactory` here:
https://github.com/apache/polaris/blob/2c2052c28f899aaa85e5f11a9131d9812ec62679/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/PolarisAuthzTestBase.java#L247


runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/IcebergCatalogHandlerAuthzTest.java
```diff
@@ -1791,12 +1788,7 @@ public void testSendNotificationSufficientPrivileges() {
 
     PolarisCallContextCatalogFactory factory =
         new PolarisCallContextCatalogFactory(
-            new RealmEntityManagerFactory(null, null, null) {
-              @Override
-              public PolarisEntityManager getOrCreateEntityManager(RealmContext realmContext) {
-                return entityManager;
-              }
-            },
+            realmEntityManagerFactory,
```
XN137, line 1791: note that `realmEntityManagerFactory` was already being used 4 lines below this one

eric-maynard (approved):

PR #2051, opened 2025-07-15
Site/dev: allow overriding the podman/docker binaries detection

snazy:
The scripts in the `bin/` directory are built to work with both Docker and podman. There are nuances in how both behave, especially wrt docker/podman-compose. Some local environment specifics require the use of `podman-compose`, others the use of `docker-compose`. The default behavior is to prefer the `podman` and `podman-compose` binaries, if those exist and fall back to `docker` and `docker-compose`. Some setups using podman however require the use of `docker-compose` even if `podman-compose` is installed. This may manifest in an error message stating that `--userns` and `--pod` cannot be used together. In that case create a file `.user-settings` in the `site/` folder and add these two lines:
```bash
DOCKER=docker
COMPOSE=docker-compose
```

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -35,6 +35,21 @@ Then point your browser to http://localhost:1313/
 Then edit files from your IDE - you will see the changes in your browser. When changing the site structure,
 it is often necessary to restart Hugo - just run `site/bin/run-hugo-in-docker.sh` again. 
 
+## Dealing with `--userns and --pod cannot be used together` errors
+
+The scripts in the `bin/` directory are built to work with both Docker and podman. There are nuances in how
+both behave, especially wrt docker/podman-compose. Some local environment specifics require the use of
+`podman-compose`, others the use of `docker-compose`. The default behavior is to prefer the `podman` and
+`podman-compose` binaries, if those exist and fall back to `docker` and `docker-compose`. Some setups using
+podman however require the use of `docker-compose` even if `podman-compose` is installed. This may manifest
+in an error message stating that `--userns` and `--pod` cannot be used together. In that case create a file
+`.user-settings` in the `site/` folder and add these two lines:
+```bash
+DOCKER=docker
+COMPOSE=docker-compose
+```
+
+
 # Building the static site
 
 ## Building a static site - served at http://localhost:8080/
--- a/site/bin/_hugo-docker-include.sh
+++ b/site/bin/_hugo-docker-include.sh
@@ -17,8 +17,10 @@
 # under the License.
 #
 
-DOCKER="$(which podman > /dev/null && echo podman || echo docker)"
-COMPOSE="$(which podman-compose > /dev/null && echo podman-compose || echo docker-compose)"
+[[ -f site/.user-settings ]] && . site/.user-settings
+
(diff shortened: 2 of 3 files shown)


adutra (approved):
Changes LGTM.

One general remark: since compose v2, `docker-compose` is just an alias of `docker compose`. The same seems to apply to `podman-compose` vs `podman compose`. So maybe we can get rid of the `COMPOSE` env var completely?

snazy:
> One general remark: since compose v2, `docker-compose` is just an alias of `docker compose`. The same seems to apply to `podman-compose` vs `podman compose`. So maybe we can get rid of the `COMPOSE` env var completely?

I don't really know what has changed. I reinstalled all the podman stuff to stick with what Ubuntu provides (had to newer podman versions before, which where not distributed by Ubuntu at that time). Sadly, the behavior seems to depend on whether you run `podman-compose` or `docker-compose`.

snazy:
Those are technically aliases in envs but not all, but :shrug:

site/README.md
```diff
@@ -35,6 +35,21 @@ Then point your browser to http://localhost:1313/
...
+The scripts in the `bin/` directory are built to work with both Docker and podman. There are nuances in how
+both behave, especially wrt docker/podman-compose. Some local environment specifics require the use of
+`podman-compose`, others the use of `docker-compose`. The default behavior is to prefer the `podman` and
+`podman-compose` binaries, if those exist and fall back to `docker` and `docker-compose`. Some setups using
+podman however require the use of `docker-compose` even if `podman-compose` is installed. This may manifest
+in an error message stating that `--userns` and `--pod` cannot be used together. In that case create a file
+`.user-settings` in the `site/` folder and add these two lines:
+```bash
+DOCKER=docker
+COMPOSE=docker-compose
```
eric-maynard, line 49: nit: I wonder if a var like `DOCKER_COMMAND` or `DOCKER_ALIAS` or something might be less likely to collide with another env var.

eric-maynard (approved):

site/bin/_hugo-docker-include.sh
```diff
@@ -17,8 +17,10 @@
 # under the License.
 #
 
-DOCKER="$(which podman > /dev/null && echo podman || echo docker)"
-COMPOSE="$(which podman-compose > /dev/null && echo podman-compose || echo docker-compose)"
+[[ -f site/.user-settings ]] && . site/.user-settings
+
+[[ -z ${DOCKER} ]] && DOCKER="$(which podman > /dev/null && echo podman || echo docker)"
```
MonkeyCanCode, line 22: nit: maybe use more compact version such as following:
```
DOCKER="${DOCKER:-$(which podman >/dev/null && echo podman || echo docker)}"
COMPOSE="${COMPOSE:-$(which podman-compose >/dev/null && echo podman-compose || echo docker-compose)}"
```

site/README.md
snazy: Yea, I've been thinking about environment variables.
But then people have to declare those elsewhere, which is also not great.

snazy: OTOH, it's an internal "hack", we can change it anytime.

PR #1891, opened 2025-06-13
Enable patch version updates for maintained Polaris version

snazy:
Polaris 1.x will be a supported/maintained release. It is crucial to apply bug and security fixes to such releases.

Therefore, this change enables patch-version updates for Polaris 1.*

PR #2052, opened 2025-07-15
Attempt to make Renovate work again

snazy:
Looks that I accidentally broke Renovate with #1891. This was made under the impression of the [Renovate change to support `baseBranches` in forking-renovate] (https://github.com/renovatebot/renovate/pull/36054). However, a [later Renovate change](https://github.com/renovatebot/renovate/pull/35579)  seems to break that.

The plan here is to:
1. remove the regex from our `baseBranches` option - if that doesn't work then
2. just use the default branch

Relevant [renovate issue](https://github.com/renovatebot/renovate/issues/28700#issuecomment-3073312040)

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -30,9 +30,10 @@
 
   // Added for posterity how to let Renovate manage version-branches, assuming that release branches
   // have the `release/` prefix.
+  // BE CAREFUL WITH THIS OPTION and make sure that the changes work.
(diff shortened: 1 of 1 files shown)


jbonofre:
In order to unblock renovate (to see if this change helps), I'm merging this PR even if it's short notice/review, I'm kindly asking to allow me to "break the good practices" this time to unblock renovate 😄

PR #2085, opened 2025-07-15
Disable renovatebot on release branches

eric-maynard:
Per the mailing list thread "[DISCUSS] Disable renovatebot on release branches", we should not do automatic dependency upgrades for release branches. Since it seems `release/1.0.x` is a release branch, we can remove this regex from renovate's list.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -79,13 +79,6 @@
         "io.quarkus:io.quarkus.gradle.plugin",
       ],
     },
-
-    // Turn off major & minor version updates on version-branches
-    {
-      matchBaseBranches: ["/^release/.*/"],
-      matchUpdateTypes: ["major", "minor"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -33,7 +33,6 @@
   // BE CAREFUL WITH THIS OPTION and make sure that the changes work.
   baseBranches: [
     "main",
-    "release/1.0.x",
   ],
   additionalBranchPrefix: "{{baseBranch}}/",
   commitMessagePrefix: "{{baseBranch}}: ",
```
eric-maynard, line 38: Currently it looks like we have `main` twice in the PR names, so separately from this PR we may also want to remove the prefix here

eric-maynard: Looks like #2085 will fix this

Issue #1327, opened 2025-04-06
Document Polaris Integration with Third-Party Identity Providers (IDPs)


flyrain:
**We need to document how Polaris integrates with third-party Identity Providers (IDPs)** to support authentication and authorization workflows.

#### Integration Overview  
- High-level flow of how Polaris delegates authentication to external IDPs  
- Supported protocols (e.g., OIDC, SAML)

#### Setup Instructions  
- Step-by-step guide for integrating with common IDPs (e.g., **Okta**, **Auth0**, **Azure AD**)  
- How to **register Polaris as a client application** with the IDP  
- Required configuration fields (e.g., client ID, secret, redirect URIs)  
- Polaris-side configuration (env variables, config files, etc.)

#### Potential Code Changes  
- Highlight any parts of Polaris that might require customization or extension for integration  
- Where hooks or plugins might be inserted for custom logic

#### Identifier Sync  
- Outline options for syncing user identifiers or roles  
   - Manual user provisioning  
   - Automatic sync (e.g., via SCIM or IDP claims)  
- Best practices for mapping IDP groups/roles to Polaris permissions

adutra:
This feature is currently not available for Polaris at all, so it's not just a matter of documentation, it's a matter of _implementing_ the feature.

The only way today would be to implement a custom `Authenticator` and `TokenBroker` – but with just Polaris OSS code, it's impossible. And implementing a proper authentication layer with OAuth2 token introspection, JWT claim mapping and all that jazz, is not really a trivial task.

This issue is possibly a duplicate of:

https://github.com/apache/polaris/issues/336

Also related:

https://github.com/apache/polaris/issues/12
https://github.com/apache/polaris/issues/441

flyrain:
It's not only a doc change for sure as complete solution. Given this is highly demanded feature, I think it's still valuable to document workarounds before #441 was implemented. So that users can try to integrate by themselves.

adutra:
This has been partially addressed by #2013, although that PR does not cover how to setup the IDP for Polaris (users, groups, roles, etc.) or how to integrate with common IDPs (Okta, Azure AD, etc.).

@flyrain is it OK to close this, or do you think we should improve the current docs with the missing items?

cc @markhoerth

flyrain:
Closed this as #2013 fixed it.

adutra:
FYI I'm adding a getting-started example: #2244

PR #2104, opened 2025-07-15
feat(helm): Add support for external authentication

adutra:


helm/polaris/templates/_helpers.tpl
```diff
@@ -150,18 +150,18 @@ line breaks, they will be escaped and a multi-line option will be printed.
...
 {{- end -}}
 
 {{/*
-Convert a dict into a string formed by a comma-separated list of key-value pairs: key1=value1,key2=value2, ...
+Escapes a property key to be used in a configmap, conforming with the Java parsisng rules for
+property files: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Properties.html#load(java.io.Reader)
+- Escapes all backslashes.
+- Escapes all key termination charaters: '=', ':' and whitespace.
 */}}
-{{- define "polaris.dictToString" -}}
```
adutra, line 159: The template `polaris.dictToString` was unused.

helm/polaris/templates/_helpers.tpl
```diff
@@ -317,3 +300,74 @@ ports:
...
+{{- $_ = set $map (printf "%s.active-roles-provider.type" $prefix) (dig "activeRolesProvider" "type" "default" $auth) -}}
+{{- $_ = set $map (printf "%s.active-roles-provider.type" $prefix) (dig "activeRolesProvider" "type" "default" $auth) -}}
+{{- if (or (eq $authType "mixed") (eq $authType "internal")) -}}
+{{- $tokenBrokerType := dig "tokenBroker" "type" "rsa-key-pair" $auth -}}
+{{- $_ = set $map (printf "%s.token-service.type" $prefix) (dig "tokenService" "type" "default" $auth) -}}
+{{- $_ = set $map (printf "%s.token-broker.type" $prefix) $tokenBrokerType -}}
+{{- $_ = set $map (printf "%s.token-broker.max-token-generation" $prefix) (dig "tokenBroker" "maxTokenGeneration" "PT1H" $auth) -}}
+{{- $secretName := dig "tokenBroker" "secret" "name" "" $auth -}}
+{{- if $secretName -}}
+{{- $subpath := empty $realm | ternary "" (printf "%s/" (urlquery $realm)) -}}
```
adutra, line 328: This is a best-effort to produce valid filesystem paths if the realm ID contains weird characters. The only function available in Helm that vaguely serves this purpose is `urlquery`.

adutra: Another option is to output the hash the realm ID. It's safer, but less readable. Lmk if that's preferable.

helm/polaris/templates/_helpers.tpl
```diff
@@ -150,18 +150,18 @@ line breaks, they will be escaped and a multi-line option will be printed.
 {{- $valAsString = $valAsString | nindent 4 | replace "\n" "\\\n" -}}
 {{- end -}}
 {{- end -}}
-{{ print $key "=" $valAsString }}
+{{ print (include "polaris.escapeConfigOptionKey" $key) "=" $valAsString }}
 {{- end -}}
 
 {{/*
-Convert a dict into a string formed by a comma-separated list of key-value pairs: key1=value1,key2=value2, ...
+Escapes a property key to be used in a configmap, conforming with the Java parsisng rules for
```
adutra, line 157: We were missing until now a proper escape mechanism for config option keys. Before this change, any realm containing a key termination character such as a space or `=`, would produce invalid configuration.

helm/polaris/templates/_helpers.tpl
dimas-b: `urlquery` LGTM :+1:

helm/polaris/values.yaml
```diff
@@ -560,15 +560,19 @@ storage:
...
-  # test is not recommended for production.
+  # -- The `ActiveRolesProvider` implementation to use. Only one built-in type is supported: default.
+  activeRolesProvider:
+    type: default
+  # -- The token service (`IcebergRestOAuth2ApiService`) implementation to use. Two built-in types are supported: default and disabled.
+  # The disabled type is always used when `authentication.type` is set to external.
   tokenService:
     type: default
-  # -- The type of token broker to use. Two built-in types are supported: rsa-key-pair and symmetric-key.
+  # -- The `TokenBroker` implementation to use. Two built-in types are supported: rsa-key-pair and symmetric-key.
```
dimas-b, line 575: nit: `TokenBroker` is _not_ applicable to the "external" auth type, right? Should we support disabling it?

helm/polaris/values.yaml
```diff
@@ -596,7 +600,63 @@ authentication:
...
+# configuration using the `advancedConfig` section.
+oidc:
+  # -- The authentication server URL. Must be provided if at least one realm is configured for external
+  # authentication.
+  authServeUrl: ~  # https://auth.example.com/realms/polaris
+  # -- The client to use when authenticating with the authentication server.
+  client:
+    # -- The client ID to use when authenticating with the authentication server.
+    id: polaris
+    # -- The secret to pull the client secret from.
```
dimas-b: nit: the OIDC secret is optional, right?

adutra: Yes. If `secret.name` is nil, no client secret is included in the configuration. This follows a pattern that is already used for other secrets.

helm/polaris/values.yaml
adutra: `tokenBroker` and `tokenService` are indeed only relevant when using internal (or mixed) authentication.

Setting these options when using external auth doesn't hurt, but in practice, the configmap won't even contain those config options if the authentication type is `external`.

helm/polaris/values.yaml
dimas-b: Do you think we could mention that in the doc comment for clarity? Current text feels like the secret might be required :thinking:

helm/polaris/values.yaml
dimas-b: ah, thx for the explanation!.. I missed that behaviour :facepalm:

helm/polaris/values.yaml
adutra: Done. I seized the opportunity to clarify `tokenBroker` and `tokenService` as well.

dimas-b (dismissed):
LGTM :+1:

helm/polaris/README.md
```diff
@@ -189,21 +189,24 @@ ct install --namespace polaris --charts ./helm/polaris
 |-----|------|---------|-------------|
 | advancedConfig | object | `{}` | Advanced configuration. You can pass here any valid Polaris or Quarkus configuration property. Any property that is defined here takes precedence over all the other configuration values generated by this chart. Properties can be passed "flattened" or as nested YAML objects (see examples below). Note: values should be strings; avoid using numbers, booleans, or other types. |
```
MonkeyCanCode, line 190: Can we copy this to `site/content/in-dev/unreleased/helm.md` as well?

MonkeyCanCode (dismissed):

helm/polaris/tests/configmap_test.yaml
```diff
@@ -336,3 +427,33 @@ tests:
     asserts:
       - matchRegex: { path: 'data["application.properties"]', pattern: "polaris.tasks.max-concurrent-tasks=10" }
       - matchRegex: { path: 'data["application.properties"]', pattern: "polaris.tasks.max-queued-tasks=20" }
+
+  - it: should configure OIDC
```
MonkeyCanCode, line 431: Probably not in the scope of this PR. But we should add the doc for OIDC https://github.com/apache/polaris/issues/1327 and have some docker compose with it (e.g. keycloak?)

helm/polaris/README.md
adutra: @MonkeyCanCode isn't this copy done automatically? Or should I make a manual copy-paste?

helm/polaris/tests/configmap_test.yaml
adutra: We already have some docs for OIDC / external IDP: 

https://github.com/apache/polaris/blob/45040bd5b80a6ab3dc5dc343182e8a6435d1e709/site/content/in-dev/unreleased/external-idp.md

helm/polaris/README.md
adutra: Okay I copied the file manually. We should look into automate this.

MonkeyCanCode: Yes, the automation is in the Makefile which is still under review: https://github.com/apache/polaris/pull/2027 (make helm-doc-generate)

PR #2105, opened 2025-07-15
Add TestFileIOFactory helper

XN137:
also cleanup warnings in `TableCleanupTaskHandlerTest`

### DIFF
--- a/service/common/src/main/java/org/apache/polaris/service/catalog/io/FileIOFactory.java
+++ b/service/common/src/main/java/org/apache/polaris/service/catalog/io/FileIOFactory.java
@@ -58,4 +58,14 @@ public interface FileIOFactory {
       @Nonnull Set<String> tableLocations,
       @Nonnull Set<PolarisStorageActions> storageActions,
       @Nonnull PolarisResolvedPathWrapper resolvedEntityPath);
+
+  static FileIOFactory wrapExisting(FileIO fileIO) {
+    return (callContext,
+        ioImplClassName,
+        properties,
+        identifier,
+        tableLocations,
+        storageActions,
+        resolvedEntityPath) -> fileIO;
+  }
 }
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/task/BatchFileCleanupTaskHandlerTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/task/BatchFileCleanupTaskHandlerTest.java
@@ -23,13 +23,11 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatPredicate;
 
 import io.quarkus.test.junit.QuarkusTest;
-import jakarta.annotation.Nonnull;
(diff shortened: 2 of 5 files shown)


service/common/src/main/java/org/apache/polaris/service/catalog/io/FileIOFactory.java
```diff
@@ -58,4 +58,14 @@ FileIO loadFileIO(
       @Nonnull Set<String> tableLocations,
       @Nonnull Set<PolarisStorageActions> storageActions,
       @Nonnull PolarisResolvedPathWrapper resolvedEntityPath);
+
+  static FileIOFactory wrapExisting(FileIO fileIO) {
```
eric-maynard: We used to have a `TestFileIOFactory`, I wonder if this would be a good time to reintroduce that? The method looks to be used only for tests.

XN137: i would argue this is shorter and more discoverable, also while its true that its only used from tests right now, the logic is still generally correct outside of tests.

if you insist, i can still move it to new `TestFileIOFactory` (any specific module/package) ?

eric-maynard: I would recommend the `org.apache.polaris.service` package in `testFixtures` alongside `TestServices`.

The reason that I am a bit unsure about putting it outside of test code is that we did in fact design FileIOFactory with a sort of a sort of "wrapper" pattern in mind (see `WasbTranslatingFileIOFactory` for an example) and I worry that having introducing a method like will have implications on that pattern or on how maintainers perceive that pattern is meant to be implemented.

XN137: just to double check: are you aware the method is static? i.e. it will never become relevant to implementers of the interface afaict

eric-maynard: Yes, I see that it's static (and therefore `final`). I think this would function well as a static method or a constructor in `TestFileIOFactory`.

XN137: > I worry that having introducing a method like will have implications on that pattern or on how maintainers perceive that pattern is meant to be implemented

my point was that since the method is static this worry is unjustified, the "pattern" is not affected at all by the static method in the interface.
that being said, i have now done as you had requested.

however taking a look at it now, not sure if the cleanup is really worth it as it adds a new class that isnt that easy to discover and its no longer a clear win in terms of lines of code, wdyt?

eric-maynard: It's up to you, I personally still think the cleanup is a good idea as it should make it easier to write new tests. I'm more concerned about lowering the cognitive load needed to set up a new test than about lowering LOC.

dimas-b (approved):

PR #2106, opened 2025-07-15
Abstract polaris-runtime-service tests for all persistence implementations

snazy:
The NoSQL persistence implementation has to run the Iceberg table & view catalog plus the Polaris specific tests as well. Reusing existing tests is beneficial to avoid a lot of code duplcation.

This change moves the actual tests to `Abstract*` classes and refactors the existing tests to extend those. The NoSQL persistence work extends the same `Abstract*` classes but runs with different Quarkus test profiles.

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractIcebergCatalogTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractIcebergCatalogTest.java
@@ -34,8 +34,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
 import io.quarkus.test.junit.QuarkusMock;
-import io.quarkus.test.junit.QuarkusTestProfile;
-import io.quarkus.test.junit.TestProfile;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import jakarta.inject.Inject;
@@ -113,7 +111,7 @@ import org.apache.polaris.core.persistence.MetaStoreManagerFactory;
 import org.apache.polaris.core.persistence.PolarisEntityManager;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.PolarisResolvedPathWrapper;
-import org.apache.polaris.core.persistence.cache.InMemoryEntityCache;
+import org.apache.polaris.core.persistence.cache.EntityCache;
 import org.apache.polaris.core.persistence.dao.entity.BaseResult;
 import org.apache.polaris.core.persistence.dao.entity.EntityResult;
 import org.apache.polaris.core.persistence.pagination.Page;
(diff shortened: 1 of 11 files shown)


runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractIcebergCatalogTest.java
```diff
@@ -298,6 +288,11 @@ public void before(TestInfo testInfo) {
             storageCredentialCache,
             createEntityCache(polarisContext.getRealmConfig(), metaStoreManager));
 
+    // Damn ThreadLocal :rage:
```
adam-christian-software: Probably remove this.

adam-christian-software (approved):
Overall, looks good.

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractIcebergCatalogTest.java
snazy: oops, personal opinion

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractPolicyCatalogTest.java
```diff
@@ -108,24 +103,7 @@
...
-          "true",
-          "polaris.features.\"SUPPORTED_CATALOG_STORAGE_TYPES\"",
-          "[\"FILE\"]",
-          "polaris.features.\"ALLOW_INSECURE_STORAGE_TYPES\"",
-          "true",
-          "polaris.readiness.ignore-severe-issues",
-          "true");
-    }
-  }
+public abstract class AbstractPolicyCatalogTest {
```
eric-maynard, line 106: The pattern we follow in other tests seems to be `...Base` such as in `BootstrapCommandTestBase`, perhaps we'd better follow that convention rather than introduce a new `Abstract...` one.

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractIcebergCatalogTest.java
```diff
@@ -301,6 +289,10 @@ public void before(TestInfo testInfo) {
             storageCredentialCache,
             createEntityCache(polarisContext.getRealmConfig(), metaStoreManager));
 
+    // LocalPolarisMetaStoreManagerFactory.bootstrapServiceAndCreatePolarisPrincipalForRealm sets
+    // the CallContext.setCurrentContext() but never clears it, whereas the NoSQL one resets it.
```
eric-maynard, line 293: There is no NoSQL one right now, so perhaps better not to mention it in a comment here

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractPolicyCatalogTest.java
snazy: Not sure we have a commen nomenclature. There are two `Base...Test` and one `...BaseTest`.

runtime/service/src/test/java/org/apache/polaris/service/quarkus/catalog/AbstractIcebergCatalogTest.java
snazy: It just explains the reason for this.

PR #1912, opened 2025-06-19
Add support for catalog federation in the CLI

eric-maynard:
The CLI currently only supports the version of EXTERNAL catalogs that was present in 0.9.0. Now, EXTERNAL catalogs can be configured with various configurations relating to federation. This PR updates the CLI to better match the REST API so that federated catalogs can be easily set up in the CLI.

PR #2121, opened 2025-07-15
Add IMPLICIT authentication support to the CLI

poojanilangekar:
PRs #1925 and #1912 were merged around the same time.  This PR connects the two changes and enables the CLI to accept IMPLICIT authentication type. 

Since Hadoop federated catalogs rely purely on IMPLICIT authentication, the CLI parsing test has been updated to reflect the same.

### DIFF
--- a/client/python/test/test_cli_parsing.py
+++ b/client/python/test/test_cli_parsing.py
@@ -507,19 +507,6 @@ class TestCliParsing(unittest.TestCase):
         check_arguments(
             mock_execute(['catalogs', 'create', 'my-catalog', '--type', 'external',
                           '--storage-type', 'gcs', '--default-base-location', 'dbl',
-                          '--catalog-connection-type', 'hadoop', '--hadoop-warehouse', 'h',
-                          '--catalog-uri', 'u', '--catalog-authentication-type', 'bearer',
-                          '--catalog-bearer-token', 'b']),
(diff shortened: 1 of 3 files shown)


poojanilangekar:
CC @eric-maynard

eric-maynard (dismissed):

client/python/test/test_cli_parsing.py
```diff
@@ -530,24 +517,6 @@ def get(obj, arg_string):
                 (0, 'catalog.connection_config_info.remote_catalog_name'): 'i',
                 (0, 'catalog.connection_config_info.uri'): 'u',
             })
-        check_arguments(
```
eric-maynard, line 533: Instead of removing these tests, can you change them to use a different connection type?

poojanilangekar: Should I change it to `iceberg-rest`? (That would be repeating the tests in line520:532 and line551:585).

eric-maynard: Looks like the test on line 507 is indeed a duplicate of the one of 520 so that one can go, but which is the duplicate of the line 551 test? That should be the only OAUTH one.

poojanilangekar: Done.

eric-maynard (approved):

PR #2127, opened 2025-07-17
Python client: add support for endpoint, sts-endpoint, path-style-access

snazy:
Amends #1913 and #2012


adutra (dismissed):

client/python/cli/command/catalogs.py
```diff
@@ -64,6 +64,9 @@ class CatalogsCommand(Command):
     remove_properties: List[str]
     hadoop_warehouse: str
     iceberg_remote_catalog_name: str
+    endpoint: str
```
eric-maynard, line 67: We'll also need to amend `_has_aws_storage_info` to check for these new members

client/python/cli/constants.py
```diff
@@ -222,6 +225,11 @@ class Create:
             EXTERNAL_ID = "(Only for S3) The external ID to use when connecting to S3"
             REGION = "(Only for S3) The region to use when connecting to S3"
             USER_ARN = "(Only for S3) A user ARN to use when connecting to S3"
+            ENDPOINT = "(Only for S3) The S3 endpoint to use when connecting to S3"
+            STS_ENDPOINT = (
+                "(Only for S3) The STS endpoint to use when connecting to STS"
+            )
+            PATH_STYLE_ACCESS = "(Only for S3) whether to use path-style-access for S3"
```
eric-maynard:
```suggestion
            PATH_STYLE_ACCESS = "(Only for S3) Whether to use path-style-access for S3"
```

client/python/cli/command/catalogs.py
MonkeyCanCode: Here is the current implementation: 
```
    def _has_aws_storage_info(self):
        return self.role_arn or self.external_id or self.user_arn or self.region
```

which check the essential stuff. Those 3 are optionals I think, thus we don't need them here right?

eric-maynard: The original intent was to make sure that you're not mixing e.g. azure and aws options, e.g. `endpoint` and `tenant_id`. To that end, it might be good to add these here.

eric-maynard: (region is also optional and was added recently)

snazy: Thx!
Added

eric-maynard (approved):

Issue #844, opened 2025-01-22
Add Amazon DynamoDB as persistence backend


jackye1995:
Similar to #650, try using Amazon DynamoDB as a NoSQL persistence backend to see if there is anything that needs to be changed to make Polaris work with it.

jackye1995:
Read through https://github.com/apache/polaris/issues/775, https://github.com/apache/polaris/issues/766, https://github.com/apache/polaris/issues/763, looks like it is a major blocker and won't be solved without big refactoring.

I think it is probably better to pursue a JDBC + serializable isolation approach that #649 is doing

dimas-b:
Related `dev` discussion: https://lists.apache.org/thread/obvrrh0w36f87rwljptzwonpt89qkcm1

singhpk234:
I am not sure if there i an active work in this stream since we are targeting 14th Oct probably better to move this to ~ 1.4

dimas-b:
Unless someone is actively working on this, I'd propose to unassign the milestone for now, and assign it again when we have a PR. Polaris is following the release train model, so any contributions that merge, get released in the next "train". Setting milestones is not critical unless a PR is almost ready.

BACtaki:
Hi @dimas-b @snazy ! I'd like to work on this if the issue is still open. 

I see that Robert implemented the MongoDB backend [here](https://github.com/apache/polaris/pull/2992/files). I'm thinking the DynamoDB backend would be in a similar vein? The idea behind this issue is to offer a nosql AWS-native persistence option, correct?

dimas-b:
@BACtaki : Yes, supporting DynamoDB under the NoSQL framework would be quite natural... I wonder, though, if @snazy might have some code for that already 🤔 if not, a contribution from your side would be very welcome 👍

BACtaki:
Sounds good! I'll wait to hear from @snazy in that case. Would love to get this started or contribute to an existing effort if one exists already.

snazy:
In fact I do have the code for a couple more database backends, just haven't pushed it. It's all in the most recent commit in [this branch](https://github.com/snazy/polaris/tree/persistence-nosql-all-backends), which is based on the #1189 branch.

The backends I mentioned are not fully wired up the "whole way" via `:polaris-persistence-nosql-cdi-quarkus` to `:polaris-runtime[-service]`. The unit test suites for these backends are passing and the heavier correctness-tests are also passing against single-node instances of the "real" implementations.

[...]

Issue #650, opened 2025-01-10
Add MongoDB persistence backend


jbonofre:

snazy:
I've been looking through the Polaris code base on how this effort is achievable and the result is not satisfying.
There are a bunch of issues, for example #775, #766, #763, that currently prevent this effort. I suspect there is a lot of work to be done before this effort can actually be started.

adam-christian-software:
@snazy & I are starting on this work now as we have a proposal and had a community meeting:

1. Proposal: https://docs.google.com/document/d/1POUWe0xMZOBoaJ6Rgiw35ziEoc6OEYCiW7Zk6bR9H6M/edit?tab=t.0
2. Initial implementation PR: https://github.com/apache/polaris/pull/1189 
3. Community meeting recording: https://drive.google.com/file/d/1XN1ng8ORB3-jCAeWmCt0vOn3zyZmvRWh/view?usp=sharing
4. Community meeting presentation: https://docs.google.com/presentation/d/1lX2EdvM0SeyuOdO_u1idlWfmnlH3hFE16JEyWo45Bdo/edit?slide=id.p1#slide=id.p1

dimas-b:
The bulk of MongoDB persistence got finalized by #3396.

Remaining maintenance tasks are provided by #3395.

PR #2131, opened 2025-07-18
feat(idgen): Start Implementation of NoSQL with the ID Generation Framework

adam-christian-software:
# Motivation
This implementation starts the series of PRs to implement the NoSQL work presented in https://docs.google.com/document/d/1POUWe0xMZOBoaJ6Rgiw35ziEoc6OEYCiW7Zk6bR9H6M/edit?tab=t.0#heading=h.nx9vzhg2x8v2 with the first implementation [here](https://github.com/apache/polaris/pull/1189).

# Description
This PR is dedicated to creating the ID Generation Framework.

Related to #[650](https://github.com/apache/polaris/issues/650) & #[844](https://github.com/apache/polaris/issues/844)

### DIFF
--- a/persistence/nosql/idgen/spi/build.gradle.kts
+++ b/persistence/nosql/idgen/spi/build.gradle.kts
@@ -36,7 +36,6 @@ dependencies {
   compileOnly(platform(libs.quarkus.bom))
   compileOnly("io.quarkus:quarkus-core")
 
-  compileOnly(project(":polaris-immutables"))
   annotationProcessor(project(":polaris-immutables", configuration = "processor"))
 
   implementation(platform(libs.jackson.bom))


persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+
+  default String describeId(long id) {
+    return Long.toString(id);
+  }
+
+  IdGenerator NONE =
+      new IdGenerator() {
+        @Override
+        public long generateId() {
+          throw new UnsupportedOperationException("NONE IdGenerator");
```
dimas-b: nit: sometimes exception messages are separated from class names. Having a more descriptive message would be nice. Suggestion: `NONE IdGenerator cannot actually generate IDs`

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/MonotonicClock.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  Instant currentInstant();
+
+  /** Monotonically increasing timestamp with nanosecond precision, not related to wall clock. */
+  long nanoTime();
+
+  void sleepMillis(long millis);
+
+  @Override
+  void close();
```
dimas-b, line 66: why override? We do not add anything new here :thinking: 

Note: it can still throw `RuntimeException`.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+package org.apache.polaris.ids.api;
+
+import jakarta.annotation.Nonnull;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.util.UUID;
+
+public interface SnowflakeIdGenerator extends IdGenerator {
+  /** Offset of the snowflake ID generator since the 1970-01-01T00:00:00Z epoch instant. */
+  Instant EPOCH_OFFSET =
```
dimas-b: nit: this is not technically an offset, but an exact moment in time. How about just `EPOCH`?

persistence/nosql/idgen/spi/src/main/java/org/apache/polaris/ids/spi/IdGeneratorFactory.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.polaris.ids.api.IdGenerator;
+
+public interface IdGeneratorFactory<I extends IdGenerator> {
+  String name();
+
+  void validateParameters(Map<String, String> params, IdGeneratorSource idGeneratorSource);
+
+  I buildIdGenerator(Map<String, String> params, IdGeneratorSource idGeneratorSource);
+
+  I buildSystemIdGenerator(Map<String, String> params, LongSupplier clockMillis);
```
dimas-b: It would be nice to add javadoc about the differences between `buildSystemIdGenerator` and `buildIdGenerator`.

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorFactory.java
```diff
@@ -0,0 +1,107 @@
...
+    if (offset != null) {
+      offsetMillis = Instant.parse(offset).toEpochMilli();
+    }
+
+    validateArguments(timestampBits, sequenceBits, nodeIdBits, offsetMillis, idGeneratorSource);
+  }
+
+  @Override
+  public SnowflakeIdGenerator buildSystemIdGenerator(
+      Map<String, String> params, LongSupplier clockMillis) {
```
dimas-b: `clockMillis` is unused and callers always have it as a constant `0`. Do we need this parameter?

persistence/nosql/idgen/impl/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd">
+    <!-- File required by Weld (used for testing), not by Quarkus -->
```
dimas-b, line 23: Would it be possible to move this file to the test source tree?

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+ * under the License.
+ */
+package org.apache.polaris.ids.api;
+
+public interface IdGenerator {
+  /** Generate a new, unique ID. */
+  long generateId();
+
+  /** Generate the system ID for a node, solely used by/for node management purposes. */
+  long systemIdForNode(int nodeId);
```
dimas-b, line 27: Is this ID expected to be the same for all `IdGenerator` implementations? What if a future impl. is not node-based?

Should this be pushed down to the Snowflake ID code?

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/MonotonicClock.java
snazy: Eliminiate the _checked_ exception.

persistence/nosql/idgen/impl/src/main/resources/META-INF/beans.xml
snazy: no, there's an application scoped bean in here

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGenerator.java
snazy: This function isn't specific to the particular implementation.
See the upcoming node-id-lease stuff: there is _one_ implementation with _one_ constant configuration per setup.

persistence/nosql/idgen/spi/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+  compileOnly(libs.jakarta.annotation.api)
+  compileOnly(libs.jakarta.validation.api)
+  compileOnly(libs.jakarta.inject.api)
+  compileOnly(libs.jakarta.enterprise.cdi.api)
+
+  compileOnly(libs.smallrye.config.core)
+  compileOnly(platform(libs.quarkus.bom))
+  compileOnly("io.quarkus:quarkus-core")
+
+  compileOnly(project(":polaris-immutables"))
```
adutra: This dependency seems unused. I wonder if you didn't forget to annotate `IdGeneratorSource` with `@PolarisImmutable`? It seems like a good candidate for that (lots of anonymous classes implementing this interface).

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/MonotonicClock.java
```diff
@@ -0,0 +1,69 @@
...
+
+  void waitUntilTimeMillisAdvanced();
```
adutra, line 68: Could be good to add javadocs here (and above), it's not immediately clear what this method is supposed to do (spin-wait until the clock ticks?).

Also neither this method nor `sleepMillis` throw `InterruptedException`, which is surprising since they are clearly blocking. It could be good to add an `@implSpec` note about how the interrupt flag is expected to be handled.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGeneratorSpec.java
```diff
@@ -0,0 +1,52 @@
...
+@JsonSerialize(as = ImmutableIdGeneratorSpec.class)
+@JsonDeserialize(as = ImmutableIdGeneratorSpec.class)
+public interface IdGeneratorSpec {
+  @WithDefault("snowflake")
+  String type();
+
+  Map<String, String> params();
+
+  @PolarisImmutable
+  interface BuildableIdGeneratorSpec extends IdGeneratorSpec {
```
adutra, line 38: This class looks odd? Afaict we could use `ImmutableIdGeneratorSpec.builder()` instead. The only difference seems to be that there is a default type here, whereas in `IdGeneratorSpec` there isn't, but it's certainly possible to overcome this limitation somehow.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+
+  int DEFAULT_NODE_ID_BITS = 10;
+  int DEFAULT_TIMESTAMP_BITS = 41;
+  int DEFAULT_SEQUENCE_BITS = 12;
+
+  long constructId(long timestamp, long sequence, long node);
+
+  long timestampFromId(long id);
+
+  long timestampUtcFromId(long id);
```
adutra, line 43: "Timestamp UTC" does not make sense to me 🤔
From the implementation class, it seems you meant "timestamp since Unix Epoch" instead.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  long EPOCH_OFFSET_MILLIS = EPOCH_OFFSET.toEpochMilli();
+
+  int DEFAULT_NODE_ID_BITS = 10;
+  int DEFAULT_TIMESTAMP_BITS = 41;
+  int DEFAULT_SEQUENCE_BITS = 12;
+
+  long constructId(long timestamp, long sequence, long node);
+
+  long timestampFromId(long id);
```
adutra, line 41: I suppose this is understood as "since the Snowflake Epoch (2025-03-01)". Could be good to add javadocs to clarify.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
adutra: Also, the expression is unnecessarily complex, the below one is 100% equivalent and imho easier to grasp:

```java
Instant.parse("2025-03-01T00:00:00Z")
```

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        params,
+        new IdGeneratorSource() {
+          @Override
+          public int nodeId() {
+            return 0;
+          }
+
+          @Override
+          public long currentTimeMillis() {
+            return SnowflakeIdGenerator.EPOCH_OFFSET_MILLIS;
```
adutra: This looks suspicious as it is returning a fixed timestamp. Did you mean `clockMillis.getAsLong()`?

```suggestion
            return clockMillis.getAsLong();
```

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorImpl.java
```diff
@@ -0,0 +1,408 @@
...
+  public long systemIdForNode(int nodeId) {
+    return constructIdUnsafe(timestampMax - 1, 0, nodeId);
+  }
+
+  private long constructIdUnsafe(long timestamp, long sequence, long nodeId) {
+    return (timestamp << timestampShift) | (nodeId << sequenceBits) | sequence;
+  }
+
+  @Override
+  public long constructId(long timestamp, long sequence, long nodeId) {
```
adutra, line 169: This method effectively bypasses the Id generator source. Maybe it shouldn't be public? Having two methods in the same API that generate IDs using different mechanisms seems error-prone. OTOH this method seems only called from `timeUuidToId`, in this same class, so it seems it could be private.

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorImpl.java
```diff
@@ -0,0 +1,408 @@
...
+        (sequence & sequenceMask) == sequence, "sequence argument %s out of range", sequence);
+    checkArgument((nodeId & nodeMask) == nodeId, "nodeId argument %s out of range", nodeId);
+    return constructIdUnsafe(timestamp, sequence, nodeId);
+  }
+
+  @Override
+  public long generateId() {
+    var nodeId = idGeneratorSource.nodeId();
+    checkState(nodeId >= 0, "Cannot generate a new ID, shutting down?");
+    var nodeIdPattern = ((long) nodeId) << sequenceBits;
```
adutra, line 184: Nit: move this variable declaration to line 228. It's not needed before.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/MonotonicClock.java
dimas-b: noted

persistence/nosql/idgen/impl/src/main/resources/META-INF/beans.xml
dimas-b: sorry for the false alarm :sweat_smile:

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGenerator.java
adam-christian-software: Actually, the more that I read about the node-id lease stuff. I don't know whether this should be in this module. Here's my thinking:
1. It seems that NodeID generation is special and not a Snowflake ID generation. The implementation is only based upon the node id passed in and some system-wide variables such as SnowflakeIdGeneratorImpl#timestampMax, SnowflakeIdGeneratorImpl#timestampShift, & SnowflakeIdGeneratorImpl#sequenceBits. So, in practice, it's really just the Node ID passed in.
2. So, given that, I think we could pull this out and just put it into the node leasing modules. That way, we can keep the IdGenerator clean for the cases that require the distributed id generation.

What do y'all think?

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGenerator.java
adam-christian-software: Updated.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
adam-christian-software: @dimas-b  - At the risk of sounding too verbose, how about ID_GENERATOR_EPOCH? I'd like to distinguish this from Unix Epoch?

And, I agree @adutra . I'll update.

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorFactory.java
adam-christian-software: I don't believe so. Removing.

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorFactory.java
adam-christian-software: So, I think that there are some special-casing that is happening and is leaking into our interfaces and implementation. See here as well: https://github.com/apache/polaris/pull/2131/files#r2214779625

We need the Snowflake ID Generation for Commit IDs and Entity Object IDs.

We need a Node ID. Node IDs cannot be Snowflake IDs because Snowflake IDs require a NodeID.

So, I actually think that we need a NodeIdGenerator concept for Node IDs and the concepts presented here for the other IDs. Right now, we are blending them into one interface and I think that is making things a bit confusing.

For example:
1. IdGeneratorFactory#buildSystemIdGenerator
2. IdGenerator#systemIdForNode

I think that we probably need to separate these concepts.

What do y'all think, @dimas-b & @snazy?

persistence/nosql/idgen/spi/src/main/java/org/apache/polaris/ids/spi/IdGeneratorFactory.java
adam-christian-software: This is related to https://github.com/apache/polaris/pull/2131/files#r2215597303

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/IdGenerator.java
dimas-b: +1 to Adam's point 1.

... however, I have a bigger concern. Suppose we run with Snowflake IDs for a while and then change to another ID generator. Assume `generateId()` outputs do not clash. Still, do we expect `systemIdForNode(X)` to return the same value for all generator implementations and for all possible values of `X`?

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
dimas-b: `ID_GENERATOR_EPOCH` LGTM. Alt: `ID_EPOCH`.

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorFactory.java
dimas-b: I tend to think node ID generation should be internal to the Snowflake ID code. A sharable `NodeIdGenerator` generator might be an overkill, but that is a secondary concern.

persistence/nosql/idgen/spi/src/main/java/org/apache/polaris/ids/spi/IdGeneratorFactory.java
dimas-b: Closing this thread for brevity. Let's continue on the linked thread.

persistence/nosql/idgen/api/src/main/java/org/apache/polaris/ids/api/SnowflakeIdGenerator.java
adam-christian-software: I like ID_EPOCH. I'll do that.

persistence/nosql/idgen/impl/src/main/java/org/apache/polaris/ids/impl/SnowflakeIdGeneratorFactory.java
adutra: My initial comment stems from the fact that the `clockMillis` parameter is part of the method signature (thus part of the API) but is not used, so I don't know why it is there.

dimas-b:
I propose to merge this PR "as is" and address comments after we get the whole NoSQL persistence merged.

I'll deal with merge conflicts presently.

snazy (approved):

dimas-b:
updated to the latest `main` and resolved conflicts.

dimas-b (dismissed):

dimas-b:
If no objections, I propose to merge and open issue for follow-up on the topics discovered during the review.

snazy (dismissed):

adutra (approved):

dimas-b:
Resolved conflicts in `gradle/libs.versions.toml`

adutra (dismissed):
LGTM.

I think my previous remark still holds though:

https://github.com/apache/polaris/pull/2131/files#r2215576536

persistence/nosql/idgen/spi/build.gradle.kts
dimas-b: `IdGeneratorSource` in this PR matches https://github.com/snazy/polaris/blob/persistence-nosql/persistence/nosql/idgen/spi/src/main/java/org/apache/polaris/ids/spi/IdGeneratorSource.java

I believe the plan is to address node ID concerns in the ID generator after we merge the whole NoSQL code. I'll file an issue for that.

I'll remove the unused dep on `polaris-immutables` for now.

adutra (approved):

persistence/nosql/idgen/spi/build.gradle.kts
dimas-b: #2494

PR #2143, opened 2025-07-20
Add podman support

MonkeyCanCode:
Similar to https://github.com/apache/polaris/pull/2051/files, this PR add support for podman for python client regeneration script. Also, as `DOCKER` seems to be preferred, I changed the `Makefile` to use the same to avoid confusion later.

### DIFF
--- a/client/templates/regenerate.sh
+++ b/client/templates/regenerate.sh
@@ -20,6 +20,8 @@
 
 set -e
 
+[[ -z ${DOCKER} ]] && DOCKER="$(which podman > /dev/null && echo podman || echo docker)"
+
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 pushd "$SCRIPT_DIR/../python" > /dev/null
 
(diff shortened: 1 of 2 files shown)


client/templates/regenerate.sh
```diff
@@ -20,6 +20,8 @@
 
 set -e
 
+[[ -z ${DOCKER} ]] && DOCKER="$(which podman > /dev/null && echo podman || echo docker)"
```
dimas-b, line 23: nit: Comments say `docker` is preferred, but this logic prefers `podman` :)

dimas-b (approved):
Changes LGTM, but I do not have an opinion in terms of preference for docker v.s podman... Let's wait for other people to comment.

client/templates/regenerate.sh
MonkeyCanCode: so `DOCKER` here refs to the variable name used in the Makefile as well as those script to change (people can use this variable to set to podman etc. in the current Makefile, I am using `CONTAINER_TOOL` as oppose to `DOCKER` to avoid confusion). But from the other PR with @snazy, podman is more preferred.

dimas-b: thx, all clear now :)

PR #2148, opened 2025-07-21
Extract ResolverFactory from PolarisEntityManager

XN137:
`PolarisEntityManager.prepareResolver` was the only place that used the
`EntityCache`.
By introducing an application-scoped `ResolverFactory` the creation of
an `EntityCache` is decided only once in `QuarkusProducers`, which will become
the only place where `MetaStoreManagerFactory.getOrCreateEntityCache` gets called.

Note that the little left-over functionality of `PolarisEntityManager` can
be split out to smaller but more dedicated interfaces in a follow-up most likely.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
@@ -46,7 +46,7 @@ public class PolarisEntityManager {
 
   /**
    * @param metaStoreManager the metastore manager for the current realm
-   * @param resolverFactory the resolver factor to use
+   * @param resolverFactory the resolver factory to use
    */
   public PolarisEntityManager(
       @Nonnull PolarisMetaStoreManager metaStoreManager, @Nonnull ResolverFactory resolverFactory) {
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
@@ -29,9 +29,8 @@ import org.apache.polaris.core.entity.PolarisEntitySubType;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisGrantRecord;
 import org.apache.polaris.core.entity.PolarisPrivilege;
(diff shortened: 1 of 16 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
```diff
@@ -40,39 +39,28 @@
...
-  private final EntityCache entityCache;
+  private final ResolverFactory resolverFactory;
 
   // Lazily instantiated only a single time per entity manager.
   private ResolvedPolarisEntity implicitResolvedRootContainerEntity = null;
 
   /**
    * @param metaStoreManager the metastore manager for the current realm
-   * @param entityCache the entity cache to use (it may be {@code null}).
+   * @param resolverFactory the resolver factor to use
```
snazy:
```suggestion
   * @param resolverFactory the resolver factory to use
```

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
```diff
@@ -40,39 +39,28 @@
...
    */
   public PolarisEntityManager(
-      @Nonnull PolarisMetaStoreManager metaStoreManager, @Nullable EntityCache entityCache) {
+      @Nonnull PolarisMetaStoreManager metaStoreManager, @Nonnull ResolverFactory resolverFactory) {
```
snazy, line 52: Nit:
```suggestion
      @Nonnull PolarisMetaStoreManager metaStoreManager, ResolverFactory resolverFactory) {
```

We kind-of agreed that parameters a not-null by default. But I don't mind having it explicit.

snazy (commented):
LGTM

Minor comments

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisEntityManager.java
XN137: while i agree that assuming non-null would be the best default it would seem weird to have only 1 of the parameters annotated like that.
also it would be unjustified to remove the annotation from the 1st parameter as part of this PR, so i add it to the 2nd parameter for consistency.

long term we should starting using https://jspecify.dev/docs/applying/#2-add-nullmarked

snazy (approved):

Issue #1801, opened 2025-06-03
Azure credential subscoping


eric-maynard:
### Is your feature request related to a problem? Please describe.

Polaris [currently doesn't subscope Azure credentials based on the provided locations](https://github.com/apache/polaris/blob/90c2580103e129b70ce3b128f1fdee3bceaa8edb/polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java#L84), which is different from the other two storage formats.

raveendra11:
hi @eric-maynard I'll check on this, could you assign?

raveendra11:
> ### Describe the solution you'd like
> Polaris [currently doesn't subscope Azure credentials based on the provided locations](https://github.com/apache/polaris/blob/90c2580103e129b70ce3b128f1fdee3bceaa8edb/polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java#L84), which is different from the other two storage formats.

are we talking about aws/gcp here for other two storages?

eric-maynard:
Yes @raveendra11, those support credential subscoping.

raveendra11:
Shall we do ? - 
 *Replace findAny() with iterator().next() in the getSubscopedCreds method to always pick the first location in the list. This makes the choice consistent every time.
 *Add a check at the start of getSubscopedCreds to throw an error called IllegalArgumentException.
 *Add a new helper method called extractCommonPathPrefix to find a shared path among all the provided locations. like, if locations are /path1/file1 and /path1/file2, the common path is path1.

eric-maynard:
@raveendra11 the issue is that the credentials are not subscoped to the base location for Azure

raveendra11:
@eric-maynard So, the main issue is not just about picking one of the allowed locations, but that we need to generate the Azure SAS token in a way that only allows access to the shared base path from all the given read/write locations. Is it?

If so, we'll find the common path shared by all allowed locations. and use that common path to scope the SAS token, so access is limited just to that path.

eric-maynard:
If possible, we should prefer to subscope the credentials to a list of locations, rather than to a shared prefix.

raveendra11:
> If possible, we should prefer to subscope the credentials to a list of locations, rather than to a shared prefix.

Hi @eric-maynard Really sorry, I'm occupied with my academics. Couldn't be able to check anymore on this as of now. if this is critical, you could please re-assign to someone.
Thank you.

PR #2149, opened 2025-07-21
Remove redundant locations when constructing access policies

eric-maynard:
Iceberg tables can technically store data across any number of paths, but Polaris currently uses 3 different locations for credential vending:
1. The table's base location
2. The table's `write.data.path`, if set
3. The table's `write.metadata.path`, if set

This was intended to capture scenarios where e.g. (2) is not a child path of (1), so that the vended credentials can still be valid for reading the entire table. However, there are systems that seem to always set (2) and (3), such as:

1. `s3:/my-bucket/base/iceberg`
2. `s3:/my-bucket/base/iceberg/data`
3. `s3:/my-bucket/base/iceberg/metadata`

In such cases the extra paths (e.g. extra resources in the AWS Policy) are redundant. In one such case, these redundant paths caused the policy to exceed the maximum allowable 2048 characters.

This PR removes redundant paths -- those that are the child of another path -- from the list of accessible locations tracked for a given table and does some slight refactoring to consolidate the logic for extracting these paths from a TableMetadata.


polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
```diff
@@ -62,4 +68,50 @@ public class StorageUtil {
...
+      String baseLocation, Map<String, String> properties) {
+    Set<String> locations = new HashSet<>();
+    locations.add(baseLocation);
+    if (properties.containsKey(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_DATA_LOCATION_KEY)) {
+      locations.add(properties.get(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_DATA_LOCATION_KEY));
+    }
+    if (properties.containsKey(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY)) {
+      locations.add(
+          properties.get(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY));
+    }
```
snazy: Can be simplified by blindly adding those to the `Set` and at the end do a `.remove(null)`.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
```diff
@@ -62,4 +68,50 @@ public class StorageUtil {
...
+
+  /** Removes "redundant" locations, so {/a/b/, /a/b/c, /a/b/d} will be reduced to just {/a/b/} */
+  private static @Nonnull Set<String> removeRedundantLocations(Set<String> locationStrings) {
+    HashSet<String> result = new HashSet<>(locationStrings);
+    result.remove(null);
+
+    for (String potentialParent : locationStrings) {
+      for (String potentialChild : locationStrings) {
+        if (potentialParent != null
+            && potentialChild != null
```
snazy: Shouldn't the elements never be `null`?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
```diff
@@ -62,4 +68,50 @@ public class StorageUtil {
...
+  }
+
+  /** Given a ViewMetadata, extracts the locations where the view's [meta]data might be found. */
+  public static @Nonnull Set<String> getLocationsAllowedToBeAccessed(ViewMetadata viewMetadata) {
+    return Set.of(viewMetadata.location());
+  }
+
+  /** Removes "redundant" locations, so {/a/b/, /a/b/c, /a/b/d} will be reduced to just {/a/b/} */
+  private static @Nonnull Set<String> removeRedundantLocations(Set<String> locationStrings) {
+    HashSet<String> result = new HashSet<>(locationStrings);
```
snazy, line 96: Since this is a new collection, it can be produced by
```java
locationStrings.stream()
  .filter(Objects::nonNull)
  .map(StorageLocation::of)
  .collect(Collectors.toCollection(HashSet::new));
```

polaris-core/src/test/java/org/apache/polaris/core/storage/StorageUtilTest.java
```diff
@@ -66,4 +68,37 @@ public void testAuthorityWithPort() {
     Assertions.assertThat(StorageUtil.getBucket("s3://bucket:8080/path/file.txt"))
         .isEqualTo("bucket:8080");
   }
+
+  @Test
+  public void getLocationsAllowedToBeAccessed() {
```
snazy, line 73: Mind adding test cases with both location keys?

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1491,24 +1466,9 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
...
-                .get(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY)
-            != null) {
-          dataLocations.add(
-              metadata
-                  .properties()
-                  .get(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY));
-        }
+        Set<String> dataLocations =
+            StorageUtil.getLocationsAllowedToBeAccessed(metadata.location(), metadata.properties());
+        new HashSet<>();
```
snazy: why this?

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -2612,16 +2572,6 @@ protected FileIO loadFileIO(String ioImpl, Map<String, String> properties) {
         callContext, ioImpl, properties, identifier, locations, storageActions, resolvedPath);
   }
 
-  private void blockedUserSpecifiedWriteLocation(Map<String, String> properties) {
```
snazy: `loadFileIO` is also unused and should be removed as well.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
eric-maynard: No, there's a call to `getLocationsAllowedToBeAccessed` where the `baseLocation` is null. In the future, other callsites could appear. I think a check here is a good idea.

eric-maynard: Oh, just saw your other comment -- if we _remove_ null from the input collection as you suggest, then yes this check can go away!

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
eric-maynard: That would remove duplicate locations, but not redundant locations like we want to. We'd still need to loop over the collection.

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: Agreed that it can be removed, but unlike `blockedUserSpecifiedWriteLocation` it's not related to this PR so I think we should do that separately.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
snazy: Yes, but you'd save the exponential instantiation of SotrageLocation objects

snazy: Actually you could safe the inner loop with a sorted collection, if the locations end with a `/`.

eric-maynard: Oh true, let's try to save the extra `StorageLocation` calls. Just pushed an update, let me know what you think!

I don't think we need to get too crazy trying to optimize this loop, considering the real number of locations is <=3 in practice and that we loop over them elsewhere when doing the overlap check.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
```diff
@@ -62,4 +70,46 @@ public class StorageUtil {
...
+        locationStrings.stream()
+            .filter(Objects::nonNull)
+            .map(StorageLocation::of)
+            .collect(Collectors.toSet());
+
+    for (StorageLocation potentialParent : locations) {
+      for (StorageLocation potentialChild : locations) {
+        if (!potentialParent.equals(potentialChild)) {
+          if (potentialChild.isChildOf(potentialParent)) {
+            locations.remove(potentialChild);
```
snazy: Looks like this can run into multiple error conditions: the `Set` being not mutable (`There are no guarantees on the type, mutability, ... of the Set returned` from `Collectors.toSet()`), a `ConcurrentModificationException` in the `Iterator`s of the nested loops, or one of the iterators (depending on the `Set` implementation) skipping an element after a `.remove()`

eric-maynard: We can revert [the change](https://github.com/apache/polaris/pull/2149#discussion_r2224689200), but what about just wrapping the set in `new HashSet`?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
eric-maynard: It turns out we don't actually want to loop over `StorageLocation`s, since we want to preserve the original strings as they came in rather than round-trip them through `StorageLocation.of(...).toString()`. This change actually breaks the new test as locations like `/foo/` get turned into `file:///foo/`

It would still be possible to save the extra calls to `StorageLocation.of` by precomputing some mapping between string locations and storage locations, but I'm not sure that the extra complexity is worth the handful of calls saved.

snazy: Looked a bit deeper into this and figured out some things need to be clarified in the Javadocs.

Although this reduction is technically not needed for Azure (actually: not possible :sigh:), it's probably worth to have some tests for this.

Opened https://github.com/eric-maynard/polaris/pull/5 with what I had in mind, feel free to take what you want/need.

eric-maynard: Wait, why is the reduction not possible for Azure?

snazy: Not saying it's not possible (in the code here), but there's no IAM for individual paths.

eric-maynard: Ah, yes, that's true -- basically #1801 is needed.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
eric-maynard: I went with a revert, because it seems that the logic to do this is actually much more complicated than originally thought. The [logic](https://github.com/eric-maynard/polaris/pull/5/files#diff-92d056b323510408ca45ed715cec73a34f1cd0446575b559e44de1564c423bb5R117) becomes more stateful, more imperative, and just generally longer. It seems that the extra 6 calls to `StorageLocation.of` could be preferable.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
eric-maynard: The main change to the javadocs seems to be comments like `Must be overridden by subclasses`, but I don't think that's true. In fact it's the opposite -- to date we've only created subclasses where we wish to override, but we could create subclasses where it's not necessary. For example, a `LocalStorageLocation`.

snazy: True, "must" might be too strong. The implementations may however be very specific to the storage-type - that's what I've been thinking of.

eric-maynard: That's a good point, let me make this explicit in a javadoc

PR #2154, opened 2025-07-22
Enable python client debug flag

MonkeyCanCode:
Enable request tracing via money patching:

Original:
```
➜  polaris git:(python_client_debug_flag) ./polaris --profile dev principals list
{"name": "root", "clientId": "root", "properties": {}, "createTimestamp": 1753151091780, "lastUpdateTimestamp": 1753151091780, "entityVersion": 1}
```

With debug mode (dummy password from the getting start example):
```
➜  polaris git:(python_client_debug_flag) ./polaris --profile dev --debug principals list
Request: POST http://localhost:8181/api/catalog/v1/oauth/tokens
Headers: HTTPHeaderDict({'Content-Type': 'application/x-www-form-urlencoded'})
Body: grant_type=client_credentials&client_id=root&client_secret=s3cr3t&scope=PRINCIPAL_ROLE%3AALL

Request: GET http://localhost:8181/api/management/v1/principals
Headers: {'Accept': 'application/json', 'User-Agent': 'OpenAPI-Generator/1.0.0/python', 'Authorization': 'Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJwb2xhcmlzIiwic3ViIjoiMSIsImlhdCI6MTc1MzE1NjUwMiwiZXhwIjoxNzUzMTYwMTAyLCJqdGkiOiIyMjc3NTc4Yi1mOGQ2LTQ3NjctODc2Yy0xZjg0YjVhOTFjOWIiLCJhY3RpdmUiOnRydWUsImNsaWVudF9pZCI6InJvb3QiLCJwcmluY2lwYWxJZCI6MSwic2NvcGUiOiJQUklOQ0lQQUxfUk9MRTpBTEwifQ.CXHMjqgGtGZZ9hUtS_Tqi4hr2Zt5kIdifqWVH0sWO2Ycbju3J6f4KWu4TgBB7FT6PQUbFqIAZiNBOAkZRpCDygUPEZ3-bt3cbTt9LWfkQYRXFnJF7-fM42zSK3Go7t1AVfYShDLozexgBvZGg700SYdcffKdCpARke7IX8VjLl_0uqZ4ygtjWwmoIHxfVH8WCTXajdvTOH7MF6aPgdVr5_y4KiCfZoTu5DG9gSJLYYAu1sVfUtQ07hGWs6wJUENLt9ZBjBBJI80Um70Oq3U508MC4y10FBWIKRrs8655le976TfcRjY_BPxpJaekBQIrxP7qz1hqP8XGSPRuFTgWbQ'}

{"name": "root", "clientId": "root", "properties": {}, "createTimestamp": 1753151091780, "lastUpdateTimestamp": 1753151091780, "entityVersion": 1}
```

Also, as the current code will requires an user to provide an argument even if the type is bool. To fix that (assuming that is not intensional), a code fix is provided (otherwise, we will need to do `--debug XXX`.

### DIFF
--- a/client/python/cli/options/parser.py
+++ b/client/python/cli/options/parser.py
@@ -57,6 +57,7 @@ class Parser(object):
         ),
         Argument(Arguments.PROFILE, str, hint="profile for token-based authentication"),
         Argument(Arguments.PROXY, str, hint="proxy URL"),
+        Argument(Arguments.DEBUG, bool, hint="Enable debug mode"),
     ]
 
     @staticmethod
@@ -64,15 +65,16 @@ class Parser(object):
         parser = TreeHelpParser(description="Polaris CLI")
 
         for arg in Parser._ROOT_ARGUMENTS:
+            kwargs = {"help": arg.hint}
             if arg.default is not None:
-                parser.add_argument(
-                    arg.get_flag_name(),
-                    type=arg.type,
-                    help=arg.hint,
-                    default=arg.default,
-                )
+                kwargs["default"] = arg.default
+
+            if arg.type is bool:
+                kwargs["action"] = "store_true"
             else:
-                parser.add_argument(arg.get_flag_name(), type=arg.type, help=arg.hint)
+                kwargs["type"] = arg.type
+
+            parser.add_argument(arg.get_flag_name(), **kwargs)
 
         # Add everything from the option tree to the parser:
         def add_arguments(parser, args: List[Argument]):
--- a/client/python/cli/polaris_cli.py
+++ b/client/python/cli/polaris_cli.py
@@ -67,8 +67,6 @@ class PolarisCli:
             command = Command.from_options(options)
             command.execute()
         else:
-            if options.debug:
-                PolarisCli._enable_api_request_logging()
             client_builder = PolarisCli._get_client_builder(options)
             with client_builder() as api_client:
                 try:
@@ -76,6 +74,8 @@ class PolarisCli:
 
                     admin_api = PolarisDefaultApi(api_client)
                     command = Command.from_options(options)
+                    if options.debug:
+                        PolarisCli._enable_api_request_logging()
                     command.execute(admin_api)
                 except Exception as e:
                     PolarisCli._try_print_exception(e)
@@ -90,12 +90,12 @@ class PolarisCli:
         # Define the wrapper function
         @functools.wraps(urllib3.PoolManager.original_urlopen)
         def urlopen_wrapper(self, method, url, **kwargs):
-            print(f"Request: {method} {url}")
+            sys.stderr.write(f"Request: {method} {url}\n")
             if "headers" in kwargs:
-                print(f"Headers: {kwargs['headers']}")
+                sys.stderr.write(f"Headers: {kwargs['headers']}\n")
             if "body" in kwargs:
-                print(f"Body: {kwargs['body']}")
-            print()
+                sys.stderr.write(f"Body: {kwargs['body']}\n")
+            sys.stderr.write("\n")
             # Call the original urlopen method
             return urllib3.PoolManager.original_urlopen(self, method, url, **kwargs)
 
--- a/client/python/cli/polaris_cli.py
+++ b/client/python/cli/polaris_cli.py
@@ -21,6 +21,8 @@ import json
 import os
 import sys
 from json import JSONDecodeError
+import urllib3
+import functools
 
 from typing import Dict
 
@@ -65,6 +67,8 @@ class PolarisCli:
             command = Command.from_options(options)
             command.execute()
         else:
+            if options.debug:
+                PolarisCli._enable_api_request_logging()
             client_builder = PolarisCli._get_client_builder(options)
             with client_builder() as api_client:
                 try:
@@ -76,6 +80,26 @@ class PolarisCli:
                 except Exception as e:
                     PolarisCli._try_print_exception(e)
                     sys.exit(1)
+
+    @staticmethod
+    def _enable_api_request_logging():
+        # Store the original urlopen method
+        if not hasattr(urllib3.PoolManager, "original_urlopen"):
+            urllib3.PoolManager.original_urlopen = urllib3.PoolManager.urlopen
+
+        # Define the wrapper function
+        @functools.wraps(urllib3.PoolManager.original_urlopen)
+        def urlopen_wrapper(self, method, url, **kwargs):
+            print(f"Request: {method} {url}")
+            if "headers" in kwargs:
+                print(f"Headers: {kwargs['headers']}")
+            if "body" in kwargs:
+                print(f"Body: {kwargs['body']}")
+            print()
+            # Call the original urlopen method
+            return urllib3.PoolManager.original_urlopen(self, method, url, **kwargs)
+
+        urllib3.PoolManager.urlopen = urlopen_wrapper
 
     @staticmethod
     def _try_print_exception(e):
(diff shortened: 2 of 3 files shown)


client/python/cli/options/parser.py
```diff
@@ -57,22 +57,24 @@ class Parser(object):
...
     ]
 
     @staticmethod
     def _build_parser() -> argparse.ArgumentParser:
         parser = TreeHelpParser(description="Polaris CLI")
 
         for arg in Parser._ROOT_ARGUMENTS:
+            kwargs = {"help": arg.hint}
             if arg.default is not None:
-                parser.add_argument(
```
eric-maynard, line 68: Is this needed because this is the first root arg that's binary?

client/python/cli/polaris_cli.py
```diff
@@ -65,6 +67,8 @@ def execute(args=None):
             command = Command.from_options(options)
             command.execute()
         else:
+            if options.debug:
+                PolarisCli._enable_api_request_logging()
```
eric-maynard: I think we should do this after the command is parsed but before it's executed... or, if possible, in the middle of command execution. As the logic is right now it looks like there are cases where we'd emit the logging but not actually make the request.

client/python/cli/options/parser.py
```diff
@@ -57,22 +57,24 @@ class Parser(object):
         ),
         Argument(Arguments.PROFILE, str, hint="profile for token-based authentication"),
         Argument(Arguments.PROXY, str, hint="proxy URL"),
+        Argument(Arguments.DEBUG, bool, hint="Enable debug mode"),
```
eric-maynard, line 60: Thinking this over, it might be nice to have multiple modes, e.g. `default`, `debug`, `dry-run`. Sometimes you just want to see the command without running it, sometimes you want to see it and run it, and sometimes you just want to run the commands.

client/python/cli/polaris_cli.py
```diff
@@ -77,6 +81,26 @@ def execute(args=None):
...
+        # Store the original urlopen method
+        if not hasattr(urllib3.PoolManager, "original_urlopen"):
+            urllib3.PoolManager.original_urlopen = urllib3.PoolManager.urlopen
+
+        # Define the wrapper function
+        @functools.wraps(urllib3.PoolManager.original_urlopen)
+        def urlopen_wrapper(self, method, url, **kwargs):
+            print(f"Request: {method} {url}")
+            if "headers" in kwargs:
+                print(f"Headers: {kwargs['headers']}")
```
eric-maynard: Does it make sense to use STDERR so you can still handle stdout if need be?

client/python/cli/options/parser.py
MonkeyCanCode: so a single mode flag and 3 modes supported? What will dry-run do? show the request?

client/python/cli/options/parser.py
MonkeyCanCode: So with current code, when we provide bool flag, it will still expected an argument. Thus, I made this change. When no bool flag is provided, it is False. If bool flag is ever provided, it is True (which matched to the behavior in add_arguments).

client/python/cli/polaris_cli.py
MonkeyCanCode: k. Let me move this one.

client/python/cli/polaris_cli.py
MonkeyCanCode: Actually, that is a very good idea for CLI. I will make this change.

client/python/cli/options/parser.py
eric-maynard: Yeah, just show the request. This isn't a blocker, I'm just wondering if it might be useful because 
1. dry-run would be very useful for testing
2. the current mode ("debug"?) is a little fragile because it non-transactionally does the printing and executes the command

client/python/cli/options/parser.py
eric-maynard: But it looks like we only changed it for the root arguments, while the other boolean arguments (e.g. `CASCADE`) are unchanged

MonkeyCanCode: that is correct, the above only change for the root one as it is the one that is enforcing everything to have an value for a given argument. The non-root arguments should work with following:
```
                if arg.type is bool:
                    del kwargs['type']
                    parser.add_argument(arg.get_flag_name(), **kwargs, action='store_true')
````

E.g.
```
➜  polaris git:(python_client_debug_flag) ✗ ./polaris --profile dev privileges catalog revoke --catalog quickstart_catalog --catalog-role quickstart_catalog_role --cascade CATALOG_MANAGE_CONTENT
Exception when communicating with the Polaris server. 501: Not Implemented
```

File: client/python/cli/options/parser.py (hunk 3 above)
Comment (MonkeyCanCode) [reply client/python/cli/options/parser.py:60]: so looking at implementing the dry-run mode, it turns out a bit more work will be needed. For example, for a single update principal property request, it will need to run following requests:
1. get principal
2. update principal

Now as there is no native way (as far as I know) to do this dry-run, we will then need to do another monkey patching on the urlopen. Now the challenge is as a single CLI request will be convert to multiple actual API requests, we will then need to create proper mock responses for each of the request to move forward until we hit the last one. 

If this is the feature we would like to support (and no other better solution other than doing a bunch mock responses), I can proceed with those implementation. The only catch that I can think off is later on if we changed API on diff version, we will need to keep those up to date (e.g. have mock for v1 and v2 as they may have diff response etc.). Unlike the debug mode, where we are really just enrich the output but showing request/header/body. 

Let me know what you think.

Comment (eric-maynard) [reply client/python/cli/options/parser.py:60]: Oh, I see. For some commands it's not easy to support dry-run. I would be fine with just failing in that case, or if we want to be really fancy we can let the user manually run the command and paste the response into stdin.

Anyway, if this all sounds like too much work let's just skip it -- was only an idea.

File: client/python/cli/polaris_cli.py [hunk 5]
```diff
@@ -72,11 +74,33 @@ def execute(args=None):
...
+
+        # Define the wrapper function
+        @functools.wraps(urllib3.PoolManager.original_urlopen)
+        def urlopen_wrapper(self, method, url, **kwargs):
+            sys.stderr.write(f"Request: {method} {url}\n")
+            if "headers" in kwargs:
+                sys.stderr.write(f"Headers: {kwargs['headers']}\n")
+            if "body" in kwargs:
+                sys.stderr.write(f"Body: {kwargs['body']}\n")
+            sys.stderr.write("\n")
```
eric-maynard, line 98: nit: looks like this newline is not needed

client/python/cli/polaris_cli.py
```diff
@@ -72,11 +74,33 @@ def execute(args=None):
...
+    @staticmethod
+    def _enable_api_request_logging():
+        # Store the original urlopen method
+        if not hasattr(urllib3.PoolManager, "original_urlopen"):
+            urllib3.PoolManager.original_urlopen = urllib3.PoolManager.urlopen
+
+        # Define the wrapper function
+        @functools.wraps(urllib3.PoolManager.original_urlopen)
+        def urlopen_wrapper(self, method, url, **kwargs):
+            sys.stderr.write(f"Request: {method} {url}\n")
```
eric-maynard, line 93: I don't know if there's a standard format for this kind of thing -- what you have looks good to me though. If we were taking the dry-run approach, it might be nice to print a runnable command like a curl.

eric-maynard (approved):

client/python/cli/options/parser.py
MonkeyCanCode: Yes, maybe come back later for this implementation if this is helpful for debugging.

client/python/cli/polaris_cli.py
MonkeyCanCode: Okay, I can do that in the next PR.

PR #2156, opened 2025-07-22
Automate the release guide

### LINKED ISSUE #2017: Automate drafting, verifying, and publish the release

A typical Apache Software Foundation release has the following workflow:

1. Draft a release
2. Start a VOTE on the dev mailing list
3. If the VOTE fails, the release has failed - "go to step 1"
4. If the VOTE passes, publish the release

As of this writing (Jul 9, 2025), the process is manual and undocumented. https://github.com/apache/polaris/issues/648 attempts to address the documentation. The Apache Trusted Releases effort (as a part of the larger ASF Tooling Initiative) dictates that all releases should be automated. See https://tooling.apache.org/trusted-release.html for more details. To that end, this ticket proposes automating the mechanisms for the release including automated processes for:

1. Drafting a release
2. Verifying major portions of the release
3. Publishing the release

The benefit of this issue are:

1. Reduce the friction of being a release manager
2. Ensure higher quality for releases through automated verification
3. Ensure less release mistakes by reducing manual steps
4. Enable the project to release more often
5. Allow releases to be created on “trusted instances” rather than user’s computers

There are various ways to do the automated release. The major acceptance criteria are:

1. Automated testing as part of the release verification process
2. Failing to publish the release if verification process fails
3. The ability to redraft a release if verification fails or voting fails
4. Drafting, verification, & publishing should all be able to be triggered via a single command
5. Drafting a release notifies the dev mailing list for a vote
6. The automation must have a descriptive README.md that covers the automation framework.
7. The releases should not be built upon a user’s hardware, but only on “trusted instances” (for example, GitHub).
8. A user can verify that the code was built on a “trusted instance”.
9. A user can verify that the code was built from a specific Git commit.
10. A user can verify the code was not modified.
11. A user can verify the code was uploaded by a “trusted instance”. For example, all binaries could be uploaded via a GitHub workflow.
12. In general, the process is public, reproducible, and contains all the information to trust the release.

The team has been working on documenting the manual processes required for release, but this has the issues noted above.

### Additional context

Relevant Issues:

1. https://github.com/apache/polaris/issues/648
2. https://github.com/apache/polaris/issues/584 

Relevant Links:

1. https://github.com/apache/polaris/milestones - Milestones for Polaris
2. https://github.com/apache/polaris/pull/485 - A WIP ideation
3. https://tooling.apache.org/trusted-release.html  - The Apache Trusted Release Platform
4. https://www.apache.org/legal/release-policy - Apache Foundation release policy

In addition to this work, there are two other related efforts ongoing:

1. GitHub Workflow for Helm Publishing
2. GitHub Workflow for Docker Images

pingtimeout:
### Description

This PR automates most of the commands listed in the [release guide](https://polaris.apache.org/community/release-guide/).  It fixes #2017.  See the README.md files for information on how the code is structured and how it can be used.  The code includes:

* Machine configuration for releasing
* RC branch and tag creation
* Polaris build
* Integration tests (only the non-cloud IT)
* Packages upload to dist.apache.org and Nexus

Parts not automated:

* Run the cloud-specific integration tests
* Close the staging repository
* Anything starting at "Start the vote thread", which is about sending e-mails

This is a first implementation that aims at being a 1:1 mapping with the current release guide.

Note that it includes the instructions from #2039 that has not been merged yet, at the time of opening this PR, given that the high likelihood of that PR being merged before this one.

### Hints for code review
Recommendation for reviewers: start with the `libs/` folder to get an idea of all utility functions, then look at the `test/` folder to see what each script is expected to do, then review the scripts themselves.

### How to test
By default, the scripts run in dry-run mode.  This is intended in order to avoid accidentally performing releases during testing.

All the other parameters used by the scripts can be overridden using environment variables.  For example,

* if you want to create release branch and tags on your own Git fork, say `pingtimeout`, run `export APACHE_REMOTE_NAME=pingtimeout APACHE_REMOTE_URL=git@github.com:pingtimeout/polaris.git`
* if your GPG key is not in `https://downloads.apache.org/incubator/polaris/KEYS`, create a KEYS file locally with your key, run a small http server (e.g. `python -m http.server`) from your laptop to serve it, and set `export KEYS_URL=http://localhost:8000/KEYS`
* Etc...

Then, imply run each script in order.  The output will contain messages that describe the commands that would be executed, if dry-run mode was disabled.  See example below:

Capture d’écran 2025-08-01 à 10 26 37

### How to view all commands that would be executed
Each script in the `test/` directory verifies the commands that the associated release script would execute, if it was not in dry-run mode.  See for example [test-02-create-release-branch.sh#L87-L103](https://github.com/pingtimeout/polaris/blob/716d32497f33556b35cadafa9cd203afb23f3ab2/releasey/test/test-02-create-release-branch.sh#L87-L103) for the part that creates the release branch and updates the version files.


releasesv2/04-build-and-test.sh
```diff
@@ -0,0 +1,118 @@
...
+exec_process mkdir -p ./regtests/output
+exec_process chmod -R 777 ./regtests/output
+
+# Generate Python client
+print_info "Regenerating Python client..."
+exec_process ./gradlew regeneratePythonClient
+
+# Build container image
+print_info "Building Polaris container image..."
+exec_process ./gradlew \
```
jbonofre: I guess this step is "only" for tests, correct ?

releasesv2/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,166 @@
...
+exec_process svn co "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
+
+print_info "Copying files to destination directory..."
+version_dir="${dist_dev_dir}/${version}"
+helm_chart_version_dir="${dist_dev_dir}/helm-chart/${version}"
+exec_process mkdir -p "${version_dir}"
+exec_process mkdir -p "${helm_chart_version_dir}"
+exec_process cp "${source_dist_dir}"/* "${version_dir}/"
+exec_process cp "${binary_dist_dir}"/* "${version_dir}/"
+exec_process cp -r "${helm_chart_dir}" "${helm_chart_version_dir}/"
```
jbonofre: This step should be updated (it's correct with the current release guide, but I guess we want to use `helm create` command to create helm "package").

releasesv2/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,166 @@
...
+exec_process svn add "${version_dir}"
+exec_process svn add "${helm_chart_version_dir}"
+
+# Commit changes
+print_info "Committing changes..."
+exec_process svn commit -m "Stage Apache Polaris ${version} RC${rc_number}"
+
+# Publish Maven artifacts
+exec_process ./gradlew publishToApache -Prelease -PuseGpgAgent
+
```
jbonofre: We should publish the Docker images RC here (see #2039 ) here.

releasesv2/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,166 @@
...
+echo
+print_success "🎉 Distributions built and staged successfully!"
+echo
+print_info "Staged artifacts:"
+print_info "- Source and binary distributions: ${version_dir}"
+print_info "- Helm charts: ${helm_chart_version_dir}"
+echo
+print_info "Next steps:"
+print_info "1. Build and stage Maven artifacts using: ./gradlew publishToApache -Prelease -PuseGpgAgent"
```
jbonofre: This step is done at line 153 already.

releasesv2/04-build-and-test.sh
pingtimeout: Correct, it is only required so that we can run the regression tests as part of the release process.  Robert mentioned that, since we should be releasing from commits that have been validated by CI, we could skip this entirely?

releasesv2/05-build-and-stage-distributions.sh
pingtimeout: -1 for now.

The purpose of this automation is to be a 1:1 mapping with the release guide.  This will facilitate review.

If the release guide is updated, we can update these scripts as well and keep the 1:1 mapping.

releasesv2/05-build-and-stage-distributions.sh
pingtimeout: Indeed, fixing.

releasesv2/05-build-and-stage-distributions.sh
pingtimeout: I am going to update this PR accordingly.  Then we need to merge #2039 before this one.

pingtimeout: Done

releasey/libs/_log.sh
```diff
@@ -0,0 +1,68 @@
...
+#
+# Common Logging Functions
+#
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
```
snazy: Isn't the common name for this `RESET`?

releasey/libs/_log.sh
```diff
@@ -0,0 +1,68 @@
...
+
+#
+# Common Logging Functions
+#
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
```
snazy: Honestly, I think those should not be overwritten to not interfere with personalized settings (standard settings ESC sequences can make text hardly unreadable - dark vs bright or even more colorful terminal profiles).

releasey/libs/_log.sh
```diff
@@ -0,0 +1,68 @@
...
+  # script.  It either prints its arguments as-is to file descriptor 3, if it
+  # is open, or prepends "DEBUG: " and prints them to stdout if it is unset.
+  #
+  # This allows the programmatic verification of all commands that are executed
+  # by a script, by running it as follows:
+  #
+  #   $ ./script.sh 3>/tmp/script.log
+  #   $ cat /tmp/script.log
+  #   ./gradlew clean build
+  #   rm -rf ./regtests/output
```
snazy, line 70: regtests?

releasey/libs/_nexus.sh
```diff
@@ -0,0 +1,96 @@
...
+  # Check if Apache credentials are configured in ~/.gradle/gradle.properties, return non-zero if not
+  print_info "Checking for Apache credentials in ${GRADLE_PROPERTIES_FILE}..."
+
+  if [[ ! -f "${GRADLE_PROPERTIES_FILE}" ]]; then
+    print_warning "Gradle properties file not found: ${GRADLE_PROPERTIES_FILE}"
+    return 1
+  fi
+
+  if grep -q "^apacheUsername=" "${GRADLE_PROPERTIES_FILE}" 2>/dev/null &&
+     grep -q "^apachePassword=" "${GRADLE_PROPERTIES_FILE}" 2>/dev/null; then
```
snazy, line 52: Actually, you should not put clear text credentials in a plain text file (`~/.gradle/gradle.properties`).

In CI we'd have environment variables populated from GitHub secrets.

releasey/libs/_nexus.sh
```diff
@@ -0,0 +1,96 @@
...
+}
+
+function ensure_credentials_are_configured() {
+  # Ensure that Apache credentials are configured in either gradle.properties or environment variables, return non-zero if not
+  print_info "Checking Apache credentials configuration..."
+  
+  if check_apache_credentials_are_set_in_gradle_properties; then
+    print_success "All Maven/Gradle credential checks passed! Your Apache credentials are properly configured for releases."
+    return 0
+  fi
```
snazy, line 82: I'd leave out this step entirely and require those to be environment variables.

releasey/libs/_docker.sh
```diff
@@ -0,0 +1,120 @@
...
+    return 1
+  fi
+  print_success "✓ Docker buildx is available"
+  return 0
+}
+
+function check_docker_hub_login_status() {
+  # Check if user is logged in to Docker Hub (optional check)
+  print_info "Checking Docker Hub login status..."
+  if docker info 2>/dev/null | grep -q "Username:"; then
```
snazy, line 81: Not sure this is technically correct. You can be "logged in" (correct: have credentials) to many repos simultaneously.

This steps fails for me, although I've got all credentials setup, just via a credentials helper.

releasey/libs/_files.sh
```diff
@@ -0,0 +1,50 @@
...
+    exec_process echo ${version} >$VERSION_FILE
+  else
+    exec_process "echo ${version} > $VERSION_FILE"
+  fi
+}
+
+function ensure_cwd_is_project_root() {
+  # This function verifies that the script is executed from the project root
+  # directory and that the gradle wrapper script is present.
+  if [[ ! -f "gradlew" ]] || [[ ! -d ".git" ]]; then
```
snazy: `.git` doesn't need to be a directory, it's a file for Git worktrees.

releasey/libs/_constants.sh
```diff
@@ -0,0 +1,43 @@
...
+#
+
+libs_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# GPG constants
+KEYS_URL=${KEYS_URL:-"https://downloads.apache.org/incubator/polaris/KEYS"}
+KEYSERVER=${KEYSERVER:-"hkps://keyserver.ubuntu.com"}
+
+# Git/SVN repository constants
+APACHE_REMOTE_NAME=${APACHE_REMOTE_NAME:-"apache"}
```
snazy, line 28: Hm, the remote name is by default (Github workflows) `origin`.

releasey/libs/_constants.sh
```diff
@@ -0,0 +1,43 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+libs_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# GPG constants
+KEYS_URL=${KEYS_URL:-"https://downloads.apache.org/incubator/polaris/KEYS"}
+KEYSERVER=${KEYSERVER:-"hkps://keyserver.ubuntu.com"}
```
snazy, line 25: Why do we need the Ubuntu keyserver?

releasey/06-build-and-stage-docker-images.sh
```diff
@@ -0,0 +1,137 @@
...
+  print_error "Expected format: apache-polaris-x.y.z-incubating-rcN"
+  exit 1
+fi
+
+# Extract version components from regex match
+major="${BASH_REMATCH[1]}"
+minor="${BASH_REMATCH[2]}"
+patch="${BASH_REMATCH[3]}"
+rc_number="${BASH_REMATCH[4]}"
+version="${major}.${minor}.${patch}-incubating"
```
snazy, line 92: Can you make `-incubating` a constant in `_constants.sh`?

releasey/06-build-and-stage-docker-images.sh
```diff
@@ -0,0 +1,137 @@
...
+
+# Extract version components from git tag in one regex match
+git_tag_regex="^apache-polaris-([0-9]+)\.([0-9]+)\.([0-9]+)-incubating-rc([0-9]+)$"
+if [[ ! ${git_tag} =~ ${git_tag_regex} ]]; then
+  print_error "Invalid git tag format: ${git_tag}"
+  print_error "Expected format: apache-polaris-x.y.z-incubating-rcN"
+  exit 1
+fi
+
+# Extract version components from regex match
```
snazy: Looks identical to what's in `05-...`. Worth unifying?

releasey/06-build-and-stage-docker-images.sh
```diff
@@ -0,0 +1,137 @@
...
+print_info "Published Docker images:"
+print_info "- polaris-server: ${DOCKER_REGISTRY}/apache/polaris:${docker_tag}"
+print_info "- polaris-admin: ${DOCKER_REGISTRY}/apache/polaris-admin-tool:${docker_tag}"
+echo
+print_info "Docker Hub links:"
+print_info "- ${DOCKER_HUB_URL}/r/apache/polaris/tags"
+print_info "- ${DOCKER_HUB_URL}/r/apache/polaris-admin-tool/tags"
+echo
+print_info "Next steps:"
+print_info "1. Start the vote thread"
```
snazy, line 130: I think it's worth to have a proposal for the subject and body of the email printed here.
Actually, maybe even in it's own shell script.

releasey/03-create-release-tag.sh
```diff
@@ -0,0 +1,150 @@
+#!/bin/bash
```
snazy: Strictly speaking the it's `03-create-release-candidate-tag.sh`.

releasey/03-create-release-tag.sh
```diff
@@ -0,0 +1,150 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Create Release Tag Script
+#
+# Automates the "Create release tag" section of the release guide.
```
snazy: "release candidate", and some occurrences below

releasey/03-create-release-tag.sh
```diff
@@ -0,0 +1,150 @@
...
+  Creates a release tag for a release candidate.
+
+  Options:
+    --version VERSION
+        The release version in format x.y.z-incubating-rcN where N is the RC number
+    -h --help
+        Print usage information.
+
+  Examples:
+    $(basename "$0") --version 1.0.0-incubating-rc1
```
snazy: I'd prefer to split version and rc numbers and make those separate parameters.
`-incubating` _must_ be present, so it can be a constant.

releasey/libs/_files.sh
```diff
@@ -0,0 +1,50 @@
...
+  # redirection, we cannot use exec_process with the exact command that will be
+  # executed.
+  if [[ ${DRY_RUN:-1} -ne 1 ]]; then
+    exec_process echo ${version} >$VERSION_FILE
+  else
+    exec_process "echo ${version} > $VERSION_FILE"
+  fi
+}
+
+function ensure_cwd_is_project_root() {
```
snazy: Actually, I don't think we need this one.
The location of the `01...` scripts is well known, so you can infer the "project root directory":

People would probably naively run it like this:
```bash
cd releasy
./01-...
./02-...
```

It's more convenient to let the scripts figure out what they can figure out.

releasey/04-build-and-test.sh
```diff
@@ -0,0 +1,118 @@
...
+      ;;
+  esac
+done
+
+print_info "Starting build and test process..."
+echo
+
+# Clean and build Polaris
+print_info "Building Polaris..."
+exec_process ./gradlew clean build
```
snazy, line 77: If you want a really clean working directory, you'd need the `git clean -xdf` hammer.

snazy: Note: you need neither the `gradlew clean` nor the `git clean...` in GH workflows.

releasey/04-build-and-test.sh
```diff
@@ -0,0 +1,118 @@
...
+# under the License.
+#
+
+#
+# Build and Test Script
+#
+# Builds Polaris completely and runs regression tests to verify the release.
+# Cloud-specific tests are disabled by default and require proper credentials
+# to be configured in the environment.
+#
```
snazy: Strictly speaking, these tests shouldn't be necessary, because a release should never be created from a Git commit that's not verified by CI.

Note: there's a [GH API endpoint](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#get-the-combined-status-for-a-specific-reference) to check the combined status.

I'd prefer a "commit status check" over mirroring what the CI workflows do.

releasey/03-create-release-tag.sh
pingtimeout: Renamed

releasey/03-create-release-tag.sh
pingtimeout: Fixed

releasey/03-create-release-tag.sh
pingtimeout: I considered pulling that part from your PR, indeed.  But decided to not reuse it so that the code is overall simpler.  The current approach is less smart, as it pushes the responsibility of adding `-incubating` to the release manager and enforces it via a regex validation.

The alternative would be:
* Have a `--version x.y.z` parameter that accepts `x.y.z` as well as `x.y.z-incubating`
* Have a `--rc N` parameter
* Have a logic that detects whether `-incubating` is already present and adds it, if missing

Overall, I would rather have a single `--version x.y.z-incubating-rcN` parameter validated by a regex.  Simpler code.

The only downside, really, is that when we graduate, we will have to update multiple shell scripts to remove the `-incubating` part.  I feel that it is acceptable.

releasey/04-build-and-test.sh
pingtimeout: TIL the GH API endpoint.  Let me look into this.  It would certainly be faster to do it that way.  And also it would enforce the fact that the commit must be validated by CI, as opposed to what the current `--commit` parameter allows.

releasey/04-build-and-test.sh
pingtimeout: I agree with you on the unnecessary aspect of those commands.  That being said, I think that until we develop the habit of only releasing via GH workflows, _it is possible_ that a release manager performs a release from their machine.

I am going to add a check that ensures there is no uncommitted change in the repo.  That way, combined with the verification in GH API that the commit has been validated by CI, we should be good.

Both the "working directory is clean" and `./gradlew clean` steps will become no-op in GH workflows anyway.

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,186 @@
...
+if ! git rev-parse --verify "${commit}" >/dev/null 2>&1; then
+  print_error "Invalid Git commit: ${commit}"
+  usage >&2
+  exit 1
+fi
+
+# Validate version format: x.y.z-incubating
+# TODO: Remove incubating when we are a TLP
+version_regex="^([0-9]+)\.([0-9]+)\.([0-9]+)-incubating$"
+if [[ ! ${version} =~ ${version_regex} ]]; then
```
adutra: Aren't we going to publish RCs?

releasey/libs/_exec.sh
```diff
@@ -0,0 +1,39 @@
...
+libs_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+source "$libs_dir/_constants.sh"
+source "$libs_dir/_log.sh"
+
+function exec_process {
+  if [[ ${DRY_RUN:-1} -ne 1 ]]; then
+    print_command "Executing '${*}'"
+    "$@"
+  elif { >&3; } 2>/dev/null; then
```
adutra, line 35: Took me a while to realize that file descriptor 3 is an optional log file. Would it be possible to use named file descriptors instead?

releasey/06-build-and-stage-docker-images.sh
pingtimeout: Good point, implementing this...

releasey/06-build-and-stage-docker-images.sh
pingtimeout: I could but it would not bring any benefit.  We would have a constant for the incubating suffix, sure.  Which means that as soon as Polaris graduates, that constant becomes an empty string and should be removed entirely from the scripts.  This means that instead of having to remove `-incubating` everywhere, we would have to remove `${INCUBATING_SUFFIX}` everywhere.  Are you sure we should do that?

releasey/06-build-and-stage-docker-images.sh
pingtimeout: Wdym a proposal for the subject and body of the e-mail?

I don't think we need a shell script just to run `./gradlew releaseEmailTemplate`.  Are you talking about anything else?

releasey/libs/_constants.sh
pingtimeout: That is part of the release guide.

> To send the key to the Ubuntu key-server, Apache Nexus needs it to validate published artifacts:
> 
> gpg --keyserver hkps://keyserver.ubuntu.com --send-keys <YOUR KEY ID HERE>

releasey/libs/_constants.sh
pingtimeout: So, on one hand what you are saying is perfectly valid and we should use `origin`.  On the other hand, the release guide in its current state has been written considering that individuals may be performing releases and refers to it as `apache`

> The release should be executed against https://github.com/apache/polaris.git repository (not a fork). Set it as remote with name apache for release if it’s not already set up.

With the current variable, we can override it in a GH workflow to `origin`.  And once we completely adopt the GH workflow based releasing process, we can switch the default value.

releasey/02-create-release-branch.sh
pingtimeout: I am not sure I understand the question.  AFAICT, we are going to create a single branch per release.  I.e. we won't create a release branch for different RC of the same version.

releasey/libs/_docker.sh
pingtimeout: Do you think we should just assume docker hub login is properly configured?  I wanted this to be a quick and easy check.  But if it turns out to create more friction than benefits, let's drop it.

releasey/libs/_exec.sh
pingtimeout: I cannot seem to get something working with named FD.  Ideally, test scripts should be able to use them as follows:

```
DRY_RUN=1 \
  "${releases_dir}/02-create-release-branch.sh" \
  --version 42.41.40-incubating \
  {COMMANDS_LOG}>"$temp_file"
```

Instead of

```
DRY_RUN=1 \
  "${releases_dir}/02-create-release-branch.sh" \
  --version 42.41.40-incubating \
  3>"$temp_file"
```

But I cannot seem to get the underlying script to recognize `COMMANDS_LOG`.  Do you have an example for a main script that can set a named fd for a subscript?

releasey/libs/_files.sh
pingtimeout: Good point, removing that.  Just relying on gradlew existence is enough.

releasey/libs/_files.sh
pingtimeout: True, with the downside that now the scripts have to `pushd/popd` in order to run any gradlew command.  Let me check if I can get that changed in a not too-invasive way.  If not, it's probably ok as-is given that GH Workflow will be the main execution flow for the scripts.

releasey/libs/_log.sh
pingtimeout: Ack.  Fixing this.

releasey/libs/_log.sh
pingtimeout: Ack.  Fixing this.

releasey/libs/_log.sh
pingtimeout: https://github.com/apache/polaris/blob/main/regtests/README.md#prerequisites

releasey/libs/_nexus.sh
pingtimeout: That's what the release guide says.  Could you open an issue to track that?

releasey/libs/_nexus.sh
pingtimeout: Let's do this as soon as the release guide is updated.

releasey/02-create-release-branch.sh
adutra: My comment is just because the regex doesn't allow versions like `1.2.3-rc4-incubating`.

releasey/libs/_exec.sh
adutra: I think it probably doesn't work for inter-script references. Let's drop the idea, numeric descriptors is fine.

releasey/02-create-release-branch.sh
pingtimeout: This is actually correct (and intended).  Let me explain.  The goal is to make it clear that a release branch should only be created for an `x.y.z` version.  So the release branch script should only be called with a version that does not include rc.

That being said, now I am having second thoughts.  If we trigger releases via GH workflows, this script will always be invoked, right?  We won't have a workflow for "cut RC1" and another workflow for "cut another RC".  So maybe I should update this script to enable that and only create the release branch if this is about RC1...  Let me see what I can come up with.

pingtimeout: See the last commit, now the script accepts versions like `apache-polaris-1.1.0-incubating-rc1` and `apache-polaris-1.1.0-incubating-rc2` but only does something for rc1.

releasey/libs/_nexus.sh
snazy: #2173

releasey/libs/_constants.sh
snazy: `apache` is the remote name on someone's machine, not on everybody's Git clones.

Strictly speaking, the release guide is broken, as you cannot expect that the `apache` Git remote name is present ;)

releasey/libs/_constants.sh
snazy: Uh, doesn't the ASF have a keyserver?

releasey/06-build-and-stage-docker-images.sh
snazy: Nah - the idea is to change the constant definition
```
INCUBATING_SUFFIX="-incubating"
```
to
```
INCUBATING_SUFFIX=""
```
in one place when it happens.

releasey/06-build-and-stage-docker-images.sh
snazy: Actually, I'm thinking of removing the `releaseEmailTemplate` (and related) tasks and do that in the release scripts. Mean, the scripts already have all the information, so the indirection via Gradle, which has to figure out the details again, isn't necessary.

releasey/02-create-release-branch.sh
snazy: > My comment is just because the regex doesn't allow versions like 1.2.3-rc4-incubating.

Simply speaking: The GA version would be `1.2.3-incubating`. The 42nd RC would therefore be `1.2.3-incubating-rc42` (`-incubating` stays where it is.

releasey/libs/_constants.sh
pingtimeout: I will clarify that with JB during the 1.0.1 release and update accordingly.  In any case, it is possible to use the current scripts and override APACHE_REMOTE_NAME=origin as environment variable if needed.

releasey/04-build-and-test.sh
pingtimeout: @snazy Just to be sure, could you help me confirm that "Combined status" is not the right thing for what we are trying to do?  When I look at the combined status of any commit from `main`, the `state` property is always `pending`.

Now, digging around, I found that there is a REST API endpoint to [verify the "Checks" that have been run against a particular commit](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference).

Example for a commit of this PR where `./gradlew rat` failed one of the checks: there is one `failure`.

```
$ curl -s -H "Authorization: Bearer ${GITHUB_REPO_STATUS_TOKEN}" -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/apache/polaris/commits/5d0c1c2/check-runs | jq '.check_runs[].conclusion'
"skipped"
"failure"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
```

And for a Renovate commit that was just pushed to `main`: everything is `success`.

```
$ curl -s -H "Authorization: Bearer ${GITHUB_REPO_STATUS_TOKEN}" -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/apache/polaris/commits/98d42205e/check-runs | jq '.check_runs[].conclusion'
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
"success"
```

So I am thinking that this verification should be "the conclusion of all checks should be `success`".  Is that what you had in mind?

snazy: Ah, I recall now. Statuses and check-runs are different things. And there's no "combined check-runs conclusion".
Yes, you'd need to "manually combine" via all `conclusion` fields (not the `status` field). 

"Successful conclusions are both `success` _and_ `skipped`, everything else **including `null`** is in this use case a failure.

The annoying thing about check-runs is that the `name` of the check-runs is only the name of the job of a workflow, not the "full name" (with the name of the workflow). I think, that's okay-ish to only say that some check failed.

Considering all check-runs of a commit is probably okay for now, as all checks are required. However, if there is ever a check (like a workflow job) that is _not_ required and that fails, the "ready-to-release" verification would fail as well, because it's a bit more complicated to know which checks are required by a branch protection rule.

releasey/libs/_log.sh
pingtimeout: I realize that there is some misunderstanding here.  This is an example that shows that, when invoking a `script.sh` (a test script, really), and redirect fd `3` to `/tmp/script.log`, then that file contains all the commands that the script would execute.  It is just an example, don't read too much into that.

releasey/04-build-and-test.sh
pingtimeout: Added.  Now all checks must either have the `success` or `skipped` conclusion.  I don't like allowing skipped, but I understand some checks may be skipped if some parts of the codebase have not been touched since the last check?

releasey/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+This directory contains scripts to automate the Apache Polaris release process, following the [official release guide](https://github.com/apache/polaris/blob/main/docs/release-guide.md).
+
+## Release Scripts
+
+These scripts are used to perform actual releases:
+
+### 1. Prerequisites Verification
+```bash
+./01-prerequisites.sh
```
flyrain, line 49: I think this should be `./releasey/01-prerequisites.sh`, otherwise, it always throws:
```
ERROR: This script must be executed from the project root directory.
ERROR: Current directory: /Users/ygu/asf/polaris/releasey
```

This also applies to other commands.

releasey/libs/_files.sh
pingtimeout: Fixed

releasey/04-build-and-test.sh
pingtimeout: I added a check in `01-prerequisites.sh` that ensures the working directory is clean (`git status --porcelain`)

releasey/README.md
pingtimeout: Just a heads up, I updated the code so that it does not require to be at the project root anymore.  So this works now.

pingtimeout:
I updated the code to include the current changes being discussed in the Helm PR and to integrate code review feedback.  I am going to update the PR description with output example so that understanding this PR is much easier.

pingtimeout:
@snazy @flyrain @adutra This PR is already quite large, so I am thinking about postponing any Github-workflow related work.  We could iron out these scripts for the August release and iterate on them while we also polish the release process.  Feel free to review at your convenience, the code is ready now.

flyrain:
Thanks @pingtimeout , will take a look today!

releasey/README.md
```diff
@@ -0,0 +1,131 @@
...
+Example:
+
+- `DRY_RUN=0` - Disable dry-run mode (default: enabled)
+- `APACHE_REMOTE_NAME=my-apache-remote` - Name of git remote where release branches and tags are pushed (default: "apache")
+- `KEYSERVER` - GPG keyserver URL (default: "hkps://keyserver.ubuntu.com")
+- ...
+
+See the content of `libs/_constants.sh` for the list of all environment variables and their default values.
+
+## Prerequisites
```
flyrain: nit: can we move this section to the top, so that users can set them before running any script?

releasey/README.md
```diff
@@ -0,0 +1,131 @@
...
+```
+Verifies that your environment is ready for releases:
+- GPG setup and key configuration
+- Maven/Gradle credentials
+- Git remote setup for Apache repository
+- GitHub token for test status verification
+
+### 2. Create Release Branch
+```bash
+./02-create-release-branch.sh --version 1.0.0-incubating [--commit HEAD] [--recreate]
```
flyrain: Nit: I think the version string needs `rcN`, it will error out otherwise.
```
./02-create-release-branch.sh --version 1.0.0-incubating-rc1
```

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,201 @@
...
+update_version "${polaris_version}"
+
+print_info "Committing and pushing version change"
+exec_process git add "$VERSION_FILE" "$HELM_CHART_YAML_FILE" "$HELM_README_FILE" "$HELM_VALUES_FILE"
+exec_process git commit -m "[chore] Bump version to ${polaris_version} for release"
+exec_process git push
+
+print_info "Updating CHANGELOG.md"
+exec_process cd "${releasey_dir}/.."
+exec_process ./gradlew patchChangelog
```
flyrain, line 187: This script exists at line 139 in case of RC2+. Should we re-run the changelog for any RC2+?

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,201 @@
...
+  fi
+fi
+
+print_info "Starting release branch creation..."
+print_info "Version: ${version}"
+print_info "Polaris version: ${polaris_version}"
+print_info "From commit: ${commit}"
+echo
+
+# Check if release branch already exists
```
flyrain, line 150: +1

releasey/03-create-release-candidate-tag.sh
```diff
@@ -0,0 +1,142 @@
...
+  previous_rc=$((rc_number - 1))
+  previous_tag="apache-polaris-${version%-rc*}-rc${previous_rc}"
+  
+  print_info "Checking for previous RC tag: ${previous_tag}"
+  if ! git tag -l "${previous_tag}" | grep -q "^${previous_tag}$"; then
+    print_error "Previous RC tag ${previous_tag} does not exist. Cannot create RC${rc_number} without RC${previous_rc}."
+    exit 1
+  fi
+  print_info "Previous RC tag ${previous_tag} found"
+fi
```
flyrain: minor: Do we have to care if the previous tag exists or not? what if the previous RC tag has been deleted manually already?

releasey/04-build-and-test.sh
```diff
@@ -0,0 +1,90 @@
...
+  esac
+done
+
+print_info "Starting build and test process..."
+echo
+
+# Clean and build Polaris
+print_info "Building Polaris..."
+exec_process cd "${releasey_dir}/.."
+exec_process ./gradlew clean build
```
flyrain, line 77: Question: Is this script file needed if all it does is to execute this single command?
Can we just put something like this in the README.md?
```
### 4. Build and Verify Tests `./gradlew clean build`
```

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+exec_process svn add "${version_dir}"
+exec_process svn add "${helm_chart_version_dir}"
+
+print_info "Committing changes..."
+exec_process svn commit -m "Stage Apache Polaris ${polaris_version} RC${rc_number}"
+
+print_info "Updating Helm index..."
+exec_process cd "${dist_dev_dir}/helm-chart"
+exec_process helm repo index .
+exec_process cd "${releasey_dir}/.."
```
flyrain, line 137: We will need to commit change after helm indexing, otherwise, the index file is not updated.

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+echo
+print_success "🎉 Distributions built and staged successfully!"
+echo
+print_info "Staged artifacts:"
+print_info "- Source and binary distributions: ${version_dir}"
+print_info "- Helm charts: ${helm_chart_version_dir}"
+print_info "- Maven artifacts: Published to Apache staging repository"
+echo
+print_info "Next steps:"
+print_info "1. Close the staging repository in Nexus"
```
flyrain, line 152: Is there a way to automate this part? Otherwise, we may call it out here, that release manager has to sign in the [Nexus website](https://repository.apache.org/#stagingRepositories) with their ASF credential to close it.

flyrain (commented):
Thanks @pingtimeout for doing this. LGTM overall with a few comments.

releasey/02-create-release-branch.sh
pingtimeout: That's a good question.  I am not 100% sure about the answer, to be honest.  The [release guide](https://polaris.apache.org/community/release-guide/) says that the changelog should be manually updated:

> Manually add an -rcN suffix to the previously generated versioned CHANGELOG section.
> Rerun the patchChangelog command
> Manually remove RC sections from the CHANGELOG
> Submit a PR to propagate CHANGELOG updates from the release branch to main.

And IIRC there are still some discussions on changelog updates, whether those updates should be merged to main, etc...

This is the reason why the script exits for RC>1.

My goal with this PR is to have a 1:1 automation of what the release guide says, so that we can start having some semi-automated (albeit imperfect) release process for the August release.  But I suspect we will iterate on this as we learn more about our practices.

releasey/03-create-release-candidate-tag.sh
pingtimeout: I don't think this should be a concern.  My rationale is that no RC tag should ever be deleted.  I don't think it is allowed per ASF principles, as it means a release candidate, for which a vote was called, could disappear.

Wdyt?

releasey/05-build-and-stage-distributions.sh
pingtimeout: There should be a way to automate that, indeed.  The grand vision is to have a fully automated release process that runs in Github Workflows.  However there are some things that needs discussed before this can happen.  Things like :

* Should we trigger when a Git tag is created or when an empty Github Release with a specific tag pattern is created, ...
* As you pointed out in another comment, should we (can we?) automatically update the Changelog?
* Should we implement fixes on branches of the supported release(s) first and then merge those branches up to main (in which case the changelog updates are easy)?  Or do we want to cherry-pick commits to add certain fixes to release branches (in which case it requires manual updates)?
* Does the ASF supports automatically sending announcement/vote/cancel e-mails?  In which case this could be embedded as part of the release automation
* Etc...

As you can see, we are not quite there yet.  I suggest we start with the current scripts, which provide a semi-automated model.  And in parallel, we can start discussion to achieve the big goal.

releasey/04-build-and-test.sh
pingtimeout: I don't think this command runs the regression tests, though.  Does it?  The script also contains a verification step against Github, see L82 below.  The verification ensures that there is no failed Github check, which gives us the guarantee that regression tests passed.

Does this make sense?

releasey/05-build-and-stage-distributions.sh
pingtimeout: Good point, let me try a few things locally and update the script.

releasey/06-build-and-stage-docker-images.sh
pingtimeout: My point, exactly.  This would mean that everywhere in the scripts, instead of having:

```
version="${major}.${minor}.${patch}-incubating"
```

We would have

```
version="${major}.${minor}.${patch}-${INCUBATING_SUFFIX}"
```

The `INCUBATING_SUFFIX` variable would be present all over the place, even though it will serve no purpose at all after graduation.  Hence, no readability benefit whatsoever.

releasey/README.md
pingtimeout: Good point, fixing...

releasey/README.md
pingtimeout: Fixing...

releasey/02-create-release-branch.sh
flyrain: I'm fine to improve it later. cc @dimas-b and @snazy, who has more experience with changelog

releasey/03-create-release-candidate-tag.sh
flyrain: I didn’t see anything in the ASF principles that explicitly calls that out—let me know if you have a link. If it’s not covered in ASF policy, the main reason to keep RC tags would just be for historical reference. That said, folks might still delete them to keep the tag list clean.

In any case, this isn’t a blocker. My thinking is that if the behavior isn’t defined, we don’t need previous tag checks.

releasey/05-build-and-stage-distributions.sh
flyrain: Agreed to automate it later. Can we provide a bit more detailed message how to close it? Just in case people get lost.
```suggestion
print_info "1. close the staging repository in Nexus manually per instructions in https://polaris.apache.org/community/release-guide/ "
```

releasey/README.md
```diff
@@ -0,0 +1,131 @@
...
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Polaris Release Automation Scripts
+
+This directory contains scripts to automate the Apache Polaris release process, following the [official release guide](https://github.com/apache/polaris/blob/main/docs/release-guide.md).
```
flyrain: the link is broken.

releasey/libs/_version.sh
```diff
@@ -0,0 +1,97 @@
...
+  # This function validates the format of a release candidate version and extracts its components (major.minor.patch and rc number).
+  # It returns 0 if the version is valid and sets the global variables major, minor, patch, and rc_number.
+  # Otherwise, it returns 1.
+  local version="$1"
+
+  if [[ ! ${version} =~ ${VERSION_REGEX_RC} ]]; then
+    return 1
+  fi
+
+  major="${BASH_REMATCH[1]}"
```
snazy, line 41: Wonder whether the global variables should be explicitly declared in a "lib", commenting where/how those are populated. WDYT?

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,201 @@
...
+
+# Validate that the commit exists
+if ! git rev-parse --verify "${commit}" >/dev/null 2>&1; then
+  print_error "Invalid Git commit: ${commit}"
+  usage >&2
+  exit 1
+fi
+
+# Validate version format: x.y.z-incubating-rcN
+# TODO: Remove incubating when we are a TLP
```
snazy, line 122: Guess the TODO can go away, or have it everywhere (which isn't great either).
Maybe have a constant `VERSION_INCUBATING="-incubating"`, so there's only one place to change in that case.

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,201 @@
...
+exec_process git commit -m "[chore] Bump version to ${polaris_version} for release"
+exec_process git push
+
+print_info "Updating CHANGELOG.md"
+exec_process cd "${releasey_dir}/.."
+exec_process ./gradlew patchChangelog
+
+print_info "Committing changelog change"
+exec_process git add "$CHANGELOG_FILE"
+exec_process git commit -m "[chore] Update changelog for release"
```
snazy, line 191:
```suggestion
exec_process git commit -m "[chore] Update changelog for ${version}"
```

releasey/02-create-release-branch.sh
snazy: TL;DR `CHANGELOG.md` has to be (re)generated for every version/RC/branch independently.

Content of `CHANGELOG.md` is maintained manually w/ the PRs being merged, it's the contributor's/reviewer's who update `CHANGELOG.md`. So changes to that file already exist and nothing has to be "merged back", and in fact cannot as the contents on "all the branches" can be different.

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,201 @@
...
+exec_process git push
+
+echo
+print_success "🎉 Release branch ${release_branch} created successfully!"
+echo
+print_info "Next steps:"
+print_info "* Submit a PR on main branch to propagate CHANGELOG updates from the release branch to main (https://github.com/${APACHE_REMOTE_NAME}/polaris/pull/new/${release_branch})."
```
snazy: Let's remove this, see my comment above.

releasey/03-create-release-candidate-tag.sh
```diff
@@ -0,0 +1,142 @@
...
+
+print_info "Starting release candidate tag creation..."
+print_info "Version: ${version}"
+print_info "Polaris version: ${polaris_version}"
+print_info "RC number: ${rc_number}"
+echo
+
+# If we're creating RC2 or later, verify previous RC tag exists
+if [[ ${rc_number} -gt 1 ]]; then
+  previous_rc=$((rc_number - 1))
```
snazy, line 104: Nit: `local`

releasey/03-create-release-candidate-tag.sh
```diff
@@ -0,0 +1,142 @@
...
+  previous_tag="apache-polaris-${version%-rc*}-rc${previous_rc}"
```
snazy: Super nit: maybe extract `${version%-rc*}` into a separate variable, just for clarity.

releasey/03-create-release-candidate-tag.sh
snazy: Deleting tags is not good practice and should really never happen, it breaks things.
Tags are referred from released artifacts/poms, that alone actually forbids deleting those.

releasey/04-build-and-test.sh
snazy, line 77: Might check for a clean working copy a `[[ -z "$(git status --porcelain)" ]]` for extra safety.

snazy, line 77: You actually only need
```suggestion
exec_process ./gradlew assemble
```

No need to run tests (with `build``) as the Git commit status is verified elsewhere.

releasey/04-build-and-test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+current_commit=$(git rev-parse HEAD)
+print_info "Current commit: ${current_commit}"
+
+if ! check_github_checks_passed "${current_commit}"; then
+  print_error "Integration and regression tests have not all passed for commit ${current_commit}"
+  print_error "Please ensure all CI workflows are successful before proceeding with the release"
+  print_info "You can check the status at: https://github.com/${GITHUB_REPO_OWNER}/${GITHUB_REPO_NAME}/commit/${current_commit}"
+  exit 1
+fi
```
snazy, line 87: Might want to move this before the Gradle build (why build something that's known to not pass CI).

releasey/04-build-and-test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+echo
+print_success "🎉 Build and test verification completed successfully!"
```
snazy, line 90: No next steps here? ;)

releasey/01-prerequisites.sh
```diff
@@ -0,0 +1,135 @@
...
+# 2. Maven/Gradle credentials verification
+# 3. Git remote setup verification
+# 4. Docker setup verification (for Docker image releases)
+#
+# Returns non-zero exit code if any setup verification fails.
+#
+
+set -euo pipefail
+
+releasey_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
snazy, line 35:
```suggestion
releasy_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
;)

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+
+print_info "Starting build and stage distributions process..."
+print_info "Version: ${polaris_version}"
+print_info "RC number: ${rc_number}"
+echo
+
+# Build distributions
+print_info "Building source and binary distributions..."
+exec_process cd "${releasey_dir}/.."
+exec_process ./gradlew build sourceTarball -Prelease -PuseGpgAgent -x test -x intTest
```
snazy, line 96:
```suggestion
exec_process ./gradlew sourceTarball -Prelease -PuseGpgAgent
```
No need to repeat what's been done in `04-...sh`

releasey/04-build-and-test.sh
snazy: Technically, you rather want 
```suggestion
exec_process ./gradlew assemble -Prelease
```
here, likely also with ` -PuseGpgAgent`

releasey/05-build-and-stage-distributions.sh
snazy: Also note that `test` and `intTest` are not conventions, there are more test tasks, so using `build` is rather discouraged.

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+
+# Create Helm package
+print_info "Creating Helm package..."
+exec_process cd "${releasey_dir}/../helm"
+exec_process helm package polaris
+exec_process helm gpg sign polaris-${polaris_version}.tgz
```
snazy: `helm gpg sign` implicitly uses the GPG agent, if necessary? I suspect the answer is "yes".

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+exec_process gpg --armor --output polaris-${polaris_version}.tgz.asc --detach-sig polaris-${polaris_version}.tgz
+calculate_sha512 polaris-${polaris_version}.tgz.prov polaris-${polaris_version}.tgz.prov.sha512
+exec_process gpg --armor --output polaris-${polaris_version}.tgz.prov.asc --detach-sig polaris-${polaris_version}.tgz.prov
+
+# Stage to Apache dist dev repository
+print_info "Staging artifacts to Apache dist dev repository..."
+
+dist_dev_dir=${releasey_dir}/polaris-dist-dev
+print_info "Checking out ${APACHE_DIST_URL}${APACHE_DIST_PATH} to ${dist_dev_dir}..."
+exec_process svn co "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
```
snazy, line 113:
```suggestion
exec_process svn checkout "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
```
Nit: use full names for clarity

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+
+print_info "Copying files to destination directory..."
+version_dir="${dist_dev_dir}/${polaris_version}"
+helm_chart_version_dir="${dist_dev_dir}/helm-chart/${polaris_version}"
```
snazy: Extra safety:
Remove existing files in `version_dir` and `helm_chart_version_dir`, in case something temporary was left there before.

snazy: Actually we should move the helm-chart into `version_dir`, so we can clearly distinguish two parallel release processes (say: 4.2 and 5.0). And it would be "safe" to if there's something else than helm-charts that deserves it's own subdirectory.

releasey/05-build-and-stage-distributions.sh
```diff
@@ -0,0 +1,155 @@
...
+exec_process svn commit -m "Stage Apache Polaris ${polaris_version} RC${rc_number}"
+
+print_info "Updating Helm index..."
+exec_process cd "${dist_dev_dir}/helm-chart"
+exec_process helm repo index .
+exec_process cd "${releasey_dir}/.."
+
+# Publish Maven artifacts
+print_info "Publishing Maven artifacts to Apache staging repository..."
+exec_process ./gradlew publishToApache -Prelease -PuseGpgAgent -Dorg.gradle.parallel=false
```
snazy, line 141: The Nexus staging repository URL + ID will be required.

You can get those from the Gradle output (`tee` to a log file) and grep for `Created staging repository`.
`s/^Created staging repository .([a-z0-9-]+). at (.*)/\1/` gives you the staging repo ID, `\2` the URL.

I'd also suggest to write both values to files in the Git tree and commit & push those to be included in the Git tag, so that data isn't lost. Maybe into `releasy/current-release/nexus-repo-id` and `releasy/current-release/nexus-repo-url`?

releasey/05-build-and-stage-distributions.sh
snazy: I propose to already start with semi-automatic releases. The required GH workflows require the scripts to be different, leading to two different implementations that have to be both verified and kept in sync, which is quite a big effort.

releasey/06-build-and-stage-docker-images.sh
snazy: I'd prefer to avoid the `-incubating` string everywhere, the "empty" constant use can be cleaned up later. Or maybe another project borrows our scripts?

releasey/libs/_constants.sh
snazy: I think sticking to Git defaults is better, the name `apache` is a custom/personal setting.

releasey/libs/_docker.sh
snazy: Yea - let's drop it.
How a login happens and particularly whether and how that appears in `docker info` is not standardized at all. It can actually be a very different login (different repo), so the check isn't that helpful.

releasey/02-create-release-branch.sh
```diff
@@ -0,0 +1,201 @@
...
+    exit 1
+  fi
+fi
+
+print_info "Starting release branch creation..."
+print_info "Version: ${version}"
+print_info "Polaris version: ${polaris_version}"
+print_info "From commit: ${commit}"
+echo
+
```
snazy, line 149: I guess a `git fetch [remote]` doesn't hurt here, just to be safe.

releasey/03-create-release-candidate-tag.sh
pingtimeout: Ack that this is not a blocker 👍

Let's rediscuss this when JB is back from holidays.  I think that's a fair question.

releasey/libs/_github.sh
```diff
@@ -0,0 +1,154 @@
...
+
+function ensure_no_uncommitted_change() {
+  if [[ -n $(git status --porcelain) ]]; then
+    print_error "There are uncommitted changes in the repository."
+    print_error "A release can only be performed from a clean state."
+    return 1
+  fi
+}
+
+function add_apache_remote() {
```
snazy, line 61: Not a fan of this.
Having multiple remotes, especially those that point to the exact same repo, break developer experience in IDEs. Things appear twice for example.

releasey/libs/_github.sh
```diff
@@ -0,0 +1,154 @@
...
+function ensure_github_token_is_configured() {
+  # Ensure that GitHub token is configured, return non-zero if not
+  print_info "Checking GitHub token configuration..."
+
+  if check_github_token; then
+    print_success "GitHub token is configured"
+    return 0
+  else
+    print_error "GITHUB_TOKEN environment variable is not set"
+    print_error "A GitHub token is required to check CI status for releases"
```
snazy, line 124: It's technically not required. The "anonymous" rate limit should be sufficient for a few runs of the script (Commit status checks). Setting up a GITHUB_TOKEN may lead to unnecessarily creating PATs or the like.
I suggest to change this to a warning.

releasey/libs/_github.sh
snazy: You can figure out the "right" remote name via `git remote -v` and looking for `https://github.com/apache/polaris.git`.

releasey/04-build-and-test.sh
pingtimeout: Actually I think this is a good point.  The 05 build-and-stage-distributions script includes `./gradlew build sourceTarball -Prelease -PuseGpgAgent -x test -x intTest`.  So we could reduce the 04 build-and-test script to just checking the status of the Github checks for the SHA that is about to be released.  Let me do that.

releasey/libs/_gpg.sh
```diff
@@ -0,0 +1,188 @@
...
+
+function get_secret_keys() {
+  # Get list of secret keys (private keys) that can be used for signing
+  gpg --list-secret-keys --with-colons 2>/dev/null | grep '^sec:' | cut -d: -f5
+}
+
+function get_gradle_signing_key_id() {
+  # Get the signing key ID from gradle.properties
+  grep '^signing.gnupg.keyName' "${GRADLE_PROPERTIES_FILE}" 2>/dev/null | cut -d'=' -f2
+}
```
snazy, line 50: Having this setting isn't technically required. It's rather the release manager's responsibility to ensure that the key is properly setup. Nothing prevents release managers to configure only one GPG key at all and/or set `default-key` in `~/.gnupg/gpg.conf`.

snazy: It also forces a _global_ setting, which can easily interfere with other stuff that people have to do with Gradle signing. So this actually adds burden to people who sign artifacts with Gradle for multiple projects.

snazy: For (locally executed) scripts, you can list the available secrets and explicitly ask the release manager to explicitly confirm that their settings are correct.

snazy: This won't be an issue in GH workflows, where the workflows do the right thing anyways.

releasey/04-build-and-test.sh
pingtimeout: It is already done as part of prerequisites

releasey/05-build-and-stage-distributions.sh
pingtimeout: Correct

releasey/02-create-release-branch.sh
flyrain: Just curious, if nothing has to be "merged back", why do we need this command?
```
exec_process ./gradlew patchChangelog
```

releasey/03-create-release-candidate-tag.sh
pingtimeout: local can only be used in a function

releasey/03-create-release-candidate-tag.sh
pingtimeout: Actually it already exists as the `polaris_version` variable, using that instead

pingtimeout:
@flyrain @snazy I think there is some misunderstanding.  I believe that only PMC members have write access to SVN distribution directory as well as permissions to create Nexus repositories.  This makes testing and implementation tricky for me, for steps ≥05.  Typically I can see this in the email template that is generated:

```
Convenience binary artifacts are staged on Nexus. The Maven repository URL is:
* NO STAGING REPOSITORY (java.lang.IllegalStateException: No staging repository with name apache created)
```

I can only rely on the information that is in the release guide, or on the hints I can find in the build file (like assuming that the `./gradlew initializeApacheStagingRepository` command does the right thing for PMC members).

snazy:
> @flyrain @snazy I think there is some misunderstanding. I believe that only PMC members have write access to SVN distribution directory as well as permissions to create Nexus repositories. This makes testing and implementation tricky for me, for steps ≥05. Typically I can see this in the email template that is generated:
> 
> ```
> Convenience binary artifacts are staged on Nexus. The Maven repository URL is:
> * NO STAGING REPOSITORY (java.lang.IllegalStateException: No staging repository with name apache created)
> ```
> 
> I can only rely on the information that is in the release guide, or on the hints I can find in the build file (like assuming that the `./gradlew initializeApacheStagingRepository` command does the right thing for PMC members).

The message likely refers to the Nexus repo, not SVN?

releasey/01-prerequisites.sh
pingtimeout: Let me mark this as resolved for now given that the folder name is `releasey` and not `releasy`.

releasey/libs/_constants.sh
RussellSpitzer: The "apache" here makes sense to me because most folk's "origin" is their own fork of the repository while "apache" is this one. At least thats how I usually configure things and I think I learned that habit from other Apache projects I've forked and worked on.

pingtimeout:
@snazy This message says that I cannot test as many things as PMC members, and that I can only rely on the information in the current release guide.  So if a PMC member requests changes that deviate from the release guide, I have to take the request at face value.

pingtimeout:
Just a heads up, I am giving a try at a Github-workflow based release automation.  Hopefully this should reduce the amount of code, given that it will only be runnable by Github.  If it fails, we can always come back to these scripts.

releasey/libs/_github.sh
RussellSpitzer: I agree there shouldn't be an "add remote" function. But i'd also note that the git -v output can also be a little tricky since if you don't use https 
It would look like
```
apache	git@github.com:apache/polaris.git (fetch)
apache	git@github.com:apache/polaris.git (push)
```

I think you are safe if you just look for `apache/polaris.git`

releasey/libs/_exec.sh
```diff
@@ -0,0 +1,40 @@
...
+# Execution wrapper function for system-modifying commands
+#
+
+LIBS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+source "$LIBS_DIR/_constants.sh"
+source "$LIBS_DIR/_log.sh"
+
+function exec_process {
+  # See print_command for information on why fd 3 is used.
```
RussellSpitzer, line 31: I'd probably just copy the note from print_ command to here since it's not exactly easy to jump between the two files

releasey/libs/_github.sh
```diff
@@ -0,0 +1,154 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Git Remote Setup Script
+#
+# This script ensures that the Git repository has the proper remote configuration for Apache Polaris releases:
+# 1. Checks whether there is a Git remote named 'apache' pointing to https://github.com/apache/polaris.git
```
RussellSpitzer, line 25: I think it's probably better to just work backwards from the assumption "we need the correct target" and not require the other conditions.

1. Checks remotes
2. If one doesn't exist with the correct URI for apache/polaris, error out
3. If it does exist return the name of the remote which will be used

releasey/libs/_constants.sh
```diff
@@ -0,0 +1,55 @@
...
+
+LIBS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# GPG constants
+KEYS_URL=${KEYS_URL:-"https://downloads.apache.org/incubator/polaris/KEYS"}
+KEYSERVER=${KEYSERVER:-"hkps://keyserver.ubuntu.com"}
+
+# Git/SVN repository constants
+APACHE_REMOTE_NAME=${APACHE_REMOTE_NAME:-"apache"}
+APACHE_REMOTE_URL=${APACHE_REMOTE_URL:-"https://github.com/apache/polaris.git"}
```
RussellSpitzer, line 29: This is only valid for HTTPS clones, a ssh clone would look like

`git@github.com:apache/polaris.git`

So it may be good to keep this just to "apache/polaris.git" since I believe we are only using this for validation?

releasey/libs/_github.sh
```diff
@@ -0,0 +1,154 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Git Remote Setup Script
```
RussellSpitzer, line 22: Quick note on this doc, this script also has a bunch of other utility methods for checking other project specific statuses. I have no problem with the organization but we should probably change the doc here to note that this also contains other utility functions for working with git. I also wouldn't be opposed if you wanted to separate out the self executing bits from the utility bits but honestly I have no strong feelings on that.

releasey/test/test-04-build-and-test.sh
```diff
@@ -0,0 +1,101 @@
...
+print_info "Verifying output content..."
+# Read the actual content
+actual_content=$(cat "$temp_file")
+
+expected_content="cd ${releasey_dir}/..
+./gradlew clean build
+curl -s -H \"Authorization: Bearer dummy_token\" -H \"Accept: application/vnd.github+json\" -H \"X-GitHub-Api-Version: 2022-11-28\" https://api.github.com/repos/apache/polaris/commits/$(git rev-parse HEAD)/check-runs | jq -r '[.check_runs[].conclusion | select(. != \"success\" and . != \"skipped\")] | length'"
+
+# Compare content
+if [[ "$actual_content" == "$expected_content" ]]; then
```
RussellSpitzer, line 92: could probably extract this out into a function since it's used in all the regressions if you like

RussellSpitzer:
> Just a heads up, I am giving a try at a Github-workflow based release automation. Hopefully this should reduce the amount of code, given that it will only be runnable by Github. If it fails, we can always come back to these scripts.

That's pretty awesome too although I like the idea of having a set of scripts that could potentially work locally as well but I'd hate to put more work on you, so if it's simpler to just do the Github Workflow approach that seems very reasonable.

pingtimeout:
@RussellSpitzer thanks for the review !

> I like the idea of having a set of scripts that could potentially work locally as well

My initial goal was that these scripts would be short-lived and replaced as soon as we had a robust release guide that paved the way for a Github workflow automation.

As Robert hinted, it is already quite complex to deal with Linux/MacOS differences (hello GNU `sed` vs BSD `sed`, etc...) as well as custom setups.  I think that as soon as we have a set of Github Workflows for releases, we should delete the scripts and keep the codebase as small as possible.

I am not saying that nobody should ever be able to perform a release locally.  What I am saying is that it will be rare enough that we should not be optimizing for it.

flyrain:
> @flyrain @snazy I think there is some misunderstanding. I believe that only PMC members have write access to SVN distribution directory as well as permissions to create Nexus repositories. This makes testing and implementation tricky for me, for steps ≥05. Typically I can see this in the email template that is generated:
> 
> ```
> Convenience binary artifacts are staged on Nexus. The Maven repository URL is:
> * NO STAGING REPOSITORY (java.lang.IllegalStateException: No staging repository with name apache created)
> ```
> 
> I can only rely on the information that is in the release guide, or on the hints I can find in the build file (like assuming that the `./gradlew initializeApacheStagingRepository` command does the right thing for PMC members).

Committers should have the permission for both svn and Nexus usually. Could you check with ASF infra for this?

releasey/02-create-release-branch.sh
dimas-b: If there are CHANGELOG.md changes on `main` while the release is in progress, then automatically handling the release sub-sections in that file will likely lead to merge conflicts.

If we want to completely automate this process, I'd suggest to identify CHANGELOG entries with PR numbers and make custom tooling for generating release sub-sections. The tool can use the PR number IDs for reconciling `main` and `release` branches... not sure if there's appetite for this :sweat_smile: My original idea was to leverage the existing gradle plugin for that, but it apparently does not fit the multi-branch use case.

pingtimeout:
Closing this PR as it is replaced by #2383

PR #2162, opened 2025-07-23
Simplify checkPolarisServiceBootstrappedForRealm

XN137:
the `PolarisMetaStoreManager` lookup can just happen inside the method

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/LocalPolarisMetaStoreManagerFactory.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/LocalPolarisMetaStoreManagerFactory.java
@@ -149,8 +149,7 @@ public abstract class LocalPolarisMetaStoreManagerFactory<StoreType>
       RealmContext realmContext) {
     if (!metaStoreManagerMap.containsKey(realmContext.getRealmIdentifier())) {
       initializeForRealm(realmContext, null);
-      checkPolarisServiceBootstrappedForRealm(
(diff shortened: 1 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/LocalPolarisMetaStoreManagerFactory.java
```diff
@@ -160,12 +159,8 @@ public synchronized Supplier<TransactionalPersistence> getOrCreateSessionSupplie
       RealmContext realmContext) {
     if (!sessionSupplierMap.containsKey(realmContext.getRealmIdentifier())) {
       initializeForRealm(realmContext, null);
-      checkPolarisServiceBootstrappedForRealm(
-          realmContext, metaStoreManagerMap.get(realmContext.getRealmIdentifier()));
-    } else {
-      checkPolarisServiceBootstrappedForRealm(
-          realmContext, metaStoreManagerMap.get(realmContext.getRealmIdentifier()));
     }
+    checkPolarisServiceBootstrappedForRealm(realmContext);
```
XN137, line 163: note that since b069591c19e99743ace09fdf2cfda65654728473 this call was happening in both if branches, so we can move it to the outside.
but imo this fact raises the question why `getOrCreateMetaStoreManager` still only calls it on the inside.

(same problem exists in the copy-pasted jdbc variant)

snazy (approved):

dimas-b (approved):

PR #2164, opened 2025-07-23
Add PolarisObjectMapper

XN137:
all static methods that took a `PolarisCallContext` as first parameter can be moved to the new class.

if we did not need to log json-related exceptions to `PolarisDiagnostics` (which we dont do in other places) then we could keep everything as static util methods.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/TaskEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/TaskEntity.java
@@ -20,7 +20,6 @@ package org.apache.polaris.core.entity;
 
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.context.CallContext;
-import org.apache.polaris.core.persistence.PolarisObjectMapperUtil;
 
 /**
  * Represents an asynchronous task entity in the persistence layer. A task executor is responsible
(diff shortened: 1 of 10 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisObjectMapperUtil.java
```diff
@@ -187,8 +108,4 @@ public int getAttemptCount() {
       return null;
```
XN137, line 108: if you look above this line, json errors around task state dont get reported to `PolarisDiagnostics` but are simply logged

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisObjectMapperUtil.java
```diff
@@ -20,98 +20,19 @@
...
-import org.apache.iceberg.rest.RESTSerializers;
-import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
 import org.apache.polaris.core.entity.PolarisTaskConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** A mapper to serialize/deserialize polaris objects. */
-public class PolarisObjectMapperUtil {
+public final class PolarisObjectMapperUtil {
```
XN137, line 31: if we go forward with this PR we might want to rename this class as it now only deals with task (execution) state?

polaris-core/src/main/java/org/apache/polaris/core/entity/TaskEntity.java
```diff
@@ -41,16 +40,16 @@ public static TaskEntity of(PolarisBaseEntity polarisEntity) {
 
   public <T> T readData(Class<T> klass) {
     PolarisCallContext polarisCallContext = CallContext.getCurrentContext().getPolarisCallContext();
```
XN137, line 42: if `PolarisObjectMapperUtil` methods did not require the `PolarisDiagnostics` for reporting json-related errors, then this usage of `CallContext.getCurrentContext()` could go away (also below in this file).

XN137: found this gem a well:
https://github.com/apache/polaris/blob/987c554d1c45a38bad2d2bd6b1557dc619188885/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java#L219-L221

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisObjectMapperUtil.java
dimas-b: Either way works for me :)

dimas-b (commented):
This looks like a nice refactoring... Did you mean to remove the "draft" state and open this PR for reviews?

XN137:
> Did you mean to remove the "draft" state and open this PR for reviews?

not yet, i will probably make an alternative PR where i remove `PolarisDiagnostics` from all json-parsing related helpers and show the transitive cleanup this enables.

update: alternative PR (which is better imo) now available https://github.com/apache/polaris/pull/2176

XN137:
closing as the alternative PR got merged

PR #2170, opened 2025-07-24
Fix Namespace resolution on grant/revoke privilege operations

poojanilangekar:
I encountered the bug while implementing RBAC for federation. 
For grant/revoke operations on tables, polaris checks the subtype of the leaf entity and make sure it matches. However, for namespace the only check is that the resolved path is not null. 

Ideally we should check that the entire path matches the one sent in the request. Further, all entities except the root should be of the type NAMESPACE. 

Testing: 
Added unit tests.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapper.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapper.java
@@ -18,9 +18,12 @@
  */
 package org.apache.polaris.core.persistence;
 
+import jakarta.annotation.Nonnull;
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.iceberg.catalog.Namespace;
 import org.apache.polaris.core.entity.PolarisEntity;
+import org.apache.polaris.core.entity.PolarisEntityType;
 
 /**
  * Holds fully-resolved path of PolarisEntities representing the targetEntity with all its grants
@@ -75,6 +78,38 @@ public class PolarisResolvedPathWrapper {
     return getResolvedParentPath().stream()
         .map(ResolvedPolarisEntity::getEntity)
         .collect(Collectors.toList());
+  }
+
+  /**
+   * Checks if a namespace is fully resolved.
+   *
+   * @param catalogName the name of the catalog
+   * @param namespace the namespace we're trying to resolve
+   * @return true if the namespace is considered fully resolved for the given catalog type
+   */
+  public boolean isFullyResolvedNamespace(
+      @Nonnull String catalogName, @Nonnull Namespace namespace) {
+    if (resolvedPath == null) {
+      return false;
+    }
+
+    List<PolarisEntity> fullPath = getRawFullPath();
+    int expectedPathLength = 1 + namespace.levels().length;
+    if (fullPath.size() != expectedPathLength) {
+      return false;
+    }
+
+    if (!fullPath.get(0).getName().equals(catalogName)) {
+      return false;
+    }
+
+    for (int i = 0; i < namespace.levels().length; i++) {
+      if (!fullPath.get(i + 1).getName().equals(namespace.levels()[i])
+          || fullPath.get(i + 1).getType() != PolarisEntityType.NAMESPACE) {
+        return false;
+      }
+    }
+    return true;
   }
 
   @Override
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapper.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapper.java
@@ -19,6 +19,7 @@
 package org.apache.polaris.core.persistence;
 
 import jakarta.annotation.Nonnull;
+import java.util.Iterator;
 import java.util.List;
 import java.util.stream.Collectors;
 import org.apache.iceberg.catalog.Namespace;
@@ -103,9 +104,14 @@ public class PolarisResolvedPathWrapper {
       return false;
     }
 
-    for (int i = 0; i < namespace.levels().length; i++) {
-      if (!fullPath.get(i + 1).getName().equals(namespace.levels()[i])
-          || fullPath.get(i + 1).getType() != PolarisEntityType.NAMESPACE) {
+    String[] namespaceLevels = namespace.levels();
+    int levelsLength = namespaceLevels.length;
+    Iterator<PolarisEntity> fullPathIterator = fullPath.iterator();
+    fullPathIterator.next();
+    for (int i = 0; i < levelsLength; i++) {
+      PolarisEntity entity = fullPathIterator.next();
+      if (!entity.getName().equals(namespaceLevels[i])
+          || entity.getType() != PolarisEntityType.NAMESPACE) {
         return false;
       }
     }
--- a/polaris-core/src/test/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapperTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapperTest.java
@@ -1,0 +1,145 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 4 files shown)


poojanilangekar:
CC: @eric-maynard @dennishuo

service/common/src/test/java/org/apache/polaris/service/admin/PolarisAdminServiceTest.java
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  @Test
+  void testIsNamespaceFullyResolved_NullWrapper() throws Exception {
+    boolean result =
+        (boolean)
+            isNamespaceFullyResolvedMethod.invoke(
+                adminService, null, "test-catalog", Namespace.of("ns1"));
+
+    assertThat(result).isFalse();
```
dimas-b: nit: Why not inline `result`? The expression is already multi-line :thinking:

service/common/src/test/java/org/apache/polaris/service/admin/PolarisAdminServiceTest.java
```diff
@@ -0,0 +1,411 @@
...
+
+    PolarisEntity catalogEntity = createEntity(catalogName, PolarisEntityType.CATALOG);
+    List<PolarisEntity> shortPath = List.of(catalogEntity);
+
+    when(resolvedPathWrapper.getRawFullPath()).thenReturn(shortPath);
+
+    boolean result =
+        (boolean)
+            isNamespaceFullyResolvedMethod.invoke(
+                adminService, resolvedPathWrapper, catalogName, namespace);
```
dimas-b: This pattern is repeated... Maybe add a help method in this test class?

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1672,6 +1672,42 @@ public boolean revokePrivilegeOnCatalogFromRole(
...
+      @Nullable PolarisResolvedPathWrapper resolvedPathWrapper,
+      @Nonnull String catalogName,
+      @Nonnull Namespace namespace) {
+    if (resolvedPathWrapper == null) {
+      return false;
+    }
+
+    List<PolarisEntity> fullPath = resolvedPathWrapper.getRawFullPath();
+    int expectedPathLength = 1 + namespace.levels().length;
+    if (fullPath.size() < expectedPathLength) {
```
dimas-b: why not `==`?

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1672,6 +1672,42 @@ public boolean revokePrivilegeOnCatalogFromRole(
...
 
+  /**
+   * Checks if a namespace is fully resolved.
+   *
+   * @param resolvedPathWrapper the resolved path wrapper from resolution
+   * @param catalogName the name of the catalog
+   * @param namespace the namespace we're trying to resolve
+   * @return true if the namespace is considered fully resolved for the given catalog type
+   */
+  private boolean isNamespaceFullyResolved(
```
dimas-b: Why keep it as a private instance method? It does not use any state from `this`... It could be a package-private `static` method in a utility class.

dimas-b (commented):
Hi @poojanilangekar , this fix makes sense to me, but I've got some comments/suggestions :)

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: ... or perhaps a `public` method in `PolarisResolvedPathWrapper`? :thinking:

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
poojanilangekar: Done, changed it to !=

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java

service/common/src/test/java/org/apache/polaris/service/admin/PolarisAdminServiceTest.java
poojanilangekar: Done.

service/common/src/test/java/org/apache/polaris/service/admin/PolarisAdminServiceTest.java
poojanilangekar: I've moved the test, so that fixed it.

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapper.java
```diff
@@ -77,6 +80,38 @@ public List<PolarisEntity> getRawParentPath() {
...
+    if (fullPath.size() != expectedPathLength) {
+      return false;
+    }
+
+    if (!fullPath.get(0).getName().equals(catalogName)) {
+      return false;
+    }
+
+    for (int i = 0; i < namespace.levels().length; i++) {
+      if (!fullPath.get(i + 1).getName().equals(namespace.levels()[i])
```
dimas-b: This may not be a big deal, but since we're always scanning this list from head to tail, using iterators might be more portable and easier to understand. Probably as efficient as `.get(N)` for array-based lists, but potentially more efficient for linked lists.

It might be preferable to move `namespace.levels()` to a local variable. Iterate it in `for`, but do simple `iterator.next()` for `fullPath`.

polaris-core/src/test/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapperTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+public class PolarisResolvedPathWrapperTest {
+
+  @Test
+  void testIsFullyResolvedNamespace_NullResolvedPath() {
+    PolarisResolvedPathWrapper wrapper = new PolarisResolvedPathWrapper(null);
+
+    boolean result = wrapper.isFullyResolvedNamespace("test-catalog", Namespace.of("ns1"));
+
+    assertThat(result).isFalse();
```
dimas-b: nit: inline?

service/common/src/test/java/org/apache/polaris/service/admin/PolarisAdminServiceTest.java
```diff
@@ -0,0 +1,282 @@
...
+        .isInstanceOf(NotFoundException.class)
+        .hasMessageContaining("Namespace " + namespace + " not found");
+  }
+
+  @Test
+  void testGrantPrivilegeOnNamespaceToRole_IncompleteNamespaceThrowsNamespaceNotFoundException()
+      throws Exception {
+    String catalogName = "test-catalog";
+    String catalogRoleName = "test-role";
+    Namespace namespace = Namespace.of("incomplete-ns");
```
dimas-b: This is a single-level namespace... How is "incomplete" different from "non existent"?

polaris-core/src/test/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapperTest.java
```diff
@@ -0,0 +1,145 @@
...
+    assertTrue(wrapper.isFullyResolvedNamespace(catalogName, namespace));
+  }
+
+  @Test
+  void testIsFullyResolvedNamespace_WrongEntityType() {
+    String catalogName = "test-catalog";
+    Namespace namespace = Namespace.of("ns1");
+
+    PolarisEntity catalogEntity = createEntity(catalogName, PolarisEntityType.CATALOG);
+    PolarisEntity wrongTypeEntity = createEntity("ns1", PolarisEntityType.TABLE_LIKE);
```
dimas-b: For completeness this test should probably also have some NS level with the right type and some with another type.

poojanilangekar: Done.

service/common/src/test/java/org/apache/polaris/service/admin/PolarisAdminServiceTest.java
poojanilangekar: Done.

polaris-core/src/test/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapperTest.java
poojanilangekar: Done.

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisResolvedPathWrapper.java
poojanilangekar: Done.

dimas-b (approved):
Thanks, @poojanilangekar ! I'll be merging late Friday if there are no more review (or Mon if I forget :) )

poojanilangekar:
Hi @dimas-b, could you please merge this PR?

PR #2179, opened 2025-07-24
Update the Release Guide about the Helm Chart package

jbonofre:
The Release Guide is not up to date regarding what we did in the Polaris 1.0.0 release about Helm Chart.

This PR updates the Release Guide and "fix" the way to create Helm Chart package.


site/content/release-guide.md
```diff
@@ -113,9 +113,20 @@ git push apache release/x.y.z
...
+and update the version in the Helm Chart in:
+
+* `helm/polaris/Chart.yaml`
+* `helm/polaris/README.md`
+* `helm/polaris/values.yaml`
+
+and commit/push the version bump:
+
+```
+git commit -a -m "Bump version to x.y.z"
```
adutra:
```suggestion
git commit -a -m "Bump Helm Chart version to x.y.z"
```

site/content/release-guide.md
```diff
@@ -188,9 +199,46 @@ cd polaris-dist-dev
...
+shasum -a 512 polaris-x.y.z.tgz.prov > polaris-x.y.z.tgz.prov
+gpg --armor --output polaris-x.y.z.tgz.prov.asc --detach-sig polaris-x.y.z.tgz.prov
+```
+
+Copy and Add the Helm package files to dist folder:
+
+```
+cd polaris-dist-dev
+mkdir helm-chart/x.y.z
+cp helm/*.tgz*  helm-chart/x.y.z
```
adutra: You just `cd polaris-dist-dev` so not sure this `cp` command is correct.

site/content/release-guide.md
jbonofre: As this commit also "include" the `version.txt` update, I used `Bump version` (not only Helm Chart).

site/content/release-guide.md
jbonofre: Good point, let me improve that. Thanks !

site/content/release-guide.md
```diff
@@ -188,9 +199,46 @@ cd polaris-dist-dev
...
+
+```
+cd helm-chart
+helm repo index .
+```
+
+Dist repository is now "complete" and we can push/commit:
+
+```
 svn commit -m"Stage Apache Polaris x.y.z RCx"
```
snazy: Nitpick:
```suggestion
svn commit -m "Stage Apache Polaris x.y.z RCx"
```

site/content/release-guide.md
```diff
@@ -113,9 +113,20 @@ git push apache release/x.y.z
...
+
+* `helm/polaris/Chart.yaml`
+* `helm/polaris/README.md`
+* `helm/polaris/values.yaml`
+
+and commit/push the version bump:
+
+```
+git commit -a -m "Bump version to x.y.z"
 git push
```
snazy: Maybe defer the `git push` to after (or before?) the `svn commit` - just in case something breaks.

pingtimeout: This particular `git push` should be removed as there is already one L137.

Also, @jbonofre would it be possible to replace the `git commit -a` command L136 by 

```
git add CHANGELOG.md
git commit -m "Patch changelog for release x.y.z"
```

This is to be 100% sure that if somebody performs a release manually but the working directory is not 100% clean, we don't accidentally add files to the repository.

site/content/release-guide.md
```diff
@@ -188,9 +199,46 @@ cd polaris-dist-dev
...
 svn add x.y.z
+```
+
+### Stage Helm Chart package
+
+You can now create a Helm package with the following command in the Polaris source folder:
+
+```
+cd helm
+helm package --sign --key 'My Key' --keyring ~/.gnupg/secring.kbx polaris
```
pingtimeout: What should `My Key` be, exactly?  Should it be the key ID?  If that's the case, for consistency with the rest of the release guide, the command should be:

```suggestion
helm package --sign --key '<YOUR KEY ID HERE>' --keyring ~/.gnupg/secring.kbx polaris
```

jbonofre: It's the release manager key. Ok for use brackets.

site/content/release-guide.md
jbonofre: Yeah sure. It makes sense.

jbonofre: It's more than the changelog though: versions bump as well. But let me list the files to update in the commit.

site/content/release-guide.md
```diff
@@ -113,16 +113,26 @@ git push apache release/x.y.z
...
+
+```
+git commit -m "Bump version to x.y.z" version.txt helm/polaris/Chart.yaml helm/polaris/README.md helm/polaris/values.yaml
 ```
 
 Update `CHANGELOG.md`:
 ```
 ./gradlew patchChangelog
-git commit -a
+git commit CHANGELOG.md
```
pingtimeout: Should this command include a `-m "Update CHANGELOG for x.y.z release"` parameter just like the other `git commit` command above?

jbonofre: Good catch. Let me do it.

site/content/release-guide.md
```diff
@@ -188,18 +198,55 @@ cd polaris-dist-dev
...
 svn add x.y.z
+```
+
+### Stage Helm Chart package
+
+You can now create a Helm package with the following command in the Polaris source folder:
+
+```
+cd helm
+helm package --sign --key '<YOUR KEY ID HERE>' --keyring ~/.gnupg/secring.kbx polaris
```
flyrain: the key id didn't work for me, I have to use the name like this: 
```
helm package helm/polaris \
  --destination build/helm \
  --sign \
  --key 'Yufei Gu' \
  --keyring ~/.gnupg/helm-secring.gpg
```

flyrain: I think it's also worth to mention how key ring is generated

site/content/release-guide.md
```diff
@@ -188,18 +198,55 @@ cd polaris-dist-dev
...
+cp helm/*.tgz*  helm-chart/x.y.z
 svn add helm-chart/x.y.z
-svn commit -m"Stage Apache Polaris x.y.z RCx"
+```
+
+You can now update the Helm index:
+
+```
+cd helm-chart
+helm repo index .
```
flyrain, line 240: Indexing is also required after the we move it to the release repo, as the links are different between the staging and release.

site/content/release-guide.md
jbonofre: Keyring is part of GPG (so classic GPG command), but I can add commands to generate it (it should be done already for GPG key for release manager).

site/content/release-guide.md
jbonofre: Actually, I think we should remove index on dist dev and only do that on dist release. I will update the release guide in that way (for discussion).

site/content/release-guide.md
flyrain: A normal key generation doesn't produce key ring. Can you also double check the key-name-vs-key-id thing? The key id never works for me. This is a quite weird thing that Helm does, I hope they fixed it.

site/content/release-guide.md
flyrain: I think it's still valuable to index the staging. Otherwise, the Helm Chart validation may fail. For example, the following commands won't work
```
helm repo add polaris-dev https://dist.apache.org/repos/dist/dev/incubator/polaris/helm-chart/
helm repo update
helm install mypolaris polaris-dev/polaris --version 1.0.0-incubating
```

site/content/release-guide.md
pingtimeout: @flyrain The first GPG command that interacts with an empty `~/.gnupg` will create a keyring.  So if it does not exist, it means that you have a specific setup (see below for an example).

Personally, I do not have a `secring.gpg` file anywhere in `~/.gnupg` because my secret keys are stored on a Yubikey.  So the command does not work for me either.  I would not be able to run the `helm package ... --keyring ~/.gnupg/helm-secring.gpg` command as-is.  Is it possible that you also have a specific setup where you private keys are not stored on your computer?

FWIW I keep running into https://github.com/helm/helm/issues/2843 because of an incompatibility between Helm and the last GPG versions.  So for me the process would look like this, relying on [the helm-gpg plugin](https://github.com/technosophos/helm-gpg):

```
$ helm package --destination build/helm polaris
$ helm gpg sign build/helm/polaris-1.0.1-incubating.tgz
```

flyrain: The plugin does seem a nice workaround, so that both keyring and key-name-vs-key-id issue can be avoided. The problem I keep hitting is that `--key '<YOUR KEY ID HERE>'` doesn't work, reported here: https://github.com/helm/helm/issues/7631.

site/content/release-guide.md
```diff
@@ -188,18 +198,55 @@ cd polaris-dist-dev
...
+cd helm
+helm package --sign --key '<YOUR KEY ID HERE>' --keyring ~/.gnupg/secring.kbx polaris
+```
+
+Create the signature and checksum for the Helm package tgz and prov files:
+
+```
+shasum -a 512 polaris-x.y.z.tgz > polaris-x.y.z.tgz.sha512
+gpg --armor --output polaris-x.y.z.tgz.asc --detach-sig polaris-x.y.z.tgz
+shasum -a 512 polaris-x.y.z.tgz.prov > polaris-x.y.z.tgz.prov
```
pingtimeout:
```suggestion
shasum -a 512 polaris-x.y.z.tgz.prov > polaris-x.y.z.tgz.prov.sha512
```

site/content/release-guide.md
pingtimeout: How far do we want to go on the gpg-setup side of things?  Like, should we assume (from the prerequisites) that a maintainer has properly configured and published their GPG keys?  Or should we include a howto in this guide?

Personally, I would keep this guide as focused as possible on the release and assume it is the responsibility of the maintainer to have a properly configured key.  Which means that we could replace this command by:

```suggestion
helm package polaris
helm gpg sign polaris-x.y.z.tgz
```

wdyt?

jbonofre: That looks like a good proposal indeed. I propose to commit Pierre's suggestion. Thoughts ?

site/content/release-guide.md
jbonofre: Good point. The index in staging is only useful for testing RC (as it's not the one used after release). Let me update here.

site/content/release-guide.md
flyrain: The GPG key generation guide (https://infra.apache.org/release-signing.html#generate) is pretty good, we may just ensure people to follow the instructions strictly in this guide. There is no need to add extra information if they are covered already.

site/content/release-guide.md
pingtimeout: @flyrain I think I implemented the necessary changes, could you double check?

flyrain (dismissed):

adutra (approved):

dimas-b (approved):

PR #2180, opened 2025-07-24
Async & reliable tasks API, SPI, Store interfaces

snazy:
Just the interfaces and value types needed for "Async & reliable tasks", the first three modules as described in `tasks/README.md`. The 3 different modules serve different "audiences":
* API is for call sites that submit tasks via the `Tasks` interface.
* SPI is for task/behavior implementations
* Store is for persistence abstraction, only for tasks implementations

The only entry point is the `org.apache.polaris.tasks.api.Tasks` interface with a function defining the behavior and providing a parameter object (if necessary), returning a `TaskSubmission`. Call sites _may_ subscribe to a `CompletionStage`, but the idea is that it's rather "fire and forget" and the task behavior does "everything that's needed". This allows the task to be executed on any node. There's no guarantee in any form that a task will run "locally" or any other specific node. Every Polaris node can handle task execution and perform failure/retry handling. Polaris nodes may use a "server" implementation or a "client" implementation or a "remote" implementation - that's defined upon deployment or by configuration (TBD).

Implementations would be abstracted similarly:
* Store implementation with NoSQL specifics, especially for the indexed/sorted lookups
* Store implementation with JDBC specifics (can leverabe `BasePersistence`)
* Tasks implementation for servers, clients and remote


tasks/README.md
```diff
@@ -0,0 +1,155 @@
...
+### Service instance
+
+Provides a `Tasks` implementation, which can submit tasks and execute tasks.
+
+### Client instance
+
+Provides a `Tasks` implementation, which can submit tasks but not execute tasks. This kind is intended for client
+application use cases, which should not be "bothered" task executions from other (Polaris service) instances. For
+example, clients should not "purge table" tasks from all users, despite that clients cannot be expected to have access
+to all service related resources.
```
pingtimeout: Should this paragraph be reworded?  My English might not be the best one, I don't understand the "_client
application use cases, which should not be "bothered" task executions from other (Polaris service) instances_" sentence.

Does it mean that client applications should not _borrow_ tasks?  Or something else?

And the sentence "_clients should not "purge table" tasks from all users, despite that clients cannot be expected to have access to all service related resources._" sounds weird with the should not/despite cannot formulation.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskBehaviorId.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
+
+/** Task behavior ID, as a type safe holder. */
+@PolarisImmutable
+@JsonSerialize(as = ImmutableTaskBehaviorId.class)
+@JsonDeserialize(as = ImmutableTaskBehaviorId.class)
+public interface TaskBehaviorId {
```
pingtimeout, line 33: I suspect that the use of an `interface` here instead of a `record` is to allow extensibility, as records cannot be extended from?

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskParameterTypeId.java
```diff
@@ -0,0 +1,39 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import jakarta.annotation.Nonnull;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Provide the {@linkplain TaskBehaviorId task behavior ID} using the
```
pingtimeout:
```suggestion
 * Provides the {@linkplain TaskBehaviorId task behavior ID} using the
```

Is this sentence correct?  This javadoc is identical to that of `TaskResultTypeId`, but neither class uses nor are used by the `TaskBehaviorId` class at all.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskResultTypeId.java
```diff
@@ -0,0 +1,39 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import jakarta.annotation.Nonnull;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Provide the {@linkplain TaskBehaviorId task behavior ID} using the
```
pingtimeout:
```suggestion
 * Provides the {@linkplain TaskBehaviorId task behavior ID} using the
```

Same question as before: Is this sentence correct?  This javadoc is identical to that of `TaskParameterTypeId`, but neither class touches the `TaskBehaviorId` class at all.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskSerializationResolver.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Optional;
+
+/**
+ * Interface needed by {@link TaskParameterTypeIdResolver} and {@link TaskParameterTypeIdResolver}
+ * to resolve type-IDs to Java types. This is used in/with CDI.
+ */
+public interface TaskSerializationResolver {
+
+  /**
+   * Resolve the task parameter type using the ID defined in {@link TaskParameterTypeId#value()}.
```
pingtimeout:
```suggestion
   * Resolves the task parameter type using the ID defined in {@link TaskParameterTypeId#value()}.
   *
   * @param parameterTypeId the task parameter type ID
   * @return the task parameter class, or an empty optional if not found
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskSerializationResolver.java
```diff
@@ -0,0 +1,37 @@
...
+   */
+  Optional<Class<? extends TaskParameter>> taskParameterType(@Nonnull String parameterTypeId);
+
+  /** Resolve the task result type using the ID defined in {@link TaskResultTypeId#value()}. */
```
pingtimeout:
```suggestion
  /**
   * Resolves the task result type using the ID defined in {@link TaskResultTypeId#value()}.
   *
   * @param resultTypeId the task result type ID
   * @return the task result class, or an empty optional if not found
   */
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskSubmission.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Use this completion stage to subscribe to the result of a task execution, do not use this when
+   * the task execution is about to happen "far in the future" or is a recurring task.
+   *
+   * <p>It only makes sense to subscribe to a task execution of a non-repeating task. The behavior
+   * of the returned completion stage is undefined for re-scheduled tasks. Implementations may
+   * return the result of the "current"/next task run or not yield a result as long as the task gets
+   * rescheduled.
+   */
+  CompletionStage<RESULT> completionStage();
```
pingtimeout, line 43: Suggestion (multi-line so I cannot offer that directly in Github):

```
  /**
   * Returns a completion stage to subscribe to the result of a task execution.
   *
   * <p>This should only be used for tasks that will execute in the near future. Do not use this for
   * tasks scheduled far in the future or for recurring tasks.
   *
   * <p>It only makes sense to subscribe to the execution of a non-repeating task. The behavior of
   * the returned completion stage is undefined for rescheduled tasks. Implementations may return
   * the result of the current or next task run, or may not yield a result as long as the task gets
   * rescheduled.
   */
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/Tasks.java
```diff
@@ -0,0 +1,66 @@
...
+package org.apache.polaris.tasks.api;
+
+/**
+ * User-facing task submission API.
+ *
+ * <p>This interface intentionally only provides a function to submit tasks.
+ */
+public interface Tasks {
+  /**
+   * Submit the given task request for execution.
```
pingtimeout:
```suggestion
   * Submits the given task request for execution.
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/Tasks.java
```diff
@@ -0,0 +1,66 @@
...
+   *
+   * <p>This API only guarantees that, upon successful execution, the task request will eventually
+   * be executed. When exactly and where exactly, think: system process, the task will be executed
+   * depends on configuration specifics and task state.
```
pingtimeout: Multi-line suggestion for clarity:

```
   * <p>This API only guarantees that, upon successful return from this method, the task request
   * will eventually be executed. The exact timing and location (i.e., which system process) of the
   * task execution depends on configuration specifics and task state.
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/Tasks.java
```diff
@@ -0,0 +1,66 @@
...
+   * @param behaviorId ID of task behavior being submitted. An {@link IllegalArgumentException} is
+   *     thrown if no behavior for this ID is registered.
+   * @param realmId realm ID for which the task is submitted
+   * @param param task parameter value, task behavior implementations derive the actual task ID
+   *     based on their implementation and potentially from this task parameter value. The caller is
+   *     responsible for providing a correct, compatible parameter value for the given task
+   *     behavior.
+   * @param resultType type of the result value object. The caller is responsible for providing the
+   *     correct, compatible result type for the given task behavior.
+   * @return asynchronous execution result
```
pingtimeout: Multi-line suggestion (distinguish the `<PARAM>` and `<RESULT>` as types.

```
   * @param <PARAM> task input parameter type
   * @param <RESULT> task execution result type
   * @param behaviorId ID of task behavior being submitted. An {@link IllegalArgumentException} is
   *     thrown if no behavior for this ID is registered.
   * @param realmId realm ID for which the task is submitted
   * @param param task parameter value. Task behavior implementations derive the actual task ID
   *     based on their implementation and potentially from this task parameter value. The caller is
   *     responsible for providing a correct, compatible parameter value for the given task
   *     behavior.
   * @param resultType type of the result value object. The caller is responsible for providing the
   *     correct, compatible result type for the given task behavior.
   * @return the task submission handle containing the task ID
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+@JsonSerialize(as = ImmutableTasksConfiguration.class)
+@JsonDeserialize(as = ImmutableTasksConfiguration.class)
+@JsonTypeName(TasksConfiguration.TYPE_ID)
+public interface TasksConfiguration {
+  String TYPE_ID = "coordinated-tasks";
+
+  int DEFAULT_RETAINED_HISTORY_LIMIT = 5;
+
+  /** Number of history entries for task scheduling/submission updates. */
+  @WithDefault("" + DEFAULT_RETAINED_HISTORY_LIMIT)
```
pingtimeout, line 45: What is the rationale for `DEFAULT_RETAINED_HISTORY_LIMIT` being an `int` instead of a `String`, and then converted to a string using a `"" + ` concatenation?

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+  @WithDefault(DEFAULT_UPDATE_STATE_INTERVAL_STRING)
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
+  @JsonFormat(shape = JsonFormat.Shape.STRING)
+  Optional<Duration> updateStateInterval();
+
+  String DEFAULT_LOST_NOT_BEFORE_STRING = "PT5M";
+  Duration DEFAULT_LOST_NOT_BEFORE = Duration.parse(DEFAULT_LOST_NOT_BEFORE_STRING);
+
+  /**
+   * Tasks in status "running" are considered as "lost" when the last state-update was before the
```
pingtimeout:
```suggestion
   * Tasks in "running" status are considered "lost" when the last state-update was before the
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+   * duration specified by this parameter.
+   *
+   * <p>Running tasks regularly update the task store with their current state (see {@code
+   * updateStateInterval}). Tasks states with a {@code lostNotBefore} in the past are considered as
+   * "dead", which means that the task's execution can be resumed by another node.
```
pingtimeout: Suggestion: `Tasks states` -> `Tasks state`

Also, AFAIU we do not have the ability to **resume** a task execution (think: from a checkpoint, à-la MapReduce).  Instead, we have the ability to **retry** a task execution, with the at-least-once semantics.  Should the `resumed` word be replaced?

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+  @WithDefault(DEFAULT_LOST_NOT_BEFORE_STRING)
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
+  @JsonFormat(shape = JsonFormat.Shape.STRING)
+  Optional<Duration> lostNotBeforeDuration();
+
+  String DEFAULT_RUNNABLE_TASKS_POLL_INTERVAL_STRING = "PT5S";
+  Duration DEFAULT_RUNNABLE_TASKS_POLL_INTERVAL =
+      Duration.parse(DEFAULT_RUNNABLE_TASKS_POLL_INTERVAL_STRING);
+
+  /** Interval at which nodes check for scheduled tasks that eligible to run. */
```
pingtimeout, line 84:
```suggestion
  /** Interval at which nodes check for scheduled tasks that are eligible to run. */
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
+  @JsonFormat(shape = JsonFormat.Shape.STRING)
+  Optional<Duration> runnableTasksPollInterval();
+
+  String DEFAULT_REMOTE_TASK_POLL_INTERVAL_STRING = "PT0.5S";
+  Duration DEFAULT_REMOTE_TASK_POLL_INTERVAL =
+      Duration.parse(DEFAULT_REMOTE_TASK_POLL_INTERVAL_STRING);
+
+  /**
+   * The interval at which the state of a task that is running on another node is being polled.
```
pingtimeout:
```suggestion
   * The interval at which the state of a task running on another node is polled.
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * <p>Call sites that subscribe to the result of a task, which runs on a different node, use this
```
pingtimeout:
```suggestion
   * <p>Call sites that subscribe to the result of a task running on a different node use this
```

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
+  @JsonFormat(shape = JsonFormat.Shape.STRING)
+  Optional<Duration> remoteTaskPollInterval();
+
+  String DEFAULT_LOCAL_SCHEDULING_INTERVAL_STRING = "PT2S";
+  Duration DEFAULT_LOCAL_SCHEDULING_INTERVAL =
+      Duration.parse(DEFAULT_LOCAL_SCHEDULING_INTERVAL_STRING);
+
+  /**
+   * New submitted that are scheduled with this interval will be scheduled for non-local execution,
```
pingtimeout:
```suggestion
   * Newly submitted tasks that are scheduled with this interval will be scheduled for non-local execution,
```

I am a bit puzzled by this parameter.  I could not find any description of a guarantee for non-local execution in the readme.  And given that the implementations will be submitted in subsequent PRs, it is difficult for me to grasp the concept described here.  Should either the javadoc or the readme be updated to provide more information?

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /** Globally unique ID of the task behavior. */
+  TaskBehaviorId id();
+
+  Class<PARAM> paramType();
+
+  Class<RESULT> resultType();
+
+  /**
+   * Provide a task-runnable that can perform the task behavior's operation.
```
pingtimeout:
```suggestion
   * Provides a task function that can perform the task behavior's operation.
```

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * <p>Implementations must <em>never</em> assume that any values or context information from a
+   * "scheduling" context (think: CDI request context, even propagated) is available. This is
+   * neither supported by the CDI specification nor practically doable, especially considering that
+   * task functions are executed "far" in the future and/or on a different node.
+   */
+  TaskFunction<PARAM, RESULT> function();
+
+  /**
+   * Generate a task ID.
```
pingtimeout, line 64:
```suggestion
   * Generates a task ID.
```

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * <p>Task behavior implementations produce either new, globally unique task IDs per submission or
+   * deterministic task IDs based on the task parameters.
+   *
+   * <p>Implementations must include the given {@code realmId} when generating or calculating a task
+   * ID.
```
pingtimeout, line 70: Is this true 100% of the time?  Or could there be some cross-realm tasks (like Polaris overall cleanup tasks) that would be realm-independent?  In which case, `must` should be replaced by `should`

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Used to generate a deterministic task ID, from the contents of a {@link TaskParameter}.
+   *
+   * <p>Deterministic task IDs are useful to converge requests for the same task to a single task
+   * execution. {@link Tasks} implementations ensure that task submissions against the same task ID
+   * are executed only once and all submissions share the same {@linkplain TaskResult task result}.
+   *
+   * <p>Implementations must include the given {@code realmId} in the task ID generation
+   * respectively the calculation.
```
pingtimeout, line 44: The `respectively the calculation` part sounds odd, is this intended?

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  // TODO this makes Guava a part of this API - acceptable in this case?
+  @SuppressWarnings("UnstableApiUsage")
+  TaskId generateDeterministicId(String realmId, Consumer<PrimitiveSink> funnel);
```
pingtimeout, line 48: That part confuses me the most.  The Javadoc says that this method consumes the content of a `TaskParameter` but the method itself does not accept that kind of object.  Shouldn't it accept zero or more `TaskParameter` instead of a consumer of a guava class that itself consumes primitive types?

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskHandle.java
```diff
@@ -0,0 +1,26 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.tasks.store;
+
+/**
+ * Represents the store implementation-specific and opaque handle for one specific task execution.
+ *
+ * <p>Implementation note: must implement {@link Object#equals(Object)}!
```
pingtimeout:
```suggestion
 * <p>Implementation note: implementations must override {@link Object#equals(Object)} and {@link
 * Object#hashCode()}.
```

I assume we want to follow good Java practices here :-)

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskParameter.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+/**
+ * Base interface for task parameter objects. This is the necessary base type for Jackson type
+ * polymorphism.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
+public interface TaskParameter {
```
pingtimeout, line 32: When I first reviewed this PR, I had the impression that this class was representing a single parameter of a given task.  And that, if a task had multiple parameters, it would have multiple `TaskParameter` objects.

Now, I am thinking that each task behaviour can have only one `TaskParameter` and that, despite the singular name of the `TaskParameter` class, it holds *all* the parameters of a given task.  I.e. it is the JSON root of an object containing all parameters.  If so, would it make sense updating the documentation to reflect that?

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskState.java
```diff
@@ -0,0 +1,146 @@
...
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
+  Optional<TaskResult> taskResult();
+
+  static ImmutableTaskState.Builder builder() {
+    return ImmutableTaskState.builder();
+  }
+
+  @SuppressWarnings("UnnecessaryDefault")
+  @JsonIgnore
+  default boolean isScheduled(Instant now) {
```
pingtimeout: It seems to me that this method returns `true` if the current task is scheduled **in the future**.  However it it is `SCHEDULED` but the current date is after the scheduled date, then this method returns `false`.  Is this intended?  Maybe add a Javadoc to explain how this method is expected to be used?

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskStore.java
```diff
@@ -0,0 +1,121 @@
...
+   *
+   * @param now maximum schedule-not-before timestamp. The implementation filters out entries that
+   *     have no or a higher schedule-not-before timestamp before calling the {@code filter}, which
+   *     allows pushing down this value as a predicate to the persistence layer.
+   * @param filter filter to be matched, only entries for which this predicate yields {@code true}
+   *     are returned.
+   * @param limit maximum number of entries to return, empty means unlimited. Only eligible tasks
+   *     that match the maximum schedule-not-before timestamp and pass the filter test are counted
+   *     towards this value.
+   * @return matching tasks, the order of elements within the returned list is undefined
```
pingtimeout, line 57: nit: given that `List` allows index-based access, and the order of elements is undefined, it might make sense to return a `Collection` instead of a `List`.

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskStore.java
```diff
@@ -0,0 +1,121 @@
...
+  Optional<TaskState> fetchTaskState(@Nonnull TaskHandle taskHandle);
+
+  /**
+   * Update the task state for the given ID.
+   *
+   * <p>The {@code updater} function might be called multiple times, for example, when
+   * commit-retries happen. Therefore {@code updater} function must be free of side effects.
+   *
+   * @param updater Update function must be idempotent and expect to be invoked multiple times. Gets
+   *     the current state as its input may throw
```
pingtimeout:
```
   * @param updater the update function, which must be idempotent and expect to be invoked multiple
   *     times. It receives the current state as input and returns the desired change.  It may
   *     throw an exception in the event of a failure.
```

I assume the semantics here are as follows:
* If the updater function returns a `TaskStoreResult`, this new state should be persisted to the task store
* If it returns an empty optional, then the task should be deleted from the task store
* If it throws, then an error occurred and an updated state should be recomputed

Is this correct?  Those semantics should probably be described in the Javadoc.

pingtimeout (commented):
Thanks @snazy for the thorough work !  The PR looks good to me.  I have made some suggestions for Javadoc clarity.  I consider them all as `nit`.  Feel free to disregard any that is not valuable enough.

tasks/README.md
snazy: Yea, the sentence is a weird. I actually removed it entirely.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskBehaviorId.java
snazy: No specific reason, but all other types are interfaces.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskParameterTypeId.java
snazy: True, this part is currently not wired anywhere.
It will be used to _dynamically_ find the right implementations as implementations are pluggable. It is later used in the async & reliable tasks implementation to find the right Jackson serializer/deserializer (Jackson polymorphism).

tasks/api/src/main/java/org/apache/polaris/tasks/api/Tasks.java
snazy: Tip: select multiple lines, then you can add multiple-line suggestions.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
snazy: That way implementations can directly refer to the `int` value instead of a string value, while smallrye-config expects a string value.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
snazy: replaced "resumed" w/ "picked up"

tasks/api/src/main/java/org/apache/polaris/tasks/api/TasksConfiguration.java
snazy: Good catch. This is rather an optimization to avoid the (more expensive) task-submission/enqueuing + acquisition for tasks that are to be executed pretty much immediately. Updated the whole paragraph.

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskBehavior.java
snazy: "Must" is actually correct here (it must not be null). Implementations defer to any of the 'TaskIdGenerator` functions. `generateNonDeterministicUniqueId` delegates ID generation to the persistence being used.

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
snazy: Well, the default implementation uses a SHA, which is rather a calculation.

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
snazy: Call sites (behavior implementations) use the attributes of the concrete `TaskParameter` **sub**-types that are used to generated a deterministic ID.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskParameter.java
snazy: Added.

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskState.java
snazy: Technically, this function is used to determine whether a task _can_ be scheduled. Updated this a bit.

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskStore.java
snazy: Mostly.
A `TaskStateUpdater` function returns a `TaskChange` "instruction" (no change, create-or-update or update-and-unschedule).
There is no "delete" passed back.
It's the `TaskStore`'s responsibility to perform the actual change and yield the outcome.

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskParameterTypeId.java
pingtimeout: Ok so then would it make sense for the Javadoc to be "Provides the **Task Parameter ID** using the Jackson-polymorphism-mechanism" instead of  "Provides the **task behavior ID** using the Jackson-polymorphism-mechanism"?

I am not sure how it will be used.  I suppose this and the `TaskResultTypeId` annotation could be used to find a task behaviour which uses matching parameter id and result type id.  In which case the Javadoc should still be updated as the annotation does not directly provide the task behaviour id.  Am I missing something?

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
pingtimeout: Oh the calculation makes total sense, but that's not what I am questioning.  Rather the grammatically akward sentence: "_Implementations must include the given {@code realmId} in the task ID generation respectively the calculation._"

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
pingtimeout: Then shouldn't this method be like that?

```
  <E extends TaskParameter> TaskId generateDeterministicId(String realmId, Consumer<E> funnel);
```

tasks/store/src/main/java/org/apache/polaris/tasks/store/TaskState.java
pingtimeout: My point is that a task status can be SCHEDULED and the method name isSchedule() suggests that it is a shortcut for status() == SCHEDULED. But the method does something else. Maybe it could be good naming it canBeScheduled() given your answer?

snazy: Seen the commit from Dec 20?

tasks/api/src/main/java/org/apache/polaris/tasks/api/TaskParameterTypeId.java
snazy: I think it will become clearer with the implementation. In case of any doubts we can certainly change it later.

tasks/spi/src/main/java/org/apache/polaris/tasks/spi/TaskIdGenerator.java
snazy: No, `PrimitiveSink` is a Guava type

PR #2190, opened 2025-07-28
SigV4 Auth Support for Catalog Federation - Part 2: Connection Config Persistence

XJDKC:
## Milestones

This is Part 2 of the [[Splitting] Initial SigV4 Auth Support for Catalog Federation](https://github.com/apache/polaris/pull/1805). Upcoming parts will build on this system:
* ~~#1899~~
* #2190
* SigV4 Auth Support for Catalog Federation - Part 3: Service Identity Info Injection

## Introduction

This PR introduces DPOs (data persistence objects) that allow Polaris to persist SigV4 authentication parameters and service identity references associated with remote catalog connections.

The core idea is to persist a reference to Polaris's own service credentials (e.g. AWS IAM user) rather than the credentials themselves. This enables secure, pluggable credential resolution from external secret stores (e.g. a vault or secret manager) and supports the ability to assume user-specified roles at runtime via SigV4.

## Design Overview

Each `ConnectionConfigInfoDpo` (used for remote catalog federation) now contains a ServiceIdentityInfoDpo, which in turn holds a ServiceSecretReference. This design allows:
* Polaris to store only references to its service identity (e.g. AWS IAM user)
* The actual credentials to be stored securely in a vault or secret manager
* Runtime resolution of credentials based on these references
* Role assumption using `SigV4AuthenticationParametersDpo` (supplied by the user)

This separation of identity metadata and authentication parameters provides a secure and flexible foundation for credential management.

## Key Components
* `SigV4AuthenticationParametersDpo`: Holds user-supplied role assumption parameters like:
    * `roleArn`
    * `roleSessionName`: optional
    * `externalId`: optional
* `ServiceIdentityInfoDpo`: Stores metadata about the Polaris-side service identity, including a secret reference.
* `AwsIamServiceIdentityInfoDpo`:
*  `ServiceSecretReference`: Specialized version of ServiceIdentityInfoDpo for AWS IAM. It includes:
    * `iamArn`: Polaris's AWS user or role
    * `ServiceSecretReference`: (points to credentials in a vault)
* `ServiceSecretReference`: Points to a credential (e.g., in a vault). This is a logical URN or identifier — Polaris never persists raw secrets.

## Flowchart
Catalog Federation - Creds Management


polaris-core/src/main/java/org/apache/polaris/core/secrets/ServiceSecretReference.java
XJDKC, line 1: I extended `UserSecretReference` and introduced a new class `ServiceSecretReference`, to better reflect the purpose of the reference. I'm open to renaming it.

The URN format check remains the same (`urn:polaris-secret:xxx:xxx`). To make this more flexible and reusable, we could introduce a base class called `SecretReference`. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/secrets/ServiceSecretReference.java
```diff
@@ -0,0 +1,53 @@
...
+import jakarta.annotation.Nullable;
+import java.util.Map;
+
+/**
+ * Represents a "wrapped reference" to a service-owned secret that holds an identifier to retrieve
+ * possibly remotely-stored secret material, along with an open-ended "referencePayload" that is
+ * specific to an implementation of the secret storage and which is needed "unwrap" the actual
+ * secret in combination with whatever is stored in the remote secrets storage.
+ */
+public class ServiceSecretReference extends UserSecretReference {
```
XJDKC: I extended `UserSecretReference` and introduced a new class `ServiceSecretReference`, to better reflect the purpose of the reference. I'm open to renaming it.

The URN format check remains the same (`urn:polaris-secret:xxx:xxx`). To make this more flexible and reusable, we could introduce a base class called `SecretReference`. WDYT?

dennishuo: If the members will be exactly the same as `UserSecretReference` and this only differs by type, we should probably just rename `UserSecretReference` to `SecretReference` without any subclasses (or only extend in the future if we really need extended or modified functionality for one of the types).

XJDKC: Agree!

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionConfigInfoDpo.java
```diff
@@ -64,22 +66,29 @@ public abstract class ConnectionConfigInfoDpo implements IcebergCatalogPropertie
...
+      @JsonProperty(value = "serviceIdentity", required = false) @Nonnull
+          ServiceIdentityInfoDpo serviceIdentity) {
+    this(connectionTypeCode, uri, authenticationParameters, serviceIdentity, true);
   }
 
   protected ConnectionConfigInfoDpo(
       int connectionTypeCode,
       @Nonnull String uri,
       @Nonnull AuthenticationParametersDpo authenticationParameters,
+      @Nonnull ServiceIdentityInfoDpo serviceIdentity,
```
dennishuo: Is this actually `@Nullable`? Here and above

polaris-core/src/main/java/org/apache/polaris/core/secrets/SecretReference.java
```diff
@@ -51,7 +51,7 @@
  * the stored "secret material" as well as the referencePayload and any associated keys used for
  * encryption.
  */
-public class UserSecretReference {
+public class SecretReference {
```
dimas-b, line 54: nit: Is this rename critical to adding SigV4 support? It would have been nicer to do it in a separate PR without functional changes (to reduce distraction) :slightly_smiling_face:

polaris-core/src/main/java/org/apache/polaris/core/identity/ServiceIdentityType.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * The internal persistence-object counterpart to ServiceIdentityTypeInfo.ServiceIdentityTypeEnum
+ * defined in the API model. We define integer type codes in this enum for better compatibility
+ * within persisted data in case the names of enum types are ever changed in place.
+ *
+ * <p>Important: Codes must be kept in-sync with JsonSubTypes annotated within {@link
+ * ServiceIdentityInfoDpo}.
+ */
+public enum ServiceIdentityType {
+  NULL_TYPE(0),
```
dimas-b, line 34: What is the purpose of `NULL_TYPE`?

polaris-core/src/main/java/org/apache/polaris/core/identity/ServiceIdentityType.java
```diff
@@ -0,0 +1,83 @@
...
+   * Given the code associated with the type, return the associated ServiceIdentityType. Return
+   * NULL_TYPE if not found
+   *
+   * @param identityTypeCode code associated to the service identity type
+   * @return ServiceIdentityType corresponding to that code or null if mapping not found
+   */
+  public static @Nonnull ServiceIdentityType fromCode(int identityTypeCode) {
+    // ensure it is within bounds
+    if (identityTypeCode < 0 || identityTypeCode >= REVERSE_MAPPING_ARRAY.length) {
+      return NULL_TYPE;
```
dimas-b, line 73: What is the purpose of `NULL_TYPE`? It has a specific type code (`0`) but we map to it from other specific type codes... how is that meant to work end-to-end? :thinking:

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -338,6 +338,14 @@ public Builder setConnectionConfigInfoDpoWithSecrets(
       return this;
     }
 
+    public Builder setConnectionConfigInfoDpo(
```
dimas-b, line 342: This method appears to be unused :thinking:

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionConfigInfoDpo.java
```diff
@@ -173,6 +194,15 @@ public static ConnectionConfigInfoDpo fromConnectionConfigInfoModelWithSecrets(
     return config;
   }
 
+  /**
+   * Creates a new copy of the ConnectionConfigInfoDpo with the given service identity info.
+   *
+   * @param serviceIdentityInfo The service identity info to set.
+   * @return A new copy of the ConnectionConfigInfoDpo with the given service identity info.
+   */
+  public abstract ConnectionConfigInfoDpo withServiceIdentity(
```
dimas-b, line 203: This method appears to be never called... Why add it in this PR?

dimas-b (commented):
LGTM with some minor comments.

polaris-core/src/main/java/org/apache/polaris/core/identity/ServiceIdentityType.java
dimas-b: I understand that this code follows an existing pattern, so I'm not asking to redo the pattern... but I still do not understand the purpose of this enum value :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/secrets/SecretReference.java
XJDKC: Ack, will remove this to another PR!

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
XJDKC: Yeah, it's just a place holder for the future service identity injection. Forgot to remove it from this PR. Will remove it!

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionConfigInfoDpo.java
XJDKC: It's just a place holder for part 3

polaris-core/src/main/java/org/apache/polaris/core/secrets/SecretReference.java
dimas-b: No worries, I already read through all these changes :)

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionConfigInfoDpo.java
dimas-b: Ok, if it's used in part 3, it's ok to keep

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
dimas-b: if it's used in part 3, it's ok to keep

dimas-b:
@XJDKC : looks like my previous comments are totally optional... Please fix conflicts, though.

polaris-core/src/main/java/org/apache/polaris/core/identity/ServiceIdentityType.java
XJDKC: It's just a enum value for all the out-of-bounds codes

With this, we don't need to check `serviceIdentityType != null` since it will always return a nonnull value!

dimas-b: I do not see any Polaris code where `serviceIdentityType` is a target of method calls at all (where NPE could be possible) :)

dimas-b: My point in the original method is that if the idea is to keep working without error, then we're losing information in `fromCode()`. What is the benefit of returning a non-null value, when the original type code's meaning is lost?

XJDKC: Right now, ServiceIdentityType is not used. But In part 3, ServiceIdentityInfo will be assigned and injected to catalog entity and then serviceIdentityType is not null and will be used for the further identity resolution (resolve the identity, get the credential and then make a remote call)

dimas-b: To put it another way: why bother with an enum in this case when the type code is an `int` and code's behaviour appears to be based on `ServiceIdentityInfoDpo` sub-types' details ?

XJDKC: Just to make it slightly easier for downstream implementations, without needing to check whether the returned value is null. But yes, the original type code's meaning is lost.

But we should be able to original code if the returned type is `NULL` right, since the caller passes the code to this method

XJDKC: This  enum is a bridge between serialized (int) and runtime (class) representation (we don't want to compare the int with 1/2/3, the code is very confusing and hard to read).

We will use this `ServiceIdentityType` in both dpo and the runtime `ResolvedServiceIdentity` class (will be added in part 3).

dimas-b:
@dennishuo : waiting for your approval :)

dennishuo (approved):
LGTM! Sorry for the delay, forgot to hit submit

PR #2049, opened 2025-07-14
Package polaris client as python package

MonkeyCanCode:
Currently we are using `<git_root_dir>/polaris` as the entrypoint to interact with polaris CLI. This is a okay for local usage but it is not ideal for others who doesn't want to downlaod the entire code base to just use the CLI as well as making polaris client distributable as part of release. To address this issue, we will need to public polaris as an CLI without depends on the bash script.

This PR addressed this problem with following changes:
1. create main in `polaris_cli.py` as that will be the entrypoint for us to use
2. set scripts directly (which is point to item 1) and include both polaris and cli directories as part of packaging. Due to `client/python/polais` is auto generated from OpenAPI and `poetry.core.masonry.api` doe build backend (with `poetry.core.masonry.api` as build backend, it only includes files that are tracked by git. in this case, then auto generated files won't be track without including adding an include clause)
3. Update the invoke command when calling python `polaris` within bash script `polaris`

Here is a quick demo on the packaging:

setup poetry then perform a poetry build (which we can use later to distribute the client code as well as push to [Python Package Index](https://pypi.org/)):
```sh
➜  polaris git:(polaris_cli_package) python3 -m venv venv
➜  polaris git:(polaris_cli_package) source venv/bin/activate
(venv) ➜  polaris git:(polaris_cli_package) pip install -r regtests/requirements.txt
...
[notice] A new release of pip is available: 24.2 -> 25.1.1
[notice] To update, run: pip install --upgrade pip
(venv) ➜  python git:(polaris_cli_package) poetry build
Building polaris (1.0.0)
Building sdist
  - Building sdist
  - Built polaris-1.0.0.tar.gz
Building wheel
  - Building wheel
  - Built polaris-1.0.0-py3-none-any.whl
```

now assuming I am an user who doesn't want to download the code base but want to use the cli:
```sh
➜  Desktop python3 -m venv venv_dist
➜  Desktop source venv_dist/bin/activate
(venv_dist) ➜  Desktop pip freeze
(venv_dist) ➜  Desktop
(venv_dist) ➜  Desktop pip install GitHome/polaris/client/python/dist/polaris-1.0.0.tar.gz
...
Building wheels for collected packages: polaris
  Building wheel for polaris (pyproject.toml) ... done
  Created wheel for polaris: filename=polaris-1.0.0-py3-none-any.whl size=520413 sha256=9c9a25c6edb2a0b642666ab07abf5bc52f0a51f939095b18f12404bb118ec2e5
[...]

PR #2192, opened 2025-07-28
Python client auto generate

MonkeyCanCode:
This is a follow up PR based on https://github.com/apache/polaris/pull/2049. To give more context, currently we are using `client/templates/regenerate.sh` to regenerate code for python cli from openapi within docker. Then on gradle side, we have a CMD registers as `regeneratePythonClient` to provide user to regenerate the python code.

Based on the discussion from above PR as well as the experimental PR (https://github.com/apache/polaris/pull/1885) from @HonahX, I then raised this PR to move python code generation away from gradle and make it as part of poetry and remove dependency for docker. 

Unlikes the experimental PR, this PR has the following as extra:
1. ensure openapi only generate source code and another else (the one in the experimental PR will setup a bunch files and try to modify many existed files as well including pyproject.toml and README.md)
2. makefile and polaris bash script are updated to not depends on gradle for client code generation
3. remove client/templates/regenerate.sh as that is no longer needed (all functionalities are ported to the client/python/generate_clients.py)
4. copy header templates into docker for testing
5. minor fix in the Docker file with `&&`


MonkeyCanCode:
@snazy one of the efforts we talked about in the other PR where we will be moving away from gradle for non-java based workload such as python client related.

client/python/generate_clients.py
```diff
@@ -0,0 +1,267 @@
...
+            "generate",
+            "-i",
+            POLARIS_CATALOG_SPEC,
+            "-g",
+            "python",
+            "-o",
+            CLIENT_DIR,
+            PACKAGE_NAME_POLARIS_CATALOG,
+            "--additional-properties=apiNameSuffix=",
+            PYTHON_VERSION,
```
HonahX, line 158: [Not a blocker] I am thinking if we should also skip generating the api docs and model docs
```
"--global-property=apiDocs=false",
"--global-property=modelDocs=false",
```
I do not think these docs are very useful, especially when they are now auto-generated and not tracked by git

client/python/generate_clients.py
```diff
@@ -0,0 +1,267 @@
...
+            header_content = hf.read()
+        with open(file_path, "r+") as f:
+            original_content = f.read()
+            f.seek(0)
+            f.write(header_content + original_content)
+    except IOError as e:
+        print(f"Error prepending header to {file_path}: {e}")
+
+
+def prepend_licenses() -> None:
```
HonahX, line 204: Not sure if we need to add license header to auto-generated codes that are not tracked by git, but we can preserve the existing behavior and make a decision later

HonahX (approved):
@MonkeyCanCode Thanks for finalizing the auto-generation solution, I think this will also make it easier for us to publish the package in the future.

HonahX (commented):
I think we also need to update the scripts in getting-started 
https://github.com/apache/polaris/blob/5983c816596bdf7827ce6e7557835e1b5ed8dd09/getting-started/spark/launch-docker.sh#L21

https://github.com/apache/polaris/blob/5983c816596bdf7827ce6e7557835e1b5ed8dd09/spec/README.md?plain=1#L54-L56

HonahX (changes requested):
Temporarily block merging to prevent breaking getting-started examples

client/python/generate_clients.py
MonkeyCanCode: Yeah. I would suggest to keep the same then clean up those in the following PR. What do u think?

client/python/generate_clients.py
MonkeyCanCode: Resolved with `"--global-property apiDocs=false modelDocs=false"`

MonkeyCanCode:
> I think we also need to update the scripts in getting-started examples, otherwise it will break
> 
> https://github.com/apache/polaris/blob/5983c816596bdf7827ce6e7557835e1b5ed8dd09/getting-started/spark/launch-docker.sh#L21
> 
> https://github.com/apache/polaris/blob/5983c816596bdf7827ce6e7557835e1b5ed8dd09/spec/README.md?plain=1#L54-L56

will do.

MonkeyCanCode:
> > I think we also need to update the scripts in getting-started examples, otherwise it will break
> > https://github.com/apache/polaris/blob/5983c816596bdf7827ce6e7557835e1b5ed8dd09/getting-started/spark/launch-docker.sh#L21
> > 
> > https://github.com/apache/polaris/blob/5983c816596bdf7827ce6e7557835e1b5ed8dd09/spec/README.md?plain=1#L54-L56
> 
> will do.

This is resolved.

client/python/generate_clients.py
MonkeyCanCode: I had to revert the change for not auto generated doc. Seems to be causing issues on python 3.9 during build (i am on python 3.13 and no issue there). I can revisit this one later.

HonahX (approved):
Thanks for updating !

.github/workflows/python-client.yml
```diff
@@ -50,9 +50,6 @@ jobs:
           distribution: 'temurin'
           java-version: '21'
 
-      - name: Run regeneratePythonClient
-        run: ./gradlew regeneratePythonClient
-
```
HonahX, line 55: Could we also add a task in the CI to do sanity check that 
```
./polaris --help
```
works

MonkeyCanCode: Sure thing. I will add this tonight.

MonkeyCanCode: Sry for the delay. A bit busy with work lately. This is added.

eric-maynard (approved):

snazy:
Anything left on this PR?

client/python/generate_clients.py
MonkeyCanCode: PR: https://github.com/apache/polaris/pull/2439

PR #2194, opened 2025-07-28
Remove PolarisCallContext from BaseMetaStoreManager.extractStorageConfiguration

XN137:
as a static helper method its strongly preferable to not depend on the full `PolarisCallContext` if not necessary.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
@@ -1608,7 +1608,8 @@ public class AtomicOperationMetaStoreManager extends BaseMetaStoreManager {
             entityId);
 
     PolarisStorageConfigurationInfo storageConfigurationInfo =
-        BaseMetaStoreManager.extractStorageConfiguration(callCtx, reloadedEntity.getEntity());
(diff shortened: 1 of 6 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -1607,8 +1607,9 @@ private void revokeGrantRecord(
             catalogId,
             entityId);
 
+    PolarisBaseEntity reloadedEntity1 = reloadedEntity.getEntity();
```
dimas-b: do we really need this local var (given that the name is odd and easily confused with `reloadedEntity`)?

dimas-b (commented):
This change LGTM overall... A could of minor concerns with code changes below.

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
XN137: no we dont, I had used a inline method refactoring and it decided to introduce those vars without me checking the result sorry. i have removed them again.

dimas-b (approved):

adutra (approved):

PR #2195, opened 2025-07-28
Remove PolarisCallContext from TreeMapMetaStore

XN137:
the concrete class `TreeMapMetaStore` only uses `PolarisDiagnostics` from the given `PolarisCallContext`.

we also remember `initialDiagnosticServices` and reset it after a transaction for consistency (not sure if it will ever get used outside a transaction context though)

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapMetaStore.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapMetaStore.java
@@ -193,6 +193,7 @@ public class TreeMapMetaStore {
   private Transaction tr;
 
   // diagnostic services
+  private final PolarisDiagnostics initialDiagnosticServices;
   private PolarisDiagnostics diagnosticServices;
 
   // all entities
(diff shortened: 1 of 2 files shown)


dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapMetaStore.java
```diff
@@ -460,54 +454,47 @@ public void runActionInTransaction(
...
       try {
         // init diagnostic services
-        this.diagnosticServices = callCtx.getDiagServices();
+        this.diagnosticServices = diagnostics;
         this.startReadTransaction();
         return transactionCode.get();
       } finally {
         this.tr = null;
-        this.diagnosticServices = null;
+        this.diagnosticServices = this.initialDiagnosticServices;
```
eric-maynard, line 479: This aspect is a behavior change, but it seems to be a pretty reasonable one

eric-maynard (approved):

Issue #2010, opened 2025-07-08
Support database bootstrap in the admin tool


dimas-b:
### Is your feature request related to a problem? Please describe.

Currently the [Admin Tool](https://github.com/apache/polaris/tree/main/runtime/admin) has a `bootstrap` command that initializes "realms".

However, currently the JDBC persistence allows multiple realms to be stored within the same JDBC schema. 

Therefore, managing the database schema is a high-level concern than managing realms.

It would be nice to have a separate command in the Admin Tool dedicated specifically to running one-time Persistence setup tasks (e.g. creating JDBC schema).

sclee01:
@dimas-b  Makes sense. Schema setup should ideally come before realm init.
Mind if I pick this up and work on a separate command?

dimas-b:
@sclee01 : absolutely - thx for contributing :)

sclee01:
@dimas-b 

I looked into this and it seems that schema setup isn’t handled solely by the Admin Tool.  
The `bootstrapRealms()` method in `MetaStoreManagerFactory` currently runs the schema initialization script every time it's called — for each realm.

To support a proper one-time setup via a new Admin Tool command, we’d need to refactor this logic and move schema setup into a separate method (e.g. `setupSchema(SchemaOptions)`).  
Since `MetaStoreManagerFactory` is an interface, this would also require changes across related implementations and test classes.

Just wanted to check if you're okay with this broader scope before proceeding further.

dimas-b:
@sclee01 : I agree with your assessment :)

I'd personally prefer to introduce new interfaces dedicated to bootstrapping and schema setup so as to keep `MetaStoreManagerFactory` to routine runtime concerns (those that pertain to every API call).

Since this change is going to be touching old core interfaces (which is ok, IMHO), it might be worth mentioning your PR on the `dev` email list (when you have it) for visibility.

PR #2196, opened 2025-07-28
add the schema-setup for admin tool

sclee01:
### Motivation

This change is related to improving the Polaris Admin Tool.

Currently, the `bootstrap` command in the Admin Tool handles both schema initialization and realm bootstrapping. However, in JDBC persistence, multiple realms can share a single schema. Therefore, schema setup should be treated as a higher-level, one-time operation.

This PR addresses that by introducing a dedicated command.

### Description

- Introduced `SetupCommand` in the Admin Tool for running one-time persistence setup tasks (e.g., schema creation)
- Added a new interface: `SchemaInitializer`, with methods:
  - `setupSchema(SchemaOptions)`
  - `isSchemaPresent(String schemaName)`
- Refactored existing logic to delegate schema setup outside of `MetaStoreManagerFactory`
- Maintains backward compatibility (bootstrap still performs schema init if needed)

### Context

This change was proposed in the following issue:
- [Support database bootstrap in the admin tool #2010](https://github.com/apache/polaris/issues/2010
)

Based on discussion with @dimas-b.

### Next Steps

- I’m planning to mention this on the dev mailing list for broader visibility, in case there are any concerns or suggestions.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -41,7 +41,6 @@ import java.util.function.Consumer;
 import java.util.stream.Stream;
 import javax.sql.DataSource;
 import org.apache.polaris.core.persistence.EntityAlreadyExistsException;
-import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
 import org.apache.polaris.persistence.relational.jdbc.models.Converter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -117,7 +116,7 @@ public class DatasourceOperations {
     }
   }
 
-  public boolean checkSchemaExists(SchemaOptions schemaOptions) {
+  public boolean checkSchemaExists(String schemaName) {
     final boolean[] exists = {false};
 
     try {
@@ -125,7 +124,7 @@ public class DatasourceOperations {
           connection -> {
             try (ResultSet schemas = connection.getMetaData().getSchemas()) {
               while (schemas.next()) {
-                if (schemaOptions.schemaName().equalsIgnoreCase(schemas.getString("TABLE_SCHEM"))) {
+                if (schemaName.equalsIgnoreCase(schemas.getString("TABLE_SCHEM"))) {
                   exists[0] = true;
                   break;
                 }
@@ -134,8 +133,7 @@ public class DatasourceOperations {
             return true;
           });
     } catch (SQLException e) {
-      throw new RuntimeException(
-          "Failed to check schema existence: " + schemaOptions.schemaName(), e);
+      throw new RuntimeException("Failed to check schema existence: " + schemaName, e);
     }
 
     return exists[0];
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -41,6 +41,7 @@ import java.util.function.Consumer;
 import java.util.stream.Stream;
 import javax.sql.DataSource;
 import org.apache.polaris.core.persistence.EntityAlreadyExistsException;
+import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
 import org.apache.polaris.persistence.relational.jdbc.models.Converter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -116,27 +117,32 @@ public class DatasourceOperations {
     }
   }
 
-  public boolean checkSchemaExists(@Nonnull String schemaName) {
-    final String sql = "SELECT 1 FROM information_schema.schemata WHERE schema_name = ?";
+  public boolean checkSchemaExists(SchemaOptions schemaOptions) {
     final boolean[] exists = {false};
 
     try {
-      runWithinTransaction(
-          connection -> {
-            try (PreparedStatement statement = connection.prepareStatement(sql)) {
-              statement.setString(1, schemaName);
-              try (ResultSet resultSet = statement.executeQuery()) {
-                exists[0] = resultSet.next();
-              }
+      runWithinTransaction(connection -> {
+        try (ResultSet schemas = connection.getMetaData().getSchemas()) {
+          while (schemas.next()) {
+            if (schemaOptions.schemaName().equalsIgnoreCase(schemas.getString("TABLE_SCHEM"))) {
+              exists[0] = true;
+              break;
             }
-            return true;
-          });
+          }
+        }
+        return true;
+      });
     } catch (SQLException e) {
-      throw new RuntimeException(e);
+      throw new RuntimeException("Failed to check schema existence: " + schemaOptions.schemaName(), e);
     }
 
     return exists[0];
   }
+
+
+
+
+
 
   /**
    * Executes SELECT Query and returns the results after applying a transformer
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -116,6 +116,31 @@ public class DatasourceOperations {
     }
   }
 
+
+  public boolean checkSchemaExists(@Nonnull String schemaName) {
+    final String sql = "SELECT 1 FROM information_schema.schemata WHERE schema_name = ?";
+    final boolean[] exists = {false};
+
+    try {
+      runWithinTransaction(connection -> {
+        try (PreparedStatement statement = connection.prepareStatement(sql)) {
+          statement.setString(1, schemaName);
+          try (ResultSet resultSet = statement.executeQuery()) {
+            exists[0] = resultSet.next();
+          }
+        }
+        return true;
+      });
+    } catch (SQLException e) {
+      throw new RuntimeException(e);
+    }
+
+    return exists[0];
+  }
+
+
+
+
   /**
    * Executes SELECT Query and returns the results after applying a transformer
    *
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -121,28 +121,25 @@ public class DatasourceOperations {
     final boolean[] exists = {false};
 
     try {
-      runWithinTransaction(connection -> {
-        try (ResultSet schemas = connection.getMetaData().getSchemas()) {
-          while (schemas.next()) {
-            if (schemaOptions.schemaName().equalsIgnoreCase(schemas.getString("TABLE_SCHEM"))) {
-              exists[0] = true;
-              break;
+      runWithinTransaction(
+          connection -> {
+            try (ResultSet schemas = connection.getMetaData().getSchemas()) {
+              while (schemas.next()) {
+                if (schemaOptions.schemaName().equalsIgnoreCase(schemas.getString("TABLE_SCHEM"))) {
+                  exists[0] = true;
+                  break;
+                }
+              }
             }
-          }
-        }
-        return true;
-      });
+            return true;
+          });
     } catch (SQLException e) {
-      throw new RuntimeException("Failed to check schema existence: " + schemaOptions.schemaName(), e);
+      throw new RuntimeException(
+          "Failed to check schema existence: " + schemaOptions.schemaName(), e);
     }
 
     return exists[0];
   }
-
-
-
-
-
 
   /**
    * Executes SELECT Query and returns the results after applying a transformer
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -116,30 +116,27 @@ public class DatasourceOperations {
     }
   }
 
-
   public boolean checkSchemaExists(@Nonnull String schemaName) {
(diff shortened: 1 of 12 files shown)


sclee01:
Fix the code style to pass the unit test.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -116,6 +116,28 @@ public void executeScript(InputStream scriptInputStream) throws SQLException {
     }
   }
 
+  public boolean checkSchemaExists(@Nonnull String schemaName) {
+    final String sql = "SELECT 1 FROM information_schema.schemata WHERE schema_name = ?";
```
dimas-b: Is `information_schema.schemata` standard? :thinking: Could it be preferable to use JDBC APIs (`Connection.getMetaData()`) for  this and let the driver figure out the right source for this data?

polaris-core/src/main/java/org/apache/polaris/core/persistence/SchemaInitializer.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.polaris.core.persistence;
+
+import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
+
+public interface SchemaInitializer {
+
+  boolean setupSchema(SchemaOptions schemaOptions);
+
+  boolean isSchemaPresent(String schemaName);
```
dimas-b: `schemaName` looks like a concept leak in this interface. `SchemaOptions` does not define a "name"... How can the caller know what schema name to ask for?

runtime/admin/src/main/java/org/apache/polaris/admintool/BaseCommand.java
```diff
@@ -29,8 +30,11 @@ public abstract class BaseCommand implements Callable<Integer> {
   public static final int EXIT_CODE_USAGE = 2;
   public static final int EXIT_CODE_BOOTSTRAP_ERROR = 3;
   public static final int EXIT_CODE_PURGE_ERROR = 4;
+  public static final int EXIT_CODE_SCHEMA_ERROR = 5;
 
   @Inject MetaStoreManagerFactory metaStoreManagerFactory;
 
+  @Inject SchemaInitializer schemaInitializer;
```
dimas-b: Is this common to all commands? I tend to think it's only for `bootstrap` and `setup` :thinking:

runtime/admin/src/main/java/org/apache/polaris/admintool/config/QuarkusProducers.java
```diff
@@ -43,6 +44,12 @@ public MetaStoreManagerFactory metaStoreManagerFactory(
     return metaStoreManagerFactories.select(Identifier.Literal.of(persistenceType)).get();
   }
 
+  @Produces
+  @ApplicationScoped
+  public SchemaInitializer schemaInitializer(MetaStoreManagerFactory metaStoreManagerFactory) {
+    return (SchemaInitializer) metaStoreManagerFactory;
```
dimas-b, line 50: Please use `.select()` as in `metaStoreManagerFactory()` for proper type handling.

dimas-b (commented):
Thanks for working on this, @sclee01 ! Some comments below.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
sclee01: I thought this would stay PostgreSQL-only, so I used the query directly. But updated to use getMetaData() for better portability.

runtime/admin/src/main/java/org/apache/polaris/admintool/BaseCommand.java
sclee01: Moved it to SetupCommand since it's not used elsewhere.

polaris-core/src/main/java/org/apache/polaris/core/persistence/SchemaInitializer.java
sclee01: I changed it to take SchemaOptions instead for better encapsulation.

runtime/admin/src/main/java/org/apache/polaris/admintool/config/QuarkusProducers.java
sclee01: I didn't use `.select()` because only `JdbcMetaStoreManagerFactory` implements that interface.
If I use `.select()`, it could cause errors at runtime when using other `MetaStoreManagerFactory` implementations that don't implement `SchemaInitializer`. Given that, I kept the cast with a type check to avoid breaking compatibility. Open to suggestions if there's a safer way to handle this.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
dimas-b: PostgreSQL is the only currently supported database. I would not want Polaris to be limited to PostgreSQL forever :)

runtime/admin/src/main/java/org/apache/polaris/admintool/config/QuarkusProducers.java
dimas-b: ... but having those runtime errors is good, IMHO :) CDI errors should only happen if the tool setup calls for a `SchemaInitializer` but the configuration does not allow for an implementation to be found. All in all, this would lead to a clear error, whereas, a type case would lead to a very obscure error.

polaris-core/src/main/java/org/apache/polaris/core/persistence/SchemaInitializer.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.polaris.core.persistence;
+
+import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
+
+public interface SchemaInitializer {
+
+  boolean setupSchema(SchemaOptions schemaOptions);
+
+  boolean isSchemaPresent(SchemaOptions schemaName);
```
dimas-b: nit: `schemaName` -> `schemaOptions`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/SchemaOptions.java
```diff
@@ -31,6 +32,17 @@ public interface SchemaOptions {
...
 
+  @Value.Default
+  default boolean setupSchema() {
+    return true;
+  }
+
+  @NotNull
+  @Value.Default
+  default String schemaName() {
+    return "POLARIS_SCHEMA";
```
dimas-b: Isn't this kind of asymmetric now? `SchemaOptions` allow callers to set any schema name, but persistence implementations do not use it for actually creating the schema :thinking:  ... or did I miss it?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -116,6 +116,29 @@ public void executeScript(InputStream scriptInputStream) throws SQLException {
...
 
+  public boolean checkSchemaExists(String schemaName) {
+    final boolean[] exists = {false};
+
+    try {
+      runWithinTransaction(
+          connection -> {
+            try (ResultSet schemas = connection.getMetaData().getSchemas()) {
+              while (schemas.next()) {
+                if (schemaName.equalsIgnoreCase(schemas.getString("TABLE_SCHEM"))) {
```
adutra, line 127:
```suggestion
                if (schemaName.equalsIgnoreCase(schemas.getString("TABLE_SCHEMA"))) {
```

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -143,15 +146,16 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
-          // Run the set-up script to create the tables.
-          datasourceOperations.executeScript(
-              datasourceOperations
-                  .getDatabaseType()
-                  .openInitScriptResource(bootstrapOptions.schemaOptions()));
-        } catch (SQLException e) {
-          throw new RuntimeException(
-              String.format("Error executing sql script: %s", e.getMessage()), e);
+        if (bootstrapOptions.schemaOptions().setupSchema()) {
+          setupSchema(bootstrapOptions.schemaOptions());
```
adutra, line 150: We are doing this setup per realm, but the schema cannot change from one realm to another. I think this should be done in a startup method. This would be beneficial to both the admin tool and the server imho.

polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/SchemaOptions.java
```diff
@@ -31,6 +31,11 @@ public interface SchemaOptions {
   @Nullable
   String schemaFile();
 
+  @Value.Default
+  default boolean setupSchema() {
```
adutra, line 35: I fell like this should go into application.properties instead. We are creating a second layer of configuration that is only accessible to the admin tool (I don't see how the server runtime can change this value).

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
```diff
@@ -80,6 +80,12 @@ static class FileInputOptions {
     }
 
     static class SchemaInputOptions {
+      @CommandLine.Option(
```
adutra, line 83: The schema init scripts _must_ be idempotent, because the server runtime executes them whenever it loads a new realm. So why do we need to specify this here? Couldn't we just assume that schema should be set up always? If we want to avoid re-running init scripts on an already-initialized database, this is a concern that affects both the admin tool and the server. IOW this should be controlled by a configuration option in application.properties.

adutra (commented):
> Currently, the bootstrap command in the Admin Tool handles both schema initialization and realm bootstrapping. However, in JDBC persistence, multiple realms can share a single schema. Therefore, schema setup should be treated as a higher-level, one-time operation.

I agree with the rationale, but to me this isn't a concern of the admin tool only. The server too is running schema setup for each realm it creates AFAIU. I think we need to extract schema setup to a startup method (that observes the `@Startup` event), then that method would inspect some configuration property to determine if setup should be performed. This would benefit both the server and the admin tool. WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
adutra: If we want to support a schemas-per-realm kind of deployment (which is not the case today), then we'd need another config property to indicate which kind of schema we are using, and when to set it up (at startup vs for each realm).

adutra:
@sclee01 thank you for your commitment!

To clarify what I have in mind: I think a configuration interface like below would be the right way to proceed:

```java
@ConfigMapping(prefix = "polaris.schema")
public interface SchemaSetupConfiguration {

  /**
   * Whether to set up the schema at startup.
   * 
   * <p>This is useful for development and testing purposes, but should be disabled in production
   * environments where the schema is managed externally.
   *
   * @return true if the schema should be set up at startup, false otherwise.
   */
  @WithName("setup-at-startup")
  @WithDefault("true")
  boolean setupSchemaAtStartup();
}
```

If you really want to give users to choice when to do the setup, then something like below could work:

```java
@ConfigMapping(prefix = "polaris.schema")
public interface SchemaSetupConfiguration {

  enum SchemaSetupMode {
    AT_STARTUP,
    PER_REALM,
    NEVER
  }

  /**
   * The mode of schema setup to use.
   *
   * @return the schema setup mode
   */
  @WithDefault("AT_STARTUP")
  SchemaSetupMode mode();
}
```

sclee01:
@adutra  @dimas-b  
I’d love to get your thoughts on this.

I initially added this to support explicit one-time schema setup from the Admin Tool side. But given that the scripts are idempotent, I’m now wondering if separating this logic is really necessary — maybe it’s fine to run the setup every time a realm is created.

That said, if the separation is implemented based on the reviews, I’m also thinking it could be beneficial in the long run — for example, it gives us flexibility if we ever want to support different schema strategies (e.g. start up, realm .. ) or control setup behavior more explicitly.

Let me know what you think — happy to drop it if it feels unnecessary, but also open to keeping it if it adds value.

adutra:
Agreed let's gather some consensus. I'd also like to hear from @eric-maynard and @singhpk234.

(Maybe this warrants a DISCUSS thread on the ML btw)

snazy:
The most user-friendly way IMO is to let Polaris perform the schema initialization on bootstrap automatically. @adutra already provided some useful additional configuration options for "special" scenarios.

dimas-b:
I think it's a good idea for the `bootstrap` command to also setup the schema by default (nicer user experience as @snazy commented).

Still, I believe the `bootstrap` command show have the option to _not_ try to create the schema (it may or may not be idempotent), in which case the admin user can explicitly run the `setup` command.

Subsequently, there is already an existing [mechanism](https://github.com/apache/polaris/blob/ca85339deee9645f0682194141ae8445ab3af768/runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java#L238) for Polaris Servers to perform "bootstrap" automatically on startup. 

With all three options combined, I think, the "getting started" experience can remain simple, while more advanced scenarios are still possible for users that need them.

That said, I do _not_ think that JDBC Persistence code should run any DDL on the main call path (i.e. outside of "bootstrap" or "setup" use cases).

sclee01:
@dimas-b I agree with your suggestion that JDBC Persistence code should not run DDL on the main call path — having it configurable makes a lot of sense.
That said, I think it’s worth hearing a few more thoughts before we settle on the direction.
Maybe we can spin up a DISCUSS thread if needed.

adutra:
@sclee01 I went ahead and started the thread: https://lists.apache.org/thread/6c7lzgy3675jrrtbrrxsm8l0xd1fo1mj

Please chime in! 🙏

Issue #2168, opened 2025-07-23
Consolidate usage of CommitFailedException and CommitConflictException


dimas-b:
### Is your feature request related to a problem? Please describe.

See discussions in #1989 for context.

Cf. https://github.com/apache/polaris/pull/1989#discussion_r2226045794

tmater:
I just want to confirm that I understand the issue correctly.

Currently, in some places, we're using `CommitFailedException`. While it generally describes the situation and returns the correct 409 response, it's not the best fit since it's an Iceberg-specific exception. In some of these cases, the conflict actually originates from Polaris.

We should review where `CommitFailedException` is used and consider replacing it with a more appropriate and specific `CommitConflictException` when the conflict comes from Polaris.

dimas-b:
Yes, `CommitFailedException` is currently used by Polaris code to represent eventual 409 errors in the REST API. However, Polaris uses a lot of non-REST Iceberg code and `CommitFailedException` may potentially occur in other contexts, where it is not clear what the appropriate REST status code should be.

IMHO, it is preferable to use `CommitConflictException`, which is owned by Polaris and is intended to clearly map to 409 at the REST level.

Issue #624, opened 2025-01-08
Root user cannot rotate principal credentials


Gerrit-K:
### Describe the bug

When trying to rotate a principal's credentials via the root user, an HTTP error with code 403 is returned:
```
{"error":{"message":"Principal 'root' with activated PrincipalRoles '[]' and activated grants via '[service_admin]' is not authorized for op ROTATE_CREDENTIALS","type":"ForbiddenException","code":403}}
```

### To Reproduce

1. Deploy locally via `docker compose up --build`
2. Note down the root principal credentials and store them in a shell variable, e.g.
    ```shell
    CLIENT_ID=f69f1990657205d9
    CLIENT_SECRET=e89026b4e87d60b7bd6ea75adb16e6f8
    ```
3. Get an access token for the root principal
    ```shell
    TOKEN="$(curl --request POST "http://localhost:8181/api/catalog/v1/oauth/tokens?grant_type=client_credentials&scope=PRINCIPAL_ROLE%3AALL&client_id=${CLIENT_ID}&client_secret=${CLIENT_SECRET}" \
    --header 'content-type: application/x-www-form-urlencoded' \
    | jq -r .access_token)"
    ```
4. Create a new principal
    ```shell
    curl --request POST http://localhost:8181/api/management/v1/principals \
    --header "Authorization: Bearer $TOKEN" \
    --header "Content-Type: application/json" \
    -d '{"name": "test"}')"
    ```
5. Try to rotate the credentials of the new principal using the same token as it was created with
    ```shell
    curl --request POST http://localhost:8181/api/management/v1/principals/test/rotate \
    --header "Authorization: Bearer $TOKEN"
    ```

### Actual Behavior

The service returns a 403 response
```
{"error":{"message":"Principal 'root' with activated PrincipalRoles '[]' and activated grants via '[service_admin]' is not authorized for op ROTATE_CREDENTIALS","type":"ForbiddenException","code":403}}
```

### Expected Behavior

The principal credentials should successfully be rotated

eric-maynard:
IIRC this is by design, users can rotate their own credentials but root cannot change them.

Gerrit-K:
Hm, that seems a bit strange to me, to be honest 🤔 A not-so-rare use case for secret rotation is to restore access after credentials have been lost (by whatever means). If the root principal cannot do this, then there is literally no way to recover credentials, so the principal needs to be deleted and a new one set up, including setting up all the roles and permissions.

dimas-b:
[...]

PR #2197, opened 2025-07-28
Expose resetCredentials via api to allow root user to reset credentials for an existing principal with custom values 

fivetran-arunsuri:
Background:
See Issue -https://github.com/apache/polaris/issues/1929 and based on Dev email discussion
WHAT

1. Exposes the resetCredentials operation via the api
2. Only Root user can reset the credentials of the existing principal
3. Requires the custom clientId and custom secret via root user to reset the random credentials already created via CreatePrincipal API
4. Needs ENFORCE_PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_CHECKING to be set to true

**Local Testing:**
Happy Scenario
image

Failure scenario:
image

### DIFF
--- a/persistence/eclipselink/src/main/java/org/apache/polaris/extension/persistence/impl/eclipselink/PolarisEclipseLinkMetaStoreSessionImpl.java
+++ b/persistence/eclipselink/src/main/java/org/apache/polaris/extension/persistence/impl/eclipselink/PolarisEclipseLinkMetaStoreSessionImpl.java
@@ -785,4 +785,10 @@ public class PolarisEclipseLinkMetaStoreSessionImpl extends AbstractTransactiona
           @Nonnull PolarisCallContext callContext, T entity) {
     return Optional.empty();
   }
+
+  @Nullable
+  @Override
+  public PolarisPrincipalSecrets resetPrincipalSecrets(@Nonnull PolarisCallContext callCtx, @Nonnull String clientId, long principalId, boolean reset, @Nonnull String oldSecretHash, String customClientId, String customClientSecret) {
+    return null;
+  }
 }
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -774,38 +774,37 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
     return principalSecrets;
   }
 
-
   @Nullable
   @Override
   public PolarisPrincipalSecrets resetPrincipalSecrets(
-          @Nonnull PolarisCallContext callCtx,
-          @Nonnull String clientId,
-          long principalId,
-          boolean reset,
-          @Nonnull String oldSecretHash,
-          String customClientId,
-          String customClientSecret) {
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      boolean reset,
+      @Nonnull String oldSecretHash,
+      String customClientId,
+      String customClientSecret) {
     PolarisPrincipalSecrets principalSecrets = loadPrincipalSecrets(callCtx, clientId);
 
     // should be found
     callCtx
-            .getDiagServices()
-            .checkNotNull(
-                    principalSecrets,
-                    "cannot_find_secrets",
-                    "client_id={} principalId={}",
-                    clientId,
-                    principalId);
+        .getDiagServices()
+        .checkNotNull(
+            principalSecrets,
+            "cannot_find_secrets",
+            "client_id={} principalId={}",
+            clientId,
+            principalId);
 
     // ensure principal id is matching
     callCtx
-            .getDiagServices()
-            .check(
-                    principalId == principalSecrets.getPrincipalId(),
-                    "principal_id_mismatch",
-                    "expectedId={} id={}",
-                    principalId,
-                    principalSecrets.getPrincipalId());
+        .getDiagServices()
+        .check(
+            principalId == principalSecrets.getPrincipalId(),
+            "principal_id_mismatch",
+            "expectedId={} id={}",
+            principalId,
+            principalSecrets.getPrincipalId());
 
     principalSecrets.setPrincipalClientId(customClientId);
     principalSecrets.resetSecrets(customClientSecret);
@@ -813,25 +812,25 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
     Map<String, Object> params = Map.of("principal_client_id", clientId, "realm_id", realmId);
     try {
       ModelPrincipalAuthenticationData modelPrincipalAuthenticationData =
-              ModelPrincipalAuthenticationData.fromPrincipalAuthenticationData(principalSecrets);
+          ModelPrincipalAuthenticationData.fromPrincipalAuthenticationData(principalSecrets);
       datasourceOperations.executeUpdate(
-              QueryGenerator.generateUpdateQuery(
-                      ModelPrincipalAuthenticationData.ALL_COLUMNS,
-                      ModelPrincipalAuthenticationData.TABLE_NAME,
-                      modelPrincipalAuthenticationData
-                              .toMap(datasourceOperations.getDatabaseType())
-                              .values()
-                              .stream()
-                              .toList(),
-                      params));
+          QueryGenerator.generateUpdateQuery(
+              ModelPrincipalAuthenticationData.ALL_COLUMNS,
+              ModelPrincipalAuthenticationData.TABLE_NAME,
+              modelPrincipalAuthenticationData
+                  .toMap(datasourceOperations.getDatabaseType())
+                  .values()
+                  .stream()
+                  .toList(),
+              params));
     } catch (SQLException e) {
       LOGGER.error(
-              "Failed to reset PrincipalSecrets  for clientId: {}, due to {}",
-              clientId,
-              e.getMessage(),
-              e);
+          "Failed to reset PrincipalSecrets  for clientId: {}, due to {}",
+          clientId,
+          e.getMessage(),
+          e);
       throw new RuntimeException(
-              String.format("Failed to reset PrincipalSecrets for clientId: %s", clientId), e);
+          String.format("Failed to reset PrincipalSecrets for clientId: %s", clientId), e);
     }
 
     // return those
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -774,6 +774,70 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
     return principalSecrets;
   }
 
+
+  @Nullable
+  @Override
+  public PolarisPrincipalSecrets resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret) {
+    PolarisPrincipalSecrets principalSecrets = loadPrincipalSecrets(callCtx, clientId);
+
+    // should be found
+    callCtx
+            .getDiagServices()
+            .checkNotNull(
+                    principalSecrets,
+                    "cannot_find_secrets",
+                    "client_id={} principalId={}",
+                    clientId,
+                    principalId);
+
+    // ensure principal id is matching
+    callCtx
+            .getDiagServices()
+            .check(
+                    principalId == principalSecrets.getPrincipalId(),
+                    "principal_id_mismatch",
+                    "expectedId={} id={}",
+                    principalId,
+                    principalSecrets.getPrincipalId());
+
+    principalSecrets.setPrincipalClientId(customClientId);
+    principalSecrets.resetSecrets(customClientSecret);
+
+    Map<String, Object> params = Map.of("principal_client_id", clientId, "realm_id", realmId);
+    try {
+      ModelPrincipalAuthenticationData modelPrincipalAuthenticationData =
+              ModelPrincipalAuthenticationData.fromPrincipalAuthenticationData(principalSecrets);
+      datasourceOperations.executeUpdate(
+              QueryGenerator.generateUpdateQuery(
+                      ModelPrincipalAuthenticationData.ALL_COLUMNS,
+                      ModelPrincipalAuthenticationData.TABLE_NAME,
+                      modelPrincipalAuthenticationData
+                              .toMap(datasourceOperations.getDatabaseType())
+                              .values()
+                              .stream()
+                              .toList(),
+                      params));
+    } catch (SQLException e) {
+      LOGGER.error(
+              "Failed to reset PrincipalSecrets  for clientId: {}, due to {}",
+              clientId,
+              e.getMessage(),
+              e);
+      throw new RuntimeException(
+              String.format("Failed to reset PrincipalSecrets for clientId: %s", clientId), e);
+    }
+
+    // return those
+    return principalSecrets;
+  }
+
   @Nullable
   @Override
   public PolarisPrincipalSecrets rotatePrincipalSecrets(
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecretsManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecretsManager.java
@@ -54,4 +54,18 @@ public interface PolarisSecretsManager {
       long principalId,
       boolean reset,
       @Nonnull String oldSecretHash);
+
+  @Nonnull
+  PrincipalSecretsResult resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret);
+
+
 }
+
+
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
@@ -36,7 +36,7 @@ public class PolarisPrincipalSecrets {
   private final long principalId;
 
   // the client id for that principal
-  private final String principalClientId;
+  private String principalClientId;
 
   // the main secret hash for that principal
   private String mainSecret;
@@ -145,6 +145,19 @@ public class PolarisPrincipalSecrets {
 
     this.mainSecret = this.generateRandomHexString(32);
     this.mainSecretHash = hashSecret(mainSecret);
+  }
+
+  /** Reset the main secrets */
+  public void resetSecrets(String customClientSecret) {
+    this.mainSecret = customClientSecret;
+    this.mainSecretHash = hashSecret(mainSecret);
+
+    this.secondarySecret = null;
+    this.secondarySecretHash = hashSecret(mainSecret);
+  }
+
+  public void setPrincipalClientId(String customClientId) {
+    this.principalClientId = customClientId;
   }
 
   public long getPrincipalId() {
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
@@ -897,6 +897,59 @@ public class AtomicOperationMetaStoreManager extends BaseMetaStoreManager {
         : new PrincipalSecretsResult(secrets);
   }
 
+  @Override
+  public @Nonnull PrincipalSecretsResult resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret) {
+    // get metastore we should be using
+    BasePersistence ms = callCtx.getMetaStore();
+
+    // if not found, the principal must have been dropped
+    EntityResult loadEntityResult =
+            loadEntity(
+                    callCtx, PolarisEntityConstants.getNullId(), principalId, PolarisEntityType.PRINCIPAL);
+    if (loadEntityResult.getReturnStatus() != BaseResult.ReturnStatus.SUCCESS) {
+      return new PrincipalSecretsResult(BaseResult.ReturnStatus.ENTITY_NOT_FOUND, null);
+    }
+
+    PolarisBaseEntity principal = loadEntityResult.getEntity();
+    Map<String, String> internalProps =
+            PolarisObjectMapperUtil.deserializeProperties(
+                    principal.getInternalProperties() == null ? "{}" : principal.getInternalProperties());
+
+    boolean doReset =
+            reset
+                    || internalProps.get(
+                    PolarisEntityConstants.PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_STATE)
+                    != null;
+    PolarisPrincipalSecrets secrets =
+            ((IntegrationPersistence) ms)
+                    .resetPrincipalSecrets(callCtx, clientId, principalId, doReset, oldSecretHash,customClientId, customClientSecret);
+
+    PolarisBaseEntity.Builder principalBuilder = new PolarisBaseEntity.Builder(principal);
+    if (reset
+            && !internalProps.containsKey(
+            PolarisEntityConstants.PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_STATE)
+            && customClientId != null
+            && customClientSecret != null) {
+      internalProps.put(
+              PolarisEntityConstants.PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_STATE, "true");
+      principalBuilder.internalProperties(
+              PolarisObjectMapperUtil.serializeProperties(internalProps));
+      principalBuilder.entityVersion(principal.getEntityVersion() + 1);
+      ms.writeEntity(callCtx, principalBuilder.build(), true, principal);
+    }
+
+    return (secrets == null)
+            ? new PrincipalSecretsResult(BaseResult.ReturnStatus.ENTITY_NOT_FOUND, null)
+            : new PrincipalSecretsResult(secrets);
+  }
+
   /** {@inheritDoc} */
   @Override
   public @Nonnull EntityResult createEntityIfNotExists(
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/IntegrationPersistence.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/IntegrationPersistence.java
@@ -82,6 +82,29 @@ public interface IntegrationPersistence {
       @Nonnull String oldSecretHash);
 
   /**
+   * Reset the secrets of a principal entity, i.e. make the specified secrets as main and secondary
+   * and assign a new client id
+   *
+   * @param callCtx call context
+   * @param clientId principal client id
+   * @param principalId principal id
+   * @param reset true if the principal secrets should be disabled and replaced with a one-time
+   *     password
+   * @param oldSecretHash the principal secret's old main secret hash
+   * @param customClientId the principal secret's old main secret hash
+   * @param customClientSecret the principal secret's old main secret hash
+   */
+  @Nullable
+  PolarisPrincipalSecrets resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret);
+
+  /**
    * When dropping a principal, we also need to drop the secrets of that principal
    *
    * @param callCtx the call context
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
@@ -952,6 +952,22 @@ public class TransactionalMetaStoreManagerImpl extends BaseMetaStoreManager {
         : new PrincipalSecretsResult(secrets);
   }
 
+  @Override
+  public @Nonnull PrincipalSecretsResult resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret) {
+    // get metastore we should be using
+    callCtx
+            .getDiagServices()
+            .fail("illegal_method_in_transaction_workspace", "resetPrincipalSecrets");
+    return null;
+  }
+
   /** {@inheritDoc} */
   @Override
   public @Nonnull CreateCatalogResult createCatalog(
--- a/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
+++ b/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
@@ -245,8 +245,7 @@ public class PolarisAdminService {
         resolutionManifest.getResolvedTopLevelEntity(topLevelEntityName, entityType);
 
     if (op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS)) {
-      boolean isRoot = authenticatedPrincipal.getPrincipalEntity().getId() == getRootEntityId()
-              && getRootPrincipalName().equals(authenticatedPrincipal.getPrincipalEntity().getName());
+      boolean isRoot = getRootPrincipalName().equals(authenticatedPrincipal.getPrincipalEntity().getName());
       if (!isRoot) {
         throw new ForbiddenException("Only root principal can reset credentials");
       }
--- a/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
+++ b/service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
@@ -20,10 +20,13 @@ package org.apache.polaris.service.admin;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static org.apache.polaris.core.entity.PolarisEntityConstants.getRootEntityId;
+import static org.apache.polaris.core.entity.PolarisEntityConstants.getRootPrincipalName;
 
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotNull;
+import jakarta.ws.rs.ForbiddenException;
 import jakarta.ws.rs.core.SecurityContext;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -63,6 +66,7 @@ import org.apache.polaris.core.admin.model.PolicyGrant;
 import org.apache.polaris.core.admin.model.PolicyPrivilege;
 import org.apache.polaris.core.admin.model.PrincipalWithCredentials;
 import org.apache.polaris.core.admin.model.PrincipalWithCredentialsCredentials;
+import org.apache.polaris.core.admin.model.ResetPrincipalRequest;
 import org.apache.polaris.core.admin.model.TableGrant;
 import org.apache.polaris.core.admin.model.TablePrivilege;
 import org.apache.polaris.core.admin.model.UpdateCatalogRequest;
@@ -240,12 +244,24 @@ public class PolarisAdminService {
     PolarisResolvedPathWrapper topLevelEntityWrapper =
         resolutionManifest.getResolvedTopLevelEntity(topLevelEntityName, entityType);
 
+    if (op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS)) {
+      boolean isRoot = authenticatedPrincipal.getPrincipalEntity().getId() == getRootEntityId()
+              && getRootPrincipalName().equals(authenticatedPrincipal.getPrincipalEntity().getName());
+      if (!isRoot) {
+        throw new ForbiddenException("Only root principal can reset credentials");
+      }
+      LOGGER
+              .atDebug()
+              .addKeyValue("principalName", topLevelEntityName)
+              .log("Root principal allowed to reset credentials");
+      return;
+    }
+
     // TODO: If we do add more "self" privilege operations for PRINCIPAL targets this should
     // be extracted into an EnumSet and/or pushed down into PolarisAuthorizer.
     if (topLevelEntityWrapper.getResolvedLeafEntity().getEntity().getId()
             == authenticatedPrincipal.getPrincipalEntity().getId()
-        && (op.equals(PolarisAuthorizableOperation.ROTATE_CREDENTIALS)
-            || op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS))) {
+        && (op.equals(PolarisAuthorizableOperation.ROTATE_CREDENTIALS))) {
       LOGGER
           .atDebug()
           .addKeyValue("principalName", topLevelEntityName)
@@ -651,7 +667,7 @@ public class PolarisAdminService {
    * references to the stored secret.
    */
   private Map<String, UserSecretReference> extractSecretReferences(
-      CreateCatalogRequest catalogRequest, PolarisEntity forEntity) {
+          CreateCatalogRequest catalogRequest, PolarisEntity forEntity) {
     Map<String, UserSecretReference> secretReferences = new HashMap<>();
     Catalog catalog = catalogRequest.getCatalog();
     UserSecretsManager secretsManager = getUserSecretsManager();
@@ -1120,6 +1136,70 @@ public class PolarisAdminService {
             newSecrets.getPrincipalClientId(), newSecrets.getMainSecret()));
   }
 
+  private @Nonnull PrincipalWithCredentials resetCredentialsHelper(
+          String principalName, boolean shouldReset, String customClientId, String customClientSecret) {
+    PrincipalEntity currentPrincipalEntity =
+            findPrincipalByName(principalName)
+                    .orElseThrow(() -> new NotFoundException("Principal %s not found", principalName));
+
+    if (FederatedEntities.isFederated(currentPrincipalEntity)) {
+      throw new ValidationException(
+              "Cannot reset credentials for a federated principal: %s", principalName);
+    }
+    PolarisPrincipalSecrets currentSecrets =
+            metaStoreManager
+                    .loadPrincipalSecrets(getCurrentPolarisContext(), currentPrincipalEntity.getClientId())
+                    .getPrincipalSecrets();
+    if (currentSecrets == null) {
+      throw new IllegalArgumentException(
+              String.format("Failed to load current secrets for principal '%s'", principalName));
+    }
+    PolarisPrincipalSecrets newSecrets =
+            metaStoreManager
+                    .resetPrincipalSecrets(
+                            getCurrentPolarisContext(),
+                            currentPrincipalEntity.getClientId(),
+                            currentPrincipalEntity.getId(),
+                            shouldReset,
+                            currentSecrets.getMainSecretHash(),
+                            customClientId,
+                            customClientSecret)
+                    .getPrincipalSecrets();
+    if (newSecrets == null) {
+      throw new IllegalStateException(
+              String.format("Failed to %s secrets for principal '%s'", "reset", principalName));
+    }
+    PolarisEntity newPrincipal =
+            PolarisEntity.of(
+                    metaStoreManager.loadEntity(
+                            getCurrentPolarisContext(),
+                            0L,
+                            currentPrincipalEntity.getId(),
+                            currentPrincipalEntity.getType()));
+
+    PrincipalEntity newPrincipalEntity = PrincipalEntity.of(newPrincipal);
+    if (customClientId != null && customClientSecret != null) {
+      PrincipalEntity.Builder updateBuilder = new PrincipalEntity.Builder(newPrincipalEntity);
+      updateBuilder.setClientId(newSecrets.getPrincipalClientId());
+      PrincipalEntity updatedNewPrincipalEntity = updateBuilder.build();
+      updatedNewPrincipalEntity =
+              Optional.ofNullable(
+                              PrincipalEntity.of(
+                                      PolarisEntity.of(
+                                              metaStoreManager.updateEntityPropertiesIfNotChanged(
+                                                      getCurrentPolarisContext(), null, updatedNewPrincipalEntity))))
+                      .orElseThrow(
+                              () ->
+                                      new CommitFailedException(
+                                              "Concurrent modification on Principal '%s'; retry later", principalName));
+      newPrincipalEntity = updatedNewPrincipalEntity;
+    }
+    return new PrincipalWithCredentials(
+            newPrincipalEntity.asPrincipal(),
+            new PrincipalWithCredentialsCredentials(
+                    newSecrets.getPrincipalClientId(), newSecrets.getMainSecret()));
+  }
+
   public @Nonnull PrincipalWithCredentials rotateCredentials(String principalName) {
     PolarisAuthorizableOperation op = PolarisAuthorizableOperation.ROTATE_CREDENTIALS;
     authorizeBasicTopLevelEntityOperationOrThrow(op, principalName, PolarisEntityType.PRINCIPAL);
@@ -1127,11 +1207,12 @@ public class PolarisAdminService {
     return rotateOrResetCredentialsHelper(principalName, false);
   }
 
-  public @Nonnull PrincipalWithCredentials resetCredentials(String principalName) {
+  public @Nonnull PrincipalWithCredentials resetCredentials(String principalName, ResetPrincipalRequest resetPrincipalRequest) {
     PolarisAuthorizableOperation op = PolarisAuthorizableOperation.RESET_CREDENTIALS;
     authorizeBasicTopLevelEntityOperationOrThrow(op, principalName, PolarisEntityType.PRINCIPAL);
-
-    return rotateOrResetCredentialsHelper(principalName, true);
+    var customClientId = resetPrincipalRequest.getClientId();
+    var customClientSecret = resetPrincipalRequest.getClientSecret();
+    return resetCredentialsHelper(principalName, true, customClientId, customClientSecret);
   }
 
   public List<PolarisEntity> listPrincipals() {
--- a/spec/polaris-management-service.yml
+++ b/spec/polaris-management-service.yml
@@ -199,6 +199,42 @@ paths:
         404:
           description: "The catalog or principal does not exist"
 
+  /principals/{principalName}/reset:
+    parameters:
+      - name: principalName
+        in: path
+        description: The principal's name
+        required: true
+        schema:
+          type: string
+          minLength: 1
+          maxLength: 256
+          pattern: '^(?!\s*[s|S][y|Y][s|S][t|T][e|E][m|M]\$).*$'
+    post:
+      operationId: resetCredentials
+      description: >
+        Reset a principal's credentials to a new set. By default, the system generates random credentials unless
+        explicitly allowed to accept user-provided credentials via configuration.  
+        This API is *not* idempotent and will return the newly created credentials.
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ResetPrincipalRequest'
+      responses:
+        200:
+          description: The principal details along with the newly reset credentials
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/PrincipalWithCredentials"
+        403:
+          description: The caller does not have permission to reset credentials
+        404:
+          description: The principal does not exist
+
+
     put:
       operationId: updatePrincipal
       description: Update an existing principal
@@ -1226,6 +1262,22 @@ components:
       required:
         - currentEntityVersion
         - properties
+
+    ResetPrincipalRequest:
+      type: object
+      properties:
+        clientId:
+          type: string
+          description: Optional client ID to set for the principal.
+          minLength: 16
+          maxLength: 16
+          pattern: '^[0-9a-f]{16}$'
+        clientSecret:
+          type: string
+          description: Optional client secret to set for the principal.
+          minLength: 32
+          maxLength: 32
+          pattern: '^[0-9a-f]{32}$'
 
     PrincipalRoles:
       type: object
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/TransactionWorkspaceMetaStoreManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/TransactionWorkspaceMetaStoreManager.java
@@ -165,6 +165,21 @@ public class TransactionWorkspaceMetaStoreManager implements PolarisMetaStoreMan
   }
 
   @Override
+  public PrincipalSecretsResult resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret) {
+    callCtx
+            .getDiagServices()
+            .fail("illegal_method_in_transaction_workspace", "resetPrincipalSecrets");
+    return null;
+  }
+
+  @Override
   public CreateCatalogResult createCatalog(
       @Nonnull PolarisCallContext callCtx,
       @Nonnull PolarisBaseEntity catalog,
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapTransactionalPersistenceImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapTransactionalPersistenceImpl.java
@@ -528,6 +528,22 @@ public class TreeMapTransactionalPersistenceImpl extends AbstractTransactionalPe
     return principalSecrets;
   }
 
+  @Override
+  public @Nonnull PolarisPrincipalSecrets resetPrincipalSecrets(
+          @Nonnull PolarisCallContext callCtx,
+          @Nonnull String clientId,
+          long principalId,
+          boolean reset,
+          @Nonnull String oldSecretHash,
+          String customClientId,
+          String customClientSecret) {
+    // get metastore we should be using
+    callCtx
+            .getDiagServices()
+            .fail("illegal_method_in_transaction_workspace", "resetPrincipalSecrets");
+    return null;
+  }
+
   /** {@inheritDoc} */
   @Override
   public void deletePrincipalSecretsInCurrentTxn(
--- a/service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
+++ b/service/common/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
@@ -51,6 +51,7 @@ import org.apache.polaris.core.admin.model.PrincipalRole;
 import org.apache.polaris.core.admin.model.PrincipalRoles;
 import org.apache.polaris.core.admin.model.PrincipalWithCredentials;
 import org.apache.polaris.core.admin.model.Principals;
+import org.apache.polaris.core.admin.model.ResetPrincipalRequest;
 import org.apache.polaris.core.admin.model.RevokeGrantRequest;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
 import org.apache.polaris.core.admin.model.TableGrant;
@@ -142,7 +143,7 @@ public class PolarisServiceImpl
   /** From PolarisCatalogsApiService */
   @Override
   public Response createCatalog(
-      CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
+          CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
     PolarisAdminService adminService = newAdminService(realmContext, securityContext);
     Catalog catalog = request.getCatalog();
     validateStorageConfig(catalog.getStorageConfigInfo());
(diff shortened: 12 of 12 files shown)


spec/polaris-management-service.yml
```diff
@@ -199,6 +199,42 @@ paths:
         404:
           description: "The catalog or principal does not exist"
 
+  /principals/{principalName}/reset:
```
dimas-b: Is it placed correctly in this file? It seems to cut in-between `get` and `put` of  `/principals/{principalName)` :thinking:

spec/polaris-management-service.yml
```diff
@@ -1227,6 +1263,22 @@ components:
...
         - properties
 
+    ResetPrincipalRequest:
+      type: object
+      properties:
+        clientId:
+          type: string
+          description: Optional client ID to set for the principal.
+          minLength: 16
+          maxLength: 16
```
dimas-b: I'm not sure it is a good idea to define restrictions on `clientId` (and `clientSecret`) in the API definition. Those are largely implementation concerns.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -240,12 +244,23 @@ private void authorizeBasicTopLevelEntityOperationOrThrow(
     PolarisResolvedPathWrapper topLevelEntityWrapper =
         resolutionManifest.getResolvedTopLevelEntity(topLevelEntityName, entityType);
 
+    if (op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS)) {
+      boolean isRoot = getRootPrincipalName().equals(authenticatedPrincipal.getPrincipalEntity().getName());
+      if (!isRoot) {
+        throw new ForbiddenException("Only root principal can reset credentials");
```
dimas-b: I believe it is preferable to delegate this kind of check to `PolarisAuthorizer`.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -240,12 +244,23 @@ private void authorizeBasicTopLevelEntityOperationOrThrow(
...
+              .log("Root principal allowed to reset credentials");
+      return;
+    }
+
     // TODO: If we do add more "self" privilege operations for PRINCIPAL targets this should
     // be extracted into an EnumSet and/or pushed down into PolarisAuthorizer.
     if (topLevelEntityWrapper.getResolvedLeafEntity().getEntity().getId()
             == authenticatedPrincipal.getPrincipalEntity().getId()
-        && (op.equals(PolarisAuthorizableOperation.ROTATE_CREDENTIALS)
-            || op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS))) {
```
dimas-b: This change LGTM as `RESET_CREDENTIALS` has no actual use case in code before this PR.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -774,6 +774,69 @@ public PolarisPrincipalSecrets generateNewPrincipalSecrets(
...
   }
 
+  @Nullable
+  @Override
+  public PolarisPrincipalSecrets resetPrincipalSecrets(
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      boolean reset,
+      @Nonnull String oldSecretHash,
```
dimas-b: `oldSecretHash` is not used?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
```diff
@@ -147,6 +147,19 @@ public void rotateSecrets(String newSecondaryHash) {
     this.mainSecretHash = hashSecret(mainSecret);
   }
 
+  /** Reset the main secrets */
+  public void resetSecrets(String customClientSecret) {
+    this.mainSecret = customClientSecret;
+    this.mainSecretHash = hashSecret(mainSecret);
+
+    this.secondarySecret = null;
+    this.secondarySecretHash = hashSecret(mainSecret);
```
dimas-b: `this.secondarySecretHash` becomes equal to `this.mainSecretHash`?.. why?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
```diff
@@ -147,6 +147,19 @@ public void rotateSecrets(String newSecondaryHash) {
     this.mainSecretHash = hashSecret(mainSecret);
   }
 
+  /** Reset the main secrets */
+  public void resetSecrets(String customClientSecret) {
+    this.mainSecret = customClientSecret;
```
dimas-b: why store `customClientSecret` (plain text) even in memory?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -774,6 +774,69 @@ public PolarisPrincipalSecrets generateNewPrincipalSecrets(
...
+        .getDiagServices()
+        .check(
+            principalId == principalSecrets.getPrincipalId(),
+            "principal_id_mismatch",
+            "expectedId={} id={}",
+            principalId,
+            principalSecrets.getPrincipalId());
+
+    principalSecrets.setPrincipalClientId(customClientId);
+    principalSecrets.resetSecrets(customClientSecret);
```
dimas-b: I believe it is preferable to keep the `principalSecrets` object immutable and create another instance with the new credentials.

dimas-b:
@fivetran-arunsuri : please check CI errors.

spec/polaris-management-service.yml
fivetran-arunsuri: changed this

spec/polaris-management-service.yml
fivetran-arunsuri: Thanks for the feedback — I understand your point about keeping implementation concerns out of the API definition.
The reason I added minLength:  and maxLength: was to make the expected format of clientId and clientSecret clear to API consumers, since these credentials are generated and validated with a fixed length on our side.

That said, if we prefer to enforce this strictly at the implementation level, I’m happy to remove the constraints from the API definition and document the expectations separately. Let me know what you think.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
fivetran-arunsuri: done

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: removed

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
fivetran-arunsuri: Should be generate random creds for secondary hash? We are not using secondary secrets so wanted to keep it the same. let me know your thoughts about it?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
fivetran-arunsuri: Just to return the values as the api response, I think

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: Thanks for the suggestion — keeping principalSecrets immutable does make sense in general.
However, we already have a rotateCredentials operation that modifies the existing secret in place, so the pattern of mutating the current instance is already established in this case.
Let me know if my thinking about the same is wrong?

fivetran-arunsuri:
> Thanks for your contribution, @fivetran-arunsuri !
> 
> As far as I understand, principal secrets rotation is implemented only for JDBC persistence... I think EclipseLink (being deprecated) is ok to exclude from this feature, but the tree map (in-memory) persistence should probably have an implementation too (to support full API with in-memory servers).
> 
> Could you also add integration tests exercising both the happy path and authorization errors?

@dimas-b I’ve added the in-memory TreeMap implementation as well as the corresponding integration tests. However, I’m currently having trouble running the integration tests locally. I couldn’t find much detail around this in the README either.

Could you please guide me on how to set them up or point me to any relevant documentation?

The error I am seeing is as follows:
`org.junit.jupiter.api.extension.ParameterResolutionException: Failed to resolve parameter [org.apache.polaris.service.it.env.PolarisApiEndpoints arg0] in method [public static void org.apache.polaris.service.it.test.PolarisManagementServiceIntegrationTest.setup(org.apache.polaris.service.it.env.PolarisApiEndpoints,org.apache.polaris.service.it.env.ClientCredentials)]

	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	Suppressed: java.lang.NullPointerException: Cannot invoke "org.apache.polaris.service.it.env.PolarisClient.close()" because "org.apache.polaris.service.it.test.PolarisManagementServiceIntegrationTest.client" is null
		at org.apache.polaris.service.it.test.PolarisManagementServiceIntegrationTest.close(PolarisManagementServiceIntegrationTest.java:127)
		at java.base/java.lang.reflect.Method.invoke(Method.java:580)
		at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
		at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1116)
		... 1 more
Caused by: java.lang.ExceptionInInitializerError
	at org.apache.polaris.service.it.ext.PolarisIntegrationTestExtension.lambda$env$0(PolarisIntegrationTestExtension.java:72)
	at org.apache.polaris.service.it.ext.PolarisIntegrationTestExtension.env(PolarisIntegrationTestExtension.java:71)
	at org.apache.polaris.service.it.ext.PolarisIntegrationTestExtension.resolveParameter(PolarisIntegrationTestExtension.java:56)
	... 1 more
Caused by: java.lang.IllegalStateException: PolarisServerManager not found
	at org.apache.polaris.service.it.ext.PolarisServerManagerLoader.lambda$static$0(PolarisServerManagerLoader.java:28)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.apache.polaris.service.it.ext.PolarisServerManagerLoader.<clinit>(PolarisServerManagerLoader.java:28)
	... 4 more`

fivetran-arunsuri:
> * #### Helm testsExpected — Waiting for status to be reported

@dimas-b Also , I don't have access to run the ci checks by myself

fivetran-arunsuri:
@dimas-b 've addressed most of the comments on the PR and would appreciate your suggestions on a few remaining points. Looking forward to your review and hoping to get this merged soon. Thanks!

fivetran-arunsuri:
> @dimas-b 've addressed most of the comments on the PR and would appreciate your suggestions on a few remaining points. Looking forward to your review and hoping to get this merged soon. Thanks!

@dimas-b Following up on this

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: While that is true, I still think it is preferable to keep credential objects immutable. I do not insist on fixing prior use cases in this PR, but I'd prefer to avoid adding new code that mutates credentials in-place :)

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
dimas-b: ah, right :facepalm:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: TBH, my real worry is not object mutation per se, but the addition the `resetSecrets` method, which make the impression that mutating `PolarisPrincipalSecrets` in-place is the accepted approach going forward :thinking:

dimas-b:
> Caused by: java.lang.IllegalStateException: PolarisServerManager not found

This looks like a class path issue to me. I'd _guess_ you attempted to run the test(s) via IDE's own unit test runner. I'd suggest to use the gradle-based test runner. The overhead is really small, but the classpath is certain to be set according to module dependencies.

Side note: CI has some failures too.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
dimas-b: I'd say we put the old "main" hash into `secondarySecretHash` and set `mainSecretHash` according to the new secret.

`secondarySecretHash` is used in `matchesSecret()`.

dimas-b: Actually, for the reset flow, it probably makes sense to drop all old secret hashes and only keep the new hash as `mainSecretHash`. 

Setting `secondarySecretHash` to `null` in this case makes sense to me.

spec/polaris-management-service.yml
dimas-b: Polaris allows for downstream projects to inject/override many things, including `PolarisSecretsManager`, but the REST API remains the same... so I do not think setting these limits for all possible implementations is justified.

dimas-b: From my POV adding an API doc statement that inputs for these fields must come from a previous Polaris client ID generated by the same service, should be sufficient.

spec/polaris-management-service.yml
```diff
@@ -261,6 +262,41 @@ paths:
...
+        description: The principal's name
+        required: true
+        schema:
+          type: string
+          minLength: 1
+          maxLength: 256
+          pattern: '^(?!\s*[s|S][y|Y][s|S][t|T][e|E][m|M]\$).*$'
+    post:
+      operationId: resetCredentials
+      description: >
```
dimas-b: nit: maybe `>-` (no new line at the end)?

persistence/eclipselink/src/main/java/org/apache/polaris/extension/persistence/impl/eclipselink/PolarisEclipseLinkMetaStoreSessionImpl.java
```diff
@@ -785,4 +785,16 @@ Optional<Optional<String>> hasOverlappingSiblings(
...
+  @Nullable
+  @Override
+  public PolarisPrincipalSecrets resetPrincipalSecrets(
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      boolean reset,
+      String customClientId,
+      String customClientSecret) {
+    return null;
```
dimas-b: nit: maybe `throw new UnsupportedOperationException("...")`?

fivetran-arunsuri:
I have fixed the CI failures

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecretsManager.java
```diff
@@ -54,4 +54,14 @@ PrincipalSecretsResult rotatePrincipalSecrets(
       long principalId,
       boolean reset,
       @Nonnull String oldSecretHash);
+
+  @Nonnull
+  PrincipalSecretsResult resetPrincipalSecrets(
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      boolean reset,
```
dimas-b: The value of `reset` appears to be always `true` :thinking:

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
```diff
@@ -147,6 +147,19 @@ public void rotateSecrets(String newSecondaryHash) {
...
+    this.mainSecretHash = hashSecret(customClientSecret);
+
+    this.secondarySecret = null;
+    this.secondarySecretHash = hashSecret(customClientSecret);
+  }
+
+  public void setPrincipalClientId(String customClientId) {
+    this.principalClientId = customClientId;
```
dimas-b: It looks like the `customClientId` may come here from `ResetPrincipalRequest`, where it is options, so it may be `null`... Is that possible? If so, it does not look like a valid call path to me, since client ID ought to be kept the same if not provided by the API caller, right?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -283,12 +285,20 @@ private void authorizeBasicTopLevelEntityOperationOrThrow(
     PolarisResolvedPathWrapper topLevelEntityWrapper =
         resolutionManifest.getResolvedTopLevelEntity(topLevelEntityName, entityType);
 
+    if (op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS)) {
+      authorizer.authorizeOrThrow(authenticatedPrincipal);
```
dimas-b: Why not let the Authroizer enforce `RESET_CREDENTIALS` via the old `authorizeOrThrow(..., PolarisAuthorizableOperation authzOp, ...)` method?

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
```diff
@@ -873,6 +873,53 @@ public void testCreatePrincipalAndRotateCredentials() {
...
+    assertThat(resetCreds.getCredentials().getClientId()).isEqualTo("f174b76a7e1a99e2");
+    assertThat(resetCreds.getCredentials().getClientSecret())
+        .isEqualTo("27029d236abc08e204922b0a07031bc2");
+
+    // Principal itself tries to reset with custom creds → should fail (403 Forbidden)
+    String principalToken = client.obtainToken(resetCreds);
+    customBody =
+        Map.of(
+            "clientId", "f174b76a7e1a99e3",
+            "clientSecret", "27029d236abc08e204922b0a07031bc3");
```
dimas-b: nit: maybe use something completely different from lines 890-891 for ease of reading?

fivetran-arunsuri: The creds are diff though, may look similar 🤔

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
fivetran-arunsuri: May be other params are not required for this particular case

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
fivetran-arunsuri: Let me check if that's possible

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: The `Authroizer` interface is meant to handle all authZ cases via the existing method parameters (operation plus related entities). I do not see a need for adding a new method for this use case. Existing methods appear to be sufficient.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -599,6 +604,16 @@ public void authorizeOrThrow(
           authenticatedPrincipal.getActivatedPrincipalRoleNames(),
           activatedEntities.stream().map(PolarisEntityCore::getName).collect(Collectors.toSet()),
           authzOp);
+    } else if (authzOp == PolarisAuthorizableOperation.RESET_CREDENTIALS) {
+      if (!isRoot) {
+        throw new ForbiddenException(
+            "Only %s principal can reset credentials",
```
dimas-b: Why do we call `isAuthorized(...)` above for `RESET_CREDENTIALS`? The auth logic is very specific here and clearly does not rely on role/privilege assignments (which is what `isAuthorized()` checks) :thinking:

spec/polaris-management-service.yml
fivetran-arunsuri: Made changes to handle validation in Impl

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: made changes, Can you please take a look

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecretsManager.java
fivetran-arunsuri: removed

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrincipalSecrets.java
fivetran-arunsuri: Earlier my reset logic was that if clientId and clientSecret were not passed in the request, we should prevent resetting. I’ve now tweaked it so that: if both clientId and clientSecret are provided and valid, we reset and return the updated principal; if not, we rotate only the secret while preserving the existing clientId. Let me know if you have any concerns.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
fivetran-arunsuri: done

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
fivetran-arunsuri: it was mostly WIP for testing out, I have moved it above 💯

fivetran-arunsuri:
@dimas-b Sorry for the delay, I was tied up with some high-priority incidents. Could you please take a look? I’m planning to merge this week

fivetran-arunsuri:
@dimas-b Please let me know if more changes are needed in this flow?

fivetran-arunsuri:
Bumping up for a re review

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -590,6 +595,15 @@ public void authorizeOrThrow(
       throw new ForbiddenException(
           "Principal '%s' is not authorized for op %s due to PRINCIPAL_CREDENTIAL_ROTATION_REQUIRED_STATE",
           authenticatedPrincipal.getName(), authzOp);
+    } else if (authzOp == PolarisAuthorizableOperation.RESET_CREDENTIALS) {
+      if (!isRoot) {
+        throw new ForbiddenException(
+            "Only Root principal(service-admin) can reset credentials for op %s", authzOp);
```
dimas-b: nit: `reset credentials` is redundant (implied by `authzOp`). I'd propose: `Only the root principal can perform %`

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -167,6 +168,18 @@ public Response createCatalog(
     return Response.status(Response.Status.CREATED).build();
   }
 
+  private void validateResetClientId(String clientId) {
```
dimas-b: suggestion: `validateClientId` - even though we do not call it in other cases, the logic in this method applies to all client IDs (same for `validateClientSecret`)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -773,6 +773,73 @@ public PolarisPrincipalSecrets generateNewPrincipalSecrets(
...
+      long principalId,
+      String customClientId,
+      String customClientSecret,
+      boolean customReset) {
+    PolarisPrincipalSecrets principalSecrets = loadPrincipalSecrets(callCtx, clientId);
+
+    // should be found
+    callCtx
+        .getDiagServices()
+        .checkNotNull(
```
dimas-b: This will produce a NPE, which will map to 500 at the REST API level, if I'm not mistaken, but the right response in this case would be 404... WDYT?

dimas-b: By the way, do we have a test for 404 reponses?

polaris-core/src/main/java/org/apache/polaris/core/persistence/IntegrationPersistence.java
```diff
@@ -81,6 +81,26 @@ PolarisPrincipalSecrets rotatePrincipalSecrets(
...
+   * @param customReset
+   */
+  @Nullable
+  PolarisPrincipalSecrets resetPrincipalSecrets(
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      String customClientId,
+      String customClientSecret,
+      boolean customReset);
```
dimas-b: The value for `customReset` is always derived as `customClientId != null && customClientSecret != null`... Is the parameter really required?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -773,6 +773,73 @@ public PolarisPrincipalSecrets generateNewPrincipalSecrets(
...
+            "expectedId={} id={}",
+            principalId,
+            principalSecrets.getPrincipalId());
+
+    if (customReset) {
+      principalSecrets =
+          new PolarisPrincipalSecrets(
+              principalSecrets.getPrincipalId(), customClientId, customClientSecret, null);
+    } else {
+      principalSecrets.rotateSecrets(principalSecrets.getMainSecretHash());
```
dimas-b: Rotating credentials when fresh client ID and secret are not provided in the "reset" API looks odd to me. I'd prefer to treat this case as a user error (400 response).

More specifically, if `customClientId` is provided -  update it. If `customClientSecret` is provided - update it. If neither is provided - error. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -893,6 +893,50 @@ private void revokeGrantRecord(
...
+    EntityResult loadEntityResult =
+        loadEntity(
+            callCtx, PolarisEntityConstants.getNullId(), principalId, PolarisEntityType.PRINCIPAL);
+    if (loadEntityResult.getReturnStatus() != BaseResult.ReturnStatus.SUCCESS) {
+      return new PrincipalSecretsResult(BaseResult.ReturnStatus.ENTITY_NOT_FOUND, null);
+    }
+
+    PolarisBaseEntity principal = loadEntityResult.getEntity();
+    Map<String, String> internalProps =
+        PolarisObjectMapperUtil.deserializeProperties(
```
dimas-b: Please use `principal.getInternalPropertiesAsMap()`. Cf. #2347.

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -893,6 +893,50 @@ private void revokeGrantRecord(
...
+
+    PolarisBaseEntity.Builder principalBuilder = new PolarisBaseEntity.Builder(principal);
+    principalBuilder.internalProperties(PolarisObjectMapperUtil.serializeProperties(internalProps));
+    // To avoid incrementing entity version twice
+    if (customReset) {
+      principalBuilder.entityVersion(principal.getEntityVersion());
+    } else {
+      principalBuilder.entityVersion(principal.getEntityVersion() + 1);
+    }
+    ms.writeEntity(callCtx, principalBuilder.build(), true, principal);
```
dimas-b: Why is this write necessary? I do not see any change to the `principal` entity :thinking:

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -893,6 +893,50 @@ private void revokeGrantRecord(
         : new PrincipalSecretsResult(secrets);
   }
 
+  @Override
+  public @Nonnull PrincipalSecretsResult resetPrincipalSecrets(
```
dimas-b, line 925: This method looks generic enough... Would it be possible to implement it in `BaseMetaStoreManager`?

spec/polaris-management-service.yml
```diff
@@ -199,6 +199,7 @@ paths:
         404:
           description: "The catalog or principal does not exist"
 
+
```
dimas-b: nit: spurious new line?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1135,18 +1136,83 @@ public void deletePrincipal(String name) {
...
+            .getPrincipalSecrets();
+    if (currentSecrets == null) {
+      throw new IllegalArgumentException(
+          String.format("Failed to load current secrets for principal '%s'", principalName));
+    }
+    PolarisPrincipalSecrets newSecrets =
+        metaStoreManager
+            .resetPrincipalSecrets(
+                getCurrentPolarisContext(),
+                currentPrincipalEntity.getClientId(),
```
dimas-b: Could you add a test that `currentPrincipalEntity.getClientId()` is properly updated by the reset operation?

dimas-b (commented):
Sorry about the delay. I have some more comments, though.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: We have implemented the logic based 2 things:
1. Custom Credential Reset:
When a user provides a custom clientId and clientSecret, the system fully resets the principal credentials with the new values with validation. This behaves similarly to a registerPrincipal flow.

2. Random Credential Generation (No Custom Input):
When no custom credentials are provided, the system generates random credentials while retaining the existing clientId. This ensures backward compatibility and resolves [Polaris issue #624](https://github.com/apache/polaris/issues/624)

fivetran-arunsuri: Having said this we could even make the suggested change as well, mentioned here: 

> `More specifically, if customClientId is provided - update it. If customClientSecret is provided - update it. If neither is provided - error. `

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
fivetran-arunsuri: this flow should be covered as part of integration tests

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: This flow should only be triggered if the principal exists, correct me If I am wrong. In case principal does not exist it should throw 404 but we are handling it here in PolarisAdminService in case current principal is not present 
image

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
fivetran-arunsuri: I have followed the placing of rotateCredentials method, So Secret management methods are present mostly in PolarisSecretsManager and their impl in respective classes. Correct me if I am wrong

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
fivetran-arunsuri: version update, moved inside the else block

fivetran-arunsuri:
@dimas-b I’ve addressed most comments. Could you clarify the remaining points so we can ensure everything is ready for merge

dimas-b:
@fivetran-arunsuri : please resolve merge conflicts, otherwise CI will not run :shrug:

spec/polaris-management-service.yml
```diff
@@ -1248,6 +1283,20 @@ components:
...
+        clientId:
+          type: string
+          description: >
+            Optional client ID to set for the principal.
+            Must be a valid client ID previously generated by this service.
+        clientSecret:
+          type: string
+          description: >
+            Optional client secret to set for the principal.
+            Must match the secret issued for the given client ID.
```
dimas-b: note sure about the "must match" part... How about "Polaris service implementations may impose extra requirements on what is accepted as a secret (special chars, length, etc.)".

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Good point... still, if the principal is found, but do not have credentials in Polaris (i.e. it's a "federated" principal), returning 500 is not nice in this case. 

Unfortunately, I do not see well-defined existing cases for reporting nice errors in cases like this. I'm ok with the current impl. but we'll probably have to revisit it at some point.

fivetran-arunsuri:
Fixed merge conflicts

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: I'm fine with generating and returning random credentials from the reset API. However, the API is still a bit confusing because in `ResetPrincipalRequest` either `clientId` and `clientSecret` may be set without the other, but in code (line 806) execution changes only when both are set. In effect, invoking `resetCredentials` only with `clientSecret` still generates a random secret, if I'm not mistaken.

Secondary point: calling `principalSecrets.rotateSecrets()` on the "reset" call path is still awkward since credential "rotation" may have very specific logic in `PolarisPrincipalSecrets`.

I propose to add a new method `PolarisPrincipalSecrets.resetSecrets(String clientId, String secret)` returning a new `PolarisPrincipalSecrets`. The handling of `null` values in `clientId` and `secret` will be inside that `resetSecrets()` method. It will keep `this` immutable and return a new `PolarisPrincipalSecrets` object possibly reusing values from the old one if they did not get reset. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: `testCreatePrincipalAndResetCredentialsWithCustomValues()` only validates that the client ID in _credentials_ got reset properly, but it does not validate that the client ID in the principal is in sync with that... or did I miss it? :thinking:

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -892,6 +892,48 @@ private void revokeGrantRecord(
...
+    Map<String, String> internalProps = principal.getInternalPropertiesAsMap();
+
+    boolean customReset = customClientId != null && customClientSecret != null;
+    PolarisPrincipalSecrets secrets =
+        ((IntegrationPersistence) ms)
+            .resetPrincipalSecrets(
+                callCtx, clientId, principalId, customClientId, customClientSecret);
+
+    PolarisBaseEntity.Builder principalBuilder = new PolarisBaseEntity.Builder(principal);
+    principalBuilder.internalProperties(PolarisObjectMapperUtil.serializeProperties(internalProps));
```
dimas-b: Maybe `principalBuilder.internalPropertiesAsMap(...)`?

TBH, I'm at a loss as to the purpose of this piece of code. I do not see any change to internal properties. Did I miss something?

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
dimas-b: fair enough

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: Made following changes to the flow:

- **Partial updates supported**: clientId and clientSecret can be set individually; null values are handled gracefully.
- **Main secret reset only**: clientSecret is replaced or generated; secondary secret remains unchanged. 
- **Principal entity sync**: clientId in the principal is updated to match the reset credentials. 
- **Immutability**: A new PolarisPrincipalSecrets object is returned; the old one remains unchanged.
- **Access control**: Only authorized users (admins) can reset credentials; principals cannot reset their own credentials

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
fivetran-arunsuri: missed to add that check to verify the clientId tied to the principal, added

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: > reusing values from the old one if they did not get reset


@dimas-b Only thing changed here as per your suggestion is kept the clientId as same if not set but regenerate random secret if not set for security reasons

fivetran-arunsuri:
@dimas-b Please take a look at the current approach

dimas-b:
@dennishuo: FYI `PolarisSecretsManager` interface changes

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1088,6 +1089,69 @@ public void deletePrincipal(String name) {
...
+      PrincipalEntity updatedNewPrincipalEntity = updateBuilder.build();
+      updatedNewPrincipalEntity =
+          Optional.ofNullable(
+                  PrincipalEntity.of(
+                      PolarisEntity.of(
+                          metaStoreManager.updateEntityPropertiesIfNotChanged(
+                              getCurrentPolarisContext(), null, updatedNewPrincipalEntity))))
+              .orElseThrow(
+                  () ->
+                      new CommitFailedException(
```
dimas-b: I'd prefer `CommitConflictException` as noted in #2168

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1088,6 +1089,69 @@ public void deletePrincipal(String name) {
...
+    PrincipalEntity newPrincipalEntity = PrincipalEntity.of(newPrincipal);
+    if (customClientId != null) {
+      PrincipalEntity.Builder updateBuilder = new PrincipalEntity.Builder(newPrincipalEntity);
+      updateBuilder.setClientId(newSecrets.getPrincipalClientId());
+      PrincipalEntity updatedNewPrincipalEntity = updateBuilder.build();
+      updatedNewPrincipalEntity =
+          Optional.ofNullable(
+                  PrincipalEntity.of(
+                      PolarisEntity.of(
+                          metaStoreManager.updateEntityPropertiesIfNotChanged(
```
dimas-b, line 1097: This is risky WRT `metaStoreManager.resetPrincipalSecrets()` (line 1112). If the request supplies a new client ID and `resetPrincipalSecrets` succeeds, but `updateEntityPropertiesIfNotChanged` fails, then the principal entity's client ID will be out of sync with the `PolarisPrincipalSecrets` client ID. This will likely break all subsequent authentication and reset credentials attempts.

I'd propose to make the inner "reset" methods robust in case the credential is not found by client ID (i.e. create a new credential / hash).

Also, I'd update the entity first as it is more likely to hit conflicts than the credential reset itself.

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -892,6 +892,45 @@ private void revokeGrantRecord(
...
+                callCtx, clientId, principalId, customClientId, customClientSecret);
+
+    PolarisBaseEntity.Builder principalBuilder = new PolarisBaseEntity.Builder(principal);
+    var newEntityVersion =
+        (customClientId != null) ? principal.getEntityVersion() : principal.getEntityVersion() + 1;
+
+    principalBuilder.entityVersion(newEntityVersion);
+    // Only write if version changed
+    if (customClientId == null) {
+      ms.writeEntity(callCtx, principalBuilder.build(), true, principal);
```
dimas-b: I'm sorry, but I still do not understand why this write is necessary. What are we changing in this `PolarisBaseEntity` aside from its version?

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -892,6 +892,45 @@ private void revokeGrantRecord(
         : new PrincipalSecretsResult(secrets);
   }
 
+  @Override
+  public @Nonnull PrincipalSecretsResult resetPrincipalSecrets(
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      @Nonnull String oldSecretHash,
```
dimas-b: `oldSecretHash` appears to be unused (in all implementations).

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: It might be worth doing the reset in 3 phases:
1) get and delete old secrets
2) change the client ID in the entity
3) create new secrets

This way, failures at any stage are recoverable by re-trying the reset operation, I believe... at least the entity and secrets can be re-aligned (perhaps with ID or secrete being re-generated).

WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: thx, the new `PolarisPrincipalSecrets` code LGTM. Resolving this thread.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Thx! The new `PolarisPrincipalSecrets.resetSecrets()` and caller code LGTM. Resolving this thread.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
fivetran-arunsuri: Thanks for the catch @dimas-b 
The other approach could be 
- Update the entity first (because conflicts are more common there).
- Only then reset secrets.
- On failure during secrets reset, roll back entity’s clientId to the old one

but again a down side what if rollback to entity's clientId fails, again it will be inconsistent state
This will also give it some idempotency. Your suggested 3-phase flow is more robust and avoids this risk. I’ll go ahead and make the change — thanks!

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
fivetran-arunsuri: is there a lighter approach for version update?

dimas-b: What is it that we're changing in the entity with this write?

fivetran-arunsuri: That said, you raise a fair point — if the only intent is to reflect the version bump, maybe there’s a lighter approach than writing back the full PolarisBaseEntity

fivetran-arunsuri:
@dimas-b I made the suggested changes but had to revert some of our earlier ones; functionality is working fine based on my testing. Could you take a look and let me know if you see any areas that still need refactoring?

fivetran-arunsuri:
Following up on the above

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
dimas-b: I wonder why we need a version change without any other property changes? :thinking:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -773,6 +773,49 @@ public PolarisPrincipalSecrets generateNewPrincipalSecrets(
...
+      principalSecrets = new PolarisPrincipalSecrets(principalId, clientId, customClientSecret);
+    } else {
+      principalSecrets =
+          new PolarisPrincipalSecrets(principalId, customClientId, customClientSecret);
+    }
+    try {
+      ModelPrincipalAuthenticationData modelPrincipalAuthenticationData =
+          ModelPrincipalAuthenticationData.fromPrincipalAuthenticationData(principalSecrets);
+      datasourceOperations.executeUpdate(
+          QueryGenerator.generateInsertQuery(
```
dimas-b, line 789: While INSERT works with current higher-level code (that calls this method). The impl. does not appear to be aligned with the javadoc of `IntegrationPersistence.resetPrincipalSecrets()`. According to the javadoc, this method may be called even when `ModelPrincipalAuthenticationData` still exists for the given client ID (which is part of the primary key).

WDYT about renaming to `storePrincipalSecrets` and updating javadoc to indicate that credentials for the given client ID are not expected to exist?

With that, we should also be able to have only one client ID method parameter.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -773,6 +773,49 @@ public PolarisPrincipalSecrets generateNewPrincipalSecrets(
...
+  public PolarisPrincipalSecrets resetPrincipalSecrets(
+      @Nonnull PolarisCallContext callCtx,
+      @Nonnull String clientId,
+      long principalId,
+      String customClientId,
+      String customClientSecret) {
+
+    PolarisPrincipalSecrets principalSecrets;
+    if (customClientId == null) {
+      principalSecrets = new PolarisPrincipalSecrets(principalId, clientId, customClientSecret);
```
dimas-b: I believe `customClientSecret` may still be `null` here (when the REST request does not have it). Should be generate a random secret when it is not provided by the user?

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -916,6 +927,53 @@ private void bootstrapPolarisService(
...
+    // if not found, the principal must have been dropped
+    EntityResult loadEntityResult =
+        loadEntity(
+            callCtx,
+            ms,
+            PolarisEntityConstants.getNullId(),
+            principalId,
+            PolarisEntityType.PRINCIPAL.getCode());
+    if (loadEntityResult.getReturnStatus() != BaseResult.ReturnStatus.SUCCESS) {
+      return null;
```
dimas-b: The return value is annotated as `@Nonnull`

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecretsManager.java
```diff
@@ -54,4 +54,16 @@ PrincipalSecretsResult rotatePrincipalSecrets(
       long principalId,
       boolean reset,
       @Nonnull String oldSecretHash);
+
+  @Nonnull
+  PrincipalSecretsResult resetPrincipalSecrets(
```
dimas-b, line 74: Please add javadoc

dimas-b (commented):
I think we're pretty close to a mergeable state, but I've got a couple of new concerns (on new code).

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: Sorry I didn't get this, The constructor already generates a random secret when customClientSecret is null, so we don’t need to handle that separately here. Please let me know if I’m missing something

dimas-b: My mistake - I missed that change :facepalm: All good here.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: Made the change as suggested, I am not too sure about passing only 1 client id as param as they serve their purpose. clientId param is tied to the principal as old client Id(And it should be used, if not provided by user) and customClientId is user provided clientId

dimas-b: I mean that at the level of `storePrincipalSecrets()` we already know what the client ID should be. I think it should be fine to resolve `clientId` vs. `customClientId` at a higher level (before calling this method).

dimas-b: As for me, it looks like `PolarisAdminService.resetCredentialsHelper()` already can resolve `clientId` vs. `customClientId` and all subsequent "reset" calls only need one (effective) client ID. WDYT?

dimas-b:
I'm generally ok with this PR (only minor concerns remain). Added a few other people, who might be interested (just for visibility from my POV).

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
fivetran-arunsuri: Makes sense, made the change

dimas-b (approved):
LGTM :+1: Thanks for your work on this @fivetran-arunsuri !

Let's wait at least a day for other reviewers to comment (if they are interested).

dimas-b:
Good point about change log. Adding a feature flag also sound good to me.

As for adding non-root permissions, it does make sense, but I propose to handle that as a follow-up.

fivetran-arunsuri:
Thanks for the review @dimas-b @eric-maynard 

I’ll update the changelog as suggested.

On the feature flag — since this API is already restricted to root, I’m not sure a flag adds much additional protection. If you feel it does, could you clarify how you envision feature-flagging this change? It feels like privilege-based access (as you mentioned) would be the more scalable long-term solution, so that specific roles can be granted reset access instead of relying on a global on/off switch.

Happy to add a toggle if you feel strongly, but my preference would be to keep the API available and focus on evolving the privilege model as followup PR

dimas-b:
Re: feature flag - I cannot speak for @eric-maynard , so just posting my view of this:

Guarding credential reset with a feature flag for people that deploy Apache Polaris binary is probably not very valuable. So if we add it, I'd suggest defaulting to "enabled".

However, for downstream projects that provide managed Polaris service it could be valuable to disable it, so that only the "service owner" has control over credentials.

I do not insist on adding this feature flag, but since it came up during reviews, it might worth adding it proactively.

CHANGELOG.md
```diff
@@ -83,6 +83,9 @@ at locations that better optimize for object storage.
 
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
+- Added Management API endpoint to reset existing principal credentials with custom values(currently restricted to root).
+
+
```
dimas-b, line 90: nit: unnecessary new line

dimas-b (approved):

eric-maynard:
Yeah @dimas-b summarized my thoughts well. 

As a catalog admin, I might feel assured that not even service admins with access to the root principal can rotate my credentials. Plus, as I noted in my comment, eventually I'd actually like to see us extend this capability to principles other than root!

Besides this rationale, I'm also just a fan of flagging changes so that service admins can keep from exercising new code paths should some regression appear.

fivetran-arunsuri:
I added the flag (default: enabled) so managed deployments can opt out if needed, while self-hosted users see no change. Let me know if anything else is needed, otherwise we can proceed with merge.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -116,6 +116,7 @@ polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE"]
 # polaris.features."ENABLE_CATALOG_FEDERATION"=true
 polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
 polaris.features."SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES"=["OAUTH", "BEARER"]
+polaris.features."ENABLE_CREDENTIAL_RESET"=true
```
dimas-b: nit: this setting is not necessary as it matches the code-level default.

dimas-b (approved):
LGTM :+1: Thanks again, @fivetran-arunsuri !

eric-maynard (dismissed):
LGTM! Awesome to see this coming together.

At some point we should update the [synchronizer tool](https://github.com/apache/polaris-tools/tree/main/polaris-synchronizer) to take advantage of this feature during migrations.

runtime/defaults/src/main/resources/application.properties
flyrain: Agreed with @dimas-b. We don't need config a default value explicitly.

spec/polaris-management-service.yml
```diff
@@ -261,6 +261,41 @@ paths:
...
+    parameters:
+      - name: principalName
+        in: path
+        description: The principal's name
+        required: true
+        schema:
+          type: string
+          minLength: 1
+          maxLength: 256
+          pattern: '^(?!\s*[s|S][y|Y][s|S][t|T][e|E][m|M]\$).*$'
```
flyrain, line 274: Can we create a component for principal name schema? So that we don't need to copy it in multiple places?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -170,6 +171,18 @@ public Response createCatalog(
     return Response.status(Response.Status.CREATED).build();
   }
 
+  private void validateClientId(String clientId) {
+    if (!clientId.matches("^[0-9a-f]{16}$")) {
```
flyrain, line 175: Why do we only allow `[0-9a-f]`?

dimas-b: I believe, that matches current client ID values as generated by Polaris.

We can certainly change validation rules later. I do not think it is critical for this PR.

spec/polaris-management-service.yml
fivetran-arunsuri: Good point 👍 We can extract principal name into a shared component. I’d prefer to handle it in a follow-up PR to keep this one focused, unless you feel strongly it should go in here

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
fivetran-arunsuri: Yes, this matches the current client ID format generated by Polaris. We can relax/update the regex in a follow-up PR if the format changes

runtime/defaults/src/main/resources/application.properties
fivetran-arunsuri: done

fivetran-arunsuri:
@dimas-b @eric-maynard I rebased my branch, which removed your approval. Could you please take another look? If everything looks good, let’s move ahead with the merge — this feature is currently blocking our migration.

flyrain (commented):
Hi @fivetran-arunsuri , thanks a lot for working on it. Sorry for the late review. Appreciated if you can answer a few questions:
1. What's the use case of reset a client id of a principal? I understand that other systems like Keycloak support that. But Polaris doesn't have to mimic them. 
2. Can we reuse the credential rotate endpoint(`/principals/{principalName}/rotate`) instead of having a new endpoint? We can still enforce only root can rotate other principal's credentials. It's also acceptable to add an optional field to allow client to input credentials.

dimas-b:
@flyrain : FYI: https://lists.apache.org/thread/wc710p1xd25l9k9t9jplnnn8njf15lq0

dimas-b:
> Can we reuse the credential rotate endpoint(/principals/{principalName}/rotate) instead of having a new endpoint?

My reading of the current Polaris Management API spec is that the "rotate" endpoint has quite different functionality from "reset". Implementing the feature proposed in this PR via the old "reset" endpoint could technically be done (I guess), but I think it will overload the old "reset" functionality too much and make its logic complicated unnecessarily. I believe adding a new endpoint is preferable (as implemented in this PR).

spec/polaris-management-service.yml
dimas-b: Doing that in a follow-up PR sounds reasonable to me.

flyrain:
> - Enable a *safe, service-admin only way* to inject known credentials
   via the API during the transition phase with validations of course

I didn't find a use case mentioned in the mail thread to reset the client id. Correct me if I'm wrong, @fivetran-arunsuri.

There is no old `reset` endpoint. Only rotate endpoint and principal creation endpoint return back the credentials. 

> My reading of the current Polaris Management API spec is that the "rotate" endpoint has quite different functionality from "reset".

How is that quite different? For credential refresh, the `rotate` endpoint can just accept an optional input to allow manual credential refresh.

dimas-b:
> There is no old reset endpoint.

Sorry, typo on my part. I meant "old rotate".

> How is that quite different? For credential refresh, the rotate endpoint can just accept an optional input to allow manual credential refresh.

The "rotate" endpoint keeps old credentials effective (they become secondary). 

The "rotate" endpoint implies that the client ID never changes.

If we're to allow completely replacing both the client ID and all associated secrets, that's no longer "rotation", IMHO.

flyrain:
The new endpoint makes sense to me if we need update client id. Are there use cases for client id update now?

fivetran-arunsuri:
Hi @flyrain, Based on the usecase explained in the email thread and as @dimas-b explained earlier.

> Hi @fivetran-arunsuri , thanks a lot for working on it. Sorry for the late review. Appreciated if you can answer a few questions:
> 
> 1. What's the use case of reset a client id of a principal? I understand that other systems like Keycloak support that. But Polaris doesn't have to mimic them.
 
This isn’t about mimicking Keycloak or expanding Polaris into a full IdP. The primary driver is our migration path from Polaris 0.9 → 1.0. Since we’ll be running both catalog servers in parallel during the cutover, we need to reuse the same clientId / clientSecret across both instances. This ensures:

- Clients can authenticate seamlessly without coordinating a mass credential rotation.
- We can safely switch traffic between catalogs and roll back instantly if needed.
- Zero downtime cutover is possible.

Without the ability to inject known credentials, we’d either have to rotate secrets across all clients (not feasible at scale) or risk schema/hash inconsistencies when trying to copy raw tables

Also this automatically answers your second question, As Rotate API keeps the clientId as same

fivetran-arunsuri:
This change is mainly to help users migrate from 0.9 (EclipseLink) → 1.0 (JDBC) with schema changes, without risky manual DB manipulation by registering existing sets of users keeping creds as same. I believe it could be quite useful for adoption. Currently the functionality is only restricted to root user

fivetran-arunsuri:
@flyrain Pasting what we discussed in DM around the reset vs rotate:

I see your point about client_id semantics and the flexibility of reusing the rotate endpoint.
That said, I’d still lean towards keeping a separate reset API, mainly for clarity of intent from a client perspective:

- rotate suggests replacing credentials while keeping the same client identity, whereas reset communicates a stronger action (e.g., provisioning a completely fresh set).

- users often expect reset when dealing with credentials (similar to password reset vs password rotate), so the explicit endpoint helps reduce confusion.

- Future flexibility: if Polaris later wants to evolve different behaviors (e.g., stricter validation or revoking old client IDs), having both endpoints gives us that room without introducing breaking changes.

- On top of that, the current implementation is limited to the root principal. rotate doesn’t work with the root principal (only with its own creds), so reset is the right fit here. It also provides a way to reset forgotten credentials for a particular principal without touching the existing implementation.

Finally, this direction was already discussed and agreed in the Dev email thread to support a separate endpoint, so keeping both aligns with that decision—rather than using the same API as createPrincipal by passing custom creds, as I initially proposed.

Would you be open to keeping both endpoints? Can we go ahead with the PR?

flyrain:
Thanks @fivetran-arunsuri! I'm fine with new reset endpoint for the clarity of intent from a client perspective.

PR #2198, opened 2025-07-28
Replace CommitFailedException with CommitConflictException

### LINKED ISSUE #2168: Consolidate usage of CommitFailedException and CommitConflictException

tmater:
In some cases, we were using CommitFailedException to represent commit conflicts, which returns the correct 409 response but is tied to Iceberg.

However, some of these conflicts originate from Polaris, making CommitConflictException a more appropriate and accurate choice.

This change updates those instances to improve clarity and exception handling semantics.

Resolves #2168

### DIFF
--- a/service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
+++ b/service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
@@ -51,7 +51,6 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.catalog.ViewCatalog;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.BadRequestException;
-import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.ForbiddenException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.hadoop.HadoopCatalog;
@@ -84,6 +83,7 @@ import org.apache.polaris.core.context.CallContext;
 import org.apache.polaris.core.entity.CatalogEntity;
 import org.apache.polaris.core.entity.PolarisEntitySubType;
 import org.apache.polaris.core.entity.table.IcebergTableLikeEntity;
+import org.apache.polaris.core.exceptions.CommitConflictException;
 import org.apache.polaris.core.persistence.PolarisEntityManager;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.PolarisResolvedPathWrapper;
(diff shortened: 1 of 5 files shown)


service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -997,7 +997,7 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
             callContext.getPolarisCallContext(), pendingUpdates);
     if (!result.isSuccess()) {
       // TODO: Retries and server-side cleanup on failure
-      throw new CommitFailedException(
+      throw new CommitConflictException(
           "Transaction commit failed with status: %s, extraInfo: %s",
           result.getReturnStatus(), result.getExtraInformation());
```
dimas-b, line 1002: Strictly speaking the majority of possible status codes do not fit 409 responses, from my POV. It might be preferable to handle this case by inspecting `result.getReturnStatus()` (probably in the exception mapper).

That said, such a change will break current de-facto API contracts.... I'm not sure whether there is appetite in the community for clarifying this.

spec/generated/bundled-polaris-catalog-service.yaml
```diff
@@ -635,7 +635,7 @@ paths:
                 TableToUpdateDoesNotExist:
                   $ref: '#/components/examples/NoSuchTableError'
         '409':
-          description: Conflict - CommitFailedException, one or more requirements failed. The client may retry.
+          description: Conflict - CommitConflictException, one or more requirements failed. The client may retry.
```
dimas-b: I'd prefer to avoid mentioning rather internal Polaris exception type names in the API doc.

dimas-b (commented):
Thanks for your contribution, @tmater ! The changes LGTM overall.

service/common/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
tmater: Yeah, I agree — I reverted the change and added a note to also keep an eye on the exceptions.

spec/generated/bundled-polaris-catalog-service.yaml
tmater: Good point, removed the exception and rephrased the description a bit.

dimas-b (dismissed):
LGTM :+1: Thanks, @tmater !

adutra (approved):

spec/generated/bundled-polaris-catalog-service.yaml
```diff
@@ -635,7 +635,7 @@ paths:
                 TableToUpdateDoesNotExist:
                   $ref: '#/components/examples/NoSuchTableError'
         '409':
-          description: Conflict - CommitFailedException, one or more requirements failed. The client may retry.
+          description: Conflict - Transaction commit failed due to concurrent modifications. The client may retry.
```
eric-maynard: This file is generated; the change should go into the appropriate spec and then this file needs to be regenerated.

dimas-b: Good point about spec change.

However, the code fix in this PR does not require spec changes. I believe the client-facing behaviour is not changed by this PR.

dimas-b: @tmater : please rollback changes to this file.

tmater: Makes sense, reverted it.

PR #2206, opened 2025-07-29
Add retry logic to handle concurrent updates for namespace properties

### LINKED ISSUE #2183: Add retry logic to handle concurrent updates for namespace properties

See discussions in https://github.com/apache/polaris/pull/1989 for context.

dev email thread https://lists.apache.org/thread/gp4f84ld08z1khrd09klzdzflv7l7fgh

_No response_

_No response_

### Additional context

_No response_

fabio-rizzo-01:
Retry logic only added for update of namespace properties. Added the new methods in PolarisMetaStoreManger and retry logic implemented only for TransactionMetastoreManagerImpl.

Fixes #2183

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
@@ -23,6 +23,8 @@ import jakarta.annotation.Nullable;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import javax.validation.constraints.NotNull;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.auth.PolarisGrantManager;
 import org.apache.polaris.core.auth.PolarisSecretsManager;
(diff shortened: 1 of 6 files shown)


fabio-rizzo-01:
@dimas-b  please have a look when you have time to see if this is going in the direction you were suggesting?

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -247,6 +251,48 @@ EntityResult updateEntityPropertiesIfNotChanged(
...
+   * @param entity entity to update, cannot be null
+   * @return the entity we updated or null if the client should retry
+   */
+  @Nonnull
+  EntityResult retryUpdateEntityProperties(
+      @Nonnull PolarisCallContext callCtx,
+      @Nullable List<PolarisEntityCore> catalogPath,
+      @Nonnull PolarisBaseEntity entity,
+      @NotNull Map<String, String> properties,
+      int numOfRetries);
```
dimas-b, line 273: I'm not sure adding another method to `PolarisMetaStoreManager` for `numOfRetries` is the best approach.

For most entities, updates should be based on the specific previous version of the entity, so retrying at the MetaStore level will violate the normal optimistic locking approach.

Namespaces are special only from the Iceberg REST Catalog API perspective. That API is defined such that Namespace property updates could be executed in the "last write wins" manner. However, in that case, I believe, retries should be handled in `IcebergCatalog` (or at a higher level, perhaps).

dimas-b (commented):
Thanks for working on this @fabio-rizzo-01 ! Unfortunately, I am not sure about the approach taken in the first revision on this PR... more in the following comment.

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
fabio-rizzo-01: thank you for the feedback. I was initially unsure about the appropriate level for implementing the retry logic. I considered placing it in the IcebergCatalog but ultimately decided it made more sense at the metastore level. I'll update the pull request to move the logic to the IcebergCatalog or at an higher level. If anyone has additional suggestions, please feel free to comment

PR #2210, opened 2025-07-30
Replace PolarisEntityManager with ResolutionManifestFactory

XN137:
after 95358a99221778b85905e5ea51afe90474378ce7 very little functionality was left in `PolarisEntityManager`.

by splitting out the more dedicated `ResolutionManifestFactory` we can remove `PolarisEntityManager` and `RealmEntityManagerFactory` completely it seems.


polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -63,13 +64,6 @@ public class PolarisResolutionManifest implements PolarisResolutionManifestCatal
 
   private final Map<Object, ResolverPath> passthroughPaths = new HashMap<>();
 
-  // For applicable operations, this represents the topmost root entity which services as an
-  // authorization parent for all other entities that reside at the root level, such as
-  // Catalog, Principal, and PrincipalRole.
-  // This simulated entity will be used if the actual resolver fails to resolve the rootContainer
-  // on the backend due to compatibility mismatches.
-  private ResolvedPolarisEntity simulatedResolvedRootContainerEntity = null;
```
XN137, line 71: TBD if we need a replacement for this or not, wanted to see CI results first without it

dimas-b: Removing `simulatedResolvedRootContainerEntity` looks reasonable to me.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolutionManifestFactoryImpl.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Nonnull
+  @Override
+  public PolarisResolutionManifest createResolutionManifest(
+      @Nonnull CallContext callContext,
+      @Nonnull SecurityContext securityContext,
+      @Nullable String referenceCatalogName) {
+    return new PolarisResolutionManifest(
+        callContext, resolverFactory, securityContext, referenceCatalogName);
```
XN137, line 42: do we need the `ResolutionManifestFactory` at all or should we call this constructor directly instead everywhere?
callers will still need a reference to `resolverFactory`.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
snazy: agree

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
```diff
@@ -127,6 +127,12 @@ public ResolverFactory resolverFactory(
     };
   }
 
+  @Produces
+  @ApplicationScoped
```
snazy, line 131: Do we need the separate producer? Seems sufficient to annotate the implementation.

snazy: Actually, if you make the factory-impl `@RequestScoped`, you can already omit the `CallContext` function parameter of `createResolutionManifest`

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolutionManifestFactoryImpl.java
dimas-b: In current PR the factory does look unnecessary. However, from my POV the ultimate goal of CDI-related refactorings is to get rid of `CallContext` in method _parameters_... In that case the factory may still be valuable as a request-scoped producer for on-demand (not injected) `PolarisResolutionManifest` instances.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
dimas-b: That would be great if we did not have `CallContext` in method _parameters_. However, it is still passed around explicitly in many cases. So, IMHO, making this factory `@RequestScoped` now adds risk of mismatching `CallContext` objects. I'd prefer to delay that change until we're ready to remove `CallContext` from method _parameters_ everywhere.

dimas-b (dismissed):
In its current state, this change looks like a step in the right direction with respect to code clarity and CDI "friendliness".... Pending comment resolution, of course.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/config/QuarkusProducers.java
snazy: Ah, bummer

XN137:
rebased after trivial merge conflict in `PolarisAdminService.java`

XN137:
rebased again after trivial conflict in `AbstractIcebergCatalogViewTest.java`

snazy (approved):

PR #2213, opened 2025-07-31
feat: Add `endpointInternal` to `AwsStorageConfigInfo`

dimas-b:
This API change is backward compatible with older clients and server using old storage configuration.

* The `endpointInternal` allows Polaris Servers to use a different host name (or IP address) for accessing S3 storage than clients. This is not a common use case, but may be relevant is more complex environments.

* If not set `endpointInternal` defaults to `endpoint`.

* The STS endpoint default changes to `endpointInternal`.

Contributes to #1530

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
@@ -58,6 +58,10 @@ public class AwsStorageConfigurationInfo extends PolarisStorageConfigurationInfo
   @JsonProperty(value = "endpoint")
   private @Nullable String endpoint;
 
+  /** Endpoint URI for internal Polaris calls to S3 API */
+  @JsonProperty(value = "endpointInternal")
+  private @Nullable String endpointInternal;
+
   /** Endpoint URI for STS API calls */
   @JsonProperty(value = "stsEndpoint")
   private @Nullable String stsEndpoint;
@@ -76,7 +80,8 @@ public class AwsStorageConfigurationInfo extends PolarisStorageConfigurationInfo
       @JsonProperty(value = "region", required = false) @Nullable String region,
       @JsonProperty(value = "endpoint") @Nullable String endpoint,
       @JsonProperty(value = "stsEndpoint") @Nullable String stsEndpoint,
-      @JsonProperty(value = "pathStyleAccess") @Nullable Boolean pathStyleAccess) {
+      @JsonProperty(value = "pathStyleAccess") @Nullable Boolean pathStyleAccess,
+      @JsonProperty(value = "endpointInternal") @Nullable String endpointInternal) {
     super(storageType, allowedLocations);
     this.roleARN = roleARN;
     this.externalId = externalId;
@@ -84,6 +89,7 @@ public class AwsStorageConfigurationInfo extends PolarisStorageConfigurationInfo
     this.endpoint = endpoint;
     this.stsEndpoint = stsEndpoint;
     this.pathStyleAccess = pathStyleAccess;
+    this.endpointInternal = endpointInternal;
   }
 
   public AwsStorageConfigurationInfo(
@@ -91,7 +97,7 @@ public class AwsStorageConfigurationInfo extends PolarisStorageConfigurationInfo
       @Nonnull List<String> allowedLocations,
       @Nonnull String roleARN,
       @Nullable String region) {
-    this(storageType, allowedLocations, roleARN, null, region, null, null, null);
+    this(storageType, allowedLocations, roleARN, null, region, null, null, null, null);
   }
 
   public AwsStorageConfigurationInfo(
@@ -100,7 +106,7 @@ public class AwsStorageConfigurationInfo extends PolarisStorageConfigurationInfo
       @Nonnull String roleARN,
       @Nullable String externalId,
       @Nullable String region) {
-    this(storageType, allowedLocations, roleARN, externalId, region, null, null, null);
+    this(storageType, allowedLocations, roleARN, externalId, region, null, null, null, null);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -160,6 +166,12 @@ public URI getEndpointUri() {
     return endpoint == null ? null : URI.create(endpoint);
   }
 
+  @JsonIgnore
+  @Nullable
+  public URI getInternalEndpointUri() {
```
adutra, line 171: nit: there is a minor naming inconsistency between `getInternalEndpointUri` and `endpointInternal`.

spec/polaris-management-service.yml
```diff
@@ -1054,8 +1054,15 @@ components:
               example: "https://s3.example.com:1234"
             stsEndpoint:
               type: string
-              description: endpoint for STS requests (optional). If not set, defaults to 'endpoint'.
+              description: endpoint for STS requests (optional). If not set, defaults to 'endpointInternal'.
```
adutra: Is this  for `stsEndpoint` or `endpoint` ?

snazy: I think the description needs to clarify:
* defaults to `endpointInternal`
* if `endpointInternal` is not specified, defaults to `endpoint`

spec/polaris-management-service.yml
```diff
@@ -1054,8 +1054,15 @@ components:
...
               example: "https://sts.example.com:1234"
+            endpointInternal:
+              type: string
+              description: >-
+                endpoint for S3 requests made by the Polaris Server (optional). If set, Polaris Service will use
+                this value instead of "endpoint". If not set, defaults to 'endpoint'. Iceberg REST API clients never
+                see this value.
```
snazy: Probably worth to update the description for `endpoint` as well, mentioning that there's `endpointInternal` in case the (client) needs to use a separate value than the service.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
dimas-b: I know, but `getEndpointInternalUri()` sounds awkward to me :sweat_smile: I'm open to renaming :) Let's see what other people think.

spec/polaris-management-service.yml
dimas-b: updated

spec/polaris-management-service.yml
dimas-b: STS requests are always sent by the Polaris Server. So if `stsEndpoint` is not set, the server will first default to `endpointInternal` for STS requests, then to `endpoint`. I've updated the doc comment to clarify.

snazy: Right, my comment's about telling that our users

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
snazy: the method name doesn't affect the serialized representation - so i've got no opinion here ;)

snazy: (although a 3rd non-native English speaker's opinion might be interesting ;) )

spec/polaris-management-service.yml
```diff
@@ -1050,12 +1050,23 @@ components:
               example: "us-east-2"
             endpoint:
               type: string
-              description: endpoint for S3 requests (optional)
+              description: >-
```
snazy, line 1053: Nit: I think you don't need the `>-` thingy

spec/polaris-management-service.yml
```diff
@@ -1050,12 +1050,23 @@ components:
...
+                endpoint for S3 requests (optional). Clients always see this value (if it is set). Polaris Servers
```
snazy, line 1054: Nit-pick
```suggestion
                Endpoint for S3 requests (optional). Clients always see this value (if it is set). Polaris Servers
```

spec/polaris-management-service.yml
```diff
@@ -1050,12 +1050,23 @@ components:
...
+                may be configured to use a different endpoint URI via the `endpointInternal` property.
               example: "https://s3.example.com:1234"
             stsEndpoint:
               type: string
-              description: endpoint for STS requests (optional). If not set, defaults to 'endpoint'.
+              description: >-
+                endpoint for STS requests made by the Polaris Server (optional). If not set, defaults to
```
snazy, line 1060: Nit-pick
```suggestion
                Endpoint for STS requests made by the Polaris Server (optional). If not set, defaults to
```

spec/polaris-management-service.yml
```diff
@@ -1050,12 +1050,23 @@ components:
...
+                'endpointInternal' (which in turn defaults to `endpoint`).
               example: "https://sts.example.com:1234"
+            endpointInternal:
+              type: string
+              description: >-
+                endpoint for S3 requests made by the Polaris Server (optional). If set, Polaris Service will use
+                this value instead of `endpoint`. If not set, defaults to `endpoint`. Iceberg REST API clients never
```
snazy, line 1067: Nit-pick
```suggestion
                Endpoint for S3 requests made by the Polaris Server (optional). If set, Polaris Service will use
                this value instead of `endpoint`. If not set, defaults to `endpoint`. Iceberg REST API clients never
```

spec/polaris-management-service.yml
dimas-b: All the other doc comments start with lowercase letters ;) :shrug:

spec/polaris-management-service.yml
dimas-b: probably, but I think it's marginally cleaner and clearer :smiley: based on https://yaml-multiline.info/

dimas-b:
If no more comments, I'll be merging on Aug 1.

spec/polaris-management-service.yml
snazy: Well (nit-pick-alert!)...
It's a (nearly) complete sentence, so :D

flyrain:
As a spec change, we will need to conclude the dev ML discussion or vote before merging. I don't think there is a consensus in the discussion thread, https://lists.apache.org/thread/9nggnwp8qb12of74zwh76l2qzkos3kjc.

dimas-b:
@flyrain : feel free to revert

PR #2214, opened 2025-07-31
Call MetaStoreManagerFactory.getOrCreateSession once in QuarkusProducers

XN137:
we already have a RequestScoped `BasePersistence` that we should use but its not possible as some `MetaStoreManagerFactory` classes cast the `PolarisCallContext.getMetaStore` to a narrower type unconditionally, which is not possible on the `BasePersistence`-proxy created by quarkus.

so we have to do it the other way around.


XN137:
closing as it will become obsolete anyway with other ongoing PRs most likely

PR #2217, opened 2025-07-31
JdbcMetaStoreManagerFactory determines schemaVersion once per realm

XN137:
otherwise every call to `MetaStoreManagerFactory.getOrCreateSession` was running a query to load the schema version.
i am assuming the version can not change at runtime.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -96,6 +96,8 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
     // Materialize realmId so that background tasks that don't have an active
     // RealmContext (request-scoped bean) can still create a JdbcBasePersistenceImpl
     String realmId = realmContext.getRealmIdentifier();
+    // determine schemaVersion once per realm
+    final int schemaVersion = JdbcBasePersistenceImpl.loadSchemaVersion(datasourceOperations);
     sessionSupplierMap.put(
         realmId,
         () ->
@@ -103,7 +105,8 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
                 datasourceOperations,
                 secretsGenerator(realmId, rootCredentialsSet),
                 storageIntegrationProvider,
-                realmId));
+                realmId,
+                schemaVersion));
 
     PolarisMetaStoreManager metaStoreManager = createNewMetaStoreManager();
     metaStoreManagerMap.put(realmId, metaStoreManager);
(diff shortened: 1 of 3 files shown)


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -96,14 +96,17 @@ private void initializeForRealm(
     // Materialize realmId so that background tasks that don't have an active
     // RealmContext (request-scoped bean) can still create a JdbcBasePersistenceImpl
     String realmId = realmContext.getRealmIdentifier();
+    // determine schemaVersion once per realm
+    final int schemaVersion = JdbcBasePersistenceImpl.loadSchemaVersion(datasourceOperations);
```
adutra, line 100: Couldn't this be once per JVM?

XN137: it seems possible yeah but we also run the init script once per realm it seems?

https://github.com/apache/polaris/blob/2117dbd08e8352b32a2c948ed6c166d7c77da50a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java#L148-L158

and since `DatasourceOperations` is the input parameter i thought its better to do it per realm as well.
as suggested it should still massively reduce the number of fetch schema version queries.

adutra: > we also run the init script once per realm it seems?

I didn't know 😓 That's unnecessary imho, and only works because the scripts are idempotent. We should do it in a startup method imho (IOW a method with `@Observes Startup event`).

adutra: Related: https://github.com/apache/polaris/pull/2196#pullrequestreview-3074638635

dimas-b: IMHO, we should not run DDL in Server runtime at all and assume DDL is executed _before_ the server starts (e.g. by the admin tool).

dimas-b: Re: per-realm, I believe we discussed at some point supporting multiple data sources (N:M for realms)... but it did not get implemented... However, I'd prefer to avoid assuming that the schema is the same for all realms.

adutra: In this, case, I think we need to go with some configuration option like the one I suggested in #2196:

https://github.com/apache/polaris/pull/2196#issuecomment-3139388635
`

snazy: @adutra not sure - do you request to do the other change or this change as ?

adutra: What I meant is that there is an ML thread about schema setup that will likely revisit this again: https://lists.apache.org/thread/6c7lzgy3675jrrtbrrxsm8l0xd1fo1mj

But I'm OK to merge this PR as is.

snazy: Yea - that thread looks orthogonal IMO.

PR #2219, opened 2025-07-31
JDBC: SERIALIZABLE/EntityNotFoundException

snazy:
With serializable isolation level, if either the primary-key or the check-constraint are violated, the lookup of the conflicting entity is not guaranteed to yield a non-null result (transaction start matters). This changes works around this situation w/ serializable isolation.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -183,11 +183,17 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
                   entity.getParentId(),
                   entity.getTypeCode(),
                   entity.getName());
-          throw new EntityAlreadyExistsException(existingEntity, e);
-        } else {
-          throw new RuntimeException(
-              String.format("Failed to write entity due to %s", e.getMessage()), e);
+          // This happens in two scenarios:
+          // 1. PRIMARY KEY violated
+          // 2. UNIQUE CONSTRAINT on (realm_id, catalog_id, parent_id, type_code, name) violated
+          // With SERIALIZABLE isolation, the conflicting entity may _not_ be visible and
+          // existingEntity can be null, which would cause an NPE in
+          // EntityAlreadyExistsException.message().
+          throw new EntityAlreadyExistsException(
+              existingEntity != null ? existingEntity : entity, e);
         }
+        throw new RuntimeException(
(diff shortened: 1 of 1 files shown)


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -183,11 +183,17 @@ private void persistEntity(
...
-          throw new RuntimeException(
-              String.format("Failed to write entity due to %s", e.getMessage()), e);
+          // This happens in two scenarios:
+          // 1. PRIMARY KEY violated
+          // 2. UNIQUE CONSTRAINT on (realm_id, catalog_id, parent_id, type_code, name) violated
+          // With SERIALIZABLE isolation, the conflicting entity may _not_ be visible and
+          // existingEntity can be null, which would cause an NPE in
+          // EntityAlreadyExistsException.message().
+          throw new EntityAlreadyExistsException(
+              existingEntity != null ? existingEntity : entity, e);
```
singhpk234, line 193: Thank you for catching this, wondering if we can add a coverage for this

snazy: That should IMHO be added as part of testing against different isolation level against the implementation.

dimas-b: I suppose a test to expose this bug is probabilistic in nature and, and as such does not fit very well into unit tests. I think it makes sense to merge the fix "as is" (the explanation makes sense to me) and work on a more comprehensive test framework separately.

dimas-b (approved):

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
singhpk234: is sending entity which is not committed i.e `entity`  the right thing to throw back in the exception 
https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/persistence/EntityAlreadyExistsException.java#L33

Error message has the entity id message in it and the message in this case is not correct, i don;t know what else to throw though 

Hence was requesting to add a test to see if this has repurcursions ?

snazy: Created #2263 to track this

adutra (approved):

PR #2228, opened 2025-08-01
Fix CLI error polaris: not found

eric-maynard:
After #2049, some users reported errors when running the CLI:
```
./polaris
./polaris: line 71: exec: polaris: not found
```

I was able to reproduce this and found that reverting one change from #2049 fixes the regression.


polaris
```diff
@@ -68,4 +68,4 @@ if [ -z "$VIRTUAL_ENV" ] || [ "$(realpath "$VIRTUAL_ENV")" != "$(realpath "${dir
 fi
 
 export SCRIPT_DIR="${dir}"
-exec polaris "$@"
+env PYTHONPATH=client/python SCRIPT_DIR="$dir" "${dir}"/polaris-venv/bin/python3 client/python/cli/polaris_cli.py "$@"
```
MonkeyCanCode, line 71: This one should work as it is now using the installed package instead of invoking the class directly. Do we have a reproducible or I can take a closer look a bit later this weekend (assuming the user removes the old virtualenv and setup the new one)

MonkeyCanCode: @eric-maynard 

I think it may be better to ask users to run `polaris --repair` instead add back the original code above as this will be a blocker for us to move away from `polaris` bash script:
```
➜  polaris git:(main) ./polaris --repair

➜  polaris git:(main) ./polaris
Traceback (most recent call last):
  File "/Users/yong/Desktop/GitHome/test/polaris/polaris-venv/bin/polaris", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/yong/Desktop/GitHome/test/polaris/client/python/cli/polaris_cli.py", line 219, in main
    PolarisCli.execute()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/yong/Desktop/GitHome/test/polaris/client/python/cli/polaris_cli.py", line 70, in execute
    client_builder = PolarisCli._get_client_builder(options)
  File "/Users/yong/Desktop/GitHome/test/polaris/client/python/cli/polaris_cli.py", line 175, in _get_client_builder
    raise Exception(
    ...<5 lines>...
    )
Exception: Please provide credentials via either --client-id & --client-secret or --access-token. Alternatively, you may set the environment variables CLIENT_ID & CLIENT_SECRET.

➜  polaris git:(main) cat polaris | grep exec
exec polaris "$@"
```

eric-maynard: Let me ask, good idea. Separately, IIUC the podman requirement can go away in the future too?

MonkeyCanCode: Podman (docker) requirement for codegen via openapi? If yes, it is already addressed in https://github.com/apache/polaris/pull/2192 but pending for review.

eric-maynard:
closing for now to confirm that --repair can address this -- thanks @MonkeyCanCode !

PR #2229, opened 2025-08-01
Remove PolarisCallContext.getMetaStore

XN137:
unfinished WIP

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -200,16 +200,6 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   }
 
   @Override
-  public synchronized BasePersistence getOrCreateSession(RealmContext realmContext) {
-    if (!sessionSupplierMap.containsKey(realmContext.getRealmIdentifier())) {
-      DatasourceOperations datasourceOperations = getDatasourceOperations();
-      initializeForRealm(datasourceOperations, realmContext, null);
-    }
-    checkPolarisServiceBootstrappedForRealm(realmContext);
-    return sessionSupplierMap.get(realmContext.getRealmIdentifier()).get();
-  }
-
-  @Override
   public synchronized EntityCache getOrCreateEntityCache(
       RealmContext realmContext, RealmConfig realmConfig) {
     if (!entityCacheMap.containsKey(realmContext.getRealmIdentifier())) {
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/LocalPolarisMetaStoreManagerFactory.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/LocalPolarisMetaStoreManagerFactory.java
@@ -154,15 +154,6 @@ public abstract class LocalPolarisMetaStoreManagerFactory<StoreType>
   }
 
   @Override
-  public synchronized TransactionalPersistence getOrCreateSession(RealmContext realmContext) {
-    if (!sessionSupplierMap.containsKey(realmContext.getRealmIdentifier())) {
-      initializeForRealm(realmContext, null);
-    }
-    checkPolarisServiceBootstrappedForRealm(realmContext);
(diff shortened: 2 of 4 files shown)


polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java
```diff
@@ -2903,9 +2900,10 @@ void testPolicyMappingCleanup() {
     // Drop N1_N2_T1, the corresponding policy mapping should be cleaned-up
     this.dropEntity(List.of(catalog, N1, N1_N2), N1_N2_T3);
 
-    BasePersistence ms = polarisCallContext.getMetaStore();
+    // hackish way to inspect internal metastore
+    BasePersistence metaStore = ((BaseMetaStoreManager<?>) polarisMetaStoreManager).getMetaStore();
```
snazy, line 2904:
```suggestion
    BasePersistence metaStore = getBasePersistence();
```

plus the implementation
```
  protected BasePersistence getBasePersistence() {
    return ((BaseMetaStoreManager<?>) polarisMetaStoreManager).getMetaStore();
  }
```

polaris-core/src/main/java/org/apache/polaris/core/persistence/BaseMetaStoreManager.java
```diff
@@ -53,6 +56,16 @@ public static PolarisStorageConfigurationInfo extractStorageConfiguration(
     return PolarisStorageConfigurationInfo.deserialize(storageConfigInfoStr);
   }
 
+  private final Supplier<MS> metaStoreSupplier;
```
dimas-b, line 59: nit: move above methods?

polaris-core/src/main/java/org/apache/polaris/core/persistence/BaseMetaStoreManager.java
```diff
@@ -53,6 +56,16 @@ public static PolarisStorageConfigurationInfo extractStorageConfiguration(
...
+
+  public BaseMetaStoreManager(Supplier<MS> metaStoreSupplier) {
+    this.metaStoreSupplier = Suppliers.memoize(metaStoreSupplier::get);
```
XN137, line 62: it is still unclear what the actual lifetime scope of the `BasePersistence` should be here.

on main its lifetime is bound to the `PolarisCallContext`.
when we use `memoize` here, we'd have a single instance per realm.
when we dont use `memoize` here, we would be creating more instances than before, because for example `AtomicOperationMetaStoreManager#loadResolvedEntityByName` calls `AtomicOperationMetaStoreManager#createEntityIfNotExists` internally, both of which call `getMetaStore()`.

i am wondering if the subclasses of BaseMetaStoreManager would need to tell the base class how to use the Supplier, depending on whether their specific BasePersistence impl instance has state or not.

theoretically we could also let BaseMetaStoreManager have a cache keyed by PolarisCallContext to keep exactly the existing "instance per callcontext" semantics though it would look a bit hacky

snazy:
Hm,
we might have to defer this one for a bit. Looks like there are some cases that could change their behavior (although not covered in test cases) - the cases when the "logical operation" relies on the same `BasePersistence` being used for a couple operations - we have to at least verify that those cases do not change their behavior.

polaris-core/src/main/java/org/apache/polaris/core/persistence/BaseMetaStoreManager.java
dennishuo: FYI A lot of the machinery behind the MetaStoreManager/BasePersistence interactions/factories exists precisely to enforce the contract that BasePersistence instances are bound to a PolarisCallContext lifetime and are RequestScoped.

If the request scoping wasn't necessary, indeed none of the indirection would be necessary. Similarly, machinery for making MetaStoreManager realm-scoped goes beyond vanilla `@ApplicationScoped` annotations in order to define/enforce realm-scoping.

XN137: as per my comment this is still WIP and i am figuring out the actual lifetime of the `BasePersistence` object.

if what you are saying is true and the currently suggested change would break an existing contract, how is it that no test in CI is failing?
can you help adding a test that covers this contract explicitly?
or at least describe how such a test would need to look like?
that would help a lot!

XN137:
closing as this will be replaced by https://github.com/apache/polaris/pull/2555

PR #2233, opened 2025-08-01
Merge polaris-service-common into polaris-runtime-service

adutra:
As discussed in the ML here is a PR that merges polaris-service-common into polaris-runtime-service.

99% of the changes are file moves or renamings:

- The first commit merges modules
- The second commit merges packages
- The third commit renames classes

Commits can be reviewed independently.

Only the 3rd one contains changes of interest.

For example, all classes like `QuarkusXYZConfiguration` were merged into `XYZConfiguration` if such class exists, rather than renamed.

One particular interface was problematic: `FeaturesConfiguration` and its subtype `ResolvedFeaturesConfiguration`. The merger resulted in a cyclic reference when resolving CDI beans. The same issue also happened with `BehaviorChangesConfiguration` and `ResolvedBehaviorChangesConfiguration`.

To solve both cyclic reference issues, I created the following hierarchy:

image

BTW @eric-maynard what is `BehaviorChangesConfiguration` used for? It seems unused.

Also: I renamed `QuarkusProducers` in both modules that had such class. They became: `ServiceProducers` and `AdminToolProducers`. I also renamed `QuarkusSparkIT` to `SparkIT`.


adutra:
The failing test does not fail to me locally: `testCreateGcpCredentialsFromStaticToken` – I am investigating.

EDIT: not failing anymore 😆 🤷‍♂️

dimas-b (approved):
LGTM in general, but I did not check every single file :sweat_smile:

adutra:
> The failing test does not fail to me locally: `testCreateGcpCredentialsFromStaticToken` – I am investigating.
> 
> EDIT: not failing anymore 😆 🤷‍♂️

Failing again :-( 

The error is:

```
Expecting actual: 1754133695923L to be between: [1754133696234L, 1754133697234L]
```

IOW, the token expires 311 milliseconds before the lowest bound.

I think this test is probably flaky: If a GC pause happens between the token issuance and the call to `Instant.now()`, we can have more than 500 milliseconds of difference. I will provide a fix.

adutra:
> I did not check every single file 😅

I hope you didn't ! 😆

adutra:
FYI: the flaky test should be fixed by #2241.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -36,13 +44,15 @@
...
+import org.apache.polaris.service.tracing.TracingFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Given a list of registered {@link TaskHandler}s, execute tasks asynchronously with the provided
  * {@link CallContext}.
  */
+@ApplicationScoped
 public class TaskExecutorImpl implements TaskExecutor {
```
snazy, line 57:
```suggestion
class TaskExecutorImpl implements TaskExecutor {
```

Probably similar for other implementations.

If it works, the constructors can be package-private as well.

runtime/service/src/main/java/org/apache/polaris/service/config/BehaviorChangesConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.config;
+
+/**
+ * FIXME: this seems unused.
```
snazy, line 22: Maybe remove?

runtime/service/src/main/java/org/apache/polaris/service/config/RawBehaviorChangesConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.config;
+
+import io.smallrye.config.ConfigMapping;
+
+@ConfigMapping(prefix = "polaris.behavior-changes")
```
snazy, line 23: Why not have this on the base interface?

runtime/service/src/main/java/org/apache/polaris/service/config/RawFeaturesConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.config;
+
+import io.smallrye.config.ConfigMapping;
+
+@ConfigMapping(prefix = "polaris.features")
```
snazy, line 23: Why not have this on the base interface?

runtime/service/src/main/java/org/apache/polaris/service/config/JacksonConfig.java
```diff
@@ -31,20 +31,19 @@
...
 import jakarta.inject.Singleton;
 import org.apache.iceberg.rest.RESTSerializers;
-import org.apache.polaris.service.config.Serializers;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @Singleton
-public class QuarkusJacksonConfig implements ObjectMapperCustomizer {
+public class JacksonConfig implements ObjectMapperCustomizer {
```
snazy: Nomenclature:
```suggestion
public class PolarisIcebergObjectMapperCustomizer implements ObjectMapperCustomizer {
```
... because this configures the global `ObjectMapper` specific to Iceberg + inherited Polaris specifics.

runtime/service/src/main/java/org/apache/polaris/service/legacy/LegacyManagementEndpoints.java
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.polaris.service.quarkus.legacy;
+package org.apache.polaris.service.legacy;
```
snazy, line 19: Unrelated to this PR:
I suspect we can remove this one for the next release.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -52,18 +62,28 @@ public class TaskExecutorImpl implements TaskExecutor {
   private final TaskFileIOSupplier fileIOSupplier;
   private final List<TaskHandler> taskHandlers = new CopyOnWriteArrayList<>();
   private final PolarisEventListener polarisEventListener;
+  private final Tracer tracer;
```
snazy: I'd prefer to make this a 
```suggestion
  private final Optional<Tracer> tracer;
```

I'm a bit cautious when using otel code, been bitten hard w/ static initializers in the past, although the particular test doesn't seem to harm.

runtime/service/src/test/java/org/apache/polaris/service/auth/external/PrincipalAuthInfoAugmentorTest.java
```diff
@@ -32,39 +32,38 @@
...
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 class PrincipalAuthInfoAugmentorTest {
 
   private PrincipalAuthInfoAugmentor augmentor;
-  private org.apache.polaris.service.quarkus.auth.external.mapping.PrincipalMapper principalMapper;
-  private org.apache.polaris.service.quarkus.auth.external.mapping.PrincipalRolesMapper
+  private org.apache.polaris.service.auth.external.mapping.PrincipalMapper principalMapper;
+  private org.apache.polaris.service.auth.external.mapping.PrincipalRolesMapper
```
snazy, line 46: Do we still need the fully qualified types here?

runtime/service/src/test/java/org/apache/polaris/service/exception/ExceptionMapperTest.java
```diff
@@ -48,6 +48,7 @@ public class ExceptionMapperTest {
   public void testFullExceptionIsLogged(
       ExceptionMapper<Exception> mapper, Exception exception, Level level) {
     Logger logger = (Logger) LoggerFactory.getLogger(mapper.getClass());
+    logger.setLevel(level);
```
snazy, line 51: Huh? What's the reason for adding this?

runtime/service/src/main/java/org/apache/polaris/service/config/RawBehaviorChangesConfiguration.java
adutra: Because this bean is injected in `ResolvedBehaviorChangesConfiguration` which also implements BehaviorChangesConfiguration, so that would result in an infinite recursion 🤷‍♂️

runtime/service/src/main/java/org/apache/polaris/service/config/BehaviorChangesConfiguration.java
adutra: Well, I added this FIXME because I'm truly puzzled by this class. @eric-maynard what is this used for?

runtime/service/src/main/java/org/apache/polaris/service/config/RawFeaturesConfiguration.java
adutra: See my comment: https://github.com/apache/polaris/pull/2233#discussion_r2251113794

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
adutra: But how would I inject this, `Instance<Tracer>`? I'm not sure that's better.

runtime/service/src/test/java/org/apache/polaris/service/exception/ExceptionMapperTest.java
adutra: Yes :-( there are similarly-named inner classes in `OidcTenantConfiguration`.

runtime/service/src/test/java/org/apache/polaris/service/auth/external/PrincipalAuthInfoAugmentorTest.java
adutra: Yes :-( there are similarly-named inner classes in OidcTenantConfiguration.

runtime/service/src/test/java/org/apache/polaris/service/exception/ExceptionMapperTest.java
adutra: This test is brittle because it assumes that the logger has the tested level enabled, but this is not the case anymore.

Anyways, this test is not very Quarkus-friendly. I will refactor it later.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
snazy: Hm,
`Instance.select().stream().findFirst()` seems to be the only way...

snazy (approved):

snazy (approved):

adutra:
Hi, since this PR is generating constant rebase conflicts with other ongoing PRs, I'm going to go ahead and merge it. I hope that's ok for everybody. 🙏

PR #2235, opened 2025-08-01
Remove config parameter from `PolarisStorageIntegration#getSubscopedCreds`

snazy:
Instances of `PolarisStorageIntegration` are created for a particular `PolarisStorageConfigurationInfo`, the same value is then passed into `PSI.getSubscopedCreds()`.

This change removes the config parameter, as it's already known when `PolarisStorageIntegration` instances are created.

### DIFF
--- a/service/common/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
+++ b/service/common/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
@@ -37,9 +37,12 @@ import org.apache.polaris.core.storage.PolarisStorageConfigurationInfo;
 import org.apache.polaris.core.storage.PolarisStorageIntegration;
 import org.apache.polaris.core.storage.PolarisStorageIntegrationProvider;
 import org.apache.polaris.core.storage.aws.AwsCredentialsStorageIntegration;
+import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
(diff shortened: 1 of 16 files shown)


adutra (approved):

service/common/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -81,27 +85,32 @@ public PolarisStorageIntegrationProviderImpl(
...
         storageIntegration =
-            (PolarisStorageIntegration<T>) new AzureCredentialsStorageIntegration();
+            (PolarisStorageIntegration<T>)
+                new AzureCredentialsStorageIntegration(
+                    (AzureStorageConfigurationInfo) polarisStorageConfigurationInfo);
         break;
       case FILE:
         storageIntegration =
-            new PolarisStorageIntegration<>("file") {
+            new PolarisStorageIntegration<>((T) polarisStorageConfigurationInfo, "file") {
```
dimas-b, line 109: `T` is `FileStorageConfigurationInfo` here, right?

snazy: Yea, but javac is a bit opinionated here ...

snazy:
Rebased after merge conflict w/ #2234

adutra (approved):

PR #2237, opened 2025-08-01
Prepare upgrade to Gradle 9

snazy:
* `shadowPub.kt` the change removes a special case that doesn't apply to Polaris
* `api/...` build scripts - changes due to nullable type handling (`Property<String>` vs `Property<String?>` - latter is ... weird)

### DIFF
--- a/api/management-model/build.gradle.kts
+++ b/api/management-model/build.gradle.kts
@@ -50,13 +50,13 @@ val generatedOpenApiSrcDir = project.layout.buildDirectory.dir("generated-openap
 openApiGenerate {
   // The OpenAPI generator does NOT resolve relative paths correctly against the Gradle project
   // directory
-  inputSpec = specsDir.file("polaris-management-service.yml").asFile.absolutePath
+  inputSpec = provider { specsDir.file("polaris-management-service.yml").asFile.absolutePath }
   generatorName = "jaxrs-resteasy"
-  outputDir = generatedDir.get().asFile.absolutePath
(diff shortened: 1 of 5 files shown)


api/management-model/build.gradle.kts
```diff
@@ -50,13 +50,13 @@ val generatedOpenApiSrcDir = project.layout.buildDirectory.dir("generated-openap
...
-  inputSpec = specsDir.file("polaris-management-service.yml").asFile.absolutePath
+  inputSpec = provider { specsDir.file("polaris-management-service.yml").asFile.absolutePath }
   generatorName = "jaxrs-resteasy"
-  outputDir = generatedDir.get().asFile.absolutePath
+  outputDir = provider { generatedDir.get().asFile.absolutePath }
   modelPackage = "org.apache.polaris.core.admin.model"
-  ignoreFileOverride = rootDir.file(".openapi-generator-ignore").asFile.absolutePath
-  removeOperationIdPrefix = true
-  templateDir = templatesDir.asFile.absolutePath
+  ignoreFileOverride.set(provider { rootDir.file(".openapi-generator-ignore").asFile.absolutePath })
```
snazy, line 57: This is because the property's declared as `Property<String?>` by the plugin extension. It doesn't make sense to declare nullable values as `Property`s - it has a value or not.

snazy (commented):
Noting: Gradle 9 also bumps to Kotlin 2.

dimas-b (approved):

adutra (approved):

ebyhr (approved):

PR #2241, opened 2025-08-02
Use application-scoped clock when generating GCP credentials

adutra:
This change also fixes a flaky test: `StorageConfigurationTest.testCreateGcpCredentialsFromStaticToken`.

### DIFF
--- a/service/common/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
+++ b/service/common/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
@@ -26,6 +26,7 @@ import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import java.time.Clock;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


snazy:
Honestly, I'm a bit confused.
Looking at the only production code call site `o.a.p.service.storage.PolarisStorageIntegrationProviderImpl#PolarisStorageIntegrationProviderImpl(o.a.p.service.storage.StorageConfiguration, o.a.p.core.storage.aws.StsClientProvider)` (an application-scoped bean), it seems the GCP access token is only ever created once?

adutra:
> Honestly, I'm a bit confused. Looking at the only production code call site `o.a.p.service.storage.PolarisStorageIntegrationProviderImpl#PolarisStorageIntegrationProviderImpl(o.a.p.service.storage.StorageConfiguration, o.a.p.core.storage.aws.StsClientProvider)` (an application-scoped bean), it seems the GCP access token is only ever created once?

Yes 😅 

I think the idea is that these are long-lived credentials for the server itself. But I agree that this is likely not ideal. We'd need something similar to `StsClientProvider`.

This PR doesn't change this though, it just introduces the injected `Clock`.

adutra:
(and I need this for #2233)

snazy:
> > Honestly, I'm a bit confused. Looking at the only production code call site `o.a.p.service.storage.PolarisStorageIntegrationProviderImpl#PolarisStorageIntegrationProviderImpl(o.a.p.service.storage.StorageConfiguration, o.a.p.core.storage.aws.StsClientProvider)` (an application-scoped bean), it seems the GCP access token is only ever created once?
> 
> Yes 😅
> 
> I think the idea is that these are long-lived credentials for the server itself. But I agree that this is likely not ideal. We'd need something similar to `StsClientProvider`.
> 
> This PR doesn't change this though, it just introduces the injected `Clock`.

I know - do we have an issue for this?
Looks like GCS is only usable for 1 hour :shrug: ?

adutra:
> I know - do we have an issue for this? Looks like GCS is only usable for 1 hour 🤷 ?

There you go: https://github.com/apache/polaris/issues/2253

PR #2250, opened 2025-08-04
Remove PolarisCallContext.getClock

XN137:
the Clock is application scoped and thus should not be put into any
realm or call specific context class.

since there were only a handful of callers this is a good opportunity to add constructors to
the `MetaStoreManager` impls.


snazy (approved):

XN137:
this will get rebased and put up for review after those PRs landed:
```
https://github.com/apache/polaris/pull/2248
https://github.com/apache/polaris/pull/2249
https://github.com/apache/polaris/pull/2251
```

XN137:
now only based on `https://github.com/apache/polaris/pull/2251`

PR #2251, opened 2025-08-04
Use injected PolarisDiagnostics in MetaStoreManagerFactory impls

XN137:
`new PolarisDefaultDiagServiceImpl()` should not be used anywhere in production code outside of `QuarkusProducers`, as it turns out a `PolarisDiagnostics` instance already was getting injected.

this actually required fixing the setup in `CatalogEntityTest` so that we dont pass in `null`.

for clarity we mark ctors that only exist for CDI as protected so that they dont get used by tests accidentally.

PR #2248, opened 2025-08-04
Stop mocking PolarisDiagnostics

XN137:
if diagnostics checks are failing in our tests we want to know about it as input validation and object invariants of production code should never be bypassed

PR #2254, opened 2025-08-04
Remove `PolarisCallContext` from `BasePersistence` et al

snazy:
Following up on #2251, this change removes the `PolarisCallContext` parameter from the interfaces `BasePersistence`, `IntegrationPersistence` and `PolicyMappingPersistence`. None of the implementations used the parameter for anything else than getting a `PolarisDiagnostics` instance, which is implicitly available in the containing type.


snazy:
~Also related to (supplement of) #2247 and #2248~

PR #2258, opened 2025-08-05
fix: Filter null entities in listPrincipals and listPrincipalRoles to prevent NPE

HotSushi:
## Summary
Similar to listCatalogs fix, these methods are non-atomic and can cause NPE when entities are deleted between listEntities and loadEntity calls. Added  .filter(Objects::nonNull) to safely handle race conditions.

Similar to:  https://github.com/apache/polaris/pull/1949

Tagging reviewers from previous pr:
@eric-maynard 
@dimas-b 
@andrew4699 

## Tests
Added tests to verify null filtering behavior when entities are deleted  after being listed but before being loaded.
✅ `./gradlew build` succeeds
✅ Added unit tests

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/ManagementServiceTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/quarkus/admin/ManagementServiceTest.java
@@ -315,4 +315,81 @@ public class ManagementServiceTest {
     assertThat(catalogs.size()).isEqualTo(1);
     assertThat(catalogs.getFirst().getId()).isEqualTo(catalog2.getCatalog().getId());
   }
+
+  /**
+   * Simulates the case when a principal is dropped after being found while listing all principals.
+   */
+  @Test
+  public void testPrincipalNotReturnedWhenDeletedAfterListBeforeGet() {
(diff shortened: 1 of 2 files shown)


XN137 (commented):
the fix indeed avoids some NPEs but the fact that similar problems had to be fixed previously (and `listCatalogRoles` is still suffering from the same problem afaict) makes me believe the problem should be addressed in a more central spot.

from a high level perspective it should not be the responsibility of `PolarisAdminService` to deal with race-condition issues in the `PolarisMetaStoreManager` for simple operations like "load all entities of type X".

i've put up https://github.com/apache/polaris/pull/2261 so that we can have a single place for the workaround while future work can look to improve the `PolarisMetaStoreManager` api in that regard.

if we think the added tests still provide additional value my suggestion would be to rebase your PR after the other PR has been merged, wdyt?

HotSushi:
Thanks @XN137, agree if there's a better place for the fix to be checked in. Thank you for jumping in. 

After that fix, I do not think tests would serve additional benefit. Do use them verbatim if you like them in your PR.

snazy:
Fixing the root cause sounds like the better path forwards. Provided a possible option in https://github.com/apache/polaris/pull/2261#pullrequestreview-3088951142

XN137:
since the other PR has been merged i'd suggest to close this PR now

PR #2261, opened 2025-08-05
Add PolarisAdminService.loadEntities helper

XN137:
`PolarisAdminService` has multiple spots where it is working around the sub-optimal `PolarisMetaStoreManager` APIs.

This results in multiple fixes like:
https://github.com/apache/polaris/pull/1949
https://github.com/apache/polaris/pull/2258

While eventually the underlying APIs should be improved, for now we can make a single central workaround and clean up some redundant code. Also we can improve the return types as callers are not interested in details of the entity layer.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
@@ -36,6 +36,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.TableIdentifier;
@@ -51,6 +52,7 @@ import org.apache.polaris.core.admin.model.BearerAuthenticationParameters;
 import org.apache.polaris.core.admin.model.Catalog;
 import org.apache.polaris.core.admin.model.CatalogGrant;
 import org.apache.polaris.core.admin.model.CatalogPrivilege;
+import org.apache.polaris.core.admin.model.CatalogRole;
 import org.apache.polaris.core.admin.model.ConnectionConfigInfo;
 import org.apache.polaris.core.admin.model.CreateCatalogRequest;
 import org.apache.polaris.core.admin.model.ExternalCatalog;
@@ -60,6 +62,8 @@ import org.apache.polaris.core.admin.model.NamespacePrivilege;
 import org.apache.polaris.core.admin.model.OAuthClientCredentialsParameters;
 import org.apache.polaris.core.admin.model.PolicyGrant;
 import org.apache.polaris.core.admin.model.PolicyPrivilege;
+import org.apache.polaris.core.admin.model.Principal;
+import org.apache.polaris.core.admin.model.PrincipalRole;
 import org.apache.polaris.core.admin.model.PrincipalWithCredentials;
 import org.apache.polaris.core.admin.model.PrincipalWithCredentialsCredentials;
 import org.apache.polaris.core.admin.model.TableGrant;
@@ -82,6 +86,7 @@ import org.apache.polaris.core.entity.CatalogRoleEntity;
 import org.apache.polaris.core.entity.NamespaceEntity;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
 import org.apache.polaris.core.entity.PolarisEntity;
+import org.apache.polaris.core.entity.PolarisEntityCore;
 import org.apache.polaris.core.entity.PolarisEntitySubType;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisGrantRecord;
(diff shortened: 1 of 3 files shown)


service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -200,6 +205,43 @@ private Optional<CatalogRoleEntity> findCatalogRoleByName(String catalogName, St
...
+      @Nonnull PolarisEntitySubType entitySubType,
+      @Nullable PolarisEntity catalogEntity,
+      @Nonnull Function<PolarisBaseEntity, T> transformer) {
+    List<PolarisEntityCore> catalogPath;
+    long catalogId;
+    if (catalogEntity == null) {
+      catalogPath = null;
+      catalogId = 0;
+    } else {
+      catalogPath = PolarisEntity.toCoreList(List.of(catalogEntity));
```
tmater: I think this could be simplified, `PolarisEntity.toCoreList()` returns `null` when the input is null or empty.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -200,6 +205,43 @@ private Optional<CatalogRoleEntity> findCatalogRoleByName(String catalogName, St
...
+            catalogPath,
+            entityType,
+            entitySubType,
+            PageToken.readEverything())
+        .getEntities()
+        .stream()
+        .map(
+            nameAndId ->
+                metaStoreManager.loadEntity(
+                    getCurrentPolarisContext(), catalogId, nameAndId.getId(), nameAndId.getType()))
```
tmater: Just a quick question, is there a specific reason we need to save the `catalogId` earlier? I'm asking because `nameAndId` already provides `getCatalogId()`.

tmater (commented):
Thanks @XN137! Nifty change, overall LGTM!

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
XN137: i had noticed that `toCoreList` has some `null` handling but if we passed `PolarisEntity.toCoreList(List.of(null))` i think it would still result in an NPE ?

imo having the explicit "if null" makes the overall flow clearer either way

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
XN137: it might be true but i am only following what the existing code was doing... whether `getCatalogId()` always returns the right value for all types of entities idk, so i was just sticking to the existing code

XN137: to elaborate: in `listCatalogsUnsafe` the returned entity should likely return the same value for `getId` and `getCatalogId` (not sure if it does) but it seems like the `loadEntity` api still requires us to pass `0` since catalogs are top level entites

snazy (commented):
The culprit here is that although all matching entities are known during the list-operation, all you get from it is this `EntityNameLookupRecord`, so you have to load the same entities again.
Having a counterpart of `PolarisMetaStoreManager#listEntities` that doesn't yield the unnecessarily reduced result but the actual entities would help.
WDYT?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -200,6 +205,43 @@ private Optional<CatalogRoleEntity> findCatalogRoleByName(String catalogName, St
...
+    List<PolarisEntityCore> catalogPath;
+    long catalogId;
+    if (catalogEntity == null) {
+      catalogPath = null;
+      catalogId = 0;
+    } else {
+      catalogPath = PolarisEntity.toCoreList(List.of(catalogEntity));
+      catalogId = catalogEntity.getId();
+    }
+    // TODO: add loadEntities method to PolarisMetaStoreManager
```
flyrain, line 222: Minor: We will need a discussion on how persistence layer is going to support it, as it affects all types of persistence. Loading everything in one call can provide a consistent view, which is nice, but there are some caveats that the uber call may be too large, so that it hits the limits(e.g., memory limit). With that, I think it's premature to consider this as a `TODO` item.

flyrain (approved):
Thanks @XN137 for doing this. LGTM!

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
tmater: Ah, yeah, makes sense, forgot about that NPE.

service/common/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
tmater: Thanks for clarifying it!

XN137:
> The culprit here is that although all matching entities are known during the list-operation, all you get from it is this EntityNameLookupRecord, so you have to load the same entities again.
Having a counterpart of PolarisMetaStoreManager#listEntities that doesn't yield the unnecessarily reduced result but the actual entities would help.
WDYT?

yeah this is mentioned in the commit message and implied by the added TODO.
but a fix/change at the persistence layer is more involved, so as a first step and immediate replacement for PR-2258 i thought centralizing the workaround is useful, as it doesnt prevent or delay the proper fix.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
XN137: afaict under the hood `listEntities` is already fetching the fully fledged `PolarisBaseEntity` instances from the database (for the jdbc case):
https://github.com/apache/polaris/blob/af69d9f9699d8a0cae1b70839e1b5c1bfc9ba511/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java#L507-L513
it includes streaming/pagination.

it just happens that the given `transformer` turns `PolarisBaseEntity` into `EntityNameLookupRecord`... and then later the record is used to look up the full entity again.
so afaict, the memory footprint would not be very different if we had a `loadEntities` function (or changed `listEntities` to not only return a `EntityNameLookupRecord`).

snazy: > so afaict, the memory footprint would not be very different if we had a loadEntities function

I agree. I also think the overall load is lower.

flyrain: To be clear, I'm not against the idea, but we will need a discussion.

PR #2268, opened 2025-08-05
Standardize logging libraries in tests

adutra:
This change enforces the following test logging patterns:

- Non-Quarkus modules use Logback Classic, configured via logback-test.xml
- Quarkus modules use JBoss Logging Manager, configured in Quarkus configuration files.

This change also introduces a workaround for the "duplicate log messages" issues with Gradle + JBoss Logging Manager. See this issue for context:

https://github.com/quarkusio/quarkus/issues/22844

The workaround implemented in this PR is very similar to the one proposed in this comment:

https://github.com/quarkusio/quarkus/issues/22844#issuecomment-1824720923

Note: it's not entirely possible imho to suppress the following message on the console:

```
The Agroal dependency is present but no JDBC datasources have been defined.
```

This is because:

1. This specific message happens during augmentation phase, not during tests
2. And it suffers from the "duplicate message" issue with (it's actually Gradle that prints those messages).

### DIFF
--- a/build-logic/src/main/kotlin/polaris-runtime.gradle.kts
+++ b/build-logic/src/main/kotlin/polaris-runtime.gradle.kts
@@ -66,7 +66,7 @@ configurations.all {
     if (group == "ch.qos.logback") {
       throw GradleException(
         "Logback dependencies are not allowed in Quarkus modules. " +
-          "Found $group:$name in ${project.name}"
+          "Found $group:$name in ${project.name}."
       )
     }
   }


runtime/defaults/src/main/resources/application.properties
```diff
@@ -68,9 +68,16 @@ quarkus.log.file.path=./logs/polaris.log
 quarkus.log.file.rotation.file-suffix=.yyyy-MM-dd.gz
 quarkus.log.file.rotation.max-file-size=10M
 quarkus.log.file.rotation.max-backup-index=14
-quarkus.log.category."org.apache.polaris".level=INFO
-quarkus.log.category."org.apache.iceberg.rest".level=INFO
-quarkus.log.category."io.smallrye.config".level=INFO
+
+# Useful loggers. Set "io.smallrye.config"to DEBUG to print the resolved configuration properties.
```
adutra: These loggers were getting in the way for integration tests, so better to comment them out.

runtime/defaults/src/main/resources/application-it.properties
```diff
@@ -17,15 +17,22 @@
...
-# Note: Quarkus integration tests cannot use QuarkusTestProfile
+# Configuration common to ALL integration tests (executed with "it" profile – Gradle "intTest" tasks).
+# Note: Quarkus integration tests cannot use QuarkusTestProfile.
+
+# Note about logging during integration tests:
+# - Logging for test code is configured in the `application-test.properties` file (!!).
+#   By default, it is set to `ERROR` level.
+# - Logging for application code is configured in the main `application.properties` file.
+#   By default, it is set to `INFO` level.
+# As far as I can tell, configuring logging levels or appenders in this file is ineffective.
```
adutra: This is the nice little surprise of the day 😆

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -47,6 +54,23 @@ testing {
   }
 }
 
+dependencies {
+  // All Quarkus projects should use JBoss LogManager with SLF4J, instead of Logback
+  implementation("org.jboss.slf4j:slf4j-jboss-logmanager")
+}
```
snazy, line 61: You may want something like this to ban logback from Quarkus.
```
configurations.all { exclude(group = "ch.qos.logback", module = "logback-classic") }
```

BTW: it would also need to be removed from the LICENSE/NOTICE

build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -107,11 +107,13 @@ testing {
       dependencies {
         implementation(project())
         implementation(testFixtures(project()))
-        runtimeOnly(
-          libs.findLibrary("logback-classic").orElseThrow {
-            GradleException("logback-classic not declared in libs.versions.toml")
-          }
-        )
+        if (!plugins.hasPlugin("io.quarkus")) {
```
snazy, line 110: Think this isn't necessary (see my comment below)

persistence/relational-jdbc/src/test/resources/logback-test.xml
```diff
@@ -19,21 +19,14 @@
...
       <pattern>%date{ISO8601} [%thread] %-5level %logger{36} - %msg%n</pattern>
     </encoder>
   </appender>
-  <root level="${log.level.console:-INFO}">
+  <root level="${test.log.level:-ERROR}">
     <appender-ref ref="console"/>
   </root>
-
-  <!--
-  Prevent the 'The Agroal dependency is present but no JDBC datasources have been defined.' build-time warning.
```
snazy, line 34: Oh!
Now I understand why that only _sometimes_ worked

persistence/relational-jdbc/src/test/resources/logback-test.xml
```diff
@@ -19,21 +19,14 @@
  under the License.
 
 -->
-<configuration>
+<configuration debug="false">
```
snazy, line 22: Should the file be removed, since logback isn't there?

plugins/spark/v3.5/spark/src/test/resources/logback-test.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" ?>
```
snazy, line 1: Should the file be removed, since logback isn't there?

runtime/service/build.gradle.kts
```diff
@@ -135,12 +136,6 @@ dependencies {
   testImplementation("software.amazon.awssdk:kms")
   testImplementation("software.amazon.awssdk:dynamodb")
 
-  runtimeOnly(project(":polaris-relational-jdbc"))
-  runtimeOnly("io.quarkus:quarkus-jdbc-postgresql") {
-    exclude(group = "org.antlr", module = "antlr4-runtime")
-    exclude(group = "org.scala-lang", module = "scala-library")
-    exclude(group = "org.scala-lang", module = "scala-reflect")
```
snazy: Those excludes are no longer necessary?

runtime/service/build.gradle.kts
```diff
@@ -135,12 +136,6 @@ dependencies {
   testImplementation("software.amazon.awssdk:kms")
   testImplementation("software.amazon.awssdk:dynamodb")
 
-  runtimeOnly(project(":polaris-relational-jdbc"))
-  runtimeOnly("io.quarkus:quarkus-jdbc-postgresql") {
```
snazy: Why do those need to be `implementation` deps now?

runtime/test-common/src/main/java/org/apache/polaris/test/commons/junit/GradleDuplicateLoggingWorkaround.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static void removeGradleHandlers() {
+    if (!DONE.getAndSet(true)) {
+      var logger = org.jboss.logmanager.LogManager.getLogManager().getLogger("");
+      for (var handler : logger.getHandlers()) {
+        if (handler.getClass().getSimpleName().equals("SLF4JBridgeHandler")) {
+          logger.removeHandler(handler);
+        }
+      }
+      LOGGER.warn("Removed SLF4JBridgeHandler to fix duplicate logging messages in Gradle tests.");
```
snazy: Nit: only log if a handler was removed?

runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
```diff
@@ -87,19 +87,27 @@ public class IcebergExceptionMapper implements ExceptionMapper<RuntimeException>
   private static final Set<String> ACCESS_DENIED_HINTS =
       Set.of("access denied", "not authorized", "forbidden");
 
-  public IcebergExceptionMapper() {}
+  private final Logger logger;
```
snazy: Nit: instead of having a field, have a function that yields `LOGGER` in production code, but is overridden in tests. Should be easier for C2 to inline.

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
adutra: This would silently remove the dependency. Isn't it better to fail the build?

Re: LICENSE and NOTICE: I checked and logback is not present in any of these files.

persistence/relational-jdbc/src/test/resources/logback-test.xml
adutra: This module is not a Quarkus module, so it uses Logback. It's OK to have this file.

plugins/spark/v3.5/spark/src/test/resources/logback-test.xml
adutra: Same: this module is not a Quarkus module.

runtime/service/build.gradle.kts
adutra: These were introduced to overcome the problems with `SparkIT` – but now `SparkIT` is isolated in a different module, so we can remove this.

But, this is an unrelated change, so I will create a separate PR for it:

https://github.com/apache/polaris/pull/2273

runtime/service/build.gradle.kts
adutra: Let's discuss in https://github.com/apache/polaris/pull/2273

runtime/service/build.gradle.kts
adutra: This change was reverted in this PR FYI.

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
snazy: > This would silently remove the dependency

Yes, in Quarkus, where you definitely don't want to have logback

persistence/relational-jdbc/src/test/resources/logback-test.xml
snazy: ah, right, ignore this comment

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
adutra: But don't you think this could confuse users? 

E.g. some user adds logback-classic to a new Quarkus module, and expects logging to be configured with logback-test.xml, but instead, the dependency is silently removed and logging is not configured as expected. 

Instead the following block fails hard on that situation:

```kotlin
configurations.all {
  // Validate that Logback dependencies are not used in Quarkus modules.
  dependencies.configureEach {
    if (group == "ch.qos.logback") {
      throw GradleException(
        "Logback dependencies are not allowed in Quarkus modules. " +
          "Found $group:$name in ${project.name}"
      )
    }
  }
}
```

Wdyt?

snazy: WFM

PR #2273, opened 2025-08-06
Nit: simplify runtime-service dependencies

adutra:
Project ":polaris-relational-jdbc" should be included like all other Polaris modules.

Dependency "io.quarkus:quarkus-jdbc-postgresql" doesn't need any excludes (these excludes were for `SparkIT` which is now isolated in a separate module).

### DIFF
--- a/runtime/service/build.gradle.kts
+++ b/runtime/service/build.gradle.kts
@@ -31,6 +31,8 @@ dependencies {
   implementation(project(":polaris-api-iceberg-service"))
   implementation(project(":polaris-api-catalog-service"))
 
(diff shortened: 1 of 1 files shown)


snazy:
> Project ":polaris-relational-jdbc" should be included like all other Polaris modules in "implementation" configuration.

But it's `runtimeOnly` ;)

adutra:
> > Project ":polaris-relational-jdbc" should be included like all other Polaris modules in "implementation" configuration.
> 
> But it's `runtimeOnly` ;)

Hehe I changed my mind in between. Probably safer to keep this one in `runtimeOnly`.

snazy (approved):

PR #2275, opened 2025-08-06
Use Mockito Java agent for mock instrumentation

adutra:
This change fixes the following warning during tests:

    Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
    WARNING: A Java agent has been loaded dynamically (.../byte-buddy-agent-1.17.5.jar)
    WARNING: If a serviceability tool is in use, please run with -XX:+EnableDynamicAgentLoading to hide this warning
    WARNING: If a serviceability tool is not in use, please run with -Djdk.instrument.traceUsage for more information
    WARNING: Dynamic loading of agents will be disallowed by default in a future release


build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -153,18 +155,22 @@ dependencies {
...
+
+  mockitoAgent(mockitoCoreLib) { isTransitive = false }
 }
 
 tasks.withType<Test>().configureEach {
   systemProperty("file.encoding", "UTF-8")
   systemProperty("user.language", "en")
   systemProperty("user.country", "US")
   systemProperty("user.variant", "")
+  jvmArgs("-javaagent:${mockitoAgent.asPath}")
```
snazy: Should use this to let the system specific path not pollute the build cache.
```suggestion
    jvmArgumentProviders.add( CommandLineArgumentProvider {  listOf("-javaagent:${mockitoAgent.asPath}") })
```

PR #2276, opened 2025-08-06
Remove legacy management endpoints

adutra:


adutra:
@flyrain FYI

snazy (approved):

flyrain:
Hi @adutra , thanks for tagging me. Will take a look soon.

adutra:
Putting this PR on hold for now to give users more time to adjust to the new endpoints.

snazy:
I think it's time to revive this PR as there was no feedback in the last two weeks, let's get it in.

adutra:
@flyrain @singhpk234 would it be OK to include this in 1.2.0?

dimas-b:
Should we make it "ready for review"? :sweat_smile:

dimas-b:
+1 to including into 1.2.0

flyrain:
Can we remove it at 2.0? I'm a bit paranoid of removing endpoints between minor versions.

dimas-b:
Endpoints on the management (Quarkus) interface are not formally considered public API, I guess (they are not callable by clients, only by infra components within the "control plane"), so it should be ok to drop old ones in a minor release, I think.

Admin users should be able to transition their monitoring solutions to the new endpoints in 1.1.0 (where both sets are supported) so upgrading to a release where the old endpoints are dropped would be transparent.

This is a different use case than supporting older clients, which may run multiple old versions at the same time.

However, it's a bit of a grey area :thinking:

singhpk234:
I would also request if possible lets move the removal to 2.0, mostly because I am not fully aware of how `/metrics` response of this api is different from the quarkus metrics if they a lot different then the dashboards, alerting made on top of these metrics etc needs to be adapted accordingly which can take sometime.

adutra:
> I am not fully aware of how /metrics response of this api is different from the quarkus metrics if they a lot different

It's exactly the same endpoint. It's just a mirror of `/q/metrics`.

snazy:
@adutra as 1.2.0 isn't yet "frozen", WDYT of deprecating these in 1.2.0 and removing in 1.3.0?

adutra:
> @adutra as 1.2.0 isn't yet "frozen", WDYT of deprecating these in 1.2.0 and removing in 1.3.0?

WFM, but how do we deprecate an endpoint? Log a warning when it's accessed the first time?

dimas-b:
> how do we deprecate an endpoint

A CHANGELOG notice is fine from my POV.

adutra:
> > how do we deprecate an endpoint
> 
> A CHANGELOG notice is fine from my POV.

There you go: #2749

adutra:
Now that 1.2.0 is out, I'm moving this PR to ready state.

dimas-b:
@fabio-rizzo-01 @fivetran-ashokborra @sungwy FYI

snazy:
Now that 1.2.0 is (nearly) out, I think it's fine to merge this PR for 1.3.

dimas-b:
Let's merge right after 1.2.0 is released :)

adutra:
Are we OK to merge this now?

PR #2277, opened 2025-08-06
Remove noop code in persistence

snazy:
The `persistStorageIntegrationIfNeeded` function does nothing in any implementation, therefore the corresponding load-function cannot yield anything ever. This change removes both functions. That allows removing the `createStorageIntegration` function as well and let its only call site `getSubscopedCredsForEntity` be implemented once and directly call the storage integration provider.

Being able to access object storage without access to entities is a requirement for all currently active tasks proposals. This is one step towards that goal.

A follow-up will entirely remove the storage-integration-provider dependency from persistence and let `StorageCredentialCache` deal with it as the only call site.


dennishuo (changes requested):
Thanks for taking a stab at refactoring!

Unfortunately, as it currently stands, this will be problematic for anyone with custom implementations of the `IntegrationPersistence` interface in which StorageIntegrations make use of the offered `persistStorageIntegrationIfNeeded` as declared by the public interface to manage custom persistent state that lives outside of StorageConfigurationInfo.

We can probably still figure out a clean way to migrate this equivalent functionality to your proposed decoupled model for async tasks, but at least one large production deployment of Polaris is known to heavily depend on the `IntegrationPersistence` methods in order to provide the layer of indirection that allows runtime binding of *different* storage integration credentials to different objects within a shared multi-tenant environment.

Is there an mailing list thread that discusses the goal of "accessing object storage without access to entities", or is this just part of the overall "Asynchronous and Reliable Tasks" thread? Can we link the ML thread in this PR description for easy access?

PR #2280, opened 2025-08-06
Add support for S3 request signing

adutra:
WIP


singhpk234:
Thank you so much for this @adutra ! 

when ready please let us know (would be really helpful if you can share some description to walk us through your thought process), happy to help with reviews :) !

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -797,4 +812,24 @@ public Response getConfig(
                 .build())
         .build();
   }
+
+  @Override
+  public Response signS3Request(
```
snazy: I think it makes sense to have this in a separate class to avoid the overhead of having unneeded things just for signing purposes. We can emit a separate signing-endpoint in a load-table, if necessary.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,154 @@
...
+import software.amazon.awssdk.http.auth.aws.signer.AwsV4HttpSigner;
+import software.amazon.awssdk.http.auth.spi.signer.SignRequest;
+import software.amazon.awssdk.http.auth.spi.signer.SignedRequest;
+import software.amazon.awssdk.services.s3.S3Client;
+
+/**
+ * S3 request signer that creates presigned URLs for S3 operations using AWS credentials obtained
+ * from the storage integration.
+ */
+public class S3RequestSignerImpl implements S3RequestSigner {
```
snazy:
```suggestion
class S3RequestSignerImpl implements S3RequestSigner {
```

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,154 @@
...
+
+  private final AwsV4HttpSigner signer = AwsV4HttpSigner.create();
+
+  private final AwsCredentialsStorageIntegration storageIntegration;
+  private final StorageCredentialCache storageCredentialCache;
+
+  public S3RequestSignerImpl(
```
snazy:
```suggestion
  S3RequestSignerImpl(
```

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    String body = signingRequest.body();
+
+    SdkHttpFullRequest.Builder requestToSign =
+        SdkHttpFullRequest.builder()
+            .uri(uri)
+            .protocol(uri.getScheme())
+            .method(method)
+            .headers(signingRequest.headers());
+
+    // FIXME is this correct?
```
snazy: nope.
See [ListObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html) + [ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) REST spec.

*Be careful:* Not all `GET /{bucket/` requests are list-objects requests!
The "v1" list-objects has no required parameter, so you cannot just check for the presence of a req param. It could also be a get-bucket-cors or get-bucket-encryption or ...-policy or ...-website or ......
Nice, no?

_Maybe_ it's okay to assume that every client uses the v2 api, which has a unique, required request param.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,154 @@
...
+      RealmConfig realmConfig,
+      AwsStorageConfigurationInfo storageConfig,
+      S3SignRequest signingRequest) {
+
+    SdkHttpMethod method = SdkHttpMethod.valueOf(signingRequest.method());
+    URI uri = signingRequest.uri();
+    String body = signingRequest.body();
+
+    SdkHttpFullRequest.Builder requestToSign =
+        SdkHttpFullRequest.builder()
```
snazy: also need the request parameters here

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,154 @@
...
+            .protocol(uri.getScheme())
+            .method(method)
+            .headers(signingRequest.headers());
+
+    // FIXME is this correct?
+    boolean allowListOperation = isReadMethod(method);
+    Set<String> allowedReadLocations = isReadMethod(method) ? Set.of(uri.toString()) : Set.of();
+    Set<String> allowedWriteLocations = isWriteMethod(method) ? Set.of(uri.toString()) : Set.of();
+
+    // TODO use the storageCredentialCache to get or generate subscoped credentials
```
snazy: how? request-signing is (mostly) used in environments that don't have sts.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    // AccessConfig accessConfig1 = storageCredentialCache.getOrGenerateSubScopeCreds();
+
+    AccessConfig accessConfig =
+        storageIntegration.getSubscopedCreds(
+            realmConfig, allowListOperation, allowedReadLocations, allowedWriteLocations);
+
+    AwsCredentials credentials = extractCredentials(accessConfig);
+
+    // FIXME is this correct?
```
snazy: clients may request a different region (hello path-style vs hostname bucket resolution)

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
adutra: Aren't they included in the request URI?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
adutra: Yep I gave up on the idea, the code has changed significantly.

adutra:
Ready for review. The test failures are unrelated to this change.

\cc @singhpk234 @snazy @dimas-b @metadaddy

api/iceberg-aws-sign-service/build.gradle.kts
```diff
@@ -0,0 +1,125 @@
...
+  additionalProperties.put("apiNameSuffix", "")
+  additionalProperties.put("metricsPrefix", "polaris")
+  serverVariables.put("basePath", "api/catalog")
+  typeMappings =
+    mapOf("S3SignRequest" to "org.apache.polaris.service.aws.sign.model.PolarisS3SignRequest")
+  importMappings =
+    mapOf(
+      "IcebergErrorResponse" to "org.apache.iceberg.rest.responses.ErrorResponse",
+      "S3SignRequest" to "org.apache.polaris.service.aws.sign.model.PolarisS3SignRequest",
+      "SignS3Request200Response" to
```
adutra: FYI The name of the logical response type is a bit weird because it is inlined, instead of being a `$ref` in the OpanAPI spec.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
```diff
@@ -146,11 +148,13 @@ public abstract class PolarisRestCatalogIntegrationBase extends CatalogTests<RES
   private RESTCatalog restCatalog;
   private String currentCatalogName;
   private Map<String, String> restCatalogConfig;
-  private URI externalCatalogBase;
+  private URI externalCatalogBaseLocation;
```
adutra, line 159: FYI I refactored this class a bit, notably to facilitate subclassing, but there is no functional change in this class.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisS3RemoteSigningIntegrationTest.java
```diff
@@ -0,0 +1,132 @@
...
+      } finally {
+        resolvingFileIO.deleteFile(fileLocation);
+      }
+    }
+  }
+
+  @Test
+  @Override
+  @Disabled("It's not possible to request an access delegation mode when registering a table.")
+  public void testRegisterTable() {
```
adutra, line 194: FYI what is happening here is:

As per the Iceberg REST spec, the register-table endpoint does not accept access delegation headers.

As a consequence, the returned `LoadTableResult` does NOT contain any access delegation enablement, e.g. it won't contain vended credentials nor remote signing enablement.

I'm not sure if this is an issue or something that engines know how to deal with.

runtime/server/build.gradle.kts
```diff
@@ -39,9 +39,6 @@ val distributionElements by
   }
 
 dependencies {
-  implementation(project(":polaris-core"))
```
adutra: Not directly referenced in this module.

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
```diff
@@ -86,7 +85,6 @@
 @QuarkusIntegrationTest
 @TestProfile(RestCatalogMinIOSpecialIT.Profile.class)
 @ExtendWith(MinioExtension.class)
-@ExtendWith(PolarisIntegrationTestExtension.class)
```
adutra, line 89: This extension should not be enabled in IT tests annotated with `@QuarkusIntegrationTest`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -297,12 +298,6 @@ public CreateNamespaceResponse createNamespace(CreateNamespaceRequest request) {
     }
   }
 
-  private static boolean isStaticFacade(CatalogEntity catalog) {
```
adutra: Refactored to become a method of `CatalogEntity`.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
```diff
@@ -72,21 +71,19 @@ public interface StorageConfiguration extends S3AccessConfig {
...
-  default AwsCredentialsProvider stsCredentials() {
+  /**
+   * Returns an {@link AwsCredentialsProvider} that provides system-wide AWS credentials. If both
+   * access key and secret key are present, it uses them directly; otherwise, it uses the default
+   * credentials provider chain.
+   *
+   * <p>The returned provider is not meant to be vended directly to clients, but rather used with
+   * STS, unless credential subscoping is disabled.
+   */
+  default AwsCredentialsProvider awsSystemCredentials() {
```
adutra: Renamed to `awsSystemCredentials` because these are not "just" credentials for STS, they are server credentials.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+      RealmContext realmContext,
+      SecurityContext securityContext) {
+    Namespace ns = decodeNamespace(namespace);
+    TableIdentifier tableIdentifier = TableIdentifier.of(ns, RESTUtil.decodeString(table));
+    return withCatalog(
+        securityContext,
+        prefix,
+        catalog -> {
+          S3SignResponse response = catalog.signS3Request(s3SignRequest, tableIdentifier);
+          return Response.status(Response.Status.OK).entity(response).build();
```
adutra, line 105: In theory the response should include Cache-Control headers. This is left for a later improvement.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public PolarisS3SignResponse signS3Request(
+      PolarisS3SignRequest s3SignRequest, TableIdentifier tableIdentifier) {
+
+    LOGGER.debug("Requesting s3 signing for {}: {}", tableIdentifier, s3SignRequest);
+
+    // TODO authorize based on the request's method?
+
+    try {
+      authorizeBasicTableLikeOperationOrThrow(
```
adutra: The RBAC checks are admittedly very coarse-grained for now.

adutra:
Re: integration test failure. The test passes locally. There is some classloading snafu going on, I will investigate later:

```
Caused by: java.lang.IllegalStateException: PolarisServerManager not found
	at org.apache.polaris.service.it.ext.PolarisServerManagerLoader.lambda$static$0(PolarisServerManagerLoader.java:28)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.apache.polaris.service.it.ext.PolarisServerManagerLoader.&lt;clinit&gt;(PolarisServerManagerLoader.java:28)
	... 4 more
```

**EDIT**

Problem solved.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/DefaultFileIOFactory.java
```diff
@@ -101,6 +110,32 @@ public FileIO loadFileIO(
       properties.putAll(accessConfig.get().credentials());
       properties.putAll(accessConfig.get().extraProperties());
       properties.putAll(accessConfig.get().internalProperties());
+    } else {
+      // If no subscoped creds were produced, use system-wide AWS or GCP credentials if available.
```
adutra: This bit is not directly related to remote signing, but I realized that, when credentials subscoping is disabled, the server is unable to use the credentials configured in application.properties.

This change allows that.

That being said, the server would still be unable to use e.g. a custom S3 endpoint, simply because this property is not exposed in application.properties.

IMHO each property that can be configured in `AccessConfig` for a specific catalog should also be configurable by a system-wide property in application.properties.

metadaddy:
Looks good. When I get an opportunity, I'll try setting this up with Backblaze B2 as S3-compatible storage and Trino as the Iceberg REST catalog client. Thanks!

api/iceberg-aws-sign-service/src/main/java/org/apache/polaris/service/aws/sign/model/PolarisS3SignRequest.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * <p>Copy of {@link S3SignRequest}, because the original does not have Jackson annotations.
+ */
+@PolarisImmutable
+@JsonDeserialize(as = ImmutablePolarisS3SignRequest.class)
+@JsonSerialize(as = ImmutablePolarisS3SignRequest.class)
+public interface PolarisS3SignRequest extends S3SignRequest {
+
+  @Value.Default
+  @Nullable // Replace javax.annotation.Nullable with jakarta.annotation.Nullable
```
snazy:
```suggestion
  @Nullable // Replace javax.annotation.Nullable from 'S3SignRequest` with jakarta.annotation.Nullable
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisS3RemoteSigningIntegrationTest.java
snazy: I _think_ it's worth to try out whether the register-table endpoint receives the `X-Iceberg-Access-Delegation` header. In Nessie, we accept that header in the register-table endpoint and haven't ignored this test.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -349,4 +349,20 @@ public static void enforceFeatureEnabledOrThrow(
                   + "it is still possible to enforce the uniqueness of table locations within a catalog.")
           .defaultValue(false)
           .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Boolean> REMOTE_SIGNING_ENABLED =
+      PolarisConfiguration.<Boolean>builder()
+          .key("REMOTE_SIGNING_ENABLED")
+          .catalogConfig("polaris.config.remote-signing.enabled")
+          .description(
+              "If true, the remote signing endpoints are enabled either globally, or for a specific catalog.")
```
snazy: Maybe add a note that the feature is experimental, elaborating why on the `CHANGELOG.md?

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -205,11 +206,19 @@ public Catalog.TypeEnum getCatalogType() {
         .orElse(null);
   }
 
+  public boolean isExternal() {
+    return Objects.equals(getCatalogType(), Catalog.TypeEnum.EXTERNAL);
```
snazy:
```suggestion
    return getCatalogType() == Catalog.TypeEnum.EXTERNAL;
```
enum values have object-identity ;)

polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisEndpoints.java
```diff
@@ -94,4 +101,18 @@ public static Set<Endpoint> getSupportedPolicyEndpoints(CallContext callContext)
...
+  /**
+   * Get the remote signing endpoints. Returns {@link #REMOTE_SIGNING_ENDPOINTS} if {@link
+   * FeatureConfiguration#REMOTE_SIGNING_ENABLED} is set globally to true or if the catalog enables
+   * remote signing; otherwise, returns an empty set.
+   */
+  public static Set<Endpoint> getSupportedRemoteSigningEndpoints(
+      CallContext callContext, CatalogEntity catalogEntity) {
+    boolean remoteSigningEnabled =
+        callContext
+            .getRealmConfig()
```
snazy:
```suggestion
      RealmConfig realmConfig, CatalogEntity catalogEntity) {
    boolean remoteSigningEnabled =
        realmConfig
```

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -39,6 +42,21 @@ public enum StorageAccessProperty {
       Boolean.class, "s3.path-style-access", "whether to use S3 path style access", false),
   CLIENT_REGION(
       String.class, "client.region", "region to configure client for making requests to AWS"),
+  AWS_REMOTE_SIGNING_ENABLED(
+      Boolean.class,
+      S3FileIOProperties.REMOTE_SIGNING_ENABLED,
+      "whether to enable remote signing for S3 requests",
```
snazy, line 60: Maybe add some text saying that this depends on the "global" flag.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -866,13 +888,52 @@ public AccessConfig getAccessConfig(
...
+  }
+
+  @Override
+  public AccessConfig getAccessConfigForRemoteSigning(TableIdentifier tableIdentifier) {
+
+    Optional<PolarisStorageConfigurationInfo> configurationInfo =
+        findStorageInfo(tableIdentifier)
+            .map(
+                info ->
+                    deserialize(
```
snazy: Nit: move the deserialze() to a separate .map() for readability.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -710,30 +697,26 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
...
+    CatalogEntity catalogEntity = getCatalogEntity();
+
+    if (delegationModes.contains(AccessDelegationMode.VENDED_CREDENTIALS)) {
+
+      LOGGER.info(
+          "allow external catalog credential vending: {}",
+          callContext
+              .getRealmConfig()
+              .getConfig(
+                  FeatureConfiguration.ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING, catalogEntity));
```
snazy: Nit: extract this value (duplicated) to a variable.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -774,21 +757,55 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
...
+      Map<String, String> credentialConfig = accessConfig.credentials();
+      responseBuilder.addAllConfig(credentialConfig);
+      responseBuilder.addAllConfig(accessConfig.extraProperties());
+      if (!credentialConfig.isEmpty()) {
+        responseBuilder.addCredential(
+            ImmutableCredential.builder()
+                .prefix(tableMetadata.location())
+                .config(credentialConfig)
+                .build());
+      }
```
snazy: Looks like this part is duplicated. Any chance to remove the duplciated code?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogAdapter.java
snazy: ... but that would require clients to respect those 🤷

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,118 @@
...
+          PolarisAuthorizableOperation.SIGN_S3_REQUEST,
+          PolarisEntitySubType.ICEBERG_TABLE,
+          tableIdentifier);
+    } catch (NoSuchTableException e) {
+      // FIXME can we do better here?
+      authorizeCreateTableLikeUnderNamespaceOperationOrThrow(
+          PolarisAuthorizableOperation.SIGN_S3_REQUEST, tableIdentifier);
+    }
+
+    throwIfRemoteSigningNotEnabled(callContext.getRealmConfig(), catalogEntity);
```
snazy, line 99: Nit: move before the `try {`

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/PolarisCallContextCatalogFactory.java
```diff
@@ -92,7 +104,10 @@ public Catalog createCallContextCatalog(
             securityContext,
             taskExecutor,
             fileIOFactory,
-            polarisEventListener);
+            polarisEventListener,
+            storageIntegrationProvider,
+            prefixParser,
+            uriInfo);
```
snazy: (Not related to this PR)
The consistently increasing amount of parameters passed to these types worries me.

snazy (commented):
Overall LGTM.
Not an "always boring" change" though - therefore a few more comments.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisS3RemoteSigningIntegrationTest.java
adutra: Created an issue: https://github.com/apache/polaris/issues/2351

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
adutra: Actually I don't think that's how it works. Looking at this logic:

https://github.com/apache/polaris/blob/15f23cae8e678f8a13c20c5012919565288525a0/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java#L116-L142

It seems that remote signing could be _globally_ disabled, but _enabled_ for a specific catalog.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/PolarisCallContextCatalogFactory.java
adutra: Agreed. We could look into making it a CDI bean.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
adutra: Actually, not possible: you need to perform RBAC first, so that the catalog entity is fetched.

api/iceberg-aws-sign-service/src/main/java/org/apache/polaris/service/aws/sign/model/PolarisS3SignRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import jakarta.annotation.Nullable;
+import org.apache.iceberg.aws.s3.signer.S3SignRequest;
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
+
+/**
+ * Request for S3 signing requests.
```
singhpk234:
```suggestion
 * Request for S3 signing.
```

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
snazy: Yea - so you can enable it, but it mustn't be disabled globally. Anyway, not a must-have.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
snazy: hm - probably worth to mention in the changelog

snazy (commented):
I think, it's very very close :)

singhpk234:
Hey @adutra thank you the change, i was reading the description and it turns out we are introducing new priviledges, IMHO this requires broader set of eyes and feedbacks as its just no longer implementation of the specification of IRC which can be contained in the PR.

Can you please send your Proposal in the dev-list (if you have already have one, can you please link this to the PR)
I only see this thread presently - https://lists.apache.org/thread/qvzwc3qxlfrk9vr7yfbx6zxfhz9lhlbc

snazy (approved):

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -143,6 +143,37 @@ public AccessConfig getSubscopedCreds(
...
+      accessConfig.put(StorageAccessProperty.CLIENT_REGION, region);
+    }
+
+    URI endpointUri = storageConfig.getEndpointUri();
+    if (endpointUri != null) {
+      accessConfig.put(StorageAccessProperty.AWS_ENDPOINT, endpointUri.toString());
+    }
+    if (Boolean.TRUE.equals(storageConfig.getPathStyleAccess())) {
+      accessConfig.put(StorageAccessProperty.AWS_PATH_STYLE_ACCESS, Boolean.TRUE.toString());
+    }
```
singhpk234: can we put these in a helper, to be used here and the method above ?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,119 @@
...
+  protected void initializeCatalog() {
+    catalogEntity =
+        CatalogEntity.of(resolutionManifest.getResolvedReferenceCatalogEntity().getRawLeafEntity());
+    if (catalogEntity.isExternal()) {
+      throw new ForbiddenException("Cannot use S3 remote signing with federated catalogs.");
+    }
+    // no need to materialize the catalog here, as we only need the catalog entity
+  }
+
+  public PolarisS3SignResponse signS3Request(
```
singhpk234, line 87: can i have a sign priviledge without table read priviledge ? a client may directly contact the sign endpoint

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,119 @@
...
+      PolarisS3SignRequest s3SignRequest, TableIdentifier tableIdentifier) {
+
+    LOGGER.debug("Requesting s3 signing for {}: {}", tableIdentifier, s3SignRequest);
+
+    // TODO authorize based on the request's method?
```
singhpk234: shouldn't we ?  SIGN_S3_REQUEST shouldn't be a blanket approval for deletes ?

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RequestSignerImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    SdkHttpMethod method = SdkHttpMethod.valueOf(signingRequest.method());
+
+    SdkHttpFullRequest.Builder requestToSign =
+        SdkHttpFullRequest.builder()
+            .uri(uri)
+            .protocol(uri.getScheme())
+            .method(method)
+            .headers(signingRequest.headers());
+
+    AwsCredentials credentials = storageConfiguration.awsSystemCredentials().resolveCredentials();
```
singhpk234: How are the credentials refreshed ?

adutra: These are the server credentials – they must not need refreshing.

adutra:
FYI: I was requested to produce a design doc for this feature:

https://docs.google.com/document/d/1ygdia7u4bUHUt6n8XhZo48aKoIyyrCvKqan3XP25iB8/edit?usp=sharing

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
RussellSpitzer: Minor request here, but you could pull this and the new IntegerationTest Helpers out into a separate PR and reduce the size of this PR by a bit. I think that would also be a very fast review

runtime/server/build.gradle.kts
RussellSpitzer: This is another change I would try to pull out to another PR, it's not related to this changeset correct?

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
adutra: Yup good idea!

https://github.com/apache/polaris/pull/2384

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,111 @@
...
+        s3SignRequest.write()
+            ? PolarisAuthorizableOperation.SIGN_S3_WRITE_REQUEST
+            : PolarisAuthorizableOperation.SIGN_S3_READ_REQUEST;
+
+    authorizeRemoteSigningOperationOrThrow(authzOp, tableIdentifier);
+
+    // Must be done after the authorization check, as the auth check creates the catalog entity
+    throwIfRemoteSigningNotEnabled(callContext.getRealmConfig(), catalogEntity);
+
+    // TODO S3 location access checks
```
adutra: FYI, this is left for a follow-up PR.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -834,13 +855,50 @@ public AccessConfig getAccessConfig(
...
+    if (!(storageIntegration instanceof AwsCredentialsStorageIntegration)) {
+      LOGGER
+          .atWarn()
+          .addKeyValue("tableIdentifier", tableIdentifier)
+          .log("Table entity storage integration is not an AWS credentials storage integration");
+      return AccessConfig.builder().build();
+    }
+
+    String prefix = prefixParser.catalogNameToPrefix(callContext.getRealmContext(), catalogName);
+    URI signerUri = uriInfo.getBaseUri().resolve("api/");
```
adutra: FYI, resolving the signer URI in case of HTTP proxies is left for a follow-up PR.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
adutra: Not anymore, I refactored this bit to comply with the design doc. 

For a read request you need `TABLE_REMOTE_SIGN` + `TABLE_READ_DATA`.

For a write request you need `TABLE_REMOTE_SIGN` + `TABLE_WRITE_DATA`.

runtime/service/src/main/java/org/apache/polaris/service/storage/aws/signer/S3RemoteSigningCatalogHandler.java
adutra: See my previous comment. For a `DELETE` method you now need `TABLE_REMOTE_SIGN` + `TABLE_WRITE_DATA`.

runtime/server/build.gradle.kts
adutra: Correct:

https://github.com/apache/polaris/pull/2385

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: @RussellSpitzer FYI:

#2386
#2387

adutra:
FYI I just rebased and synchronized this PR with the design doc.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
RussellSpitzer: Thanks! It definitely helps to keep this sort of refactor out of the larger functional change

adutra:
Switching back to draft, sorry.

I finally managed to adapt this PR to the AccessConfig refactoring, but I still need to solve a few FIXMEs and wrap my head around the new design.

adutra:
This is now ready again for review!

\cc @snazy @singhpk234 @flyrain @dimas-b

This PR is equivalent to Milestone 1 in the original design doc. Let's try to work incrementally and get this in, we are already working on getting a new PR for Milestone 2 and especially the authz checks that are not yet implemented. Thanks!

singhpk234:
Hey Alex, 
Thank a ton for all the work !
Is this understanding of M1 and M2 correct ? (take design doc above)
```
M1 - Simple, slow implementation with basic RBAC
- Not production ready – marked experimental
- Does not implement access checks based on S3 location
- Does not support HTTP proxies
- Corresponds to the current PR: https://github.com/apache/polaris/pull/2280 

M2 - Fast implementation with efficient RBAC
- Production-ready
- Implements [Access checks based on the S3 location](https://docs.google.com/document/d/1ygdia7u4bUHUt6n8XhZo48aKoIyyrCvKqan3XP25iB8/edit?tab=t.0#bookmark=id.hs05149qp3eo) 
- Supports HTTP proxies
- Implements "Nessie style" signed parameters
```

If yes, I am bit confused, if we know M2 is the one we eventually wanna make and M2 is a bit orthogonal to M1 why are making / investing on M1, a couple questions that would be really helpful to get your take : 
1. How long it will take for us to get M2 post M1 ? (you mentioned M2 is already in works ?)
2. why are we making M1 if M1 is not production recommended and based on my understanding it will never be.
3. when both M2 and M1 are there when will a user use M1 over M2 ?
4. The route for M1 is `/s3-sign/v1/{prefix}/namespaces/{namespace}/tables/{table}` and for M2 its gonna be 
`/s3-sign/v1/{prefix}/namespaces/{namespace}/tables/{table}/{signContext}` or `/s3-sign/v1/{prefix}/namespaces/{namespace}/tables/{table}?signContext={signedContext}` checking this commit (please correct me If i am missing something) https://github.com/projectnessie/nessie/commit/2c6b9313c09adb0b584f6aeb6c97d94791fcef28 it will be earlier ? are we gonna introduce a new endpoint for M2 ?

adutra:
> If yes, I am bit confused, if we know M2 is the one we eventually wanna make and M2 is a bit orthogonal to M1 why are making / investing on M1, a couple questions that would be really helpful to get your take :

It's not orthogonal, M2 builds on top of M1. 

> 1. How long it will take for us to get M2 post M1 ? (you mentioned M2 is already in works ?)

It depends if we want all the features in M2 or not. I would actually personally break M2 into smaller pieces. The most important piece is "Access checks based on the S3 location", which can be done in a few man-days. The "Nessie style" signed parameters would require a bit more than that, imo.

> 2. why are we making M1 if M1 is not production recommended and based on my understanding it will never be.

For the same reason we released the Events feature in incomplete state, with "beta" status: to get users feedback, and to work incrementally.

> 3. when both M2 and M1 are there when will a user use M1 over M2 ?

These are not "flavors" of remote signing. It's just one same feature, in two levels of maturity M1: experimental, M2: production-ready.

> are we gonna introduce a new endpoint for M2 ?

Yes, if we adopt Nessie's strategy for fast remote signing as proposed in the doc (it's optional). Note: this isn't a breaking change, as it's not user facing.

Happy to discuss this topic more in depth in a Slack channel or during the next sync!

dimas-b:
The proposed phased approach (M1/M2) SGTM :+1:

api/s3-sign-service/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
...
+  implementation(libs.jakarta.annotation.api)
+  implementation(libs.jakarta.inject.api)
+  implementation(libs.jakarta.validation.api)
+  implementation(libs.swagger.annotations)
+
+  implementation(libs.jakarta.servlet.api)
+  implementation(libs.jakarta.ws.rs.api)
+
+  implementation(platform(libs.micrometer.bom))
+  implementation("io.micrometer:micrometer-core")
```
dimas-b: Is `micrometer` actually used in this module?

spec/README.md
```diff
@@ -17,14 +17,40 @@
...
+      the Notifications API
+
+    - [oauth-tokens-api.yaml](polaris-catalog-apis/oauth-tokens-api.yaml) - Contains the specification for the 
+      internal OAuth Token endpoint, extracted from the Apache Iceberg REST Catalog API.
+
+    - [policy-apis.yaml](polaris-catalog-apis/policy-apis.yaml) - Contains the specification for the Policy APIs.
+
+- [s3-sign](s3-sign) - This folder contains the specifications for S3 remote signing:
+
+  - [iceberg-s3-signer-open-api.yaml](s3-sign/iceberg-s3-signer-open-api.yaml) - Contains the specification of the
```
dimas-b, line 49: nit: maybe clarify that Polaris uses only type definitions from the spec file?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -828,6 +830,9 @@ public Response getConfig(
                         .addAll(VIEW_ENDPOINTS)
                         .addAll(PolarisEndpoints.getSupportedGenericTableEndpoints(realmConfig))
                         .addAll(PolarisEndpoints.getSupportedPolicyEndpoints(realmConfig))
+                        .addAll(
+                            PolarisEndpoints.getSupportedRemoteSigningEndpoints(
```
dimas-b: I'm not sure it's worth advertising in `ConfigResponse` because it's not a standard endpoint, for which clients might want to check the endpoints list. Clients learn about it via `loadTable` responses, I guess 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/AccessConfigProvider.java
```diff
@@ -101,4 +120,58 @@ public AccessConfig getAccessConfig(
...
+        instanceof AwsCredentialsStorageIntegration awsCredentialsStorageIntegration)) {
+      LOGGER
+          .atWarn()
+          .addKeyValue("tableIdentifier", tableIdentifier)
+          .log("Table entity storage integration is not an AWS credentials storage integration");
+      return AccessConfig.EMPTY;
+    }
+
+    String prefix = prefixParser.catalogNameToPrefix(callContext.getRealmContext(), catalogName);
+    URI signerUri = uriInfo.getBaseUri().resolve("api/");
```
dimas-b: I'm not sure about the `api/` constant... It related to our `@Path` annotations in the generated REST API classes, so it's usage is a bit obscure here. Could we make it a constant in the `s3-sign-service` module?

polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
```diff
@@ -78,4 +82,15 @@ public String genericTable(TableIdentifier ident) {
         "generic-tables",
         RESTUtil.encodeString(ident.name()));
   }
+
+  public String s3RemoteSigning(TableIdentifier ident) {
```
dimas-b, line 88: Is this a `core` concern? I suppose it is only relevant to the REST layer 🤔

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -52,6 +54,21 @@ public enum StorageAccessProperty {
...
       false),
+  AWS_REMOTE_SIGNING_ENABLED(
+      Boolean.class,
+      S3FileIOProperties.REMOTE_SIGNING_ENABLED,
+      "whether to enable remote signing for S3 requests",
+      false),
+  AWS_REMOTE_SIGNER_URI(
+      String.class,
+      S3V4RestSignerClient.S3_SIGNER_URI,
+      "the base URI for the remote signer service, used for signing S3 requests",
```
dimas-b, line 65: Does this have a reference Iceberg version number where the property becomes respected?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -801,6 +806,33 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
...
+
+      AccessConfig accessConfig =
+          accessConfigProvider.getAccessConfigForRemoteSigning(
+              callContext, catalogName, tableIdentifier, resolvedStoragePath);
+
+      Map<String, String> credentialConfig = accessConfig.credentials();
+
+      if (!credentialConfig.isEmpty()) {
+        responseBuilder.addAllConfig(credentialConfig);
+        responseBuilder.addCredential(
```
dimas-b: Why `addCredential()` when we're remote-signing?

dimas-b (commented):
Lots of comments 🙂 but the PR overall looks good 👍 Thanks for implementing this, @adutra !

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/AccessConfigProvider.java
adutra: I agree, but I'm not sure a constant in `s3-sign-service` is the best solution. The `api/` path segment doesn't seem to be a first-class citizen in our APIs, rather a bi-product. It's only declared as a server variable in the OpenAPI generator config, for example:

https://github.com/apache/polaris/blob/1f4e748a96b926cd1639d6ed463997d2f9c455c4/api/iceberg-service/build.gradle.kts#L86

https://github.com/apache/polaris/blob/b6e247deda8a618136ec16c75b86eb2a2d12642b/api/polaris-catalog-service/build.gradle.kts#L116

The management APi is actually a bit different, there is not `basePath` variable in polaris-management-service.yml, so the base path is hard-coded:

https://github.com/apache/polaris/blob/04c8ee6972df948beb15401e8dc905a37b27c33d/spec/polaris-management-service.yml#L26

But the common characteristic is that `api/`is the first path segment of all Polaris APIs, but that is not enforced by any constraint.

I would say the right place for this segment would be in `PolarisResourcePaths`, but you will notice again that none of the paths include the `api/` segment. Same for Iceberg's `ResourcePaths`. So it's really implicit 😅 

All of this to say, let's declare a constant in `PolarisResourcePaths` and call it a day, wdyt?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
adutra: OK let's remove.

polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
adutra: Well the same could be said for all the methods in this class 😄 
The package is `org.apache.polaris.core.rest` btw - so I guess REST is a core concern 🤷‍♂️ 

I think `PolarisResourcePaths` and `PolarisEndpoints` should not live in `core` ideally. But that's the way it is today.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
adutra: Iceberg 1.2.0. That's more than 2 years old, I'm not sure we need to mention it here. Wdyt?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: Good catch, this doesn't make sense 😄

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/AccessConfigProvider.java
dimas-b: +1 to `PolarisResourcePaths`

polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
dimas-b: oops, I did not actually look at the whole class, only at the diff :sweat_smile:

Let's keep it for now and maybe refactor later.... In general, I'd like to untangle core from REST :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
dimas-b: That's fine... I was worries it might have been more recent ;)

dimas-b (approved):

singhpk234:
Hey Alex, 
Thanks for the response, TBH I am still  unclear why do we wanna introduce an endpoint for M1 only to introduce a new one in M2 ? How can an **public** endpoint not be a user facing change ? 
Apologies but I am not able to follow, As you suggested, definitely sync will be really helpful to discuss, meanwhile i request you to please bump this in the thread https://lists.apache.org/thread/xpf60ko823c7mlg5bhnddq3btvdsq5s1 too for more eyes, in case I am missing something.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -394,6 +394,47 @@ protected void authorizeRenameTableLikeOperationOrThrow(
...
+  protected void authorizeRemoteSigningOrThrow(
+      EnumSet<PolarisAuthorizableOperation> ops, TableIdentifier identifier) {
+
+    ensureResolutionManifestForTable(identifier);
+    PolarisResolvedPathWrapper target =
+        resolutionManifest.getResolvedPath(
+            identifier, PolarisEntityType.TABLE_LIKE, PolarisEntitySubType.ICEBERG_TABLE, true);
+
+    // If the table doesn't exist, we still need to check allowed locations from the parent
+    // namespace.
```
singhpk234, line 406: [doubt] is this a valid state ?

CHANGELOG.md
```diff
@@ -36,6 +36,14 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 [Iceberg Metrics Reporting]: https://iceberg.apache.org/docs/latest/metrics-reporting/
 
+- **S3 remote request signing** has been added, allowing Polaris to work with S3-compatible object storage systems.
+  *Remote signing is currently experimental and not enabled by default*. In particular, RBAC checks are currently not
+  production-ready. One new table privilege was introduced: `TABLE_REMOTE_SIGN`. To enable remote signing:
+    1. Set the system-wide property `REMOTE_SIGNING_ENABLED` or the catalog-level `polaris.request-signing.enabled`
+       property to `true`.
+    2. Grant the `TABLE_REMOTE_SIGN` privilege to a catalog role. The role must also be granted the `TABLE_READ_DATA`
+       and `TABLE_WRITE_DATA` privileges.
```
singhpk234: [doubt] is there an assertion when even TABLE_REMOTE_SIGN is granted these two privieldges would be there ? 

nit : 
```suggestion
    2. Grant the `TABLE_REMOTE_SIGN` privilege to a catalog role. The catalog role must also be granted the `TABLE_READ_DATA`
       and `TABLE_WRITE_DATA` privileges.
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -101,4 +120,58 @@ public AccessConfig getAccessConfig(
...
+        instanceof AwsCredentialsStorageIntegration awsCredentialsStorageIntegration)) {
+      LOGGER
+          .atWarn()
+          .addKeyValue("tableIdentifier", tableIdentifier)
+          .log("Table entity storage integration is not an AWS credentials storage integration");
+      return AccessConfig.EMPTY;
+    }
+
+    String prefix = prefixParser.catalogNameToPrefix(callContext.getRealmContext(), catalogName);
+    URI signerUri = uriInfo.getBaseUriBuilder().path(PolarisResourcePaths.API_PATH_SEGMENT).build();
```
singhpk234, line 287: can we add a TODO in code to handle proxies ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -844,6 +869,37 @@ ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING, getResolvedCatalogEntity())) {
...
+      return delegationModes.iterator().next();
+    }
+
+    if (delegationModes.contains(VENDED_CREDENTIALS) && delegationModes.contains(REMOTE_SIGNING)) {
+
+      boolean skipCredIndirection =
+          realmConfig.getConfig(FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
+
+      boolean credentialSubscopingAllowed =
+          baseCatalog instanceof IcebergCatalog
```
singhpk234, line 940: what if its an IcebergCatalog but doesn't support STS ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -844,6 +869,37 @@ ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING, getResolvedCatalogEntity())) {
     return responseBuilder;
   }
 
+  private AccessDelegationMode selectAccessDelegationMode(
+      Set<AccessDelegationMode> delegationModes) {
```
singhpk234, line 923: does linking this part of spec https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L1859 helps in putting some context that server is allowed to decide which one to choose ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
singhpk234, line 287: can we add a TODO to handle proxies here ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,198 @@
...
+      throw new ForbiddenException("Cannot use S3 remote signing with federated catalogs.");
+    }
+    baseCatalog =
+        catalogFactory.createCallContextCatalog(callContext, polarisPrincipal, resolutionManifest);
+  }
+
+  public PolarisS3SignResponse signS3Request(
+      PolarisS3SignRequest s3SignRequest, TableIdentifier tableIdentifier) {
+
+    LOGGER.debug("Requesting s3 signing for {}: {}", tableIdentifier, s3SignRequest);
```
singhpk234: do we need to explicity log it here or quarkus supports Access log, asking because i don't see such logging in Iceberg Catalog Handler

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        return StorageUtil.getLocationsUsedByTable(baseTable.operations().current());
+      }
+
+      throw new ForbiddenException("No storage configuration found for table.");
+
+    } else {
+
+      // If the table or view doesn't exist, the engine might be writing the manifests before the
+      // table creation is committed. In this case, we still need to check allowed locations from
+      // the parent entities.
```
singhpk234: is this the case of staged table ? i wondering if we should block on this case ? becasue it may happen the table is different from namespace location ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,198 @@
...
+  private Set<String> getTargetLocations(PolarisS3SignRequest s3SignRequest) {
+    // TODO M2: map http URI to s3 URI
+    return Set.of();
+  }
+
+  public static void throwIfRemoteSigningNotEnabled(
+      RealmConfig realmConfig, CatalogEntity catalogEntity) {
+    if (catalogEntity.isExternal()) {
+      throw new ForbiddenException("Remote signing is not enabled for external catalogs.");
+    }
```
singhpk234, line 151: I understand we discussed this before for federated catalog we should disable remote signing but during that time we didn;t support cred vending for them either, i wonder if its  a good time to reconsider that remote signing would be possible too (ofc in a new PR)

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,198 @@
...
+
+    return s3SignResponse;
+  }
+
+  private Collection<String> getAllowedLocations(TableIdentifier tableIdentifier) {
+
+    if (baseCatalog.tableExists(tableIdentifier)) {
+
+      // If the table exists, get allowed locations from the table metadata
+      Table table = baseCatalog.loadTable(tableIdentifier);
```
singhpk234, line 126: for each SIGN request a loadTable call is super expensive, if someone enables this in prod by accident for a table million of files the service might become unresponsive ? 
[suggestion / optional to address] wondering for now if we should add a prod readiness check to keep this disabled for prod ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
```diff
@@ -0,0 +1,198 @@
...
+      if (configurationInfo.isEmpty()) {
+        throw new ForbiddenException("No storage configuration found for table.");
+      }
+
+      return configurationInfo.get().getAllowedLocations();
+    }
+  }
+
+  private Set<PolarisStorageActions> getStorageActions(PolarisS3SignRequest s3SignRequest) {
+    // TODO M2: better handling of DELETE and LIST
```
singhpk234: [for my understanding] can you please elaborate this case more ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
adutra: Yes. When creating a new table, engines write the manifest files first, then commit the table. The manifest writes must be signed too. Thus, a request to sign can arrive at the server before the table is created.

CHANGELOG.md
adutra: > is there an assertion when even TABLE_REMOTE_SIGN is granted these two privieldges would be there ?

Currently no.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: That's a good question! 

It's complex to check this at this level because you need to find the `AwsStorageConfigurationInfo` corresponding to the resolved path in order to know if STS is unavailable.

I think we shouldn't make this logic too complex, because it's not common for a client to request more than one access delegation mode anyways.

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: Quarkus indeed supports access logs. We must be careful with logging here because this endpoint will be called very, very frequently.

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: Totally agree, I think we could discuss this possibility later on.

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: > for each SIGN request a loadTable call is super expensive

Indeed, that's where M2 will help because it will save us this `loadTable` call. 

I will add a FIXME for now.

> [suggestion / optional to address] wondering for now if we should add a prod readiness check to keep this disabled for prod ?

Good idea, will do 👍

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: By just looking at the request URI, it's not easy to know if the request is a write or a delete, or if it is a read or a list.

For example: the [ListObjects](https://redocly.github.io/redoc/?url=https://api.apis.guru/v2/specs/amazonaws.com/s3/2006-03-01/openapi.json#operation/ListObjects) operation is conceptually a LIST operation, and [GetObject](https://redocly.github.io/redoc/?url=https://api.apis.guru/v2/specs/amazonaws.com/s3/2006-03-01/openapi.json#operation/GetObject) is conceptually a READ. But both requests use the GET method, so it's not easy to disambiguate.

Another example: the [DeleteObject](https://redocly.github.io/redoc/?url=https://api.apis.guru/v2/specs/amazonaws.com/s3/2006-03-01/openapi.json#operation/DeleteObject) operation uses the DELETE method, but the [DeleteObjects](https://redocly.github.io/redoc/?url=https://api.apis.guru/v2/specs/amazonaws.com/s3/2006-03-01/openapi.json#operation/DeleteObjects) operation uses... the POST method 🤷‍♂️ 

I will add some comments to clarify.

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: That's a very good question, let me check if we have tests for staged table creation + remote signing.

adutra: After investigating a bit, it seems that staged vs direct doesn't affect remote signing.

However I found a few feature flags that could interfere with remote signing:

- `ALLOW_UNSTRUCTURED_TABLE_LOCATION`
- `ALLOW_EXTERNAL_TABLE_LOCATION`
- `ALLOW_EXTERNAL_METADATA_FILE_LOCATION`
- `DEFAULT_LOCATION_OBJECT_STORAGE_PREFIX_ENABLED`

I will try to add tests for these flags + remote signing.

adutra: A few updates on this. I added a few more tests and verified that all the above flags are working with remote signing.

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: Update on this topic: I was able to remove the `loadTable` call, replaced with a direct check of the `PolarisEntity` properties. That's going to be faster imho.

CHANGELOG.md
singhpk234: Are we then failing during Runtime ? is this done mostly like at the time of granting there was both READ & WRITE but then admin revoked one of them so this grant needs to be revoked too ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
singhpk234: should we add a ticket for this so we can tackle this later, WDYT ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
singhpk234: Totally Agree, so is the conclusion not to log this in the debug anymore ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
singhpk234: should we file a ticket so we can tackle this later as a community, WDYT ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
singhpk234: do we store the 3 properties in the entity properties asking because 
i only see the StorageUtil#getLocationsUsedByTable passing TableMetada#properties and also not seeing it actually persisted in StorageUtil, am i missing something ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: This call costs almost nothing: all parameters are constants, so the call will immediately return if the DEBUG level is not active for this logger. The JVM can generally inline this call.

But, if you prefer, I will remove the call.

CHANGELOG.md
adutra: This is done by Polaris RBAC enforcement. The two `PoalrisAuthorizableOperation` instances used in remote signing are declared as follows:

```java
  SIGN_S3_READ_REQUEST(EnumSet.of(TABLE_REMOTE_SIGN, TABLE_READ_DATA)),
  SIGN_S3_WRITE_REQUEST(EnumSet.of(TABLE_REMOTE_SIGN, TABLE_WRITE_DATA)),
```

Thus as you can see to authorize these operations, the role must have been granted 2 privileges: `TABLE_REMOTE_SIGN` and either `TABLE_READ_DATA` (for reads) or `TABLE_WRITE_DATA` (for writes).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: https://github.com/apache/polaris/issues/3090

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: https://github.com/apache/polaris/issues/3091

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: Sorry, which 3 properties? I'm not sure I understand your question 😅

adutra:
Heads up: since this PR creates constant merge conflicts, I squashed the older commits to facilitate conflict resolution.

binarycat0:
Hello everyone 👋 I would like to propose my help in implementing M2 ([from proposed milestones](https://docs.google.com/document/d/1ygdia7u4bUHUt6n8XhZo48aKoIyyrCvKqan3XP25iB8/edit?tab=t.0#heading=h.76hegagb5gud)). 
As I see M2 can be split into 3 independent features: locations-check, nessir-style-signed-params, support-proxies. 
I already have a solution for check-locations and ready to create a follow-up PR after this will be merged. 

Could we agree on a timeline to merge?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
singhpk234: iceberg properties location, write.data.path, write.metadata.path, which is used for credential vending, I am not sure we persist them Entity, as the function StorageUtil#getLocationsUsedByTable has always worked in TableMetadata and not on Entity properties

CHANGELOG.md
singhpk234: It seems like either READ or WRITE works but the CHANGELOG gives an impression both is required ? should we update the wording ?

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
singhpk234: TBH its fine, I just wanted to know whats your preference :)

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
adutra: Ah now I get it! Good question! I think there is a problem indeed.

For the base location, we're good, it is stored in the properties map:

https://github.com/apache/polaris/blob/07edd301f6cbb87ff2cf875bc5ea0b99875acded/polaris-core/src/main/java/org/apache/polaris/core/entity/table/IcebergTableLikeEntity.java#L119

But afaict `write.data.path` and `write.metadata.path` are not stored in the entity 😞 

I see a few options:

1. We could store `write.data.path` and `write.metadata.path` in the entity, but that will be a problem for catalogs created before the change.
2. We could go back to `baseCatalog.loadTable(tableIdentifier)` but that would be very slow.
3. We could just add a FIXME here + create a GitHub issue and wait for a better solution with M2.

Wdyt?

I would go with 3) since we know that this code path will be improved, but I'm fine with 2) as well.

CHANGELOG.md
adutra: OK I updated the changelog!

runtime/service/src/main/java/org/apache/polaris/service/storage/s3/sign/S3RemoteSigningCatalogHandler.java
singhpk234: I would personally prefer 2 over 3 as IMHO 3 is not quite correct ? 

one thing i was thinking is every sign request will make a corresponding loadTable to get signer configs, should be backfill with an updateEntity if these props are missing i.e during read, wdyt ?

adutra: Sorry for the late reply!

I went back to option 2 : `baseCatalog.loadTable(tableIdentifier)`.

> one thing i was thinking is every sign request will make a corresponding loadTable to get signer configs, should be backfill with an updateEntity if these props are missing i.e during read, wdyt ?

That's an interesting idea, however, it would still be a temporary measure until we get faster signings. So all in all, I'm not sure if it's worth investing on this change.

Wdyt?

adutra:
A few updates on remote signing:

The community decided to push some changes down to the Iceberg code, which resulted in two PRs:

* https://github.com/apache/iceberg/pull/15112
* https://github.com/apache/iceberg/pull/15168 

I have been working actively on a new version that incorporates the above changes, and also provides the full feature, including the "M2" topics:

* Reverse proxy support
* Encrypted signing token
* Full authz checks including table locations

I think it makes sense to close this PR and open a new one later, not only because it is stale, but because the authorization model will change significantly (it will go from a model that checks on every signing request to a model that checks authz only once at load table).

I'm confident we could include the Iceberg changes in the upcoming Iceberg 1.11 release, then have remote signing released (as beta of course) for Polaris 1.5. **But this requires upgrading Polaris to Java 17**.

\cc @singhpk234

PR #2282, opened 2025-08-06
Fix Hadoop federation to initialize the configuration

poojanilangekar:
This PR addresses the gap in Hadoop federation to ensure that we initialize the hadoop configuration prior to initializing the catalog object.

The iceberg library expects the hadoop configuration to be initialized before creating a HadoopCatalog object. This change ensures that Polaris is compatible with the underlying iceberg library. Additionally, since the config initialization is based on the underlying (default) core-site.xml file, the change ensures that the federated catalog was created using `IMPLICIT` authentication mode. 

Testing: 
Due to current limitations in the current test setup, tested the change manually.
[TODO] Add a regtest with hadoop federation once the change is baked into the apache/polaris docker image.


poojanilangekar:
CC : @eric-maynard

PR #2283, opened 2025-08-06
(WIP) Separate Cloud Integration Tests

tmater:
This change separates the cloud tests into a new Gradle task called cloudTest, allowing them to be executed independently. Previously, these tests were part of the intTest task but were skipped automatically when cloud credentials weren't available. Moving them to a separate task avoids running unnecessary test setup in local and CI environments where cloud integration isn't needed. Tested locally by running both intTest and cloudTest.


build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -52,6 +52,14 @@ testing {
       }
       sources { java.srcDirs(tasks.named("quarkusGenerateCodeTests")) }
     }
+    register<JvmTestSuite>("cloudTest") {
+      targets.all {
+        tasks.named("compileCloudTestJava").configure {
+          dependsOn(tasks.named("compileQuarkusTestGeneratedSourcesJava"))
+        }
+      }
```
snazy:
```suggestion
      tasks.named("compileCloudTestJava").configure {
        dependsOn("compileQuarkusTestGeneratedSourcesJava")
      }
```
(`tasks` is `Project.tasks`, not part of `JvmTestSuite` - not to confuse w/ `JvmTestSuite.testTask` )

Same change can be applied above.

runtime/service/build.gradle.kts
```diff
@@ -212,50 +212,55 @@ tasks.named<Test>("test").configure {
...
-            !"quarkus.test.arg-line".equals(it.key) &&
-            !"quarkus.log.file.path".equals(it.key)
-        }
-        .forEach { args.add("${it.key}=${it.value}") }
-
-      args
+tasks
+  .withType<Test>()
+  .matching { it.name in listOf("intTest", "cloudTest") }
+  .configureEach {
```
snazy:
```suggestion
listOf("intTest", "cloudTest").map { tasks.named<Test>(it) }.forEach { it.configure {
```

The `.matching { it.name` causes an eager configuration of all test tasks.

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -52,6 +52,14 @@ testing {
       }
       sources { java.srcDirs(tasks.named("quarkusGenerateCodeTests")) }
     }
+    register<JvmTestSuite>("cloudTest") {
+      targets.all {
+        tasks.named("compileCloudTestJava").configure {
```
snazy:
```suggestion
      tasks.named(sources.compileJavaTaskName).configure {
```

should work as well

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -78,7 +86,14 @@ configurations.named("intTestImplementation").configure {
   extendsFrom(configurations.getByName("testImplementation"))
 }
 
-dependencies { add("intTestImplementation", java.sourceSets.getByName("test").output.dirs) }
+configurations.named("cloudTestImplementation").configure {
+  extendsFrom(configurations.getByName("testImplementation"))
+}
```
snazy: Can probably moved into the test suites `.register` closure as
```kotlin
      configurations.named(sources.implementationConfigurationName).configure {
        extendsFrom(configurations.get("testImplementation"))
      }
```

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -78,7 +86,14 @@ configurations.named("intTestImplementation").configure {
...
+
+dependencies {
+  add("intTestImplementation", java.sourceSets.getByName("test").output.dirs)
+  add("cloudTestImplementation", java.sourceSets.getByName("test").output.dirs)
```
snazy: Can probably moved into the test suites `.register` closure as
``` kotlin
      dependencies {
        implementation(java.sourceSets.getByName("test").output.dirs)
      }
```

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -52,6 +52,14 @@ testing {
       }
       sources { java.srcDirs(tasks.named("quarkusGenerateCodeTests")) }
     }
+    register<JvmTestSuite>("cloudTest") {
```
snazy: I _think_ with all the suggestions below, it could be unified as:
```kotlin
    fun intTestSuiteConfigure(testSuite: JvmTestSuite) = testSuite.run {
      targets.all { 
        testTask.configure {
// TODO configuration from lines 37..42 went here
          // For Quarkus...
          //
          // io.quarkus.test.junit.IntegrationTestUtil.determineBuildOutputDirectory(java.net.URL)
          // is not smart enough :(
          systemProperty("build.output.directory", layout.buildDirectory.asFile.get())
          dependsOn(tasks.named("quarkusBuild"))
        }
      }
      tasks.named(sources.compileJavaTaskName).configure {
        dependsOn("compileQuarkusTestGeneratedSourcesJava")
      }
      configurations.named(sources.runtimeOnlyConfigurationName).configure {
        extendsFrom(configurations.getByName("testRuntimeOnly"))
      }
      configurations.named(sources.implementationConfigurationName).configure {
// Let the test's implementation config extend testImplementation, so it also inherits the
// project's "main" implementation dependencies (not just the "api" configuration)
        extendsFrom(configurations.getByName("testImplementation"))
      }
      sources { java.srcDirs(tasks.named("quarkusGenerateCodeTests")) }
    }
    listOf("intTest", "cloudTest").forEach { register<JvmTestSuite>(it).configure { intTestSuiteConfigure(this) } }
```

snazy (commented):
Thanks for tackling this, @tmater !

Some comments on the Gradle parts, just potential simplifications.

Would you mind making the super classes of the moved test classes `abstract`, and maybe rename those to make clear that those are meant to be overridden?

To disable the tests, it's also possible to add something like the following to the test "top-level" classes like `RestCatalogAwsIT`, so even the `@BeforeAll` doesn't run.
```java
@EnabledIfEnvironmentVariable(named = "INTEGRATION_TEST_S3_PATH", matches=".+")
```

tmater:
Thanks @snazy for the thorough review and great suggestions!
 - Added the Gradle function described by you, I hope I used it properly
 - Made base classes abstract, suffixed them with `Base`
 - Added annotation to skip test class execution and removed the `shouldSkip` method
 - Replaced the `matching` with the suggested lazy init

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
```diff
@@ -72,14 +82,6 @@ configurations.all {
   }
 }
 
-// Let the test's implementation config extend testImplementation, so it also inherits the
-// project's "main" implementation dependencies (not just the "api" configuration)
-configurations.named("intTestImplementation").configure {
-  extendsFrom(configurations.getByName("testImplementation"))
-}
-
-dependencies { add("intTestImplementation", java.sourceSets.getByName("test").output.dirs) }
```
adutra, line 81: It seems that this line got removed. Isn't it necessary anymore?

snazy: Eagle eye - however, tests pass 🤷

adutra: Exactly, so maybe not needed after all.

adutra (approved):

build-logic/src/main/kotlin/polaris-runtime.gradle.kts
tmater: Oh, nice catch!
I was tinkering with this a bit early afternoon, but then I won't update it.
Thank you for the review! :)

PR #2286, opened 2025-08-06
Add a regression test for Catalog Federation

poojanilangekar:
This change adds a PR to test the end-to-end catalog federation functionality. 

Steps involved: 
1. Create a `new-user` account. 
2. Create an `INTERNAL` catalog, called `test-catalog-local`. 
3. Grant the following privileges on `test-catalog-local` :`TABLE_WRITE_DATA` -> `catalog_admin` -> `service_admin` -> `new-user`. 
4. Create an `EXTERNAL` catalog called `test-catalog-external` that points to `test-catalog-local` and uses the client-id and client-secret credentials for the `new_user` account. 
5. Grant the following privileges on `test-catalog-external` :`TABLE_WRITE_DATA` -> `catalog_admin` -> `service_admin`.
6. Connect to `test-catalog-local` and use it to create namespaces/tables and insert some data. 
7. Connect to `test-catalog-external` and select the existing data and insert new rows. 
8. Re-connect to `test-catalog-local` and verify that the changes made in `test-catalog-external` are visible. 

This change disables URL overlap check since we are federating to a single Polaris instance.

### DIFF
--- a/regtests/README.md
+++ b/regtests/README.md
@@ -85,6 +85,15 @@ project, just run:
 env POLARIS_HOST=localhost ./regtests/run.sh
 ```
 
+The catalog federation tests rely on the following configurations in `application.properties` to
+be set in order to succeed.
+
+```
+polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE", "FILE"]
+polaris.features."ENABLE_CATALOG_FEDERATION"=true
+polaris.features."ALLOW_OVERLAPPING_CATALOG_URLS"=true
+```
+
 To run the tests in verbose mode, with test stdout printing to console, set the `VERBOSE`
 environment variable to `1`; you can also choose to run only a subset of tests by specifying the
 test directories as arguments to `run.sh`. For example, to run only the `t_spark_sql` tests in
--- a/regtests/t_catalog_federation/src/catalog_federation.sh
+++ b/regtests/t_catalog_federation/src/catalog_federation.sh
@@ -1,0 +1,201 @@
+#!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# This test creates an INTERNAL catalog and an EXTERNAL catalog with passthrough facade
+# to demonstrate true catalog federation.
+
+set -e
+
+
+export POLARIS_TOKEN=$(curl -s http://localhost:8181/api/catalog/v1/oauth/tokens \
+    --user root:s3cr3t \
+    -d 'grant_type=client_credentials' \
+    -d 'scope=PRINCIPAL_ROLE:ALL' | jq -r .access_token)
+
+
+SPARK_BEARER_TOKEN="${POLARIS_TOKEN}"
+
+echo "=== Setting up Catalog Federation Test ==="
+
+# Step 1: Create a new principal
+echo "Creating new principal..."
+PRINCIPAL_RESPONSE=$(curl -s -X POST -H "Authorization: Bearer ${SPARK_BEARER_TOKEN}" -H 'Content-Type: application/json' \
+  http://${POLARIS_HOST:-localhost}:8181/api/management/v1/principals \
+  -d '{
+    "principal": {
+      "name": "new-user"
+    }
+  }')
+
+NEW_CLIENT_ID=$(echo "$PRINCIPAL_RESPONSE" | jq -r '.credentials.clientId')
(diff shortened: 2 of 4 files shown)


poojanilangekar:
CC @dennishuo @eric-maynard

regtests/README.md
```diff
@@ -85,6 +85,15 @@ project, just run:
 env POLARIS_HOST=localhost ./regtests/run.sh
 ```
 
+The catalog federation tests rely on the following configurations in `application.properties` to
+be set in order to succeed.
+
+```
+polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE", "FILE"]
```
eric-maynard: super nit:
```suggestion
polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"=["S3","GCS","AZURE","FILE"]
```

eric-maynard: Do we not also rely on `readiness.ignore-severe-issues`?

eric-maynard: And `ALLOW_INSECURE_STORAGE_TYPES`

regtests/t_catalog_federation/src/catalog_federation.sh
```diff
@@ -0,0 +1,196 @@
...
+use polaris;
+show namespaces;
+select * from ns1.test_table order by id;
+insert into ns1.test_table values (3, 'Charlie');
+select * from ns2.test_table order by id;
+insert into ns2.test_table values (3, 'Apache Polaris');
+EOF
+
+echo ""
+echo "=== Verifying federation via LOCAL catalog ==="
```
eric-maynard, line 166: Can we do some verification of RBAC as well?

regtests/README.md
poojanilangekar: The two flags are set as default JVM args

https://github.com/apache/polaris/blob/1b9316932458e80f7d16a685f4df9c13fbe9110e/runtime/server/build.gradle.kts#L77

regtests/t_catalog_federation/src/catalog_federation.sh
poojanilangekar: I will update this test once #2223 is merged. For now, we can only do catalog-level RBACs for federated catalogs.

regtests/README.md
poojanilangekar: I just realized that we don't need the "FILE" parameter either because it's included in the JVM args. I've removed it.

eric-maynard (approved):

dimas-b (commented):
The same test seems possible to do under the JUnit5 framework (as an `intTest`). 

That approach is preferable from my POV because it is more easily discoverable in a  java IDE and would show regressions at part of the usual gradle validation tasks.

WDYT?

poojanilangekar:
This is a regression test (which is still run by GitHub CI). Can I send out a separate PR for an integration test if that works for you? At that point we can remove this test if necessary. Currently we have no tests for catalog federation, so it might be a good idea to keep this around while I work on the IT test as you suggested. 

For some context, I won't be able to dedicate as much time to the project going forward so I expect that to take some time, in the meantime it is a good idea to have some test that check that federation still works and flag PRs that potentially break federation. 

Please let me know what you think.

dimas-b:
From my POV running docker-based tests is more cumbersome than running JUnit5-based tests. So, if the same functionality can be tested under the JUnit framework, I'd prefer that approach.

I do not oppose merging this PR, but I believe it too heavy for the job it performs.

eric-maynard:
@dimas-b I agree that should add more integration tests on top of this , but we'll need docker-based regression tests for federation in general (imagine testing federation to Hive, you need to actually run HMS somewhere). Accordingly I think it's okay or even prudent to have a regression test for IRC federation for the sake of parity.

snazy:
Having tests for more complex setups is totally legit. I just don't think that containers are are a requirement for that.

This is where for example https://github.com/apache/polaris-tools/pull/18 comes into play - it runs a Polaris server (actually "anything Quarkus") from the "local" code base or a specific distributed/released version, which you can then use in an integration test to test the federation use case.

From a machine resource usage it's even more lightweight as only the processes are run, but no container infra.

And you'd be able to run it from your IDE and, with some option tweaking, can debug the processes.

dennishuo (approved):
LGTM

I agree we'll want both flavors longer-term (java tests and black-box regtests) because long-term they'll be able to test different scopes of functionality. In general having this black-box end-to-end is still needed for building confidence that it really works from the perspective of an external caller without JVM-local "tricks" that might creep into a JUnit test, especially when the nature of the feature (such as Federation) innately deals with how things interact beyond the single JVM boundary.

Re: heavyweight containers, it doesn't look like this PR actually *requires* anything related to Docker -- the `sh` test just extends the pre-existing patterns we already have for other `sh` tests and AFAICT is equally happy to run against a direct `./gradlew run` or IDE-run or `java -jar` local deployment vs having Polaris run in the Docker container. So the Docker parts are just making sure the pre-existing Docker-based runs such as for precommit/CI are also able to run the new test successfully.

snazy:
I'm not against this PR in principle, however it was mentioned that "we'll need docker-based regression tests". I do think that the state of the regression tests as of today. do nothing that could not be done in a Java based integration test.

OTOH, playing devil's advocate here, do we have good enough JUnit (integration) test coverage for federation?
Another devil: are the regression tests actually really testing a "full black box end to end test" including all the things that end users do (real database, external IdP, secured Polaris setup, external x/y/z)?

dennishuo:
Yeah, I agree the "Docker" discussion is worth expanding in another thread, even though it's orthogonal to this PR. I'm definitely also in favor of supporting the non-Docker scenarios like running the server in an IDE or anywhere else and still being able to run tests against those, and we can make sure the structure of regtests continues supporting that pattern.

Also +1 to expanding the black-box testing scope to really exercise the things a real user will do; having this PR as a starting point will hopefully reduce the barrier of entry for folks interested in exercising those complex scenarios.

Since it sounds like no one is opposed to merging this PR, I'll go ahead and merge it to unblock expanding test coverage.

PR #2291, opened 2025-08-07
Fix REST responses for failed Admin operations

XN137:
the `boolean` return values of many methods in `PolarisAdminService`
were often simply not getting used at all, thus the REST api returned
success in those cases even though the `PrivilegeResult` was marked
as failed.
due to this fix a silently failing test now needs to be adjusted.
    
we return the `PrivilegeResult` instead of a `boolean` to give the
client at least some indication of what has gone wrong on the server
side.

note that some of the other operations were throwing Expcetions already,
which are already reported back correctly to the client.

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
@@ -124,7 +124,9 @@ public class PolarisManagementServiceIntegrationTest {
 
   @AfterAll
   public static void close() throws Exception {
-    client.close();
+    if (client != null) {
+      client.close();
+    }
   }
 
   @AfterEach
@@ -1728,7 +1730,7 @@ public class PolarisManagementServiceIntegrationTest {
             .managementApi(catalogAdminToken)
             .request(
                 "v1/principal-roles/"
-                    + principalRoleName
+                    + principalRoleName2
                     + "/catalog-roles/"
                     + catalogName
                     + "/"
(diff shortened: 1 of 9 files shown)


runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -640,7 +661,7 @@ public Response addGrantToCatalogRole(
             .log("Don't know how to handle privilege grant: {}", grantRequest);
         return Response.status(Response.Status.BAD_REQUEST).build();
     }
-    return Response.status(Response.Status.CREATED).build();
+    return toResponse(result, Response.Status.CREATED);
```
snazy, line 672: This means that errors were not propagated to clients, right?

XN137: yeah, some methods do throw exceptions internally (and those get propagated to clients correctly) but the returned `boolean` value of the methods was often ignored.
as part of this PR we have to fix a test in `PolarisManagementServiceIntegrationTest.java` which was silently failing before.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
```diff
@@ -1743,14 +1745,32 @@ public void testCatalogAdminGrantAndRevokeCatalogRoles() {
...
+                    + "/catalog-roles/"
+                    + catalogName
+                    + "/"
+                    + catalogRoleName)
+            .delete()) {
+      assertThat(response)
+          .returns(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), Response::getStatus);
+      assertThat(response.hasEntity()).isTrue();
+      ErrorResponse errorResponse = response.readEntity(ErrorResponse.class);
+      assertThat(errorResponse.message()).contains("Operation failed: GRANT_NOT_FOUND");
```
XN137, line 1772: there could be an argument that `GRANT_NOT_FOUND` should be a 404 instead of internal server error but it would mean we need to translate all `BaseResult.ReturnStatus` values to a proper HTTP code.
i think this can be figured out in a followup and we should first stop silently failing in those cases.

snazy: Hm - HTTP/500 feels odd, as it's rather a user (argument) error. HTTP/400 would be better IMHO.

XN137: ok switched to HTTP/400

snazy (approved):

dimas-b (approved):

adutra (approved):

snazy:
If nobody objects, I'll merge later today.

PR #1817, opened 2025-06-05
Use isolated request contexts for task execution

dimas-b:
* Avoid propagating request contexts into threads running background tasks.

* Use fresh request context for each task instead.

* Manually propagate realm context into a task from the request that submits the task.

* Update DefaultConfigurationStore to cross-check realm IDs in request context and explicit method parameters.

* Fix tests to set thread-local data to match CDI context data.

Next step: Simplify DefaultConfigurationStore API to remove contextual arguments and use CDI context instead.

PR #2294, opened 2025-08-07
Remove `CallContext.copy()`

snazy:
The function is only needed to create a "safe clone" of `PolarisCallContext` in `TaskExecutorImpl`, which doesn't have references to CDI request scoped beans.

To still let `TaskExecutorImpl` making "safe clones", a functionality to get (fresh) instances of `RealmContext` is required. To enable this, the `RealmContextResolver` has been enhanced with "`RealmContext` lookups" by realm-ID. That in turn led to splitting the HTTP/REST-to-realm-context resolution into two parts: HTTP/REST-to-realm-ID and realm-ID-to-context.

This change is part of the effort to allow tasks run run in different JVMs.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
@@ -18,6 +18,7 @@
  */
 package org.apache.polaris.service.task;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
@@ -37,12 +38,17 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import org.apache.polaris.core.PolarisCallContext;
+import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.context.CallContext;
+import org.apache.polaris.core.context.RealmContext;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.TaskEntity;
+import org.apache.polaris.core.persistence.BasePersistence;
 import org.apache.polaris.core.persistence.MetaStoreManagerFactory;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
+import org.apache.polaris.service.context.RealmContextResolver;
 import org.apache.polaris.service.events.AfterTaskAttemptedEvent;
 import org.apache.polaris.service.events.BeforeTaskAttemptedEvent;
 import org.apache.polaris.service.events.PolarisEventListener;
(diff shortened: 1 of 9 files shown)


runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -122,27 +131,58 @@ public void addTaskHandlerContext(long taskEntityId, CallContext callContext) {
     // the task is still running.
     // Note: PolarisCallContext has request-scoped beans as well, and must be cloned.
     // FIXME replace with context propagation?
-    CallContext clone = callContext.copy();
-    tryHandleTask(taskEntityId, clone, null, 1);
+    tryHandleTask(taskEntityId, new TaskContext(callContext), null, 1);
+  }
+
+  record TaskContext(String realmId, PolarisDiagnostics diagnostics) {
```
gh-yzou, line 137: have we done some testing (at least manually to verify the task is still executing correctly)? as far as I know, we currently do not have a good integration test that actually helps verify the background task. The regtests t_pyspark/test_spark_sql_s3_with_privileges.py contains a tests that helps verify the background purge task, but not running in the CI today. 

Can we follow up the readme here https://github.com/apache/polaris/blob/main/regtests/README.md to run the test against aws to verify things are still working?

dimas-b: Actually, @gh-yzou 's comment made me think that we probably need to start a new request context for the task via `@ActivateRequestContext` as in #1817 and also block context propagation on the task thread pool. That should give us proper CDI context isolation. WDYT?

snazy: CDI context propagation does not work across different JVMs for the tasks-proposals.

snazy: > have we done some testing

I mean, sure? There are tests in the code base for this.

> The regtests t_pyspark/test_spark_sql_s3_with_privileges.py ... not running in the CI today.

I'd suggest to fix that independently. Mind taking a stab on that, @gh-yzou ? MinIO, Azurite and Google emulator help there.

eric-maynard: > I mean, sure? There are tests in the code base for this.

Our task tests are not very complete right now, so this doesn't give me much confidence.

gh-yzou: t_pyspark/test_spark_sql_s3_with_privileges.py can only be run locally with your avaiable S3 account, @jbonofre is working on getting an S3 account for cli testing.

However, I am not sure what would be the extra benefit this approach would give us. The copy() method is used to propagate the whole call Context and RealmContext to the task execution, and the id look up seems also only used by task execution and requires an extra re-construction step. Furthermore, this approach doesn't seem scalable in the future if we are adding more information to callContext or RealmContext (such as user specific information), especially for RealmContext, it may require us to implement a RealmContextManager for lookup.

If the concern is that this function is only used by task executor, i recall @dimas-b was POC something that leverages CDI feature to propagate the CallContext to background thread, which seems a cleaner way to do that. @dimas-b maybe we can resume that work?

PR #1528, opened 2025-05-05
Interface changes for pagination

eric-maynard:
The Apache Iceberg REST catalog spec [describes](https://github.com/apache/iceberg/blob/f7c6d57a03282516c6f988fa23a809416cd765b1/open-api/rest-catalog-open-api.yaml#L258) how APIs like `listNamespaces` can implement pagination, but Polaris currently does not implement pagination.

This PR introduces a framework that will allow us to enable pagination for these APIs and that can also be extended to other APIs like `listCatalogs`.

<hr>

This is a modified version of #273 that removes the actual pagination implementation at the persistence level.

PR #2296, opened 2025-08-07
Remove unused LIST_PAGINATION_ENABLED feature configuration

adutra:


eric-maynard:
Why is this unused?

adutra:
> Why is this unused?

I have no idea. It was introduced in #1528, that's all I know.

eric-maynard:
I think that this should have been used in #1938; does pagination now ignore this flag?

adutra:
> I think that this should have been used in #1938; does pagination now ignore this flag?

@snazy do you know?

dimas-b (approved):

dimas-b:
Adding if/else on the paginated call paths looks a bit of an overkill to me. As currently implemented, IIRC, pagination is controlled by the client. If the client does not request a call to be paginated, the server will work as before #1938.

IMHO, the risk of breaking clients with current pagination code is low.

eric-maynard:
We have feature flags for tons of things controlled by the client -- like table location overlap, generic tables, drop-with-purge, policies, etc. This flag was put in place specifically to guard against regressions caused by the new code path introduced by pagination.

dimas-b:
Fair point. I'll look into that presently.

adutra:
Moving this to draft while @dimas-b looks into this issue.

dimas-b:
Following up in #2401

adutra:
Closing then, thanks @dimas-b !

PR #2301, opened 2025-08-07
Modularize calls to federated catalogs

poojanilangekar:
Based on the discussion in the community sync, this PR modularizes federation calls to a separate factory. The specific instances of the NonRESTCatalogFactory is loaded only when a non-REST catalog is initialized and accessed. Beyond the catalog initialization, all the calls are identical for internal and federated catalogs. 

After this change, Polaris needs to be compiled with the `NonRESTCatalogs` property set to the list of catalogs for which are implemented/supported. Unless a catalog type is a part of the flag, Polaris will not load its implementation at runtime. 

Testing:
Applied the regtest in PR #2286.


poojanilangekar:
CC @eric-maynard @dennishuo

poojanilangekar:
The conditional runtime dependency could rely on one of the two options: 

1.  Based on the `NonRESTCatalogs` property (which an additional property to be set in `gradle.properties` or as a JVM argument).
1.  Reuse the existing `polaris.features.SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES` but require the user to declare it in `gradle.properties` instead of `application.properties`.

The current implementation uses option 1. 

I can change the option or name if necessary. If you have other possible suggestions, please let me know.

polaris-core/src/main/java/org/apache/polaris/core/catalog/NonRESTCatalogFactory.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.polaris.core.connection.ConnectionConfigInfoDpo;
+import org.apache.polaris.core.secrets.UserSecretsManager;
+
+/**
+ * Factory interface for creating non-REST catalog handles based on connection configuration.
+ *
+ * <p>Implementations should be annotated with CDI annotations and use the @Identifier annotation to
+ * specify which connection type they support.
+ */
+public interface NonRESTCatalogFactory {
```
eric-maynard: IIUC this is basically referring to anything _other_ than an Iceberg REST catalog? Just "REST" might be confusing.

What's more, can we just have one interface (e.g. `ExternalCatalogFactory`) and provide an implementation that gives you back an IRC for the sake of uniformity?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -219,7 +221,6 @@ protected void initializeCatalog() {
       Catalog federatedCatalog;
       ConnectionType connectionType =
           ConnectionType.fromCode(connectionConfigInfoDpo.getConnectionTypeCode());
-
       switch (connectionType) {
         case ICEBERG_REST:
```
eric-maynard: Here's where we would plug in the extra wrapper layer for the factory that produces a Catalog impl for IRC

polaris-core/src/main/java/org/apache/polaris/core/catalog/NonRESTCatalogFactory.java
poojanilangekar: Done.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
poojanilangekar: Done.

runtime/service/build.gradle.kts
```diff
@@ -30,6 +30,9 @@ dependencies {
   implementation(project(":polaris-api-management-service"))
   implementation(project(":polaris-api-iceberg-service"))
   implementation(project(":polaris-api-catalog-service"))
+  if ((project.findProperty("NonRESTCatalogs") as String?)?.contains("HADOOP") == true) {
```
snazy: This is a breaking change to Polaris that effectively _removes_ functionality in the Polaris server.

snazy (commented):
Looks like this PR introduces a breaking change that _removes_ functionality.

runtime/service/build.gradle.kts
eric-maynard: IIUC the functionality is still there, but the way you need to build Polaris to use that functionality does change (across versions).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -220,42 +218,29 @@ protected void initializeCatalog() {
...
-              new RESTCatalog(
-                  context,
-                  (config) ->
-                      HTTPClient.builder(config)
-                          .uri(config.get(org.apache.iceberg.CatalogProperties.URI))
-                          .build());
-          federatedCatalog.initialize(
-              ((IcebergRestConnectionConfigInfoDpo) connectionConfigInfoDpo).getRemoteCatalogName(),
-              connectionConfigInfoDpo.asIcebergCatalogProperties(getUserSecretsManager()));
+          factoryIdentifier = "iceberg-rest";
```
eric-maynard: Not a fan of these magic strings which basically are just another flavor of `connectionType`.  Can we use `connectionType` directly or, barring that, use a constant? e.g. `HadoopFederatedCatalogFactory.CATALOG_TYPE = "hadoop"`

extensions/federation/hadoop/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+
+  // Logging
+  implementation(libs.slf4j.api)
+}
+
+java { toolchain { languageVersion.set(JavaLanguageVersion.of(21)) } }
+
+tasks.withType<JavaCompile> { options.encoding = "UTF-8" }
+
+tasks.test { useJUnitPlatform() }
```
adutra: These lines shouldn't be required.

runtime/service/build.gradle.kts
adutra: In any case, shouldn't this block be declared rather in `runtime/server`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
poojanilangekar: Done.

extensions/federation/hadoop/build.gradle.kts
poojanilangekar: Done.

runtime/service/build.gradle.kts
poojanilangekar: Regarding runtime/server module: We reference the ExternalCatalogFactory in the IcebergCatalogHandler which is in runtime/service module. So isn't this the correct location?

adutra: But afaict `ExternalCatalogFactory` is an interface in `polaris-core`. We don't reference `HadoopFederatedCatalogFactory` directly in code.

In `runtime/server` we already have other similar `runtimeOnly` declarations:

```
  runtimeOnly(project(":polaris-eclipselink"))
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -220,42 +218,27 @@ protected void initializeCatalog() {
...
-          federatedCatalog = new HadoopCatalog(conf, warehouse);
-          federatedCatalog.initialize(
-              warehouse,
-              connectionConfigInfoDpo.asIcebergCatalogProperties(getUserSecretsManager()));
+          factoryIdentifier = connectionType.getFactoryIdentifier();
           break;
         default:
-          throw new UnsupportedOperationException(
-              "Connection type not supported: " + connectionType);
+          throw new UnsupportedOperationException("Unsupported connection type: " + connectionType);
```
eric-maynard: Sorry to keep asking for changes here, but it looks like the `switch` is basically not needed anymore. Can we try to remove the `switch`, add error handling to `getFactoryIdentifier` and use something like:
```java
Instance<ExternalCatalogFactory> externalCatalogFactory =
          externalCatalogFactoryRegistry.select(connectionType. getFactoryIdentifier())
```

poojanilangekar: Done.

eric-maynard (dismissed):

runtime/service/build.gradle.kts
poojanilangekar: I had that until https://github.com/apache/polaris/pull/2301/commits/c73966cda2d73396aeda70738b1b7133a545ba24 

I can do either, from the Polaris OSS sync last week (discussion about regarding hive federation), my takeaway was that we wanted to avoid having the default Polaris JAR depend on anything hadoop. 


However, if you'd prefer compiling it each time (and only loading if necessary), I can revert that change. I will send out a separate PR without dynamic compilation and update a README.md for this PR. Please pick the option that's best suited according to you.

dennishuo: +1 to adding something to the README/getting-started or similar to better make it an easy copy/paste command to decide whether to compile with or without the extended dependencies.

My understanding is indeed that this check was directly to address the concern others had about having Hadoop (or Hive in the future) compile-time dependencies be always present for all Polaris builds.

Personally I don't feel too strongly either way, so I'm okay with or without the additional compilation property.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -220,42 +218,18 @@ protected void initializeCatalog() {
...
-              connectionConfigInfoDpo.asIcebergCatalogProperties(getUserSecretsManager()));
-          break;
-        default:
-          throw new UnsupportedOperationException(
-              "Connection type not supported: " + connectionType);
+      // Use the unified factory pattern for all external catalog types
+      Instance<ExternalCatalogFactory> externalCatalogFactory =
+          externalCatalogFactories.select(
+              Identifier.Literal.of(connectionType.getFactoryIdentifier()));
+      if (!externalCatalogFactory.isUnsatisfied()) {
```
dennishuo: Generally I like to avoid double-negatives where possible, so could be worth considering swapping the `if/else` here to get rid of the double-negative.

dennishuo (dismissed):

runtime/service/build.gradle.kts
dennishuo: Also, per offline discussion, +1 to what @adutra said about putting the Hadoop extension into `runtime/server` if it works correctly for Quarkus finding it in the runtime assembly.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
poojanilangekar: Done, I've used `isResolvable()` instead.

runtime/service/build.gradle.kts
poojanilangekar: Done.

dennishuo (approved):

poojanilangekar:
@snazy I've sent out an alternate in #2332 which does not remove the functionality by default but rather creates a separate module that can be removed at a later point. If you oppose removing the functionality, then simply modularizing it makes no user visible changes. We need one of these two options to move forward with Hive federation (which I've proposed over a month ago and we agreed upon during the last community sync). Hence I request you to pick one of the two options so we can make progress on supporting various federation options for our users.

PR #2302, opened 2025-08-08
Add type-check to PolarisEntity ctors

XN137:
WIP

PR #2304, opened 2025-08-08
Fix TableIdentifier in TaskFileIOSupplier

XN137:
we cant just convert a `TaskEntity` to a `IcebergTableLikeEntity` as the
`getTableIdentifier()` method will not return a correct value by using
the name of the task and its parent namespace (which is empty?).

task handlers instead need to pass in the `TableIdentifier` that they
already inferred via `TaskEntity.readData`.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/task/BatchFileCleanupTaskHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/task/BatchFileCleanupTaskHandler.java
@@ -53,7 +53,7 @@ public class BatchFileCleanupTaskHandler extends FileCleanupTaskHandler {
     BatchFileCleanupTask cleanupTask = task.readData(BatchFileCleanupTask.class);
     TableIdentifier tableId = cleanupTask.tableId();
     List<String> batchFiles = cleanupTask.batchFiles();
-    try (FileIO authorizedFileIO = fileIOSupplier.apply(task, callContext)) {
+    try (FileIO authorizedFileIO = fileIOSupplier.apply(task, tableId, callContext)) {
       List<String> validFiles =
           batchFiles.stream().filter(file -> TaskUtils.exists(file, authorizedFileIO)).toList();
       if (validFiles.isEmpty()) {
--- a/runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
@@ -21,6 +21,7 @@ package org.apache.polaris.service.task;
 import java.time.Clock;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -36,7 +37,6 @@ import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.context.CallContext;
 import org.apache.polaris.core.entity.AsyncTaskType;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
-import org.apache.polaris.core.entity.PolarisEntity;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.TaskEntity;
 import org.apache.polaris.core.entity.table.IcebergTableLikeEntity;
@@ -71,21 +71,19 @@ public class TableCleanupTaskHandler implements TaskHandler {
 
   @Override
   public boolean canHandleTask(TaskEntity task) {
-    return task.getTaskType() == AsyncTaskType.ENTITY_CLEANUP_SCHEDULER && taskEntityIsTable(task);
+    return task.getTaskType() == AsyncTaskType.ENTITY_CLEANUP_SCHEDULER
+        && tryGetTableEntity(task).isPresent();
   }
 
-  private boolean taskEntityIsTable(TaskEntity task) {
-    PolarisEntity entity = PolarisEntity.of((task.readData(PolarisBaseEntity.class)));
-    return entity.getType().equals(PolarisEntityType.TABLE_LIKE);
+  private Optional<IcebergTableLikeEntity> tryGetTableEntity(TaskEntity task) {
+    return Optional.ofNullable(task.readData(PolarisBaseEntity.class))
+        .filter(entity -> entity.getType().equals(PolarisEntityType.TABLE_LIKE))
+        .map(IcebergTableLikeEntity::of);
   }
 
   @Override
   public boolean handleTask(TaskEntity cleanupTask, CallContext callContext) {
-    PolarisBaseEntity entity = cleanupTask.readData(PolarisBaseEntity.class);
-    PolarisMetaStoreManager metaStoreManager =
-        metaStoreManagerFactory.getOrCreateMetaStoreManager(callContext.getRealmContext());
-    IcebergTableLikeEntity tableEntity = IcebergTableLikeEntity.of(entity);
-    PolarisCallContext polarisCallContext = callContext.getPolarisCallContext();
+    IcebergTableLikeEntity tableEntity = tryGetTableEntity(cleanupTask).orElseThrow();
     LOGGER
         .atInfo()
         .addKeyValue("tableIdentifier", tableEntity.getTableIdentifier())
@@ -95,7 +93,8 @@ public class TableCleanupTaskHandler implements TaskHandler {
     // It's likely the cleanupTask has already been completed, but wasn't dropped successfully.
     // Log a
     // warning and move on
-    try (FileIO fileIO = fileIOSupplier.apply(cleanupTask, callContext)) {
+    try (FileIO fileIO =
+        fileIOSupplier.apply(cleanupTask, tableEntity.getTableIdentifier(), callContext)) {
       if (!TaskUtils.exists(tableEntity.getMetadataLocation(), fileIO)) {
         LOGGER
             .atWarn()
@@ -107,6 +106,10 @@ public class TableCleanupTaskHandler implements TaskHandler {
 
       TableMetadata tableMetadata =
           TableMetadataParser.read(fileIO, tableEntity.getMetadataLocation());
+
+      PolarisMetaStoreManager metaStoreManager =
+          metaStoreManagerFactory.getOrCreateMetaStoreManager(callContext.getRealmContext());
+      PolarisCallContext polarisCallContext = callContext.getPolarisCallContext();
 
       Stream<TaskEntity> manifestCleanupTasks =
           getManifestTaskStream(
--- a/runtime/service/src/main/java/org/apache/polaris/service/task/ManifestFileCleanupTaskHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/task/ManifestFileCleanupTaskHandler.java
@@ -63,7 +63,7 @@ public class ManifestFileCleanupTaskHandler extends FileCleanupTaskHandler {
   public boolean handleTask(TaskEntity task, CallContext callContext) {
     ManifestCleanupTask cleanupTask = task.readData(ManifestCleanupTask.class);
     TableIdentifier tableId = cleanupTask.tableId();
-    try (FileIO authorizedFileIO = fileIOSupplier.apply(task, callContext)) {
+    try (FileIO authorizedFileIO = fileIOSupplier.apply(task, tableId, callContext)) {
       ManifestFile manifestFile = decodeManifestData(cleanupTask.manifestFileData());
       return cleanUpManifestFile(manifestFile, authorizedFileIO, tableId);
     }
--- a/runtime/service/src/main/java/org/apache/polaris/service/task/TaskFileIOSupplier.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/task/TaskFileIOSupplier.java
@@ -24,21 +24,21 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.BiFunction;
+import org.apache.commons.lang3.function.TriFunction;
 import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.io.FileIO;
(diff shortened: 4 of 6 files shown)


runtime/service/src/main/java/org/apache/polaris/service/task/BatchFileCleanupTaskHandler.java
```diff
@@ -53,7 +53,7 @@ public boolean handleTask(TaskEntity task, CallContext callContext) {
     BatchFileCleanupTask cleanupTask = task.readData(BatchFileCleanupTask.class);
     TableIdentifier tableId = cleanupTask.tableId();
     List<String> batchFiles = cleanupTask.batchFiles();
-    try (FileIO authorizedFileIO = fileIOSupplier.apply(task, callContext)) {
+    try (FileIO authorizedFileIO = fileIOSupplier.apply(task, tableId, callContext)) {
```
eric-maynard, line 56: Should the tableId just be a part of the TaskEntity?

XN137: depends on whether there will ever be tasks that dont operate on a single iceberg table... currently the `TaskEntity` design seems to leave this open

XN137: thinking about this again, since the `TableIdentifier` is downstream only getting used for logging, it would seem an disproportionate amount of effort to change the `TaskEntity` to avoid passing in the tableidentifier here (that is already available in each task handler)

eric-maynard: Why would the existence of a task that involves multiple tables mean that a given task can't include a table name in its properties?

XN137: sorry i might not be fully understanding what you had been asking for.
could you please provide a more detailed design of how we would be making the tableId part of the TaskEntity?
and then also why that would be a better solution that simply passing in the tableId that is already available in the task handlers? (which is taken from `TaskEntity.readData` so its already part of the `TaskEntity` in a way)

eric-maynard: If the TableIdentifier is a piece of information relevant to the task, the task itself should contain that information. I don't see why we would break an interface instead of just adding that information.

XN137: it already contains that information, readable via `TaskEntity.readData`.

please provide a detailed suggestion of how you would want the `TaskEntity` to hold the table identifier in a different way, as I am unable to comprehend what you are asking for and how that would be an improvement.

`TaskEntity` is a `PolarisEntity` and thus afaict the only places where we can add "new information" is in the `properties` or `internalProperties` fields and `TaskEntity.readData` is already stored in the `properties` and holds the `TableIdentifier` (for tasks that need it).

eric-maynard: > it already contains that information, readable via TaskEntity.readData.

If a table identifier is indeed already in TASK_DATA already for a given task, then this change is not needed. Task handling code should extract the TASK_DATA with `readData`, which should contain a TableIdentifier (possibly via a TableLikeEntity) if the handling code needs a TableIdentifier. The `TaskFileIOSupplier` type should not require modification.

> TaskEntity is a PolarisEntity and thus afaict the only places where we can add "new information" is in the properties or internalProperties fields and TaskEntity.readData is already stored in the properties and holds the TableIdentifier (for tasks that need it).

Yes, that's true. What's the question here?

XN137: > Task handling code should extract the TASK_DATA with readData, which should contain a TableIdentifier (possibly via a TableLikeEntity) if the handling code needs a TableIdentifier. 

this is already happening: all callers of `TaskFileIOSupplier.apply` are task handlers and they already get the `TableIdentifier` via `TaskEntity.readData`. with this PR they are simply passing that `TableIdentifier` into the `TaskFileIOSupplier`.

> The TaskFileIOSupplier type should not require modification.

the `TaskFileIOSupplier` is what is requiring a `TableIdentifer` for logging purposes (as its a parameter for `FileIOFactory.loadFileIO`).
without this PR task handlers are simply logging the wrong value because it treats a `TaskEntity` as a `IcebergTableLikeEntity` and generates a wrong `TableIdentifier` (pointing to the task entity, instead of the table).

> Yes, that's true. What's the question here?

I am trying to explain to you that what you are asking for is already happening i.e. tasks already store the `TableIdentifier` in the `TASK_DATA` property of the `TaskEntity`.

eric-maynard: > without this PR task handlers are simply logging the wrong value because it treats a TaskEntity as a IcebergTableLikeEntity and generates a wrong TableIdentifier (pointing to the task entity, instead of the table).

Yes, this logic is definitely wrong -- but we should fix it to extract the TASK_DATA and get the qualified table identifier from there. AFAICT that doesn't require any change to `TaskFileIOSupplier` -- am I missing something?

XN137: I had asked you multiple times already for a more detailed suggestion of how you would want to achieve that.

So I can just guess that you want to add something like a `Optional<TableIdentifier> getTableIdentifier()` method to `TaskEntity` ?
but when we think about how it would need to be implemented, it would mean that it has to contain full knowledge of ALL TYPES of tasks and what kind of objects each one stores in `TASK_DATA` to de-serialize them and get the `TaskIdentifier`.

but that same knowledge ALREADY EXISTS in the task handlers (and its where we already have the `TableIdentifier` alongside the other task parameters), so again, just letting the task handlers pass that value into the `TaskFileIOSupplier` seems like the right approach to me.

can you help me understand: why is adding another parameter to `TaskFileIOSupplier` so bad it has to block the fixing of a bug?

as an alternative we could also just stop passing `TableIdentifier` into `FileIOFactory.loadFileIO` as downstream its only for logging (if there is an error creating the fileIO, the task handlers could log the tableIdentifier they already have as well alongside the exception)...

eric-maynard: > So I can just guess that you want to add something like a Optional<TableIdentifier> getTableIdentifier() method to TaskEntity ?

I don't think that's necessary, we could just add the fully-qualified table identifier to the task as a property and extract it (if it's present) when it's needed for logging.

> why is adding another parameter to TaskFileIOSupplier so bad it has to block the fixing of a bug?

As you noted, this is just fixing some logging -- and it's also a PR against main not a release. Hardly a time-sensitive bugfix, so let's do this the right way.

> as an alternative we could also just stop passing TableIdentifier into FileIOFactory.loadFileIO as downstream its only for logging 

I think this is fine too, if the change to add the tableidentifier as a property is a headache.

eric-maynard: Do you have a good way to test this, by the way? I ran through the existing `TableCleanupTaskHandlerTest` with the debugger and the logged identifiers (e.g. `db1.schema1.table1`) look correct

XN137: > we could just add the fully-qualified table identifier to the task as a property and extract it (if it's present) when it's needed for logging

so you suggest to have another `TaskEntity` property and duplicate the value that is already there in `TASK_DATA`.
would that mean we remove it from `TASK_DATA` or keep it there as well?

will the new property be optional or required?
if optional, should the `TaskFileIOSupplier` fail hard if its missing?
if required, are there any persisted task entities out there that need to be "migrated" somehow?

> Hardly a time-sensitive bugfix, so let's do this the right way.

this is delaying some follow-up work i was doing i.e. https://github.com/apache/polaris/pull/2302 but yeah as stated above, the right way for me is passing the already available value.

> I think this is fine too, if the change to add the tableidentifier as a property is a headache.

so changing the `FileIOFactory` interface would be fine but changing `TaskFileIOSupplier` is a no-go? whats the criteria?

>  I ran through the existing TableCleanupTaskHandlerTest with the debugger and the logged identifiers (e.g. db1.schema1.table1) look correct

not sure what you did locally but for me the identifier is wrong in the debugger:
Screenshot From 2025-08-13 10-27-33
note how it has an empty `namespace` and the `name` is just the task name.

eric-maynard: In your screenshot, `identifier.toString` looks correct to me -- `cleanup_db1.schema1.table1`. Should it be something else?

XN137: > In your screenshot, identifier.toString looks correct to me -- cleanup_db1.schema1.table1. Should it be something else?

in the test we are building the `TaskEntity` like so:
https://github.com/apache/polaris/blob/ef177b51cad07411d6b2b673fe3b5cc8a01c6d70/runtime/service/src/test/java/org/apache/polaris/service/task/TableCleanupTaskHandlerTest.java#L110-L112

so the `cleanup_` prefix being in the `TableIdentifier` is not correct.
also a proper `TableIdentifier` would have a non-empty `Namespace` of `db1.schema1`:
https://github.com/apache/polaris/blob/ef177b51cad07411d6b2b673fe3b5cc8a01c6d70/runtime/service/src/test/java/org/apache/polaris/service/task/TableCleanupTaskHandlerTest.java#L93

eric-maynard: Ah, sorry, I completely missed that. Anyway, [here](https://github.com/apache/polaris/compare/main...eric-maynard:polaris:fix-task-tableid?expand=1) is a one-line fix that doesn't change any APIs. From my debugger:
Screenshot 2025-08-13 at 11 15 05 AM

eric-maynard: This correctly extracts the TableIdentifier (via the TableLikeEntity) from the TASK_DATA like I mentioned above.

XN137: > Anyway, [here](https://github.com/apache/polaris/compare/main...eric-maynard:polaris:fix-task-tableid?expand=1) is a one-line fix that doesn't change any APIs

afaict your "one-line fix" only works for table cleanup tasks, where we happen to store a `IcebergTableLikeEntity` into the `TASK_DATA`:
https://github.com/apache/polaris/blob/1277eff3e1e400821a73015d3c3b3c7eb4831dbb/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java#L1433-L1437

for other tasks, it would be throwing an error as we store something else, for example:
https://github.com/apache/polaris/blob/20febdaede19fb7c46e120652fdd1a262c2138e4/runtime/service/src/main/java/org/apache/polaris/service/task/ManifestFileCleanupTaskHandler.java#L64-L66

eric-maynard: Yeah, that's what I was trying to say above -- if there is a TableIdentifier in the TASK_DATA we should be extracting and using that. The current "cast" of a TaskEntity into a TableLikeEntity is definitely wrong, but if we can fix this the easy way we should.

If there is a task that _doesn't_ have a TableIdentifier in TASK_DATA we should either add it there or remove the need for it (i.e. pass in `null` to `loadFileIO` and mark it `Nullable`) if that's easier.

XN137: > if there is a TableIdentifier in the TASK_DATA we should be extracting and using that.

it seems like we have circled back to a previous point in our discussion, so i am posting my reply again:

> So I can just guess that you want to add something like a Optional<TableIdentifier> getTableIdentifier() method to TaskEntity ?
but when we think about how it would need to be implemented, it would mean that it has to contain full knowledge of ALL TYPES of tasks and what kind of objects each one stores in TASK_DATA to de-serialize them and get the TaskIdentifier.
>
> but that same knowledge ALREADY EXISTS in the task handlers (and its where we already have the TableIdentifier alongside the other task parameters), so again, just letting the task handlers pass that value into the TaskFileIOSupplier seems like the right approach to me.

i still dont have an answer as to why changing `TaskFileIOSupplier` is a no-go when all callers have the `TableIdentifier` already and the underlying `FileIOFactory` (in its current form) requires it.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
```diff
@@ -71,21 +71,19 @@ public TableCleanupTaskHandler(
...
   public boolean canHandleTask(TaskEntity task) {
-    return task.getTaskType() == AsyncTaskType.ENTITY_CLEANUP_SCHEDULER && taskEntityIsTable(task);
+    return task.getTaskType() == AsyncTaskType.ENTITY_CLEANUP_SCHEDULER
+        && tryGetTableEntity(task).isPresent();
   }
 
-  private boolean taskEntityIsTable(TaskEntity task) {
-    PolarisEntity entity = PolarisEntity.of((task.readData(PolarisBaseEntity.class)));
-    return entity.getType().equals(PolarisEntityType.TABLE_LIKE);
+  private Optional<IcebergTableLikeEntity> tryGetTableEntity(TaskEntity task) {
```
dimas-b, line 78: nit: the "try" prefix makes me wonder whether the method may or may not produce an "effect" (as in `Semaphore.tryAcquire()`), but in fact it always operates the same way without side effects... I'd prefer something like `asTableEntity(TaskEntity task)` or `asOptionalTableEntity()`.

dimas-b (approved):
LGTM :+1:

PR #2307, opened 2025-08-08
Refactor Authenticator and PolarisPrincipal

adutra:
This PR contains no added or modified functionality.

The main goal of this change is to facilitate future integration of federated principals:

- `AuthenticatedPolarisPrincipal` becomes an interface `PolarisPrincipal`, as the original class leaks implementation details (references to `PrincipalEntity` and thus to the storage layer). The new interface does not reference the storage layer. This is one step further towards easy pluggability of authentication in Polaris.

- The `Authenticator.authenticate()` method does not return an `Optional` anymore, as this was ambiguous (returning `Optional.empty()` vs throwing `NotAuthorizedException`).

- Also the `Authenticator` interface is not generic anymore. This was an artifact of times when there were two kinds of `Authenticators` in Polaris (one for internal auth, the other for external) and is not necessary anymore.


adutra:
FYI @collado-mike @dennishuo @flyrain

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
```diff
@@ -0,0 +1,89 @@
...
+        .name(name)
+        .properties(properties)
+        .roles(roles)
+        .build();
+  }
+
+  /**
+   * Returns the unique identifier of the principal.
+   *
+   * <p>This identifier is used to uniquely identify the principal within the Polaris system.
```
dimas-b: It's unique within a realm, though, right?

dimas-b (dismissed):

collado-mike:
I'd like to take a look at this, but I won't have cycles until Friday. Can we please hold off on merging before then?

runtime/service/src/main/java/org/apache/polaris/service/auth/external/OidcPolarisCredentialAugmentor.java
```diff
@@ -78,26 +78,20 @@ public Uni<SecurityIdentity> augment(
             .select(Identifier.Literal.of(config.principalRolesMapper().type()))
             .get();
     return Uni.createFrom()
-        .item(() -> setPrincipalAuthInfo(identity, principalMapper, principalRolesMapper));
+        .item(() -> setPolarisCredential(identity, principalMapper, principalRolesMapper));
```
snazy: Nit: mind adding a comment that the two mappers may do expensive work, hence justifying the async call here?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
```diff
@@ -0,0 +1,89 @@
...
+        .roles(roles)
+        .build();
+  }
+
+  /**
+   * Returns the unique identifier of the principal.
+   *
+   * <p>This identifier is used to uniquely identify the principal within a Polaris realm.
+   */
+  long getId();
```
snazy, line 73: We differentiate between persisted (`PersistedPolarisPrincipal`) and non-persistent ones. I wonder whether this getter should be present at all here. It could at least return an `OptionalLong` to make it clear that only persisted principals have an ID. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/auth/PersistedPolarisPrincipal.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Override
+  public final long getId() {
+    return getEntity().getId();
+  }
+
+  @Override
+  public final Map<String, String> getProperties() {
+    return getEntity().getInternalPropertiesAsMap();
```
snazy, line 64: Maybe make this a `@Value.Lazy` so the implicit JSON deserialization happens only once.

runtime/service/src/main/java/org/apache/polaris/service/auth/PolarisCredential.java
```diff
@@ -21,19 +21,25 @@
...
+public interface PolarisCredential extends Credential {
+
+  static PolarisCredential of(
+      @Nullable Long getPrincipalId,
+      @Nullable String getPrincipalName,
+      Set<String> getPrincipalRoles) {
+    return ImmutablePolarisCredential.builder()
+        .principalId(getPrincipalId)
+        .principalName(getPrincipalName)
+        .principalRoles(getPrincipalRoles)
```
snazy:
```suggestion
  static PolarisCredential of(
      @Nullable Long principalId,
      @Nullable String principalName,
      Set<String> principalRoles) {
    return ImmutablePolarisCredential.builder()
        .principalId(principalId)
        .principalName(principalName)
        .principalRoles(principalRoles)
```

snazy (commented):
Overall LGTM, it's a mostly "boring" change.

I wonder though whether `PolarisPrincipal` should have the `id` property removed or become optional (see reason below).

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
adutra: The problem is that the Resolver requires the principal to have an id, since it resolves principals by id. Thus it seemed to me that it was legit to require from all principals to have at least a numeric identifier.

runtime/service/src/main/java/org/apache/polaris/service/auth/external/OidcPolarisCredentialAugmentor.java
adutra: You are right but in this case we need `context.runBlocking`

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
adutra: As a follow-up task, we could attempt to modify the Resolver so that it makes principal lookups by name. WDYT @collado-mike ?

dimas-b:
Shall we mark it as "draft" until @collado-mike reviews? (just to avoid the impression that it's mergeable)

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
collado-mike: I think we should be able to move past the Resolver needing to resolve the principal at all. This was largely present because the Resolver fetched the accessible PrincipalRoles based on grant records, but we have moved that requirement out so that we can rely on roles defined in the token instead. I think it's reasonable to change the Resolver to skip resolving the principal and roles entirely.

runtime/service/src/main/java/org/apache/polaris/service/auth/DefaultActiveRolesProvider.java
```diff
@@ -54,11 +55,18 @@ public class DefaultActiveRolesProvider implements ActiveRolesProvider {
   @Inject MetaStoreManagerFactory metaStoreManagerFactory;
 
   @Override
-  public Set<String> getActiveRoles(AuthenticatedPolarisPrincipal principal) {
+  public Set<String> getActiveRoles(PolarisPrincipal principal) {
+    if (!(principal instanceof PersistedPolarisPrincipal persistedPolarisPrincipal)) {
```
collado-mike, line 59: huh. I thought the variable here was always scoped to within the if block! TIL

collado-mike (approved):
Thanks for holding off. This looks good to me.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
adutra: Oh that's an interesting improvement for sure! I will try to do that in a follow-up task 👍

adutra:
Had to rebase to fix merge conflicts – PTAL again.

PR #2315, opened 2025-08-11
Doc: Add example on using Polaris with PyIceberg

FredKhayat:
It is not obvious how to connect PyIceberg to a Polaris catalog.

This PR clears that up by providing an example in the getting-started section of the documentation.


dimas-b:
If no other comments, I suppose this can be merged tomorrow.

site/content/in-dev/unreleased/getting-started/using-polaris.md
```diff
@@ -285,6 +285,51 @@ SELECT * FROM iceberg.quickstart_schema.quickstart_table;
...
+catalog = load_catalog(
+    type='rest',
+    uri='http://localhost:8181/api/catalog',
+    warehouse='quickstart_catalog',
+    token=token,
+)
+```
+
+It is possible to use `load_catalog` function by providing an authorization token directly. This method is useful when using an external identity provider (e.g. Google Identity).
+
```
flyrain, line 332: I'd suggest to provide some sample commands of pyIceberg, like list table, list files, etc. Not a blocker though.

eric-maynard: eh, the other sections for other clients doesn't get into too much detail either -- it could be nice to have, but I think a regtest would be even better to cover a full set of operations in pyiceberg. The docs can stay lightweight

PR #2317, opened 2025-08-11
Add entitySubType param to BasePersistence.listEntities

XN137:
`BasePersistence.listEntities` has 3 variants:
```
Page<EntityNameLookupRecord> listEntities(..., PageToken);

Page<EntityNameLookupRecord> listEntities(..., Predicate<PolarisBaseEntity>, PageToken)

<T> Page<T> listEntities(..., Predicate<PolarisBaseEntity>, Function<PolarisBaseEntity, T>, PageToken);
```

the 1st method exists to only return the subset of entity properties required to build an `EntityNameLookupRecord`.

the 3rd method supports a predicate and transformer function on the underlying `PolarisBaseEntity`, which means it has to load all entity properties.

the 2nd method is weird as it supports a full `Predicate<PolarisBaseEntity>`, which means it has to load all entity properties under the hood for filtering but then throws most of them away to return a `EntityNameLookupRecord`.
this explains why the implementations of the 2nd method simply forward to the 3rd method usually.
any performance benefits of returning a `EntityNameLookupRecord` are lost.

as it turns out the 2nd method is only used, because methods 1 and 3 dont support passing a `PolarisEntitySubType` parameter to filter down the retrieved data.
Note that the sub type property is available from both the `PolarisBaseEntity` as well as the `EntityNameLookupRecord`.

By adding this parameter, the 2nd method can go away completely.
we can even push down the sub type filtering into the queries of some of our persistence implementations.
other existing implementations are free to decide whether they want to push it down as well or filter on the query results in memory.

note that since we have no `TransactionalPersistence` implementation in the codebase that provides an optimized variant of method 1 we can have a default method in the interface that forwards to method 3.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -44,6 +44,7 @@ import org.apache.polaris.core.entity.PolarisChangeTrackingVersions;
 import org.apache.polaris.core.entity.PolarisEntity;
 import org.apache.polaris.core.entity.PolarisEntityCore;
 import org.apache.polaris.core.entity.PolarisEntityId;
+import org.apache.polaris.core.entity.PolarisEntitySubType;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisGrantRecord;
 import org.apache.polaris.core.entity.PolarisPrincipalSecrets;
@@ -427,32 +428,16 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
       long catalogId,
       long parentId,
       @Nonnull PolarisEntityType entityType,
+      @Nonnull PolarisEntitySubType entitySubType,
       @Nonnull PageToken pageToken) {
+    // TODO: only fetch the properties required for creating an EntityNameLookupRecord
     return listEntities(
         callCtx,
         catalogId,
         parentId,
         entityType,
+        entitySubType,
         entity -> true,
-        EntityNameLookupRecord::new,
-        pageToken);
-  }
-
-  @Nonnull
-  @Override
-  public Page<EntityNameLookupRecord> listEntities(
-      @Nonnull PolarisCallContext callCtx,
-      long catalogId,
-      long parentId,
-      @Nonnull PolarisEntityType entityType,
-      @Nonnull Predicate<PolarisBaseEntity> entityFilter,
-      @Nonnull PageToken pageToken) {
-    return listEntities(
-        callCtx,
-        catalogId,
-        parentId,
-        entityType,
-        entityFilter,
         EntityNameLookupRecord::new,
         pageToken);
   }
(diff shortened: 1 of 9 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -710,20 +709,14 @@ private void bootstrapPolarisService(
...
-
     // return list of active entities
     Page<EntityNameLookupRecord> resultPage =
         ms.listEntitiesInCurrentTxn(
             callCtx,
             resolver.getCatalogIdOrNull(),
             resolver.getParentId(),
             entityType,
-            filter,
+            entitySubType,
```
XN137, line 719: note that due to this change calls to `PolarisMetaStoreManager.listEntities` like here:
https://github.com/apache/polaris/blob/2f985ab93ccd65b0c98fdd8902e217792c622a68/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L2574-L2591
no longer have to load the full `PolarisBaseEntity` to apply the `Predicate<PolarisBaseEntity>` and thus they can actually take advantage of optimized implementations of `Page<EntityNameLookupRecord> listEntities` in `BasePersistence` that load only the required properties.

adutra (approved):
Awesome improvement!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -427,43 +428,28 @@ public Page<EntityNameLookupRecord> listEntities(
       long catalogId,
       long parentId,
       @Nonnull PolarisEntityType entityType,
+      @Nonnull PolarisEntitySubType entitySubType,
       @Nonnull PageToken pageToken) {
+    // TODO: only fetch the properties required for creating an EntityNameLookupRecord
```
dennishuo, line 433: To help us remember as well, can we add `and update the underlying CONSTRAINT clause to make the implicit index INCLUDE the subTypeCode` to this TODO?

dennishuo (approved):
LGTM, thanks!

I also would like to adjust the transformer/filter version of `listEntities` to better reflect it as less of an overloaded method and more of a totally distinct method, but perhaps that fits in better in your other PR https://github.com/apache/polaris/pull/2290

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
XN137: i'll take a stab at this as an immediate follow-up or file it as a separate issue where i will take note of this additional "index coverage" requirement

dimas-b (approved):

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
XN137: filed the ticket: https://github.com/apache/polaris/issues/2352

dimas-b: Related `dev` ML discussion: https://lists.apache.org/thread/kxyqp287zmxmorod7l205ojkwfqgrhkv

Issue #85, opened 2024-08-03
[FEATURE REQUEST] Add HDFS storage option for catalogs


rdsarvar:
**Is your feature request related to a problem? Please describe.**
Currently it appears that the storage options are geared towards cloud providers. To support companies running on premise I would like to request HDFS support.

**Describe the solution you'd like**
Catalog R/W support HDFS as a storage option.

flyrain:
We can add HDFS support, but handling credentials might be difficult. @sfc-gh-schen, can you provide more insight? I think it may not be possible for credential vending.

rdsarvar:
Would permissive HDFS be a simpler first ask with potential evolutions in the future to support credentials? Aka start off assuming that the HDFS cluster is open (no Kerberos, etc), running on-premise protected by networking, and then consider a proper strategy in the future?

Sorry if I'm off the mark on what you meant by your last comment but I assumed it's relating to authn/authz against HDFS (and not mapping to an internal strategy in Polaris).

flyrain:
Yup, it's reasonable to start with non-authentication and non-authorization for HDFS.

rdsarvar:
I don't mind trying my hand at this for the simple case and provide a baseline for people to extend in the future.

I think in order for this to work, due to the DFS is created inside of Iceberg core with the Hadoop configuration object we initialize, we'll need to rely on the `HADOOP_USER_NAME` environment variable (which will mean all DFS interaction would be under the same username). Otherwise it'll default access to be the user running the service.

I can see if I find time one of these days to throw something together in a PR (as an non-tested PoC) just to get feedback on if it's something we want to move forward

rdsarvar:
@flyrain QQ about the repo which I noticed starting to write the changes required by this MR:

1. Do the regtests have Open API Python templates that aren't commit into the repo? When I generate the files with the provided commands in the README it results in generated Python files without the license header and in some cases it breaks the functionality. An example diff of `regtests/client/python/polaris/management/models/aws_storage_config_info.py`:
```
         _obj = cls.model_validate({
             "storageType": obj.get("storageType"),
-            "allowedLocations": obj.get("allowedLocations"),
-            "roleArn": obj.get("roleArn")
+            "allowedLocations": obj.get("allowedLocations")
         })
```

[...]

PR #2322, opened 2025-08-11
feat: adding unauthenticated HDFS storage support to catalogs (#85)

rdsarvar:
# TODO - Description

Initially contains per catalog resources and username configurations that enable Polaris to configure and impersonate against an HDFS cluster. This initial unauthenticated workflow supports #85

Draft MR opened to showcase potential functionality of supporting HDFS leveraging internal catalog properties as the store of information. Please feel free to critique or suggest better implementation workflows

### DIFF
--- a/client/python/cli/command/__init__.py
+++ b/client/python/cli/command/__init__.py
@@ -82,7 +82,9 @@ class Command(ABC):
                 catalog_role_session_name=options_get(Arguments.CATALOG_ROLE_SESSION_NAME),
                 catalog_external_id=options_get(Arguments.CATALOG_EXTERNAL_ID),
                 catalog_signing_region=options_get(Arguments.CATALOG_SIGNING_REGION),
-                catalog_signing_name=options_get(Arguments.CATALOG_SIGNING_NAME)
+                catalog_signing_name=options_get(Arguments.CATALOG_SIGNING_NAME),
+                hadoop_resources=options_get(Arguments.HADOOP_RESOURCES),
+                hadoop_username=options_get(Arguments.HADOOP_USERNAME)
             )
         elif options.command == Commands.PRINCIPALS:
             from cli.command.principals import PrincipalsCommand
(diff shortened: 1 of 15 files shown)


snazy:
We have been (more precisely: are) discussing support for Hive, and implicitly HDFS, in Polaris.
There are a couple of concerns about adding (more) Hadoop stuff to Polaris.

Hadoop types like `org.apache.hadoop.security.UserGroupInformation` et al rely on functionality that has already been deprecated since Java 17 ([JEP-411](https://openjdk.org/jeps/411)) and is already  _removed_ in Java since version 24 ([JEP-486](https://openjdk.org/jeps/486))). Although we technically do not need 24 (or the upcoming Java 25) yet, we might have to.

As we want to support these use cases, the approach would be to have a Polaris instance that _federates_ to an instance, which can then have have hard "max Java version" restrictions but also do other "restrictive" things like Kerberos.

This all falls into the "catalog federation" bucket. We'd appreciate your input on the dev-mailing-list and in-person discussions in our biweekly community sync!

rdsarvar:
> We have been (more precisely: are) discussing support for Hive, and implicitly HDFS, in Polaris. There are a couple of concerns about adding (more) Hadoop stuff to Polaris.
> 
> Hadoop types like `org.apache.hadoop.security.UserGroupInformation` et al rely on functionality that has already been deprecated since Java 17 ([JEP-411](https://openjdk.org/jeps/411)) and is already _removed_ in Java since version 24 ([JEP-486](https://openjdk.org/jeps/486))). Although we technically do not need 24 (or the upcoming Java 25) yet, we might have to.
> 
> As we want to support these use cases, the approach would be to have a Polaris instance that _federates_ to an instance, which can then have have hard "max Java version" restrictions but also do other "restrictive" things like Kerberos.
> 
> This all falls into the "catalog federation" bucket. We'd appreciate your input on the dev-mailing-list and in-person discussions in our biweekly community sync!

Hi @snazy thanks for the response -- 

Interesting solution, so just to recap to make sure I'm on the same page the thought is to run a separate sidecar / service which itself can be pegged to a lower Java version and support any HDFS related operations there? So in essence the workflow would now look like:

1. User makes request to Polaris catalog (ex: new table)
2. Polaris determines the storage type is HDFS and proxies the request to the HDFS catalog service (REST API I assume)
3. The HDFS catalog service would perform required actions on behalf of Polaris, returning what happened through structured JSON response
4. Polaris would respond to the user and make any required persistence changes

Thus decoupling the potentially unsafe (or lesser version) Hadoop usage to users that are explicitly requiring it.

I'll join the Slack workspace, do you have reference to the channel (or thread) that this is being discussed? Or is it mostly in the syncs / mailing list

snazy:
That's (pretty much) correct.
But let's keep the discussion on the dev-ML.

rdsarvar:
Sounds good, for reference of anyone coming into this PR it can be found here:  https://lists.apache.org/thread/5qktjv6rzd8pghcl6f4oohko798o2p2g

PR #2327, opened 2025-08-12
Let CI archive html test reports

XN137:
when having to debug CI test failures its much more convenient to be able to download the html report compared to the XML reports (as the latter requires to you find the right file/failure manually).

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -73,6 +73,7 @@ jobs:
           name: upload-unit-test-artifacts
           path: |
             **/build/test-results/**
+            **/build/reports/tests/**
 
   quarkus-tests:
     name: Quarkus Tests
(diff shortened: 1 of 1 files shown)


XN137:
recent example from another PR:
image

.github/workflows/gradle.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         with:
           name: upload-unit-test-artifacts
           path: |
-            **/build/test-results/**
+            **/build/reports/tests/**
```
snazy, line 76: Both the test-results _and_ the reports are useful. Why not archive both?

XN137: ok added back

snazy:
I don't think this is in any way controversial, merging.

PR #2332, opened 2025-08-12
Modularize federation (Option 2)

poojanilangekar:
Based on the discussion in the community sync, this PR modularizes federation calls to a separate factory. The specific instances of the NonRESTCatalogFactory is loaded only when a non-REST catalog is initialized and accessed. Beyond the catalog initialization, all the calls are identical for internal and federated catalogs. 

Based on the discussion in #2301, this PR removes the compile-time flag used to build the `HadoopFederatedCatalogFactory`. 

Testing:
Applied the regtest in PR https://github.com/apache/polaris/pull/2286.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
@@ -20,6 +20,7 @@ package org.apache.polaris.core.connection;
 
 import jakarta.annotation.Nonnull;
 import java.util.Arrays;
+import java.util.Locale;
 
 /**
  * The internal persistence-object counterpart to ConnectionConfigInfo.ConnectionTypeEnum defined in
@@ -35,8 +36,8 @@ public enum ConnectionType {
   HADOOP(2),
   ;
 
-  public static final String ICEBERG_REST_FACTORY_IDENTIFIER = "ICEBERG_REST";
-  public static final String HADOOP_FACTORY_IDENTIFIER = "HADOOP";
+  public static final String ICEBERG_REST_FACTORY_IDENTIFIER = ICEBERG_REST.name().toLowerCase(Locale.ROOT);
+  public static final String HADOOP_FACTORY_IDENTIFIER = HADOOP.name().toLowerCase(Locale.ROOT);
 
   private static final ConnectionType[] REVERSE_MAPPING_ARRAY;
 
--- a/polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
@@ -35,6 +35,9 @@ public enum ConnectionType {
   HADOOP(2),
   ;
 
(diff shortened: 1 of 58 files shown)


poojanilangekar:
CC: @eric-maynard @dennishuo  @adutra

eric-maynard:
If @snazy is okay with using this approach for HMS federation, then I think it's preferable to option 1

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
```diff
@@ -35,6 +35,9 @@ public enum ConnectionType {
   HADOOP(2),
   ;
 
+  public static final String ICEBERG_REST_FACTORY_IDENTIFIER = "ICEBERG_REST";
+  public static final String HADOOP_FACTORY_IDENTIFIER = "HADOOP";
```
snazy: THose are superfluous, as those are the enum names, and can therefore be removed.

snazy (commented):
Yea, thanks for removing the conditional build parameter. One comment though. But otherwise I'm fine with it.

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
```diff
@@ -77,4 +80,22 @@ public enum ConnectionType {
...
+    switch (this) {
+      case ICEBERG_REST:
+        return ICEBERG_REST_FACTORY_IDENTIFIER;
+      case HADOOP:
+        return HADOOP_FACTORY_IDENTIFIER;
+      default:
+        throw new UnsupportedOperationException(
+            "No factory identifier for connection type: " + this);
+    }
+  }
```
adutra, line 100: You may want to just return `name().toLowerCase(Locale.ROOT)` – since most CDI identifiers in Polaris (if not all) are lower-case.

```suggestion
  public String getFactoryIdentifier() {
    return name().toLowerCase(Locale.ROOT);
  }
```

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
poojanilangekar: We still need these because we can't use ConnnectionType.ICEBERG_REST.name() or ConnectionType.HADOOP.name(). The annotations needs to be constant expressions. Here's the error for your reference.

The value for annotation attribute Identifier.value must be a constant expressionJava(536871525)

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionType.java
poojanilangekar: We can't use the name().toLowerCase() directly because of two reasons:

1. We need the annotations to be constants. (Please see my comment above)
2. We need to explicitly handle the NULL_TYPE/default case. 

To make things consistent, I've changed the declarations to use lower case.

dennishuo (approved):
LGTM

dennishuo:
Looks like other comments have also been addressed, merging now.

PR #2336, opened 2025-08-13
Replace CallContext with RealmConfig in CatalogEntity

XN137:
the concrete `setStorageConfigurationInfo` method in the `Builder` only uses the `RealmConfig` so we should pass that instead.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
@@ -40,8 +40,8 @@ import org.apache.polaris.core.admin.model.GcpStorageConfigInfo;
 import org.apache.polaris.core.admin.model.PolarisCatalog;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
 import org.apache.polaris.core.config.BehaviorChangeConfiguration;
+import org.apache.polaris.core.config.RealmConfig;
 import org.apache.polaris.core.connection.ConnectionConfigInfoDpo;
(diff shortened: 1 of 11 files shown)


snazy (dismissed):
+1 on this change, there's no functional change.

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -79,7 +79,7 @@ public static CatalogEntity of(PolarisBaseEntity sourceEntity) {
     return null;
   }
 
-  public static CatalogEntity fromCatalog(CallContext callContext, Catalog catalog) {
+  public static CatalogEntity fromCatalog(RealmConfig realmConfig, Catalog catalog) {
```
dimas-b, line 82: nit: I tend to think the PR title should mention "replace CallContext with RealmConfig ..." :)

dimas-b (approved):
Code changes LGTM :+1: Thx @XN137 !

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
XN137: ok changed the PR title

XN137:
rebased after trivial conflict in `CatalogEntityTest.java`

dimas-b (approved):

XN137:
rebased after trivial conflict in `PolarisStorageConfigurationInfo.java`

snazy (approved):

adutra (approved):

PR #2337, opened 2025-08-13
Reduce get realm config calls

XN137:
DRAFT as based on `https://github.com/apache/polaris/pull/2336` currently

classes with a `CallContext` field should call `getRealmConfig` once and
store it as a field as well.
the idea is that long term we would want to stop relying on the
`CallContext` itself but instead inject its individual items.


snazy (dismissed):

XN137:
had to rebase due to a conflict and also found more spots we can adjust (i.e. add `RealmConfig` to `TestServices`).
PTAL again

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -138,8 +138,8 @@ public class IcebergCatalogAdapter
           .build();
 
   private final RealmContext realmContext;
-  private final CallContext callContext;
   private final RealmConfig realmConfig;
+  private final CallContext callContext;
```
snazy: ? looks unused

XN137: its not unused, both fields are getting used. the PR just changes the order, but ok reverted this change

dimas-b (dismissed):
This change looks valuable for future direct injection of `RealmConfig` into objects that require it. Reducing dependency on `CallContext` also looks valuable to me from the perspective of leveraging CDI more and custom context code less, which in turn improves maintainability, IMHO.

dimas-b:
This change does not appear to change any code behaviour. Given that it has been in review for 5 days, I'm going to merge it later today, unless new comments are posted.

eric-maynard (dismissed):

dimas-b:
@XN137 : unfortunately, there's a merge conflict now... Could you fix it, please?

XN137:
rebased after tiny conflict in `CatalogHandler.java`

snazy (approved):

PR #2340, opened 2025-08-13
Make PolarisAuthorizer RequestScoped

XN137:
all methods in `PolarisAuthorizer` currently have a `CallContext` parameter.
in its only implementation only `CallContext.getRealmConfig` is getting used.

so since `PolarisAuthorizer` cant be used outside a request, we can simply make it request-scoped and inject the request-scoped `RealmConfig` directly.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
@@ -134,12 +134,6 @@ public class ServiceProducers {
   }
 
   @Produces
-  @ApplicationScoped
-  public PolarisAuthorizer polarisAuthorizer() {
-    return new PolarisAuthorizerImpl();
-  }
-
-  @Produces
   @Singleton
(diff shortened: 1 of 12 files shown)


runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -170,6 +164,12 @@ public RealmConfig realmConfig(CallContext callContext) {
     return callContext.getRealmConfig();
   }
 
+  @Produces
+  @RequestScoped
+  public PolarisAuthorizer polarisAuthorizer(RealmConfig realmConfig) {
+    return new PolarisAuthorizerImpl(realmConfig);
```
XN137, line 170: note: if any custom implementations of `PolarisAuthorizer` are out there and still need the `CallContext` for whatever reason, they are free to still inject it here.

snazy:
The PR got 4 approvals and no objections, merging.

PR #1164, opened 2025-03-12
add refresh credentials property to loadTableResult

wolflex888:
This PR will add necessary refresh-vended-credentials properties to `LoadTableResponse` to allow each IoFile will get appropriate properties to initiate `VendedCredentialsProvider`.

Issue #1325, opened 2025-04-06
Document how Polaris work with different cloud storage(S3, Azure Blob Storage, GCS).


flyrain:
### Is your feature request related to a problem? Please describe.

The topic was briefly mentioned in https://polaris.apache.org/in-dev/unreleased/configuration/ and https://polaris.apache.org/in-dev/unreleased/overview/#storage-configuration. They don't provide enough clear instruction how Polaris service provider to set them. Having a dedicated page with more detailed information will be helpful.

Abhijoyshirovmukherjee:
This is something necessary. Can anybody share any document or piece of how the polaris catalog would be integrated with the underlying storage to be queried with spark or flink.

Currently getting the following error while inserting or reading data from the iceberg table in blob using flink:
No File system for scheme: abfss

Note: Have tried to run it with spark with the same dependencies and it has worked for inserting but not for reading.

@flyrain

frHimanshu:
I'm up to give a pr on this issue.

adnanhemani:
Given #1435 merged, should we close this @flyrain ?

flyrain:
Per offline discussion with @adnanhemani, we will still need more doc on how different storage setup works.

mj006648:
Hi @flyrain, following up on your note that more docs are needed for storage backends beyond what landed in #1435. If no one is currently working on the Azure Blob Storage and GCS sides, I'd like to pick those up.

Tentative plan, mirroring the structure of the existing S3 / Spark docs:

- Azure Blob Storage configuration page
  - `STORAGE_CONFIG_INFO` example (tenant id, client id, secret vs. Workload Identity)
  - ADLS Gen2 endpoint conventions
  - Vended Credentials behavior (cross-ref to #418 if still relevant)
- Google Cloud Storage configuration page
  - service-account-key vs. ADC vs. HMAC
  - uniform vs. bucket-level IAM notes
- A short section on the client-side dependencies each backend needs (hadoop-azure, gcs-connector, etc.) when used via Spark / Trino / PyIceberg.

Does this match what you had in mind, or would you prefer a single consolidated "storage backends" page instead of per-backend pages? Happy to follow whichever structure fits the existing docs better.

flyrain:
Thanks for picking this up, @mj006648. My idea is to provide end to end instructions that include both Polaris server side and client side configuration for different object storage systems, including AWS S3, Azure, and GCS. What you described here is already pretty close.

[...]

PR #2341, opened 2025-08-13
add refresh credentials property to loadTableResult

jasonf20:
Rebased version of https://github.com/apache/polaris/pull/1164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,9 @@ at locations that better optimize for object storage.
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
 - Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
-automatic storage credential refresh per table on the client side.
+automatic storage credential refresh per table on the client side. Java client version >= 1.8.0 is required.
+The endpoint path is always returned when using vended credentials, but the client must configure 
+REFRESH\_ENDPOINT\_ENABLED = "true" in the catalog properties for the client to use the endpoint.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,9 +83,8 @@ at locations that better optimize for object storage.
 
 - Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
 automatic storage credential refresh per table on the client side. Java client version >= 1.8.0 is required.
-The endpoint path is always returned when using vended credentials, but the client must configure
-"refresh-credentials-enabled" = "true" in the catalog properties for the client to use the endpoint.
-Note, the "refresh-credentials-enabled" configuration prefix is specific to each storage provider.
+The endpoint path is always returned when using vended credentials, but clients must enable the 
+refresh-credentials flag for the desired storage provider.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,7 @@ at locations that better optimize for object storage.
 
 - Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
 automatic storage credential refresh per table on the client side. Java client version >= 1.8.0 is required.
-The endpoint path is always returned when using vended credentials, but the client must configure 
+The endpoint path is always returned when using vended credentials, but the client must configure
 REFRESH\_ENDPOINT\_ENABLED = "true" in the catalog properties for the client to use the endpoint.
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,7 +84,8 @@ at locations that better optimize for object storage.
 - Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
 automatic storage credential refresh per table on the client side. Java client version >= 1.8.0 is required.
 The endpoint path is always returned when using vended credentials, but the client must configure
-REFRESH\_ENDPOINT\_ENABLED = "true" in the catalog properties for the client to use the endpoint.
+"refresh-credentials-enabled" = "true" in the catalog properties for the client to use the endpoint.
+Note, the "refresh-credentials-enabled" configuration prefix is specific to each storage provider.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,7 @@ at locations that better optimize for object storage.
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
 - Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
-automatic credential refresh per table on the client side.
+automatic storage credential refresh per table on the client side.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,9 @@ at locations that better optimize for object storage.
 
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
+- Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
+automatic credential refresh per table on the client side.
+
 ### Changes
 
 - Polaris Management API clients must be prepared to deal with new attributes in `AwsStorageConfigInfo` objects.
--- a/polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
@@ -57,6 +57,17 @@ public class PolarisResourcePaths {
         "polaris", "v1", prefix, "namespaces", RESTUtil.encodeNamespace(ns), "generic-tables");
   }
 
+  public String credentialsPath(TableIdentifier ident) {
+    return SLASH.join(
+        "v1",
+        prefix,
+        "namespaces",
+        RESTUtil.encodeNamespace(ident.namespace()),
+        "tables",
+        RESTUtil.encodeString(ident.name()),
+        "credentials");
+  }
+
   public String genericTable(TableIdentifier ident) {
     return SLASH.join(
         "polaris",
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
@@ -47,5 +47,6 @@ public interface PolarisCredentialVendor {
       PolarisEntityType entityType,
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
-      @Nonnull Set<String> allowedWriteLocations);
+      @Nonnull Set<String> allowedWriteLocations,
+      String refreshCredentialsEndpoint);
 }
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
@@ -38,6 +38,10 @@ public interface PolarisCredentialVendor {
    *     allowedWriteLocations
    * @param allowedReadLocations a set of allowed to read locations
    * @param allowedWriteLocations a set of allowed to write locations
+   * @param refreshCredentialsEndpoint an optional endpoint to use for refreshing credentials. If
+   *     supported by the storage type it will be returned to the client in the appropriate
+   *     properties. The endpoint may be relative to the base URI and the client is responsible for
+   *     handling the relative path
    * @return an enum map containing the scoped credentials
    */
   @Nonnull
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
@@ -19,6 +19,7 @@
 package org.apache.polaris.core.storage;
 
 import jakarta.annotation.Nonnull;
+import java.util.Optional;
 import java.util.Set;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.entity.PolarisEntityType;
@@ -48,5 +49,5 @@ public interface PolarisCredentialVendor {
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
       @Nonnull Set<String> allowedWriteLocations,
-      String refreshCredentialsEndpoint);
+      Optional<String> refreshCredentialsEndpoint);
 }
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
@@ -61,7 +61,8 @@ public abstract class PolarisStorageIntegration<T extends PolarisStorageConfigur
       @Nonnull RealmConfig realmConfig,
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
-      @Nonnull Set<String> allowedWriteLocations);
+      @Nonnull Set<String> allowedWriteLocations,
+      String refreshCredentialsEndpoint);
 
   /**
    * Validate access for the provided operation actions and locations.
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
@@ -56,6 +56,10 @@ public abstract class PolarisStorageIntegration<T extends PolarisStorageConfigur
    *     locations
    * @param allowedReadLocations a set of allowed to read locations
    * @param allowedWriteLocations a set of allowed to write locations
+   * @param refreshCredentialsEndpoint an optional endpoint to use for refreshing credentials. If
+   *     supported by the storage type it will be returned to the client in the appropriate
+   *     properties. The endpoint may be relative to the base URI and the client is responsible for
+   *     handling the relative path
    * @return An enum map including the scoped credentials
    */
   public abstract AccessConfig getSubscopedCreds(
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
@@ -20,6 +20,7 @@ package org.apache.polaris.core.storage;
 
 import jakarta.annotation.Nonnull;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import org.apache.polaris.core.config.RealmConfig;
 
@@ -62,7 +63,7 @@ public abstract class PolarisStorageIntegration<T extends PolarisStorageConfigur
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
       @Nonnull Set<String> allowedWriteLocations,
-      String refreshCredentialsEndpoint);
+      Optional<String> refreshCredentialsEndpoint);
 
   /**
    * Validate access for the provided operation actions and locations.
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
@@ -46,13 +46,13 @@ public enum StorageAccessProperty {
       Boolean.class,
       AwsClientProperties.REFRESH_CREDENTIALS_ENABLED,
       "whether to enable automatic refresh of credentials",
-      true,
+      false,
       false),
   AWS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       AwsClientProperties.REFRESH_CREDENTIALS_ENDPOINT,
       "the endpoint to use for refreshing credentials",
-      true,
+      false,
       false),
 
   GCS_ACCESS_TOKEN(String.class, "gcs.oauth2.token", "the gcs scoped access token"),
@@ -66,13 +66,13 @@ public enum StorageAccessProperty {
       Boolean.class,
       GCPProperties.GCS_OAUTH2_REFRESH_CREDENTIALS_ENABLED,
       "whether to enable automatic refresh of credentials",
-      true,
+      false,
       false),
   GCS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       GCPProperties.GCS_OAUTH2_REFRESH_CREDENTIALS_ENDPOINT,
       "the endpoint to use for refreshing credentials",
-      true,
+      false,
       false),
 
   // Currently not using ACCESS TOKEN as the ResolvingFileIO is using ADLSFileIO for azure case and
@@ -83,13 +83,13 @@ public enum StorageAccessProperty {
       Boolean.class,
       "adls.refresh-credentials-enabled",
       "whether to enable automatic refresh of credentials",
-      true,
+      false,
       false),
   AZURE_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       "adls.refresh-credentials-endpoint",
       "the endpoint to use for refreshing credentials",
-      true,
+      false,
       false),
   EXPIRATION_TIME(
       Long.class,
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
@@ -42,12 +42,6 @@ public enum StorageAccessProperty {
       Boolean.class, "s3.path-style-access", "whether to use S3 path style access", false),
   CLIENT_REGION(
       String.class, "client.region", "region to configure client for making requests to AWS"),
-  AWS_REFRESH_CREDENTIALS_ENABLED(
-      Boolean.class,
-      AwsClientProperties.REFRESH_CREDENTIALS_ENABLED,
-      "whether to enable automatic refresh of credentials",
-      false,
-      false),
   AWS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       AwsClientProperties.REFRESH_CREDENTIALS_ENDPOINT,
@@ -62,12 +56,6 @@ public enum StorageAccessProperty {
       "the time the gcs access token expires, in milliseconds",
       true,
       true),
-  GCS_REFRESH_CREDENTIALS_ENABLED(
-      Boolean.class,
-      GCPProperties.GCS_OAUTH2_REFRESH_CREDENTIALS_ENABLED,
-      "whether to enable automatic refresh of credentials",
-      false,
-      false),
   GCS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       GCPProperties.GCS_OAUTH2_REFRESH_CREDENTIALS_ENDPOINT,
@@ -79,12 +67,6 @@ public enum StorageAccessProperty {
   // it expects for SAS
   AZURE_ACCESS_TOKEN(String.class, "", "the azure scoped access token"),
   AZURE_SAS_TOKEN(String.class, "adls.sas-token.", "an azure shared access signature token"),
-  AZURE_REFRESH_CREDENTIALS_ENABLED(
-      Boolean.class,
-      "adls.refresh-credentials-enabled",
-      "whether to enable automatic refresh of credentials",
-      false,
-      false),
   AZURE_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       "adls.refresh-credentials-endpoint",
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
@@ -51,7 +51,7 @@ public enum StorageAccessProperty {
   AWS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       AwsClientProperties.REFRESH_CREDENTIALS_ENDPOINT,
-      "the endpoint to use for refreshing credentials",
+      "the endpoint to load vended credentials for a table from the catalog",
       false,
       false),
 
@@ -71,7 +71,7 @@ public enum StorageAccessProperty {
   GCS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       GCPProperties.GCS_OAUTH2_REFRESH_CREDENTIALS_ENDPOINT,
-      "the endpoint to use for refreshing credentials",
+      "the endpoint to load vended credentials for a table from the catalog",
       false,
       false),
 
@@ -88,7 +88,7 @@ public enum StorageAccessProperty {
   AZURE_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       "adls.refresh-credentials-endpoint",
-      "the endpoint to use for refreshing credentials",
+      "the endpoint to load vended credentials for a table from the catalog",
       false,
       false),
   EXPIRATION_TIME(
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
@@ -19,6 +19,7 @@
 package org.apache.polaris.core.storage;
 
 import org.apache.iceberg.aws.AwsClientProperties;
+import org.apache.iceberg.gcp.GCPProperties;
 
 /**
  * A subset of Iceberg catalog properties recognized by Polaris.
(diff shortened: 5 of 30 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -420,16 +423,45 @@ public Response loadTable(
...
         });
   }
 
+  private LoadTableResponse injectRefreshVendedCredentialProperties(
+      LoadTableResponse originalResponse, String credentialsEndpoint) {
+    LoadTableResponse.Builder loadResponseBuilder =
+        LoadTableResponse.builder().withTableMetadata(originalResponse.tableMetadata());
+    loadResponseBuilder.addAllConfig(originalResponse.config());
+    loadResponseBuilder.addAllCredentials(originalResponse.credentials());
+    loadResponseBuilder.addConfig(
```
singhpk234: what about azure ? do we want to do it in follow-up ? also should we only inject this when creds are for aws ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -420,16 +423,45 @@ public Response loadTable(
...
           }
 
           return tryInsertETagHeader(Response.ok(response), response, namespace, table).build();
         });
   }
 
+  private LoadTableResponse injectRefreshVendedCredentialProperties(
+      LoadTableResponse originalResponse, String credentialsEndpoint) {
+    LoadTableResponse.Builder loadResponseBuilder =
+        LoadTableResponse.builder().withTableMetadata(originalResponse.tableMetadata());
```
singhpk234: do we need to update metadata-location too ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
smaheshwar-pltr: FWIW, I think enabling the Azure refresh properties currently causes Azure credential vending to immediately fail due to https://github.com/apache/iceberg/issues/13733, so could maybe wait for that fix to release?

smaheshwar-pltr: > also should we only inject this when creds are for aws?

Could maybe lift the AWS refresh check below to `loadTable` itself, to avoid e.g. reconstructing the `LoadTableResponse` if not needed?

jasonf20: I think it's better to provide the URL even if it's not AWS that way the user can override `AwsClientProperties.REFRESH_CREDENTIALS_ENABLED` if he is working with a system that he knows supports it despite the server not being updated to reflect this yet. For example, an updated Azure iceberg SDK like mentioned above.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
jasonf20: Update it to what value?

Do you mean is it redundant with the copyConfigs below? Seems like the metadata location is stored in a dedicated field and is not part of the config.

singhpk234: nvm, it seems like the withTableMetadata sets both the tableMetada and the metadatLocation 
```
        public Builder withTableMetadata(TableMetadata tableMetadata) {
            this.metadataLocation = tableMetadata.metadataFileLocation();
            this.metadata = tableMetadata;
            return this;
        }
 ```
so it would not be required

File: runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java (hunk 1 above)
Comment (singhpk234) [reply runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java]: AWS cred refresh properties are : 

```
client.refresh-credentials-enabled
client.refresh-credentials-endpoint
```

Azure credential refresh properties are : 

```
adls.refresh-credentials-enabled
adls.refresh-credentials-endpoint
```

I am not sure how can one re-use aws cred endpoint prop to enable azure ? nevertheless why would server send back azure properties for an aws storage ?

Comment (jasonf20) [reply runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java]: Oh, good to know. I assumed it's the same properties. I think it's not a good choice to use different properties, even if the client level has different settings. The client/server protocol should be standard. However, it seems like that boat has sailed. 

I would prefer having the endpoint URL always served even if it's not used, that way power users have the info available to them.

Despite that I've added a commit that only performs the inject code if the type is AWS. Let me know if that's better.

File: runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java [hunk 3]
```diff
@@ -430,16 +433,48 @@ public Response loadTable(
...
+  private LoadTableResponse injectRefreshVendedCredentialProperties(
+      LoadTableResponse originalResponse, String credentialsEndpoint) {
+    // Only enable credential refresh for currently supported credential types
+    if (originalResponse.credentials().stream()
+        .anyMatch(
+            credential ->
+                credential
+                    .prefix()
+                    .toLowerCase()
+                    .startsWith(StorageConfigInfo.StorageTypeEnum.S3.name().toLowerCase()))) {
```
singhpk234: i would also recommend adding this for azure, even if the client has bug, we would be ready from the server side.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
singhpk234: That i agree, unfortunately I not part of the discussions where these decisions were being made.
My recommendation is to also add similar handling for Azure, i understand the client has a bug with Azure, but this should not stop us at the server end to be ready with the changes.

dimas-b:
Please address CI failures :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
jasonf20: Alright, but since the properties are different for each provider, doing this in a clean way requires delegating this to the storage integration, so the code change scope becomes much larger. Please see the latest commit which does this.

polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
```diff
@@ -57,6 +57,17 @@ public String genericTables(Namespace ns) {
...
 
+  public String credentialsPath(TableIdentifier ident) {
+    return SLASH.join(
+        "v1",
+        prefix,
+        "namespaces",
+        RESTUtil.encodeNamespace(ident.namespace()),
+        "tables",
+        RESTUtil.encodeString(ident.name()),
+        "credentials");
```
dimas-b, line 68: Just to confirm: relative URIs are correct in this case, right?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -430,9 +438,14 @@ public Response loadTable(
                     .loadTableIfStale(tableIdentifier, ifNoneMatch, snapshots)
                     .orElseThrow(() -> new WebApplicationException(Response.Status.NOT_MODIFIED));
           } else {
+            String refreshCredentialsEndpoint =
+                (delegationModes.contains(VENDED_CREDENTIALS))
+                    ? new PolarisResourcePaths(prefix).credentialsPath(tableIdentifier)
```
dimas-b: Might be worth extracting into a shared method (cf. line 366).

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -130,7 +131,8 @@ public AccessConfig getOrGenerateSubScopeCreds(
                   polarisEntity.getType(),
                   k.allowedListAction(),
                   k.allowedReadLocations(),
-                  k.allowedWriteLocations());
+                  k.allowedWriteLocations(),
+                  refreshCredentialsEndpoint);
```
dimas-b: `refreshCredentialsEndpoint` should be part of the cache key because `ScopedCredentialsResult` is a function also of `refreshCredentialsEndpoint`.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -75,7 +75,8 @@ public AccessConfig getSubscopedCreds(
       @Nonnull RealmConfig realmConfig,
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
-      @Nonnull Set<String> allowedWriteLocations) {
+      @Nonnull Set<String> allowedWriteLocations,
+      String refreshCredentialsEndpoint) {
```
dimas-b: Why is `refreshCredentialsEndpoint` used for AWS and Azure but not for GCP?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -39,6 +41,18 @@ public enum StorageAccessProperty {
       Boolean.class, "s3.path-style-access", "whether to use S3 path style access", false),
   CLIENT_REGION(
       String.class, "client.region", "region to configure client for making requests to AWS"),
+  AWS_REFRESH_CREDENTIALS_ENABLED(
+      Boolean.class,
+      AwsClientProperties.REFRESH_CREDENTIALS_ENABLED,
+      "whether to enable automatic refresh of credentials",
+      true,
```
dimas-b: I do not think this property is a credential property (logically)

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -39,6 +41,18 @@ public enum StorageAccessProperty {
...
+      false),
+  AWS_REFRESH_CREDENTIALS_ENDPOINT(
+      String.class,
+      AwsClientProperties.REFRESH_CREDENTIALS_ENDPOINT,
+      "the endpoint to use for refreshing credentials",
+      true,
```
dimas-b: I do not think this property is a credential property (logically)

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -52,6 +66,18 @@ public enum StorageAccessProperty {
   // it expects for SAS
   AZURE_ACCESS_TOKEN(String.class, "", "the azure scoped access token"),
   AZURE_SAS_TOKEN(String.class, "adls.sas-token.", "an azure shared access signature token"),
+  AZURE_REFRESH_CREDENTIALS_ENABLED(
+      Boolean.class,
+      "adls.refresh-credentials-enabled",
+      "whether to enable automatic refresh of credentials",
+      true,
```
dimas-b: I do not think this property is a credential property (logically)

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -52,6 +66,18 @@ public enum StorageAccessProperty {
...
+      false),
+  AZURE_REFRESH_CREDENTIALS_ENDPOINT(
+      String.class,
+      "adls.refresh-credentials-endpoint",
+      "the endpoint to use for refreshing credentials",
+      true,
```
dimas-b: I do not think this property is a credential property (logically)

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
```diff
@@ -61,7 +61,8 @@ public abstract AccessConfig getSubscopedCreds(
       @Nonnull RealmConfig realmConfig,
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
-      @Nonnull Set<String> allowedWriteLocations);
+      @Nonnull Set<String> allowedWriteLocations,
+      String refreshCredentialsEndpoint);
```
dimas-b: Maybe `Optional<String>`?

polaris-core/src/test/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegrationTest.java
```diff
@@ -32,20 +32,20 @@ public class AzureCredentialsStorageIntegrationTest {
   public void testAzureCredentialFormatting() {
     Instant expiresAt = Instant.ofEpochMilli(Long.MAX_VALUE);
 
-    AccessConfig noSuffixResult = toAccessConfig("sasToken", "some_account", expiresAt);
+    AccessConfig noSuffixResult = toAccessConfig("sasToken", "some_account", expiresAt, null);
```
dimas-b: Could you add tests for non-null refresh endpoint values? (sorry, if I missed them).

dimas-b (commented):
LGTM overall :+1: Thanks for working on this @jasonf20 !

Do we have existing integration tests for the credential refresh endpoint? If not, could you add a test in `RestCatalogMinIOSpecialIT` (probably). I suppose we could using the test REST Client to validate refresh credential responses. I know this is specific to the S3 storage integration, but it will hit the common call path and we can validate Azure `AccessConfig` in unit tests. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/rest/PolarisResourcePaths.java
jasonf20: Yes, from what I understand the recent version of the client automatically appends the base URI if a relative path is returned from the server.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
jasonf20: Since it's calculated statically I think it won't make a difference but I can add it. Perhaps if it's called with vended-credentials enabled and then again disabled it has some effect. Will add to the key.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
jasonf20: If it's supported in GCP I can use it, but previous comments mentioned it's only supported for AWS and Azure right now.

polaris-core/src/test/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegrationTest.java
jasonf20: I changed one of the tests to use a non null value already, I'll see if there are more places worth doing it.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
jasonf20: From my understanding of the client code:
1. The `AwsClientFactory` is initialized ahead of time with generic properties retrieved from the "configuration" endpoint. This doesn't include specific endpoints for credential refresh per table and/or access keys. 
2. Later on when loading the fileIO per table, the base properties + additional table properties provided from the `List<Credential>` are [added here](https://github.com/apache/iceberg/blob/2012f661a19e613abec2ef0024d7b340c9620f14/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L450).

For the property to be available in `Credential.config()` down the line this needs to be set to true.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
jasonf20: done

jasonf20:
@dimas-b I've added the integration test and unit test for azure as you suggested and made other fixes based on your comments.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
dimas-b: Interesting... but ok, let's implement what we can :)

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
dimas-b: Thanks for the code pointer! In this case, it does make sense to mark these properties as "credentials", but please add a code comment for future readers to be aware of what Iceberg java clients expect.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
```diff
@@ -47,5 +48,6 @@ ScopedCredentialsResult getSubscopedCredsForEntity(
       PolarisEntityType entityType,
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
-      @Nonnull Set<String> allowedWriteLocations);
+      @Nonnull Set<String> allowedWriteLocations,
+      Optional<String> refreshCredentialsEndpoint);
```
dimas-b, line 56: Please add javadoc for this property and mention that it may be a relative URI.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
```diff
@@ -61,7 +62,8 @@ public abstract AccessConfig getSubscopedCreds(
       @Nonnull RealmConfig realmConfig,
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
-      @Nonnull Set<String> allowedWriteLocations);
+      @Nonnull Set<String> allowedWriteLocations,
+      Optional<String> refreshCredentialsEndpoint);
```
dimas-b, line 70: Same here: please add javadoc

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: Thx for adding it to the cache key! Would you mind accessing it here via `k.` for clarity?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -430,16 +438,28 @@ public Response loadTable(
...
 
           return tryInsertETagHeader(Response.ok(response), response, namespace, table).build();
         });
   }
 
+  private static Optional<String> getRefreshCredentialsEndpoint(
+      EnumSet<AccessDelegationMode> delegationModes,
+      String prefix,
+      TableIdentifier tableIdentifier) {
+    return delegationModes.contains(AccessDelegationMode.VENDED_CREDENTIALS)
```
dimas-b, line 458: This logic is ok from my POV, but I wonder if other community members prefer a feature flag to switch the new behaviour on/off.

Please send an email about this to the `dev` ML.

dimas-b (commented):
This PR LGTM in its current state :+1: ... with some minor comments :slightly_smiling_face: 

However, I believe it is worth sending a `dev` email about the new feature before merging (as commented below).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
dimas-b: Actually, I'll send the `dev` email myself :)

dimas-b: https://lists.apache.org/thread/dvrx5lg36ovo7387pt6dm06wgmswstvc

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -39,6 +41,18 @@ public enum StorageAccessProperty {
...
+  AWS_REFRESH_CREDENTIALS_ENABLED(
+      Boolean.class,
+      AwsClientProperties.REFRESH_CREDENTIALS_ENABLED,
+      "whether to enable automatic refresh of credentials",
+      true,
+      false),
+  AWS_REFRESH_CREDENTIALS_ENDPOINT(
+      String.class,
+      AwsClientProperties.REFRESH_CREDENTIALS_ENDPOINT,
+      "the endpoint to use for refreshing credentials",
```
singhpk234: ideally we are not refreshing the credentials, we are getting brand new one how about we say something like 
https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L119

singhpk234 (commented):
Mostly LGTM as well 👍 ! we are very close :) !

dimas-b:
@jasonf20 : Could you add an entry about this change to `CHANGELOG.md`, section `New Features`?

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
dimas-b: It looks like GCP credential refresh _is_ supported: https://github.com/lakekeeper/lakekeeper/pull/1227/files#diff-8d3c6564c330f473c1e4556b7dbbf79dddae3a49a9d9c55352c7a614fc8b3fb5R88 WDYT?

dimas-b: Cf. https://github.com/apache/iceberg/blob/b82dac4858fb2d15929a797660252c50f8e00ed8/gcp/src/main/java/org/apache/iceberg/gcp/GCPProperties.java#L50

dimas-b: Iceberg support for GCP credential refresh appears to be available since 1.7.0, according to https://github.com/apache/iceberg/pull/11282

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
jasonf20: I had another look at this. I think I was wrong since it also passes in `tableConf` which if set initializes a new fileIO with those properties, so this can be false. I'll update those.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
jasonf20: done

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
```diff
@@ -37,6 +38,10 @@ public interface PolarisCredentialVendor {
    *     allowedWriteLocations
    * @param allowedReadLocations a set of allowed to read locations
    * @param allowedWriteLocations a set of allowed to write locations
+   * @param refreshCredentialsEndpoint an optional endpoint to use for refreshing credentials. If
+   *     supported by the storage type it will be returned to the client in the appropriate
+   *     properties. The endpoint may be relative to the base URI and the client is responsible for
```
singhpk234, line 43: can we return anything except relative path ? my understanding is base uri is always catalog uri https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/s3/VendedCredentialsProvider.java#L92  is this statement in intentionally open ended since spec is not explicit ?

dimas-b (dismissed):
LGTM :+1: Thanks again for you contribution, @jasonf20 !

Since this PR was already mentioned on the [dev ML](https://lists.apache.org/thread/dvrx5lg36ovo7387pt6dm06wgmswstvc) on Aug 21 and no concerned were raised, I think it is good to merge. I'll wait until Aug 26 before merging, in case any last minute comments come up.

CHANGELOG.md
```diff
@@ -81,6 +81,9 @@ at locations that better optimize for object storage.
 
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
+- Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
+automatic credential refresh per table on the client side.
```
singhpk234: nit: I understand the intent but i think imho its better to be explicit here, as there.
```suggestion
automatic storage credential refresh per table on the client side.
```

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
dimas-b: thx !

singhpk234 (approved):
LGTM overall (have some small cosmetic suggestion for documentation), Thank you @jasonf20 ! much needed feature for long running jobs !

will wait for @dimas-b review and approval  before merge

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
dimas-b: Thanks for your diligence! :slightly_smiling_face:

singhpk234:
Looks like we already have @dimas-b ;) review, we were parallely reviewing, we good to go from my end !

CHANGELOG.md
dimas-b: The suggestion LGTM :+1:

dimas-b:
Oops, CI failure :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
jasonf20: At one point the client required full paths, now it works with partial paths. I'm not sure if the spec identifies one as the correct method over the other so I left it like this. If one is definitely the "correct" way I can update the comment.

dimas-b: @jasonf20 : do you know the Iceberg version / PR where relative paths started working?

jasonf20: It was linked in the previous PR. https://github.com/apache/polaris/pull/1164#discussion_r2007984981

Seems like it's since 1.8.0

jasonf20:
> Oops, CI failure 😅

fixed

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
singhpk234: Oh no this made a lot of things super tricky, i checked iceberg python for refresh keyword but i didn't any prs for that so i am assuming that it doesn't support, taking only the iceberg java sdk reference 
if we always send relative url it only works for >= 1.8
fails for < 1.8 
I know there is a header that the java sdk sends which has SDK version `X-Client-Version` header we may want to salvage this now on when to return the absolute vs relative, https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java#L77

seems like the only broken release would be 1.7 

Its a bit surprising to me, but i think for now we can park this by just documenting this gotcha, WDYT @dimas-b ?

dimas-b: Getting full URI is a bit trickier - as discussed in #1164.

From my POV, let's merge this "as is" and open a GH issue that it does not work properly for Iceberg < 1.8. When someone has time, let's improve.

Getting this feature enabled for Iceberg clients >= 1.8 is valuable, I think.

dimas-b: Since old client version behave differently, I think we ought to have a feature flag now... just in case someone runs into client incompatibility issues and wants to disable credential refresh properties completely (i.e. revert to old Polaris behaviour). WDYT?

singhpk234: That would be ideal IMHO (specially now), but i think the jobs without us sending the credentials endpoint would still have failed because my understanding is we would have not made the /credentials call unless the creds in loadTable expires https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/s3/VendedCredentialsProvider.java#L63, its just we might see an HTTP error rather than the creds expired exception.
But agree, we might want to have a feature flag to fallback to the old behaviour if we unexpectedly fail.

jasonf20: I don't think a feature flag is needed. 

1. The client is already broken, if they reach the need to refresh credentials it already fails today. 
2. A feature flag will not solve the problem. Enabling it will fix the issue for some and create a different issue for others, so you'll never be able to enable it anymore than you can without the flag. 
3. Clients can simply upgrade to 1.8.0 to get a working version. Their current version doesn't work anyway so they can upgrade at their own time when this becomes important to them. 
4. If you would like we can remove the `refresh_credentials_enabled` property and let the client set that one in his catalog settings. That way he can control if this gets used or not. This would be better than a flag and is controlled client side. not server side.

jasonf20: I've added a commit that does 4.

dimas-b (commented):
@jasonf20 : I'm a bit lost now... Why do we need to remove the "enabled" flag in https://github.com/apache/polaris/pull/2341/commits/2a123d1df204c748622ff4b85f3a8906daaa8a9c?

In my earlier comments I mean that we need to add a new flag to `FeatureConfiguration` and use it to control whether to add the credential refresh endpoint and the related "enabled" flag to `AccessConfig`... WDYT?

dimas-b (dismissed):
After re-reading @jasonf20 comment, I think removing of the "enabled" properties from `AccessConfig` is a sound approach.

CHANGELOG.md
```diff
@@ -81,6 +81,9 @@ at locations that better optimize for object storage.
 
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
+- Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
+automatic storage credential refresh per table on the client side.
```
singhpk234: Lets call out the client version limitation here and the suggestion to the caller to set the client side override to enable this.

singhpk234 (dismissed):

CHANGELOG.md
dimas-b: good points!

dimas-b: since the release is imminent, if the round trip cycle is slow, I suppose we can still merge this PR and later adjust the changelog between @singhpk234 and myself :) Let's give it a few more hours.

jasonf20: I updated the changelog. Seems a little verbose now for a changelog but It's probably fine. 

p.s. I believe you should be able to push changes to my branch yourself since I have the "Allow edits and access to secrets by maintainers" checkbox enabled. But I never tried using that myself.

CHANGELOG.md
```diff
@@ -82,7 +82,9 @@ at locations that better optimize for object storage.
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
 - Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
-automatic storage credential refresh per table on the client side.
+automatic storage credential refresh per table on the client side. Java client version >= 1.8.0 is required.
+The endpoint path is always returned when using vended credentials, but the client must configure 
+REFRESH\_ENDPOINT\_ENABLED = "true" in the catalog properties for the client to use the endpoint.
```
singhpk234: the properties are unfortunately different for the all the cloud providers, not sure if we wanna call this out, just saying enabled never the less `refresh-credentials-enabled` should be endpoint 

how about we say that clients must enable refresh-credentials flag for their respective cloud provider ?

jasonf20: Yea that's less verbose. I changed it to that.

dimas-b: lgtm :+1:

jasonf20:
Thanks for your reviews @dimas-b @singhpk234 !

CHANGELOG.md
```diff
@@ -81,6 +81,11 @@ at locations that better optimize for object storage.
 
 - Introduced bootstrap command options to specify custom schema files for database initialization.
 
+- Added refresh credentials endpoint configuration to LoadTableResponse for AWS, Azure, and GCP. Enabling
+automatic storage credential refresh per table on the client side. Java client version >= 1.8.0 is required.
+The endpoint path is always returned when using vended credentials, but clients must enable the
```
flyrain, line 86: We will need to document this. Unfortunately, the best place to doc this is a page for storage setting and credential vending, which doesn't exist yet. Here is the issue filed long time ago, https://github.com/apache/polaris/issues/1325.  Not a blocker for this PR though.

PR #2343, opened 2025-08-13
Add integration tests with Keycloak

adutra:


runtime/test-common/build.gradle.kts
```diff
@@ -32,10 +32,21 @@ configurations.all {
...
   implementation(platform(libs.testcontainers.bom))
   implementation("org.testcontainers:testcontainers")
   implementation("org.testcontainers:postgresql")
+
+  implementation(libs.testcontainers.keycloak) {
+    exclude(group = "org.keycloak", module = "keycloak-admin-client")
+  }
+  // Keycloak Admin Client brings RESTEasy Classic, which conflicts with Quarkus RESTEasy Reactive;
+  // it must not be present during Quarkus augmentation otherwise Quarkus tests won't start.
+  compileOnly(libs.keycloak.admin.client)
```
adutra, line 49: I spent 5 hours today trying to find a better solution for this, but in the end, having both RESTEasy Classic and Reactive on the classpath gets blocked here:

https://github.com/quarkusio/quarkus/blob/072896e711640752ffc3176677591c6d263360e8/extensions/resteasy-reactive/rest-common/deployment/src/main/java/io/quarkus/resteasy/reactive/common/deployment/ResteasyReactiveCommonProcessor.java#L109-L110

Maybe a better solution would be to isolate the Keycloak container in a separate module, and shade RESTEasy Classic.

dimas-b: Having a server runner that launches pre-built jars would help here, I believe... There was a discussion or PR about that somewhere, right?

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogKeycloakFileIT.java
```diff
@@ -0,0 +1,81 @@
...
+  KeycloakAccess keycloak;
+
+  @Override
+  protected ClientPrincipal createTestPrincipal(
+      PolarisClient client, String principalName, String principalRole) {
+    ClientPrincipal principal = super.createTestPrincipal(client, principalName, principalRole);
+    keycloak.createRole(principalRole);
+    keycloak.createUser(
+        principalName,
+        // Use the same password as the client secret
```
dimas-b: Why? As for me, it detracts from the test's validity because now it is not obvious whether authentication happens locally (client secret) or via Keycloak :thinking:

dimas-b (commented):
Nice improvement :+1: ... just one minor comment :)

runtime/test-common/build.gradle.kts
snazy: I recall that RESTEasy classic vs reactive issue from Nessie, where it's resolved [this way](https://github.com/projectnessie/nessie/blob/b7bd05b1a1c7a519fd5e3b121ce787d4e0cb812b/servers/quarkus-server/build.gradle.kts#L148-L151). But the effect's the same: exclude RESTEasy Classic.

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogKeycloakFileIT.java
adutra: I'm not sure I get your point. This password is never used outside of this class (see method below).

adutra: Anyways, I replaced this with a fixed password for all users. Let me know if that's better.

runtime/test-common/build.gradle.kts
adutra: I suggest to leave it this way for now and I'll investigate if shading is an option.

runtime/test-common/src/main/java/org/apache/polaris/test/commons/keycloak/KeycloakContainer.java
```diff
@@ -0,0 +1,181 @@
...
+  @Override
+  public void start() {
+    super.start();
+    keycloakAdminClient = getKeycloakAdminClient();
+    URI rootUrl = URI.create(getAuthServerUrl());
+    issuerUrl = rootUrl.resolve(CONTEXT_PATH);
+    tokenEndpoint = issuerUrl.resolve("protocol/openid-connect/token");
+    createRole("ALL");
+    createUser("root");
+    assignRoleToUser("ALL", "root");
```
adutra, line 69: FYI I initially forgot to assign `PRINCIPAL_ROLE:ALL` to `root` and yet, the tests passed: this is because `ActiveRolesProvider` automatically infers "all granted roles" when no role is found in the token. This is another reason why I am advocating for merging `ActiveRolesProvider` into `Authenticator`.

runtime/test-common/build.gradle.kts
snazy: +1, shading feels too much.

snazy (approved):
LGTM 

Letting @dimas-b take another look though.

runtime/test-common/build.gradle.kts
dimas-b: shading is too much for this case :) I'd go with a custom runner... not ok "as is" for now

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogKeycloakFileIT.java
dimas-b: I guessed the password was not used outside, but that's also a reason for not using the same password, right? ;) Thx for changing!

dimas-b: My point was that if the password is the same (with a specific comment) it creates an expectation that it might be reused somewhere... which is what I wanted to avoid :sweat_smile:

dimas-b (approved):

PR #2344, opened 2025-08-13
Integration tests for Catalog Federation

poojanilangekar:
Adds a Junit5 integration test for catalog federation.

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
@@ -114,10 +114,6 @@ public class CatalogFederationIntegrationTest {
   @Test
   void testCatalogFederation() {
     newUserCredentials = managementApi.createPrincipal(PRINCIPAL_NAME);
-    assertThat(newUserCredentials).isNotNull();
-    assertThat(newUserCredentials.getCredentials()).isNotNull();
-    assertThat(newUserCredentials.getCredentials().getClientId()).isNotEmpty();
-    assertThat(newUserCredentials.getCredentials().getClientSecret()).isNotEmpty();
 
     FileStorageConfigInfo storageConfig =
         FileStorageConfigInfo.builder()
@@ -135,9 +131,6 @@ public class CatalogFederationIntegrationTest {
             .setStorageConfigInfo(storageConfig)
             .build();
     managementApi.createCatalog(localCatalog);
-    assertThat(managementApi.getCatalog(LOCAL_CATALOG_NAME)).isNotNull();
-    assertThat(managementApi.getCatalog(LOCAL_CATALOG_NAME).getType())
-        .isEqualTo(Catalog.TypeEnum.INTERNAL);
 
     CatalogGrant catalogGrant =
         CatalogGrant.builder()
@@ -145,8 +138,6 @@ public class CatalogFederationIntegrationTest {
             .setPrivilege(CatalogPrivilege.TABLE_WRITE_DATA)
             .build();
     managementApi.addGrant(LOCAL_CATALOG_NAME, CATALOG_ROLE_NAME, catalogGrant);
-    assertThat(managementApi.listGrants(LOCAL_CATALOG_NAME, CATALOG_ROLE_NAME).getGrants())
-        .contains(catalogGrant);
     managementApi.assignPrincipalRole(PRINCIPAL_NAME, PRINCIPAL_ROLE_NAME);
     CatalogRole localCatalogAdminRole =
         managementApi.getCatalogRole(LOCAL_CATALOG_NAME, CATALOG_ROLE_NAME);
@@ -177,13 +168,8 @@ public class CatalogFederationIntegrationTest {
             .setStorageConfigInfo(storageConfig)
             .build();
     managementApi.createCatalog(externalCatalog);
-    assertThat(managementApi.getCatalog(EXTERNAL_CATALOG_NAME)).isNotNull();
-    assertThat(managementApi.getCatalog(EXTERNAL_CATALOG_NAME).getType())
-        .isEqualTo(Catalog.TypeEnum.EXTERNAL);
 
     managementApi.addGrant(EXTERNAL_CATALOG_NAME, CATALOG_ROLE_NAME, catalogGrant);
-    assertThat(managementApi.listGrants(EXTERNAL_CATALOG_NAME, CATALOG_ROLE_NAME).getGrants())
-        .contains(catalogGrant);
     CatalogRole externalCatalogAdminRole =
         managementApi.getCatalogRole(EXTERNAL_CATALOG_NAME, CATALOG_ROLE_NAME);
     managementApi.grantCatalogRoleToPrincipalRole(
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
@@ -1,0 +1,253 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.it.test;
+
+import static org.apache.polaris.service.it.env.PolarisClient.polarisClient;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.net.URI;
+import java.util.List;
+import org.apache.polaris.core.admin.model.AuthenticationParameters;
+import org.apache.polaris.core.admin.model.Catalog;
+import org.apache.polaris.core.admin.model.CatalogGrant;
+import org.apache.polaris.core.admin.model.CatalogPrivilege;
+import org.apache.polaris.core.admin.model.CatalogProperties;
+import org.apache.polaris.core.admin.model.CatalogRole;
+import org.apache.polaris.core.admin.model.ConnectionConfigInfo;
+import org.apache.polaris.core.admin.model.ExternalCatalog;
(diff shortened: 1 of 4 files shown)


poojanilangekar:
CC: @eric-maynard @dimas-b

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
```diff
@@ -0,0 +1,253 @@
...
+    managementApi.deletePrincipal(PRINCIPAL_NAME);
+  }
+
+  @Test
+  void testCatalogFederation() {
+    newUserCredentials = managementApi.createPrincipal(PRINCIPAL_NAME);
+    assertThat(newUserCredentials).isNotNull();
+    assertThat(newUserCredentials.getCredentials()).isNotNull();
+    assertThat(newUserCredentials.getCredentials().getClientId()).isNotEmpty();
+    assertThat(newUserCredentials.getCredentials().getClientSecret()).isNotEmpty();
```
dimas-b: nit: why bother asserting this?

runtime/service/src/intTest/java/org/apache/polaris/service/it/CatalogFederationIT.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.quarkus.test.junit.QuarkusIntegrationTest;
+import io.quarkus.test.junit.QuarkusTestProfile;
+import io.quarkus.test.junit.TestProfile;
+import java.util.Map;
+import org.apache.polaris.service.it.test.CatalogFederationIntegrationTest;
+
+@TestProfile(CatalogFederationIT.CatalogFederationProfile.class)
+@QuarkusIntegrationTest
+public class CatalogFederationIT extends CatalogFederationIntegrationTest {
```
dimas-b: This test probably belongs in the `spark-tests` module, same as `SparkIT`... WDYT?

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
```diff
@@ -0,0 +1,253 @@
...
+        .isEqualTo(Catalog.TypeEnum.INTERNAL);
+
+    CatalogGrant catalogGrant =
+        CatalogGrant.builder()
+            .setType(CatalogGrant.TypeEnum.CATALOG)
+            .setPrivilege(CatalogPrivilege.TABLE_WRITE_DATA)
+            .build();
+    managementApi.addGrant(LOCAL_CATALOG_NAME, CATALOG_ROLE_NAME, catalogGrant);
+    assertThat(managementApi.listGrants(LOCAL_CATALOG_NAME, CATALOG_ROLE_NAME).getGrants())
+        .contains(catalogGrant);
```
dimas-b: nit: this assert validates grant handling, not catalog federation... It's ok to have it, but it overloads the test's meaning.

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
```diff
@@ -0,0 +1,253 @@
...
+        PolarisCatalog.builder()
+            .setType(Catalog.TypeEnum.INTERNAL)
+            .setName(LOCAL_CATALOG_NAME)
+            .setProperties(catalogProperties)
+            .setStorageConfigInfo(storageConfig)
+            .build();
+    managementApi.createCatalog(localCatalog);
+    assertThat(managementApi.getCatalog(LOCAL_CATALOG_NAME)).isNotNull();
+    assertThat(managementApi.getCatalog(LOCAL_CATALOG_NAME).getType())
+        .isEqualTo(Catalog.TypeEnum.INTERNAL);
```
dimas-b: nit: these look like redundant asserts to be... The test will fail anyway if the catalog is not created. If we want to enhance test coverage of the management API (as is validate that types are not lost on API round-trips) we should probably do that in dedicated (and simpler) tests.

dimas-b (commented):
Thanks, @poojanilangekar ! The PR LGTM overall, but I have a concern with the IT test location (as commented).

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
poojanilangekar: Done.

runtime/service/src/intTest/java/org/apache/polaris/service/it/CatalogFederationIT.java
poojanilangekar: Done.

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
poojanilangekar: Done.

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
poojanilangekar: Done.

runtime/spark-tests/src/intTest/java/org/apache/polaris/service/spark/it/CatalogFederationIT.java
```diff
@@ -0,0 +1,39 @@
...
+@TestProfile(CatalogFederationIT.CatalogFederationProfile.class)
+@QuarkusIntegrationTest
+public class CatalogFederationIT extends CatalogFederationIntegrationTest {
+
+  public static class CatalogFederationProfile implements QuarkusTestProfile {
+    @Override
+    public Map<String, String> getConfigOverrides() {
+      return Map.of(
+          "polaris.features.\"ENABLE_CATALOG_FEDERATION\"", "true",
+          "polaris.features.\"ALLOW_OVERLAPPING_CATALOG_URLS\"", "true");
```
adutra, line 37: Would it be possible to use 2 different base locations + 2 different storage configs for the two catalogs, and thus avoid setting this?

poojanilangekar: We need to use the same base location because we are using a single Polaris instance. So unless we have two different Polaris instances (which we can't do in the current integration test setup), we need this configuration.

runtime/spark-tests/src/intTest/java/org/apache/polaris/service/spark/it/CatalogFederationIT.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.polaris.service.it.test.CatalogFederationIntegrationTest;
+
+@TestProfile(CatalogFederationIT.CatalogFederationProfile.class)
+@QuarkusIntegrationTest
+public class CatalogFederationIT extends CatalogFederationIntegrationTest {
+
+  public static class CatalogFederationProfile implements QuarkusTestProfile {
+    @Override
+    public Map<String, String> getConfigOverrides() {
+      return Map.of(
```
eric-maynard, line 34: It might be good to explicitly set `SUPPORTED_CATALOG_CONNECTION_TYPES`

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
```diff
@@ -0,0 +1,239 @@
...
+            .setProperties(catalogProperties)
+            .setStorageConfigInfo(storageConfig)
+            .build();
+    managementApi.createCatalog(externalCatalog);
+
+    managementApi.addGrant(EXTERNAL_CATALOG_NAME, CATALOG_ROLE_NAME, catalogGrant);
+    CatalogRole externalCatalogAdminRole =
+        managementApi.getCatalogRole(EXTERNAL_CATALOG_NAME, CATALOG_ROLE_NAME);
+    managementApi.grantCatalogRoleToPrincipalRole(
+        PRINCIPAL_ROLE_NAME, EXTERNAL_CATALOG_NAME, externalCatalogAdminRole);
```
eric-maynard, line 176: We only have one test right now so it's fine, but if we add more this should get refactored out

eric-maynard (approved):

runtime/spark-tests/src/intTest/java/org/apache/polaris/service/spark/it/CatalogFederationIT.java
poojanilangekar: Done.

runtime/spark-tests/src/intTest/java/org/apache/polaris/service/spark/it/CatalogFederationIT.java
dimas-b: @poojanilangekar : Do you mean that federated catalogs check for location overlaps with internal catalogs?

poojanilangekar: Yes, that is correct. 

Unless the `ALLOW_OVERLAPPING_CATALOG_URLS` flag is set, Polaris always checks if the base location of a CreateCatalog (irrespective of whether its INTERNAL vs EXTERNAL) request overlaps with an existing catalog. Therefore we need this flag to be set, at least for the JUnit tests.

eric-maynard: (This has come up before, but I think the fix is that federating EXTERNAL catalogs shouldn't need or support a location)

poojanilangekar: Yes, this is correct. In case of "real" federation, none of the local catalogs will have an overlapping base location.

dimas-b: As @eric-maynard noted, it does look like federated catalogs should _not_ have a location defined in Polaris. Is it possible to avoid declaring locations in federated catalogs now?

poojanilangekar: Not at the moment. Making this change would require changing the API spec and making subsequent changes in the CreateCatalog validations.

dimas-b: Would you be able to open an issue for sorting this out later?

poojanilangekar: Done: #2356

dimas-b: thx :+1:

dimas-b: This PR is good to merge from my POV :+1:

PR #2348, opened 2025-08-14
Replace CallContext with RealmConfig in enforceFeatureEnabledOrThrow

XN137:
this works towards breaking down `CallContext` usage and contents


snazy (approved):

flyrain (approved):
+1 Thanks for the PR.

PR #2349, opened 2025-08-14
Fix Keycloak getting-started example

adutra:
The `polaris-setup` container was erroneously including a non-existent scope when fetching a token from Keycloak.

### DIFF
--- a/getting-started/keycloak/docker-compose.yml
+++ b/getting-started/keycloak/docker-compose.yml
@@ -67,13 +67,16 @@ services:
       - CLIENT_SECRET=s3cr3t
     volumes:
       - ../assets/polaris/:/polaris
-    entrypoint: |
-      /bin/sh -c "apk add --no-cache jq && \
-        chmod +x /polaris/create-catalog.sh && \
-        token=$$(curl http://keycloak:8080/realms/iceberg/protocol/openid-connect/token --user client1:s3cr3t -d 'grant_type=client_credentials' -d 'scope=catalog' | jq -r .access_token) && \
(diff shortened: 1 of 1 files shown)


adutra:
@singhpk234 would you mind having a look again? I don't know how this bug slipped through the cracks 😄

getting-started/keycloak/docker-compose.yml
```diff
@@ -67,13 +67,16 @@ services:
...
-        /polaris/create-catalog.sh realm-internal && \
-        /polaris/create-catalog.sh realm-external $$token && \
-        /polaris/create-catalog.sh realm-mixed $$token"
+    entrypoint: "/bin/sh"
+    command:
+      - "-c"
+      - >-
+        apk add --no-cache jq && 
+        chmod +x /polaris/create-catalog.sh && 
+        token=$$(curl http://keycloak:8080/realms/iceberg/protocol/openid-connect/token --user client1:s3cr3t -d 'grant_type=client_credentials' | jq -r .access_token) &&
```
adutra, line 76: The only real difference is the removal of `-d 'scope=catalog'` – the rest is formatting.

adutra:
> Not sure how this got missed, i did get a 200 from keycloak, and do remeber we discussed removing the scopes.
> for my understanding how does the error shows up ?

The create-catalog.sh script fails to create the catalog.

What I think happened is that at some point I removed the `catalog` scope from Keycloak config (since it was unused), but I already had a Keycloak container created, so when I started the container again, it started with the old config, and it worked.

PR #1466, opened 2025-04-25
Implement federation to HadoopCatalog

eric-maynard:
Polaris recently added support for catalog federation in #1305, which allows a Polaris instance to federate requests against a given catalog to another Iceberg REST catalog This is useful for deployments with multiple Iceberg REST catalogs, but we need to introduce federation to other catalogs in order to support a more diverge range of deployments.

This implements federation to a `Hadoop` catalog via the new `HADOOP` type of `ConnectionConfigInfo`.

PR #1305, opened 2025-04-04
Initial MVP implementation of Catalog Federation to remote Iceberg REST Catalogs

dennishuo:
Initial MVP for https://github.com/apache/polaris/issues/540

This initial implementation of Catalog Federation is described in this document: https://docs.google.com/document/d/1Q6eEytxb0btpOPcL8RtkULskOlYUCo_3FLvFRnHkzBY/edit?tab=t.0#bookmark=id.o00h4wkebh07 - Section "Securables Federation with Catalog Facade only"

One key building block is managing the user-specified credentials to use for connecting to remote catalogs, explained in https://docs.google.com/document/d/1JPNx5vL4vM8DqwRwnBIPiQxwN4MXOdGx4Ki0j7vgwSM/edit?tab=t.0

Key elements implemented here:

- Creation of ExternalCatalogs with ConnectionConfigInfo defined to point at remote Iceberg REST Catalogs
- UserSecretsManager interface and UserSecretReference to define a safe control flow where secrets specified or referenced in API requests are canonicalized into non-secret "reference" objects to be persisted with relevant entities; provide a default UnsafeInMemorySecretsManager which exercises the relevant control flows
- Instantiation of outbound Catalog clients in IcebergCatalogHandler for federated catalogs and re-routing all read and write requests to the remote catalog instead of using a Polaris-local IcebergCatalog
- Updated Resolver logic to accommodate "partial resolution" in a way that already supports correct Catalog-level RBAC now, and will support finer-grained RBAC in the future once we add JIT-creation of entities (as described in Milestone 2 of the proposal document)

This provides the functionality originally demoed as a prototype here: https://youtu.be/4iU5wXQrAn4?t=1022

All the behaviors are gated behind this FeatureConfiguration: `ENABLE_CATALOG_FEDERATION`

Special thanks to @XJDKC for co-authoring this PR with me!

PR #2355, opened 2025-08-14
Support HMS Federation

poojanilangekar:
Supports federating to HiveCatalog using the Iceberg REST library. 
All hive dependencies are added in an independent module, i.e., `polaris-extensions-federation-hive` and can be removed/converted to a compile time flag if necessary. 
Similar to HadoopCatalog, HMS federation support is currently restricted to `IMPLICIT` auth. The underlying authentication can be any form that Hive supports, however Polaris will not store and manage any of these credentials. Again, similar to HadoopCatalog this version supports federating to a single Hive instance. 

This PR relies on Polaris discovering the `hive-site.xml` file to get the configuration options from the classpath (including `HADOOP_CONF_DIR`).

Testing: 
Modified the regression test to locally test that Hive federation works as expected. The next step would be to add a regression test once the change is baked into the Polaris docker image (for CI builds).

This PR primarily builds on #1305 and #1466. Thank you @dennishuo and @eric-maynard for helping out with this.


poojanilangekar:
CC: @dennishuo @eric-maynard

eric-maynard (commented):
Can we get a regression test like the one in #2286?

poojanilangekar:
Yes, I have a test that I used locally. Unfortunately, I just can't add it to the test until we get this PR baked into the Docker image. :(

poojanilangekar:
Here is a [jupyter notebook](https://github.com/poojanilangekar/polaris/blob/hive-demo/regtests/hive_federation_demo.ipynb) I've used in the past to demo the feature. (in lieu of the regtests that I will add after this PR is merged).

eric-maynard (dismissed):

extensions/federation/hive/src/main/java/org/apache/polaris/extensions/federation/hive/HiveFederatedCatalogFactory.java
```diff
@@ -0,0 +1,62 @@
...
+    AuthenticationParametersDpo authenticationParametersDpo =
+        connectionConfigInfoDpo.getAuthenticationParameters();
+    if (authenticationParametersDpo.getAuthenticationTypeCode()
+        != AuthenticationType.IMPLICIT.getCode()) {
+      throw new IllegalStateException("Hive federation only supports IMPLICIT authentication.");
+    }
+    String warehouse = ((HiveConnectionConfigInfoDpo) connectionConfigInfoDpo).getWarehouse();
+    // Unlike Hadoop, HiveCatalog does not require us to create a Configuration object, the iceberg
+    // rest library find the default configuration by reading hive-site.xml in the classpath
+    // (including HADOOP_CONF_DIR classpath).
```
dennishuo, line 56: Could be worth adding a TODO here to better qualify the assertion about not requiring Configuration and requiring hive-site.xml (and so we don't forget), that if we *do* want Hadoop Configuration to come from runtime Polaris properties instead, that we need to call `hiveCatalog.setConf` *before* the call to `hiveCatalog.initialize` and our constructed `Configuration` should begin with `new Configuration(false)` to avoid loading the default hive-site.xml?

AFAICT doing those two actions *should* be sufficient in potential multi-catalog environments to at least prevent basic conf-leakage between HiveCatalog instances.

dennishuo (dismissed):
LGTM, though is this the case where we wanted compile-time flags to allow the Polaris user to choose to exclude the whole `org.apache.hive:hive-metastore` dependency, even if we chose not to use a compile-time flag for `HadoopCatalog` due to that being a breaking change?

extensions/federation/hive/src/main/java/org/apache/polaris/extensions/federation/hive/HiveFederatedCatalogFactory.java
poojanilangekar: Done.

sfc-gh-dhuo (approved):

dennishuo (approved):

PR #2361, opened 2025-08-15
Add PolarisDiagnostics field to TransactionalMetaStoreManagerImpl

XN137:

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.polaris.core.PolarisCallContext;
+import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.entity.AsyncTaskType;
 import org.apache.polaris.core.entity.EntityNameLookupRecord;
 import org.apache.polaris.core.entity.LocationBasedEntity;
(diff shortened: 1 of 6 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -268,14 +271,13 @@ private void dropEntity(
       @Nonnull PolarisPrivilege priv) {
 
     // validate non null arguments
-    callCtx.getDiagServices().checkNotNull(securable, "unexpected_null_securable");
-    callCtx.getDiagServices().checkNotNull(grantee, "unexpected_null_grantee");
-    callCtx.getDiagServices().checkNotNull(priv, "unexpected_null_priv");
+    diagnostics.checkNotNull(securable, "unexpected_null_securable");
```
dimas-b, line 274: This effectively makes `PolarisDiagnostics` realm-scoped at most.  In current Apache Polaris code `PolarisDiagnostics` is actually `@ApllicationScoped` so this change is fine from my POV.

I wonder what the intended use case for `PolarisDiagnostics` in downstream projects was initially... @dennishuo @eric-maynard WDYT?

Based on current codebase, from my POV, `PolarisDiagnostics` can remain `@ApllicationScoped`.

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
eric-maynard: From what I know, `PolarisDiagnostics` is intended to serve as an extensible way to handle asserts/errors -- so integrators can implement anything from custom logging to automatically-filed tickets when an issue occurs. To that end, the mapping between CallContext and PolarisDiagnostics is useful if the custom logic will be somehow dependent on the caller or the realm.

Having said that, looking at the current usage, I agree that it seems like `@ApplicationScoped` is appropriate.

dimas-b: Thanks for the insight, @eric-maynard !

PR #2364, opened 2025-08-15
Fix soft-merge conflict on `main`

snazy:


XN137 (approved):

runtime/service/src/test/java/org/apache/polaris/service/catalog/AbstractPolicyCatalogTest.java
```diff
@@ -194,7 +194,7 @@ public void before(TestInfo testInfo) {
             metaStoreManager,
             userSecretsManager,
             securityContext,
-            new PolarisAuthorizerImpl(callContext.getRealmConfig()),
+            new PolarisAuthorizerImpl(polarisContext.getRealmConfig()),
```
XN137: actually we have a `realmConfig` field now that we should use

PR #1619, opened 2025-05-19
Set the default value of DROP WITH PURGE to false

williamhyun:
As per #1617 , we will be setting the default value of DROP WITH PURGE to false.

Issue #1617, opened 2025-05-19
DROP WITH PURGE should be disabled by default


eric-maynard:
### Describe the bug

As mentioned in #774 and elsewhere, DROP WITH PURGE is currently somewhat unreliable. While there is some ongoing work to make this operation more scalable and robust, we should disable this by default for the time being. A flag was implemented in #270 to disable DROP WITH PURGE, so we only need to flip the default now.

dimas-b:
Can this be closed now that #1619 got merged?

PR #2369, opened 2025-08-16
Add feature config to allow dropping views without purging

### LINKED ISSUE #2367: Unable to drop view with default server configuration
### Describe the bug

Run Polaris from recent `main` with all config defaults.

In Spark SQL do:
```
spark-sql (default)> use polaris;
spark-sql ()> create namespace ns;
spark-sql ()> create view ns.v1 as select 'abc';

spark-sql ()> drop view ns.v1;
25/08/15 13:18:09 ERROR SparkSQLDriver: Failed in [drop view ns.v1]
org.apache.iceberg.exceptions.ForbiddenException: Forbidden: Unable to purge entity: v1. To enable this feature, set the Polaris configuration DROP_WITH_PURGE_ENABLED or the catalog configuration polaris.config.drop-with-purge.enabled
```

Note that the client-side SQL does not request "PURGE".

### To Reproduce

_No response_

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Additional context

_No response_

### System information

_No response_

dimas-b:
With tables, the client can decide whether to purge the table on drop or not. However, Polaris Servers used to unconditionally perform the purge on dropping a view.

After #1619 that behaviour effectively prevents dropping views if the admin user does not set `DROP_WITH_PURGE_ENABLED`. The latter, though, is not currently advisable per #1617.

This change introduces a new feature configuration (`PURGE_VIEWS_ON_DROP`) that allows the admin user to instruct Polaris servers to drop views without purging to achieve operational parity with tables.

Fixes #2367


dimas-b:
(made the branch at the main repo accidentally :facepalm: will delete it when the PR is done)

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -191,6 +191,15 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> PURGE_VIEWS_ON_DROP =
+      PolarisConfiguration.<Boolean>builder()
+          .key("PURGE_VIEWS_ON_DROP")
+          .catalogConfig("polaris.config.purge-views-on-drop")
+          .description(
+              "Indicates whether Polaris should purge view metadata files when a view is dropped.")
```
eric-maynard:
```suggestion
              "If set to true, Polaris will attempt to delete metadata files when a view is dropped.")
```

dimas-b:
@eric-maynard : I renamed the feature flag to `PURGE_VIEW_METADATA_ON_DROP` per email discussion. Since there were no strong opinions, I'm open to going back to the previous name or any other name you suggest :)

https://lists.apache.org/thread/k94o7k0f6d2cqgqx24v8j21z01ho34sl

eric-maynard (approved):
Still LGTM!

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: @eric-maynard : I lost this change and had to re-apply it. Please approve again :sweat_smile:

dimas-b:
rebased and resolved conflicts

snazy (approved):

PR #2370, opened 2025-08-16
Optimize PolicyCatalog.listPolicies

XN137:
this is a follow-up to https://github.com/apache/polaris/pull/2290

the optimization is to use `listEntities` instead of `loadEntities` when
there is no `policyType` filter to apply

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/PolicyCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/PolicyCatalog.java
@@ -24,6 +24,7 @@ import static org.apache.polaris.service.types.PolicyAttachmentTarget.TypeEnum.C
 
 import com.google.common.base.Strings;
 import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/PolicyCatalog.java
```diff
@@ -153,24 +157,46 @@ public Policy createPolicy(
...
     }
 
-    List<PolarisEntity> catalogPath = resolvedEntities.getRawFullPath();
-    // TODO: when the "policyType" filter is null we should only call "listEntities" instead
+    List<PolarisEntityCore> catalogPath =
+        PolarisEntity.toCoreList(resolvedEntities.getRawFullPath());
+    if (policyType == null) {
+      // without a policyType filter we can call listEntities to acquire the entity names
+      ListEntitiesResult listEntitiesResult =
+          metaStoreManager.listEntities(
```
dimas-b, line 172: I believe this deserves a test to ensure the results are the consistent on both call paths.

dimas-b (commented):
LGTM overall :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/PolicyCatalog.java
XN137: i did not mention this but i had checked before that both call paths are already covered by tests [here](https://github.com/apache/polaris/blob/bc5d0c49f788e3520761ba4aae1c19288a60452f/runtime/service/src/test/java/org/apache/polaris/service/catalog/AbstractPolicyCatalogTest.java#L312-L356) afaict

dimas-b: Nice :+1: Thx!

dimas-b (approved):

pingtimeout (approved):

PR #2372, opened 2025-08-18
Add PolarisDiagnostics field to AbstractTransactionalPersistence

XN137:


polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/AbstractTransactionalPersistence.java
```diff
@@ -51,6 +52,13 @@
  * the BasePersistence methods in terms of lower-level methods that subclasses must implement.
  */
 public abstract class AbstractTransactionalPersistence implements TransactionalPersistence {
+
+  protected final PolarisDiagnostics diagnostics;
```
dimas-b: I'd prefer a `private` field with a `protected` accessor. Direct field access complicates debugging and code analysis. Also, this class is likely to be reused downstream, which makes accessors more relevant, IMHO. WDYT?

XN137: ok made the field private and switched to accessor

dimas-b (approved):

PR #2380, opened 2025-08-18
feat: Add Pod Disruption Budget support to Helm chart

### LINKED ISSUE #2345: Helm chart should support provision of PodDisruptionBudged

When Polaris is deployed via Helm with autoscaling.minReplicas > 1 or replicaCount > 1 it should be able to provision PodDisruptionBudged https://kubernetes.io/docs/tasks/run-application/configure-pdb/ to improve fault tolerance.

Easiest solution is to create PDB with opt-in configuration. Some trivial configuration like
```
podDisruptionBudget:
  enabled: true
  minAvailable: <count-or-percent>
  maxUnavailable: <count-or-percent>
```

which will generate correct PodDisruptionBudget with `selector.matchLabels` to the deployed Polaris

_No response_

### Additional context

_No response_

bmlyr:
Add PodDisruptionBudget support to Helm chart (https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
This PR should resolve #2345

### DIFF
--- a/helm/polaris/tests/poddisruptionbudget_test.yaml
+++ b/helm/polaris/tests/poddisruptionbudget_test.yaml
@@ -200,14 +200,11 @@ tests:
             app.kubernetes.io/instance: polaris-release
 
   # validation tests
-  - it: should not allow both minAvailable and maxUnavailable (minAvailable takes precedence)
+  - it: should fail when both minAvailable and maxUnavailable are set
     set:
       podDisruptionBudget.enabled: true
       podDisruptionBudget.minAvailable: 1
       podDisruptionBudget.maxUnavailable: 1
     asserts:
-      - equal:
-          path: spec.minAvailable
-          value: 1
-      - isNull:
-          path: spec.maxUnavailable
+      - failedTemplate:
+          errorMessage: "podDisruptionBudget.minAvailable and podDisruptionBudget.maxUnavailable cannot be both set."
--- a/helm/polaris/values.yaml
+++ b/helm/polaris/values.yaml
@@ -68,11 +68,11 @@ podDisruptionBudget:
   enabled: false
   # -- The minimum number of pods that should remain available during disruptions.
   # Can be an absolute number (ex: 5) or a percentage of desired pods (ex: 50%).
-  # IMPORTANT: Cannot be used simultaneously with maxUnavailable. If both are set, minAvailable takes precedence.
+  # IMPORTANT: Cannot be used simultaneously with maxUnavailable.
   minAvailable: ~
   # -- The maximum number of pods that can be unavailable during disruptions.
   # Can be an absolute number (ex: 5) or a percentage of desired pods (ex: 50%).
-  # IMPORTANT: Cannot be used simultaneously with minAvailable. If both are set, minAvailable takes precedence.
+  # IMPORTANT: Cannot be used simultaneously with minAvailable.
   maxUnavailable: ~
   # -- Annotations to add to the pod disruption budget.
   annotations: {}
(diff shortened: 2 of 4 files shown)


helm/polaris/values.yaml
```diff
@@ -62,6 +62,21 @@ podLabels: {}
...
+  # -- Specifies whether a pod disruption budget should be created.
+  enabled: false
+  # -- The minimum number of pods that should remain available during disruptions.
+  # Can be an absolute number (ex: 5) or a percentage of desired pods (ex: 50%).
+  # Cannot be used if maxUnavailable is set.
+  minAvailable: ~
+  # -- The maximum number of pods that can be unavailable during disruptions.
+  # Can be an absolute number (ex: 5) or a percentage of desired pods (ex: 50%).
+  # Cannot be used if minAvailable is set.
+  maxUnavailable: 1
```
adutra: Setting this to 1 by default forces users to unset this value, if they want to set `minAvailable`. How about we don't specify any default value for any of the fields?

adutra:
@bmlyr I think you need to rebase your PR, there is some strange compilation failure right now.

helm/polaris/values.yaml
bmlyr: Sure it's better, I have set this value to ~ now

bmlyr:
@adutra Might be better now

helm/polaris/tests/poddisruptionbudget_test.yaml
```diff
@@ -0,0 +1,214 @@
...
+          value: "50%"
+      - isNull:
+          path: spec.minAvailable
+
+  # spec.minAvailable
+  - it: should set minAvailable and unset maxUnavailable
+    set:
+      podDisruptionBudget.enabled: true
+      podDisruptionBudget.minAvailable: 2
+      podDisruptionBudget.maxUnavailable: null
```
adutra, line 152: Explicitly unsetting this is not necessary anymore, but it doesn't hurt to keep this line either.

helm/polaris/tests/poddisruptionbudget_test.yaml
```diff
@@ -0,0 +1,214 @@
...
+      fullnameOverride: polaris-override
+    asserts:
+      - equal:
+          path: spec.selector.matchLabels
+          value:
+            app.kubernetes.io/name: polaris
+            app.kubernetes.io/instance: polaris-release
+
+  # validation tests
+  - it: should handle both minAvailable and maxUnavailable set (minAvailable takes precedence)
```
adutra: This would be rather a pathological case imho, but OK to test it.

adutra (approved):

helm/polaris/tests/poddisruptionbudget_test.yaml
adutra: Apparently, this would fail to apply:

```
one or more objects failed to apply, reason: PodDisruptionBudget.policy [...] is invalid: [...] minAvailable and maxUnavailable cannot be both set.
```

bmlyr: Should we prevent the user from doing this? Maybe by selecting only minAvailable?

adutra: Maybe? We can also fail hard e.g.

```
{{- if and .Values.podDisruptionBudget.minAvailable .Values.podDisruptionBudget.maxUnavailable }}
{{- fail "podDisruptionBudget.minAvailable and podDisruptionBudget.maxUnavailable cannot be both set." -}}
{{- end }}
```

bmlyr: I made a few changes to avoid that. Let me know if you prefer this version

bmlyr: You're right, failing is a better option

bmlyr: Might be better now

adutra (approved):

helm/polaris/tests/poddisruptionbudget_test.yaml
adutra: Very nice, thank you 👍

adutra:
Thank you for your contribution @bmlyr !

PR #2381, opened 2025-08-19
Add PolarisDiagnostics field to BaseMetaStoreManager

XN137:

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
@@ -93,6 +93,7 @@ public class TransactionalMetaStoreManagerImpl extends BaseMetaStoreManager {
   private final PolarisDiagnostics diagnostics;
 
   public TransactionalMetaStoreManagerImpl(Clock clock, PolarisDiagnostics diagnostics) {
+    super(diagnostics);
     this.clock = clock;
     this.diagnostics = diagnostics;
   }
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
@@ -90,12 +90,10 @@ public class TransactionalMetaStoreManagerImpl extends BaseMetaStoreManager {
       LoggerFactory.getLogger(TransactionalMetaStoreManagerImpl.class);
 
   private final Clock clock;
-  private final PolarisDiagnostics diagnostics;
 
   public TransactionalMetaStoreManagerImpl(Clock clock, PolarisDiagnostics diagnostics) {
     super(diagnostics);
     this.clock = clock;
-    this.diagnostics = diagnostics;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -93,6 +93,7 @@ public class TransactionalMetaStoreManagerImpl extends BaseMetaStoreManager {
   private final PolarisDiagnostics diagnostics;
 
   public TransactionalMetaStoreManagerImpl(Clock clock, PolarisDiagnostics diagnostics) {
+    super(diagnostics);
```
XN137, line 95: how to handle this case?
in https://github.com/apache/polaris/pull/2372 @dimas-b suggested to make fields in the base class private and have a protected getter...
when we do this in this file as well, we need to adjust 49 usage spots of the field.
is it worth it for consistency?

snazy: nah, let's keep the protected field

snazy (dismissed):

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
XN137: the existing field in this class is private (and the one we are adding to the base class is private as well).
they are currently initialized with the same value, which may be odd.

snazy: Right, I meant to make the field in `BaseMetaStoreManager` protected.

XN137: sorry i am not following, do you want me to make the field in `BaseMetaStoreManager` protected and avoid the getter? currently its private and has a protected getter (as suggested in the other PR).

the thread here is about the existing field in `TransactionalMetaStoreManagerImpl` which is now kind of redundant.

dimas-b: I would not mind keeping the `private` field in this class to avoid exposing fields from the base class.

IMHO, direct field access to the base class in this case will complicate the work of establishing a "clean" Polaris java API later (soon).

dimas-b: I'm also fine with using a `protected` getter method from the base class.

dimas-b: I slightly prefer the `protected getDiagnostics()` method for the sake of consistency with #2372.

XN137: ok i've removed the field and using `getDiagnostics()` now

XN137:
rebased after small conflict in `PolarisTreeMapAtomicOperationMetaStoreManagerTest.java`

dimas-b (approved):

dimas-b:
If there are no fresh comments, I propose to merge on Aug 22.

PR #2383, opened 2025-08-19
Automate the release guide - Take 2 - Github workflows

pingtimeout:
This PR is the next iteration of release automation.  It builds on top of #2156 and reuses the common bash libraries that were defined.

### Differences from the initial PR
* Release automation can only be triggered via Github Workflows.  It is not possible to perform a semi-automated release from a committer/PMC computer.
* It assumes that the following secrets are defined:
  * `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` - the credentials that can be used to push Docker images to Dockerhub
  * `GPG_PRIVATE_KEY` and `GPG_PASSPHRASE` - the ASCII armored private key that should be used to sign artifacts and its associated passphrase.  The associated public key is assumed to be added to the [`KEYS`](https://downloads.apache.org/incubator/polaris/KEYS) file prior to being entered here.
  * `APACHE_USERNAME` and `APACHE_PASSWORD` - the credentials that can be used to connect to the ASF SVN server **as well as** to the ASF Nexus server.
* All code that was used to perform release steps locally has been removed to keep the PR as small as possible.

### Similarities with the initial PR
This PR builds on the same assumptions from the initial PR:

* Release cannot be fully automated as of today as there are concerns that the release guide may not be comprehensive.  Hence full automation is not desirable yet.
* Release can only be semi-automated given that certain operations must be manually performed by the release manager.

### Remaining known-unknowns
I can see that `apache/polaris` has [a `Nightly build` github workflow](https://github.com/apache/polaris/actions/runs/17040500750) that publishes snapshots every night to the Apache Nexus repository.  In this workflow's definition, I can find references two Nexus credentials (`secrets.NEXUS_USER` and `secrets.NEXUS_PW`).  However I cannot find any such secret defined on https://github.com/apache/polaris/settings/secrets/actions, nor am I sure whether those credentials can be used to interact with the ASF SVN server.  Some clarification is needed to ensure proper credentials configuration.

### Example runs
I have used this PR to simulate the release of Polaris 99.98.97-incubating-rc1 on my own fork.  No upload was performed, this is just to prove out that things should be working as expected.  You can find links to the following workflow executions:

* This [Create Release Branch](https://github.com/pingtimeout/polaris/actions/runs/17233943316) workflow was used to cut the release branch with proper naming pattern: `release/99.98.97-incubating`.  It was run with `dry-run=0` so that the release branch was actually created.
* This [Update version and Changelog for Release Candidate](https://github.com/pingtimeout/polaris/actions/runs/17233984671) workflow was used to set the Polaris version to `99.98.97-incubating`, update the changelog and push the RC1 tag.  It was also run with `dry-run=0` so that the modifications were actually performed (on my fork only)
* For comparison, this [Update version and Changelog for Release Candidate](https://github.com/pingtimeout/polaris/actions/runs/17208581924) workflow is what happens when we try to cut an RC from a commit where some Github checks have failed (e.g. CI)
* From there, all subsequent workflows were run with `dry-run=1` so that no interaction with any ASF server happened.
* This [Build and Publish Release Artifacts](https://github.com/pingtimeout/polaris/actions/runs/17234090721) workflow was run with `dry-run=1` to check the commands that would be executed, if an actual release was to be performed.  We can see the binaries and Helm Charts would be built, signed, checksum'ed and publish to ASF SVN and Nexus repositories.  Docker images would be built but *not* published.
* And finally, this [Publish Release After Vote Success](https://github.com/pingtimeout/polaris/actions/runs/17234155255) workflow shows how the artifacts would be moved from the dist dev space to the dist release space on ASF SVN, the final release tag would be created and the Nexus repository would be automatically released.  This is also where the Github Release itself would be created.

### Next steps
If this PR is approved, it should be possible to use the `Create Release Branch` and `Update Release Candidate` workflows to respectively cut the release branch and set the RC version as well as create the RC tag.

### Recommendations for reviewers
* Please review the `releasey/release-process-flowchart.png` flowchart first.  It shows how each workflow participates to the overall release process.  It should substantially help make sense of the Github workflows for the rest of the review.

### DIFF
--- a/.github/workflows/release-3-build-and-publish-artifacts.yml
+++ b/.github/workflows/release-3-build-and-publish-artifacts.yml
@@ -251,6 +251,16 @@ jobs:
           echo "RELEASEY_DIR=$(pwd)/releasey" >> $GITHUB_ENV
           echo "LIBS_DIR=$(pwd)/releasey/libs" >> $GITHUB_ENV
 
+      - name: Set up Docker Buildx
+        run: |
+          docker buildx use default
+          docker buildx create \
+          --platform linux/amd64,linux/arm64 \
+          --use \
+          --name polarisbuild \
+          --driver-opt network=host || docker buildx use polarisbuild
+          docker buildx inspect
+
       - name: Set up Java
         uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5
         with:
--- a/.github/workflows/release-4-publish-release.yml
+++ b/.github/workflows/release-4-publish-release.yml
@@ -243,7 +243,15 @@ jobs:
           Final release tag \`${final_release_tag}\` created and pushed
           EOT
 
-
+      - name: Set up Docker Buildx
+        run: |
+          docker buildx use default
+          docker buildx create \
+          --platform linux/amd64,linux/arm64 \
+          --use \
+          --name polarisbuild \
+          --driver-opt network=host || docker buildx use polarisbuild
+          docker buildx inspect
 
       - name: Set up Java
         uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5


.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,261 @@
...
+          source "${LIBS_DIR}/_log.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          print_info "Creating RC tag ${release_tag} at commit ${new_tag_ref}"
+
+          # Create the tag at the new commit
+          exec_process git tag "${release_tag}" "${new_tag_ref}"
+          exec_process git push origin "${release_tag}"
+
+          print_success "🎉 RC tag ${release_tag} created successfully at commit ${new_tag_ref}!"
```
snazy: This is an example of information that would be useful to have in the GH workflow summary.

.github/workflows/release-1-create-release-branch.yml
```diff
@@ -0,0 +1,119 @@
...
+            echo "DRY_RUN mode enabled - no actual changes will be made"
+          else
+            echo "DRY_RUN=0" >> $GITHUB_ENV
+            echo "DRY_RUN mode disabled - actual changes will be made"
+          fi
+
+      - name: Validate release parameters
+        run: |
+          releasey_dir="$(pwd)/releasey"
+          LIBS_DIR="${releasey_dir}/libs"
```
snazy: Those can be moved into one step that exports the two variables.

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,261 @@
...
+    name: Release - 2 - Update version and Changelog for Release Candidate
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          # Fetch full history for proper branch operations
```
snazy: Nit:
```suggestion
          # Fetch full history. Branch operations require this.
```

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,261 @@
...
+      - name: Configure Git
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Set DRY_RUN environment variable
+        run: |
+          if [[ "${{ github.event.inputs.dry_run }}" == "true" ]]; then
+            echo "DRY_RUN=1" >> $GITHUB_ENV
+            echo "DRY_RUN mode enabled - no actual changes will be made"
```
snazy: This information is also useful to see in the summary - maybe as a "big red warning" that it's a dry run.

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,261 @@
...
+            print_error "Invalid release branch version format: ${branch_version}"
+            print_error "Expected format: x.y.z-incubating"
+            exit 1
+          fi
+
+          print_info "Parsed version components from branch:"
+          print_info "  Major: ${major}"
+          print_info "  Minor: ${minor}"
+          print_info "  Patch: ${patch}"
+          print_info "  Version without RC: ${version_without_rc}"
```
snazy: Also useful in the WF summary.

Mind adding the version also as a concatenated string with and without the RC suffix? For the lazy copy-paste-people out there.

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,261 @@
...
+
+          if [[ ${rc_number} -eq 1 ]]; then
+            print_info "Creating new GitHub release: ${release_title}"
+            exec_process gh release create "${release_tag}" \
+              --title "${release_title}" \
+              --prerelease \
+              --target "${new_tag_ref}"
+
+            print_success "🎉 GitHub release created: ${release_title}"
+          else
```
snazy: I think what's missing here is the cleanup of the previous RC, like purging data in SVN and deleting the previous's RC Nexus staging repo.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,195 @@
...
+      - name: Build source and binary distributions
+        run: |
+          releasey_dir="$(pwd)/releasey"
+          LIBS_DIR="${releasey_dir}/libs"
+
+          source "${LIBS_DIR}/_log.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          print_info "Building source and binary distributions..."
+          exec_process ./gradlew build sourceTarball -Prelease -PuseGpgAgent -x test -x intTest
```
snazy:
```suggestion
          exec_process ./gradlew assemble sourceTarball -Prelease -PuseGpgAgent
```
`test` and `intTest` are not the only check tasks. `build` depends on the `assemble` and `check` tasks. But we don't need to run any checks here, as the Git commit status is checked.

pingtimeout:
Thanks for the review @snazy, I have updated the code to leverage step summaries and so that workflows 3, 4 and 5 are merged into jobs of a single workflow.  For now, I am keeping the separation between workflows 2 and 3 for testing purposes: it is possible to run workflow 2 in real mode (create branch, update changelog, ...) but workflows 3 and above should only be run in dry-run mode (no publication to Apache, etc...) until we have the infra setup to perform automatic releases (secrets).

I am going to run a bunch of test runs with those changes and will keep this PR updated.

EDIT: I updated this PR description with the new workflow runs that contain github step summaries

pingtimeout:
I compared the list of commands that JB executed for the 1.1.0 release with the ones that would have been executed by the workflow and found the following differences:

1. The release branch is not identical.  As per the release guide, it should be `release/1.1.0-incubating`, but the one JB used was `release/1.1.x`, which is probably better, should we want to use that branch to cut new patch releases.
2. JB commands contained a `./gradlew clean build`.  This is unnecessary as we can only execute the workflow against a commit which has been verified by CI.
3. JB commands contained a `./gradlew build -x test -x intTest`.  In this PR, @snazy pointed out that this is unnecessary and could be replaced by `./gradlew assemble`, which is what the workflow produced
4. The workflow creates two SVN commits on dist/dev: one for the binaries and one for the Helm charts.  JB added everything into a single commit.  I am going to update the workflows accordingly.
5. JB closed the Nexus repository manually, whereas it _should_ be possible to automate this with `./gradlew closeApacheStagingRepository`, which is what the workflows use.
6. The workflows try to add the `index.yaml` file from the wrong directory.

Points (1), (4) and (6) should be fixed.  Points (2) and (3) can imo be ignored.  @jbonofre may I request you to verify the closeApacheStagingRepository command from point (5) please?

jbonofre:
The branch name is similar with the previous ones.

jbonofre:
I will take a look on the Nexus API to close the repo.

pingtimeout:
Actually, item (4) might not be doable without increasing the complexity of the workflow.  The two commits are caused by an implementation detail: as @snazy suggested, we separate the job that builds the binaries from the job that builds the Helm charts so that they can be independently retried, should any of them fail.

Which means that to have a single commit for both the binary and the Helm charts, we would have to start sharing directories across jobs.  And that, in turn, means that even if the job that builds binaries succeeds, those binaries may not immediately be uploaded on dist/dev.

So I suggest we keep the two separate commits.

Wdyt?

pingtimeout:
@jbonofre Could you confirm the name of the secrets that should be used by the workflows?  So far, these are the ones that are used:

* `APACHE_USERNAME` and `APACHE_PASSWORD` are the SVN credentials that should be used to checkout the dist dev and dist release repositories and add binaries
* `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` are the credentials used to publish the final Docker images to DockerHub
* `GPG_PRIVATE_KEY` is the ASCII armored private key that should be used to sign artifacts
* `GPG_PASSPHRASE` is the passphrase associated to the private key

If that's ok, let's move forward with creating the missing secrets on Github and prepare the execution of those workflows for the October release.

.github/workflows/release-4-publish-release.yml
```diff
@@ -0,0 +1,340 @@
...
+          exec_process svn export --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive \
+            "${release_artifacts_url}" "${artifacts_dir}/artifacts"
+
+          # Download Helm chart artifacts
+          exec_process svn export --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive \
+            "${release_helm_url}" "${artifacts_dir}/helm"
+
+          # Create GitHub release
+          release_title="Release ${version_without_rc}"
+          release_notes="Apache Polaris ${version_without_rc} Release
```
snazy, line 322: Can you replace this variable assignment with a `cat << EOT` or so? Or surround with some `# ************************` comment lines?
It's a bit hard to find the closing/ending `"`.

.github/workflows/release-4-publish-release.yml
```diff
@@ -0,0 +1,340 @@
...
+
+      - name: Release candidate repository on Nexus
+        env:
+          ORG_GRADLE_PROJECT_apacheUsername: ${{ secrets.APACHE_USERNAME }}
+          ORG_GRADLE_PROJECT_apachePassword: ${{ secrets.APACHE_PASSWORD }}
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+
+          # Use the Gradle task to release the Apache staging repository
+          exec_process ./gradlew releaseApacheStagingRepository
```
snazy: I'm not sure this works without additional information like the staging repository ID (see [here](https://github.com/apache/polaris/pull/485/files#diff-064f97061477cc170b443791d7292896086582753d2a3b8ebf74bb69196a562aR287-R288)) and capturing those values [in release-3...](https://github.com/apache/polaris/pull/2383/files#diff-b103fc0293df91eb81d72ea7c5bcf5ab5b3fa9ea7b943ae09ad5791431087a9cR182) (see  [here](https://github.com/apache/polaris/pull/485/files#diff-4d004b07a60437298ae0dad3ddfb2d0eed301a1a1a5a277420286ef7f60b0836R331-R345)).

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,352 @@
...
+
+          dist_dev_dir=${RELEASEY_DIR}/polaris-dist-dev
+          exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
+
+          version_dir="${dist_dev_dir}/${version_without_rc}"
+          exec_process mkdir -p "${version_dir}"
+          exec_process cp build/distributions/* "${version_dir}/"
+          exec_process cp runtime/distribution/build/distributions/* "${version_dir}/"
+
+          exec_process cd "${dist_dev_dir}"
```
snazy, line 178: Does the `cd` become effective in this shell? (IIRC a shell function execution runs in its own shell.)

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,352 @@
...
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+
+      - name: Create Helm package
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+
+          exec_process cd helm
```
snazy, line 359: (same `cd` comment as above)

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,352 @@
...
+          source "${LIBS_DIR}/_constants.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          dist_dev_dir=${RELEASEY_DIR}/polaris-dist-dev
+          exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
+
+          exec_process mkdir -p "${dist_dev_dir}/helm-chart/${version_without_rc}"
+          exec_process cp helm/polaris-${version_without_rc}.tgz* "${dist_dev_dir}/helm-chart/${version_without_rc}/"
+
+          exec_process cd "${dist_dev_dir}/helm-chart"
```
snazy, line 384: (same `cd` comment as above)

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,352 @@
...
+          exec_process helm repo index .
+
+          exec_process cd "${dist_dev_dir}"
```
snazy, line 387: (same `cd` comment as above)

.github/workflows/release-4-publish-release.yml
```diff
@@ -0,0 +1,340 @@
...
+          source "${LIBS_DIR}/_constants.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          # Checkout the release Helm chart directory
+          release_helm_dir="${RELEASEY_DIR}/polaris-dist-release-helm-chart"
+          release_helm_url="${APACHE_DIST_URL}/release/incubator/polaris/helm-chart"
+
+          exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${release_helm_url}" "${release_helm_dir}"
+
+          exec_process cd "${release_helm_dir}"
```
snazy, line 220: (same `cd` comment as above)

releasey/libs/_github.sh
```diff
@@ -0,0 +1,102 @@
...
+source "${LIBS_DIR}/_constants.sh"
+source "${LIBS_DIR}/_exec.sh"
+
+function get_remote_url() {
+  # Get the URL for a given remote, return non-zero if remote doesn't exist
+  local remote_name="$1"
+  git remote get-url "${remote_name}" 2>/dev/null || return 1
+}
+
+function get_github_repo_info() {
```
snazy: Why not just use [`GITHUB_REPOSITORY`](https://docs.github.com/en/actions/reference/workflows-and-actions/variables)?

releasey/libs/_github.sh
```diff
@@ -0,0 +1,102 @@
...
+
+  # Get repository information from the current Git repository
+  local repo_info
+  repo_info=$(get_github_repo_info) || {
+    print_error "Failed to determine GitHub repository information"
+    return 1
+  }
+
+  local num_invalid_checks
+  local num_invalid_checks_retrieval_command="curl -s -H \"Authorization: Bearer ${GITHUB_TOKEN}\" -H \"Accept: application/vnd.github+json\" -H \"X-GitHub-Api-Version: 2022-11-28\" https://api.github.com/repos/${repo_info}/commits/${commit_sha}/check-runs | jq -r '[.check_runs[] | select(.conclusion != \"success\" and .conclusion != \"skipped\" and (.name | startswith(\"Release - \") | not))] | length'"
```
snazy: Nit: Can be simplified by using `gh api` command.

releasey/libs/_log.sh
```diff
@@ -0,0 +1,79 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Common Logging Functions
+#
+
+# Colors for output - only use colors if terminal supports them and user hasn't disabled them
```
snazy: Super nit:
```suggestion
# Colors for output - only use colors if terminal supports them
```

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: I believe the directory change is indeed in the current shell.  I was able to verify this locally:

```
bash-5.2$ source releasey/libs/_exec.sh
bash-5.2$ pwd
/Users/plaporte/env/home/dremio/projects/polaris
bash-5.2$ DRY_RUN=1 exec_process cd tools
DEBUG: Dry-run, WOULD execute 'cd tools'
bash-5.2$ DRY_RUN=0 exec_process cd tools
DEBUG: Executing 'cd tools'
bash-5.2$ pwd
/Users/plaporte/env/home/dremio/projects/polaris/tools
```

I suggest we keep it that way for now, until we have the opportunity to execute the workflow for real.  Rationale: merging the commands on a single line makes the thing a lot less readable:

```
          exec_process bash -c "cd '${dist_dev_dir}' && svn add '${version_without_rc}' && svn commit --username '$SVN_USERNAME' --password '$SVN_PASSWORD' --non-interactive -m 'Stage Apache Polaris ${version_without_rc} RC${rc_number}'"
```

So I would like to give it a try as-is, considering with a local bash shell it works.  We can always merge the commands later if it turns out to be a bug.  Wdyt?

releasey/libs/_github.sh
pingtimeout: TIL.  Also this function was created in the first version when the scripts had to be executable locally.  It is not necessary anymore, fixing.

.github/workflows/release-3-build-and-publish-artifacts.yml
snazy: WFM - your local bash should work the same a bash in GH runners ;)

.github/workflows/release-4-publish-release.yml
pingtimeout: Excellent !  That's exactly the kind of information I lacked because I do not have Nexus permissions to test this.  I am going to copy this pattern in the workflows.

pingtimeout: Done.

jbonofre:
I can create the secrets on the repository:
* `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` - the credentials that can be used to push Docker images to Dockerhub
* `GPG_PRIVATE_KEY` and `GPG_PASSPHRASE` - the ASCII armored private key that should be used to sign artifacts and its associated passphrase. The associated public key is assumed to be added to the [KEYS](https://downloads.apache.org/incubator/polaris/KEYS) file prior to being entered here.
* `APACHE_USERNAME` and `APACHE_PASSWORD` - the credentials that can be used to connect to the ASF SVN server as well as to the ASF Nexus server.

@pingtimeout I guess you assume we have a "release bot" account, right ? So it means that I have to create this `release_bot` account right ?

pingtimeout:
@jbonofre We could have a release bot but it's not assumed.  The workflows can (and are expected to) be triggered manually, at this time.

pingtimeout:
@snazy may I request a +1 for this PR?

jbonofre:
FYI: https://release-test.apache.org/tutorial

.github/workflows/gradle.yml
```diff
@@ -30,7 +30,7 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
+    branches: [ "main", "release/*" ]
```
jbonofre, line 33: NB: we should not protect the release branch for now. Having PRs is totally fine, but it would be painful for the release manager to protect the release branch (just fyi).

snazy: @jbonofre these changes only let CI run against those branches, those are not protected via `.asf.yaml`'s `github.protected_branches`.

NB: (about release-branches-protection) IMO changes to a release branch should be minimal compared to the `main` branch. If we ever do point-releases and consider some release an LTS release, we should apply the same level of protection to such a branch as we do for `main`. The more commits are on a release-branch, the more has to be verified for every release. Being confident that the changes went through a "standard" review process and pass CI is crucial IMO. WDYT?

pingtimeout: @jbonofre well, if you check [release-2-update-release-candidate.yml L131-L135](https://github.com/apache/polaris/pull/2383/files#diff-b48bf3b0ffc794c60c5c47deb82b150e3b440785ebde41cd2a32f1d7e191b4f9R131-R135), you can see that RC can only be drafted from commits that have been validated by CI.  This is so that we do not introduce any issue with a bad cherry-pick.  Removing the `release/*` part that you highlighted would change that.  And I believe this is one of the main features from #485 that @snazy wanted me to keep.

Wdyt?

releasey/libs/_version.sh
```diff
@@ -0,0 +1,176 @@
...
+  local existing_tags
+  existing_tags=$(git tag -l "${tag_pattern}" | sort -V)
+
+  if [[ -z "${existing_tags}" ]]; then
+    # No existing RC tags, start with RC0
+    rc_number=0
+  else
+    # Extract the highest RC number and increment
+    local highest_rc
+    highest_rc=$(echo "${existing_tags}" | sed "s/apache-polaris-${version_without_rc}-rc//" | sort -n | tail -1)
```
snazy, line 138: Nit: could use an array and leverage that it's already sorted.
```suggestion
  existing_tags=($(git tag -l "${tag_pattern}" | sort -V))

  if [[ ${#existing_tags[@]} -eq 0 ]]; then
    # No existing RC tags, start with RC0
    rc_number=0
  else
    # Extract the highest RC number and increment
    local highest_rc
    highest_rc=$(echo "${existing_tags[-1]//apache-polaris-${version_without_rc}-rc/")
```

releasey/libs/_version.sh
```diff
@@ -0,0 +1,176 @@
...
+  # This function finds the next available patch number for a given major.minor version.
+  # It returns 0 and sets the global variable patch to the next available patch number.
+  # Patch numbers start from 0. It takes major and minor as input (e.g., "1", "0").
+  local major="$1"
+  local minor="$2"
+
+  # Get all existing tags for this major.minor version
+  local tag_pattern="apache-polaris-${major}.${minor}.*-incubating-rc*"
+  local existing_tags
+  existing_tags=$(git tag -l "${tag_pattern}" | sort -V)
```
snazy, line 155: Nit: could leverage array type here as well.

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,200 @@
...
+
+          # Validate branch version format and extract components
+          if ! validate_and_extract_branch_version "${branch_version}"; then
+            echo "❌ Invalid release branch version format: \`${branch_version}\`, expected: major.minor.x" >> $GITHUB_STEP_SUMMARY
+            exit 1
+          fi
+
+          # Find the next available patch number for this major.minor version
+          find_next_patch_number "${major}" "${minor}"
+
```
snazy, line 95: Just to prevent some potential debugging round trips, it might be nice to print the results like
```suggestion
          echo "Next patch number: ${patch}"

```
here and after the other calls that "return" (change) a global variable. We can later remove that, if necessary. I'm just thinking that the initial runs may go sideways - Murphy's law and such... ;)

releasey/libs/_version.sh
```diff
@@ -0,0 +1,176 @@
...
+  minor="${BASH_REMATCH[2]}"
+  patch="${BASH_REMATCH[3]}"
+  version_without_rc="${major}.${minor}.${patch}-incubating"
+
+  return 0
+}
+
+function update_version {
+  local version="$1"
+  local current_version=$(cat "$VERSION_FILE")
```
snazy: To [prevent "masking" the return value](https://github.com/koalaman/shellcheck/wiki/SC2155) of `cat`
```suggestion
  local current_version
  current_version=$(cat "$VERSION_FILE")
```

releasey/libs/_version.sh
```diff
@@ -0,0 +1,176 @@
...
+
+function update_helm_version {
+  local old_version="$1"
+  local new_version="$2"
+  exec_process sed -i~ "s/${old_version}/${new_version}/g" "$HELM_CHART_YAML_FILE"
+  exec_process sed -i~ "s/${old_version}/${new_version}/g" "$HELM_README_FILE"
+  # The readme file may contain version with double dash for shields.io badges
+  # We need a second `sed` command to ensure that the version replacement preserves this double-dash syntax.
+  local current_version_with_dash=$(echo "$old_version" | sed 's/-/--/g')
+  local version_with_dash=$(echo "$version" | sed 's/-/--/g')
```
snazy: Nit: simplify ((SC2001)[https://github.com/koalaman/shellcheck/wiki/SC2001])
```suggestion
  local current_version_with_dash
  local version_with_dash
  current_version_with_dash="$(echo "${old_version//-/--}")"
  version_with_dash="$(echo "$version//-/--}")"
```

releasey/libs/_version.sh
```diff
@@ -0,0 +1,176 @@
...
+  update_helm_version "${current_version}" "${version}"
+}
+
+function update_version_txt {
+  local version="$1"
+  # This function is only there for dry-run support.  Because of the
+  # redirection, we cannot use exec_process with the exact command that will be
+  # executed.
+  if [[ ${DRY_RUN:-1} -ne 1 ]]; then
+    exec_process echo ${version} >$VERSION_FILE
```
snazy:
```suggestion
    exec_process echo "${version}" >$VERSION_FILE
```

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -0,0 +1,200 @@
...
+          EOT
+
+      - name: Commit and push changes
+        run: |
+          source "${LIBS_DIR}/_constants.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          # Commit version files and changelog
+          exec_process git add "$VERSION_FILE" "$HELM_CHART_YAML_FILE" "$HELM_README_FILE"
+          exec_process git add "$CHANGELOG_FILE"
```
snazy: Super nit:
```suggestion
          exec_process git add \
            "$VERSION_FILE" \
            "$HELM_CHART_YAML_FILE" \
            "$HELM_README_FILE" \
            "$CHANGELOG_FILE"
```

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+      - name: Stage artifacts to Apache dist dev repository
+        env:
+          SVN_USERNAME: ${{ secrets.APACHE_USERNAME }}
+          SVN_PASSWORD: ${{ secrets.APACHE_PASSWORD }}
+        run: |
+          source "${LIBS_DIR}/_constants.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          dist_dev_dir=${RELEASEY_DIR}/polaris-dist-dev
+          exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
```
snazy, line 171: This one worries me, as `exec_process` logs all arguments. I'm not sure how good GH's secrets masking is 🤷 

Sadly I couldn't find an easy way to provide SVN the credentials. It doesn't seem to support env vars for this and updating the auth-cache is also not that easy.

Mind adding an `exec_process` variant that doesn't log the whole command line but rather a provided string? Something like
```
exec_process2 "Performing svn checkout ..." svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
```

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+
+          version_dir="${dist_dev_dir}/${version_without_rc}"
+          exec_process mkdir -p "${version_dir}"
+          exec_process cp build/distributions/* "${version_dir}/"
+          exec_process cp runtime/distribution/build/distributions/* "${version_dir}/"
+
+          exec_process cd "${dist_dev_dir}"
+          exec_process svn add "${version_without_rc}"
+
+          exec_process svn commit --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive -m "Stage Apache Polaris ${version_without_rc} RC${rc_number}"
```
snazy, line 181: Similar here (`exec_process` + credentials)

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+
+      - name: Publish and close Apache Nexus staging repository
+        env:
+          ORG_GRADLE_PROJECT_apacheUsername: ${{ secrets.APACHE_USERNAME }}
+          ORG_GRADLE_PROJECT_apachePassword: ${{ secrets.APACHE_PASSWORD }}
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+
+          # Publish artifacts to staging repository
+          exec_process ./gradlew publishToApache closeApacheStagingRepository -Prelease -PuseGpgAgent --info > gradle_publish_output.txt 2>&1
```
snazy: Nit: maybe use 'tee` so the Gradle output is captured on the console.
```suggestion
          exec_process ./gradlew publishToApache closeApacheStagingRepository -Prelease -PuseGpgAgent --info 2>&1 | tee gradle_publish_output.txt
```

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+
+      - name: Build Polaris Server Docker image
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+
+          exec_process ./gradlew :polaris-server:assemble :polaris-server:quarkusAppPartsBuild --rerun \
+            -Dquarkus.container-image.build=true \
+            -Dquarkus.container-image.push=false \
+            -Dquarkus.docker.buildx.platform="linux/amd64,linux/arm64" \
+            -Dquarkus.container-image.tag="${git_tag}"
```
snazy: Should this be 
```suggestion
            -Dquarkus.container-image.tag="${version}"
```
?

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+
+      - name: Build Polaris Admin Tool Docker image
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+
+          exec_process ./gradlew :polaris-admin:assemble :polaris-admin:quarkusAppPartsBuild --rerun \
+            -Dquarkus.container-image.build=true \
+            -Dquarkus.container-image.push=false \
+            -Dquarkus.docker.buildx.platform="linux/amd64,linux/arm64" \
+            -Dquarkus.container-image.tag="${git_tag}"
```
snazy:
```suggestion
            -Dquarkus.container-image.tag="${version}"
```
?

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+          echo "LIBS_DIR=$(pwd)/releasey/libs" >> $GITHUB_ENV
+
+      - name: Set up Helm
+        uses: azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4
+        with:
+          version: 'latest'
+
+      - name: Install Helm GPG plugin
+        run: |
+          helm plugin install https://github.com/technosophos/helm-gpg || true
```
snazy: Why's the `|| true` needed? Mind adding a comment?

releasey/libs/_exec.sh
```diff
@@ -0,0 +1,46 @@
...
+function calculate_sha512 {
+  local source_file="$1"
+  local target_file="$2"
+  # This function is only there for dry-run support.  Because of the
+  # redirection, we cannot use exec_process with the exact command that will be
+  # executed.
+  if [[ ${DRY_RUN:-1} -ne 1 ]]; then
+    exec_process shasum -a 512 "${source_file}" > "${target_file}"
+  else
+    exec_process "shasum -a 512 ${source_file} > ${target_file}"
```
snazy, line 44: Nit: could remove the 2nd argument
```suggestion
    exec_process shasum -a 512 "${source_file}" > "${source_file}.sha512"
  else
    exec_process "shasum -a 512 ${source_file} > ${source_file}.sha512"
```

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+      - name: Stage Helm chart to Apache dist dev repository
+        env:
+          SVN_USERNAME: ${{ secrets.APACHE_USERNAME }}
+          SVN_PASSWORD: ${{ secrets.APACHE_PASSWORD }}
+        run: |
+          source "${LIBS_DIR}/_constants.sh"
+          source "${LIBS_DIR}/_exec.sh"
+
+          dist_dev_dir=${RELEASEY_DIR}/polaris-dist-dev
+          exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
```
snazy, line 379: (similar svn/credentials as above)

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+          exec_process mkdir -p "${dist_dev_dir}/helm-chart/${version_without_rc}"
+          exec_process cp helm/polaris-${version_without_rc}.tgz* "${dist_dev_dir}/helm-chart/${version_without_rc}/"
+
+          exec_process cd "${dist_dev_dir}/helm-chart"
+          exec_process helm repo index .
+
+          exec_process cd "${dist_dev_dir}"
+          exec_process svn add "helm-chart/${version_without_rc}"
+
+          exec_process svn commit --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive -m "Stage Apache Polaris Helm chart ${version_without_rc} RC${rc_number}"
```
snazy, line 390: (similar svn/credentials as above)

.github/workflows/release-4-publish-release.yml
```diff
@@ -0,0 +1,359 @@
...
+    inputs:
+      dry_run:
+        description: 'Dry run mode (check to enable, uncheck to perform actual operations)'
+        required: false
+        type: boolean
+        default: true
+      staging_repository_id:
+        description: 'Nexus staging repository ID to release (e.g., orgapachepolaris-1234)'
+        required: true
+        type: string
```
snazy, line 33: For the future:
I think we could store the staging repo ID as a separate file in SVN and read it in this workflow. WDYT?

.github/workflows/release-4-publish-release.yml
```diff
@@ -0,0 +1,359 @@
...
+          source "${LIBS_DIR}/_exec.sh"
+
+          # Define source and destination URLs
+          dev_artifacts_url="${APACHE_DIST_URL}/dev/incubator/polaris/${version_without_rc}"
+          release_artifacts_url="${APACHE_DIST_URL}/release/incubator/polaris/${version_without_rc}"
+
+          dev_helm_url="${APACHE_DIST_URL}/dev/incubator/polaris/helm-chart/${version_without_rc}"
+          release_helm_url="${APACHE_DIST_URL}/release/incubator/polaris/helm-chart/${version_without_rc}"
+
+          exec_process svn mv --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive \
```
snazy, line 186: (exec_process / svn creds, as above)
and `svn` below

.github/workflows/release-1-create-release-branch.yml
```diff
@@ -0,0 +1,119 @@
...
+
+          echo "## Mode" >> $GITHUB_STEP_SUMMARY
+          if [[ "${{ github.event.inputs.dry_run }}" == "true" ]]; then
+            echo "DRY_RUN=1" >> $GITHUB_ENV
+            echo "‼️ DRY_RUN mode enabled - No actual changes will be made" >> $GITHUB_STEP_SUMMARY
+          else
+            echo "DRY_RUN=0" >> $GITHUB_ENV
+            echo "DRY_RUN mode disabled - Performing actual operations" >> $GITHUB_STEP_SUMMARY
+          fi
+
```
snazy, line 67: Nit: Mind adding this to all jobs that may pull Docker images to use GH's docker.io mirror?
```
      - name: Setup test environment
        uses: ./.github/actions/setup-test-env
```

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,357 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run mode (check to enable, uncheck to perform actual operations)'
+        required: false
+        type: boolean
+        default: true
+
+jobs:
```
snazy, line 31: Mind adding a final job (maybe in a follow-up) to print the proposed `[VOTE]` email subject and content?

.github/workflows/release-4-publish-release.yml
```diff
@@ -0,0 +1,359 @@
...
+        description: 'Dry run mode (check to enable, uncheck to perform actual operations)'
+        required: false
+        type: boolean
+        default: true
+      staging_repository_id:
+        description: 'Nexus staging repository ID to release (e.g., orgapachepolaris-1234)'
+        required: true
+        type: string
+
+jobs:
```
snazy, line 35: Mind adding a final job (maybe in a follow-up) to print the proposed `[ANNOUNCE]` and `[VOTE]` email subject and content?

releasey/libs/_version.sh
pingtimeout: I don't think we should do that.  The code change contains a `Fruits` variable that looks like a typo.  And also, the default sort for `git tag` is lexicographic, not numeric, hence the `| sort -n` part of the command.

.github/workflows/release-2-update-release-candidate.yml
pingtimeout: I believe it is already done.  I removed all intermediate `echo` in favor of the final `cat <<EOT >> $GITHUB_STEP_SUMMARY` so that we do not need to unfold each step to look at console output.

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: I am not certain that it would work.  The `grep` commands that extract the staging repo id assume that there is only one matching line.  Did you verify locally that with the `2>&1 | tee` part, this is still valid?

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: Oh that's a very good point.  Thanks for catching it!  I am not sure we should add a specific function just for that, as there is only one secret that requires masking: the SVN password.  The other secrets are not passed to `exec_process`.  So instead, what I can do is add a [`::add-mask::`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#masking-a-value-in-a-log) command as the first command of each step that uses the SVN password so that Github replaces the value by asterisks in the output.

.github/workflows/release-4-publish-release.yml
pingtimeout: I cannot find a target to write the announcement?  So I would say definitely in a follow up.

.github/workflows/release-1-create-release-branch.yml
pingtimeout: Done

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: Added a comment.  It might be worth noting that this pattern is also present (but uncommented) in `.github/workflows/helm.yml`.

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: I implemented the `::add-mask::` part but I don't think it was necessary.  [This run is a dry-run](https://github.com/pingtimeout/polaris/actions/runs/19171315668/job/54804409204) on my fork where I created fake secrets.  And although I only added `::add-mask::` for the svn password field, it seems that the username is also masked.

image

I suppose we should keep the add-mask directive for an extra layer of safety.  Marking this as resolved.

.github/workflows/release-4-publish-release.yml
snazy: I'm ok to do it in a follow-up.

Just throwing in arbitrary, not fleshed-out ideas here...
1. Use a GH issue to track release progress and let the workflow add a comment with the email suggestions there.
    Requires some GH API interaction - either REST (`gh api ...`) or [scripted](https://github.com/actions/github-script).
2. Emit the email suggestions as a GH step summary
3. Both?

.github/workflows/release-3-build-and-publish-artifacts.yml
snazy: Updated the following comment just to close this one.

.github/workflows/release-3-build-and-publish-artifacts.yml
snazy: Yes, it works.
I think it would be much better to have all the Gradle output in the workflow console, otherwise we're "blind" to what happened.
Might also be worth to add the `--info` Gradle argument, just in case we need some information for debugging purposes.

* Create `bar.sh`
  ```
  #!/usr/bin/env bash
  echo "TO STDOUT"
  echo "TO STDERR" > /dev/stderr
  ```
* and `foo.sh`
  ```
  #!/usr/bin/env bash
  function exec_process {
    print_command "Executing '${*}'"
    "$@"
  }
  exec_process ./bar.sh 2>&1 | tee logging
  ```
* `chmod 755 foo.sh bar.sh`
* `rm -f logging ; ./foo.sh ; echo "" ; echo "" ; cat logging`
* shows this on the console and in the log file
   ```
   ./foo.sh: line 3: print_command: command not found
   TO STDOUT
   TO STDERR
   ```
* without the `2>&1` all stderr output is lost

releasey/libs/_version.sh
```diff
@@ -0,0 +1,179 @@
...
+  local old_version="$1"
+  local new_version="$2"
+  exec_process sed -i~ "s/${old_version}/${new_version}/g" "$HELM_CHART_YAML_FILE"
+  exec_process sed -i~ "s/${old_version}/${new_version}/g" "$HELM_README_FILE"
+  # The readme file may contain version with double dash for shields.io badges
+  # We need a second `sed` command to ensure that the version replacement preserves this double-dash syntax.
+  local current_version_with_dash
+  local version_with_dash
+  current_version_with_dash="$(echo "${old_version//-/--}")"
+  version_with_dash="$(echo "$version//-/--}")"
```
snazy:
```suggestion
  current_version_with_dash="${old_version//-/--}"
  version_with_dash="${version//-/--}"
```
Bug fix (bracket for the 2nd line) + simplification

releasey/libs/_version.sh
```diff
@@ -0,0 +1,179 @@
...
+
+  return 0
+}
+
+function update_version {
+  local version="$1"
+  local current_version
+  current_version=$(cat "$VERSION_FILE")
+  update_version_txt "${version}"
+  update_helm_version "${current_version}" "${version}"
```
snazy: Hm, this doesn't work.
`version.txt` contains `1.3.0-incubating-SNAPSHOT` - hence `update_helm_version` can't replace the version pattern.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,374 @@
...
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+
+      - name: Create Helm package
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+
+          exec_process cd helm
+          exec_process helm package polaris
+          exec_process helm gpg sign polaris-${version_without_rc}.tgz
```
snazy: Could you cross-check this one?

It failed for me locally:
```
$ exec_process helm gpg sign polaris-2.3.0-incubating-SNAPSHOT.tgz 
DEBUG: Executing 'helm gpg sign polaris-2.3.0-incubating-SNAPSHOT.tgz'
Signing polaris-2.3.0-incubating-SNAPSHOT.tgz
tar: Pattern matching characters used in file names
tar: Use --wildcards to enable pattern matching, or --no-wildcards to suppress this warning
tar: */Chart.yaml: Not found in archive
tar: Exiting with failure status due to previous errors
Error: plugin "gpg" exited with error
```
The tarball contains these files (abbreviated list):
```
polaris/Chart.yaml
polaris/values.yaml
...
```

releasey/libs/_version.sh
```diff
@@ -0,0 +1,179 @@
...
+  exec_process sed -i~ "s/${old_version}/${new_version}/g" "$HELM_CHART_YAML_FILE"
+  exec_process sed -i~ "s/${old_version}/${new_version}/g" "$HELM_README_FILE"
+  # The readme file may contain version with double dash for shields.io badges
+  # We need a second `sed` command to ensure that the version replacement preserves this double-dash syntax.
+  local current_version_with_dash
+  local version_with_dash
+  current_version_with_dash="$(echo "${old_version//-/--}")"
+  version_with_dash="$(echo "$version//-/--}")"
+  exec_process sed -i~ "s/${current_version_with_dash}/${version_with_dash}/" "$HELM_README_FILE"
+}
```
snazy, line 119: (suggestion in the PR as well now)
```suggestion
function update_version {
  local version="$1"
  update_version_txt "${version}"
  update_helm_version "${version}"
}

function update_version_txt {
  local version="$1"
  # This function is only there for dry-run support.  Because of the
  # redirection, we cannot use exec_process with the exact command that will be
  # executed.
  if [[ ${DRY_RUN:-1} -ne 1 ]]; then
    exec_process echo "${version}" >$VERSION_FILE
  else
    exec_process "echo ${version} > $VERSION_FILE"
  fi
}

function update_helm_version {
  local new_version="$1"
  exec_process sed -E -i~ "s/^version: .+/version: ${new_version}/g" "$HELM_CHART_YAML_FILE"
  exec_process sed -E -i~ "s/^appVersion: .+/appVersion: ${new_version}/g" "$HELM_CHART_YAML_FILE"
  exec_process sed -E -i~ "s/[0-9]+[.][0-9]+([.][0-9]+)?(-incubating)-SNAPSHOT/${new_version}/g" "$HELM_README_FILE"
  # The readme file may contain version with double dash for shields.io badges
  # We need a second `sed` command to ensure that the version replacement preserves this double-dash syntax.
  local current_version_with_dash
  local version_with_dash
  current_version_with_dash="${old_version//-/--}"
  version_with_dash="${version//-/--}"
  exec_process sed -E -i~ "s/[0-9]+[.][0-9]+([.][0-9]+)?(--incubating)--SNAPSHOT/${version_with_dash}/g" "$HELM_README_FILE"
}
```

releasey/libs/_version.sh
pingtimeout: Ack.  With you change it works.  That being said, I think it is a problem of the release guide not asking for the versions in `helm/` to be bumped after a release.  Like, they are currently still `1.2.0-SNAPSHOT`, which has already been released final, so this is an error.

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: I wonder if it is linked to a `tar` version?  I do not have any warning on my end.

```
$ tar tzf polaris-99.98.97-incubating.tgz | head -n 5
polaris/Chart.yaml
polaris/values.yaml
polaris/templates/_helpers.tpl
polaris/templates/configmap.yaml
polaris/templates/deployment.yaml

$ helm gpg sign polaris-99.98.97-incubating.tgz
Signing polaris-99.98.97-incubating.tgz

$ echo $?
0
```

releasey/libs/_version.sh
pingtimeout: I believe this as been merged with your PR.  Marking this as resolved.  Let me know if that's not what you meant.

.github/workflows/release-4-publish-release.yml
pingtimeout: So, the first version of these workflows started by creating a Github draft release.  But IIRC JB explained that as per a GH release, the tag is already published and so are some artifacts.  And that is a problem as a vote has not yet happened.  In other words, a Github release can only be created once the vote succeeds.

snazy: I mentioned GitHub _issue_ not _release_ ;) ;)

snazy: But maybe better to align with what ATR does. IIRC they have something planned wrt voting.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -151,6 +151,8 @@ jobs:
           SVN_USERNAME: ${{ secrets.APACHE_USERNAME }}
           SVN_PASSWORD: ${{ secrets.APACHE_PASSWORD }}
         run: |
+          echo "::add-mask::$SVN_PASSWORD"
```
snazy, line 165: TIL!
Although GH should do this with all secrets out of the box. But it doesn't hurt.

.github/workflows/release-3-build-and-publish-artifacts.yml
snazy: I _think_ [this](https://github.com/snazy/helm-gpg/blob/2bf58051350b16508a52c6f10d36ae5b2550836d/.github/workflows/ci.yml#L63-L87) could work with `helm package --sign` and we can get rid of the helm-gpg-plugin, which is unmaintained for 5 years now.

snazy: @jbonofre could you provide the details how you generated generated the gpg key? Then we can do another round of testing simulating the whole roundtrip (key -> secret -> workflow -> get artifacts -> verify) to be sure that it works.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,380 @@
...
+
+      - name: Create Helm package
+        env:
+          GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+        run: |
+          echo "::add-mask::$GPG_PASSPHRASE"
+
+          source "${LIBS_DIR}/_exec.sh"
+
+          gpg --export-secret-keys > $GNUPGHOME/secring.gpg
```
snazy:
```suggestion
          # Make sure these files are always deleted
          trap "rm -f $GNUPGHOME/secring.gpg $GNUPGHOME/pubring.gpg $GNUPGHOME/passphrase" EXIT

          gpg --export-secret-keys > $GNUPGHOME/secring.gpg
```

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -0,0 +1,380 @@
...
+          exec_process helm package polaris --sign --key "." --keyring $GNUPGHOME/secring.gpg --passphrase-file $GNUPGHOME/passphrase
+          exec_process helm verify polaris-${version_without_rc}.tgz
+
+          calculate_sha512 polaris-${version_without_rc}.tgz
+          exec_process gpg --armor --output polaris-${version_without_rc}.tgz.asc --detach-sig polaris-${version_without_rc}.tgz
+          calculate_sha512 polaris-${version_without_rc}.tgz.prov
+          exec_process gpg --armor --output polaris-${version_without_rc}.tgz.prov.asc --detach-sig polaris-${version_without_rc}.tgz.prov
+
+          # Clean up temporary GPG directory
+          exec_process rm -f $GNUPGHOME/secring.gpg $GNUPGHOME/pubring.gpg $GNUPGHOME/passphrase
```
snazy:
```suggestion
```

snazy (approved):
This is really a big improvement!

The change now LGTM, so +1!

We just need to do a final safety cross-check with a dummy generated GPG key built in the same way as the "real" GPG key. But after that we should merge this PR.

Then we can test this (nearly) end-to-end and clean up the release-test artifacts. What we cannot test is the final release publication, because there is no CI for this use case but the good old try-n-error approach with some fingers-crossed.

pingtimeout:
@snazy There were a couple of things that needed fixed, like the fact that `svn` is not installed by default on Github Runners.  I pushed a few new commits.  Here are the links to workflow runs that passed:

* [Create release branch](https://github.com/pingtimeout/polaris/actions/runs/19459214595)
* [Update version and changelog for RC](https://github.com/pingtimeout/polaris/actions/runs/19459229571)
* [Build and publish release artifacts](https://github.com/pingtimeout/polaris/actions/runs/19462981258)

That last one correctly created a staging repository on Nexus (see screenshot below).  It also successfully pushed artifacts to dist dev.  I verified that those artifacts were in SVN and deleted them.  I will drop the staging repository next.

image

The only thing that is not tested is the last release workflow to publish artifacts after a release is successful.  We will have to wait for the official 1.3.0 release to test it given its real-life impacts.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -193,7 +193,7 @@ jobs:
           source "${LIBS_DIR}/_exec.sh"
 
           # Publish artifacts to staging repository
-          exec_process ./gradlew publishToApache closeApacheStagingRepository -Prelease -PuseGpgAgent --info > gradle_publish_output.txt 2>&1
+          exec_process ./gradlew publishToApache closeApacheStagingRepository -Prelease -PuseGpgAgent --info 2>&1 | tee gradle_publish_output.txt
```
snazy, line 196: Just FYI (no change needed/requested): an even shorter version of `2>&1` is `&>` (see [here](https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Redirecting-Standard-Output-and-Standard-Error)). I recently learned that as well.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -257,6 +257,12 @@ jobs:
           distribution: 'temurin'
           java-version: '21'
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
snazy: Not an [approved action](https://github.com/apache/infrastructure-actions/blob/main/approved_patterns.yml).
It it really needed? IIRC buildx is [already installed](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools).

snazy: I guess all that's needed is [this part](https://github.com/snazy/maxio-daily/blob/e717a87525e4d2ebe2e71245d6b2ed1cf386f5bc/src/buildx.sh#L29-L35), potentially adapted?
[This line](https://github.com/snazy/maxio-daily/blob/e717a87525e4d2ebe2e71245d6b2ed1cf386f5bc/src/buildx.sh#L39) might help w/ debugging.

snazy: (same for the 4th workflow)

pingtimeout: Fixed.  Thanks for the pointer !

.github/workflows/release-3-build-and-publish-artifacts.yml
pingtimeout: I am kind of -1 on this syntax because I personally _always_ forget it and have to try `&>` and `>&`.  So even though `2>&1` is longer, it is a very common pattern that is easy to recognize.

pingtimeout:
Run with the manual `docker buildx` setup: https://github.com/pingtimeout/polaris/actions/runs/19470515841, all good

.github/workflows/release-3-build-and-publish-artifacts.yml
snazy: Yea, haven't seen that much in the wild either.

.github/workflows/release-2-update-release-candidate.yml
snazy: #3080

PR #2388, opened 2025-08-19
Remove numeric identifier from PolarisPrincipal

adutra:
This change removes the requirement for Polaris principals to have a numeric identifier, by removing the only sites where such an identifier was required: 

- In the `Resolver`. Instead of lookups by id, the `Resolver` now performs lookups by principal name.
- In  `PolarisAdminService`. Instead of comparing entity ids, the code now compares the principal name against the entity name and type.

Note: the lookup in the `Resolver` is still necessary, because the `Resolver` also needs to fetch the grant records.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
@@ -251,11 +251,8 @@ public class PolarisAdminService {
     PolarisResolvedPathWrapper topLevelEntityWrapper =
         resolutionManifest.getResolvedTopLevelEntity(topLevelEntityName, entityType);
 
-    // TODO: If we do add more "self" privilege operations for PRINCIPAL targets this should
-    // be extracted into an EnumSet and/or pushed down into PolarisAuthorizer.
-    if (topLevelEntityWrapper.getResolvedLeafEntity().getEntity().getId()
-            == polarisPrincipal.getId()
-        && (op.equals(PolarisAuthorizableOperation.ROTATE_CREDENTIALS))) {
+    PolarisEntity entity = topLevelEntityWrapper.getResolvedLeafEntity().getEntity();
+    if (isSelfEntity(entity) && isSelfOperation(op)) {
       LOGGER
           .atDebug()
           .addKeyValue("principalName", topLevelEntityName)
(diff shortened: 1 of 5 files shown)


snazy (approved):

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -305,6 +301,30 @@ private void authorizeBasicTopLevelEntityOperationOrThrow(
...
+  /**
+   * Returns true if the operation is a "self" operation, that is, an operation that is being
+   * performed by the principal on itself.
+   *
+   * <p>TODO: If we do add more "self" privilege operations for PRINCIPAL targets this should be
+   * extracted into an EnumSet and/or pushed down into PolarisAuthorizer.
+   */
+  private static boolean isSelfOperation(PolarisAuthorizableOperation op) {
+    return op.equals(PolarisAuthorizableOperation.ROTATE_CREDENTIALS)
+        || op.equals(PolarisAuthorizableOperation.RESET_CREDENTIALS);
```
dimas-b: side note: `RESET_CREDENTIALS` is no going to be a "self" operation after #2197

dimas-b (dismissed):
LGTM :+1:

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
adutra: My understanding of #2197 is that the root user will be able to reset credentials of another user, but it won't remove the possibility for any user to reset their own credentials. So in a way it it still a "self" operation. Am I missing something?

dimas-b: IIRC, in current `main` the `RESET_CREDENTIALS` operation is "dead code" (but will be restored to active service by #2197).

dimas-b: I do not think a self-reset of credentials will be allowed, though.

adutra: So, should I remove this check? IMO any changes here should be made by #2197 itself, but I don't mind proactively changing this bit in this PR.

adutra: I see the change has been made in #2197: only `ROTATE_CREDENTIALS` remains a "self-operation".

adutra:
@dennishuo @collado-mike PTAL.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: I agree that this change should be made under #2197. I commented here only as a "heads up" hoping to clarify potential merge conflicts. Sorry, if I caused confusion :sweat_smile:

PR #2389, opened 2025-08-19
Include principal name in Polaris tokens

adutra:
Summary of changes:

- Instead of including the principal id twice in the token, the principal name is now used as the subject claim. While the default authenticator doesn't need the principal name and works with just the principal id, not having the "real" principal name available could be a problem for other authenticator implementations.

- `DecodedToken` has been refactored and renamed to `InternalPolarisCredential`. It is also now a package-private component as it is an implementation detail of Polaris internal auth mechanism.

- `TokenBroker.verify()` now returns `PolarisCredential`.


runtime/service/src/test/java/org/apache/polaris/service/auth/DefaultAuthenticatorTest.java
```diff
@@ -73,10 +73,9 @@ public void testFetchPrincipalThrowsServiceExceptionOnMetastoreException() {
 
   @Test
   public void testFetchPrincipalThrowsNotAuthorizedWhenNotFound() {
-    DecodedToken token = Mockito.mock(DecodedToken.class);
+    PolarisCredential token = Mockito.mock(PolarisCredential.class);
     long principalId = 100L;
     when(token.getPrincipalId()).thenReturn(principalId);
-    when(token.getClientId()).thenReturn("abc");
```
singhpk234, line 75: is this intentional or was just redundant ?

runtime/service/src/main/java/org/apache/polaris/service/auth/InternalPolarisCredential.java
```diff
@@ -0,0 +1,72 @@
...
+ * A specialized {@link PolarisCredential} used for internal authentication, when Polaris is the
+ * identity provider.
+ *
+ * <p>Such credentials are created by the Polaris service itself, from a JWT token previously issued
+ * by Polaris itself.
+ *
+ * @see JWTBroker
+ */
+@PolarisImmutable
+abstract class InternalPolarisCredential implements PolarisCredential {
```
singhpk234: [optional : may be another name ?] isn't it still a decoded bearer token ? maybe its just me but i find it a little bit hard to disassociate with the client crendentials flow semantics, which typically requires client secret

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
```diff
@@ -157,16 +143,18 @@ public TokenResponse generateFromClientSecrets(
...
   }
 
-  private String generateTokenString(String clientId, String scope, Long principalId) {
+  private String generateTokenString(
+      String principalName, long principalId, String clientId, String scope) {
     Instant now = Instant.now();
     return JWT.create()
         .withIssuer(ISSUER_KEY)
-        .withSubject(String.valueOf(principalId))
+        .withSubject(principalName)
```
singhpk234, line 156: [for my understanding] we don't use principalId for lookups from metastore from sub and always use the prinicipal id from the claim ? if yes how do we plan to leverage principalName from the subject here ?

runtime/service/src/test/java/org/apache/polaris/service/auth/DefaultAuthenticatorTest.java
adutra: Redundant :-)

runtime/service/src/main/java/org/apache/polaris/service/auth/InternalPolarisCredential.java
adutra: I'm not opposed to other names.

My rationale is: this component represents credentials extracted from an internal polaris token, hence the name I picked.

Let's wait for more feedback from other reviewers and I'll change the name accordingly.

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
adutra: > we don't use principalId for lookups from metastore from sub and always use the prinicipal id from the claim ? 

Correct.

> how do we plan to leverage principalName from the subject here ?

The `DefaultAuthenticator` only does lookups by name if the principal id is not available. This won't happen with polaris internal tokens, of course. But other `Authenticator` implementations could require the principal name to be available. It's imho bad that the principal name is never surfaced in internal polaris tokens.

runtime/service/src/main/java/org/apache/polaris/service/auth/InternalPolarisCredential.java
adutra: Would `PolarisToken` work for you?

adutra: (or `InternalPolarisToken`)

singhpk234: > InternalPolarisToken

works for me ! This comment is totally optional to address, agree for waiting for more peoplr to  weigh in :) !

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
singhpk234: > But other Authenticator implementations could require the principal name to be available

i would like this understand this use case more, is it more from audit purpose ? or are we saying some Token Brokers would prefer look-up by name as names would be like `email-id` for them which is considered to unique ?

adutra: Yes, exactly. Some `Authenticator`s could enforce by-name lookup since principal names must be unique.

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
```diff
@@ -60,34 +58,22 @@ public abstract class JWTBroker implements TokenBroker {
...
-        @Override
-        public String getScope() {
-          return decodedJWT.getClaim("scope").asString();
-        }
-      };
-
-    } catch (JWTVerificationException e) {
+      return InternalPolarisToken.of(
+          decodedJWT.getSubject(),
+          decodedJWT.getClaim(CLAIM_KEY_PRINCIPAL_ID).asLong(),
```
dimas-b, line 72: I might have missed it, but where is this claim set (in issued tokens)?

adutra: In the `generateTokenString` method:

https://github.com/apache/polaris/blob/95ebdd359ed0956ddc7a771c37683f9a3ecb0176/runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java#L164-L177

dimas-b: Ah, it was not part of the diff in this PR :facepalm:

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
```diff
@@ -156,16 +142,18 @@ public TokenResponse generateFromClientSecrets(
...
     return new TokenResponse(
         tokenString, TokenType.ACCESS_TOKEN.getValue(), maxTokenGenerationInSeconds);
   }
 
-  private String generateTokenString(String clientId, String scope, Long principalId) {
+  private String generateTokenString(
+      String principalName, long principalId, String clientId, String scope) {
     Instant now = Instant.now();
     return JWT.create()
         .withIssuer(ISSUER_KEY)
```
dimas-b, line 155: side note: it might be worth setting `aud` to the realm ID and later validating it (for follow up). WDYT?

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
```diff
@@ -60,34 +58,22 @@ public abstract class JWTBroker implements TokenBroker {
...
-
-        @Override
-        public String getScope() {
-          return decodedJWT.getClaim("scope").asString();
-        }
-      };
-
-    } catch (JWTVerificationException e) {
+      return InternalPolarisToken.of(
+          decodedJWT.getSubject(),
```
dimas-b, line 71: I guess this might have issues during rolling upgrades as old tokens might still have `sub` == principal ID.

I do not think it is a critical issue, but might be worth mentioning in `CHANGELOG.md`. WDYT?

adutra: I think we are fine: the `DefaultAuthenticator` inspects `PolarisCredential.getPrincipalId()` and `PolarisCredential.getPrincipalName()`, and these fields are mapped as follows:

<table>
<tr><th></th><th>Source claim before this PR</th><th>Source claim after this PR</th></tr>
<tr><td>getPrincipalId()</td><td>principal_id</td><td>principal_id</td></tr>
<tr><td>getPrincipalName()</td><td>(none)</td><td>sub</td></tr>
</table>

adutra: So, if a token generated by an old `JWTBroker` is processed by a new `JWTBroker`, `getPrincipalName()` would return the string representation of the principal ID which would result in a failed lookup, but `getPrincipalId()` would still return the principal ID in numeric form, and that lookup would succeed.

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
adutra: We could definitely include the realm in the token, but I'm not sure about using the `aud` claim. In all compliance with the specs, this claim should contain the URI of the Polaris service:

https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3

But, it could also contain a simple string as well, and the interpretation of that value is application-specific.

Maybe a separate claim e.g. `polaris_realm` would be better?

runtime/service/src/main/java/org/apache/polaris/service/auth/JWTBroker.java
dimas-b: Good point! Thanks for the clarification :+1:

PR #2394, opened 2025-08-19
chore: Suppress javac deprecation warnings in SparkCatalog

dimas-b:
SparkCatalog intentionally overrides and uses deprecated methods from Spark's TableCatalog.

This PR adds suppression annotations to allow for clean compilation given that the deprecated method calls and overrides are clearly expected in this case.

### DIFF
--- a/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
+++ b/plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
@@ -145,6 +145,7 @@ public class SparkCatalog
   }
 
   @Override
+  @SuppressWarnings({"deprecation"})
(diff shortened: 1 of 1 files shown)


flyrain (approved):

adutra (approved):
Yeah it seems Spark delegates the new method to the old one, forcing implementors to implement the latter rather than the former:

In 3.5.6:

https://github.com/apache/spark/blob/v3.5.6/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/TableCatalog.java#L197-L202

In 4.0.0:

https://github.com/apache/spark/blob/v4.0.0/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/TableCatalog.java#L202-L209

🤷‍♂️ 😅

PR #2401, opened 2025-08-19
feat: enforce LIST_PAGINATION_ENABLED

dimas-b:
The enforcement of the LIST_PAGINATION_ENABLED flag was missed in #1938. This change make the flag effective as discussed in #2296.

Note: this causes a change in the default Polaris behaviour (no pagination by default) with respect to the previous state of `main`. However, there is no behaviour change with respect to 1.0.0 or 1.0.1 as previous releases did not have #1938.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
@@ -24,7 +24,7 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import jakarta.annotation.Nullable;
 import java.util.Optional;
 import java.util.OptionalInt;
-import java.util.function.Supplier;
+import java.util.function.BooleanSupplier;
 import org.apache.polaris.immutables.PolarisImmutable;
 
 /** A wrapper for pagination information passed in as part of a request. */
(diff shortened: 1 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/PageToken.java
```diff
@@ -86,7 +87,10 @@ static PageToken fromLimit(int limit) {
    * @see Page#encodedResponseToken()
    */
   static PageToken build(
-      @Nullable String serializedPageToken, @Nullable Integer requestedPageSize) {
-    return PageTokenUtil.decodePageRequest(serializedPageToken, requestedPageSize);
+      @Nullable String serializedPageToken,
+      @Nullable Integer requestedPageSize,
+      Supplier<Boolean> shouldDecodeToken) {
```
snazy: Why is this a `Supplier`? None of the call sites is really "dynamic".

dimas-b: It is not called when the token is `null`, so it avoids touching the config look up in most cases... I'm fine with an unconditional config lookup too. What is preferable from your POV?

snazy: Hm, neither `getResolvedCatalogEntity` nor `getConfig` are really simple.
Mind making this at least a `Predicate`?

dimas-b: `Predicate` is a one-argument function, but here we just need to have a deferred getter for a simple `boolean` :thinking: 

Updated to `BooleanSupplier`. WDYT?

PR #2390, opened 2025-08-19
Remove ActiveRolesProvider

adutra:
Summary of changes:

- As proposed on the ML, `ActiveRolesProvider` is removed, and `DefaultActiveRolesProvider` is merged into `DefaultAuthenticator`. `ActiveRolesAugmentor` is also merged into `AuthenticatingAugmentor`.

- The implicit convention that "no roles in credentials" == "all roles requested" is removed, as it is ambiguous. From now on, credentials must explicitly include the `PRINCIPAL_ROLE:ALL` pseudo-role in order to request all roles.

- `PersistedPolarisPrincipal` is removed. It existed merely as a means of passing the `PrincipalEntity` from the authenticator to the roles provider. This is not necessary anymore.

PR #2404, opened 2025-08-20
Deprecate ActiveRolesProvider for removal

adutra:
This change deprecates `ActiveRolesProvider` for removal. The actual removal is here: #2390.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,9 +55,8 @@ at locations that better optimize for object storage.
 
 ### Deprecations
 
-* The property `polaris.active-roles-provider.type` is deprecated in favor of
-  `polaris.authentication.active-roles-provider.type`. The old property is still supported, but will be removed in a
-  future release.
+- The property `polaris.active-roles-provider.type` is deprecated for removal.
+- The `ActiveRolesProvider` interface is deprecated for removal.
 
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


snazy (approved):

CHANGELOG.md
```diff
@@ -55,9 +55,8 @@ at locations that better optimize for object storage.
 
 ### Deprecations
 
-* The property `polaris.active-roles-provider.type` is deprecated in favor of
-  `polaris.authentication.active-roles-provider.type`. The old property is still supported, but will be removed in a
```
pingtimeout, line 59: This line has been removed, but it seems to me that it should be moved to the upgrade notes?  I.e. whenever something is deprecated and a path forward is provided, this information should be present in the changelog.  Wdyt?

adutra: In fact, #2000 was the root cause for the initial deprecation warning. 

But that PR is not released yet; and the new direction regarding `ActiveRolesProvider` is that it should be removed, instead of made realm-aware.

We _could_ revert #2000. But at this point, I'm not sure it's worth the hassle.

Instead, I'm just proposing to _not_ advertise the property naming change, in the hopes that users won't need to care about it at all. 

Does that make sense?

pingtimeout: Yes it does, thanks for the clarification

dennishuo:
LGTM!

PR #2405, opened 2025-08-20
Fix createViewWithCustomMetadataLocation tests for cloudTest task

tmater:
This patch fixes the `createViewWithCustomMetadataLocation` tests for `cloudTest` tasks. The original test was generating temp directories internally, causing `cloudTest`s to fail with `BadRequestException` instead of the expected `ForbiddenException`.

Changes:
- Switched to Hadoop's Path (Java Path removes slashes, e.g. s3://bucket/path -> s3:/bucket/path)
- Made base classes `abstract` to avoid running them
- Implemented `createViewWithCustomMetadataLocation` to allow passing a custom location

Testing:
- Verified locally

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
@@ -18,9 +18,8 @@
  */
 package org.apache.polaris.service.it.test;
 
-import com.google.common.base.Strings;
+import java.io.File;
 import java.util.List;
-import org.apache.hadoop.fs.Path;
 import org.apache.polaris.core.admin.model.AzureStorageConfigInfo;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
 import org.apache.polaris.core.storage.StorageUtil;
@@ -38,12 +37,12 @@ public abstract class PolarisRestCatalogViewAdlsIntegrationTestBase
         .setStorageType(StorageConfigInfo.StorageTypeEnum.AZURE)
         .setAllowedLocations(
             List.of(
-                StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_SUBDIR, Path.SEPARATOR)))
+                StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_SUBDIR, File.separator)))
         .build();
   }
 
   @Override
   protected String getCustomMetadataLocationDir() {
-    return new Path(BASE_LOCATION, POLARIS_IT_CUSTOM_SUBDIR).toString();
+    return StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_SUBDIR, File.separator);
   }
 }
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
@@ -18,26 +18,32 @@
  */
 package org.apache.polaris.service.it.test;
 
+import com.google.common.base.Strings;
 import java.util.List;
-import java.util.Optional;
-import org.apache.polaris.core.admin.model.AwsStorageConfigInfo;
+import org.apache.hadoop.fs.Path;
+import org.apache.polaris.core.admin.model.AzureStorageConfigInfo;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
+import org.apache.polaris.core.storage.StorageUtil;
 
-/** Runs PolarisRestCatalogIntegrationBase test on AWS. */
-public abstract class PolarisRestCatalogAwsIntegrationTestBase
-    extends PolarisRestCatalogIntegrationBase {
-  public static final String ROLE_ARN =
-      Optional.ofNullable(System.getenv("INTEGRATION_TEST_ROLE_ARN"))
-          .or(() -> Optional.ofNullable(System.getenv("INTEGRATION_TEST_S3_ROLE_ARN")))
-          .orElse("arn:aws:iam::123456789012:role/my-role");
-  public static final String BASE_LOCATION = System.getenv("INTEGRATION_TEST_S3_PATH");
+/** Runs PolarisRestCatalogViewIntegrationTest on Azure. */
+public abstract class PolarisRestCatalogViewAdlsIntegrationTestBase
+    extends PolarisRestCatalogViewIntegrationBase {
+  public static final String TENANT_ID = System.getenv("INTEGRATION_TEST_ADLS_TENANT_ID");
+  public static final String BASE_LOCATION = System.getenv("INTEGRATION_TEST_ADLS_PATH");
 
   @Override
   protected StorageConfigInfo getStorageConfigInfo() {
-    return AwsStorageConfigInfo.builder()
-        .setRoleArn(ROLE_ARN)
-        .setStorageType(StorageConfigInfo.StorageTypeEnum.S3)
-        .setAllowedLocations(List.of(BASE_LOCATION))
+    return AzureStorageConfigInfo.builder()
+        .setTenantId(TENANT_ID)
+        .setStorageType(StorageConfigInfo.StorageTypeEnum.AZURE)
+        .setAllowedLocations(
+            List.of(
+                StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_SUBDIR, Path.SEPARATOR)))
         .build();
+  }
+
+  @Override
+  protected String getCustomMetadataLocationDir() {
+    return new Path(BASE_LOCATION, POLARIS_IT_CUSTOM_SUBDIR).toString();
   }
 }
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
@@ -20,27 +20,30 @@ package org.apache.polaris.service.it.test;
 
 import com.google.common.base.Strings;
 import java.util.List;
-import java.util.stream.Stream;
-import org.apache.polaris.core.admin.model.AzureStorageConfigInfo;
+import java.util.Optional;
+import org.apache.hadoop.fs.Path;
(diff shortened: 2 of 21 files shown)


integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
```diff
@@ -20,13 +20,12 @@
...
 import java.util.Optional;
-import java.util.stream.Stream;
+import org.apache.hadoop.fs.Path;
 import org.apache.polaris.core.admin.model.AwsStorageConfigInfo;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
-import org.assertj.core.util.Strings;
 
 /** Runs PolarisRestCatalogViewIntegrationTest on AWS. */
-public class PolarisRestCatalogViewAwsIntegrationTest
+public abstract class PolarisRestCatalogViewAwsIntegrationTestBase
```
snazy: Since we're renaming this type anyway, let's better use the protocol name, not the vendor name.

```suggestion
public abstract class PolarisRestCatalogViewS3IntegrationTestBase
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
```diff
@@ -20,13 +20,12 @@
 
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Stream;
+import org.apache.hadoop.fs.Path;
```
snazy: Hm, sure, this is just a test, but could we avoid Hadoop dependencies?

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewGcpIntegrationTestBase.java
```diff
@@ -19,13 +19,12 @@
...
 import java.util.List;
-import java.util.stream.Stream;
+import org.apache.hadoop.fs.Path;
 import org.apache.polaris.core.admin.model.GcpStorageConfigInfo;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
-import org.assertj.core.util.Strings;
 
 /** Runs PolarisRestCatalogViewIntegrationTest on GCP. */
-public class PolarisRestCatalogViewGcpIntegrationTest
+public abstract class PolarisRestCatalogViewGcpIntegrationTestBase
```
snazy: Similar here
```suggestion
public abstract class PolarisRestCatalogViewGcsIntegrationTestBase
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAzureIntegrationTestBase.java
```diff
@@ -19,13 +19,12 @@
...
 import java.util.List;
-import java.util.stream.Stream;
+import org.apache.hadoop.fs.Path;
 import org.apache.polaris.core.admin.model.AzureStorageConfigInfo;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
-import org.assertj.core.util.Strings;
 
 /** Runs PolarisRestCatalogViewIntegrationTest on Azure. */
-public class PolarisRestCatalogViewAzureIntegrationTest
+public abstract class PolarisRestCatalogViewAzureIntegrationTestBase
```
snazy: Similar
```suggestion
public abstract class PolarisRestCatalogViewAdlsIntegrationTestBase
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
tmater: Shall I change these 3 (non-view test bases) as well in one go:
- PolarisRestCatalogAwsIntegrationTestBase
- PolarisRestCatalogAzureIntegrationTestBase
- PolarisRestCatalogGcpIntegrationTestBase

Also the 6 implementations:
- RestCatalogAwsIT
- RestCatalogAzureIT
- RestCatalogGcpIT
- RestCatalogViewAwsIT
- RestCatalogViewAzureIT
- RestCatalogViewGcpIT

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
tmater: The idea behind Hadoop Path came because Iceberg does something "funky" with the `metadataFileLocation()` that resembles Hadoop Path behavior: It normalizes the scheme `file:///...` to `file:/...`, but it keeps `s3://...` as `s3://...`. 
For the assertion later I couldn't use `java.nio.Path` because that normalizes the scheme and creates an incorrect `s3:/...` path.

Just dug a bit more, I found a reference to Hadoop Path in Iceberg's [LocationProviders:24](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/LocationProviders.java#L24), maybe this is how they do it as well, I could not find the exact location where it gets normalized.

I’m open to other approaches, but the only solution that came to mind was writing a custom assertion, which I’m also happy with.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
snazy: I'd support it, yea

snazy: Mean, we support MinIO and a bunch of other S3 compatible systems, not only AWS. Sure, we could argue about GCS or ADLS. However, GCS and ADLS are specific, where a vendor name or "product suite" name are not.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
snazy: Argh - the damn `file` URI scheme is so ambiguous (IMHO, because of [multiple representations](https://en.wikipedia.org/wiki/File_URI_scheme) for the same thing, legacy representations, risk of interpreting those wrong).

If it's just about eliminating all host related parts, I'd say let's just use a regex to "fix" the paths?
Something like
```java
protected static final Pattern FILE_LOCATION_PATTERN =
  Pattern.compile("file:/*(.*)");

protected String fixFileUri(String location) {
  var m = FILE_LOCATION_PATTERN.matcher(location);
  return m.matches() ? "file:/" + m.group(1) : location;
}
```

WDYT? Would that work?

adutra: Just FYI In Polaris we have:

- `org.apache.polaris.core.storage.StorageLocation` which standardizes file URIs to `"file:///`
- but `InMemoryStorageIntegration` does the opposite 😅 :

https://github.com/apache/polaris/blob/532ee5137d2c41774983d075ea31dcbd2661aa1d/polaris-core/src/main/java/org/apache/polaris/core/storage/InMemoryStorageIntegration.java#L74

snazy: Hm - maybe let's ignore the `file` scheme specialty and just assume that it's "correct" if supplied.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
tmater: Makes sense, renamed the classes and related parts.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAwsIntegrationTestBase.java
tmater: Good point, I used `java.nio.file.Path` to remove the scheme.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewIntegrationBase.java
```diff
@@ -188,33 +182,89 @@ protected boolean overridesRequestedLocation() {
     return true;
   }
 
-  @Override
+  protected String getCustomMetadataLocationDir() {
+    return "";
+  }
```
adutra, line 187:
```suggestion
  protected abstract String getCustomMetadataLocationDir();
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewIntegrationBase.java
```diff
@@ -86,6 +86,8 @@ public abstract class PolarisRestCatalogViewIntegrationBase extends ViewCatalogT
   private static PolarisApiEndpoints endpoints;
   private static PolarisClient client;
   private static ManagementApi managementApi;
+  protected static final String POLARIS_IT_SUBDIR = "polaris_it";
+  protected static final String POLARIS_IT_CUSTOM_SUBDIR = "polaris_it_custom";
```
adutra, line 90: This constant seems unused.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
```diff
@@ -18,29 +18,31 @@
...
+            List.of(
+                StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_SUBDIR, File.separator)))
         .build();
   }
 
   @Override
-  protected boolean shouldSkip() {
-    return Stream.of(BASE_LOCATION, TENANT_ID).anyMatch(Strings::isNullOrEmpty);
+  protected String getCustomMetadataLocationDir() {
+    return StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_SUBDIR, File.separator);
```
adutra, line 46: Shouldn't this be:

```suggestion
    return StorageUtil.concatFilePrefixes(BASE_LOCATION, POLARIS_IT_CUSTOM_SUBDIR, File.separator);
```

PR #2412, opened 2025-08-20
chore: fix Page javadoc

dimas-b:
Correct method args (follow-up to #2401)

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
@@ -24,6 +24,7 @@ import jakarta.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
```diff
@@ -24,6 +24,7 @@
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.function.BooleanSupplier;
```
jbonofre, line 27: Do we really need to import `BooleanSupplier` here as it's not really used ?

jbonofre (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/Page.java
dimas-b: It's referenced from javadoc :)

dimas-b: I believe using imports in javadoc is standard in Polaris... at least my IDE does that automagically :)

jbonofre: Yes, I see that, but it looks weird to import just for a javadoc 😄

dimas-b: Make the javadoc text more concise. I kind of like it personally :)

dimas-b: In most cases imports already exist for real code references, but here the method is mentioned in javadoc, but not called from code :man_shrugging:

PR #789, opened 2025-01-15
Allow custom access token generation in integration test env

dimas-b:
Polaris allows running the integration test suite against an external (managed) server.

This change further enables customizing access tokens for these tests according to the test env. needs.

PR #2413, opened 2025-08-20
chore(test): Restore PolarisAccessManager

dimas-b:
Restore `PolarisAccessManager` (with default implementation of `IcebergTokenAccessManager`) which was added as an extension point for running Polaris tests in downstream build environments under #789, but was mistakenly removed in #2343


snazy (approved):
Confirmed that this change only brings back functionality that's been removed by #2343.

PR #2420, opened 2025-08-21
Fix merge conflict in CatalogFederationIntegrationTest

eric-maynard:
#2344 added a new test for catalog federation, but it looks like an undetected conflict with concurrent changes related to authentication have broken the test in main.


sfc-gh-ygu (approved):
Thanks for the fix! This is indeed a 1.1.0 blocker.

flyrain (approved):

Issue #2417, opened 2025-08-21
Namespaces can be created outside the parent catalog's allowedLocations


eric-maynard:
### Describe the bug

The default table location overlap check relies on the idea that the logical hierarchy of tables & namespaces within a catalog is aligned with the physical arrangement of those tables & namespaces within one of the catalog's allowed locations. However, it's currently possible to create a namespace outside of the parent catalog, which breaks not only this assumption but also the fundamental constraint of a catalog having a list of allowed locations at which its tables may be found.

### To Reproduce

Create a catalog, such as with:
```bash
./polaris \
    catalogs \
    create quickstart_catalog \
    --storage-type file \
    --default-base-location 'file:///tmp/quickstart'
```

Create a namespace:
```bash
./polaris \
    namespaces \
    create \
    --catalog quickstart_catalog \
    --location 'file:///tmp/ns' \
    ns
```

PR #2422, opened 2025-08-21
Do not allow namespaces outside of allowed locations

eric-maynard:
When we create a namespace or alter its location, we must confirm that this location is within the parent location -- whether that's a catalog or a namespace -- in order for our overlap checks and assumptions around credential vending to be valid. This PR introduces this behavior and enables it by default behind a new feature flag `ALLOW_NAMESPACE_LOCATION_ESCAPE`

Fixes: #2417


polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -100,6 +100,16 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> ALLOW_NAMESPACE_LOCATION_ESCAPE =
+      PolarisConfiguration.<Boolean>builder()
+          .key("ALLOW_NAMESPACE_LOCATION_ESCAPE")
```
dimas-b: Feature flags [constitute](https://polaris.apache.org/in-dev/unreleased/evolution/) our SemVer "API" surface, so I believe we need to formally send a discussion email to the `dev` ML about this flag.

dimas-b: Proactively +1 from my side :)

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisApplicationIntegrationTest.java
```diff
@@ -641,4 +666,53 @@ public void testRequestBodyTooLarge() throws Exception {
...
+            .isInstanceOf(BadRequestException.class)
+            .hasMessageContaining("location");
+      } else {
+        assertThatCode(createBadNamespace).doesNotThrowAnyException();
+      }
+      ThrowableAssert.ThrowingCallable createBadChildGoodParent =
+          () ->
+              sessionCatalog.createNamespace(
+                  sessionContext,
+                  Namespace.of("good_namespace", "bad_child"),
```
dimas-b, line 707: Could you add a test where one of the parent namespaces has a location property set too (not just at the catalog level)?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1083,6 +1093,65 @@ private void validateNoLocationOverlap(
...
+      LOGGER.debug(
+          "Validating namespace {} given parent catalog {}",
+          namespace.getName(),
+          parentEntity.getName());
+      var storageConfigInfo = parentEntity.getStorageConfigurationInfo();
+      if (storageConfigInfo == null) {
+        throw new IllegalArgumentException(
+            "Cannot create namespace without a parent storage configuration");
+      }
+      System.out.println("#### Validating " + namespace.getName());
```
dimas-b: oops, `System.out`?

dimas-b (commented):
LGTM overall :+1:

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
eric-maynard: Is that so? That seems to create a weird incentive not to feature-flag anything.

In this particular case there's a behavior change due to the default setting of the flag so in that sense it's relevant to semver, but it's also a bugfix.

eric-maynard:
Pulling this back to draft to investigate -- we may need to just block all custom namespace locations like we do for tables in order to avoid scenarios where two namespaces overlap despite being within their parent entity

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: I'm +1 on this change already. So having extra discussion on the `dev` ML does not matter for me personally.

@flyrain WDYT?

flyrain: +1 for the change as well. If this is more a bug fix, I don't think we even need this flag, the only reason we need the config is that existing behavior can be broken due to the change.

flyrain: We could even mark it `deprecated` in the beginning(this PR).

dimas-b: @flyrain : but what's you POV about requiring `dev` discussion for feature flags in general?

flyrain: I'm supportive on having a discussion on dev mailing list. It's great to let more people know this behavior change/bug fix.

eric-maynard:
I've updated the PR to make this check more like the table one, which totally disallows custom locations by default.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1083,6 +1093,74 @@ private void validateNoLocationOverlap(
...
+          "Validating namespace {} given parent catalog {}",
+          namespace.getName(),
+          parentEntity.getName());
+      var storageConfigInfo = parentEntity.getStorageConfigurationInfo();
+      if (storageConfigInfo == null) {
+        throw new IllegalArgumentException(
+            "Cannot create namespace without a parent storage configuration");
+      }
+      boolean allowed =
+          parentEntity.getStorageConfigurationInfo().getAllowedLocations().stream()
```
dennishuo, line 1122: AllowedLocations in the StorageConfig are an independent concept vs either of:

1. Whether to allow "custom" locations outside of the default path hierarchy that uses namespace names
2. Whether to allow "overlapping" locations

The most important one is validating that anything that can vend credentials resides under the allowedLocations.

Overlap and custom hierarchies in general are more common to disable validation for.

eric-maynard: This PR isn't related to credential vending, but the location that a namespace is created at.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
eric-maynard: In this case I think we want to have a flag for parity with tables -- the "bug" here IMO is that the default behavior is too lax, not that the lax behavior is accessible at all.

I'll create a quick thread to discuss the change in more detail, but in general my position is that it's not healthy for a project to require ML discussion for every new config. This creates an incentive to not protect changes with configs, and in the ideal world where all changes are config-protected this essentially necessitates an ML thread for every PR.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -508,6 +508,11 @@ private void createNamespaceInternal(
     } else {
       LOGGER.debug("Skipping location overlap validation for namespace '{}'", namespace);
     }
+    if (!realmConfig.getConfig(
+        FeatureConfiguration.ALLOW_NAMESPACE_CUSTOM_LOCATION, catalogEntity)) {
+      LOGGER.debug("Validating that namespace {} has a location inside its parent", namespace);
```
flyrain: Minor: is it necessary to have another debugging message here given line 1106 provides a similar one?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1083,6 +1093,74 @@ private void validateNoLocationOverlap(
...
+            "Cannot create namespace without a parent storage configuration");
+      }
+      List<StorageLocation> defaultLocations =
+          parentEntity.getStorageConfigurationInfo().getAllowedLocations().stream()
+              .filter(java.util.Objects::nonNull)
+              .map(
+                  l -> {
+                    return StorageLocation.ensureTrailingSlash(
+                        StorageLocation.ensureTrailingSlash(l) + namespace.getName());
+                  })
```
flyrain: minor: 
```suggestion
                  l -> 
                    StorageLocation.ensureTrailingSlash(
                        StorageLocation.ensureTrailingSlash(l) + namespace.getName())
                  )
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -671,6 +676,11 @@ public boolean setProperties(Namespace namespace, Map<String, String> properties
     } else {
       LOGGER.debug("Skipping location overlap validation for namespace '{}'", namespace);
     }
+    if (!realmConfig.getConfig(
+        FeatureConfiguration.ALLOW_NAMESPACE_CUSTOM_LOCATION, catalogEntity)) {
+      LOGGER.debug("Validating that namespace {} has a location inside its parent", namespace);
```
flyrain: Same here

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -508,6 +508,11 @@ private void createNamespaceInternal(
     } else {
       LOGGER.debug("Skipping location overlap validation for namespace '{}'", namespace);
     }
+    if (!realmConfig.getConfig(
+        FeatureConfiguration.ALLOW_NAMESPACE_CUSTOM_LOCATION, catalogEntity)) {
```
flyrain: Why do we validate the namespace location when `ALLOW_NAMESPACE_CUSTOM_LOCATION` is false?  I think we should validate it when we allow customer location.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -508,6 +508,11 @@ private void createNamespaceInternal(
...
+      LOGGER.debug("Validating that namespace {} has a location inside its parent", namespace);
+      validateNamespaceLocation(entity, resolvedParent);
+    }
```
flyrain, line 514: Minor: It'd be nice to wrap this into a method, so that we don't duplicate them in two different places.
```
    if (!realmConfig.getConfig(FeatureConfiguration.ALLOW_NAMESPACE_LOCATION_OVERLAP)) {
      LOGGER.debug("Validating no overlap for {} with sibling tables or namespaces", namespace);
      validateNoLocationOverlap(entity, resolvedParent.getRawFullPath());
    } else {
      LOGGER.debug("Skipping location overlap validation for namespace '{}'", namespace);
    }
    if (!realmConfig.getConfig(
        FeatureConfiguration.ALLOW_NAMESPACE_CUSTOM_LOCATION, catalogEntity)) {
      LOGGER.debug("Validating that namespace {} has a location inside its parent", namespace);
      validateNamespaceLocation(entity, resolvedParent);
    }
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1083,6 +1093,74 @@ private void validateNoLocationOverlap(
...
+                + "which is not enabled. Expected location: "
+                + defaultLocation);
+      }
+    } else {
+      throw new IllegalArgumentException(
+          "Failed to validate namespace "
+              + namespace.getName()
+              + " given parent "
+              + parent.getName());
+    }
```
flyrain: Minor: the logic would be slight easier to understand if we add something like this in the beginning of the method
```
    Preconditions.checkArgument(parentType.equals(PolarisEntityType.CATALOG) || parentType.equals(PolarisEntityType.NAMESPACE), "Invalid parent type");
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1083,6 +1093,74 @@ private void validateNoLocationOverlap(
...
+                    return StorageLocation.ensureTrailingSlash(
+                        StorageLocation.ensureTrailingSlash(l) + namespace.getName());
+                  })
+              .map(StorageLocation::of)
+              .toList();
+      if (!defaultLocations.contains(namespaceLocation)) {
+        throw new IllegalArgumentException(
+            "Namespace "
+                + namespace.getName()
+                + " has a custom location, "
```
flyrain, line 1134: Can we output the invalid custom location here to make debugging a bit easier?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1083,6 +1093,74 @@ private void validateNoLocationOverlap(
...
+      StorageLocation defaultLocation =
+          StorageLocation.of(
+              StorageLocation.ensureTrailingSlash(
+                  StorageLocation.ensureTrailingSlash(parentLocation) + namespace.getName()));
+      if (!defaultLocation.equals(namespaceLocation)) {
+        throw new IllegalArgumentException(
+            "Namespace "
+                + namespace.getName()
+                + " has a custom location, "
+                + "which is not enabled. Expected location: "
```
flyrain:
```suggestion
                + "which is not enabled. Expected a location within: "
```

flyrain (commented):
Thanks @eric-maynard for the fix. Left some comments.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: No, we expected a very exact location -- not a location within that path.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: Why would we validate it when allowing a custom location?

When custom locations are disallowed, we must validate the namespace location to ensure it's not custom.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: Screenshot 2025-08-26 at 9 22 03 AM

flyrain: it works when we remove the `{}` and `;`

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
flyrain: I think we are thinking about two different types of custom location
1. Free form, the locations could be in anywhere. I believe this configure is for that. Do we have a use case for this? I feel this is a wild behavior, there is no guarantee that credential vending or other location related things will work.
2. still in the parent directory, but users can give a different sub prefix/dir name other than the default one, `parent/namespace_name`

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dennishuo: Right, I was explaining why your PR description is misleading:

> in order for our overlap checks and assumptions around credential vending to be valid

Is this PR fixing assumptions around credential vending, or just adding an option for whether to enforce "standard" locations for namespaces? In general we shouldn't need namespaces to be standard to still provide the intended guarantees about enforcement of ALLOWED_LOCATIONS.

The more important bug to fix is enforcement of ALLOWED_LOCATIONS.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dennishuo: The original design was always to still securely support a situation where people might want to remap all their namespaces/tables into UUID paths, as long as they reside under the ALLOWED_LOCATIONS defined within the StorageConfig.

In this PR we need to be very precise about what we mean by "custom location". As I understand it, "custom" just means it doesn't follow the "default" where the pathname is equal to the namespace name hierarchically:

`ns1.ns2` -> `s3://basebucket/basepath1/ns1/ns2/`
`StorageConfig.defaultBaseLocation = s3://basebucket/basepath1`
`StorageConfig.allowedLocations= [s3://basebucket/basepath1, s3://basebucket/basepath2]`

If we allow a "custom location", a user might want:
`ns1.ns2` -> `s3://basebucket/basepath1/6f8af834t/`
`StorageConfig.defaultBaseLocation = s3://basebucket/basepath1`
`StorageConfig.allowedLocations= [s3://basebucket/basepath1, s3://basebucket/basepath2]`

In such a case we must still validate that the "custom" location  `s3://basebucket/basepath1/6f8af834t/` resides under *one of* the allowed locations in the StorageConfig, and by default we'd still also perform a check to prevent overlaps.

So there's three completely distinct concepts:

1. "Custom paths" -> Use supplied paths that might be uuids, rather than "default" name-friendly paths. Disabling this is fairly commonplace, and is more a *stylistic* choice by the user
2. "Overlapping paths check" -> Whether paths are custom or follow default name-friendly paths, we perform overlap checks so that RBAC can "correctly" uniquely identify a storage prefix during credential-vending. In theory, if *all* paths only use default name-based paths, we don't need overlap checks, but since "custom paths" allows someone to already remap some other namespace to the path before creating a new namespace using "default name-based paths", in general the overlap check may still be needed. Catalog admins may disable this for performance if they have full control over path specification - this is a *declarative* security option, where a catalog admin declares whether paths are fully under trusted control
3. "Allowed locations check" -> No matter what mix of custom/default locations is used for all namespaces and tables, the primary invariant is that all credential vending *must* only be under one or more of the declared ALLOWED_LOCATIONS here. Credential vending is used *internally* as well for Polaris to access metadata json, so this also applies for CreateTable even if the caller isn't asking for vended-credentials externally. This option should *not* generally be configurable for disablement. If a catalog admin wishes for a new path to be allowed in a catalog, they should update the StorageConfig to add that ALLOWED_LOCATION.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: That clause is part of a sentence:
> When we create a namespace or alter its location, we must confirm that this location is within the parent location ... in order for our overlap checks and assumptions around credential vending to be valid.

and is used in that sentence to articulate why we care that namespaces reside within their parent's location. Indeed, if you remove the clause from that context it could be misleading.

If there's another (more important) bugfix, let's submit that in parallel with this existing bugfix.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: @flyrain the current behavior is (1), so the flag protects the behavior change and lets you revert to the current behavior. An earlier version of the PR did implement (2).

This new behavior intended to behave exactly like `ALLOW_EXTERNAL_TABLE_LOCATION` for the sake of parity. I shied away from using "external" in the name because it's now a bit of a loaded term within Polaris.

eric-maynard: For the version of this PR that implements (2), see [this commit](https://github.com/apache/polaris/pull/2422/commits/b92c7b795c3d0ec3c177d00ce33039e68fefe8f9) for reference. However, given how `ALLOW_EXTERNAL_TABLE_LOCATION` behaves, I believe the new behavior is more correct.

flyrain: Given (1) is the current behavior, I'm fine with the change. We can handle the case (2) in followup PRs. Can we enrich the description to reduce the confusing?
```
  .description("If set to true, allow namespaces with custom locations.")
```
->
```
  .description("If set to true, allow namespaces with custom locations, which could escape from parent locations, allowed catalog locations. Please be aware that credential vending may not work in that case.")
```

flyrain (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sfc-gh-dhuo: No. The second description means it's the wrong feature.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sfc-gh-dhuo: My clause was also part of a paragraph:

AllowedLocations in the StorageConfig are an independent concept vs either of:

1. Whether to allow "custom" locations outside of the default path hierarchy that uses namespace names
2. Whether to allow "overlapping" locations

The most important one is validating that anything that can vend credentials resides under the allowedLocations.

Overlap and custom hierarchies in general are more common to disable validation for.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
eric-maynard: I've demoted this to a behavior change flag (not a feature) and clarified the verbiage

dimas-b:
@eric-maynard : These tests fail on `main` after this PR... not sure how they passed on the PR :sweat_smile: Could you take a look?
```
 IcebergAllowedLocationTest > testCreateTableInsideOfCatalogAllowedLocations(Path) FAILED
    java.lang.IllegalArgumentException at IcebergAllowedLocationTest.java:352

IcebergAllowedLocationTest > testCreateTableOutSideOfCatalogAllowedLocations(Path) FAILED
    java.lang.IllegalArgumentException at IcebergAllowedLocationTest.java:352
```

eric-maynard:
Hey @dimas-b, you're totally right, my bad. These two PRs had a conflict. I put up a fix in #2511.

I've talked to @flyrain about this in the past, but it would be pretty cool if we could have a [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) and force PRs to pass CI before they actually merge.

dimas-b:
+1 to merge queue... at least we should discuss it :)

Issue #2334, opened 2025-08-13
Incorrect documentation in rsa-key-pair for production.


bacek:
https://github.com/apache/polaris/blob/8996132e8fc3347790d8009f1e7663ace048c693/site/content/in-dev/unreleased/configuring-polaris-for-production.md?plain=1#L64

It actually doesn't work with Polaris, because it will generate PCKS#1, not PCKS#8.

Error stacktrace is showing PCKS#8 usage

```
     ... 62 more                                                                                                                                
 Caused by: java.security.InvalidKeyException: Unable to decode key                                                                             
     at java.base/sun.security.pkcs.PKCS8Key.decode(PKCS8Key.java:137)                                                                          
     at java.base/sun.security.pkcs.PKCS8Key.<init>(PKCS8Key.java:96)                                                                           
     at java.base/sun.security.rsa.RSAPrivateCrtKeyImpl.<init>(RSAPrivateCrtKeyImpl.java:162)                                                   
     at java.base/sun.security.rsa.RSAPrivateCrtKeyImpl.newKey(RSAPrivateCrtKeyImpl.java:92)                                                    
     at java.base/sun.security.rsa.RSAKeyFactory.generatePrivate(RSAKeyFactory.java:348)                                                        
     at java.base/sun.security.rsa.RSAKeyFactory.engineGeneratePrivate(RSAKeyFactory.java:249)                                                  
     ... 64 more                                                                                                                                
 Caused by: java.io.IOException: algid parse error, not a sequence                                                                              
     at java.base/sun.security.x509.AlgorithmId.parse(AlgorithmId.java:390)                                                                     
     at java.base/sun.security.pkcs.PKCS8Key.decode(PKCS8Key.java:112)                                                                          
     ... 69 more                                                                                                                                
                                                                                                                                                
 2025-08-13 03:42:08,654 ERROR [org.apa.pol.ser.exc.IcebergExceptionMapper] [,POLARIS] [,,,] (executor-thread-11) Unhandled exception returning 
  INTERNAL_SERVER_ERROR: java.lang.RuntimeException: Could not reconstruct the private key                                                      
     at org.apache.polaris.service.auth.PemUtils.getPrivateKey(PemUtils.java:107)                                                               
     at org.apache.polaris.service.auth.PemUtils.readPrivateKeyFromFile(PemUtils.java:120)                                                      
[...]

PR #2427, opened 2025-08-22
Use implicitly command to generate private key with PKCS#8

MonkeyCanCode:
This is to fix https://github.com/apache/polaris/issues/2334


dimas-b (approved):
The change LGTM since ` genrsa` is superseded `genpkey` in the `openssl` man page :+1: 

Yet, as noted in https://github.com/apache/polaris/issues/2334#issuecomment-3213169417 the old command produces correct output in my env. too.

PR #2430, opened 2025-08-22
chore(docs): Clarify that {prefix} is a Catalog name.

binarycat0:
Clarify that {prefix} is a Catalog name.

Add tags into "Polaris Management Service" openapi spec.

Add .vscode catalog into .gitignore.

### DIFF
--- a/spec/iceberg-rest-catalog-open-api.yaml
+++ b/spec/iceberg-rest-catalog-open-api.yaml
@@ -28,7 +28,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.0.1
+  version: 0.0.2
   description:
     Defines the specification for the first version of the REST Catalog API.
     Implementations should ideally support both Iceberg table specs v1 and v2, with priority given to v2.
@@ -79,7 +79,7 @@ paths:
           required: false
           schema:
             type: string
-          description: Warehouse location or identifier to request from the service
+          description: Warehouse location (Catalog name) or identifier to request from the service
       description:
         "
         All REST clients should first call this route to get catalog configuration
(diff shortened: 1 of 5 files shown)


spec/iceberg-rest-catalog-open-api.yaml
```diff
@@ -79,7 +79,7 @@ paths:
           required: false
           schema:
             type: string
-          description: Warehouse location or identifier to request from the service
+          description: Warehouse location (Catalog name) or identifier to request from the service
```
dimas-b, line 82: In general, we should not edit this file. It is owned and maintained under the Iceberg project. Cf. https://github.com/apache/polaris/blob/1e8dc887d691fa1c7b45ed3ae17a4475b384056a/spec/README.md

spec/polaris-management-service.yml
```diff
@@ -28,10 +28,10 @@ servers:
     variables:
       scheme:
         description: The scheme of the URI, either http or https.
-        default: https
+        default: http
```
dimas-b, line 31: Why default to `http` (without TLS)?

pingtimeout (commented):
I wonder whether it would be better to update the Polaris documentation.  Currently, the `site/content/in-dev/unreleased/polaris-catalog-service.md` file is an immediate redirection to swagger.io.  We could replace this redirection by a page that explains the important concepts that people should be aware of, like the fact that the Iceberg `{prefix}` should be set to the catalog name, and then provide a _link_ to swagger.io .

spec/polaris-management-service.yml
binarycat0: I understand that openapi specs are usually used by devs and QA and we do not use TLS by default, so for convenience I set it as simple http.

spec/iceberg-rest-catalog-open-api.yaml
flyrain: Agreed. Although putting something doc in Polaris doc is a good option, as Polaris opinionated that `prefix` is the catalog name.

Issue #2204, opened 2025-07-29
Make all generated archives reproducible


snazy:
To be able to validate a release artifact (jar, tarball, zip, etc) against the state of the source tree, the artifacts need to be generated in a reproducible way.

There are a few, totally build related differences as of today:
1. [x] jar files: `META-INF/MANIFEST.MF` have different attributes (**fixed**)
    * `Apache-Polaris-Is-Release` (depends on the `-Prelease` vs just `-PjarWithGitinfo` property)
    * `Apache-Polaris-Build-Git-Describe` (same as above)
    * `Apache-Polaris-Build-Timestamp` - could be changed to the Git commit timestamp using UTC
    * `Apache-Polaris-Build-System` - remove this one
    * `Apache-Polaris-Build-Java-Version` - remove this one
2. [x] jar files: `META-INF/jandex.idx` are reported to be different (**fixed**)
3. [x] jar files: mtime of `META-INF/MANIFEST.MF` and `META-INF/jandex.idx` vary (**fixed**)
4. [ ] `..../quarkus/generated-bytecode.jar` class files differ. Since Quarkus 3.28.2, the generated bytecode is often the same, but there is still no guarantee.
5. [x] varying order of entries in Quarkus re-assembled `.../app/polaris-*.jar` archives (https://github.com/quarkusio/quarkus/issues/50578)
6. [x] "group write" POSIX permission varies between platforms for zip/tar archive entries (fixed via #2819 + #3146) 
With all mentioned issues addressed, it should be possible to build distribution artifacts that match binary, leading to the same sha512.
7. [x] source tarballs have entries with "current mtime". The source tarballs should use a fixed mtime (fixed via #2823)
8. [x] "Top level POM" is not reproducible, because the POM `<developers>` and `<contributors>` elements are dynamically populated at build time (fixed via #2826).
9. [ ] Helm chart package tarball (generated via `helm package`) is not reproducible, `helm package` has no options to "fix" entries' mtime and POSIX attributes. See: Releasy: make Helm package reproducible (#3086)
10. [x] Source-tarball - eliminate git-gzip risk (via #3075)
11. [ ] Build: Ensure reproducible .properties files (via #3089)
12. [X] `git archive --mtime` requires a timezone (via #3143)
13. [x] Non deterministic behavior with `withXml()` for `pom.xml` customization (via #3145)

jbonofre:
Thanks for this issue, good point indeed.

snazy:
[...]

PR #2432, opened 2025-08-22
Built: improve reproducible archive files

snazy:
As part of the effort for #2204, this change fixes a few aspects around reproducible builds:

Some Gradle projects produce archive files, but don't get the necessary Gradle archive-tasks settings applied: one not-published project but also the tarball&zip of the distribution. This change moves the logic to the new build-plugin `polaris-reproducible`.

Another change is to have some Quarkus generated jar files adhere to the same conventions, which are constant timestamps for the zip entries and a deterministic order of the entries. That's sadly not a full fix, as the classes that are generated or instumented by Quarkus differ in each build.

Contributes to #2204

### DIFF
--- a/build-logic/src/main/kotlin/Utilities.kt
+++ b/build-logic/src/main/kotlin/Utilities.kt
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+import java.io.File
+import java.io.FileOutputStream
+import java.nio.file.attribute.FileTime
+import java.util.zip.ZipFile
+import java.util.zip.ZipOutputStream
 import org.gradle.api.Project
 import org.gradle.process.JavaForkOptions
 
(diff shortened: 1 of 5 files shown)


build-logic/src/main/kotlin/Utilities.kt
```diff
@@ -61,3 +66,41 @@ fun JavaForkOptions.addSparkJvmOptions() {
...
 }
+
+/**
+ * Rewrites the given ZIP file.
+ *
+ * The timestamps of all entries are set to `1980-02-01 00:00`, zip entries appear in a
+ * deterministic order.
+ */
+fun makeZipReproducible(source: File) {
+  val t = FileTime.fromMillis(318211200_000) // 1980-02-01 00:00 GMT
```
dimas-b, line 77: Why not `0` :wink:

build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -237,9 +238,20 @@ configurations.all {
...
-// ensure jars conform to reproducible builds
-// (https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives)
-tasks.withType<AbstractArchiveTask>().configureEach {
-  isPreserveFileTimestamps = false
-  isReproducibleFileOrder = true
+if (plugins.hasPlugin("io.quarkus")) {
+  tasks.named("quarkusBuild") {
+    actions.addLast {
+      listOf(
+          "quarkus-app/quarkus-run.jar",
```
dimas-b, line 245: Does Quarkus not respect `isReproducibleFileOrder`, etc.?

build-logic/src/main/kotlin/polaris-java.gradle.kts
```diff
@@ -237,9 +238,20 @@ configurations.all {
     }
 }
 
-// ensure jars conform to reproducible builds
-// (https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives)
-tasks.withType<AbstractArchiveTask>().configureEach {
-  isPreserveFileTimestamps = false
-  isReproducibleFileOrder = true
+if (plugins.hasPlugin("io.quarkus")) {
```
dimas-b, line 241: Would it be worth moving it to `polaris-server.gradle.kts`?.. or perhaps make a separate plugin dedicated to quarkus?

build-logic/src/main/kotlin/Utilities.kt
snazy: Gradle uses the same timestamp. IIRC it's got something to do with ZIP file format specialties...

build-logic/src/main/kotlin/polaris-java.gradle.kts
snazy: Ehm - nope. That's a Gradle thing.

build-logic/src/main/kotlin/Utilities.kt
dimas-b: Ah, ok :+1:

build-logic/src/main/kotlin/polaris-java.gradle.kts
snazy: We have a dedicated build-plugin for the "distributed" Quarkus builds, but also other projects that use/build with Quarkus.... So I put it into the non-java version specific one.

PR #2434, opened 2025-08-22
Add Delegator to all API Implementations

adnanhemani:
~~Draft~~ PR of the discussion on the Dev ML regarding using the Delegator pattern to add Events instrumentation to all Polaris APIs.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.admin;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -37,7 +36,6 @@ import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiService {
 
   @Inject @Delegate PolarisCatalogsApiService delegate;
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalsEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalsEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.admin;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -34,7 +33,6 @@ import org.apache.polaris.service.admin.api.PolarisPrincipalsApiService;
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class PolarisPrincipalsEventServiceDelegator implements PolarisPrincipalsApiService {
 
   @Inject @Delegate PolarisPrincipalsApiService delegate;
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalRolesEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalRolesEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.admin;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -34,7 +33,6 @@ import org.apache.polaris.service.admin.api.PolarisPrincipalRolesApiService;
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class PolarisPrincipalRolesEventServiceDelegator implements PolarisPrincipalRolesApiService {
 
   @Inject @Delegate PolarisPrincipalRolesApiService delegate;
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.catalog.generic;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -33,7 +32,6 @@ import org.apache.polaris.service.types.CreateGenericTableRequest;
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class CatalogGenericTableEventServiceDelegator
     implements PolarisCatalogGenericTableApiService, CatalogAdapter {
 
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.catalog.iceberg;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -42,7 +41,6 @@ import org.apache.polaris.service.types.NotificationRequest;
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class IcebergRestCatalogEventServiceDelegator implements IcebergRestCatalogApiService {
 
   @Inject @Delegate IcebergCatalogAdapter delegate;
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestConfigurationEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestConfigurationEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.catalog.iceberg;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -31,7 +30,6 @@ import org.apache.polaris.service.catalog.api.IcebergRestConfigurationApiService
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class IcebergRestConfigurationEventServiceDelegator
     implements IcebergRestConfigurationApiService {
 
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/CatalogPolicyEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/CatalogPolicyEventServiceDelegator.java
@@ -22,7 +22,6 @@ package org.apache.polaris.service.catalog.policy;
 import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
 import jakarta.decorator.Delegate;
-import jakarta.enterprise.inject.Alternative;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
@@ -36,7 +35,6 @@ import org.apache.polaris.service.types.UpdatePolicyRequest;
 
 @Decorator
 @Priority(1000)
-@Alternative
 public class CatalogPolicyEventServiceDelegator
     implements PolarisCatalogPolicyApiService, CatalogAdapter {
 


runtime/service/src/main/java/org/apache/polaris/service/admin/ApiBusinessLogic.java
```diff
@@ -0,0 +1,37 @@
...
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Qualifier
+@Retention(RUNTIME)
+@Target({TYPE, METHOD, FIELD, PARAMETER})
+public @interface ApiBusinessLogic {
```
dimas-b: I'm against the term "business logic". I find it very confusing. How about `PublicApi`?

dimas-b: Side note: this interface appears to be out of scope based on the PR description.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceDefaultDelegator.java
```diff
@@ -0,0 +1,333 @@
...
+import org.apache.polaris.core.admin.model.UpdatePrincipalRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
+import org.apache.polaris.service.admin.api.PolarisPrincipalRolesApiService;
+import org.apache.polaris.service.admin.api.PolarisPrincipalsApiService;
+
+
+@RequestScoped
+@Alternative
+@Priority(1000) // Will allow vendor-specific delegators to be added and used
```
dimas-b: What is a "vendor" in this context? I propose "downstream projects". It is a more neutral term and more technically correct from my POV :)

dimas-b: But on the other hand, I do not think this comment is very critical :)

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceDefaultDelegator.java
```diff
@@ -0,0 +1,333 @@
...
+import org.apache.polaris.core.admin.model.UpdatePrincipalRequest;
+import org.apache.polaris.core.admin.model.UpdatePrincipalRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
+import org.apache.polaris.service.admin.api.PolarisPrincipalRolesApiService;
+import org.apache.polaris.service.admin.api.PolarisPrincipalsApiService;
+
+
+@RequestScoped
+@Alternative
```
dimas-b: What do we achieve by always using these delegating wrappers?

runtime/service/src/main/java/org/apache/polaris/service/admin/ApiBusinessLogic.java
adnanhemani: I'm not sure `@PublicApi` captures what we're trying to say here. What do you think about something like `@ApiBaseImplementation`?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceDefaultDelegator.java
adnanhemani: That's okay - I can agree with "downstream products" :) Pushed change in this next revision!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceDefaultDelegator.java
adnanhemani: Once we get these delegating wrappers in, we can start adding the events instrumentation to these wrappers (as discussed on the Dev ML, based on the feedback from the community).

For example, the code in #1904 would then be re-written to go into these delegator classes instead.

dimas-b: Should we have something like `PolarisServiceWithEvents`?.. or `EventProducingPolarisService`?

The impl. will follow the delegator pattern. It's the "Default Delegator" in the class name that make me confused.

runtime/service/src/main/java/org/apache/polaris/service/admin/ApiBusinessLogic.java
dimas-b: I see your idea now, I think :sweat_smile: 

However, I'm not sure this is the best approach. Current delegators tae concrete classes into their constructors - this reduces option for composing the delegators with other implementations.

Since delegation is apparently covering API interfaces, I believe it is best to have one delegator per `*ApiService` interface.

Then, I propose to use `@MainService` as a `@Qualifier` for `PolarisServiceImpl`, etc.

For completeness, we should probably use a qualifier like `@EventsService` for delegators that introduce events.

This way, both sets of services can be disambiguated in this repo and in downstream projects.

adnanhemani:
Thanks for your feedback, @dimas-b! I've updated the PR accordingly - please let me know if there is further feedback :)

runtime/service/src/main/java/org/apache/polaris/service/admin/EventsServiceDelegator.java
```diff
@@ -0,0 +1,36 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import jakarta.inject.Qualifier;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+// Used to annotate delegator classes, which are emitting events
```
adutra:
```suggestion
/** Used to annotate delegator classes, which are emitting events. */
```

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.polaris.core.admin.model.UpdateCatalogRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
+
+@RequestScoped
+@Default
+@EventsServiceDelegator
+@Alternative
+@Priority(1000) // Will allow downstream project-specific delegators to be added and used
+public class PolarisCatalogsServiceDefaultDelegator implements PolarisCatalogsApiService {
```
adutra: I like the general direction we are taking in this PR, but wanted to mention that this CDI pattern has dedicated annotations: `@Decorator` and `@Delegate`.

It's explained here:

- https://quarkus.io/guides/cdi#decorators
- https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0#decorator_bean

However, these annotations require both decorator and delegate beans to implement one common interface. This might not be possible for beans like `IcebergCatalogAdapter` since it's not an interface. We'd need to extract an interface from it in order to make the Decorator CDI pattern work.

Here, it could look like this:

```java
@RequestScoped
@Default
@EventsServiceDelegator
@Decorator
@Priority(1000) // Will allow downstream project-specific delegators to be added and used
public class PolarisCatalogsServiceDefaultDelegator implements PolarisCatalogsApiService {

  private final PolarisCatalogsApiService delegate;

  @Inject
  public PolarisCatalogsServiceDefaultDelegator(@Delegate @MainService PolarisCatalogsApiService delegate) {
    this.delegate = delegate;
  }
```

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
```diff
@@ -0,0 +1,181 @@
...
+  private final PolarisCatalogsApiService delegate;
+
+  public PolarisCatalogsServiceDefaultDelegator(@MainService PolarisCatalogsApiService delegate) {
+    this.delegate = delegate;
+  }
+
+  /** From PolarisCatalogsApiService */
```
adutra: nit: do we need these dummy javadocs?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalRolesServiceDefaultDelegator.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.polaris.core.admin.model.UpdatePrincipalRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisPrincipalRolesApiService;
+
+@RequestScoped
+@Default
+@EventsServiceDelegator
+@Alternative
+@Priority(1000) // Will allow downstream project-specific delegators to be added and used
+public class PolarisPrincipalRolesServiceDefaultDelegator
```
adutra: nit: the naming convention is a bit odd, I'd expect the decorator to be named `<DelegateName>EventServiceDelegator` or something like that.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -89,6 +89,7 @@
 
 /** Concrete implementation of the Polaris API services */
 @RequestScoped
+@MainService
```
adutra: This may not be necessary anymore when using proper interfaces and `@Decorator` / `@Delegate`.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adutra: That said, I see that most adapters generally implement one or more interfaces, e.g.

```java
@RequestScoped
@MainService
public class IcebergCatalogAdapter
    implements IcebergRestCatalogApiService, IcebergRestConfigurationApiService, CatalogAdapter {
```

So, maybe using `@Decorator` could work out of the box, if both beans (decorator and delegate) implement exactly the same interfaces (and there is no ambiguity when resolving beans).

adutra (commented):
Thanks @adnanhemani for this proposal that seems to go into a clean direction.

runtime/service/src/main/java/org/apache/polaris/service/admin/EventsServiceDelegator.java
adnanhemani: Changed!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adnanhemani: These are copied from the original file, I didn't remove them. Happy to remove if you're more comfortable without them, but I originally didn't see any reason to remove these.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalRolesServiceDefaultDelegator.java
adnanhemani: hmm - I've tried naming everything into `<DelegateName>DefaultDelegator` rather than "...EventServiceDelegator" because in the following PRs, we will add the Events instrumentation to this Delegator. In other words, generating Events will be the default delegator. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adnanhemani: This is a good suggestion, it helps make the code a bit tighter as well. Added in the next revision!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adnanhemani: You're correct - removed!

adnanhemani (commented):
Thanks @adutra for your feedback, I've push the revision addressing all the comments!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalRolesServiceDefaultDelegator.java
adutra: So... what happens if we need delegators for something else? E.g. `PolarisDiagnostics` could be turned into another delegator, e.g. `<DelegateName>Diagnostics`.

The beauty of `@Decorator` is that you can in theory have as many decorators you need, and each time the decorated interface is injected, CDI would transparently create a chain of invocations that traverse all the decorators up to the "undecorated" bean. IOW if you inject `PolarisCatalogsApiService` somewhere, calling `PolarisCatalogsApiService.createCatalog()` would actually call `<DelegateName>Diagnostics.createCatalog()` then `<DelegateName>EventServiceDelegator.createCatalog()` then `PolarisServiceImpl.createCatalog()`.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adutra: Hm what original file? I thought this file was new. This is a minor point though, so feel free to handle it as you see fit.

adutra (dismissed):
LGTM thanks @adnanhemani – the naming convention is still imo a bit strange but we can change that later.

runtime/service/src/main/java/org/apache/polaris/service/admin/EventsServiceDelegator.java
```diff
@@ -0,0 +1,36 @@
...
+
+import jakarta.inject.Qualifier;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/** Used to annotate delegator classes, which are emitting events. */
+@Qualifier
+@Retention(RUNTIME)
+@Target({TYPE, METHOD, FIELD, PARAMETER})
+public @interface EventsServiceDelegator {}
```
dimas-b: maybe nitpicky, but "delegator" feels odd to me in a qualifier. A qualifier is supposed highlight a functional aspect of a bean, IMHO. Whether or not the bean delegates to other beans or genuinely implements and API should be be exposed in a qualifier, IMHO.

Would `ServiceWithEvents` or `EventsEmitter` (I assume it's supposed to emit events eventually) work?

dimas-b:
I'm fine addressing naming concerns later if there's a general understanding that class and qualifier names will change :)

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
```diff
@@ -0,0 +1,176 @@
...
+import org.apache.polaris.core.admin.model.CreateCatalogRoleRequest;
+import org.apache.polaris.core.admin.model.RevokeGrantRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
+
+@EventsServiceDelegator
+@Default
+@Decorator
```
dimas-b: Using decorators is an interesting approach.

However, AFAIK, decorators are impossible to remove in downstream builds, right? I mean that if a decorator is discovered, it will be applied and all decorators will be applied in priority order. This means that all downstream users or `runtime/service` will always get this decorator in call paths.

This is fine, but I believe it might be wise to isolate the decorator in a new module (jar) to allow downstream projects to opt out of it.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
```diff
@@ -0,0 +1,176 @@
...
+import org.apache.polaris.core.admin.model.AddGrantRequest;
+import org.apache.polaris.core.admin.model.CreateCatalogRequest;
+import org.apache.polaris.core.admin.model.CreateCatalogRoleRequest;
+import org.apache.polaris.core.admin.model.RevokeGrantRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
+
+@EventsServiceDelegator
```
dimas-b: The idea in my previous comment when I suggested using custom qualifiers was to use them for composing the delegation chain of *ApiService implementations.

However, if we go with the decorator approach, I believe this qualifier becomes rather useless :thinking: WDYT?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adutra: That is correct. This situation is generally solved by either deactivating the decorator through some configuration flag, or, as you pointed out, by isolating the decorators in separated modules that integrators are free to include or exclude.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adutra: I asked myself the same question - but it all depends on how we are going to use these qualifiers, so I figured we would figure that out later 😅

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
dimas-b: Good point. I suppose `@IfBuildProperty(...)` might be good enough for now (without introducing new modules).

... or perhaps `@UnlessBuildProperty(...)`.

eric-maynard (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalRolesServiceDefaultDelegator.java
adnanhemani: I understand what you are saying now - I think I was originally closer to Dmitri's idea earlier in this PR by attempting to use a delegator wrapper without the Decorator pattern. Let me put a general comment on the PR regarding this and change the naming convention to more accurately reflect the Decorator pattern.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adnanhemani: Agreed here with both of you. I'm leaning towards the `@UnlessBuildProperty` - but am thinking to leave this PR without it and let a future contribution make that change if anyone requires it.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adnanhemani: Sorry, I meant it came from `PolarisServiceImpl.java`, which was the base file that I was using for creating this file. Let me remove.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adnanhemani: I think you're right @dimas-b. Let me remove it for now and then if we think it'll be useful later, we can always reintroduce it in a new change!

adnanhemani (commented):
Thanks all for the feedback! There was a gradual shift from originally using a delegator wrapper to using the Decorator pattern in this PR - but I think we have reached convergence on this pattern based on the comments in this PR.

Naming conventions are still not finalized, but I believe this is the closest we have been so far to something usable as a base. I will be releasing further PRs using this as a base shortly for further event instrumentation.

singhpk234 (approved):

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.polaris.core.admin.model.CreateCatalogRoleRequest;
+import org.apache.polaris.core.admin.model.RevokeGrantRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRoleRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
+
+@Decorator
+@Priority(1000)
+@Alternative
```
adutra: Hmm I think `@Decorator` is mutually exclusive with `@Alternative`.

Alternative beans provide a whole different bean implementation. Use it when you want to _suppress_ the original bean's behavior. It's similar to mocking.

Decorators on the other hand are used to further specialize the original bean's behavior, _without suppressing it_. It's similar to AOP interceptors.

IMO what we are trying to achieve here is the second approach where the original bean keeps doing its job, but we add behavior before/after each invocation of the bean.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
adutra: That works for me 👍

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adnanhemani: You are correct - I misread some documentation while looking up what would be required if we wanted to use `@UnlessBuildProperty` in the future and thought that `@Alternative` would be required at that time. Let me remove that property and push again!

adutra (approved):

RussellSpitzer:
Looks good to me, I learned a lot about the delegator annotations allowed in the CDI from the discussion and it seems like this is a very straightforward pattern.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsServiceDefaultDelegator.java
dimas-b: Fair enough

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisPrincipalsEventServiceDelegator.java
```diff
@@ -0,0 +1,102 @@
...
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.SecurityContext;
+import org.apache.polaris.core.admin.model.CreatePrincipalRequest;
+import org.apache.polaris.core.admin.model.GrantPrincipalRoleRequest;
+import org.apache.polaris.core.admin.model.UpdatePrincipalRequest;
+import org.apache.polaris.core.context.RealmContext;
+import org.apache.polaris.service.admin.api.PolarisPrincipalsApiService;
+
+@Decorator
+@Priority(1000)
```
dimas-b, line 35: nit: It might be worth adding a public constant for this priority value so that downstream projects could reference it... but it can be done later.

dimas-b (approved):

PR #2445, opened 2025-08-25
Add user principal tag in metrics

### LINKED ISSUE #2444: Report user principal tag in metrics

Currently we emit various metrics and have the option of tagging them with the realm-id, when `polaris.metrics.realm-id-tag.enable-in-api-metrics=true`. 
In some cases it would be useful to also annotate the metrics with a user-principal tag, such that specific issues can be better traced at a user level.

We could add a `polaris.metrics.user-principal-tag.enable-in-api-metrics=true`.

The user principal is already present in the SecurityContext, which can be easily annotated with a `MeterTag` in the api template.

We could log the catalog name but this is more complicated and would require a more intrusive change. 

### Additional context

_No response_

fivetran-kostaszoumpatianos:
Fixes #2444.

This PR adds a user principal tag in metrics and an associated configuration option that turns it on. 
This is the following: `polaris.metrics.user-principal-tag.enable-in-api-metrics`, and **by default this is set to `false`.**

To retrieve the user principal ID it uses the SecurityContext and annotates it with a `MeterTag` annotation under the key `principal`.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
@@ -44,6 +44,7 @@ import org.apache.polaris.service.context.RealmContextResolver;
 import org.apache.polaris.service.context.TestRealmContextResolver;
 import org.apache.polaris.service.events.PolarisEventListener;
 import org.apache.polaris.service.events.TestPolarisEventListener;
+import org.apache.polaris.service.metrics.MetricsConfiguration;
 import org.apache.polaris.service.persistence.InMemoryPolarisMetaStoreManagerFactory;
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigValue;
@@ -111,6 +112,17 @@ public class ProductionReadinessChecks {
             severe);
       }
     }
+  }
+
+  @Produces
+  public ProductionReadinessCheck checkMetricTags(MetricsConfiguration config) {
+    if (config.userPrincipalTag().enableInApiMetrics()) {
+      return ProductionReadinessCheck.of(
+          Error.of(
+              "Metrics configuration includes user principal name in tags.",
+              "polaris.metrics.user-principal-tag.enable-in-api-metrics"));
+    }
+    return ProductionReadinessCheck.OK;
   }
 
   @Produces
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
@@ -116,11 +116,16 @@ public class ProductionReadinessChecks {
 
   @Produces
   public ProductionReadinessCheck checkMetricTags(MetricsConfiguration config) {
-    if (config.userPrincipalTag().enableInApiMetrics()) {
+    if (config.userPrincipalTag().enableInApiMetrics()
+        && config.realmIdTag().enableInApiMetrics()) {
       return ProductionReadinessCheck.of(
           Error.of(
-              "Metrics configuration includes user principal name in tags.",
+              "Metrics configuration includes both user principal name and realm id in tags this could cause performance implications.",
               "polaris.metrics.user-principal-tag.enable-in-api-metrics"));
+    } else {
+      LOGGER.warn(
+          "Metrics configuration includes user principal name in tags. "
+              + "This could have performance implications.");
     }
     return ProductionReadinessCheck.OK;
   }
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
@@ -115,17 +115,25 @@ public class ProductionReadinessChecks {
   }
 
   @Produces
-  public ProductionReadinessCheck checkMetricTags(MetricsConfiguration config) {
+  public ProductionReadinessCheck checkUserPrincipalMetricTag(MetricsConfiguration config) {
+    if (config.userPrincipalTag().enableInApiMetrics()) {
+      return ProductionReadinessCheck.of(
+          Error.of(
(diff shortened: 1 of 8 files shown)


runtime/service/src/main/java/org/apache/polaris/service/metrics/MetricsConfiguration.java
```diff
@@ -65,4 +68,17 @@ interface RealmIdTag {
...
+  interface UserPrincipalTag {
+
+    /**
+     * Whether to include the User Principal tag in the API request metrics.
+     *
+     * <p>Beware that if the cardinality of this tag is too high, it can cause performance issues or
+     * even crash the server.
+     */
+    @WithDefault("false")
+    boolean enableInApiMetrics();
```
dimas-b, line 81: Defaulting to `false` here looks reasonable to me. Still, since this may expose sensitive information, please open a `dev` email discussion for visibility.

fivetran-kostaszoumpatianos:
> LGTM 👍 Thanks for your contribution, @fivetran-kostaszoumpatianos !

Thanks @dimas-b I will send a `dev` email.

runtime/service/src/main/java/org/apache/polaris/service/metrics/MetricsConfiguration.java
fivetran-kostaszoumpatianos: Thanks @dimas-b I sent an email to `dev`.

fivetran-kostaszoumpatianos:
@dimas-b I had to re-format the code, and your approval got removed. Could you please take another look? Thanks!

adutra:
FYI here is the ML discussion thread:

https://lists.apache.org/thread/o7of5dpmglmkjosftqsyr54x3dcfo1o4

I replied there but to summarize my opinion here: I'm not against the idea but there are security risks. Imo this feature should be opt-in (which is the case) and guarded by a production readiness check.

fivetran-kostaszoumpatianos:
> FYI here is the ML discussion thread:
> 
> https://lists.apache.org/thread/o7of5dpmglmkjosftqsyr54x3dcfo1o4
> 
> I replied there but to summarize my opinion here: I'm not against the idea but there are security risks. Imo this feature should be opt-in (which is the case) and guarded by a production readiness check.

Thank you @adutra I have added a production readiness check.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -113,6 +114,17 @@ public void warnOnFailedChecks(
     }
   }
 
+  @Produces
+  public ProductionReadinessCheck checkMetricTags(MetricsConfiguration config) {
+    if (config.userPrincipalTag().enableInApiMetrics()) {
+      return ProductionReadinessCheck.of(
+          Error.of(
+              "Metrics configuration includes user principal name in tags.",
```
dimas-b: This will effectively always produce WARN log messages in servers where principal tags are enabled, even if the admin user configured that explicitly. From my POV this is not ideal user experience, but having this WARN is probably better that accidentally exposing principal names.

I think we may want to add per-message suppression for production checks at some point.

dimas-b (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -113,6 +114,17 @@ public void warnOnFailedChecks(
     }
   }
 
+  @Produces
+  public ProductionReadinessCheck checkMetricTags(MetricsConfiguration config) {
+    if (config.userPrincipalTag().enableInApiMetrics()) {
```
adutra, line 119: I think we are missing a readiness check when _both_ tags are enabled, since this is the most dangerous situation.

runtime/service/src/test/java/org/apache/polaris/service/metrics/RealmIdTagEnabledMetricsTest.java
```diff
@@ -39,7 +39,9 @@ public Map<String, String> getConfigOverrides() {
           "polaris.metrics.realm-id-tag.enable-in-api-metrics",
           "true",
           "polaris.metrics.realm-id-tag.enable-in-http-metrics",
-          "true");
+          "true",
```
adutra, line 42: At some point I think we'll need to refactor these tests and create a sort of "parameterized Quarkus test" that tests all the combinations of (realm tag on/off) x (principal tag on/off). But we can look into that later.

eric-maynard (approved):
As long as it's off by default, this LGTM!

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
fivetran-kostaszoumpatianos: Yes, I agree. We should have that at some point. In the interim we could get away with just having the warning there. I can open an issue and work on it.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
fivetran-kostaszoumpatianos: Indeed, this is a great point. I added one.

runtime/service/src/test/java/org/apache/polaris/service/metrics/RealmIdTagEnabledMetricsTest.java
fivetran-kostaszoumpatianos: Yes, good point. Thanks!

adutra (approved):

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
fivetran-kostaszoumpatianos: @dimas-b 
For selectively turning off readiness checks specific to a given list of offending parameters, I created an issue: https://github.com/apache/polaris/issues/2471 and an associated PR: https://github.com/apache/polaris/pull/2472

PR #931, opened 2025-02-03
Add profile support for cli

MonkeyCanCode:
Helper:
```
➜  polaris git:(cli_profile) ./polaris profiles --help
input: polaris profiles --help
options:
  profiles
    create
      Positional arguments:
        profile
    delete
      Positional arguments:
        profile
    get
      Positional arguments:
        profile
    list
    update
      Positional arguments:
        profile
```

Create profile:
```
➜  polaris git:(cli_profile) ./polaris profiles create dev
Polaris Client ID: root
Polaris Client Secret: 712f1f6daf16326fd794607ef1ce90d9
Polaris Host [localhost]:
Polaris Port [8181]:
Polaris profile dev created successfully.
```

Get profile:
```
➜  polaris git:(cli_profile) ./polaris profiles get dev
Polaris profile dev: {'client_id': 'root', 'client_secret': '712f1f6daf16326fd794607ef1ce90d9', 'host': 'localhost', 'port': 8181}
 ```

List profiles:
```
➜  polaris git:(cli_profile) ./polaris profiles list
Polaris profiles:
 - dev
 ```

Delete profile:
```
➜  polaris git:(cli_profile) ./polaris profiles create test
Polaris Client ID: 1
Polaris Client Secret: 2
Polaris Host [localhost]:
Polaris Port [8181]:
Polaris profile test created successfully.
➜  polaris git:(cli_profile) ./polaris profiles list
Polaris profiles:
 - dev
 - test
➜  polaris git:(cli_profile) ./polaris profiles delete test
Polaris profile test deleted successfully.
➜  polaris git:(cli_profile) ./polaris profiles list
Polaris profiles:
 - dev
```

Update profile:
```
➜  polaris git:(cli_profile) ./polaris profiles create test
Polaris Client ID: 1
Polaris Client Secret: 2
Polaris Host [localhost]:
Polaris Port [8181]:
Polaris profile test created successfully.
➜  polaris git:(cli_profile) ./polaris profiles update test
Polaris Client ID [1]: test
Polaris Client Secret [2]: pass
Polaris Client ID [localhost]:
Polaris Client Secret [8181]:
Polaris profile test updated successfully.
➜  polaris git:(cli_profile) ./polaris profiles get test
Polaris profile test: {'client_id': 'test', 'client_secret': 'pass', 'host': 'localhost', 'port': 8181}
```

Use profile with CLI:
```
➜  polaris git:(cli_profile) ./polaris --profile dev principals create test
[...]

PR #2448, opened 2025-08-26
Remove SCRIPT_DIR and default config location to user home

MonkeyCanCode:
This PR remove the use/needed of `SCRIPT_DIR` and change the default config location to user home (~/.polaris/)

### DIFF
--- a/client/python/cli/constants.py
+++ b/client/python/cli/constants.py
@@ -371,5 +371,5 @@ CLIENT_SECRET_ENV = "CLIENT_SECRET"
 CLIENT_PROFILE_ENV = "CLIENT_PROFILE"
 DEFAULT_HOSTNAME = "localhost"
 DEFAULT_PORT = 8181
-CONFIG_DIR = os.path.expanduser("~/.polaris")
+CONFIG_DIR = (os.environ.get("POLARIS_HOME") or "").strip() or os.path.expanduser("~/.polaris")
 CONFIG_FILE = os.path.join(CONFIG_DIR, ".polaris.json")
--- a/client/python/cli/command/profiles.py
+++ b/client/python/cli/command/profiles.py
@@ -56,6 +56,7 @@ class ProfilesCommand(Command):
         if not os.path.exists(CONFIG_FILE):
             return {}
         with open(CONFIG_FILE, "r") as f:
+            print(f"Loading profiles from {CONFIG_FILE}")
             return json.load(f)
 
     def _save_profiles(self, profiles: Dict[str, Dict[str, str]]) -> None:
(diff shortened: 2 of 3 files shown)


eric-maynard:
IIRC I used the home directory in an earlier version of the CLI, but this caused issues with multiple Polaris environments conflicting. Keeping the `venv` and everything within the Polaris directory actually seems like a good thing to me, what do we gain by using home instead?

MonkeyCanCode:
I think the problem with Polaris directory is for users who only want to use CLI and not downloading the entire code base, there won't be a repo path for them to use. Maybe we can allow users to specify where to put as an overwrite?

> IIRC I used the home directory in an earlier version of the CLI, but this caused issues with multiple Polaris environments conflicting. Keeping the `venv` and everything within the Polaris directory actually seems like a good thing to me, what do we gain by using home instead?

eric-maynard:
+1 to allowing a custom location, good idea. We could also default to something like `~/.local/lib` or `/usr/local/` if you think that's better than SCRIPT_DIR as a default.

MonkeyCanCode:
Sounds good. Let me draft something later this week then ask for review again. Thanks.

> +1 to allowing a custom location, good idea. We could also default to something like `~/.local/lib` or `/usr/local/` if you think that's better than SCRIPT_DIR as a default.

MonkeyCanCode:
> to allowing a custom location, good idea. We could also default to something like `~/.local/lib` or `/usr/local/` if you think that's better than SCRIPT_DIR as a default.

@eric-maynard do you mind take another look?

client/python/cli/constants.py
```diff
@@ -371,9 +371,5 @@ class Namespaces:
 CLIENT_PROFILE_ENV = "CLIENT_PROFILE"
 DEFAULT_HOSTNAME = "localhost"
 DEFAULT_PORT = 8181
-CONFIG_DIR = os.environ.get("SCRIPT_DIR")
-if CONFIG_DIR is None:
-    raise Exception(
-        "The SCRIPT_DIR environment variable is not set. Please set it to the Polaris's script directory."
-    )
+CONFIG_DIR = (os.environ.get("POLARIS_HOME") or "").strip() or os.path.expanduser("~/.polaris")
```
HonahX, line 374: Should we fall back to use SCRIPT_DIR if given for backward compatibility?

HonahX (commented):
Thanks for the change!

HonahX (approved):

client/python/cli/constants.py
MonkeyCanCode: I am not sure if anyone is using it and we never documented this as well. If that is preferred, we can. Let me know what you think.

flyrain: The script dir was introduce by #931 in Feb 2025, which isn't far ago, also the cli was never in any binary distribution. With that, I'm OK to remove it directly.

flyrain (approved):

flyrain:
Thanks @MonkeyCanCode for the PR. Thanks @eric-maynard  and @HonahX for the review.

PR #2456, opened 2025-08-27
Remove commons-lang3 dependency

XN137:
outside of tests we can replace the functionality with jdk11 and guava. also stop using `org.assertj.core.util` as its a non-public api.


dimas-b (approved):
LGTM :+1:

eric-maynard (approved):

flyrain (approved):

PR #2462, opened 2025-08-28
Remove hadoop-common from polaris-runtime-service

XN137:
since https://github.com/apache/iceberg/commit/9bc6853938533c9c1dfac497229c8f4c30672458 passing a hadoop Configuration to `CatalogUtil.loadCatalog` is optional.

since `hadoop-common` brings in a lot of dependencies (and thus potential vulnerabilities) we should avoid depending on it in our core modules if not strictly required.

note that we were using the transitive `commons-collections` dependency in a spot that we have to adjust.

### DIFF
--- a/runtime/service/build.gradle.kts
+++ b/runtime/service/build.gradle.kts
@@ -75,22 +75,6 @@ dependencies {
 
   implementation(libs.auth0.jwt)
 
-  implementation(libs.hadoop.common) {
-    exclude("org.slf4j", "slf4j-reload4j")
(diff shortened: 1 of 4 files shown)


adutra (dismissed):

dimas-b (dismissed):

XN137:
rebased after trivial merge conflict in `IcebergCatalogHandlerAuthzTest.java`

singhpk234 (approved):

flyrain (approved):

Issue #2352, opened 2025-08-14
Optimize JdbcBasePersistenceImpl.listEntities


XN137:
### Is your feature request related to a problem? Please describe.

in https://github.com/apache/polaris/pull/2317 a TODO got added to the jdbc persistence implementation:
https://github.com/apache/polaris/blob/a8ca16b65bd0dcc013a4fc715b4111202b912c15/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java#L424-L443

currently the method is simply forwarding to the non-optimized `listEntities` variant.

### Describe the solution you'd like

the idea is that in the JDBC query we should only load the columns/properties needed for creating a `EntityNameLookupRecord` instance.
the more general `listEntities` variant can be used as a starting point and trimmed down to just the necessary parts.

also according to this [review comment](https://github.com/apache/polaris/pull/2317#discussion_r2271168766) we should make sure that the query in the optimized case can be served efficiently from the database by only having to read a single index (that contains all necessary columns) without having to access the main table/row storage (see also [this explanation](https://github.com/apache/polaris/pull/2262#pullrequestreview-3099480896)).

this may require a new or updated schema for the jdbc backend.

binarycat0:
Hello. I have capacity, please assign it to me.

snazy:
Sure, feel free to work on it

binarycat0:
The main changes of the Java code have been accepted and merged into master, but changes regarding the addition of a new JDBC Metastore index have been moved to discussion on the DEV mailinglist. 
Please join the discussion and share your Pros/Cons in favor of adding the new index - [[DISCUSS] Add JDBC Metastore table index](https://lists.apache.org/thread/xmjjot2s1v734jfx3m6ggf7djn50q1w3)

XN137:
the java changes are in and the ML discussion for adding an index didnt seem to get much traction.
admins are free to add the index if they notice slow/inefficient queries around listing tables/namespaces.
so i'll mark this ticket as closed.

PR #2465, opened 2025-08-28
perf(refactor): optimizing JdbcBasePersistenceImpl.listEntities

binarycat0:
org.apache.polaris.persistence.relational.jdbc.JdbcBasePersistenceImpl.listEntities

- Reduced Column Selection: Only 6 columns instead of 16
- Eliminated Object Creation Overhead: Direct conversion to EntityNameLookupRecord without intermediate PolarisBaseEntity
- minor refactoring

Issue: #2352 

---
Query EXPLAIN plan shown in the comment: 
https://github.com/apache/polaris/pull/2465#issuecomment-3244547998

### Follow-ups
- https://github.com/apache/polaris/issues/2487 
- https://github.com/apache/polaris/issues/2488


singhpk234:
Thank you for this change @binarycat0 do you have some perf numbers on how much this benefits compared to existing

binarycat0:
@singhpk234 Hello! Good question. I did not do any performance measurements, but potencially it should help reduce ram and cpu consumption. Could you share a link/clue how I can perform perf testing and compare previous and the current solution?

persistence/relational-jdbc/src/main/resources/postgres/schema-v2.sql
```diff
@@ -60,6 +60,7 @@ CREATE INDEX IF NOT EXISTS idx_entities ON entities (realm_id, catalog_id, id);
 CREATE INDEX IF NOT EXISTS idx_locations
     ON entities USING btree (realm_id, parent_id, location_without_scheme)
     WHERE location_without_scheme IS NOT NULL;
+CREATE INDEX IF NOT EXISTS idx_entities_lookup ON entities (realm_id, catalog_id, parent_id, type_code, sub_type_code, id, name);
```
dimas-b: Could you post an explain plan that shows how this index is used in Persistence API calls?

persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql
```diff
@@ -63,6 +63,7 @@ CREATE INDEX IF NOT EXISTS idx_locations ON entities(realm_id, catalog_id, locat
 
 -- TODO: create indexes based on all query pattern.
 CREATE INDEX IF NOT EXISTS idx_entities ON entities (realm_id, catalog_id, id);
+CREATE INDEX IF NOT EXISTS idx_entities_lookup ON entities (realm_id, catalog_id, parent_id, type_code, sub_type_code, id, name);
```
dimas-b: Do we care for this index in H2? Does anything break if it's not present? Performance is not a concern with H2, I assume.

dimas-b (commented):
LGTM overall :+1: 

Re: perf. testing, from my side, I'd ask for a test that progressively adds entities and shows various API response times as a function of time (or equivalently of the number of entities).

The hardware does not have to be very fast. The point is to compare old code with new code on the same hardware.

binarycat0:
## Test ENV
Docker image postgres:17

Rows in total = 899040.
Expected rows result = 200005.

```
CREATE INDEX IF NOT EXISTS idx_entities_lookup ON entities (realm_id, catalog_id, parent_id, type_code, sub_type_code);
```

## Explain query
```
EXPLAIN ANALYZE (SELECT id, catalog_id, parent_id, type_code, name, sub_type_code FROM POLARIS_SCHEMA.ENTITIES WHERE catalog_id = 6911759911282783572 AND sub_type_code = 0 AND realm_id = 'POLARIS' AND parent_id = 6911759911282783572 AND type_code = 6);
```

## Results
### With Index
```
Bitmap Heap Scan on entities  (cost=956.71..19256.08 rows=44861 width=50) (actual time=7.143..32.498 rows=200005 loops=1)
  Recheck Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6) AND (sub_type_code = 0))
  Heap Blocks: exact=3848
  ->  Bitmap Index Scan on idx_entities_lookup  (cost=0.00..945.49 rows=44861 width=0) (actual time=6.378..6.378 rows=200005 loops=1)
        Index Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6) AND (sub_type_code = 0))
Planning Time: 0.111 ms
Execution Time: 37.933 ms
```

### With no Index
```
Bitmap Heap Scan on entities  (cost=3960.68..22260.05 rows=44861 width=50) (actual time=16.567..45.052 rows=200005 loops=1)
  Recheck Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6))
  Filter: (sub_type_code = 0)
  Heap Blocks: exact=3848
  ->  Bitmap Index Scan on constraint_name  (cost=0.00..3949.47 rows=44861 width=0) (actual time=16.043..16.043 rows=200005 loops=1)
        Index Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6))
Planning Time: 0.150 ms
Execution Time: 51.119 ms
```

## Conclusion
Index helps reduce query cost to 75% .. 13% and reduce query execution time up to 25%;

cc: @dimas-b @singhpk234

persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql
binarycat0: I'm not sure. I was trying to follow the comment above: 
`-- TODO: create indexes based on all query pattern`

I can remove this change if it's not relevant.

binarycat0:
@singhpk234 My colleagues and I conducted several different experiments and we didn't notice significant improvement on the API side. This is most likely primarily dependent on `EntityCache`, which greatly improves performance in the current implementation. 
Therefore, I can say that the current changes don't significantly improve API performance, but the additional DB INDEX improves SQL experience and reduces query execution time. 
Also, in my tests I didn't notice a difference when measuring EXPLAIN PLAN for (select *) versus (select colA, colB...)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -421,38 +422,13 @@ public List<PolarisChangeTrackingVersions> lookupEntityVersions(
...
-        entity -> true,
-        EntityNameLookupRecord::new,
-        pageToken);
-  }
-
-  @Nonnull
-  @Override
-  public <T> Page<T> loadEntities(
-      @Nonnull PolarisCallContext callCtx,
+  private PreparedQuery buildEntityQuery(
```
XN137, line 429: wondering: can it be static?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/EntityNameLookup.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.apache.polaris.persistence.relational.jdbc.models;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Map;
+import org.apache.polaris.core.entity.EntityNameLookupRecord;
+import org.apache.polaris.persistence.relational.jdbc.DatabaseType;
+
+public class EntityNameLookup implements Converter<EntityNameLookupRecord> {
```
XN137: rename to `EntityNameLookupRecordConverter` ?

persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql
```diff
@@ -63,6 +63,9 @@ CREATE INDEX IF NOT EXISTS idx_locations ON entities(realm_id, catalog_id, locat
 
 -- TODO: create indexes based on all query pattern.
 CREATE INDEX IF NOT EXISTS idx_entities ON entities (realm_id, catalog_id, id);
+CREATE INDEX IF NOT EXISTS idx_entities_lookup
+    ON entities (realm_id, catalog_id, parent_id, type_code, sub_type_code)
+    INCLUDE (id, name);
```
XN137: according to https://h2database.com/html/commands.html#create_index `INCLUDE` can only be used on a unique index in H2

did we confirm this works?

binarycat0:
# Test ENV
OS: `24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041 arm64`
Storage: Docker image `postgres:17`
Rows in total = 899040.
Expected rows result = 200005.

# Results
## Main build version

```
EXPLAIN ANALYZE
(SELECT id, catalog_id, parent_id, type_code, name, entity_version, sub_type_code, create_timestamp, drop_timestamp, purge_timestamp, to_purge_timestamp, last_update_timestamp, properties, internal_properties, grant_records_version, location_without_scheme
 FROM POLARIS_SCHEMA.ENTITIES
 WHERE catalog_id = 6911759911282783572 AND sub_type_code = 0 AND realm_id = 'POLARIS' AND parent_id = 6911759911282783572 AND type_code = 6);
```

```
Bitmap Heap Scan on entities  (cost=3960.68..22260.05 rows=44861 width=140) (actual time=16.287..43.851 rows=200005 loops=1)
  Recheck Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6))
  Filter: (sub_type_code = 0)
  Heap Blocks: exact=3848
  ->  Bitmap Index Scan on constraint_name  (cost=0.00..3949.47 rows=44861 width=0) (actual time=15.792..15.792 rows=200005 loops=1)
        Index Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6))
Planning Time: 0.239 ms
Execution Time: 49.253 ms
```

## New build version
```
CREATE INDEX IF NOT EXISTS idx_entities_lookup
    ON entities USING btree (realm_id, catalog_id, parent_id, type_code, sub_type_code)
    INCLUDE (id, name);
```

```
EXPLAIN ANALYZE
(SELECT id, catalog_id, parent_id, type_code, name, sub_type_code 
 FROM POLARIS_SCHEMA.ENTITIES
 WHERE catalog_id = 6911759911282783572 AND sub_type_code = 0 AND realm_id = 'POLARIS' AND parent_id = 6911759911282783572 AND type_code = 6);
```

```
Index Only Scan using idx_entities_lookup on entities  (cost=0.42..2975.89 rows=42744 width=50) (actual time=0.030..29.235 rows=200005 loops=1)
  Index Cond: ((realm_id = 'POLARIS'::text) AND (catalog_id = '6911759911282783572'::bigint) AND (parent_id = '6911759911282783572'::bigint) AND (type_code = 6) AND (sub_type_code = 0))
  Heap Fetches: 0
Planning Time: 0.089 ms
Execution Time: 35.317 ms
```

# Conclusion
Using `Index Only Scan` instead of `Bitmap Heap Scan` helps:
- reduce query execution time up to 28%;
- reducing query cost up to 87%;

cc: @dimas-b @singhpk234

polaris-core/src/main/java/org/apache/polaris/core/entity/HasEntityId.java
```diff
@@ -0,0 +1,26 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.core.entity;
+
+public interface HasEntityId {
+  long getCatalogId();
```
XN137: should we add the same documentation as the fields in `PolarisEntityCore` ?

> // the id of the catalog associated to that entity. NULL_ID if this entity is top-level like
> // a catalog

polaris-core/src/main/java/org/apache/polaris/core/persistence/pagination/EntityIdToken.java
```diff
@@ -41,7 +42,7 @@ default String getT() {
     return ID;
   }
 
-  static @Nullable EntityIdToken fromEntity(PolarisBaseEntity entity) {
+  static @Nullable EntityIdToken fromEntity(HasEntityId entity) {
```
XN137: nit: mark the `entity` parameter `@Nullable`

XN137 (commented):
thanks for working on this, left some minor comments but looks good overall now.

imo since this addresses an existing TODO it shouldnt be strictly required to demonstrate elaborate perf numbers.

what we would need though is a comparison of the postgres explain plan before and after these changes for operations that use `PolarisMetaStoreManager.listEntities` i.e. listing tables below a parent namespace.
previously is should be an inefficient "table scan" where-as now it is an [index-only scan](https://www.postgresql.org/docs/current/indexes-index-only-scans.html) ...
whether that shows an immediately measurable performance improvements depends on a lot of details about benchmark design and postgres internals.

persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql
binarycat0: CI tests already failed and I removed INCLUDE instruction. I have put `id, name` under the main brackets `ON (...)`. 
My bad, I misunderstood the doc - https://www.h2database.com/html/commands.html#create_index.

persistence/relational-jdbc/src/main/resources/postgres/schema-v2.sql
binarycat0: Posted EXPLAIN plan below - https://github.com/apache/polaris/pull/2465#issuecomment-3244547998 
Please check.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
binarycat0: Definitely It can be, but it would require extend method signature due to using `this.realmId` in method. I believe it can be refactored in future when required.

polaris-core/src/main/java/org/apache/polaris/core/entity/HasEntityId.java
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ */
+
+package org.apache.polaris.core.entity;
+
+/**
+ * An interface for entity types that have an id. These entities provide a * method `getId` and
+ * `getCatalogId` to identify the entity.
+ */
+public interface HasEntityId {
```
adutra: nit: how about `Identifiable`?

persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql
```diff
@@ -63,6 +63,8 @@ CREATE INDEX IF NOT EXISTS idx_locations ON entities(realm_id, catalog_id, locat
 
 -- TODO: create indexes based on all query pattern.
 CREATE INDEX IF NOT EXISTS idx_entities ON entities (realm_id, catalog_id, id);
+CREATE INDEX IF NOT EXISTS idx_entities_lookup
+    ON entities (realm_id, catalog_id, parent_id, type_code, sub_type_code, id, name);
```
adutra: Why there is no `INCLUDE (id, name)` clause here?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/EntityNameLookupRecordConverter.java
```diff
@@ -0,0 +1,46 @@
...
+        rs.getLong("catalog_id"),
+        rs.getLong("id"),
+        rs.getLong("parent_id"),
+        rs.getString("name"),
+        rs.getInt("type_code"),
+        rs.getInt("sub_type_code"));
+  }
+
+  @Override
+  public Map<String, Object> toMap(DatabaseType databaseType) {
```
adutra, line 42: The `Converter` interface is a bit strange indeed. The method above is a pure converter utility, but this one looks like it's meant to convert the object's _internal state_ to some external representation.

This hints at the idea that this interface should be split in two. In `ResultSetIterator` for instance, we only need the `fromResultSet` method.

persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql
binarycat0: Regarding the H2 doc - `INCLUDE clause may only be specified for UNIQUE indexes`
- https://www.h2database.com/html/commands.html#create_index

This subset of fields can be not unique `(realm_id, catalog_id, parent_id, type_code, sub_type_code)` and INCLUDE can not be applied so I put (id, name) in the main index.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/EntityNameLookupRecordConverter.java
binarycat0: This class is created because `org.apache.polaris.persistence.relational.jdbc.DatasourceOperations#executeSelectOverStream` requires passing an instance of `Converter`. You probably right that separating this Interface would be preferred, but it looks like out of scope of the current issue. So I propose to create an extra minor issue if we are interested in this action.

binarycat0: I will take a look closer, maybe it will be easier to implement a new Interface for this case as I did with `Identifiable`.

binarycat0: Moved into Follow-up issue as Out Of Scope
- https://github.com/apache/polaris/issues/2487

dimas-b:
Thanks for the perf. update @binarycat0 ! However, the execution times are too small and the sample size is small too (I'm afraid) so the perf. numbers are not decisive, IMHO :sweat_smile: 

At the same time the new index will have some storage and maintenance overhead.

Let's open a `dev` ML discussion about this and see what people who use Polaris with Postgres think.

The java code changes LGTM :+1:

dimas-b:
@binarycat0 : please remove `Database Index Optimization: Covering index enables index-only scans` from the description as the new index is no longer in this PR :sweat_smile:

Issue #2471, opened 2025-08-29
Ignore user-specified readiness checks


fivetran-kostaszoumpatianos:
### Is your feature request related to a problem? Please describe.

Users might want to explicitly disable readiness checks related to specific configuration parameters without disabling readiness checks altogether. Currently we don't have such an option.

### Describe the solution you'd like

I propose that we add a `polaris.readiness.ignore-offending-properties` config which can be, for example, set up as follows:
```
polaris.readiness.ignore-offending-properties=\
  polaris.metrics.user-principal-tag.enable-in-api-metrics,\
  polaris.features."ALLOW_INSECURE_STORAGE_TYPES",\
  polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"
```

all readiness checks that blame one of the specified properties will be suppressed. 

### Describe alternatives you've considered

We could have one config per readiness check but this is an overkill.

PR #2472, opened 2025-08-29
Introducing: polaris.readiness.ignore-offending-properties

fivetran-kostaszoumpatianos:
Fixes:  #2471 

This PR introduces a `polaris.readiness.ignore-offending-properties` config that accepts a map of properties for which readiness checks are suppressed.

It can be used, for example, as follows:
```
polaris.readiness.ignore-offending-properties=\
  polaris.metrics.user-principal-tag.enable-in-api-metrics,\
  polaris.features."ALLOW_INSECURE_STORAGE_TYPES",\
  polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES"
```

The check performed is case-insesitive.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
@@ -74,7 +74,14 @@ public class ProductionReadinessChecks {
       @Observes Startup event,
       Instance<ProductionReadinessCheck> checks,
       ReadinessConfiguration config) {
-    List<Error> errors = checks.stream().flatMap(check -> check.getErrors().stream()).toList();
+    List<Error> errors =
+        checks.stream()
+            .flatMap(check -> check.getErrors().stream())
+            .filter(
+                error ->
+                    config.ignoreOffendingProperties().stream()
+                        .noneMatch(prop -> prop.equalsIgnoreCase(error.offendingProperty())))
+            .toList();
     if (!errors.isEmpty()) {
       var utf8 = Charset.defaultCharset().equals(StandardCharsets.UTF_8);
       var warning = utf8 ? WARNING_SIGN_UTF_8 : WARNING_SIGN_PLAIN;
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ReadinessConfiguration.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ReadinessConfiguration.java
@@ -21,6 +21,7 @@ package org.apache.polaris.service.config;
 import io.quarkus.runtime.annotations.StaticInitSafe;
 import io.smallrye.config.ConfigMapping;
 import io.smallrye.config.WithDefault;
+import java.util.Set;
 
 @StaticInitSafe
 @ConfigMapping(prefix = "polaris.readiness")
@@ -33,4 +34,11 @@ public interface ReadinessConfiguration {
    */
   @WithDefault("false")
   boolean ignoreSevereIssues();
+
+  /**
+   * Set of properties that, if found to be misconfigured, will be ignored when determining the
+   * production readiness.
+   */
+  @WithDefault("{}")
+  Set<String> ignoreOffendingProperties();
 }


fivetran-kostaszoumpatianos:
@dimas-b @adutra @eric-maynard could you maybe take a look at this PR? thanks!

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -74,7 +74,14 @@ public void warnOnFailedChecks(
...
       Instance<ProductionReadinessCheck> checks,
       ReadinessConfiguration config) {
-    List<Error> errors = checks.stream().flatMap(check -> check.getErrors().stream()).toList();
+    List<Error> errors =
+        checks.stream()
+            .flatMap(check -> check.getErrors().stream())
+            .filter(
+                error ->
+                    config.ignoreOffendingProperties().stream()
+                        .noneMatch(prop -> prop.equalsIgnoreCase(error.offendingProperty())))
```
dimas-b, line 83: This approach LGTM in general. However, WDYT about adding a new "ID" property to `Error`, e.g. `error.getId()`.

The ID could be a static constant for cases where there could only ever be one `Error` instance per "check" code (e.g. `checkUserPrincipalMetricTag`) or it could be a deterministic (hash) function of the "type" plus some parameters (e.g. `checkInsecureStorageSettings` could produce IDs like `storage-17af38` and `storage-46fq98`).

The idea is that admin users should suppress specific error instances, but not "ranges" of errors. This way, if an admin user suppresses one particular check cases, new checks will still be visible when Polaris adds them.  The value of `error.offendingProperty()` may still be too broad in some cases.

The "hash" part being deterministic will allow admin users to propagate the same configuration to all their deployment environments. At the same time, it is not easy to guess, which will force the admin user to review what exactly needs to be suppressed. Also, if the meaning of the error changes, we can change the ID, and it will force the admin users to reassess the implications (and re-suppress).

WDYT?

runtime/service/src/main/java/org/apache/polaris/service/config/ReadinessConfiguration.java
```diff
@@ -33,4 +34,11 @@ public interface ReadinessConfiguration {
    */
   @WithDefault("false")
   boolean ignoreSevereIssues();
+
+  /**
+   * Set of properties that, if found to be misconfigured, will be ignored when determining the
+   * production readiness.
+   */
+  @WithDefault("{}")
+  Set<String> ignoreOffendingProperties();
```
eric-maynard, line 43: Why the asymmetry between this and `ignoreSevereIssues`? IIUC this is basically a subset of severe (?) issues that the admin wants to configure the readiness check to ignore. Maybe `ignoreSelectIssues`?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
eric-maynard: > The idea is that admin users should suppress specific error instances

What would this flow look like, though? Is this to support cases like `I want to allow setting config X, but not Y, and I want to allow setting config Z to A or B but not to C.`? I fear we are at risk of overengineering this a bit. As it is, only admins have access to these configs.

dimas-b: from my POV, this is not so much about config `A=X` or `A=Y`, but more about "Polaris detected something dangerous about X". Now, if the admin user suppresses this warning, I do not want the suppression to automatically hide future warnings about "dangerous Y".

It may be related to some specific config, but may be not. I can imagine running as the `root` OS user falls under the same category of auto-detectable issues.

runtime/service/src/main/java/org/apache/polaris/service/config/ReadinessConfiguration.java
fivetran-kostaszoumpatianos: Because technically the check is based on the offending property and not the issue type. `ignoreIssuesForSelectedOffendingProperties` ? maybe too much?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
fivetran-kostaszoumpatianos: Filtering based on error ids would also fit well with the naming change that you propose @eric-maynard then we can call the ignore method: `ignoreSelectedIssues` since we will now have a way of filtering by issue. Maybe the parameter hash is a bit too much. For me, I would be ok deactivating a check altogether if I know that I have a dangerous config there. It depends on how cautious we want to be.

dimas-b: I'd be ok without the parameter hash for a start. However, having a concise but non-predictable error ID is important I think. That is to say, a user suppressing a particular error must first observe the error. It should not be easy to suppress something "proactively" :) At the same time the error ID should not be dependent on the runtime env. (i.e. be the same in all k8s pods, for example). WDYT?

PR #2473, opened 2025-08-29
Service: Validate allowed locations from Catalog Storage Config

flyrain:
We will need to validate allowed locations from Catalog Storage Config in case of structured table location path.


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1006,51 +1004,27 @@ private void validateLocationsForTableLike(
...
+                    locations.stream()
+                        .filter(
+                            location -> location.startsWith("file:") || location.startsWith("http"))
+                        .collect(Collectors.toList());
+                if (!invalidLocations.isEmpty()) {
+                  throw new ForbiddenException(
+                      "Invalid locations '%s' for identifier '%s': File locations are not allowed",
+                      invalidLocations, identifier);
+                }
+              }
```
flyrain, line 1026: This part is unchanged.

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
```diff
@@ -0,0 +1,106 @@
...
+      List<String> allowedLocations) {
+    this(storageConfigurationInfo, allowedLocations, null);
+  }
+
+  public void validate(RealmConfig realmConfig, TableIdentifier identifier, Set<String> locations) {
+    if (parentLocation != null) {
+      validateLocations(realmConfig, List.of(parentLocation), locations, identifier);
+    }
+
+    validateLocations(realmConfig, allowedLocations, locations, identifier);
```
flyrain: Ensure to validate the allowed locations regardless of `ALLOW_UNSTRUCTURED_TABLE_LOCATION` is true or false

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -160,7 +161,7 @@ public static Optional<PolarisStorageConfigurationInfo> forEntityPath(
                 Set<String> locations =
                     StorageUtil.getLocationsAllowedToBeAccessed(
                         null, entityPathReversed.get(0).getPropertiesAsMap());
-                return new StorageConfigurationOverride(
+                return new LocationRestrictions(
```
dennishuo: I think the TODO on line 160 added in https://github.com/apache/polaris/pull/1320/files#diff-607cdd8c6ea78443988359e42ffec3091a4c84bc9151575772ee3b90164e03f3 was correctly pointing out that the addition of these locations from `StorageUtil.getLocationsAllowedToBeAccessed` (formerly `userSpecifiedWriteLocations`) msimatched the intent of the validation in ways that were problematic for Views.

We should remove all this location-augmentation here and simply return the basic restrictions directly from `configInfo` in this `else` branch.

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
```diff
@@ -0,0 +1,106 @@
...
+   * <p>All locations in this list have been validated to conform to the storage type's URI scheme
+   * requirements during construction.
+   */
+  private final List<String> allowedLocations;
+
+  /**
+   * The parent location for structured table enforcement.
+   *
+   * <p>When non-null, this location represents the root under which all new tables must be created,
+   * enforcing a structured hierarchy. When null, table creation is allowed anywhere within the
```
dennishuo: "enforcing a structured hierarchy _in addition to residing under (@code allowedLocations}_" 

to make it clear that the restrictions are `AND` instead of `OR`.

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
```diff
@@ -0,0 +1,106 @@
...
+   * {@code allowedLocations}.
+   */
+  private final String parentLocation;
+
+  public LocationRestrictions(
+      @Nonnull PolarisStorageConfigurationInfo storageConfigurationInfo,
+      List<String> allowedLocations,
```
dennishuo: I think we can remove the `allowedLocations` input here based on what I commente in `PolarisStorageConfigurationInfo`. The intended use of `storageConfigurationInfo.getAllowedLocations()` should be implicit.

If we *really* wanted to allow a UNION of `configInfo.getAllowedLocations()` with some additional `alternativeAllowedLocations` we should probably name it explicitly as such to make it very clear that we're changing the total set of allowedLocations rather than paring down a smaller subset of allowedLocations.

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
```diff
@@ -0,0 +1,106 @@
...
+    this.parentLocation = parentLocation;
+  }
+
+  public LocationRestrictions(
+      @Nonnull PolarisStorageConfigurationInfo storageConfigurationInfo,
+      List<String> allowedLocations) {
+    this(storageConfigurationInfo, allowedLocations, null);
+  }
+
+  public void validate(RealmConfig realmConfig, TableIdentifier identifier, Set<String> locations) {
```
dennishuo: Javadoc comments for this method would be nice. Especially for `locations`. We might want to rename `locations` to `requestedLocations` to make it clear what the role of that method parameter is.

Same for the `locations` param to `validateLocations` below.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -160,7 +161,7 @@ public static Optional<PolarisStorageConfigurationInfo> forEntityPath(
                 Set<String> locations =
                     StorageUtil.getLocationsAllowedToBeAccessed(
```
dennishuo: While we're at it, maybe we should rename all the `StorageUtil.getLocationsAllowedToBeAccessed` methods to make it clear that it's not really telling what's "allowed", but really what's *declared* to be used by the table.

Maybe rename it to `StorageUtil.getLocationsDeclaredForTable` or `StorageUtil.getLocationsRequestedForTable` or `StorageUtil.getAllLocationsUsedByTable` or something similar.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
flyrain: Test `createViewWithCustomMetadataLocationUsingPolaris()` failed once we remove it, which was introduced by #1320. Do we need to support view with custom location(`write.metadata.path`)? I don't think it's necessary.  We can probably revert all changes in #1320 if needed. Also, we can fix it in a followup PR. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
flyrain: Fixed it

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
flyrain: Same as https://github.com/apache/polaris/pull/2473#discussion_r2310619315

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
flyrain: Fixed it

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
eric-maynard: I don't think we need to worry too much about edge cases like a custom `write.metadata.path` for views. In fact, even for tables, we all know that tables can contain data files outside of `write.metadata.path` U `write.data.path` U `location` right? We need to capture the majority of use layouts with credential vending, but not necessarily the long tail.

dennishuo: Right, regardless of whether this allows or disallows setting `write.metadata.path` for Views, that test case is currently documenting the incorrect behavior that also applies to Tables.

Probably better to fix-forward the test; even #1320 itself indicated it was just documenting current observed behavior, but we can update the `createViewWithCustomMetadataLocationUsingPolaris` to ensure it also fails with ForbiddenException even if the custom path is in the parent.

In this case we also want to make sure even with `ALLOW_UNSTRUCTURED_TABLE_LOCATION=true` for the *CreateTable* scenario (not CreateView), the creation will fail if the "custom location" isn't part of the Catalog's `allowedLocations`.

dennishuo: Note, if we want a "success" case for Views calling `super.createViewWithCustomMetadataLocation();` we can just construct the original Catalog's StorageConfig to include the custom path in `allowedLocations`

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
flyrain: Fixed

polaris-core/src/main/java/org/apache/polaris/core/storage/LocationRestrictions.java
flyrain: Fixed

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
flyrain: Fixed

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewIntegrationBase.java
```diff
@@ -201,16 +201,10 @@ public void createViewWithCustomMetadataLocationUsingPolaris(@TempDir Path tempD
...
-    String customLocationChild =
-        Paths.get(tempDir.toUri().toString(), "custom-location/child").toString();
+    String customLocation =
+        Paths.get(storageConfig.getAllowedLocations().getFirst(), "/custom-location1").toString();
 
-    catalog()
-        .createNamespace(
-            identifier.namespace(),
-            ImmutableMap.of(
-                IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY, location));
```
flyrain, line 213: It doesn't make sense to have a `write.metadata.path` for namespace, also we won't check the locations used by parent after the change in method `forEntityPath`

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewIntegrationBase.java
```diff
@@ -201,16 +201,10 @@ public void createViewWithCustomMetadataLocationUsingPolaris(@TempDir Path tempD
     TableIdentifier identifier = TableIdentifier.of("ns", "view");
 
     String location = Paths.get(tempDir.toUri().toString()).toString();
```
flyrain, line 203: This location still escape from the storage config. We will need more fixes.

dennishuo (approved):
LGTM

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewIntegrationBase.java
dennishuo: The base class does add the `@TempDir` to allowedLocations in https://github.com/apache/polaris/blob/f555f30f6f6a10df441e8da84098c82245de28a9/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewFileIntegrationTestBase.java#L43

Is this a different tempdir that escapes? If we want to exercise a failure we probably need to try to set the `location` to something else.

flyrain: Yeah, they are different tmpDir, the paths generated are different. So the namespace location should not be allowed, not sure why it still passed. I will debug it more.

flyrain: Weird. It works as expected when I moved the integration test to unit test.

flyrain: It turns out the Polaris added a structured location to the view in https://github.com/apache/polaris/pull/1966. When a rest client creates view with location A, the Polaris will automatically adds the prefix like /root/namespace/A, related code https://github.com/apache/polaris/blob/17a359b43cf961a1261a7dfa7405ff47fb06b5f3/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1244. I don’t think this is correct, it breaks the contract between Iceberg REST clients and servers. The REST Server should only modify the location when it’s missing.

The unit tests I added don't have the issue as it doesn't go through that route.

flyrain: In short, the location checking is fine. I will merge this PR, and file another to fix the view issue.

PR #2474, opened 2025-08-29
Remove commons-codec dependency

XN137:
follow-up to f8ad77a6915defdcfd4bf778f8f6123042ded31e

we can simply use guava instead and eliminate the extra dependency

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/DigestUtils.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/DigestUtils.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core;
+
+import com.google.common.hash.Hashing;
+import java.nio.charset.StandardCharsets;
+
+public final class DigestUtils {
+  private DigestUtils() {
+    // utils class
+  }
+
+  public static String sha256Hex(String input) {
+    return Hashing.sha256().hashString(input, StandardCharsets.UTF_8).toString();
+  }
+}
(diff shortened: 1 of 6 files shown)


polaris-core/src/main/java/org/apache/polaris/core/DigestUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.common.hash.Hashing;
+import java.nio.charset.StandardCharsets;
+
+public final class DigestUtils {
+  private DigestUtils() {
+    // utils class
+  }
+
+  public static String sha256Hex(String input) {
+    return Hashing.sha256().hashString(input, StandardCharsets.UTF_8).toString();
```
XN137, line 30: some basic local testing to confirm this produces the same results as `org.apache.polaris.core.DigestUtils.sha256Hex`:
```
Input: 
guava -> e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
codec -> e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
Input: foobar
guava -> c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2
codec -> c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2
Input: ajsdljaksl-92301928i390!@#$%^&*()_+
guava -> ee7a434dd148156b40dd00ad567c37dea7297ebccb6df0f7a5be064c97dcb18b
codec -> ee7a434dd148156b40dd00ad567c37dea7297ebccb6df0f7a5be064c97dcb18b
Input: 😀
guava -> f0443a342c5ef54783a111b51ba56c938e474c32324d90c3a60c9c8e3a37e2d9
codec -> f0443a342c5ef54783a111b51ba56c938e474c32324d90c3a60c9c8e3a37e2d9
```

XN137:
thanks for the review

> I would suggest to update LICENSE to remove commons-codec mention there.

i think while this PR removes a direct dependency on the library we still depend on it transitively and thus it ends up in our binary distribution:
```
unzip -l ./runtime/distribution/build/distributions/polaris-bin-1.1.0-incubating-SNAPSHOT.zip | grep commons-codec
   373045  1980-02-01 00:00   polaris-bin-1.1.0-incubating-SNAPSHOT/admin/lib/main/commons-codec.commons-codec-1.18.0.jar
   373045  1980-02-01 00:00   polaris-bin-1.1.0-incubating-SNAPSHOT/server/lib/main/commons-codec.commons-codec-1.18.0.jar
```
doesnt that mean we have to keep it in the LICENSE file?

generally wondering whether we have plans to move towards automatic creation/update of that file?
i would assume/hope this is a solved problem in the gradle ecosystem :sweat_smile:

jbonofre:
Yes, as it's still shipped in the distribution, we have to keep in the `LICENSE`.

I will open another PR to simplify a bit `LICENSE`. That's not so easy to have automatic script for that (especially for `NOTICE`). As already said, it's more an assist: the contributor still has to check what is actually embedded, and the affected part of the `NOTICE`.

By the way, I have a script to check the versions/dependency in the distributions.

dimas-b (approved):

PR #2479, opened 2025-08-31
Add Events for Policy Service APIs

adnanhemani:
This PR adds the Events instrumentation for the Policy Service APIs, surrounding the default delegated call to the business logic APIs.


runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/CatalogPolicyEventServiceDelegator.java
```diff
@@ -132,14 +199,31 @@ public Response getApplicablePolicies(
...
+        new CatalogPolicyServiceEvents.AfterGetApplicablePoliciesEvent(
+            catalogName,
+            namespace,
+            targetName,
+            policyType,
+            (GetApplicablePoliciesResponse) resp.getEntity()));
+    return resp;
+  }
+
+  private String getCatalogFromPrefix(String prefix, RealmContext realmContext) {
```
adutra: nit: inline the method.

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogPolicyServiceEvents.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.polaris.service.types.UpdatePolicyRequest;
+
+/**
+ * Event records for Catalog Policy operations. Each operation has corresponding "Before" and
+ * "After" event records.
+ */
+public class CatalogPolicyServiceEvents {
+
+  // Policy CRUD Events
+  public record BeforeCreatePolicyEvent(
```
adutra, line 36: Sorry to insist on naming, but it's still inconsistent.

In #2482 we have `(Before|After)(Object)(Verb)Event`. E.g. `BeforeCatalogListEvent`.

Here we have `(Before|After)(Verb)(Object)Event`. E.g. `BeforeCreatePolicyEvent`.

(I find verb-object more natural than object-verb btw, but I'll defer the decision to native English speakers.)

runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/CatalogPolicyEventServiceDelegator.java
adnanhemani: Changed.

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogPolicyServiceEvents.java
adnanhemani: Thinking about it again, I think verb-object is better too. Let me keep these events as-is in that case and change the rest of the PRs.

HonahX (approved):
LGTM! Do we have plan to add tests that verify all the before and after are correctly called during each request?

adutra (dismissed):

adutra:
> Do we have plan to add tests that verify all the before and after are correctly called during each request?

+1 to the idea.

HonahX (approved):

PR #2480, opened 2025-08-31
Add Events for Iceberg REST APIs

adnanhemani:
This PR adds the Events instrumentation for the Iceberg REST Service APIs, surrounding the default delegated call to the business logic APIs.


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -103,8 +192,23 @@ public Response createTable(
       String accessDelegationMode,
       RealmContext realmContext,
       SecurityContext securityContext) {
```
adnanhemani, line 210: Remove instrumentation from https://github.com/apache/polaris/pull/1844 for this.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListener.java
```diff
@@ -55,4 +55,162 @@ public void onBeforeTaskAttempted(BeforeTaskAttemptedEvent event) {}
...
+
+  /** {@link IcebergRestCatalogEvents.BeforeUpdateNamespacePropertiesEvent} */
+  public void onBeforeUpdateNamespaceProperties(
+      IcebergRestCatalogEvents.BeforeUpdateNamespacePropertiesEvent event) {}
+
+  /** {@link IcebergRestCatalogEvents.AfterUpdateNamespacePropertiesEvent} */
+  public void onAfterUpdateNamespaceProperties(
+      IcebergRestCatalogEvents.AfterUpdateNamespacePropertiesEvent event) {}
+
+  // Iceberg REST Catalog Table Events
```
adutra: Minor nits:

- We already have some table events declared above, maybe we should group them.
- Naming inconsistency: `onBeforeCreateTable` uses infinitive + noun, whereas `onAfterTableCommited` uses noun + past participle.

runtime/service/src/main/java/org/apache/polaris/service/events/IcebergRestCatalogEvents.java
```diff
@@ -0,0 +1,179 @@
...
+import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
+import org.apache.polaris.service.types.CommitTableRequest;
+import org.apache.polaris.service.types.CommitViewRequest;
+import org.apache.polaris.service.types.NotificationRequest;
+
+/**
+ * Event records for Iceberg REST Catalog operations. Each operation has corresponding "Before" and
+ * "After" event records.
+ */
+public class IcebergRestCatalogEvents {
```
adutra, line 46: Shouldn't we move here the existing events? E.g. `AfterTableCommitedEvent` etc.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -777,8 +777,8 @@ public Response sendNotification(
         securityContext,
         prefix,
         catalog -> {
-          catalog.sendNotification(tableIdentifier, notificationRequest);
-          return Response.status(Response.Status.NO_CONTENT).build();
+          boolean notificationSent = catalog.sendNotification(tableIdentifier, notificationRequest);
```
adutra: Unrelated change?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -62,26 +124,45 @@ public Response listNamespaces(
       String parent,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    return delegate.listNamespaces(
-        prefix, pageToken, pageSize, parent, realmContext, securityContext);
+    polarisEventListener.onBeforeListNamespaces(new BeforeListNamespacesEvent(prefix, parent));
+    Response resp =
+        delegate.listNamespaces(prefix, pageToken, pageSize, parent, realmContext, securityContext);
+    polarisEventListener.onAfterListNamespaces(new AfterListNamespacesEvent(prefix, parent));
```
adutra: Sorry if this has been discussed already, but what is the plan for failed events, should we wrap the decorated call in a try-catch block and fire a failed event?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -44,14 +96,24 @@
...
       CreateNamespaceRequest createNamespaceRequest,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    return delegate.createNamespace(prefix, createNamespaceRequest, realmContext, securityContext);
+    polarisEventListener.onBeforeCreateNamespace(
+        new BeforeCreateNamespaceEvent(prefix, createNamespaceRequest));
+    Response resp =
+        delegate.createNamespace(prefix, createNamespaceRequest, realmContext, securityContext);
+    CreateNamespaceResponse createNamespaceResponse =
+        resp.readEntity(CreateNamespaceResponse.class);
```
adutra: `readEntity` is dangerous: it is supposed to read the response's input stream, and therefore should only be used on the client side, not the server side. Use `getEntity` instead.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestConfigurationEventServiceDelegator.java
```diff
@@ -25,19 +25,28 @@
...
+import org.apache.iceberg.rest.responses.ConfigResponse;
 import org.apache.polaris.core.context.RealmContext;
 import org.apache.polaris.service.catalog.api.IcebergRestConfigurationApiService;
+import org.apache.polaris.service.events.IcebergRestCatalogEvents;
+import org.apache.polaris.service.events.PolarisEventListener;
 
 @Decorator
 @Priority(1000)
 public class IcebergRestConfigurationEventServiceDelegator
     implements IcebergRestConfigurationApiService {
```
adutra, line 37: I noticed that you created a decorator for the config endpoint, but there is no decorator for the token endpoint (`IcebergRestOAuth2ApiService`). Is that on purpose?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
adnanhemani: No, at the moment we do not get whether the notification was sent successfully or not. So I've introduced this change so that we can surface that information. (There are code paths possible where the notificationSent can be false but the call still returns without error).

But to be fair, I'm not sure what notificationSent really helps with. So please let me know what you think.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: That is the plan for the future, yes. Well, use try-finally rather than catch.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestConfigurationEventServiceDelegator.java
adnanhemani: I purposefully left it alone as the `getToken` endpoint is deprecated for removal. Please let me know if you'd still like me to implement it regardless.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListener.java
adnanhemani: Good point, changed.

runtime/service/src/main/java/org/apache/polaris/service/events/IcebergRestCatalogEvents.java
adnanhemani: Agreed, included in the next revision!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: Thanks for pointing this out! Let me change this in the next revision.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
adutra: In this case this requires a modification of the OpenAPI spec. See my related comment:

https://github.com/apache/polaris/pull/2482/files#r2318079257

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -307,7 +519,18 @@ public Response sendNotification(
...
+    Response resp =
+        delegate.sendNotification(
+            prefix, namespace, table, notificationRequest, realmContext, securityContext);
+    polarisEventListener.onAfterSendNotification(
+        new AfterSendNotificationEvent(catalogName, namespace, table, (boolean) resp.getEntity()));
+    return resp;
+  }
+
+  private String getCatalogFromPrefix(String prefix, RealmContext realmContext) {
+    return prefixParser.prefixToCatalogName(realmContext, prefix);
```
adutra: nit: inline.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListener.java
adutra: Thanks, that's better! But we still have inconsistencies, see my comment:

https://github.com/apache/polaris/pull/2479#discussion_r2318169836

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
adutra: Also: `status(Response.Status.NO_CONTENT).entity(notificationSent)` is contradictory: the response for a 204 status should not contain any body:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListener.java
adnanhemani: Changed the remaining to be verb-object as well. Pleas let me know if something still stands out :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: Done as part of merge from main.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: Changed.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
adnanhemani: Needing an OpenAPI spec change and ML thread regarding this is a fair ask. Let me revert this out for now - maybe we can proceed without the result of the `sendNotification` call first and then in parallel, I can start the ML thread for this.

Once we reach a consensus as a community, then we can go back and resolve the TODO.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
```diff
@@ -20,61 +20,55 @@
...
-import org.apache.polaris.service.events.BeforeViewCommitedEvent;
-import org.apache.polaris.service.events.BeforeViewRefreshedEvent;
+import org.apache.polaris.service.events.AfterAttemptTaskEvent;
+import org.apache.polaris.service.events.BeforeAttemptTaskEvent;
+import org.apache.polaris.service.events.BeforeLimitRequestRateEvent;
+import org.apache.polaris.service.events.IcebergRestCatalogEvents;
 
 public abstract class PolarisPersistenceEventListener extends PolarisEventListener {
 
   // TODO: Ensure all events (except RateLimiter ones) call `addToBuffer`
```
adutra: Is the plan to tackle this later?

adutra: BTW the comment is outdated: 
```suggestion
  // TODO: Ensure all events (except RateLimiter ones) call `processEvent`
```

adutra (approved):

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
adnanhemani: Yes, the plan is to tackle in a separate PR - I would like to keep the focus of this PR on the events itself and the transformations will go in a further PR.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -273,10 +477,21 @@ public Response replaceView(
...
+            prefix, namespace, view, commitViewRequest, realmContext, securityContext);
+    polarisEventListener.onAfterReplaceView(
+        new AfterReplaceViewEvent(
+            catalogName, namespace, view, (LoadViewResponse) resp.getEntity()));
+    return resp;
   }
 
+  /**
+   * Table Committed Events are already instrumented at a more granular level than the API itself.
+   */
```
singhpk234: how and why , are it we tracking it per table ?? this seems like it doesn't follow the pattern 
-- BeforEvent 
-- do Something 
-- After Event

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -103,8 +202,29 @@ public Response createTable(
       String accessDelegationMode,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    return delegate.createTable(
-        prefix, namespace, createTableRequest, accessDelegationMode, realmContext, securityContext);
+    String catalogName = prefixParser.prefixToCatalogName(realmContext, prefix);
+    if (!createTableRequest.stageCreate()) {
```
singhpk234: why don't we emit anything for StageCreate ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -212,8 +384,12 @@ public Response listViews(
       Integer pageSize,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    return delegate.listViews(
-        prefix, namespace, pageToken, pageSize, realmContext, securityContext);
+    String catalogName = prefixParser.prefixToCatalogName(realmContext, prefix);
+    polarisEventListener.onBeforeListViews(new BeforeListViewsEvent(catalogName, namespace));
```
singhpk234: are the nested namespace encoded by '%1F' in the final even ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: The TableCommittedEvent was previously merged and there are no changes to it as part of this PR. The idea is that all table commits that happened as part of the same Polaris API call (through the `commitTransaction` API) will already be tied together through the same request ID and so a new Event for this API is not explicitly required.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: My understanding is that when a table is stageCreated, then no actual table is made - and therefore there is no new (meta)data created. In that case, I don't think it makes sense to emit a `AfterCreateTableEvent` if a table was not truly created.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: This is a good callout - we should switch this to use `Namespace` objects instead. Doing in the next revision.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
singhpk234: But we are not emitting a onBeforeCreateTable too ? never the less what happens when i stage create and then do `/commit` ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
singhpk234: what is the on before and on after for this, would looks though ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -262,7 +480,13 @@ public Response renameView(
       RenameTableRequest renameTableRequest,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    return delegate.renameView(prefix, renameTableRequest, realmContext, securityContext);
+    String catalogName = prefixParser.prefixToCatalogName(realmContext, prefix);
+    polarisEventListener.onBeforeRenameView(
+        new BeforeRenameViewEvent(catalogName, renameTableRequest));
```
singhpk234, line 483: no namespace here ? wouldn't it be nice to know namespace before after ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: I did not realize this - have removed the flag around BeforeCreateTableEvent. Consumers of this event should see this is for a `StageCreate` table and surface that information accordingly.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: On second thought, you are correct - the Table Committed events are triggered even for non-transactional events. Let me add this.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: `renameTableRequest` has the before and after table identifiers, where an event listener can access that the namespace information.

singhpk234 (approved):
The code changes LGTM from IRC POV ! Thanks @adnanhemani 

Orthogonal to the PR - I am a bit concerned with seeing the read events being instrumented too, as my understanding was we wanted write API's only (mirroring event's API), this increases load on persistence and the application in general since we use the same JDBC connection pool and we do batching  in memory and reads can shoot these up. 
My recommendation as part of follow-ups would be : 
* Mark this feature experimental, when ever the E2E machinery is ready
* run benchmarks with polaris benchmarking tool (whats could be the potential degradation) to identify bottelnecks.

adnanhemani:
Thanks @singhpk234 - makes sense. When we finish the E2E mechanism, we will surely either run the benchmark or mark as experimental.

singhpk234 (approved):

PR #2481, opened 2025-08-31
Add Events for Generic Table APIs

adnanhemani:
This PR adds the Events instrumentation for the Generic Tables Service APIs, surrounding the default delegated call to the business logic APIs.


runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
```diff
@@ -78,7 +108,19 @@ public Response loadGenericTable(
...
+    Response resp =
+        delegate.loadGenericTable(prefix, namespace, genericTable, realmContext, securityContext);
+    polarisEventListener.onAfterLoadGenericTable(
+        new CatalogGenericTableServiceEvents.AfterLoadGenericTableEvent(
+            catalogName, namespace, ((LoadGenericTableResponse) resp.getEntity()).getTable()));
+    return resp;
+  }
+
+  private String getCatalogFromPrefix(String prefix, RealmContext realmContext) {
+    return prefixParser.prefixToCatalogName(realmContext, prefix);
```
adutra: nit: inline (the method does nothing  more than `prefixToCatalogName`)

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogGenericTableServiceEvents.java
```diff
@@ -0,0 +1,47 @@
...
+ * under the License.
+ */
+
+package org.apache.polaris.service.events;
+
+import org.apache.polaris.service.types.CreateGenericTableRequest;
+import org.apache.polaris.service.types.GenericTable;
+
+public class CatalogGenericTableServiceEvents {
+  public record BeforeCreateGenericTableEvent(
```
adutra, line 26: Same naming inconsistency with #2482.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
adnanhemani: Changed.

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogGenericTableServiceEvents.java
adnanhemani: I've left these as-is and will change the remainder.

PR #2482, opened 2025-08-31
Add Events for PolarisServiceImpl APIs

adnanhemani:
This PR adds the Events instrumentation for all APIs listed in the PolarisServiceImpl.java file, surrounding the default delegated call to the business logic APIs.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
@@ -48,28 +48,35 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
   @Override
   public Response createCatalog(
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
-    polarisEventListener.onBeforeCatalogCreated(new CatalogsServiceEvents.BeforeCatalogCreatedEvent(request.getCatalog().getName()));
+    polarisEventListener.onBeforeCatalogCreated(
+        new CatalogsServiceEvents.BeforeCatalogCreatedEvent(request.getCatalog().getName()));
     Response resp = delegate.createCatalog(request, realmContext, securityContext);
-    polarisEventListener.onAfterCatalogCreated(new CatalogsServiceEvents.AfterCatalogCreatedEvent(resp.readEntity(Catalog.class)));
-    // If we are okay to start returning the catalog in the response, then we can simply return `resp`.
+    polarisEventListener.onAfterCatalogCreated(
+        new CatalogsServiceEvents.AfterCatalogCreatedEvent(resp.readEntity(Catalog.class)));
+    // If we are okay to start returning the catalog in the response, then we can simply return
+    // `resp`.
     return Response.status(Response.Status.CREATED).build();
   }
 
   @Override
   public Response deleteCatalog(
       String catalogName, RealmContext realmContext, SecurityContext securityContext) {
-    polarisEventListener.onBeforeCatalogDeleted(new CatalogsServiceEvents.BeforeCatalogDeletedEvent(catalogName));
+    polarisEventListener.onBeforeCatalogDeleted(
+        new CatalogsServiceEvents.BeforeCatalogDeletedEvent(catalogName));
     Response resp = delegate.deleteCatalog(catalogName, realmContext, securityContext);
-    polarisEventListener.onAfterCatalogDeleted(new CatalogsServiceEvents.AfterCatalogDeletedEvent(catalogName));
+    polarisEventListener.onAfterCatalogDeleted(
+        new CatalogsServiceEvents.AfterCatalogDeletedEvent(catalogName));
     return resp;
   }
 
   @Override
   public Response getCatalog(
       String catalogName, RealmContext realmContext, SecurityContext securityContext) {
-    polarisEventListener.onBeforeCatalogGet(new CatalogsServiceEvents.BeforeCatalogGetEvent(catalogName));
+    polarisEventListener.onBeforeCatalogGet(
+        new CatalogsServiceEvents.BeforeCatalogGetEvent(catalogName));
     Response resp = delegate.getCatalog(catalogName, realmContext, securityContext);
-    polarisEventListener.onAfterCatalogGet(new CatalogsServiceEvents.AfterCatalogGetEvent(resp.readEntity(Catalog.class)));
+    polarisEventListener.onAfterCatalogGet(
+        new CatalogsServiceEvents.AfterCatalogGetEvent(resp.readEntity(Catalog.class)));
     return resp;
   }
 
@@ -79,9 +86,12 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
       UpdateCatalogRequest updateRequest,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    polarisEventListener.onBeforeCatalogUpdated(new CatalogsServiceEvents.BeforeCatalogUpdatedEvent(catalogName, updateRequest));
-    Response resp = delegate.updateCatalog(catalogName, updateRequest, realmContext, securityContext);
-    polarisEventListener.onAfterCatalogUpdated(new CatalogsServiceEvents.AfterCatalogUpdatedEvent(resp.readEntity(Catalog.class)));
+    polarisEventListener.onBeforeCatalogUpdated(
+        new CatalogsServiceEvents.BeforeCatalogUpdatedEvent(catalogName, updateRequest));
+    Response resp =
+        delegate.updateCatalog(catalogName, updateRequest, realmContext, securityContext);
+    polarisEventListener.onAfterCatalogUpdated(
+        new CatalogsServiceEvents.AfterCatalogUpdatedEvent(resp.readEntity(Catalog.class)));
     return resp;
   }
 
@@ -99,9 +109,13 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
       CreateCatalogRoleRequest request,
       RealmContext realmContext,
       SecurityContext securityContext) {
-    polarisEventListener.onBeforeCatalogRoleCreate(new CatalogsServiceEvents.BeforeCatalogRoleCreateEvent(catalogName, request.getCatalogRole().getName()));
+    polarisEventListener.onBeforeCatalogRoleCreate(
+        new CatalogsServiceEvents.BeforeCatalogRoleCreateEvent(
+            catalogName, request.getCatalogRole().getName()));
     Response resp = delegate.createCatalogRole(catalogName, request, realmContext, securityContext);
-    polarisEventListener.onAfterCatalogRoleCreate(new CatalogsServiceEvents.AfterCatalogRoleCreateEvent(catalogName, resp.readEntity(CatalogRole.class)));
+    polarisEventListener.onAfterCatalogRoleCreate(
+        new CatalogsServiceEvents.AfterCatalogRoleCreateEvent(
+            catalogName, resp.readEntity(CatalogRole.class)));
     return resp;
   }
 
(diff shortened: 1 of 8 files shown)


runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -26,36 +26,58 @@
...
 @Decorator
 @Priority(1000)
 public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiService {
 
   @Inject @Delegate PolarisCatalogsApiService delegate;
+  @Inject PolarisEventListener polarisEventListener;
 
   @Override
   public Response createCatalog(
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
```
adnanhemani, line 50: Remove instrumentation from https://github.com/apache/polaris/pull/1844 for this.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -106,14 +157,25 @@ public Response updateCatalogRole(
...
+        new CatalogsServiceEvents.AfterCatalogRoleUpdateEvent(
+            catalogName, (CatalogRole) resp.getEntity()));
+    return resp;
   }
 
   @Override
   public Response listCatalogRoles(
       String catalogName, RealmContext realmContext, SecurityContext securityContext) {
-    return delegate.listCatalogRoles(catalogName, realmContext, securityContext);
+    polarisEventListener.onBeforeCatalogList(new CatalogsServiceEvents.BeforeCatalogListEvent());
```
adutra: Is this the right event? We are listing catalog roles, not catalogs.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -123,8 +185,25 @@ public Response addGrantToCatalogRole(
...
-        catalogName, catalogRoleName, grantRequest, realmContext, securityContext);
+    polarisEventListener.onBeforeAddGrantToCatalogRole(
+        new CatalogsServiceEvents.BeforeAddGrantToCatalogRoleEvent(
+            catalogName, catalogRoleName, grantRequest));
+    Response resp =
+        delegate.addGrantToCatalogRole(
+            catalogName, catalogRoleName, grantRequest, realmContext, securityContext);
+    PolarisServiceImpl.AddGrantToCatalogRoleEntityWrapper entityWrapper =
+        (PolarisServiceImpl.AddGrantToCatalogRoleEntityWrapper) resp.getEntity();
+    if (resp.getStatus() != Response.Status.BAD_REQUEST.getStatusCode()) {
```
adutra:
```suggestion
    if (resp.getStatus() == Response.Status.CREATED.getStatusCode()) {
```

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -630,6 +635,12 @@ public Response listCatalogRolesForPrincipalRole(
     return Response.ok(catalogRoles).build();
   }
 
+  record AddGrantToCatalogRoleEntityWrapper(
```
adutra: These records have to be public, as they are potentially escaping the package (IOW, a decorator sitting outside this package would be unable to process the response returned by `addGrantToCatalogRole`).

These records are OK for now, but we need to include these structures in the OpenAPI spec for this service. Could you please add a TODO for that, or better yet, file an issue? WIthout proper inclusion in the OpenAPI spec, these structures become a form of "undocumented" API.

It seems the problematic endpoints are:

* `createCatalog`
* `createPrincipalRole`
* `createCatalogRole`
* `addGrantToCatalogRole`
* `revokeGrantFromCatalogRole`

All of them would need to have their response schema updated in the OpenAPI spec.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -135,8 +214,22 @@ public Response revokeGrantFromCatalogRole(
...
+    PolarisServiceImpl.RevokeGrantFromCatalogRoleEntityWrapper entityWrapper =
+        (PolarisServiceImpl.RevokeGrantFromCatalogRoleEntityWrapper) resp.getEntity();
+    polarisEventListener.onAfterRevokeGrantFromCatalogRole(
+        new CatalogsServiceEvents.AfterRevokeGrantFromCatalogRoleEvent(
+            catalogName,
+            catalogRoleName,
+            entityWrapper.polarisPrivilege(),
+            entityWrapper.grantResource(),
+            entityWrapper.cascade()));
+    return resp;
```
adutra: For now, let's not include the entity in the response, since that would be a violation of the spec.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -397,7 +402,7 @@ public Response createPrincipalRole(
     PrincipalRole newPrincipalRole =
         new PrincipalRoleEntity(adminService.createPrincipalRole(entity)).asPrincipalRole();
     LOGGER.info("Created new principalRole {}", newPrincipalRole);
-    return Response.status(Response.Status.CREATED).build();
+    return Response.status(Response.Status.CREATED).entity(newPrincipalRole).build();
```
adutra: nit: we could standardize the usage of `toResponse` for all endpoints.

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogsServiceEvents.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.polaris.core.admin.model.Catalog;
+import org.apache.polaris.core.admin.model.CatalogRole;
+import org.apache.polaris.core.admin.model.GrantResource;
+import org.apache.polaris.core.admin.model.RevokeGrantRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRequest;
+import org.apache.polaris.core.admin.model.UpdateCatalogRoleRequest;
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+public class CatalogsServiceEvents {
+  public record BeforeCatalogCreatedEvent(String catalogName) implements PolarisEvent {}
```
adutra: nit: still some naming inconsistencies: most events use a verb in the infinitive form (`BeforeCatalogListEvent`) but some use the past participle (`BeforeCatalogCreatedEvent`).

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogsServiceEvents.java
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Event fired after a grant is added to a catalog role in Polaris.
+   *
+   * @param catalogName the name of the catalog
+   * @param catalogRoleName the name of the catalog role
+   * @param privilege the privilege granted
+   * @param grantResource the grant resource
+   */
+  public record AfterAddGrantToCatalogRoleEvent(
```
adutra, line 101: *Ideally*, if we eventually modify the OpenAPI spec, I think each "after" event should take a response object that mirrors the "before" event and its request object. E.g.:

```java
record BeforeAddGrantToCatalogRoleEvent(String, String, AddGrantRequest)
record  AfterAddGrantToCatalogRoleEvent(String, String, AddGrantResponse)
```

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adutra: In fact, this topic imo deserves a discussion thread in the ML.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adnanhemani: Changed in next revision.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adnanhemani: Good catch!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adnanhemani: Removed per comment below.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adnanhemani: Agreed, removing as per above comment.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adnanhemani: This makes sense, I will start a new ML discussion regarding this. In the meantime, let me add TODOs for this and remove it for now from this PR to unblock the rest of the events.

runtime/service/src/main/java/org/apache/polaris/service/events/CatalogsServiceEvents.java
adnanhemani: Changed!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adnanhemani: Not opposed to this, but let's do this in a different PR :)

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -651,13 +646,13 @@ public Response addGrantToCatalogRole(
...
     PrivilegeResult result;
+    PolarisPrivilege privilege;
     switch (grantRequest.getGrant()) {
       // The per-securable-type Privilege enums must be exact String match for a subset of all
       // PolarisPrivilege values.
       case ViewGrant viewGrant:
         {
-          PolarisPrivilege privilege =
-              PolarisPrivilege.valueOf(viewGrant.getPrivilege().toString());
+          privilege = PolarisPrivilege.valueOf(viewGrant.getPrivilege().toString());
```
adnanhemani, line 578: Not a requirement, but is a helpful refactor that will help in the future if/when we change the public API.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -165,7 +160,8 @@ private static Response toResponse(BaseResult result, Response.Status successSta
           .entity(icebergErrorResponse)
           .build();
     }
-    return Response.status(successStatus).build();
+    Response.ResponseBuilder responseBuilder = Response.status(successStatus);
+    return responseBuilder.build();
```
flyrain: This change seems unnecessary.

flyrain (commented):
Thanks @adnanhemani for the PR. LGTM overall. Do we have tests for added events(catalog, principal, principal role)?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adnanhemani: Good catch, this is a hangover from some other code I removed. Reverted.

adnanhemani:
@flyrain - I haven't added the tests yet, but will add it in a follow up PR. I didn't want to saddle this PR with the extra LOCs for something that's quite straightforward and requires much closer attention to the details of the events themselves.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
adutra: This doesn't look reverted yet 🤔

adnanhemani: Interesting, I don't know what happened and why it didn't pick up. Thanks for rechecking!

Made the changes now (and have verified)!

adutra (approved):

Issue #2464, opened 2025-08-28
Support Cockroach DB as Metastore


sathesuraj:
### Is your feature request related to a problem? Please describe.

Currently, Polaris does not support CockroachDB. It only supports PostgreSQL, even though CockroachDB is highly compatible with PostgreSQL.

### Describe the solution you'd like

CockroachDB uses INT8 (64-bit signed integer) as the default storage type for INT, INTEGER, and BIGINT data types. It does not support a 32-bit integer type (INT4) as a separate storage type. This design simplifies type handling and avoids integer overflow issues.

In Polaris, a JDBC connection is used to create metastore tables by executing SQL files. These files define types as INT and BIGINT. During the bootstrap process, when Polaris attempts to read INT columns from CockroachDB as the Integer Java datatype, it fails because CockroachDB stores these as INT8, which should be read as Long in Java.

One option would be to update sql scripts to use INT4 and INT8 instead of INT, it should work with both PostgreSQL and Cockroach DB.

flyrain:
Hi @sathesuraj, it's great idea to support Cockroach. Updating sql scripts itself only solve the part of the problem. The Polaris JDBC impl. still expects a 4 bytes integer. We will need to fix these as well. cc @singhpk234

singhpk234:
> The Polaris JDBC impl. still expects a 4 bytes integer. We will need to fix these as wel

precisely, have you tried this out with E2E tests (can try enabling it via overriding the profile [here](https://github.com/apache/polaris/pull/1371/files#diff-fde7952bb8b0cc8a6bc6959316008613eac974d0ef0afe0e55a51c1d010c3eb8)) if there are additional gotchas

sathesuraj:
@flyrain / @singhpk234  - I haven't tried using E2E tests yet.

In CockroachDB, executing the schema SQL results in all INT types being created and stored as INT8. However, if you explicitly specify the type as INT4 in the SQL, the column is saved with the type alias as INT4, though it is still stored internally as INT8.

To make it work with CockroachDB, I explicitly specified the type as INT4 in the SQL files and executed them before running the bootstrap process. This created an alias for those columns as INT4, even though they are internally stored as INT8. This approach allows the columns to be read as Integer in Java, enabling the use of CockroachDB as a metastore for Polaris.

For example, the types were updated for columns in the "entities" table were - entity_version, type_code, sub_type_code, and grant_record_version. 

No code or config change was required.

flyrain:
[...]

PR #2485, opened 2025-09-01
Use INT4 type alias for CockroachDB JDBC compatibility

sathesuraj:
Explicitly defined integer columns as INT4 in SQL script so that a CockroachDB JDBC driver correctly maps them to Java's Integer type, despite them being stored internally as INT8 in CockroachDB. This enables CockroachDB support for the metastore and will continue to work for PostgreSQL.

Fixes: apache/polaris#2464

### DIFF
--- a/persistence/relational-jdbc/src/main/resources/postgres/schema-v1.sql
+++ b/persistence/relational-jdbc/src/main/resources/postgres/schema-v1.sql
@@ -21,7 +21,7 @@ SET search_path TO POLARIS_SCHEMA;
 
 CREATE TABLE IF NOT EXISTS version (
     version_key TEXT PRIMARY KEY,
-    version_value INTEGER NOT NULL
+    version_value INT4 NOT NULL
 );
 INSERT INTO version (version_key, version_value)
 VALUES ('version', 1)
@@ -35,9 +35,9 @@ CREATE TABLE IF NOT EXISTS entities (
     id BIGINT NOT NULL,
     parent_id BIGINT NOT NULL,
     name TEXT NOT NULL,
-    entity_version INT NOT NULL,
-    type_code INT NOT NULL,
-    sub_type_code INT NOT NULL,
+    entity_version INT4 NOT NULL,
+    type_code INT4 NOT NULL,
+    sub_type_code INT4 NOT NULL,
     create_timestamp BIGINT NOT NULL,
     drop_timestamp BIGINT NOT NULL,
     purge_timestamp BIGINT NOT NULL,
@@ -45,7 +45,7 @@ CREATE TABLE IF NOT EXISTS entities (
     last_update_timestamp BIGINT NOT NULL,
     properties JSONB not null default '{}'::JSONB,
     internal_properties JSONB not null default '{}'::JSONB,
-    grant_records_version INT NOT NULL,
+    grant_records_version INT4 NOT NULL,
     PRIMARY KEY (realm_id, id),
     CONSTRAINT constraint_name UNIQUE (realm_id, catalog_id, parent_id, type_code, name)
 );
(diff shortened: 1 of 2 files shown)


persistence/relational-jdbc/src/main/resources/postgres/schema-v1.sql
```diff
@@ -21,7 +21,7 @@ SET search_path TO POLARIS_SCHEMA;
 
 CREATE TABLE IF NOT EXISTS version (
     version_key TEXT PRIMARY KEY,
-    version_value INTEGER NOT NULL
+    version_value INT4 NOT NULL
```
singhpk234, line 24: seems like cocroach-db uses pg driver as we have checks for the display type here 
https://github.com/apache/polaris/blob/main/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java#L46

would be nice to have a comment.

singhpk234 (commented):
Thank you for the change @sathesuraj !

Just to double sure so that we can call Polaris CockroachDB compatible, and protect Cocroach DB integration works in future relases (any regressions is caught), my **_suggestion_** would be to add eq class for [PostgresRelationalJdbcLifeCycleManagement](https://github.com/apache/polaris/blob/main/runtime/test-common/src/main/java/org/apache/polaris/test/commons/PostgresRelationalJdbcLifeCycleManagement.java#L28C14-L28C55) , mostly we will have to change the container to use cocroach container instead of pg 
can be done via
```
    return new CockroachContainer(
        dockerImage("cockroach").asCompatibleSubstituteFor("cockroachdb/cockroach"));
 ```
 may be we can refactor the base class a lit bit too, and the wiring of it to a profile would work mostly same.
 
Really appreciate you taking time to testing it cocroach and adding this capability to Polaris !

persistence/relational-jdbc/src/main/resources/postgres/schema-v1.sql
sathesuraj: Yes, cockroach DB uses pg driver. I'll add a comment for it.

dimas-b (commented):
Thanks for your contribution @sathesuraj !

@singhpk234 made a good point about testing. As far as I can see current JDBC persistence tests use only H2. It is great for light-weight unit tests, but does not provide assurance that the code + schema actually work with PostgreSQL or CockroachDB.

Now that we're starting to make non-trivial schema changes, it would be nice to make tests similar to `JdbcRestCatalogIT`  and `RelationalJdbcBootstrapCommandTest` but using actual PG and CockroachDB docker images. See `RelationalJdbcAdminProfile` for inspiration.

Ideally, it would be nice to add those tests in this PR, but if you do not have capacity ATM, I think it should be ok to merge these schema changes "as is" and add tests later.

dimas-b:
If we do not add tests in this PR, let's keep #2464 open until we have end-to-end tests.

flyrain:
This is a nice improvement. I'd love to see this to be merged. We may also need to check the specific SQL code of Cockroach DB. The current SQL codes are all from Postgres, https://github.com/polaris-catalog/polaris/blob/main/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java#L55-L55

sathesuraj:
> This is a nice improvement. I'd love to see this to be merged. We may also need to check the specific SQL code of Cockroach DB. The current SQL codes are all from Postgres, https://github.com/polaris-catalog/polaris/blob/main/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java#L55-L55

@flyrain  - Unfortunately, I could not spend time to add end-to-end tests for this change. I'll try to work on the comments and will update this PR.

flyrain:
@sathesuraj. I'm fine with adding test in another PR by other contributors if you don't have time. Thanks a lot for the contribution!

snazy:
I think Cockroach really deserves E2E tests.
While CR has is pg-compatible on the wire, it doesn't behave exactly like "the original PostgreSQL", which is especially true for reported SQL errors, which are IIRC inspected in the code base.

singhpk234:
superceded by https://github.com/apache/polaris/pull/3352

PR #2492, opened 2025-09-03
Add subtype-check to PolarisEntity subclass ctors

XN137:
this is a follow-up to ac31963c9eef12b2e1a1c7615d950939caf3b200

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/PolarisGenericTableCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/PolarisGenericTableCatalog.java
@@ -95,9 +95,7 @@ public class PolarisGenericTableCatalog implements GenericTableCatalog {
     PolarisResolvedPathWrapper resolvedEntities =
         resolvedEntityView.getPassthroughResolvedPath(
             tableIdentifier, PolarisEntityType.TABLE_LIKE, PolarisEntitySubType.ANY_SUBTYPE);
-    GenericTableEntity entity =
-        GenericTableEntity.of(
-            resolvedEntities == null ? null : resolvedEntities.getRawLeafEntity());
+    PolarisEntity entity = resolvedEntities == null ? null : resolvedEntities.getRawLeafEntity();
     if (null == entity) {
       entity =
           new GenericTableEntity.Builder(tableIdentifier, format)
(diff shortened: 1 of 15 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/PolarisGenericTableCatalog.java
```diff
@@ -95,9 +95,7 @@ public GenericTableEntity createGenericTable(
     PolarisResolvedPathWrapper resolvedEntities =
         resolvedEntityView.getPassthroughResolvedPath(
             tableIdentifier, PolarisEntityType.TABLE_LIKE, PolarisEntitySubType.ANY_SUBTYPE);
-    GenericTableEntity entity =
-        GenericTableEntity.of(
-            resolvedEntities == null ? null : resolvedEntities.getRawLeafEntity());
```
XN137, line 100: not every table-like entity we find here is a generic table

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1182,8 +1184,11 @@ private <T extends PolarisEntity & LocationBasedEntity> void validateNoLocationO
                     tbl -> {
                       PolarisResolvedPathWrapper resolveTablePath =
                           resolutionManifest.getResolvedPath(tbl);
-                      return IcebergTableLikeEntity.of(resolveTablePath.getRawLeafEntity())
-                          .getBaseLocation();
+                      PolarisEntity tableEntity = resolveTablePath.getRawLeafEntity();
+                      if (tableEntity.getSubType() == PolarisEntitySubType.GENERIC_TABLE) {
+                        return GenericTableEntity.of(tableEntity).getBaseLocation();
```
XN137, line 1269: since the `getBaseLocation()` impl is different between `GenericTableEntity` and `IcebergTableLikeEntity` this might have actually fixed a bug?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -2412,9 +2417,9 @@ private boolean sendNotificationForTableLike(
       if (null == entity) {
         existingLocation = null;
         entity =
-            new IcebergTableLikeEntity.Builder(tableIdentifier, newLocation)
+            new IcebergTableLikeEntity.Builder(
```
adutra, line 2511: How de we know it's a table, do notifications never get sent for views?

adutra (approved):

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
XN137: good question, in this change i am just following the already existing code.

this method seems questionable in that it has a `PolarisEntitySubType` parameter but also hardcodes `PolarisEntitySubType.ICEBERG_TABLE` in multiple spots.

i guess it does not matter since this method is currently only ever called from the `@Path("/namespaces/{namespace}/tables/{table}/notifications")` REST handler (there is no equivalent for views afaict)

adutra: Thanks for the clarification 👍

PR #2507, opened 2025-09-04
Fix deprecation warnings around RandomStringUtils

XN137:
the `RandomStringUtils.random` method is deprecated since commons-lang 3.17.0

https://github.com/apache/commons-lang/commit/69cb996265ba603e0be5b1c98097775f2467a6c2

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
@@ -136,6 +136,10 @@ public class PolarisManagementServiceIntegrationTest {
     client.cleanUp(authToken);
   }
 
+  private static String newRandomString(int length) {
+    return RandomStringUtils.insecure().next(length, true, true);
+  }
+
   @Test
   public void testCatalogSerializing() throws IOException {
     CatalogProperties props = new CatalogProperties("s3://my-old-bucket/path/to/data");
@@ -215,7 +219,7 @@ public class PolarisManagementServiceIntegrationTest {
             .setAllowedLocations(List.of("s3://my-old-bucket/path/to/data"))
             .build();
 
-    String goodName = RandomStringUtils.random(MAX_IDENTIFIER_LENGTH, true, true);
+    String goodName = newRandomString(MAX_IDENTIFIER_LENGTH);
 
     Catalog catalog =
         PolarisCatalog.builder()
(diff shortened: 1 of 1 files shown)


integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
```diff
@@ -136,6 +136,10 @@ public void tearDown() {
     client.cleanUp(authToken);
   }
 
+  private static String newRandomString(int length) {
+    return RandomStringUtils.insecure().next(length, true, true);
```
dimas-b, line 140: Replacing something "deprecated" with something "insecure" does not look very nice to me :sweat_smile: 

Also (semi-related), using unseeded random input in tests could make them non-deterministic.

Do we really need random inputs here?

My impression is that we need to ensure that certain chars are permitted or not and the length is restricted. Using deterministic test parameters is probably a more natural approach here.

I understand that this PR merely attempts to avoid deprecation, but since the related test code came to focus here, WDYT about making a more substantial change to make the test deterministic and improve coverage of input cases?

XN137: afaict it is named `insecure()` to make clear that the source of randomness is not fit for cryptographic purposes.
we can also switch to `secure()` instead if you prefer?

> WDYT about making a more substantial change to make the test deterministic and improve coverage of input cases

we can do that in a followup PR imo, but if we want to improve the coverage compared to randomized input you need to give concrete examples on what inputs you think would "improve" that coverage as for me its not clear.
we can of course make the test inputs deterministic (by picking one randomized sample?) but i dont think this would improve coverage.

dimas-b: `insecure()` is technically acceptable for current PR, I think. These names do not have to be cryptographically secure.

dimas-b: I can only guess what the intention behind the randomized version might have been.

I'd start with names containing ASCII upper/lower case letters and numbers having the the max allowed name length. I think that one case is sufficient for now.

Replacing random name with a fixes name as noted above is not going to reduce the tests' validity, but will improve determinism.

If we want to test other edge cases, it might be preferable to do so in a unit test targeting the validation code (if such validation code exists).

PR #2508, opened 2025-09-04
Add loadEntities batch call and rename listFullEntities

collado-mike:
https://github.com/apache/polaris/pull/2290 introduced a new `loadEntities` variant, which is really a `listEntities` call that returns the complete `PolarisBaseEntity` rather than the `EntityNameLookupRecord`. A batch `loadEntities` call that functions similar to the `loadEntity`, when given an id, returns the identified entity, is also useful, notably for cases when you don't want to list all entities of a particular type (e.g., loading a set of Principal Roles or refreshing specific entities for the `EntityCache`).

This introduces a new `loadResolvedEntities` API and renames the previous `loadEntities` to `listFullEntities` to avoid ambiguity. The new API now mirrors the `loadResolvedEntity...`APIs, accepting either a list of `PolarisEntityId`s or `EntityNameLookupRecord`s. I used `EntityNameLookupRecord` because that is the return type for the original `listEntities` API, but also because `PolarisEntityCore` requires a `grantVersion`, which may not be present, e.g., if the caller only has the results of a `listEntities` call. I also wanted to mirror the existing `loadEntity` API, which requires an `PolarisEntityType` argument and the `PolarisEntityId` doesn't contain a type field.

I used the `ResolvedPolarisEntity` type and terminology in the API name in order to make the EntityCache API and the raw PolarisMetaStoreManager API the same. In part, this aims to start bringing the two APIs closer together so that the concept of the cache can one day be just an implementation detail, rather than part of the core business logic. The bulk load implementation in the cache mirrors the logic in the Resolver, in that it ensures that it always returns a snapshot consistent with the state of the persistence layer as it exists at a single point in time. This means that it validates that the entire batch of entities returned matches the entity versions and grant versions returned by a call to the `loadEntitiesChangeTracking` API.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -18,9 +18,25 @@
  */
 package org.apache.polaris.persistence.relational.jdbc;
 
+import static org.apache.polaris.persistence.relational.jdbc.QueryGenerator.PreparedQuery;
+
 import com.google.common.base.Preconditions;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.entity.EntityNameLookupRecord;
@@ -61,23 +77,6 @@ import org.apache.polaris.persistence.relational.jdbc.models.ModelPrincipalAuthe
 import org.apache.polaris.persistence.relational.jdbc.models.SchemaVersion;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.sql.Connection;
-import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-
-import static org.apache.polaris.persistence.relational.jdbc.QueryGenerator.PreparedQuery;
 
 public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPersistence {
 
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
@@ -2314,26 +2314,26 @@ public class TransactionalMetaStoreManagerImpl extends BaseMetaStoreManager {
                     return entities.get(i);
                   }
                 })
-            .map(
-                e -> {
-                  if (e == null) {
-                    return null;
-                  } else {
-                    // load the grant records
-                    final List<PolarisGrantRecord> grantRecordsAsSecurable =
-                        ms.loadAllGrantRecordsOnSecurableInCurrentTxn(
-                            callCtx, e.getCatalogId(), e.getId());
-                    final List<PolarisGrantRecord> grantRecordsAsGrantee =
-                        e.getType().isGrantee()
-                            ? ms.loadAllGrantRecordsOnGranteeInCurrentTxn(
-                                callCtx, e.getCatalogId(), e.getId())
-                            : List.of();
-                    return new ResolvedPolarisEntity(
-                        PolarisEntity.of(e), grantRecordsAsGrantee, grantRecordsAsSecurable);
-                  }
-                })
+            .map(e -> toResolvedPolarisEntity(callCtx, e, ms))
             .collect(Collectors.toList());
     return new ResolvedEntitiesResult(ret);
+  }
+
+  private static ResolvedPolarisEntity toResolvedPolarisEntity(
+      PolarisCallContext callCtx, PolarisBaseEntity e, TransactionalPersistence ms) {
+    if (e == null) {
+      return null;
+    } else {
+      // load the grant records
+      final List<PolarisGrantRecord> grantRecordsAsSecurable =
+          ms.loadAllGrantRecordsOnSecurableInCurrentTxn(callCtx, e.getCatalogId(), e.getId());
+      final List<PolarisGrantRecord> grantRecordsAsGrantee =
+          e.getType().isGrantee()
+              ? ms.loadAllGrantRecordsOnSecurableInCurrentTxn(callCtx, e.getCatalogId(), e.getId())
+              : List.of();
+      return new ResolvedPolarisEntity(
+          PolarisEntity.of(e), grantRecordsAsGrantee, grantRecordsAsSecurable);
+    }
   }
 
   @Nonnull
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/cache/InMemoryEntityCache.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/cache/InMemoryEntityCache.java
@@ -23,6 +23,18 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.RemovalListener;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.config.BehaviorChangeConfiguration;
@@ -40,19 +52,6 @@ import org.apache.polaris.core.persistence.dao.entity.ResolvedEntitiesResult;
 import org.apache.polaris.core.persistence.dao.entity.ResolvedEntityResult;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.AbstractMap;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
 /** An in-memory entity cache with a limit of 100k entities and a 1h TTL. */
 public class InMemoryEntityCache implements EntityCache {


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -456,7 +457,12 @@ public List<PolarisBaseEntity> lookupEntities(
     if (entityIds == null || entityIds.isEmpty()) return new ArrayList<>();
     PreparedQuery query = QueryGenerator.generateSelectQueryWithEntityIds(realmId, entityIds);
     try {
-      return datasourceOperations.executeSelect(query, new ModelEntity());
+      Map<PolarisEntityId, PolarisBaseEntity> idMap =
+          datasourceOperations.executeSelect(query, new ModelEntity()).stream()
+              .collect(
+                  Collectors.toMap(
+                      e -> new PolarisEntityId(e.getCatalogId(), e.getId()), Function.identity()));
+      return entityIds.stream().map(idMap::get).collect(Collectors.toList());
```
dimas-b, line 472: Since we're obviously enforcing the order of the returned values to match the order of requested IDs, could you add a test for that?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
```diff
@@ -339,10 +339,7 @@ public boolean equals(Object o) {
     PolarisBaseEntity that = (PolarisBaseEntity) o;
     return subTypeCode == that.subTypeCode
         && createTimestamp == that.createTimestamp
-        && dropTimestamp == that.dropTimestamp
-        && purgeTimestamp == that.purgeTimestamp
```
dimas-b, line 343: Is this change related to the stated purpose of this PR?

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -1531,14 +1532,42 @@ private void revokeGrantRecord(
...
+    // mimic the behavior of loadEntity above, return null if not found or type mismatch
+    List<PolarisBaseEntity> ret =
+        IntStream.range(0, entityLookupRecords.size())
+            .mapToObj(
+                i -> {
+                  if (entities.get(i) != null
+                      && !entities.get(i).getType().equals(entityLookupRecords.get(i).getType())) {
+                    return null;
+                  } else {
+                    return entities.get(i);
```
dimas-b: Rather than assuming fast access by position in the `entities` list, would it be simpler and more generalizable to iterate `entities` and `entityLookupRecords` in parallel?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: I mean a test case that deliberately requests loadEntities with IDs ordered differently.

dimas-b: nevermind, I think existing tests are good enough :)

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -164,12 +165,13 @@ Page<PolarisBaseEntity> loadEntities(
    * @param entitySubType subType of entities to list (or ANY_SUBTYPE)
    * @return list of all matching entities
    */
-  default @Nonnull List<PolarisBaseEntity> loadEntitiesAll(
+  default @Nonnull List<PolarisBaseEntity> loadFullEntitiesAll(
```
dennishuo: Should this be `listFullEntitiesAll` instead of `loadFullEntitiesAll`?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
collado-mike: It was necessary for some test `equals()` comparisons to pass

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -346,6 +348,21 @@ EntityResult loadEntity(
...
+   * if the input list is empty. Order in that returned list is the same as the input list. Some
+   * elements might be NULL if the entity has been dropped.
+   *
+   * @param callCtx call context
+   * @param entityLookupRecords the list of entity lookup records to load
+   * @return a non-null list of entities corresponding to the lookup keys. Some elements might be
+   *     NULL if the entity has been dropped.
+   */
+  @Nonnull
+  EntitiesResult loadEntities(
```
dennishuo: For the current main EntityCache architecture, generally anything that falls into a "point lookup" is normally supposed to be cache-cooperative, while "list" is fundamentally passthrough because the set of listed contents aren't part of the entityVersion-controlled set (i.e. creation or modification of child entities don't increment entityVersions on the parent entity).

In this way, we probably want "batch point-lookup" to align better with Resolver and EntityCache, and make this `loadResolvedEntities` returning a collection of `ResolvedEntities` instead of a collection of only the naked `PolariseBaseEntity`.

Being able to resolve the grants all at once as well would be useful for using this in the PrincipalRoles batch-lookup.

I'd be okay with having boolean input options to control whether the caller actually wants to resolve grants as well in case we're concerned about doing wasted work when callsites don't need the grants, so this method doesn't necessarily need to be more expensive than only looking up naked entities.

But aligning the interface with ResolvedEntities could be a good step towards making the Resolver/EntityCache interfaces less confusing and a better match with the MetaStoreManager interface.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
dimas-b: I do not have a strong objection, but the old equals code appears to treat these objects as "value" objects, meaning and difference in any attribute would lead to objects being different. That's just my observation on the code local to this class.

That said, I do not really think these objects are used as value objects by Polaris code other than the related tests, so the change is probably ok.

On the other hand, do tests have to compare objects in "whole"? Could ID comparison be sufficient?

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
collado-mike: I like this option - happy to move forward with this approach. If we mirror the cache APIs, do you think `EntityNameLookupRecord` still makes sense as an input type? E.g., `loadEntitiesChangeTracking` takes `PolarisEntityId` as an input type. We could introduce a similar type to allow batch load by name

polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
collado-mike: When I was deduping this code to support both load by id and by lookup record, I ended up keeping the access by index so I could use a function to get the type of the entity. I generally think it's reasonable to expect access by index for List implementations in practical use

dimas-b: Fair enough :+1:

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -416,6 +419,38 @@ ResolvedEntityResult loadResolvedEntityByName(
...
+   * empty list if the input list is empty. Order in that returned list is the same as the input
+   * list. Some elements might be NULL if the entity has been dropped.
+   *
+   * @param callCtx call context
+   * @param entityLookupRecords the list of entity lookup records to load
+   * @return a non-null list of entities corresponding to the lookup keys. Some elements might be
+   *     NULL if the entity has been dropped.
+   */
+  @Nonnull
+  ResolvedEntitiesResult loadResolvedEntities(
```
dennishuo: The existence of this method may cause confusion since the single-lookup methods have both `ByName` and `ById` variations, and an `EntityNameLookupRecord` happens to have the name in it, and yet it looks like the actual key difference between the two methods is whether we have `EntityType` per item.

I'm not sure if we already use `EntityNameLookupRecord` as an *input* argument anywhere else, but generally since it was kind of structured as an *output* argument before it seems it becomes ambiguous as an input argument.

And at first glance the unittests seems to imply that the lookup would be "by name":

    @ParameterizedTest
    @ValueSource(strings = {"id", "name"})
      ....

        if (loadType.equals("id")) {
          // Create entity ID list with the updated entity
          List<PolarisEntityId> entityIds = List.of(getPolarisEntityId(T6v2));
    
          // Call batch load - this should detect the stale version and reload
          results =
              cache.getOrLoadResolvedEntities(this.callCtx, PolarisEntityType.TABLE_LIKE, entityIds);
        } else {
          results =
              cache.getOrLoadResolvedEntities(this.callCtx, List.of(new EntityNameLookupRecord(T6v2)));
        }


To match convention with the single-item lookups can we rename these methods to say `loadResolvedEntitiesById`?

And if the difference is really just whether we pass in a per-entity EntityType, I think even parallel Lists (`List<EntityId>, List<EntityType>` would be better than reusing `EntityNameLookupRecord` just for its `catalogId, entityId, entityType`.

Alternatively, cleanest would be just having one interface entirely with `List<EntityIdAndType>` as the input argument.

dennishuo: Another thing to consider is whether we actually want to allow different entity type lookups within a single batch. It may change the atomicity semantics for persistence implementations where different types are in different atomicity domains.

Are there any callsites that actually rely on using this form of the method instead of the one with a single entityType across the whole list of EntityIds?

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -2307,6 +2310,73 @@ private PolarisEntityResolver resolveSecurableToRoleGrant(
...
+                  } else {
+                    return entities.get(i);
+                  }
+                })
+            .map(
+                e -> {
+                  if (e == null) {
+                    return null;
+                  } else {
+                    // load the grant records
```
dimas-b: nit: maybe add `toResolvedPolarisEntity()` as in `AtomicOperationMetaStoreManager`?

dimas-b (commented):
I do not have any concerns with the current state of this PR, but I'd be interested in reviewing again after comments from @dennishuo are resolved :)

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
collado-mike: TBH, I can't think of a use case where we would mix `EntityType`s in a single call. The two immediate use cases I have in mind are 

1. Batch loading the principal roles during the authentication step
2. Support for loading TableMetadata from persistence rather than from cloud storage (this is in concert with https://github.com/apache/polaris/pull/2735 and other future PRs). 

In both cases, only a single EntityType is loaded. I used the `EntityNameLookupRecord` type largely because it is the return type for the `listEntities` API, but I wanted to avoid fetching all entities in full in the case that many/most entities are already in cache. Personally, I don't like the pattern of using parallel list parameters for an API, so I would oppose the `List<EntityId>, List<EntityType>` option. I am ok with a new `EntityIdAndType` argument, but I'd also be ok with just supporting the one API that takes in the `EntityType` and `List<EntityId>` arguments and getting rid of the other option until a need arises.

dennishuo: Yeah sounds good, I think it's good to have the more opinionated "all entities in the batch are the same EntityType" method signature for now, as it's easier to ensure different impls can fulfill the interface. Signature `(EntityType, List<EntityId>)` looks good to me, and let's remove the overloaded method regarding `EntityNameLookupRecord`.

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -416,6 +418,23 @@ ResolvedEntityResult loadResolvedEntityByName(
...
+   * list is the same as the input list. Some elements might be NULL if the entity has been dropped.
+   *
+   * @param callCtx call context
+   * @param entityType the type of entities to load
+   * @param entityIds the list of entity ids to load
+   * @return a non-null list of entities corresponding to the lookup keys. Some elements might be
+   *     NULL if the entity has been dropped.
+   */
+  @Nonnull
+  ResolvedEntitiesResult loadResolvedEntities(
```
dimas-b, line 433: The name LGTM, but I guess it does not match the PR description anymore 🤔

dennishuo (approved):
LGTM

dimas-b (approved):

dennishuo (approved):

PR #2511, opened 2025-09-05
Fix failing tests in IcebergAllowedLocationTest

eric-maynard:
This test, added in #2473, fails after #2422. This fixes the config override.


dimas-b (approved):
Thanks for the quick fix!

flyrain (approved):

PR #2518, opened 2025-09-08
Use PrincipalEntity in PolarisMetaStoreManager.createPrincipal

XN137:
if we pass and return a more specific `PrincipalEntity` we can simplify the surrounding code and implementation

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/CreatePrincipalResult.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/CreatePrincipalResult.java
@@ -20,7 +20,6 @@ package org.apache.polaris.core.persistence.dao.entity;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.base.Preconditions;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
@@ -45,7 +44,6 @@ public class CreatePrincipalResult extends BaseResult {
     super(errorCode, extraInformation);
     this.principal = null;
     this.principalSecrets = null;
-    validate();
   }
 
   /**
@@ -59,7 +57,6 @@ public class CreatePrincipalResult extends BaseResult {
     super(ReturnStatus.SUCCESS);
     this.principal = principal;
     this.principalSecrets = principalSecrets;
-    validate();
   }
 
   @JsonCreator
@@ -71,18 +68,6 @@ public class CreatePrincipalResult extends BaseResult {
     super(returnStatus, extraInformation);
     this.principal = principal;
     this.principalSecrets = principalSecrets;
-    validate();
-  }
-
-  private void validate() {
-    if (getReturnStatus() == ReturnStatus.SUCCESS) {
-      Preconditions.checkNotNull(principal);
-      Preconditions.checkNotNull(principalSecrets);
-      Preconditions.checkState(getPrincipal() != null, "Entity is not a Principal");
-    } else {
-      Preconditions.checkState(principal == null);
-      Preconditions.checkState(principalSecrets == null);
-    }
   }
 
   public PrincipalEntity getPrincipal() {


polaris-core/src/main/java/org/apache/polaris/core/persistence/AtomicOperationMetaStoreManager.java
```diff
@@ -577,28 +575,22 @@ private void revokeGrantRecord(
...
-            PolarisEntitySubType.NULL_SUBTYPE,
-            PolarisEntityConstants.getRootEntityId(),
-            PolarisEntityConstants.getRootPrincipalName());
-
-    // create this principal
+    PrincipalEntity rootPrincipal =
+        new PrincipalEntity.Builder()
+            .setId(rootPrincipalId)
+            .setName(PolarisEntityConstants.getRootPrincipalName())
+            .setCreateTimestamp(System.currentTimeMillis())
```
XN137, line 582: this looks odd as we are not using the `clock` field but using the clock field causes some test to fail, so i think we need to followup where we overhaul things to prefer `Clock` everywhere.

for now this is simply keeping the existing logic from the `PolarisBaseEntity` ctor:
https://github.com/apache/polaris/blob/7af85be7f45c933a377314a669e2a16633c93532/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java#L244-L257

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -194,7 +194,7 @@ Page<PolarisBaseEntity> loadEntities(
    */
   @Nonnull
   CreatePrincipalResult createPrincipal(
-      @Nonnull PolarisCallContext callCtx, @Nonnull PolarisBaseEntity principal);
+      @Nonnull PolarisCallContext callCtx, @Nonnull PrincipalEntity principal);
```
sfc-gh-ygu, line 197: Hi @XN137 , thanks for the refactor,  the changes LGTM overall. However, this interface change impacts all downstream projects. Can you start a dev mail thread for awareness and potential deprecation process?

flyrain, line 197: Hi [@XN137](https://github.com/XN137) , thanks for the refactor, the changes LGTM overall. However, this interface change impacts all downstream projects. Can you start a dev mail thread for awareness and potential deprecation process?

dimas-b: +1 to `dev` discussion.

I agree that this change can affect downstream projects. Still, according to the current [Polaris Evolution](https://polaris.apache.org/releases/1.0.1/evolution/#using-polaris-as-a-build-time-dependency) approach, this kind of java class change is not considered an "API" change from the SemVer perspective.

XN137: sorry i was out for 1,5 weeks.
started the ML discussion here:
https://lists.apache.org/thread/cflfhx59942kx29mtmthns5oo12dyzdv

wondering: what kind of timeframe do we need to wait for incoming opinions?

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
dimas-b: Thanks for  the email, @XN137!

Given that the PR has been in review for 3 weeks, I think waiting for a couple of days after the `dev` email should be enough to get feedback from people who might have missed the GH notifications.

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/CreatePrincipalResult.java
```diff
@@ -67,10 +71,22 @@ private CreatePrincipalResult(
     super(returnStatus, extraInformation);
     this.principal = principal;
     this.principalSecrets = principalSecrets;
+    validate();
```
sfc-gh-mcollado: I'm quite hesitant to add this `validate` function here, given that the constructor is annotated to be a JSON deserializer. If the JSON structure doesn't adhere to the requirements (and, as it could have been generated by an older or newer version of the software, it's entirely plausible), we may throw an exception during response serialization rather than allowing the caller to validate the state of the response. 

I'd say just remove this method from this constructor, but it seems that this is really the _only_ constructor that could possibly need this validation, as the other two never support a non-successful state with an entity or a successful state with no entity.

sfc-gh-mcollado (changes requested):
I like the change overall, but the validation in the JSON constructor is a pretty major behavior change and potentially breaks forward/backward compatibility with systems that are serializing these responses.

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/CreatePrincipalResult.java
dimas-b: From my POV, it is fair to remove the `validate()` call from this constructor.

I wonder, though, why is it a `@JsonCreator` at all?.. I do not see how these objects could be instantiated from JSON 🤔 Did I miss something? 😅

XN137: i am in a similar boat as Dmitri as i dont immediately see how/where the JSON de-serialization is actually getting used.
i also kept the `principal` field as a `PolarisBaseEntity` so that we dont hard-break the ser/deser logic of the class.

but for me the `validate()` method enforced the "invariant" of this class (i.e. what is a valid `CreatePrincipalResult` instance).
correct me if i am wrong but afaict this has not really changed since the original addition of the class:
https://github.com/apache/polaris/blame/4b24da5ef9e3e6093fc8f17cefeefceba11e1783/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java#L289-L340
so in that sense we should be fine about backwards-compatibility.

if a system uses the json deserialization, wouldnt it want to validate that the given input translates to a valid `CreatePrincipalResult` instance?
whats the benefit of being able to create an "invalid" `CreatePrincipalResult` instance?
imo such a system should have explicit error handling about its de-serialization usage (i.e. wrap instance creation in a try-catch?)

XN137: OK i've now removed the `validate()` method as it mostly made sense to get called in this json ctor imo.

note that due the change in `getPrincipal` the returned entity now passes through validation of `PrincipalEntity.of` so if there is an error about the entity being invalid it just got moved from the ctor to the `getPrincipal` method.

collado-mike: Thanks. This did come up some time ago in a mailing list thread and in a community discussion - all of the `PolarisMetaStoreManager` return types are annotated with Jackson annotations to support HTTP-based implementations of the `PolarisMetaStoreManager` interface. I did promise to submit an implementation so that we can run tests that validate everything still works for distributed environments, but... I haven't yet... 😅

PR #2531, opened 2025-09-10
Fix arg-matching in ExceptionMapperTest.testFullExceptionIsLogged()

snazy:
The above test asserts the the log record passed to `org.jboss.logmanager.Logger#logRaw(org.jboss.logmanager.ExtLogRecord)`. For the sake of this test, both the `IcebergJsonProcessingExceptionMapper` and `IcebergExceptionMapper` have a visible-for-testing function `getLogger()` to provide a mocked logger. `getLogger()` however is used for "all the logging", which makes the behavior `verify(JBOSS_LOGGER).logRaw(argThat(...))` "interesting" - the test fails with #2461 although no related dependencies have changed in that PR.

But eventually the test failure's caused by using `getLogger()` for all logging, which may provide the _wrong_ log record and cause the test to fail.

This change renames the `getLogger()` functions to make their special meaning clear, and adds some clarifying comments in the code.

Also adding some more test cases for the various exception mapper code paths.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
@@ -263,8 +263,8 @@ public class IcebergExceptionMapper implements ExceptionMapper<RuntimeException>
   }
 
   /**
-   * This function is only present for the {@code ExceptionMapperTest} and must only be used during
-   * once any execution of {@link #toResponse(RuntimeException)}.
+   * This function is only present for the {@code ExceptionMapperTest} and must only be used once
+   * during any execution of {@link #toResponse(RuntimeException)}.
    */
   @VisibleForTesting
   Logger getLoggerForExceptionLogging() {


runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
```diff
@@ -261,8 +262,12 @@ static Optional<Integer> mapCloudExceptionToResponseCode(Throwable t) {
     return Optional.of(Status.INTERNAL_SERVER_ERROR.getStatusCode());
   }
 
+  /**
+   * This function is only present for the {@code ExceptionMapperTest} and must only be used during
+   * once any execution of {@link #toResponse(RuntimeException)}.
```
adutra:
```suggestion
   * This function is only present for the {@code ExceptionMapperTest} and must only be used once
   * during any execution of {@link #toResponse(RuntimeException)}.
```

adutra (approved):

dimas-b (approved):

PR #2533, opened 2025-09-10
Inject PolarisAdminService into PolarisServiceImpl

XN137:
`PolarisServiceImpl` already is a request-scoped bean. if we apply the same to `PolarisAdminService` we can simply inject it into `PolarisServiceImpl`.


polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecretsManager.java
```diff
@@ -21,7 +21,6 @@
 import jakarta.annotation.Nonnull;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.persistence.dao.entity.PrincipalSecretsResult;
-import software.amazon.awssdk.annotations.NotNull;
```
XN137, line 24: wrong import

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -700,7 +703,7 @@ private Map<String, SecretReference> extractSecretReferences(
    * @see #extractSecretReferences
    */
   private boolean requiresSecretReferenceExtraction(
-      @NotNull ConnectionConfigInfo connectionConfigInfo) {
+      @Nonnull ConnectionConfigInfo connectionConfigInfo) {
```
XN137, line 706: needed to be changed as quarkus thought this might be a rest endpoint or the like, quite sure `Nonnull` was the original intent (i.e. not validation here)

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -99,57 +89,21 @@ public class PolarisServiceImpl
...
+      PolarisAdminService adminService) {
+    this.realmConfig = realmConfig;
     this.reservedProperties = reservedProperties;
     this.polarisEventListener = polarisEventListener;
-  }
-
-  private PolarisAdminService newAdminService(
-      RealmContext realmContext, SecurityContext securityContext) {
-    PolarisPrincipal authenticatedPrincipal = (PolarisPrincipal) securityContext.getUserPrincipal();
-    if (authenticatedPrincipal == null) {
```
XN137, line 136: note that this checks happens in `PolarisAdminService` ctor already

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -172,7 +126,6 @@ private static Response toResponse(BaseResult result, Response.Status successSta
   @Override
   public Response createCatalog(
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
```
XN137, line 128: the `realmContext` and `securityContext` params in these methods are now all unused.
note that since this class has a `CallContext` field, we can be sure that `realmContext` is matching the one from the injected request-scoped fields.

wondering:
should we adjust openapi `server-templates` to remove one/both from the apis (now/in a followup)?

dimas-b: I'd say remove in a follow-up.

dimas-b (approved):

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
HonahX: +1 on follow-up, the same should apply to `IcebergAdapter`, `GenericTableAdapter` and `PolicyAdapter`. I think the server template update will not be a small PR

PR #2540, opened 2025-09-10
Make ResolverFactory + ResolutionManifestFactory request-scoped

XN137:
this avoids passing around the `CallContext` parameter

note that ideally the `SecurityContext` would also be injected from the request however our tests around `PolarisAuthzTestBase` are written in a way that does not easily support this currently.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
@@ -107,33 +107,6 @@ public class ServiceProducers {
   }
 
   @Produces
-  @ApplicationScoped
-  public ResolverFactory resolverFactory(
-      PolarisDiagnostics diagnostics,
-      MetaStoreManagerFactory metaStoreManagerFactory,
-      PolarisMetaStoreManager polarisMetaStoreManager) {
-    return (callContext, securityContext, referenceCatalogName) -> {
-      EntityCache entityCache =
-          metaStoreManagerFactory.getOrCreateEntityCache(
-              callContext.getRealmContext(), callContext.getRealmConfig());
-      return new Resolver(
-          diagnostics,
-          callContext.getPolarisCallContext(),
-          polarisMetaStoreManager,
-          securityContext,
-          entityCache,
-          referenceCatalogName);
-    };
-  }
-
-  @Produces
-  @ApplicationScoped
-  public ResolutionManifestFactory resolutionManifestFactory(
-      PolarisDiagnostics diagnostics, ResolverFactory resolverFactory) {
-    return new ResolutionManifestFactoryImpl(diagnostics, resolverFactory);
-  }
-
-  @Produces
   @Singleton
   public PolarisDiagnostics polarisDiagnostics() {
     return new PolarisDefaultDiagServiceImpl();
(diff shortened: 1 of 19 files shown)


dimas-b (commented):
LGTM :+1:

dimas-b (dismissed):
I agree that leveraging CDI for request-scope object injection is preferable to manually passing context object as method arguments.

Thanks for you work on this @XN137 !

dimas-b:
Good points, @HonahX !

From my POV, `ResolverFactory` could probably be replaced with a CDI producer for injecting request-scoped `Resolver` instances.

I tend to view this PR as an incremental step towards that goal :) 

That said, I guess it might be preferable to continue this kind of design discussion on the `dev` ML. I believe it is more readable and easier to discover than GH issues discussions... especially if the number of replies is large :)

HonahX:
Thanks @dimas-b for the explanation! +1 for a dev list discussion for this. 

> From my POV, ResolverFactory could probably be replaced with a CDI producer for injecting request-scoped Resolver instances.

I think this will be a good point to discuss. CDI producer could help simplify the code within the polaris, however I barely remembered that there was some discussion on making `polaris-core` a "library" that could be useful in other custom implementation/framework. Replacing the factory class with CDI producer might make it harder towards that goal. But this is just my perspective and would love to hear others' thoughts once it is on the dev list

XN137:
rebased after trivial conflict in `PolarisAdminService.java`

XN137:
sorry i was OOO for a week. thank you for the feedback everyone!

> IMHO, a "factory" makes more sense to be an ApplicationScoped bean, that once initialized, other methods/threads could use it to generate the real instance.

i dont think there is a strict requirement to have factories be application scoped.
a factory is just serving as an injectable decision point about how specific instances of an interface are supposed to be created (for a particular subset of the system).
that doesnt immediately imply requirements on the lifetime scope of the factory or the produced objects.

for example on `main` branch the `PolarisResolutionManifest` uses the `ResolverFactory` to produce a `primaryResolver` field immediately in the ctor and a "single-use" `passthroughResolver` on the fly in `getPassthroughResolvedPath`.

generally `ResolverFactory` is only used by `PolarisResolutionManifest` and `IcebergCatalogAdapter` both of which afaict have a lifetime bound to a single request.
this is indicated by both of them having a `CallContext` field, which gets initialized at construction time.

> Making factories request-scoped seem to indicate that we do not need actually need these factories, and we can directly have request-scoped Resolver and ResolutionManfiest bean. Otherwise we are generating an additional dispensible factory object for every request.

i would disagree with this because as shown in this PR both `ResolverFactory` and `ResolutionManifestFactory` cant be used without a `CallContext` which is only available in a request-scope.
also, as pointed out above, the factory is also used on the fly to generate multiple instances inside a single object / the "same request".

so while the factory currently is application-scoped it cant be used outside of a request-scope and this means all callers must pass all the required parameters around all the time, which this PR tries to avoid/clean up.

> I think in general the factory design could allow us to add potential customization in the future and maybe some of them could be cached at realm-level instead of every request.

correct me if i am wrong but i think even with making the factories request-scope (which matches their usage in the codebase) the implementation of that factory is still free to implement the object creation however it wants.
so if something is supposed to be cached at realm-level the factory is still free to do that.
(for example by injecting an application scoped bean that handles shared realm state).
but this custom implementation decision does not need to be part of the factory usage in the codebase.
a request-scoped factory can also return a singleton object if it really wants.

XN137:
> From my POV, ResolverFactory could probably be replaced with a CDI producer for injecting request-scoped Resolver instances.

not sure this is possible as explained in how `PolarisResolutionManifest` uses the `ResolverFactory` i.e. one resolver instance being created in the ctor and one on the fly in a method call.
so we cant inject Resolver instances directly (apart from maybe the `primaryResolver` one, assuming its state not being shared by multiple unrelated usage spots).

XN137:
rebased again after trivial conflict in `PolarisAdminService.java`

dimas-b (approved):
This refactoring LGTM and I think it is aligned with our code "evolution" approach, so I think it can be merged without further `dev` ML discussions.

@HonahX : WDYT?

This, of course, does not preclude `dev` list discussions about API changes and how to deal with CDI beans and factories in general. I merely mean that this particular change looks reasonable to merge from my POV 🙂

XN137:
rebased again after trivial conflict in `PolarisAdminService.java`

dimas-b:
@HonahX : Do you think we could merge this?

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -169,6 +142,34 @@ public PolarisAuthorizer polarisAuthorizer(RealmConfig realmConfig) {
...
+            securityContext,
+            entityCache,
+            referenceCatalogName);
+  }
+
+  @Produces
+  @RequestScoped
+  public ResolutionManifestFactory resolutionManifestFactory(
+      PolarisDiagnostics diagnostics, RealmContext realmContext, ResolverFactory resolverFactory) {
+    return new ResolutionManifestFactoryImpl(diagnostics, realmContext, resolverFactory);
```
adutra, line 170: Just a small nit: why do you need this bean to be produced by a producer method? Wouldn't it be enough to annotate `ResolutionManifestFactoryImpl` with `@RequestScoped`?

XN137: it's probably not strictly needed, this just keeps the same style the previous code used.
we can also do it the other way if you prefer.
i'd suggest to do it a followup if its ok with you?

dimas-b:
Given some approvals I assume lazy consensus from people who comment before but did not formally approve... merging.

Open to follow-up (if required).

PR #2543, opened 2025-09-10
Avoid using jackson method for parsing YAML from any URL in RootCredentialsSet

dimas-b:
As discussed https://github.com/FasterXML/jackson-core/issues/803 this method can lead to hidden issues and got deprecated.

Instead, we manage URL steams locally in RootCredentialsSet and permit only those URLs that do not have the host component (such as files and java resources)... which makes sense to do from a general security perspective too.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/RootCredentialsSet.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/RootCredentialsSet.java
@@ -32,7 +32,7 @@ import com.google.common.base.Strings;
 import jakarta.annotation.Nullable;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URL;
+import java.net.URI;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/RootCredentialsSet.java
```diff
@@ -131,17 +138,28 @@ static RootCredentialsSet fromList(List<String> credentialsList) {
    * </pre>
    */
   static RootCredentialsSet fromUrl(URL url) {
```
snazy: Shouldn't this be refactored to `java.net.URI` to prevent actual DNS resolution via `java.net.URL`?

dimas-b: Good point :+1: Updated.

polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/RootCredentialsSet.java
```diff
@@ -130,18 +137,29 @@ static RootCredentialsSet fromList(List<String> credentialsList) {
    * }
    * </pre>
    */
-  static RootCredentialsSet fromUrl(URL url) {
+  static RootCredentialsSet fromUri(URI uri) {
+    Preconditions.checkNotNull(uri);
+    Preconditions.checkArgument(
+        Strings.isNullOrEmpty(uri.getHost()),
+        "Remote URIs are not allowed for RootCredentialsSet: %s",
+        uri);
```
snazy, line 145: Actually thinking a bit more about this, you can still get a "host" via a `jar` URI like `jar:http://foo.bar/...` or all the `file` scheme "specialties". But that's rather a question what source locations we want to allow.

For the scope of this PR, the change as it is looks fine.

snazy (approved):

PR #2544, opened 2025-09-10
Fix deprecation warnings in GcpCredentialsStorageIntegrationTest

dimas-b:
Refactor the code and suppress the warning about readTree(URL). The rationale for its deprecation does not apply to java resources.

Cf. https://github.com/FasterXML/jackson-core/issues/803

### DIFF
--- a/polaris-core/src/test/java/org/apache/polaris/service/storage/gcp/GcpCredentialsStorageIntegrationTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/service/storage/gcp/GcpCredentialsStorageIntegrationTest.java
@@ -37,6 +37,7 @@ import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


polaris-core/src/test/java/org/apache/polaris/service/storage/gcp/GcpCredentialsStorageIntegrationTest.java
```diff
@@ -177,6 +177,12 @@ private AccessConfig subscopedCredsForOperations(
         Optional.of(REFRESH_ENDPOINT));
   }
 
+  private JsonNode readResource(ObjectMapper mapper, String name) throws IOException {
+    // readTree from a test resource is not affected by the issues that prompted its deprecation
+    //noinspection deprecation
+    return mapper.readTree(GcpCredentialsStorageIntegrationTest.class.getResource(name));
```
HonahX: [Non-blocker] For my learning purpose, is there any reason that we want to suppress it instead of fixing it by explicitly opening a stream and read the stream?

HonahX (dismissed):

polaris-core/src/test/java/org/apache/polaris/service/storage/gcp/GcpCredentialsStorageIntegrationTest.java
dimas-b: I thought it might be simpler to suppress for test code... but it's a good point. I'll update.

dimas-b: updated... it came down to the same number of lines :)

HonahX (approved):

PR #2555, opened 2025-09-12
Request-scoped PolarisMetaStoreManager

XN137:
see ML discussion:
https://lists.apache.org/thread/ot19px6t0w808pp994rrc8kk7186dfxm

the `PolarisMetaStoreManager` interface methods all take a `PolarisCallContext` parameter
because they need access to `PolarisCallContext.getMetaStore` aka the persistence session.
also occasionally the `RealmContext` and `RealmConfig` is needed.

this means `PolarisMetaStoreManager` is only usable within a request-scope.
`MetaStoreManagerFactory.getOrCreateMetaStoreManager` impls suggests that there needs to
be one shared `PolarisMetaStoreManager` instance per realm.
however if we look at the `PolarisMetaStoreManager` impls (`AtomicOperationMetaStoreManager`
and `TransactionalMetaStoreManagerImpl`) we can see that they dont have any state themselves, the fields only reference application-scoped instances.

Those 2 observations combined mean that we can make the `PolarisMetaStoreManager` request-scoped and then the `PolarisCallContext` parameter can be removed from all the methods without losing any functionality/flexibility.
note that if custom implementations would need any state sharing for a realm, they are still free to do so in their custom `MetaStoreManagerFactory` implementations, as they can
initialize the custom request-scoped `PolarisMetaStoreManager` instance with whatever (shared) data they need.

when `PolarisMetaStoreManager` is request-scoped the `getOrCreateSession` method also becomes a private implementation detail of the `MetaStoreManagerFactory`.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/JWTBroker.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/JWTBroker.java
@@ -27,7 +27,6 @@ import java.time.temporal.ChronoUnit;
 import java.util.Optional;
 import java.util.UUID;
 import org.apache.iceberg.exceptions.NotAuthorizedException;
-import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.entity.PrincipalEntity;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.dao.entity.PrincipalSecretsResult;
@@ -48,11 +47,9 @@ public abstract class JWTBroker implements TokenBroker {
   private static final String CLAIM_KEY_PRINCIPAL_ID = "principalId";
   private static final String CLAIM_KEY_SCOPE = "scope";
 
-  private final PolarisMetaStoreManager metaStoreManager;
   private final int maxTokenGenerationInSeconds;
 
-  JWTBroker(PolarisMetaStoreManager metaStoreManager, int maxTokenGenerationInSeconds) {
-    this.metaStoreManager = metaStoreManager;
+  JWTBroker(int maxTokenGenerationInSeconds) {
     this.maxTokenGenerationInSeconds = maxTokenGenerationInSeconds;
   }
 
@@ -86,7 +83,7 @@ public abstract class JWTBroker implements TokenBroker {
       String subjectToken,
       String grantType,
       String scope,
-      PolarisCallContext polarisCallContext,
+      PolarisMetaStoreManager metaStoreManager,
       TokenType requestedTokenType) {
     if (requestedTokenType != null && !TokenType.ACCESS_TOKEN.equals(requestedTokenType)) {
       return TokenResponse.of(OAuthError.invalid_request);
@@ -125,7 +122,7 @@ public abstract class JWTBroker implements TokenBroker {
       String clientSecret,
       String grantType,
       String scope,
-      PolarisCallContext polarisCallContext,
+      PolarisMetaStoreManager metaStoreManager,
       TokenType requestedTokenType) {
     // Initial sanity checks
     TokenRequestValidator validator = new TokenRequestValidator();
@@ -135,7 +132,8 @@ public abstract class JWTBroker implements TokenBroker {
       return TokenResponse.of(initialValidationResponse.get());
     }
 
-    Optional<PrincipalEntity> principal = findPrincipalEntity(clientId, clientSecret);
+    Optional<PrincipalEntity> principal =
+        findPrincipalEntity(metaStoreManager, clientId, clientSecret);
     if (principal.isEmpty()) {
       return TokenResponse.of(OAuthError.unauthorized_client);
     }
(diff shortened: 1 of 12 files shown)


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -196,52 +188,46 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
...
-  public synchronized PolarisMetaStoreManager getOrCreateMetaStoreManager(
-      RealmContext realmContext) {
-    if (!metaStoreManagerMap.containsKey(realmContext.getRealmIdentifier())) {
-      DatasourceOperations datasourceOperations = getDatasourceOperations();
-      initializeForRealm(datasourceOperations, realmContext, null);
-      checkPolarisServiceBootstrappedForRealm(realmContext);
+  public PolarisMetaStoreManager createMetaStoreManager(
+      RealmContext realmContext, @Nullable RealmConfig realmConfig) {
+    if (realmConfig == null) {
+      realmConfig = new RealmConfigImpl(configurationStore, realmContext);
```
dimas-b: Do we need the `realmConfig` parameter at all?

polaris-core/src/main/java/org/apache/polaris/core/persistence/cache/InMemoryEntityCache.java
```diff
@@ -43,19 +42,12 @@
 public class InMemoryEntityCache implements EntityCache {
 
   private final PolarisDiagnostics diagnostics;
-  private final PolarisMetaStoreManager polarisMetaStoreManager;
```
dimas-b, line 46: I suppose InMemoryEntityCache must use a PolarisMetaStoreManager from the same realm. If `PolarisMetaStoreManager` is a method parameter it become not so obvious. WDYT about keeping this field?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
XN137: as we can build it ourselves its not strictly needed no, but many call sites already have it and can pass it in.
i am fine to remove the parameter if we think thats better overall.

polaris-core/src/main/java/org/apache/polaris/core/persistence/cache/InMemoryEntityCache.java
XN137: afaict we cant keep this field as `PolarisMetaStoreManager` now represents a (per-request) persistence session where-as `InMemoryEntityCache` is supposed to be a shared cache within a realm.

the question is with what persistence session do we want to populate the cache?
a shared one or the request-scoped one given by the callers as "use this for loading in case cache entry is missing" ?

XN137:
rebased after trivial a conflict and also to adjust to `AccessConfigProvider` (though we need it to become request-scoped as well soon)

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/JWTBroker.java
```diff
@@ -50,11 +49,9 @@ public abstract class JWTBroker implements TokenBroker {
   private static final String CLAIM_KEY_PRINCIPAL_ID = "principalId";
   private static final String CLAIM_KEY_SCOPE = "scope";
 
-  private final PolarisMetaStoreManager metaStoreManager;
   private final int maxTokenGenerationInSeconds;
 
-  JWTBroker(PolarisMetaStoreManager metaStoreManager, int maxTokenGenerationInSeconds) {
-    this.metaStoreManager = metaStoreManager;
+  JWTBroker(int maxTokenGenerationInSeconds) {
```
dimas-b, line 52: `TokenBroker` is already request-scoped, why not rely on the constructor to have the correct `metaStoreManager` for  the current request?

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/TokenBrokerFactory.java
```diff
@@ -18,11 +18,13 @@
...
 import org.apache.polaris.core.context.RealmContext;
 
 /**
  * Factory that creates a {@link TokenBroker} for generating and parsing. The {@link TokenBroker} is
  * created based on the realm context.
  */
-public interface TokenBrokerFactory extends Function<RealmContext, TokenBroker> {}
+public interface TokenBrokerFactory {
+
+  TokenBroker newTokenBroker(RealmContext realmContext);
```
dimas-b, line 29: suggestion: move this into a separate PR maybe? I think it's a valuable refactoring and independent of the request scope discussions.

polaris-core/src/main/java/org/apache/polaris/core/persistence/cache/InMemoryEntityCache.java
dimas-b: If the cache answers from memory, the per-request persistence session is irrelevant. So the caller of the lookup method(s) does not know whether the current persistence session was / will be used. Some values returned from the cache might have been loaded via a different session.

The "reload if necessary" checks are based on local entity data and do not appear to take the "session" into account.

With that in mind, I think it would be preferable to resolve this ambiguity and always load from a separate session (one per `InMemoryEntityCache` instance or created on demand).

@dennishuo : WDYT?

dimas-b (commented):
The general direction of this PR LGTM (but I have not reviewed all changes yet). Dealing with `PolarisMetaStoreManager` instances as per-request objects makes sense to me (irrespective of the runtime framework in use). The metastore factory controls the creation of `PolarisMetaStoreManager` objects and associating them with request-specific data.

@dennishuo @collado-mike WDYT?

snazy:
Looks like this PR needs a rebase

PR #475, opened 2024-11-25
CI/Build: Add Gradle Build Scans

snazy:
Gradle Build Scans, free of use, collect a lot of information about a Gradle build, including the actual output of failed test. This becomes quite convenient when inspecting test failures in CI and a lot of other information about Gradle builds.

[Example build scan](https://scans.gradle.com/s/jpuykotf4hac6)

PR #2559, opened 2025-09-13
Publish build scans to develocity.apache.org for build insights

clayburn:
@flyrain - this is the PR we discussed today at Community over Code.

This PR publishes a Build Scan for every CI build and for every local build from an authenticated Apache committer. The build will not fail if publishing fails.

The Build Scans of the Apache Polaris project are published to the Develocity instance at develocity.apache.org, hosted by the Apache Software Foundation and run in partnership between the ASF and Gradle. This Develocity instance has all features and extensions enabled and is freely available for use by the Apache Polaris project and all other Apache projects.

On this Develocity instance, Apache Polaris will have access not only to all of the published Build Scans but other aggregate data features such as:

- Dashboards to view all historical Build Scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

Please let me know if there are any questions about the value of Develocity or the changes in this pull request and I’d be happy to address them.


adutra:
@clayburn this is great, thank you for this contribution!

Note: you need to run `./gradlew spotlessApply` to fix the formatting issues before we can approve.

snazy:
Thanks @clayburn for the PR.
It looks like it's addressing the same functionality as #475

clayburn:
Thanks @adutra. I have pushed up the changes to pass the spotless check.

@snazy - I did not see your PR. A few of the updates in mine are necessary, like the server URL and the secret name. You'll also find that the Common Custom User Data Gradle Plugin in my PR will apply some extra tags and values similar to what you are doing. You can see those docs [here](https://github.com/gradle/common-custom-user-data-gradle-plugin?tab=readme-ov-file#captured-data). There are other things you are trying to do in that file though, so let me know you'd like to proceed here.

flyrain:
@clayburn thanks for the PR. The develocity.apache.org looks nice to me. I think it will benefit Polaris. A basic question: does ASF Infra set the `DEVELOCITY_ACCESS_KEY` or Polaris PMC member/committer set it?

flyrain (approved):

settings.gradle.kts
```diff
@@ -109,6 +109,11 @@ pluginManagement {
   }
 }
 
+plugins {
+  id("com.gradle.develocity") version "4.1.1"
```
dimas-b, line 113: Would it be possible to use https://github.com/apache/polaris/blob/main/gradle/baselibs.versions.toml for version management (as we do for other build plugins)?

settings.gradle.kts
```diff
@@ -126,3 +131,15 @@ gradle.beforeProject {
...
 }
+
+val isCI = System.getenv("CI") != null
+
+develocity {
+  server = "https://develocity.apache.org"
+  projectId = "polaris"
+  buildScan {
+    uploadInBackground = !isCI
+    publishing.onlyIf { it.isAuthenticated }
```
dimas-b, line 142: Sending scans when authenticated SGTM... Still, it might be worth mentioning it in the main README for awareness. WDYT?

settings.gradle.kts
snazy: It's a "settings plugin", so sadly no.

settings.gradle.kts
snazy: Apache's Develocity instance requires authentication. (See my other comment).

snazy (commented):
Overall LGTM, and I can add what's really missing from #475 in a follow-up PR.

I think what's missing is the actual secret being added to the `apache/polaris` GitHub repo (I don't see build scans being published from this PR).

Guess it's worth to add a link to [the Gradle docs](https://docs.gradle.com/develocity/gradle-plugin/current/#automated_access_key_provisioning) about how to use build scans in local builds in `CONTRIBUTING.md`. That however requires an Apache account. We can add support for non-Apache-committers in a follow-up, so they can use Gradle's public instance.

settings.gradle.kts
clayburn: @snazy is correct. Unfortunately this is not doable

settings.gradle.kts
clayburn: I added a bit to the README about build scan publication to the README.

clayburn:
> A basic question: does ASF Infra set the DEVELOCITY_ACCESS_KEY or Polaris PMC member/committer set it?

> I think what's missing is the actual secret being added to the apache/polaris GitHub repo (I don't see build scans being published from this PR).

The `DEVELOCITY_ACCESS_KEY` secret is an organization secret at the `apache` org level. It is already present and used by other ASF projects. Unfortunately, with the way GitHub Actions works, pull requests from forks cannot access this secret, so they will not publish build scans. This is something we'd like to provide workflows for doing, but not something we have today.

If you want to verify this pull request, you are welcome to push this branch to the `apache/polaris` repo and run a build from that branch.

snazy:
> The `DEVELOCITY_ACCESS_KEY` secret is an organization secret at the `apache` org level.

Ah, true. Totally makes sense.

jbonofre:
For reference: https://infra.apache.org/gradle.html

jbonofre (approved):
It looks good to me, and compliant with https://cwiki.apache.org/confluence/display/INFRA/Project+Onboarding+Instructions+for+Develocity

Thanks !

dimas-b (approved):

flyrain (approved):
+1 Thanks for the context @clayburn and @jbonofre !

flyrain:
Thanks a lot @clayburn for adding the integration! Thanks @snazy @dimas-b @jbonofre for the review!

PR #2565, opened 2025-09-15
Remove DROP statements from SQL init scripts

adutra:
SQL init scripts must be idempotent, because they may be invoked several times by a Polaris server during realm bootstrapping (the script is invoked once per realm to bootstrap).

It is therefore not possible to put any DROP statements in the scripts as this would result in immediate data loss of all previously-bootstrapped realms.

### DIFF
--- a/persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
+++ b/persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
@@ -35,7 +35,6 @@ MERGE INTO version (version_key, version_value)
 -- H2 supports COMMENT, but some modes may ignore it
 COMMENT ON TABLE version IS 'the version of the JDBC schema in use';
 
-DROP TABLE IF EXISTS entities;
 CREATE TABLE IF NOT EXISTS entities (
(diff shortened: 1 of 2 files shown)


flyrain:
That's a valid concern. Thanks for catching it, @adutra ! cc @singhpk234

dimas-b (approved):
Making schema scripts idempotent is a valuable change! Thanks, @adutra !

persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
```diff
@@ -35,7 +35,6 @@ MERGE INTO version (version_key, version_value)
 -- H2 supports COMMENT, but some modes may ignore it
 COMMENT ON TABLE version IS 'the version of the JDBC schema in use';
 
-DROP TABLE IF EXISTS entities;
```
dimas-b, line 38: I guess this is worth mentioning in `CHANGELOG.md`, WDYT?

adutra: Will tackle that as a follow-up PR.

PR #2566, opened 2025-09-15
Make database column events.request_id nullable

adutra:
The request ID is an optional information coming from the REST request. This PR makes it nullable in the database schema.

This PR also annotates the `ModelEvent.principalName` and `PolarisEvent.principalName` fields as nullable in code (the corresponding column was already nullable in the database schema).

### DIFF
--- a/persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
+++ b/persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
@@ -129,7 +129,7 @@ CREATE TABLE IF NOT EXISTS events (
     realm_id TEXT NOT NULL,
     catalog_id TEXT NOT NULL,
     event_id TEXT NOT NULL,
-    request_id TEXT NOT NULL,
+    request_id TEXT,
     event_type TEXT NOT NULL,
     timestamp_ms BIGINT NOT NULL,
     principal_name TEXT,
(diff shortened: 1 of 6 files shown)


persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
```diff
@@ -129,7 +129,7 @@ CREATE TABLE IF NOT EXISTS events (
     realm_id TEXT NOT NULL,
     catalog_id TEXT NOT NULL,
     event_id TEXT NOT NULL,
```
dimas-b, line 131: side note: assuming `event_id` values are unique, this table can grow definitely, I guess... Do we have a mechanism to cap its size?

singhpk234: does capping size means not persisting the events when the number of records > threshold or failing unless its fixed, my understanding is failing unless we fix it (like removing last x months events in a cold store type of things and then for users we can make a view which does union from both the stores ?) would be a better way to go if we try to put a cap as we want to repurpose this for audit POV

singhpk234 (approved):
LGTM as well !

persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
eric-maynard: I would expect the listener to handle capping the size -- or some other maintenance service responsible for pruning this table.

dimas-b: My primary concern is about naive users who deploy pre-built binaries and start accumulating data in this table :thinking:

adutra: Created an issue to track this: https://github.com/apache/polaris/issues/2573

PR #2572, opened 2025-09-15
fix password in README.md for `./gradlew run`

dimas-b:
Use the password that matches what the `run` task actually configures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Apache Polaris is built using Gradle with Java 21+ and Docker 27+.
 - `./gradlew build` - To build and run tests. Make sure Docker is running, as the integration tests depend on it.
 - `./gradlew assemble` - To skip tests.
 - `./gradlew check` - To run all checks, including unit tests and integration tests.
-- `./gradlew run` - To run the Polaris server locally; the server is reachable at localhost:8181. This is also suitable for running regression tests, or for connecting with Spark. Set your own credentials by specifying system property `./gradlew run -Dpolaris.bootstrap.credentials=POLARIS,root,secret` where:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,7 +65,7 @@ Apache Polaris is built using Gradle with Java 21+ and Docker 27+.
 - `./gradlew build` - To build and run tests. Make sure Docker is running, as the integration tests depend on it.
 - `./gradlew assemble` - To skip tests.
 - `./gradlew check` - To run all checks, including unit tests and integration tests.
-- `./gradlew run` - To run the Polaris server locally; the server is reachable at localhost:8181. This is also suitable for running regression tests, or for connecting with Spark. Set your own credentials by specifying system property `./gradlew run -Dpolaris.bootstrap.credentials=POLARIS,root,secret` where:
+- `./gradlew run` - To run the Polaris server locally; the server is reachable at localhost:8181. This is also suitable for running regression tests, or for connecting with Spark. Set your own credentials by specifying system property `./gradlew run -Dpolaris.bootstrap.credentials=POLARIS,root,s3cr3t` where:
```
dimas-b, line 68: Cf. https://github.com/apache/polaris/blob/b6e247deda8a618136ec16c75b86eb2a2d12642b/runtime/server/build.gradle.kts#L81

flyrain: This was changed to `secret` by #2558. I think using `secret` instead of `s3cr3t` makes more sense. The command here is to show how to set your own credentials.  This part is only required when a custom credential needed `Dpolaris.bootstrap.credentials=POLARIS,root,s3cr3t`, without it, the default value is still `root, s3cr3t`. 

Another reason to use `secret` is that the follow doc says that:
```
root is the CLIENT_ID
secret is the CLIENT_SECRET
```

dimas-b: Oops, apparently I misinterpreted :facepalm: will revert momentarily

dimas-b: #2576

PR #2574, opened 2025-09-15
Introduce alternate in-memory buffering event listener

adutra:
As promised, this is a proposal for an alternative implementation of `InMemoryBufferPolarisPersistenceEventListener`, leveraging Mutiny and Smallrye Fault Tolerance, and avoiding dealing with buffering and flushing manually.

For now I'm proposing a separate CDI bean, to facilitate reviews or comparisons.

But if the approach taken in this PR is considered a better solution for buffering events, I would suggest to replace `InMemoryBufferPolarisPersistenceEventListener` completely.


adutra:
\cc @adnanhemani

runtime/service/build.gradle.kts
```diff
@@ -128,22 +128,23 @@ dependencies {
   testImplementation("io.quarkus:quarkus-junit5-mockito")
   testImplementation("io.quarkus:quarkus-rest-client")
   testImplementation("io.quarkus:quarkus-rest-client-jackson")
+  testImplementation("io.quarkus:quarkus-jdbc-h2")
```
adutra, line 132: I did some cleanup in this file which had a few duplicate dependencies.

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventProducerTest.java
```diff
@@ -0,0 +1,182 @@
...
+  private void sendAsync(String realmId, int n) {
+    for (int i = 0; i < n; i++) {
+      eventLoopGroup.next().execute(() -> producer.processEvent(realmId, event()));
+    }
+  }
+
+  private void assertRows(String realmId, int expected, Duration timeout) {
+    String query = "SELECT COUNT(*) FROM polaris_schema.events WHERE realm_id = '" + realmId + "'";
+    await()
+        .atMost(timeout)
```
adnanhemani: Seems like we are testing with actual delays. Is there any way around this? Personally, I would like if we not increase the running time of the test suite as much as possible.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @PreDestroy
+  public void shutdown() {
+    processors.asMap().values().forEach(UnicastProcessor::onComplete);
+    processors.invalidateAll(); // doesn't call the eviction listener
+  }
+
+  protected UnicastProcessor<PolarisEvent> createProcessor(String realmId) {
+    UnicastProcessor<PolarisEvent> processor = UnicastProcessor.create();
```
adnanhemani, line 108: I don't believe that `UnicastProcessor` is thread-safe for multiple producers by default. However, if we pass in a `ConcurrentLinkedQueue` during `create`, we may be able to make it become thread-safe. Thoughts?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
```diff
@@ -0,0 +1,131 @@
...
+    processor
+        .emitOn(Infrastructure.getDefaultWorkerPool())
```
adnanhemani, line 110: Question on this: is this creating a separate execution service? Is it possible for this to become a bottleneck in terms of how fast events can be ingested?

It it be better or worse to have the threads here managed by Quarkus instead?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
```diff
@@ -0,0 +1,131 @@
...
+        .group()
+        .intoLists()
+        .of(configuration.maxBufferSize(), configuration.bufferTime())
+        .subscribe()
+        .with(events -> flush(realmId, events));
```
adnanhemani: Leaving here as a note for anyone else wondering:

Even though this pattern will dequeue events one-by-one when attempting to generate the list of events, I don't believe we will have the race condition where events are being added at a higher rate than the amount of events being drained. This is because only a fixed amount of events can be requested by the subscription at a time (which seems like `Long.MAX_VALUE`). If this load is continuously sustained, then it may overwhelm the service by overflowing the queue that is backing the UnicastProcessor - but I don't think there's many other ways to get around that.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
```diff
@@ -112,12 +112,12 @@ public void onAfterCatalogCreated(AfterCatalogCreatedEvent event) {
     processEvent(polarisEvent);
   }
 
-  protected record ContextSpecificInformation(long timestamp, @Nullable String principalName) {}
+  public record ContextSpecificInformation(long timestamp, @Nullable String principalName) {}
 
-  abstract ContextSpecificInformation getContextSpecificInformation();
+  protected abstract ContextSpecificInformation getContextSpecificInformation();
```
adnanhemani, line 95: Curious as to why these changes were made?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: Very good question! But in fact, it is safe for multiple producers as the `onNext` method is synchronized. No need to use a MPSC queue.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: On a Quarkus platform, `Infrastructure.getDefaultWorkerPool()` always returns Quarkus default worker pool. So this method is not creating a separate execution service – which is why we don't need to care about shutting it down either.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: If I understand correctly the concern, what matters here is the `UnicastProcessor` queue. For now, it's unbounded, so there should never be a `BackPressureFailure` raised, but the small risk is that the queue could grow if the database cannot keep up with the number of events coming in.

I think it's OK for now,  if users report memory issues we can look into bounded queues. But I don't think the server can generate events faster than the database is able to write them.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
adutra: Because otherwise you cannot access this method from outside this package. 

Generally speaking, since this class is meant to be subclassed, even by implementors outside Polaris, it's not recommended to use package-private members.

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventProducerTest.java
adutra: Indeed, but then we need a separate test to test just the timeouts. I will do the change.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: BTW: the tests are testing this. If you look at the `sendAsync` method, it fires events from multiple threads.

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventProducerTest.java
adutra: Done. It's indeed much faster now 👍

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: I added a small safeguard: if the upstream terminates with a failure, a message is logged and the processor is evicted from the cache.

dimas-b (approved):

adutra:
> LGTM 👍
> 
> @adutra - I can ask others to review and merge this as well. Would you like me to work on gluing this together with the already existing Event Listener, or would you like to do it?

I can take care of that. Let's merge this one first if everyone agrees? I'll need re-approvals because of a merge conflict 😄

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
```diff
@@ -0,0 +1,141 @@
...
+  @SuppressWarnings("unused")
+  protected void onFlushError(String realmId, List<PolarisEvent> events, Throwable error) {
+    LOGGER.error("Failed to persist {} events for realm '{}'", events.size(), realmId, error);
+  }
+
+  protected void onProcessorError(String realmId, Throwable error) {
+    LOGGER.error(
+        "Unexpected error while processing events for realm '{}'; some events may have been dropped",
+        realmId,
+        error);
```
singhpk234, line 138: I wonder if its just okay to drop events considering they are deemed to be Audit logs ?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
singhpk234: is there a way to configure this workerPool ? lets say if i my one instance is handling a lot of realms ?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
```diff
@@ -0,0 +1,141 @@
...
+  protected void processEvent(String realmId, PolarisEvent event) {
+    var processor = Objects.requireNonNull(processors.get(realmId));
+    processor.onNext(event);
+  }
+
+  @Override
+  protected ContextSpecificInformation getContextSpecificInformation() {
+    var principal = securityContext.getUserPrincipal();
+    var principalName = principal == null ? null : principal.getName();
+    return new ContextSpecificInformation(clock.millis(), principalName);
```
singhpk234, line 92: may be this is not in the scope of the pr, but do we know why just principal and not the activated roles ?

singhpk234 (approved):
LGTM as well, thanks @adutra ! 
Added some questions for my better understanding

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adnanhemani: This one is on me, I forgot to put it in the original PR - I will enhance this in the future to include activated roles.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: Yes, since the beginning the consistency guarantee has been "at most once", so it's understood that some events may be lost.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryEventListener.java
adutra: Yes, it's configurable through the `quarkus.thread-pool.*` options:

https://quarkus.io/guides/all-config#quarkus-core_quarkus-thread-pool-core-threads

The core thread pool is responsible for running all blocking tasks in Quarkus. I think this is appropriate for this use case. Note: this thread pool does NOT handle HTTP requests.

If this becomes problematic one day, the solution would be to declare a separate `Executor` CDI bean and inject it here.

adutra:
Merging this now, I will tackle the remaining improvements next.

PR #2578, opened 2025-09-15
Avoid exceptions on ETag matches

dimas-b:
Exceptions have runtime overhead, which is avoidable in this case.


singhpk234 (approved):
LGTM, Thank for the change @dimas-b !

Not sure why we were throwing an exception when our intention was to return back NOT_MODIFIED status code

HonahX (approved):
LGTM!

PR #1102, opened 2025-03-03
Automatic license check for Python

DaniilRoman:
https://github.com/apache/polaris/issues/822

Here's what it looks like when it passes:

Screenshot 2025-03-03 at 19 08 35

And here's what it looks like when it doesn't pass:

Screenshot 2025-03-03 at 20 08 07

Issue #822, opened 2025-01-18
Automatic license checks for Python


snazy:
### Describe the bug

We already have ways to collect and verify licenses for the Java ecosystem.

It would help a lot to also have license verification for the Python ecosystem, to ensure that dependencies are used in an ASF-compliant way.

This is rather not just an "enhancement" or "improvement" but a necessity, because I suspect that not many people will manually inspect all Python direct and indirect dependencies after every dependency change or at least before every release.

License violations cause serious trouble.

Related: #821

MonkeyCanCode:
Maybe consider https://github.com/raimon49/pip-licenses for this?

MonkeyCanCode:
Close this one as https://github.com/MonkeyCanCode/polaris/commit/65495515121d52e6c63a87ce3ee05e6ad7a1b544 is merged.

PR #2580, opened 2025-09-16
Python client: add license check

MonkeyCanCode:
This is requested by @snazy a while back via https://github.com/apache/polaris/issues/822 and @DaniilRoman did the initial implementation via https://github.com/apache/polaris/pull/1102. This is the PR for merged the changes from sample PR with our GH action and Makefile. 

There are a lot more allowed licenses from ASF (https://www.apache.org/legal/resolved.html#category-x) and a bunch for which should't be included as well. For now, I put the allow list with the packages that are currently being used.

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -93,7 +93,7 @@ ignore-packages = [
     "zstandard",             # BSD-3-Clause
 ]
 partial-match = true
-# allow-only = "MIT;Apache;BSD License;PSF-2.0;ISC;The Unlicense;Python Software Foundation License;Mozilla Public License"
+allow-only = "MIT;Apache;BSD License;PSF-2.0;ISC;The Unlicense;Python Software Foundation License;Mozilla Public License"
 
 [build-system]
 requires = ["poetry-core>=2.0.0,<3.0.0", "openapi-generator-cli==7.11.0.post0"]


flyrain (approved):
+1 Thanks for working on it, @MonkeyCanCode !

Makefile
```diff
@@ -162,6 +162,12 @@ client-integration-test: client-setup-env ## Run client integration tests
 	@echo "Tearing down Docker Compose services..."
 	@$(DOCKER) compose -f $(PYTHON_CLIENT_DIR)/docker-compose.yml down || true # Ensure teardown even if tests fail
 
+.PHONY: client-license-check
+client-license-check: client-setup-env ## Run license compliance check
+	@echo "--- Starting license compliance check ---"
+	@$(ACTIVATE_AND_CD) && pip-licenses
```
HonahX, line 168: Somehow this does not work for me in an old poetry environment. 
```
pip-licenses: command not found
```
But I've verified that a clean install will work.

HonahX (approved):
LGTM! Thanks for working on this. 

I've verified this can catch non-supported license in deps. e.g.
```
--- Starting license compliance check ---
license GNU Lesser General Public License v2 or later (LGPLv2+) not in allow-only licenses was found for package chardet:5.2.0
make: *** [client-license-check] Error 1
```

Makefile
MonkeyCanCode: Maybe the previous poetry env doesn't have this dependency installed as I added it last night via this PR. In case if u want that env to work, u can source the venv then run poetry command to install all again.

HonahX: Yeah I ran make install-dependencies and I saw poetry explicitly said it installed pip-licenses yet it still could not found it. Could be some other weird issue in that env. So I ends up using a clean environment to verify : )


Using poetry run pip-licenses will work btw

snazy:
@MonkeyCanCode the merge broke CI on `main`. Mind taking a look?

snazy:
> @MonkeyCanCode the merge broke CI on `main`. Mind taking a look?

NVM, merged your fix.

Issue #2207, opened 2025-07-29
Malformed subscoped credentials error with external s3


lastranget:
### Describe the bug

I'm trying to set up a polaris catalog that points to our on-premises Pure FlashBlade s3 storage instance.

I'm getting the following error when I try to create a table via spark sql shell:

`org.apache.iceberg.exceptions.RESTException: Unable to process: Failed to get subscoped credentials: (Service: Sts, Status Code: 400, Request ID: null) (SDK Attempt Count: 1)`

This is similar to the errors reported in #1146 , but those errors have additional details as to the actual issue after "subscoped credentials:", whereas in this case, the error message is left incomplete.

I believe from https://github.com/apache/polaris/pull/1913 that external s3 providers should be supported currently.

The full stack trace is as follows:

```
spark-sql ()> CREATE NAMESPACE ICE_NS;
Time taken: 1.11 seconds
spark-sql ()> USE NAMESPACE ICE_NS;
Time taken: 0.071 seconds
spark-sql (ICE_NS)> CREATE TABLE PEOPLE (id int, name string) USING iceberg;
25/07/29 17:57:42 ERROR SparkSQLDriver: Failed in [CREATE TABLE PEOPLE (id int, name string) USING iceberg]
org.apache.iceberg.exceptions.RESTException: Unable to process: Failed to get subscoped credentials: (Service: Sts, Status Code: 400, Request ID: null) (SDK Attempt Count: 1)
	at org.apache.iceberg.rest.ErrorHandlers$DefaultErrorHandler.accept(ErrorHandlers.java:248)
	at org.apache.iceberg.rest.ErrorHandlers$TableErrorHandler.accept(ErrorHandlers.java:123)
	at org.apache.iceberg.rest.ErrorHandlers$TableErrorHandler.accept(ErrorHandlers.java:107)
	at org.apache.iceberg.rest.HTTPClient.throwFailure(HTTPClient.java:215)
	at org.apache.iceberg.rest.HTTPClient.execute(HTTPClient.java:299)
	at org.apache.iceberg.rest.BaseHTTPClient.post(BaseHTTPClient.java:88)
	at org.apache.iceberg.rest.RESTSessionCatalog$Builder.create(RESTSessionCatalog.java:771)
	at org.apache.iceberg.CachingCatalog$CachingTableBuilder.lambda$create$0(CachingCatalog.java:264)
[...]

PR #2589, opened 2025-09-17
Unify create/loadTable call paths

dimas-b:
In preparation for implementing sending non-credential config to REST Catalog clients for #2207 this PR unifies calls paths for create/load table operations.

This change does not have any differences in authorization.

This change is not expecte to have any material behaviour differences to the affected code paths.

The main idea is to consolidate decision-making for that to include into REST responses and use method parameters like `EnumSet<AccessDelegationMode> delegationModes` for driving those decisions.


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -758,11 +788,16 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
...
       TableIdentifier tableIdentifier,
       TableMetadata tableMetadata,
+      EnumSet<AccessDelegationMode> delegationModes,
       Set<PolarisStorageActions> actions,
-      String snapshots,
       Optional<String> refreshCredentialsEndpoint) {
     LoadTableResponse.Builder responseBuilder =
         LoadTableResponse.builder().withTableMetadata(tableMetadata);
+
+    if (!delegationModes.contains(VENDED_CREDENTIALS)) {
```
adutra, line 797: Probably safer when more delegation modes get supported:

```suggestion
    if (delegationModes.isEmpty()) {
```

dimas-b: I used `.isEmpty()` in other cases, but in this method the code below specifically performs credential vending, so I think this check is correct :)

We can certainly refactor when we support remote S3 request signing. WDYT?

adutra: Sure, WFM.

dimas-b:
Rebased and resolved conflicts.

@adnanhemani : please take a look to double check that I did not break any of your changes from #2480.

PR #2602, opened 2025-09-18
Generate Request IDs (if not specified); Return Request ID as a Header

adnanhemani:
This PR makes two changes:
1. Generate a Request ID for all incoming requests. It will prioritize using a Request ID specified by the incoming request - but if one is not specified, then a new one is generated. The Request ID is therefore always available in the `ContainerRequestContext` downstream.
2. Creates a new ContainerResponseFilter where the Request ID is added as a header to all responses.

I'm having some trouble with the naming of the classes, please do help suggest better names!

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
@@ -22,32 +22,35 @@ package org.apache.polaris.service.tracing;
 import com.google.common.annotations.VisibleForTesting;
 import jakarta.enterprise.context.ApplicationScoped;
 import java.util.UUID;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 @ApplicationScoped
 public class RequestIdGenerator {
-  private volatile String baseDefaultUuid = UUID.randomUUID().toString();
-  static final AtomicLong COUNTER = new AtomicLong();
   static final Long COUNTER_SOFT_MAX = Long.MAX_VALUE / 2;
 
-  public String generateRequestId() {
-    long currentCounter = COUNTER.incrementAndGet();
-    String requestId = baseDefaultUuid + "_" + currentCounter;
-    if (currentCounter >= COUNTER_SOFT_MAX) {
-      synchronized (this) {
-        if (COUNTER.get() >= COUNTER_SOFT_MAX) {
-          // If we get anywhere close to danger of overflowing Long (which, theoretically,
-          // would be extremely unlikely) rotate the UUID and start again.
-          baseDefaultUuid = UUID.randomUUID().toString();
-          COUNTER.set(0);
-        }
-      }
+  record State(String uuid, long counter) {
+
+    State() {
+      this(UUID.randomUUID().toString(), 1);
     }
-    return requestId;
+
+    String requestId() {
+      return String.format("%s_%019d", uuid, counter);
+    }
+
+    State increment() {
+      return counter >= COUNTER_SOFT_MAX ? new State() : new State(uuid, counter + 1);
+    }
+  }
+
+  final AtomicReference<State> state = new AtomicReference<>(new State());
+
+  public String generateRequestId() {
+    return state.getAndUpdate(State::increment).requestId();
   }
 
   @VisibleForTesting
   public void setCounter(long counter) {
-    COUNTER.set(counter);
+    state.set(new State(state.get().uuid, counter));
   }
 }
--- a/runtime/service/src/test/java/org/apache/polaris/service/tracing/RequestIdGeneratorTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/tracing/RequestIdGeneratorTest.java
@@ -43,7 +43,8 @@ public class RequestIdGeneratorTest {
 
     assertThat(requestId).isNotNull();
     assertThat(requestId).matches(this::isValidRequestIdFormat);
-    assertThat(requestId).contains("_1"); // First call should increment counter to 1
+    // First call should increment counter to 1
+    assertThat(extractCounterFromRequestId(requestId)).isEqualTo(1);
   }
 
   @Test
@@ -62,22 +63,22 @@ public class RequestIdGeneratorTest {
 
   @Test
   void testCounterIncrementsSequentially() {
-    requestIdGenerator.setCounter(0);
+    //    requestIdGenerator.setCounter(0);
 
     String firstId = requestIdGenerator.generateRequestId();
     String secondId = requestIdGenerator.generateRequestId();
     String thirdId = requestIdGenerator.generateRequestId();
 
-    assertThat(firstId).endsWith("_1");
-    assertThat(secondId).endsWith("_2");
-    assertThat(thirdId).endsWith("_3");
+    assertThat(extractCounterFromRequestId(firstId)).isEqualTo(1);
+    assertThat(extractCounterFromRequestId(secondId)).isEqualTo(2);
+    assertThat(extractCounterFromRequestId(thirdId)).isEqualTo(3);
   }
 
   @Test
   void testCounterRotationAtSoftMax() {
     // Set counter close to soft max
     long softMax = RequestIdGenerator.COUNTER_SOFT_MAX;
-    requestIdGenerator.setCounter(softMax - 1);
+    requestIdGenerator.setCounter(softMax);
 
     String beforeRotation = requestIdGenerator.generateRequestId();
     String afterRotation = requestIdGenerator.generateRequestId();
@@ -87,9 +88,9 @@ public class RequestIdGeneratorTest {
     String afterUuidPart = afterRotation.substring(0, afterRotation.lastIndexOf('_'));
     assertNotEquals(beforeUuidPart, afterUuidPart);
 
-    // After rotation, counter should be reset and UUID should be different
-    assertThat(beforeRotation).endsWith("_" + softMax);
-    assertThat(afterRotation).endsWith("_1"); // Counter reset to 1 (after increment from 0)
+    assertThat(extractCounterFromRequestId(beforeRotation)).isEqualTo(softMax);
+    // Counter reset to 1 (after increment from 0)
+    assertThat(extractCounterFromRequestId(afterRotation)).isEqualTo(1);
   }
 
   @Test
@@ -98,7 +99,8 @@ public class RequestIdGeneratorTest {
 
     String requestId = requestIdGenerator.generateRequestId();
 
-    assertThat(requestId).endsWith("_101"); // Should increment from set value
+    // Should increment from set value
+    assertThat(extractCounterFromRequestId(requestId)).isEqualTo(100);
   }
 
   private boolean isValidRequestIdFormat(String str) {
@@ -112,5 +114,9 @@ public class RequestIdGeneratorTest {
     } catch (IllegalArgumentException e) {
       return false;
     }
+  }
+
+  private long extractCounterFromRequestId(String requestId) {
+    return Long.parseLong(requestId.split("_")[1]);
   }
 }


runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Inject LoggingConfiguration loggingConfiguration;
+
+  @Override
+  public void filter(ContainerRequestContext rc) {
+    var requestId = rc.getHeaderString(loggingConfiguration.requestIdHeaderName());
+    if (requestId == null) {
+      requestId = UUID.randomUUID().toString();
+    }
+    MDC.put(REQUEST_ID_KEY, requestId);
```
adutra: I'm admittedly nit-picking a bit 😅 but I'd argue that the `MDC.put` statement should stay in `LoggingMDCFilter` since it's about logging.

What should be moved here is just the `rc.setProperty` call.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdResponseFilter.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.apache.polaris.service.tracing.RequestIdFilter.REQUEST_ID_KEY;
+
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerResponseContext;
+import jakarta.ws.rs.container.ContainerResponseFilter;
+import jakarta.ws.rs.ext.Provider;
+
+@Provider
+public class RequestIdResponseFilter implements ContainerResponseFilter {
+  private static final String REQUEST_ID_HEADER = "X-Request-Id";
```
adutra: The X- prefix is deprecated, FYI. See: 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers

But more importantly, why not use the header name used for incoming requests, `Polaris-Request-Id`?

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
```diff
@@ -0,0 +1,52 @@
...
+@Provider
+public class RequestIdFilter implements ContainerRequestFilter {
+
+  public static final String REQUEST_ID_KEY = "requestId";
+
+  @Inject LoggingConfiguration loggingConfiguration;
+
+  @Override
+  public void filter(ContainerRequestContext rc) {
+    var requestId = rc.getHeaderString(loggingConfiguration.requestIdHeaderName());
```
adutra, line 44: I'm wondering if we should move this setting out of the `polaris.log` space? Maybe `polaris.telemetry.request-id-header-name`, wdyt? Because it's not only for logs, it's also visible in telemetry.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
adnanhemani: Sure, I can agree to that :) Will adjust in this revision.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdResponseFilter.java
adnanhemani: TIL "X-" is deprecated! Thanks!

We can - I thought this was more of a standardized name prior. Let me change this.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
adnanhemani: I thought about it, but didn't want to make a breaking change here. Maybe we can tackle this later when we're closer to a major version upgrade? Happy to put that in a different PR :)

adutra: Fair enough, let's keep it in `polaris.log` for now.

runtime/service/src/test/java/org/apache/polaris/service/admin/RequestIdHeaderTest.java
```diff
@@ -0,0 +1,121 @@
...
+@QuarkusTest
+@TestProfile(RequestIdHeaderTest.Profile.class)
+public class RequestIdHeaderTest {
+  public static class Profile implements QuarkusTestProfile {
+    @Override
+    public Map<String, String> getConfigOverrides() {
+      return Map.of(
+          "polaris.log.request-id-header-name",
+          REQUEST_ID_HEADER,
+          "polaris.bootstrap.credentials",
```
adutra: This doesn't look right. There is no Quarkus configuration named `polaris.bootstrap.credentials`.

adnanhemani: That's odd - I found it from a different [test](https://github.com/apache/polaris/blob/main/runtime/service/src/test/java/org/apache/polaris/service/admin/RealmHeaderTest.java#L48) and the [README](https://github.com/apache/polaris/blob/main/README.md?plain=1#L105) as well.

I've removed it and the test still works 🤔 I can investigate this later as this is not the point of this PR.

adutra: The README is correct, but the other test is wrong indeed.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    if (requestId == null) {
+      requestId = UUID.randomUUID().toString();
```
dimas-b: Is exhausting the randomness pool a concern? @snazy : WDYT?

adutra: Good point. Also: `UUID.randomUUID()` is in theory a blocking call (happens only when the entropy source is empty though), and therefore should be avoided in filters.

dimas-b: If the request ID does not have to be a UUID (I'm a bit out-of-date on this) it may be worth using a per-node (or per-thread) UUID (allocated one per restart) plus a simple counter.

adutra: A few suggestions:

1. Use `org.apache.polaris.ids.impl.SnowflakeIdGeneratorImpl#idToTimeUuid`
2. Use a simple `AtomicLong` counter

dimas-b: Also UUID v7 may be worth considering (optional, for follow-up): https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-04.html#name-uuid-version-7

adutra: UUID v7 is time-ordered and database-friendly, but I don't think it would address the concern around entropy exhaustion 🤔

dimas-b: true... I wonder if we could leverage https://github.com/apache/polaris/tree/main/persistence/nosql/idgen :thinking: (edit: I see that @adutra suggested the same above :sweat_smile: )

flyrain: I might be wrong, but are we sure that `UUID.randomUUID()` is blocking call and in what chance will it exhaust the system entropy? 
It uses `/dev/urandom` by default starting from JDK1.8 and never blocked, https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/.

adnanhemani: I've added the `RequestIdGenerator`, which is a very, very simple ID generator - since we don't require the level of complexity that has been made in the `SnowflakeIdGenerator`. It's just a quick implementation of @dimas-b's suggestion above:

> If the request ID does not have to be a UUID (I'm a bit out-of-date on this) it may be worth using a per-node (or per-thread) UUID (allocated one per restart) plus a simple counter.

I'm very hesitant to introduce a complete dependency on the NoSql Persistence for the Service module and so creating the `RequestIdGenerator` is my minor effort to avoid doing that. I'm sure there may be a better reason for introducing this dependency in the future, but I don't want to sidetrack the simple goal (and simple requirements) of this PR with whether we should introduce this dependency or not.

adutra: > are we sure that UUID.randomUUID() is blocking call and in what chance will it exhaust the system entropy?

I've never got a satisfying answer for that question 😄 

Most commenters stress the fact that `dev/urandom` never blocks. 

This is certainly true, but doesn't address the fact that the entropy pool might get exhausted at some point, in which case your UUIDs will have very poor randomness.

Furthermore, the `UUID.randomUUID()` call is explicitly considered blocking by BlockHound:

https://github.com/reactor/BlockHound/issues/157

I _think_ that this is due to the fact that `new SecureRandom()` does some magic to select the random numbers provider, and even if the default provider doesn't block (it plugs into `/dev/urandom`), you still can configure your JVM with a different provider that could block.

runtime/service/src/test/java/org/apache/polaris/service/admin/RequestIdHeaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+  @Test
+  public void testRequestIdHeaderNotSpecified() {
+    Map<String, String> headers = Map.of(REALM_HEADER, REALM);
+    assertThat(sendRequest(headers)).matches(this::isValidDefaultUUID);
+  }
+
+  @Test
+  public void testRequestIdHeaderNotSpecifiedAndCounterExhausted() {
+    requestIdGenerator.setCounter(Long.MAX_VALUE / 2 + 1);
```
adutra: This test is actually testing the internals of `RequestIdGenerator`. I'd prefer to move this to a new `RequestIdGeneratorTest` class.

runtime/service/src/test/java/org/apache/polaris/service/admin/RequestIdHeaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+    String currentRequestIdBase = requestId.split("_")[0];
+    String uuidBase = currentRequestIdBase;
+
+    requestId = sendRequest(headers);
+    assertThat(requestId).matches(this::isValidDefaultUUID);
+    currentRequestIdBase = requestId.split("_")[0];
+    assertThat(currentRequestIdBase).isNotEqualTo(uuidBase);
+  }
+
+  private boolean isValidDefaultUUID(String str) {
```
adutra: Same here, I'm not sure we need to verify that request IDs have any particular structure. For all intents and purposes, the ID is an opaque string and it doesn't matter how it's generated, as long as it is unique.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import jakarta.ws.rs.ext.Provider;
+import org.apache.polaris.service.config.FilterPriorities;
+import org.apache.polaris.service.logging.LoggingConfiguration;
+
+@PreMatching
+@ApplicationScoped
+@Priority(FilterPriorities.REQUEST_ID_FILTER)
+@Provider
+public class RequestIdFilter implements ContainerRequestFilter {
+  @Inject RequestIdGenerator requestIdGenerator;
```
adutra: nit: move this line down to line 40

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.polaris.service.tracing;
+
+import com.google.common.annotations.VisibleForTesting;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+@ApplicationScoped
+public class RequestIdGenerator {
+  private static String BASE_DEFAULT_UUID = UUID.randomUUID().toString();
```
adutra:
```suggestion
  private static String baseDefaultUuid = UUID.randomUUID().toString();
```

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+  static final AtomicLong COUNTER = new AtomicLong();
+  static final Long COUNTER_SOFT_MAX = Long.MAX_VALUE / 2;
+
+  public String generateRequestId() {
+    String requestId = BASE_DEFAULT_UUID + "_" + COUNTER.incrementAndGet();
+    if (COUNTER.get() >= COUNTER_SOFT_MAX) {
```
adutra: There are many issues here: 

1. `COUNTER.get()` will not necessarily return what `COUNTER.incrementAndGet()` returned in the previous line. 
2. Many threads may enter this if block in parallel, and will all try to assign the `BASE_DEFAULT_UUID` field.
3. `BASE_DEFAULT_UUID` is not volatile nor atomic, so the value write on line 38 may not be immediately visible to other threads.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
flyrain: Here’s a post that shows how to work around BlockHound: https://stackoverflow.com/a/75687886/933856.

That said, we probably shouldn’t make decisions based on a single anecdote or assumptions about JVM configurations. What if a user runs Polaris on their own JVM and it breaks? That scenario is very likely. And what if a future JVM introduces a breaking change?

Do we need to worry about that now? Probably not. It feels like a premature optimization at this stage.

adutra: Quarkus doesn't use BlockHound – I cited BlockHound as an example. But Quarkus does have a mechanism to detect blocking calls in non-blocking contexts, we already had the problem with `RealmContextFilter`.

flyrain: I posted that link to demonstrate that people disagree with BlockHound on whether it is blocking call. Does Quarkus consider also `UUID.randomUUID()` as a blocking-call? If not, we got another reason to optimize it later.

runtime/service/src/test/java/org/apache/polaris/service/admin/RequestIdHeaderTest.java
adnanhemani: Done in the next revision!

runtime/service/src/test/java/org/apache/polaris/service/admin/RequestIdHeaderTest.java
adnanhemani: Got it - I'm still testing the format in `RequestIdGenerator` (just to make sure that what we expect to happen is still happening), but removed it from here.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
adnanhemani: TIL, thanks!

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
adnanhemani: Sorry, you are right and I'm embarrassed that I didn't catch these. I've fixed this in the next revision using an AtomicBoolean as a way to track whether a reset is in progress. Please let me know if I've missed anything else.

adnanhemani (commented):
Responded to comments from @adutra . Please let me know if there is anything else!

runtime/service/src/main/java/org/apache/polaris/service/config/FilterPriorities.java
```diff
@@ -21,6 +21,7 @@
 import jakarta.ws.rs.Priorities;
 
 public final class FilterPriorities {
+  public static final int REQUEST_ID_FILTER = Priorities.AUTHENTICATION - 101;
```
flyrain, line 24: nit:  -> `REALM_CONTEXT_FILTER - 1 `  to make it more readable?

flyrain (approved):
+1

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+
+package org.apache.polaris.service.tracing;
+
+import com.google.common.annotations.VisibleForTesting;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+@ApplicationScoped
+public class RequestIdGenerator {
```
adutra, line 28: Sorry, this is still faulty. Consider the alternative implementation below instead.

```java
@ApplicationScoped
public class RequestIdGenerator {

  private static class State {

    final String uuid = UUID.randomUUID().toString();
    long counter = 1;

    String requestId() {
      return String.format("%s:%019d", uuid, counter);
    }

    State increment() {
      if (counter == Long.MAX_VALUE) {
        return new State();
      } else {
        counter++;
        return this;
      }
    }
  }

  private final AtomicReference<State> state = new AtomicReference<>(new State());

  public String generateRequestId() {
    return state.updateAndGet(State::increment).requestId();
  }
}
```

adutra: If we are worried about the performance of `requestId()` the below version would be faster:

```java
    String requestId() {
      StringBuilder sb = new StringBuilder(56);
      sb.append(uuid).append(':');
      String string = String.valueOf(counter);
      sb.append("0".repeat(19 - string.length()));
      sb.append(string);
      return sb.toString();
    }
```

I'm using zero-padded counters because this makes IDs with the same base UUID lexicographically sortable.

sfc-gh-ahemani: I'm in favor of the zero-padded counters, happy to make the change - but can you please explain what about the current implementation makes it faulty?

I appreciate your alternative implementation, however, I think this is actually susceptible to a race condition as well. If two threads call `updateAndGet` at the same time, it is still entirely possible that (given `increment` returns back the same `State` object) both threads complete calling `updateAndGet` prior to calling `requestId()` and as a result return back the same counter value. 

If I'm reading the documentation correctly, `updateAndGet` only provide guarantees when the object is immutable or the the function is side-effect free - which I don't believe is the case with the `State` object.

While I'm sure this sort of race condition is highly unlikely to happen, I think that we will not hit that in the implementation as it is written currently - which is why I'm curious as to what you spotted.

adnanhemani: I'm in favor of the zero-padded counters, happy to make the change - but can you please explain what about the current implementation makes it faulty?

I appreciate your alternative implementation, however, I think this is actually susceptible to a race condition as well. If two threads call `updateAndGet` at the same time, it is still entirely possible that (given `increment` returns back the same `State` object) both threads complete calling `updateAndGet` prior to calling `requestId()` and as a result return back the same counter value. 

If I'm reading the documentation correctly, `updateAndGet` only provide guarantees when the object is immutable or the the function is side-effect free - which I don't believe is the case with the `State` object.

While I'm sure this sort of race condition is highly unlikely to happen, I think that we will not hit that in the implementation as it is written currently - which is why I'm curious as to what you spotted.

adutra: It's not faulty in the sense that it would throw an exception, but it has some corner cases:

1. Threads A and B increment the counter past `COUNTER_SOFT_MAX` (line 34)
2. Threads A and B evaluate `currentCounter >= COUNTER_SOFT_MAX` to true (line 36)
3. Thread A enters the synchronized block, resets `baseDefaultUuid` and `COUNTER`
4. Thread B enters the synchronized block, only to find that `currentCounter >= COUNTER_SOFT_MAX` is now false => the object's monitor was acquired for nothing.

Also:

1. Thread A increments counter to `COUNTER_SOFT_MAX` (line 34)
2. Thread B also increments counter, but enters the synchronized block first
3. Thread B resets `baseDefaultUuid` and `COUNTER`
4. Thread A reads `baseDefaultUuid` (line 35), finds the new value
5. Thread A creates a request ID with the new `baseDefaultUuid` but the old counter value => not dangerous, but weird.

That being said, you are absolutely right that my revised version above is faulty as well. I apologize for that, I wrote it too quickly. What I had in mind is the typical pattern where each invocation of the update function returns a new, immutable state. The below implementation is correct, and even much simpler since the state can be a record:

```java
@ApplicationScoped
public class RequestIdGenerator {

  record State(String uuid, long counter) {

    State() {
      this(UUID.randomUUID().toString(), 1);
    }

    String requestId() {
      return String.format("%s:%019d", uuid, counter);
    }

    State increment() {
      return counter == Long.MAX_VALUE ? new State() : new State(uuid, counter + 1);
    }
  }

  final AtomicReference<State> state = new AtomicReference<>(new State());

  public String generateRequestId() {
    return state.getAndUpdate(State::increment).requestId();
  }
}
```

dimas-b: Do we want to handle `long` overflow in the version with `getAndUpdate()`?

adnanhemani: Thanks for this, I agree with the edge cases. I don't think the first scenario matters all that much to be fair but I can agree that the second scenario is a bit odd. I will pivot to your suggestion instead.

My main hesitation was regarding pressure on the GC processes since we are consistently creating new record objects and throwing them away per request - but after some back-of-the-envelope calculations, I think the amount of data we are creating under the publicized TPS that we have on blogs, etc. is too little to notice any pressure on the GC process. So this should not be an issue IMO.

flyrain (approved):

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
```diff
@@ -0,0 +1,56 @@
...
+    State() {
+      this(UUID.randomUUID().toString(), 1);
+    }
+
+    String requestId() {
+      return String.format("%s_%019d", uuid, counter);
+    }
+
+    State increment() {
+      return counter >= COUNTER_SOFT_MAX ? new State() : new State(uuid, counter + 1);
```
adutra, line 42: Why `COUNTER_SOFT_MAX` ? There is no risk of overflowing here.

runtime/service/src/test/java/org/apache/polaris/service/tracing/RequestIdGeneratorTest.java
```diff
@@ -0,0 +1,122 @@
...
+      assertThat(generatedIds).doesNotContain(requestId);
+      generatedIds.add(requestId);
+    }
+
+    assertThat(generatedIds).hasSize(1000);
+  }
+
+  @Test
+  void testCounterIncrementsSequentially() {
+    //    requestIdGenerator.setCounter(0);
```
adutra, line 66: nit: dangling comment.

runtime/service/src/test/java/org/apache/polaris/service/tracing/RequestIdGeneratorTest.java
```diff
@@ -0,0 +1,122 @@
...
+    String afterUuidPart = afterRotation.substring(0, afterRotation.lastIndexOf('_'));
+    assertNotEquals(beforeUuidPart, afterUuidPart);
+
+    assertThat(extractCounterFromRequestId(beforeRotation)).isEqualTo(softMax);
+    // Counter reset to 1 (after increment from 0)
+    assertThat(extractCounterFromRequestId(afterRotation)).isEqualTo(1);
+  }
+
+  @Test
+  void testSetCounterChangesNextGeneratedId() {
```
adutra, line 97: Not sure this test has any value, `setCounter` is not meant to be called outside of tests, and besides, I think we can remove that method.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  final AtomicReference<State> state = new AtomicReference<>(new State());
+
+  public String generateRequestId() {
+    return state.getAndUpdate(State::increment).requestId();
+  }
+
+  @VisibleForTesting
+  public void setCounter(long counter) {
```
adutra, line 53: You don't need this method, `RequestIdGenerator.state` is already package-private.

runtime/service/src/test/java/org/apache/polaris/service/tracing/RequestIdGeneratorTest.java
```diff
@@ -0,0 +1,122 @@
...
+    String requestId = requestIdGenerator.generateRequestId();
+
+    assertThat(requestId).isNotNull();
+    assertThat(requestId).matches(this::isValidRequestIdFormat);
+    // First call should increment counter to 1
+    assertThat(extractCounterFromRequestId(requestId)).isEqualTo(1);
+  }
+
+  @Test
+  void testGenerateRequestId_ReturnsUniqueIds() {
```
adutra, line 51: This test could be made multi-threaded.

adutra:
In order to move things forward, and since @flyrain already approved, I'm going to merge this PR now even if my latest comments are still open. I will address them myself in a follow-up PR.

PR #2603, opened 2025-09-18
[OpenAPI Modification] Return created objects

adnanhemani:
As per the ML thread [here](https://lists.apache.org/thread/q7q0rrsmw5gcqv30g4hr9ffq3gtr72yk), this PR introduces the change to return all objects that are created within their respective API calls. 

~Additionally, the `revokeGrantFromCatalogRole` API will now return 204 (No Content) on a successful call rather than 201 (Created).~


spec/polaris-management-service.yml
```diff
@@ -787,7 +799,7 @@ paths:
             schema:
               $ref: "#/components/schemas/RevokeGrantRequest"
       responses:
-        201:
+        204:
```
dimas-b: This is technically a breaking change per our current "evolution" approach: https://polaris.apache.org/releases/1.0.1/evolution/#semantic-versioning

The change itself is meaningful, but if we make it, I suppose we'd have to do a major version bump (which is ok from my POV)..

adnanhemani: Removed as per ML - and will introduce it separately.

adutra (approved):
LGTM. I wasn't too worried about the 201 -> 204 change, but I agree that it's better to leave this change out.

flyrain (dismissed):
LGTM

HonahX (dismissed):
LGTM! +1 on adding a line in the CHANGELOG to reveal this. We should also update our python cli to print the entity, we can do that in a follow-up : )

singhpk234 (approved):

dimas-b:
vote thread: https://lists.apache.org/thread/rv3sqno6orfyx7h3f2llb8lkq00zofg8

PR #2615, opened 2025-09-18
Non sts s3

dimas-b:


runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
```diff
@@ -266,23 +287,45 @@ public LoadTableResponse doTestCreateTable(RESTCatalog restCatalog) throws IOExc
...
 
     restCatalog.dropTable(id);
     assertThat(restCatalog.tableExists(id)).isFalse();
     return loadTableResponse;
   }
 
   @ParameterizedTest
-  @ValueSource(booleans = {true, false})
-  public void testAppendFiles(boolean pathStyle) throws IOException {
+  // TODO:  @CsvSource("true,")
```
adutra, line 317: What is the issue with these test cases?

dimas-b: Forgot to uncomment :smile: :facepalm: - fixed

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -794,10 +794,6 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
     LoadTableResponse.Builder responseBuilder =
```
flyrain, line 794: It's a bit clutter that this method return a builder, so that caller need to invoke `build()`. The code would be cleaner if it returns a `LoadTableResponse`. Not a blocker though.

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
```diff
@@ -266,23 +287,45 @@ public LoadTableResponse doTestCreateTable(RESTCatalog restCatalog) throws IOExc
...
+          .containsEntry(
+              REFRESH_CREDENTIALS_ENDPOINT,
+              "v1/" + catalogName + "/namespaces/test-ns/tables/t1/credentials");
+      assertThat(loadTableResponse.credentials()).hasSize(1);
+    } else {
+      assertThat(loadTableResponse.config()).doesNotContainKey(SECRET_ACCESS_KEY);
+      assertThat(loadTableResponse.config()).doesNotContainKey(ACCESS_KEY_ID);
+      assertThat(loadTableResponse.config()).doesNotContainKey(REFRESH_CREDENTIALS_ENDPOINT);
+      assertThat(loadTableResponse.credentials()).isEmpty();
+    }
```
flyrain, line 309: Can we move the validation logic to the caller(line 238)? I think it makes more sense to validate them, as the `AccessDelegationMode` is one of the test parameter. 
Also I think we need to validate the ENDPOINT in case `AccessDelegationMode` is `VENDED_CREDENTIALS`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: Maybe :thinking: but this PR merely changes the impl. of this method. I did not mean to refactor related code to minimize the amount of changes. It's a `private` method, we can adjust it any time if current return type become a nuisance :slightly_smiling_face:

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
dimas-b: Endpoint is validated on line 296

dimas-b: Moved credential asserts to top-level test methods :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
flyrain: I'm fine with a followup, as it's not introduced by this PR.

eric-maynard (approved):

PR #2616, opened 2025-09-19
[OpenAPI Modification] Change revokeGrantFromCatalogRole successful status to 204 (No Content)

adnanhemani:
As per ML [thread](https://lists.apache.org/thread/q7q0rrsmw5gcqv30g4hr9ffq3gtr72yk). This should cause a major version upgrade - as such, do not merge until Polaris 2.x.x is ready.


dimas-b:
If I'm not mistaken this change stems from https://lists.apache.org/thread/wbmvf8s0sdwhfn8p91pwqsxolxd2btj4

I am personally not convinced that changing the response code is worth the possible ripple effects on the client side.

I believe it might be worth a dedicated `dev` ML discussion to highlight the rationale for the response code change as the previous email discussion does not appear to indicate why the new response code is preferable, but mentions it only as a possible alternative.

adnanhemani:
Issue for this here: #2657

dimas-b:
More specific discussion thread: https://lists.apache.org/thread/bhd1srzks0pz0odoqmm87gfyj2yz2k41

PR #2623, opened 2025-09-19
virtualenv: wider version range

snazy:
see https://github.com/apache/polaris/pull/2614#issuecomment-3311484328


snazy:
Updated to `virtualenv = ">=20.33.0,<20.35.0"` (see [poetry changes](https://python-poetry.org/history/)) to not conflict with #2560 (poetry bump to 2.2.0)

PR #922, opened 2025-01-31
Event Listeners

andrew4699:
Implementation of event listeners discussed [here](https://lists.apache.org/thread/03yz5wolkvy8l7rbcwjnqdq1bl8p065v).

I decided to keep this implementation generic and not take a dependency on Jakarta Events nor Vertx busses. It's easy to extend this, either within Polaris or in an external PolarisEventListener, and handle events however one wishes.

Some high level notes:
- PolarisEventListener is the main interface with all the event methods such as `onBeforeRequestRateLimited`
- DefaultPolarisEventListener is an empty implementation which allows users to only partially implement event handlers
- `polaris.events.type` is the config that lets you specify your event listener implementation

PR #2629, opened 2025-09-19
Fix & enhancements to the Events API hierarchy

adutra:
Summary of changes:

- Turned `PolarisEventListener` into an interface to facilitate implementation / mocking
- Added missing `implements PolarisEvent` to many event records
- Removed unused method overrides
- Added missing method overrides to `TestPolarisEventListener`

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
@@ -91,9 +91,9 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
 
   @Override
   public Response listCatalogs(RealmContext realmContext, SecurityContext securityContext) {
-    polarisEventListener.onBeforeListCatalog(new CatalogsServiceEvents.BeforeListCatalogEvent());
+    polarisEventListener.onBeforeListCatalogs(new CatalogsServiceEvents.BeforeListCatalogsEvent());
     Response resp = delegate.listCatalogs(realmContext, securityContext);
-    polarisEventListener.onAfterListCatalog(new CatalogsServiceEvents.AfterListCatalogEvent());
+    polarisEventListener.onAfterListCatalogs(new CatalogsServiceEvents.AfterListCatalogsEvent());
     return resp;
   }
 
--- a/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisEventListener.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisEventListener.java
@@ -32,518 +32,523 @@ import org.apache.polaris.service.events.PrincipalsServiceEvents;
  * Represents an event listener that can respond to notable moments during Polaris's execution.
  * Event details are documented under the event objects themselves.
  */
-public abstract class PolarisEventListener {
+public interface PolarisEventListener {
   /** {@link BeforeLimitRequestRateEvent} */
-  public void onBeforeLimitRequestRate(BeforeLimitRequestRateEvent event) {}
+  default void onBeforeLimitRequestRate(BeforeLimitRequestRateEvent event) {}
 
   /** {@link IcebergRestCatalogEvents.BeforeCommitTableEvent} */
-  public void onBeforeCommitTable(IcebergRestCatalogEvents.BeforeCommitTableEvent event) {}
+  default void onBeforeCommitTable(IcebergRestCatalogEvents.BeforeCommitTableEvent event) {}
 
   /** {@link IcebergRestCatalogEvents.AfterCommitTableEvent} */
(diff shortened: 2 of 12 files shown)


adutra:
cc @adnanhemani

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisEventListener.java
```diff
@@ -29,287 +29,287 @@
  * Represents an event listener that can respond to notable moments during Polaris's execution.
  * Event details are documented under the event objects themselves.
  */
-public abstract class PolarisEventListener {
+public interface PolarisEventListener {
```
eric-maynard, line 35: This was extensively debated at the time, but see [this comment](https://github.com/apache/polaris/pull/922#discussion_r1988472892) and related ones for why this is an ABC rather than an interface.

flyrain: Thanks for flagging it, @eric-maynard. I still hold my point, that this should be ABC, not interface.

adutra: Thank you both for reviving this discussion. I've re-read the old thread carefully, and I still do not understand what the problem is with using an interface rather than an abstract class. 

Can you please summarize your concerns here for clarity?

If it's about binary compatibility when adding more methods to the API, I'm afraid an interface and an ABC provide roughly the same linkage risks. From [JLS 13.4.16](https://docs.oracle.com/javase/specs/jls/se21/html/jls-13.html#jls-13.4.16) (emphasis mine):

> adding a default method is a binary-compatible change because it does not introduce errors at link time, even if it introduces errors at compile time or invocation time. In practice, **the risk of accidental clashes occurring by introducing a default method are similar to those associated with adding a new method to a non-final class**.

In contrast, the advantages of using interfaces rather than abstract classes in public APIs are abundantly documented. From _Effective Java, 3rd Edition_ (J. Bloch, 2017) (emphasis mine):

> Item 20: Prefer interfaces to abstract classes
> Java has two mechanisms to define a type that permits multiple implementations: interfaces and abstract classes. Since the introduction of default methods for interfaces in Java 8 [JLS 9.4.3], both mechanisms allow you to provide implementations for some instance methods. A major difference is that to implement the type defined by an abstract class, a class must be a subclass of the abstract class. **Because Java permits only single inheritance, this restriction on abstract classes severely constrains their use as type definitions**. Any class that defines all the required methods and obeys the general contract is permitted to implement an interface, regardless of where the class resides in the class hierarchy.

flyrain: Yes, binary compatibility is the concern. The upstream changes shouldn't break downstream event listeners. However, looking over the PR, it looks like all methods are now default methods, which is fine. Let's also ensure that any new methods added in the future are default as well.

Could we add a test to enforce this? It might also be helpful to include a comment on the class to explicitly state that all new methods should be default.

adutra: Hi @flyrain thanks for your flexibility 🙂 

I will add the comment and the test as suggested. 

I'd note that this kind of check would be a good fit for the `revapi` plugin – the same one we use in Iceberg to track API breaking changes. But I think Polaris is still a young project, and revapi would flag too many API changes that would be otherwise permitted by our current evolution rules:

https://github.com/apache/polaris/blob/b1eb9bd81da10b8cc7bb3f200ceb22b9990642bb/site/content/in-dev/unreleased/evolution.md?plain=1#L81-L82

So, let's stick with a small test for now 👍

adutra: Added a comment and tests as suggested.

I also re-organized the interface methods into "sections" for clarity, I hope that's OK.

The tests check that all methods are default and well-formed, and also that all events have corresponding methods in the interface. 

I added a similar check to `TestPolarisEventListener` (it was lacking many methods).

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -91,9 +91,9 @@ public Response updateCatalog(
 
   @Override
   public Response listCatalogs(RealmContext realmContext, SecurityContext securityContext) {
-    polarisEventListener.onBeforeListCatalog(new CatalogsServiceEvents.BeforeListCatalogEvent());
+    polarisEventListener.onBeforeListCatalogs(new CatalogsServiceEvents.BeforeListCatalogsEvent());
```
adutra, line 94: When I added tests for method compliance I noticed that this event was different from other "list" events because the object type being listed wasn't in the plural.

adutra:
@flyrain @eric-maynard is the code looking better now?

PR #2633, opened 2025-09-19
Azure: Fix azure expires at prefix for the credentials refresh

singhpk234:
### About the change 

For the credential refresh endpoint for the ADLS its expected the expiration time should be sent with the prefix of `AZURE_SAS_TOKEN_EXPIRES_AT_MS_PREFIX` which presently Polaris doesn't sends as the course of that the CredentialProvider responsible for the refresh fails the assertion here https://github.com/apache/iceberg/blob/main/azure/src/main/java/org/apache/iceberg/azure/adlsv2/VendedAdlsCredentialProvider.java#L123
and hence not making it usable.

P.S I think we should also use constants for iceberg SDK in the StorageAccessProperty, but this is an orthogonal discussion for now. 

### Testing 
Add a small UT

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
@@ -182,7 +182,7 @@ public class AzureCredentialsStorageIntegration
       Instant expiresAt,
       Optional<String> refreshCredentialsEndpoint) {
     AccessConfig.Builder accessConfig = AccessConfig.builder();
-    handleAzureCredential(accessConfig, sasToken, storageDnsName);
+    handleAzureCredential(accessConfig, sasToken, storageDnsName, expiresAt);
     accessConfig.put(
         StorageAccessProperty.EXPIRATION_TIME, String.valueOf(expiresAt.toEpochMilli()));
     refreshCredentialsEndpoint.ifPresent(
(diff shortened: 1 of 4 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -193,8 +193,11 @@ static AccessConfig toAccessConfig(
   }
 
   private static void handleAzureCredential(
-      AccessConfig.Builder config, String sasToken, String host) {
+      AccessConfig.Builder config, String sasToken, String host, Instant expiresAt) {
     config.putCredential(StorageAccessProperty.AZURE_SAS_TOKEN.getPropertyName() + host, sasToken);
+    config.putCredential(
+        StorageAccessProperty.AZURE_SAS_TOKEN_EXPIRES_AT_MS_PREFIX.getPropertyName() + host,
```
dimas-b, line 199: Could you add a test case for this to `AzureCredentialsStorageIntegrationTest`?

dimas-b (commented):
Thanks for catching this, @singhpk234 !

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
singhpk234: Done :) !

dimas-b (dismissed):

dimas-b (commented):
Could you also add a note to `CHANGELOG.md`?

singhpk234:
ACK, seems like we haven't frozen the CHANGELOG for 1.1, i put a PR for this here 

- https://github.com/apache/polaris/pull/2635

will add the CHANGELOG for this bug fix on the unreleased section so we can use this in prep of 1.2.

dimas-b (approved):

singhpk234:
Thank you @dimas-b !

PR #2634, opened 2025-09-19
Auth: reorganize internal authentication components

adutra:
This PR contains no functional and no user-facing change. It is merely a refactor to better organize auth code.

Summary of changes:

- Moved all internal authentication components to the `org.apache.polaris.service.auth.internal` package and subpackages
- Reduced visibility of utility classes
- Renamed `TokenBroker` class hierarchy to stick to the naming standard: `<Algorithm>JWTBroker`
- Introduced `@PolarisImmutable` whenever appropriate
- Removed unused `NoneTokenBrokerFactory` (we already have `DisabledOAuth2ApiService`)
- Removed unused `TokenBrokerFactoryConfig`


adutra:
FYI @flyrain @dennishuo @collado-mike

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/service/OAuthError.java
```diff
@@ -16,21 +16,25 @@
...
-/**
- * This is a union of configuration settings of all token brokers.
- *
- * @see TokenBrokerFactory
- */
-public interface TokenBrokerFactoryConfig {
-  @Nullable
-  String file();
+public enum OAuthError {
+  invalid_request("The request is invalid"),
```
dimas-b, line 23: nit: UPPER case or CamelCase for values? (usage sites have changes anyway).

dimas-b (approved):
LGTM 👍 Thanks, @adutra !

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/service/OAuthError.java
adutra: This enum comes from here:

https://github.com/apache/polaris/blob/20febdaede19fb7c46e120652fdd1a262c2138e4/runtime/service/src/main/java/org/apache/polaris/service/auth/OAuthTokenErrorResponse.java#L27-L45

I just renamed to `OAuthError` to avoid the name clash with `java.lang.Error`.

But otherwise, the enum constants were all lower case (snake case).

dimas-b: I'm ok either way :)

adutra: In fact, these are error codes defined in the RFC: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2
Since the enum constants are included as-is in the response, I believe we should not modify them.

dimas-b: Good point!

adutra:
Going to merge this now since it's just a refactoring, if anyone has concerns please reach out to me or open an issue.

Issue #2620, opened 2025-09-19
Polaris CLI should support specifying the realm


rukor2:
### Is your feature request related to a problem? Please describe.

This is useful if you have a non-default realm name. At the moment, it is impossible to use the CLI as it can either not generate a token or if you generate the token externally, you cant use it because it will complain that a Polaris-Realm header is not supplied.

### Describe the solution you'd like

Allow for the CLI to support custom realms

dimas-b:
Thanks for reporting this, @rukor2 !

I added the "bug" label because Polaris Servers and  the admin tool support realms as a first-class concept, therefore the CLI should be able to interact with any realm that a Polaris Server has.

binarycat0:
I briefly checked the existing CLI doc  https://polaris.apache.org/in-dev/unreleased/command-line-interface/ 
It seems there is no options to specify REALM really.

I will try to help with that. As I understand Polaris CLI is written on python based on `click` framework so it should **NOT** be hard.

dimas-b:
"should NOT be hard"? 😉

binarycat0:
I checked closely and as I see it uses custom ArgumentParser instead of `click`, but anyway it should NOT be hard

binarycat0:
I can't find any reference to REALM in API specs - https://polaris.apache.org/in-dev/unreleased/polaris-api-specs/ 🤷‍♂️ 

The [REALM](https://polaris.apache.org/in-dev/unreleased/realm/) documentation says - `a realm is resolved from request headers` but I can't find it in codebase 🤔 

@rukor2 could you share an example how to use it in API calls?

dimas-b:
Clients indicate the desired realm via a header, e.g. as in [obtain-token.sh](https://github.com/apache/polaris/blob/596239cdba8b18431aa9a0e5ec2f317fb8365b2b/getting-started/assets/polaris/obtain-token.sh#L28)

binarycat0:
I see that token `scope` is hardcoded also

https://github.com/apache/polaris/blob/596239cdba8b18431aa9a0e5ec2f317fb8365b2b/client/python/cli/polaris_cli.py#L135

dimas-b:
`PRINCIPAL_ROLE:ALL` is fine for CLI to use for now, I think. We can make it user-settable later.

rukor2:
> I can't find any reference to REALM in API specs - https://polaris.apache.org/in-dev/unreleased/polaris-api-specs/ 🤷‍♂️
> 
[...]

PR #2643, opened 2025-09-21
fix(enhancement): add --context-realm, --context-header-name root args to polaris CLI

binarycat0:
fix(enhancement): #2620 

## fix(enhancement) 
- add `--context-realm`, `--context-header-name` root args to `./polaris` CLI 

## refactoring
- minor typing hints fixes
- move ApiClient builder logic from `polaris_cli.PolarisCli` to external `api_client_builder.ApiClientBuilder` class
  - motivation: enhance code granularity


client/python/cli/options/parser.py
```diff
@@ -55,26 +55,25 @@ class Parser(object):
...
+            hint="realm header is defining header name to use as context header name. if not defined will be used default from Polaris server",
+            default=DEFAULT_REALM_HEADER
+        ),
         Argument(Arguments.PROFILE, str, hint="profile for token-based authentication"),
         Argument(Arguments.PROXY, str, hint="proxy URL"),
         Argument(Arguments.DEBUG, bool, hint="Enable debug mode"),
     ]
 
     @staticmethod
     def _build_parser() -> argparse.ArgumentParser:
```
eric-maynard, line 76: This method's body is gone, but it wasn't removed

client/python/cli/options/parser.py
```diff
@@ -114,6 +113,9 @@ def recurse_options(subparser, options: List[Option]):
                     )
                     recurse_options(children_subparser, option.children)
 
+        parser = TreeHelpParser(description="Polaris CLI")
+        add_arguments(parser, Parser._ROOT_ARGUMENTS)
```
eric-maynard, line 117: `add_arguments` isn't used for the `_ROOT_ARGUMENTS` because it affected the help string IIRC, is that somehow fixed?

client/python/cli/polaris_cli.py
```diff
@@ -67,8 +55,10 @@ def execute(args=None):
             command = Command.from_options(options)
             command.execute()
         else:
-            client_builder = PolarisCli._get_client_builder(options)
-            with client_builder() as api_client:
+            client_builder = ApiClientBuilder(
+                options, direct_authentication=PolarisCli.DIRECT_AUTHENTICATION_ENABLED
+            )
+            with client_builder.build() as api_client:
```
eric-maynard: Since we immediately call `build` I wonder if we really need the factory pattern here?

client/python/cli/polaris_cli.py
```diff
@@ -119,101 +109,6 @@ def _try_print_exception(e):
                 f"Exception when communicating with the Polaris server. {e}{os.linesep}"
             )
 
-    @staticmethod
```
eric-maynard, line 122: Refactoring these out is wise; this file has gotten really big when it should be a thin entry point into the CLI

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -81,6 +83,11 @@ If the `--host` and `--port` options are not provided, the CLI will default to c
 
 Alternatively, the `--base-url` option can be used instead of `--host` and `--port`, but both options cannot be used simultaneously. This allows specifying arbitrary Polaris URLs, including HTTPS ones, that have additional base prefixes before the `/api/*/v1` subpaths.
 
+If your polaris server is configured to use multy realm (or not default realm), you can use `--realm` option to specify the realm you want to use. If `--realm` is not provided, the CLI will check the `REALM` environment variable. If neither is provided, the CLI will not use realm context header.
```
eric-maynard:
```suggestion
If your Polaris server is configured to use a realm other than the default, you can use the `--realm` option to specify a realm. If `--realm` is not provided, the CLI will check the `REALM` environment variable. If neither is provided, the CLI will not send the realm context header.
```

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -81,6 +83,11 @@ If the `--host` and `--port` options are not provided, the CLI will default to c
...
+Also, if you polaris server uses custom realm header name, you can use `--realm-header` option to specify the header name. If `--realm-header` is not provided, the CLI will check the `REALM_HEADER` environment variable. If neither is provided, the CLI will use default header name `Polaris-Realm`.
```
eric-maynard:
```suggestion
Also, if your Polaris server uses a custom realm header name, you can use the `--realm-header` option to specify it. If `--realm-header` is not provided, the CLI will check the `REALM_HEADER` environment variable. If neither is provided, the CLI will use default header name `Polaris-Realm`.
```

client/python/cli/constants.py
```diff
@@ -369,6 +371,9 @@ class Namespaces:
 CLIENT_ID_ENV = "CLIENT_ID"
 CLIENT_SECRET_ENV = "CLIENT_SECRET"
 CLIENT_PROFILE_ENV = "CLIENT_PROFILE"
+REALM_ENV = "REALM"
+REALM_HEADER_ENV = "REALM_HEADER"
```
eric-maynard: I think the ability to specify a custom realm header is valuable, though I wonder if it is worth a new argument. In the end, the CLI can't expose every custom behavior and this may be in the realm (ha) of things we route people to using the REST API directly for.

If we do want the CLI to support a custom header name, perhaps we could consider an API like:
```bash
polaris ... --realm '{"my-header": "realm_name"}'
```

I'll be interested to see what other reviewers think about CLI support for custom header names.

client/python/cli/options/parser.py
binarycat0: I will double-check it. I moved it here because code looked duplicated.

eric-maynard: For sure, it looks very duplicated. But it has always been this way and I remember there was definitely a reason... but to be honest I don't recall exactly what it is. My guess is something to do with the help command. If we find out, it's a good idea to add a comment here :)

client/python/cli/constants.py
binarycat0: Thank you for your feedback, I will take a look on the comments. Will wait other feedbacks.
BTW This feature was requested in the issue #2620


`'{"my-header": "realm_name"}'`

IMO This variant can be, but ContextRealmHeaderName has a default value and this version requires doing extra movements 😃

client/python/cli/options/parser.py
binarycat0: I checked this, but I didn't get the comment. 
The removed code is a duplicate of the inner `add_arguments` method but with skipping some options of class `cli.options.option_tree.Argument`. 
The `add_arguments` is more comprehensive and covers more options.

I discovered this while experimenting with adding new `--realm` argument. In the early version I assumed that this option can be multiple.

client/python/cli/options/parser.py
binarycat0: I double-checked this. Everything works well, and all tests are passing 👌

client/python/cli/options/parser.py
eric-maynard: Makes sense, just wanted to check this was intentional. Thanks!

client/python/cli/constants.py
eric-maynard: Perhaps could also just add a `--header` option?

binarycat0: Yeah, I think you are right. Since the option name `--header-name` has nothing to do with `realm`, so we can use just `--header` as an option name 🤔

binarycat0:
off-topic: @eric-maynard Do you know why current CLI does not use https://click.palletsprojects.com/en/stable/ ? Licensing concerns?

client/python/cli/api_client_builder.py
```diff
@@ -0,0 +1,187 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
```
MonkeyCanCode, line 2: Minor: missing begin and end "#" for the license header.

client/python/cli/command/profiles.py
```diff
@@ -25,11 +25,13 @@
 
 from cli.command import Command
 from cli.constants import (
-    Subcommands,
-    DEFAULT_HOSTNAME,
-    DEFAULT_PORT,
-    CONFIG_DIR,
-    CONFIG_FILE,
+  Subcommands,
```
MonkeyCanCode, line 28: NIT: use `make client-lint` for code formatting (it calls `ruff format`)

MonkeyCanCode:
> off-topic: @eric-maynard Do you know why current CLI does not use https://click.palletsprojects.com/en/stable/ ? Licensing concerns?

Click is BSD-3-Clause. So no license concern here.

MonkeyCanCode (dismissed):

client/python/cli/command/profiles.py
binarycat0: I did, no changes in this part.

client/python/cli/api_client_builder.py
binarycat0: Header added automatically. Fixed.

binarycat0:
Hello colleagues, could you check this? 
@eric-maynard @MonkeyCanCode @holdenk 
Thank you.

MonkeyCanCode:
Changes looks good to me. I think @eric-maynard had some concerns earlier. Let's wait for him to check more.

> Hello colleagues, could you check this? 
> @eric-maynard @MonkeyCanCode @holdenk 
> Thank you.

client/python/cli/command/profiles.py
```diff
@@ -72,11 +74,15 @@ def _create_profile(self, name: str) -> None:
...
             host = input(f"Polaris Host [{DEFAULT_HOSTNAME}]: ") or DEFAULT_HOSTNAME
             port = input(f"Polaris Port [{DEFAULT_PORT}]: ") or DEFAULT_PORT
+            realm = input(f"Polaris Context Realm: ")
+            header = input(f"Polaris Context Header Name [{DEFAULT_HEADER}]: ") or DEFAULT_HEADER
             profiles[name] = {
                 "client_id": client_id,
                 "client_secret": client_secret,
                 "host": host,
                 "port": port,
+                Arguments.REALM: realm,
```
jbonofre, line 84: Just curious (nit): why using a different syntax here ? Why not using directly `realm` ?

client/python/cli/command/profiles.py
```diff
@@ -104,6 +110,8 @@ def _update_profile(self, name: str) -> None:
             current_client_secret = profiles[name].get("client_secret")
             current_host = profiles[name].get("host")
             current_port = profiles[name].get("port")
+            current_realm = profiles[name].get(Arguments.REALM)
```
jbonofre, line 113: Same question here.

binarycat0: Hello, I decided to use the predefined variable instead of string literals, and if everyone will be ok with such approach to apply the same in the follow-up PR. 

The idea is to avoid using string literals.

dimas-b:
@eric-maynard : Could you clarify whether your "request changes" is still critical? I hope your previous concerns have been addressed :slightly_smiling_face:

client/python/generate_clients.py
```diff
@@ -164,7 +164,7 @@ def generate_polaris_catalog_client() -> None:
             "-g",
             "python",
             "-o",
-            CLIENT_DIR,
+            str(CLIENT_DIR),
```
HonahX, line 167: Just curious, did you encounter any issue that need this fix?

client/python/cli/command/profiles.py
HonahX: +1 on using constants instead of pure string. I think this shall be a good practice in the future .

client/python/cli/api_client_builder.py
```diff
@@ -0,0 +1,188 @@
...
+
+class ApiClientBuilder:
+    def __init__(self, options: Namespace, *, direct_authentication: bool = False):
+        self.conf = BuilderConfig(options)
+        self.direct_auth_enabled = direct_authentication
+
+    def _get_token(self):
+        header_params = {"Content-Type": "application/x-www-form-urlencoded"}
+        if self.conf.header and self.conf.realm:
+            header_params[self.conf.header] = self.conf.realm
```
HonahX, line 129: In a follow-up, is it possible to add some tests for the new arguments?

HonahX (approved):
LGTM! Thanks for adding these arguments!

It seems all the concerns has been addressed. Let's give another day for some final comments from other reviewers

eric-maynard (approved):
LGTM! Thanks for the great contribution.

One nit: I think this is more of a `feat:` than a `fix:` :)

client/python/cli/options/parser.py
```diff
@@ -55,26 +55,25 @@ class Parser(object):
...
             hint="access token for token-based authentication",
         ),
+        Argument(
+            Arguments.REALM,
+            str,
+            hint="realm to use with header. if not defined will be used default realm from Polaris server. read more: https://polaris.apache.org/releases/1.1.0/configuration/",
+            default=None
+        ),
+        Argument(
+            Arguments.HEADER,
```
eric-maynard, line 65: When I suggested a `--header` flag, what I meant is that you could pass in any arbitrary header here such as `--header "X-foo: bar"`, not that the flag `--header` would be a good name for something which is just the realm context header name.

eric-maynard: To add more color here, the way you would use `--header` to achieve a custom realm would be either like `--header "Polaris-Realm: foo"` or `--header "My-Custom-Realm-Header-Name: bar"`. This supports custom realm header names without the need to introduce a flag specifically for that.

The current implementation is sort of the worst of both worlds, because we have added a generic-sounding `header` flag that actually does something really specific to just one type of header.

client/python/cli/command/profiles.py
jbonofre: Personally I would have used two PRs: in this one, we should use pure string to be consistent, and have another PR to use constants everywhere. 
But that's ok, let's do a follow-up PR.

jbonofre (approved):
LGTM. I took the liberty to change the title, because it's not a fix imho but more feature/improvement.

client/python/generate_clients.py
binarycat0: Hello. Yes, I did face an issue with impossibility to execute generating specs in case of incorrectness of types, cause I assume `__repr__` method of CLIENT_DIR object returns not string but PosixPath object. 

```
Traceback (most recent call last):
  File "/polaris/client/python/generate_clients.py", line 328, in <module>
    main()
  File "/polaris/client/python/generate_clients.py", line 324, in main
    build()
  File "/polaris/client/python/generate_clients.py", line 303, in build
    generate_polaris_management_client()
  File "/polaris/client/python/generate_clients.py", line 133, in generate_polaris_management_client
    subprocess.check_call(
  File "/usr/lib/python3.12/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['openapi-generator-cli', 'generate', '-i', '/polaris/client/python/spec/polaris-management-service.yml', '-g', 'python', '-o', PosixPath('/polaris/client/python'), '--additional-properties=packageName=polaris.management', '--additional-properties=apiNamePrefix=polaris', '--additional-properties=pythonVersion=3.9', '--additional-properties=generateSourceCodeOnly=true', '--skip-validate-spec', '--ignore-file-override', '/polaris/client/python/.openapi-generator-ignore', '--global-property=apiDocs=false,modelDocs=false,modelTests=false,apiTests=false']' returned non-zero exit status 1.

```

PR #2650, opened 2025-09-22
Enhancement : adding support for Aurora postgres AWS IAM authentication

fabio-rizzo-01:
Adding ability to connect to AWS Aurora Postgres using IAM authentication (password-less DB connection).

Tests carried out:

1.  Deployed image to EKS cluster with [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IAM Role for Service Account) enabled. Polaris is able to connect to aurora database

2. Deployed image to minikube to validate username/password auth to Postgress DB still works

Both test cases were successful.

Added example configuration in properties file

### DIFF
--- a/runtime/defaults/src/main/resources/application.properties
+++ b/runtime/defaults/src/main/resources/application.properties
@@ -132,20 +132,6 @@ quarkus.devservices.enabled=false
 quarkus.rds.sync-client.type=apache
 
 
-# AWS IAM postgres example
-#quarkus.datasource.jdbc.url=jdbc:postgresql://aurora-cluster-url:6160/polaris
-#quarkus.datasource.jdbc.additional-jdbc-properties.wrapperPlugins=iam
-#quarkus.datasource.username=dbusername
-#quarkus.datasource.active=true
-#quarkus.datasource.db-kind=postgresql
-#quarkus.datasource.jdbc.additional-jdbc-properties.ssl=true
-#quarkus.datasource.jdbc.additional-jdbc-properties.sslmode=require
-#quarkus.datasource.credentials-provider=aws
-
-#quarkus.rds.credentials-provider.aws.use-quarkus-client="true"
-#quarkus.rds.credentials-provider.aws.username=dbusername
-#quarkus.rds.credentials-provider.aws.hostname=aurora-cluster-url
-#quarkus.rds.credentials-provider.aws.port=6160
 
 polaris.file-io.type=default
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### Upgrade Notes
 
+- Amazon RDS plugin enabled, this allows polaris to connect to AWS Aurora PostgreSQL using IAM authentication.
 - The EclipseLink Persistence implementation has been deprecated since 1.0.0 and will be completely removed
   in 1.3.0 or in 2.0.0 (whichever happens earlier).
 


persistence/relational-jdbc/build.gradle.kts
```diff
@@ -35,6 +35,9 @@ dependencies {
 
   implementation(libs.smallrye.common.annotation) // @Identifier
   implementation(libs.postgresql)
+  implementation("software.amazon.awssdk:url-connection-client")
```
singhpk234: why do we need url-connection-client specifically ? can this not work with apache http client ?
per this should work - https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-rds.html#_configuring_rds_clients

runtime/defaults/src/main/resources/application.properties
```diff
@@ -122,9 +122,27 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
-# polaris.persistence.type=relational-jdbc
-
+#polaris.persistence.type=relational-jdbc
 polaris.secrets-manager.type=in-memory
+#disable localstack
+#if set to true it will try to start localstack at build and run time for the local environment
+quarkus.devservices.enabled=false
+
+
+#AWS IAM postgres example
```
singhpk234: wondering linking this additionally https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-rds.html#_configuration_reference would be helpful too

runtime/defaults/src/main/resources/application.properties
```diff
@@ -122,9 +122,27 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
+#quarkus.datasource.jdbc.additional-jdbc-properties.wrapperPlugins=iam
+#quarkus.datasource.username=dbusername
+#quarkus.datasource.active=true
+#quarkus.datasource.db-kind=postgresql
+#quarkus.datasource.jdbc.additional-jdbc-properties.ssl=true
+#quarkus.datasource.jdbc.additional-jdbc-properties.sslmode=require
+#quarkus.datasource.credentials-provider=aws
+
+#quarkus.rds.credentials-provider.aws.use-quarkus-client="true"
+#quarkus.rds.credentials-provider.aws.username=dbusername
```
singhpk234: seems like this is from this section - https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-rds.html#_iam_authentication_support_with_quarkus_credential_provider ? 
did you had to change the trust relationships etc and IAM policy etc, would be helpful adding  something about that here too !

wondering if we can move this to https://polaris.apache.org/in-dev/unreleased/metastores/ ? cc @flyrain @dimas-b  thoughts ?

singhpk234 (commented):
Thanks @fabio-rizzo-01 for taking this up !

persistence/relational-jdbc/build.gradle.kts
fabio-rizzo-01: either should work yes, you want me to change it?

runtime/defaults/src/main/resources/application.properties
fabio-rizzo-01: do you want me to add a comment with more details around the policy for the pod? or in the docs? let me know what details you want me to add

dimas-b: Instructions in docs would be nice, but IMHO, it can be a separate PR

persistence/relational-jdbc/build.gradle.kts
dimas-b: How can the user select which client to use?

dimas-b: My impression from the quarkiverse docs is that the type of HTTP client is a build time decision that is implied by the dependency included in the build (`url-connection-client` vs. `apache-client`). The latter is [already](https://github.com/apache/polaris/blob/4e2facbb4b743b4d54c1d39aaf6fc290960eca7e/runtime/service/build.gradle.kts#L92) included in Polaris.

My preference would be to use the `apache-client` for Aurora too and avoid `url-connection-client`.

WDYT? CC: @snazy

fabio-rizzo-01: it is at build time, I'll update to apache if this is the preferred way.

dimas-b: +1 to using `apache-client` from my side (reconfirming) :slightly_smiling_face:

fabio-rizzo-01: I'll update the PR

persistence/relational-jdbc/build.gradle.kts
```diff
@@ -35,6 +35,9 @@ dependencies {
 
   implementation(libs.smallrye.common.annotation) // @Identifier
   implementation(libs.postgresql)
+  implementation("software.amazon.awssdk:apache-client")
+  implementation(platform(libs.quarkus.amazon.services.bom))
+  implementation("io.quarkiverse.amazonservices:quarkus-amazon-rds")
```
dimas-b: I guess these deps are not required for building the JDBC code itself. They probably need to be in `runtime/common`.

The idea is to keep the JDBC persistence lean, but add RDS where drivers would normally be added. WDYT?

(sorry, I missed this in my previous review)

dimas-b: Note: `apache-client` is probably already present in "runtime" modules.

fabio-rizzo-01: updated

dimas-b:
Thanks for your contribution, @fabio-rizzo-01 !

runtime/defaults/src/main/resources/application.properties
```diff
@@ -122,9 +122,29 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
 # polaris.persistence.type=eclipse-link
 # polaris.persistence.type=in-memory-atomic
 polaris.persistence.type=in-memory
-# polaris.persistence.type=relational-jdbc
-
+#polaris.persistence.type=relational-jdbc
```
flyrain: minor:  this change seems unnecessary

runtime/defaults/src/main/resources/application.properties
```diff
@@ -122,9 +122,29 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
 polaris.secrets-manager.type=in-memory
+#disable localstack
```
flyrain: nit: we usually have a space between `#` and content.

fabio-rizzo-01:
> Thanks a lot @fabio-rizzo-01 for working on it. I'm glad this was figured out. The PR looks good to me overall! My only concern is that users won't be able to figure out how database configure work from a bag of comments from the file `application.properties`. Here are complexity users will face:
> 
> 1. Two options to configure JDBC db. 1. Username/Password 2. AWS IAM AuthN
> 2. Two `application.properties` files(admin/server) to configure them. And they have to be consistent with each other.
> 
> I'd propose to move these configures to a new section under "Relational JDBC" of the page https://polaris.apache.org/in-dev/unreleased/metastores/. With certain description like `there are two options. option 1 ..., option 2 ...`. WDYT?

shall we merge this and I'll create another PR for the docs?

dimas-b:
I think doing doc changes in a follow-up PR makes sense :+1:

runtime/defaults/src/main/resources/application.properties
dimas-b: true, @fabio-rizzo-01 : would you mind reverting this line for the sake of proper tranceability should someone need to see when and why it was added?

runtime/defaults/src/main/resources/application.properties
dimas-b: @fabio-rizzo-01 : could you fix this together with the other minor comment?

fabio-rizzo-01: PR updated

flyrain:
Thanks, @fabio-rizzo-01, for addressing the comment. I'm good with a follow-up doc PR. In that case, I suggest we remove the configuration changes from both `application.properties` files. We add a new jdbc connection option in this PR. We got two options now. Users might confused by which option to choose, how they set it, especially since neither file includes a complete setup for both.

fabio-rizzo-01:
> Thanks, @fabio-rizzo-01, for addressing the comment. I'm good with a follow-up doc PR. In that case, I suggest we remove the configuration changes from both `application.properties` files. We add a new jdbc connection option in this PR. We got two options now. Users might confused by which option to choose, how they set it, especially since neither file includes a complete setup for both.

I can't remove all the properties these 2 need to be there: `quarkus.rds.sync-client.type=apache quarkus.devservices.enabled=false`

the app won't run, because it needs to know which client to use to connect otherwise it defaults to the url-connection-client that is not there and for the second one it will try to start up localstack. So I can remove the db config but not these 2.

singhpk234:
> I can't remove all the properties these 2 need to be there: quarkus.rds.sync-client.type=apache quarkus.devservices.enabled=false

This seems reasonable if this is a build time property and all others are runtime properties. my understaing we were concerned mostly from two ways to configure runtime properties for the datasources. 

Orthogonally it would be really nice to add a `CHANGELOG.md` entry for this ! Thank you @fabio-rizzo-01 for working on this one.

dimas-b:
Good point about CHANGELOG (I missed that :sweat_smile: ). @fabio-rizzo-01 : could you add a brief notice that the RDS plugin is now enabled and allows connecting to Aurora DB? I do not think it needs to be too verbose since doc changes are coming later.

fabio-rizzo-01:
> Good point about CHANGELOG (I missed that 😅 ). @fabio-rizzo-01 : could you add a brief notice that the RDS plugin is now enabled and allows connecting to Aurora DB? I do not think it needs to be too verbose since doc changes are coming later.

properties removed and updated changelog file

flyrain (approved):
+1 Thanks a lot @fabio-rizzo-01 for adding this!

PR #2653, opened 2025-09-23
Add fallback in case the VERSION table is not present

eric-maynard:
Since #1942, persistence implementations -- particularly the JDBC implementation -- have a concept of a schema "version" and Polaris versions are not necessarily coupled with schema versions. As a result, we can dynamically load the schema version at runtime to determine whether certain features are safe to enable. 

However, there may be users who are running a schema version from before the concept of schema versioning was even introduced -- and those users might currently see an error when Polaris attempts to load from the version table. This PR adds a best-effort fallback to attempt to mitigate the error they would see by defaulting to schema version 0.


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -736,6 +737,9 @@ static int loadSchemaVersion(DatasourceOperations datasourceOperations) {
       return schemaVersion.getFirst().getValue();
     } catch (SQLException e) {
       LOGGER.error("Failed to load schema version due to {}", e.getMessage(), e);
+      if (fallbackOnDoesNotExist && e.getMessage().contains("does not exist")) {
+        return SchemaVersion.MINIMUM.getValue();
+      }
```
sfc-gh-ygu: Could we use `42P01` to check if table isn't there? https://www.postgresql.org/docs/current/errcodes-appendix.html

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
```diff
@@ -25,6 +25,8 @@
 import org.apache.polaris.persistence.relational.jdbc.DatabaseType;
 
 public class SchemaVersion implements Converter<SchemaVersion> {
+  public static final SchemaVersion MINIMUM = new SchemaVersion(0);
```
flyrain, line 28: https://github.com/polaris-catalog/polaris/blob/eb7618bce8b4ecb259f0030d5d24458aa2d91343/persistence/relational-jdbc/src/main/resources/postgres/schema-v1.sql#L26-L26

JDBC impl. is only with the v1 schema. There is no v0 for jdbc implementations. Should we use `1` here?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
eric-maynard: I'm not sure if we want the check to be postgresql-specific; my hope is that by keeping the check generic the fallback can work for different databases

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
eric-maynard: I believe that's why we want to use 0 -- the v1 schema file includes the VERSION table, so the lack of a VERSION table indicates an earlier (pre-1) schema.

We don't use this MINIMUM number to choose a schema file, only the reverse -- the detect what schema version an already-running metastore was bootstrapped with.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
flyrain: that's fair. Is the string `does not exist` work across different database?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
flyrain: The v1 schema released with 1.0.x doesn't include the version table. 
Also, do we want to distinguish v0 and v1 in the code base? I'd prefer not.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Support for arbitrary databases is basically missing ATM, I think.

As for "supported" databases, could we add backend-specific  "does not exist" logic to `DatabaseType`?

polaris-core/src/main/java/org/apache/polaris/core/config/BehaviorChangeConfiguration.java
```diff
@@ -84,4 +84,14 @@ protected BehaviorChangeConfiguration(
                   + " credential vending.")
           .defaultValue(false)
           .buildBehaviorChangeConfiguration();
+
+  public static final BehaviorChangeConfiguration<Boolean> SCHEMA_VERSION_FALL_BACK_ON_DNE =
+      PolarisConfiguration.<Boolean>builder()
+          .key("SCHEMA_VERSION_FALL_BACK_ON_DNE")
```
dimas-b, line 90: suggestion (optional): `FALLBACK_JDBC_SCHEMA_VERSION` (and change the value to be an `int`.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
eric-maynard: I think we do need to distinguish them. For a given Polaris version e.g. 1.1.0, we have a notion of a v1 schema as defined by the v1 schema file and that contains the version table. If there's a metastore without the version table, it definitely can't be v1.

polaris-core/src/main/java/org/apache/polaris/core/config/BehaviorChangeConfiguration.java
eric-maynard: This feels similar to letting you break glass and force your own schema version, which not the intent here (and can be done by manipulating the VERSION table anyway). Really all we want to do is just capture the fact that the lack of a version table is implicitly version 0. The flag is only added out of an abundance of caution and out of recognition of the fact that the error handling is only best-effort.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
flyrain: I might miss something. What would be behavior difference between a 1.0 v1 schema and a 1.1 v1 schema? I think they should be identical. The problem we are trying to resolve is that the 1.0 v1 schema doesn't have a version table when 1.1 Polaris deployed. The system behavior should be the same as if there is a version table with version 1.

polaris-core/src/main/java/org/apache/polaris/core/config/BehaviorChangeConfiguration.java
dimas-b: fair point. My suggestion was optional anyway (mainly trying to simplify the code that uses it).

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Isolating database-specific logic for this change is the only concern I have with this PR.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
eric-maynard: You can have two different schema versions without any behavior difference between them, so I'm not sure I follow your question.

The problem we are trying to solve is identifying the schema version for a metastore without a schema version table. The schema version table starts at version 1. A metastore without this table is therefore a pre-1 version.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
eric-maynard: Is there a particular error code we want to catch in postgres? 

If so, perhaps `DatasourceOperations` would be a good place for this check, something in the vein of `isConstraintViolation`.

In my mind this check will go away after some time, once we can reasonably expect to stop supporting pre-1 schema versions. So if it's a best-effort check that only works for a metastore backend or two, that's okay with me.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
dimas-b: I do not think we have any formal requirement for the schema versions to be comparable to each other numerically.

Current JDBC Persistence code makes some numerical comparisons, but I tend to think that this is an implementation concern. As long as the interpretation of schema version numbers is consistent with the schemas themselves, the code is fine.

With that in mind, using `0` for missing schema information should produce correct runtime behaviour, as far as I can tell.

dimas-b:
@eric-maynard : please rebase to fix markdown CI :slightly_smiling_face:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
flyrain: Since we are sticked with Postgres anyway, for example,  https://github.com/polaris-catalog/polaris/blob/06d523116fd5f5b61f95d8660166e76c00fd2a3e/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java#L58-L58. I'm fine with  this to be Postgres specific.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
flyrain: We will need to use the schema version for things proposed here, https://lists.apache.org/thread/5d9rl1l2jflbbnrl12ofmczjbcw8qv89.  "Pre-1" and "1" are going to have the same behavior in terms of how we handling the missing column `location_without_schema`.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Following the `isConstraintViolation()` patterns SGTM :+1:

dimas-b (approved):

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/SchemaVersion.java
flyrain: It's a minor comment for me. Feel free to merge it as is.

singhpk234 (approved):
I also have a slight preference for 1 over 0, considering 1 being the starting version number, but totally fine if we wanna forward with 0, The check i think we might wanna add later is < 2 than if we want to not project column `location_without_schema` which is also fine !

I will merge this PR by EOD today if no blocking comment come up to keep the ball rolling for 1.2 

Thank you so much @eric-maynard !

PR #2654, opened 2025-09-23
Doc: Add breaking changes section for 1.1 release

singhpk234:
### About the change 

Some users while upgrading from Polaris 1.0 -> 1.1 recently reported issues because of missing schema table, this would be happening because the schema version table is not created in v1 script as the course of that JDBCBasePersistenceImpl is failing and hence the whole metastore 

Thank @flyrain for the deepdive : https://apache-polaris.slack.com/archives/C084XDM50CB/p1758368000877189

while there is already works on fixing this, while discussing with @flyrain they suggested having something in the release note would be really helpful which i agree, hence putting a docs pr.

### DIFF
--- a/site/content/downloads/_index.md
+++ b/site/content/downloads/_index.md
@@ -57,8 +57,8 @@ Apache Polaris 1.1.0-incubating was released on September 19th, 2025.
 - **Breaking changes**
   - Helm chart: the default value of the `authentication.tokenBroker.secret.symmetricKey.secretKey` property has changed 
     from `symmetric.pem` to `symmetric.key`.
-  - For migrations from 1.0.x to 1.1.x, users using JDBC persistence must ensure that they and wanting to continue using v1 schema,
-    run following SQL statement under `POLARIS_SCHEMA` to make sure version table exists:
+  - For migrations from 1.0.x to 1.1.x, users using JDBC persistence and wanting to continue using v1 schema, must ensure
+    that they, run following SQL statement under `POLARIS_SCHEMA` to make sure version table exists:
     ```sql 
     CREATE TABLE IF NOT EXISTS version (
        version_key TEXT PRIMARY KEY,


CHANGELOG.md
```diff
@@ -95,6 +95,19 @@ Apache Polaris 1.1.0-incubating was released on September 19th, 2025.
 - **Breaking changes**
   - Helm chart: the default value of the `authentication.tokenBroker.secret.symmetricKey.secretKey` property has changed
       from `symmetric.pem` to `symmetric.key`.
+  - For migrations from 1.0.x to 1.1.x, users using JDBC persistence and wanting to continue using v1 schema, must ensure that they,
```
dimas-b, line 98: What's the alternative to `continue using v1 schema`?

flyrain: One option is to migrate to v2 schema, which requires more than just the version table. We will need to write a tool for that per offline discussion with @singhpk234

flyrain:
Thanks a lot for adding this, @singhpk234 ! Thanks @dimas-b for the review.

CHANGELOG.md
singhpk234: Apologies, I missed responding here: 
I think there a couple of scenarios, for example 

- if we simply move to v2 from v1 by adding the column in the entity table we will have to backfill the location without schema column 
- if we started with v2 schema from day 1 we should be good or any realm created with v2 column starting should good too, but in the application layer for the optimization to correctly work there is no way to check this 

Bottom line being my understanding is the optimized siblling check for migrated schema version might not be reliable, i was wondering if we should capture some additional metadata in a sense that `is_migrated` ? default value 0 but incase the schema table exists set this value to 1, now when we do the optimized sibbling check we will just say this is not applicable if the metastore is migrated across version, though this too has shortcommings to be honest for example v2 -> v3 migration should still work, would really appreciate your thoughts

Issue #2024, opened 2025-07-10
Website docs issues after 1.0.0 release


snazy:
### Describe the bug

Some things there...

1. "main branch warning": <img width="1251" height="472" alt="Image" src="https://github.com/user-attachments/assets/1ef8d3a6-468a-43ef-b752-858b50b08bfe" />
2. left-side navigation missing "in-dev" + 0.9.0: <img width="503" height="595" alt="Image" src="https://github.com/user-attachments/assets/035f3727-1b73-4fd6-be91-808b428e6370" />
3. "in-dev docs warning" not highlighted: <img width="733" height="283" alt="Image" src="https://github.com/user-attachments/assets/44288dad-6b42-4238-83f7-808130c5bc71" />
4. "in-dev docs" missing 1.0.0 in left-side navigation: <img width="508" height="452" alt="Image" src="https://github.com/user-attachments/assets/8098d8a2-3ceb-43a7-9453-4d68af2938d8" />
5. 1.0.0 docs missing in-dev and 0.9.0 in left-side navigation: <img width="508" height="452" alt="Image" src="https://github.com/user-attachments/assets/9e53c6a2-9b54-4107-b96f-7eef37cbbbca" />
6. 0.9.0 docs missing 1.0.0 in left-side navigation: <img width="400" height="679" alt="Image" src="https://github.com/user-attachments/assets/bc3fff24-9f22-4c93-9155-f925498db6f3" />
7. 1.0.0 docs appear "under `Releases`" (others do not) + `Releases` page is empty: <img width="850" height="642" alt="Image" src="https://github.com/user-attachments/assets/8dee00e7-714a-4d34-81b0-59b25b5ffee8" />

MonkeyCanCode:
For `"in-dev docs warning" not highlighted`, this issue I am award and reported to @jbonofre as well. This is due to changes made in https://github.com/apache/polaris/pull/1967. However, those are rendering fines on local but not with the published site (I will need a bit more info regarding the doc publish process to continue debugging). Other than this, the newly added page for `Polaris Helm Chart` also doesn't show the icon but that is getting rendering correctly as well on local setup (the url refs by the image is also corrected from the published page).

snazy:
Looks like "In Development" is missing in the left-side menu.

adam-christian-software:
I'm cleaning up some old bugs and found this one.

I believe that Issue 3 still persists. The others have been finished. I have listed out the individual verifications and attached pictures at the bottom.

[...]

PR #2666, opened 2025-09-24
fix(downloads): Fix Small Downloads Page Issues

adam-christian-software:
# Context 
This change should finish up Issue #2024 . It does a small update to add a page title and a Table of Contents to the downloads section of the site.

# Screenshot
The before is below and the after is above.

image

### DIFF
--- a/site/content/downloads/_index.md
+++ b/site/content/downloads/_index.md
@@ -17,14 +17,17 @@
 # specific language governing permissions and limitations
 # under the License.
 #
+linkTitle: 'Downloads'
+title: 'Downloads'
+type: docs
 weight: 200
-toc_hide: true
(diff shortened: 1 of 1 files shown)


flyrain:
Thanks @adam-christian-software for the PR. Do we need a huge "Download" title on the top? It feels more clutter than before. 

The more I looked at this page, the more I feel it should be called `releases` instead of  `downloads`, as it includes helm chart and all release notes. But this is just my personal preference.

jbonofre:
@flyrain inthink I agree that the page should be named releases. We should just avoid to conflict with release menu entry we have about documentation.

adam-christian-software:
@flyrain - I agree & I'll do the rename. :)

PR #2670, opened 2025-09-24
fix(auth): let ServiceFailureException bubble up for proper HTTP status mapping during auth

HotSushi:
# Summary

Previously, ServiceFailureException from persistence layer timeouts was being
wrapped in AuthenticationFailedException, causing IOException during loadEntity
operations to return 401 Unauthorized instead of the correct 503 Service Unavailable.

This change allows ServiceFailureException to bubble up to IcebergExceptionMapper
which properly maps it to 503 Service Unavailable, while preserving the existing
behavior for other authentication exceptions.

- Modified AuthenticatingAugmentor to catch ServiceFailureException separately
- Added test case to verify ServiceFailureException bubbles up correctly
- Other RuntimeExceptions still wrapped in AuthenticationFailedException for 401

# Tests
✅  Added a unit test
✅ `./gradlew build` succeeds

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/auth/AuthenticatingAugmentorTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/auth/AuthenticatingAugmentorTest.java
@@ -29,6 +29,7 @@ import io.quarkus.security.runtime.QuarkusSecurityIdentity;
 import io.smallrye.mutiny.Uni;
 import java.security.Principal;
 import org.apache.iceberg.exceptions.NotAuthorizedException;
+import org.apache.iceberg.exceptions.ServiceFailureException;
 import org.apache.polaris.core.auth.PolarisPrincipal;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


runtime/service/src/test/java/org/apache/polaris/service/auth/AuthenticatingAugmentorTest.java
```diff
@@ -92,6 +93,29 @@ public void testAugmentAuthenticationFailure() {
...
+    ServiceFailureException serviceException =
+        new ServiceFailureException("Unable to fetch principal entity");
+    when(authenticator.authenticate(credential)).thenThrow(serviceException);
+
+    // When/Then
+    assertThatThrownBy(
+            () -> augmentor.augment(identity, Uni.createFrom()::item).await().indefinitely())
+        .isInstanceOf(ServiceFailureException.class)
+        .hasMessage("Unable to fetch principal entity")
+        .isNotInstanceOf(AuthenticationFailedException.class);
```
singhpk234: is this check required ?

flyrain (approved):
Thanks @HotSushi for the change. +1 on more accurate HTTP return code.

singhpk234 (dismissed):

runtime/service/src/test/java/org/apache/polaris/service/auth/AuthenticatingAugmentorTest.java
sfc-gh-sraikar: Updated~

singhpk234 (approved):

HonahX (approved):
LGTM!

flyrain:
Thanks @HotSushi for the fix! Thanks @singhpk234 @HonahX for the review.

PR #2672, opened 2025-09-24
Support S3 storage that does not have STS

dimas-b:
This change is backward compatible with old catalogs that have storage configuration for S3 systems with STS.

* Add new property to S3 storage config: `stsUnavailable` (defaults to "available").

* Do not call STS when unavailable in `AwsCredentialsStorageIntegration`, but still put other properties (e.g. s3.endpoint) into `AccessConfig`

* Fail create/load table requests if credential vending is requested but STS is not available.


CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 - Added a Management API endpoint to reset principal credentials, controlled by the `ENABLE_CREDENTIAL_RESET` (default: true) feature flag.
 
+- Added support for S3-compatible storage that does not have STS (use `stsUavailable: true` in catalog storage configuration)
```
dennishuo, line 44: Could be worth expanding this sentence with a bit more of what's in the yaml for easy reference if people are reading the notes without wanting to read the full spec change -- specifically to clarify that it disables credential vending rather than vending some kind of root credentials after skipping STS subscoping.

How does the behavior compare to setting `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION=true`?

dimas-b: `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION=true` disables all storage config in API responses. Disabling STS only removes credentials.

dimas-b: Re: message changes: I'm not sure it's worth adding YAML to CHANGELOG... Would it not make it too verbose? In the end CHANGELOG has notes for many releases :thinking: 

How about I update it with a CLI example when I add CLI support for the new config entry?

jbonofre: I agree that from an user standpoint, CLI (or `curl` :) ) and updating the configuration list on the documentation would be more than welcome.

dimas-b: I will certainly add a docs page about this to "getting started" once the impl. is approved :)

dimas-b: @dennishuo : are you ok with this path forward? :slightly_smiling_face:

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -79,43 +79,46 @@ public AccessConfig getSubscopedCreds(
...
-        .ifPresent(
-            i -> {
-              accessConfig.put(
-                  StorageAccessProperty.EXPIRATION_TIME, String.valueOf(i.toEpochMilli()));
-              accessConfig.put(
-                  StorageAccessProperty.AWS_SESSION_TOKEN_EXPIRES_AT_MS,
-                  String.valueOf(i.toEpochMilli()));
-            });
+
+    if (storageConfig.shouldUseSts()) {
```
singhpk234: [doubt] what if the client says it wan't to do `x-Iceberg-vended-credentials`, indication it wants credentials vending, but the storage doesn't support it, should we throw a 400 instead ?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -98,6 +98,18 @@ public URI getInternalEndpointUri() {
   /** Flag indicating whether path-style bucket access should be forced in S3 clients. */
   public abstract @Nullable Boolean getPathStyleAccess();
 
+  /**
+   * Flag indicating whether STS is available or not. It is modeled in the negative to simplify
+   * support for unset values ({@code null} being interpreted as {@code false}).
```
flyrain, line 103:
```suggestion
   * support for unset values ({@code null} being interpreted as {@code true}).
```

dimas-b: Javadoc is correct, I believe 😅  A null (unset) value would be treated as `StsUnavailable` being `false` , meaning that STS is _available_ (current default). Cf. `shouldUseSts()`.

I believe interpreting `null` as `false` is quite common.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: Ideally yes, and that was in my first revision of this PR. Unfortunately many existing use cases excessively request credential vending for `FILE` storage and would be broken if we were to enforce this logic "in general".

We cannot enforce it here, because "storage integration" code has valid call paths without credential vending, where it exposes non-credential config to clients (e.g. S3 endpoints).

I'd like to enforce stricter checks as you suggested, but due to risk of breaking existing clients (I saw that in our reg. tests), I propose to defer it. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
flyrain: You are right, sorry for the misleading. I was confused by two new methods introduced, `getStsUnavailable()`, `shouldUseSts()`, I think it'd be nice to only keep one of them. For example, the check brought up by @singhpk234 here https://github.com/apache/polaris/pull/2672#discussion_r2383187673, would be reasonable to apply when we check `getStsUnavailable()` purely. Wrapping `getStsUnavailable()` with `shouldUseSts()` makes it even more ambiguous, so that the logic become less obviously when we need to check if users ask for credential vending.

dimas-b: Ok, I removed `shouldUseSts()` from the config class. @flyrain : PTAL.

flyrain: Thanks for doing that, this LGTM.

CHANGELOG.md
dennishuo: Yeah, to clarify, I didn't mean to add too much, just maybe adding a `; credential vending will then be disabled` in the parentheses so that the change note would say

    (use `stsUavailable: true` in catalog storage configuration; credential vending will then be disabled);

In any case, non-blocking.

dennishuo: Incidentally, `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION` really was only intended to pertain to the "credential" aspect, so it's unfortunate that it sort of fell into fully disabling StorageConfig mix-ins even for the non-credential-related config. But we can sort that out some other time.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
singhpk234: I see, thanks for the explanation, I totally missed taking `FILE` into consideration ! IMHO its relatively easy to start with a stricter behaviour first and relax it later rather doing the other way as now there might be clients out there who would expect the behaviour to fail open and now expect 200, which we later plan to change to 400 anyways. 

I would have appreciated to discussed this more !

dimas-b: I'll make a follow-up PR for this presently.

dimas-b: @singhpk234 : thanks for the ping and sorry about missing a related bug (should be fixed in #2711)

PR #2674, opened 2025-09-24
Docs: Add more details about v1 schema user to upgrade from 1.0 to 1.1

singhpk234:
### About this change 
 
Seems like just adding schema is not sufficient 

https://apache-polaris.slack.com/archives/C084XDM50CB/p1758754322883619?thread_ts=1758368000.877189&cid=C084XDM50CB

Also add a note that not enable Optimized sibling check in this to warn with the potential incorrect answer because of backfill

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,7 +107,10 @@ Apache Polaris 1.1.0-incubating was released on September 19th, 2025.
       ON CONFLICT (version_key) DO UPDATE
                               SET version_value = EXCLUDED.version_value;
       COMMENT ON TABLE version IS 'the version of the JDBC schema in use';
+    
+    ALTER TABLE polaris_schema.entities ADD COLUMN IF NOT EXISTS location_without_scheme TEXT;
     ```
+    - Please don't enable [OPTIMIZED_SIBLING_CHECK](https://github.com/apache/polaris/blob/740993963cb41c2c1b4638be5e04dd00f1263c98/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java#L346) feature configuration, once the above SQL statements are run. As it may lead to incorrect behavior.
 - **Deprecations**
   - The property `polaris.active-roles-provider.type` is deprecated for removal.
   - The `ActiveRolesProvider` interface is deprecated for removal.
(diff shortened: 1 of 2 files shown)


HonahX:
Seems there's some issue with Github Actions, it cannot access the image base
```
failed to solve: docker.io/apache/spark:3.5.6-java17: failed to resolve source metadata for docker.io/apache/spark:3.5.6-java17: unexpected status from HEAD request to https://registry-1.docker.io/v2/apache/spark/manifests/3.5.6-java17: 401 Unauthorized
```

CHANGELOG.md
```diff
@@ -107,7 +107,10 @@ Apache Polaris 1.1.0-incubating was released on September 19th, 2025.
       ON CONFLICT (version_key) DO UPDATE
                               SET version_value = EXCLUDED.version_value;
       COMMENT ON TABLE version IS 'the version of the JDBC schema in use';
+    
+    ALTER TABLE polaris_schema.entities ADD COLUMN IF NOT EXISTS location_without_scheme TEXT;
     ```
+    - Please don't enable [OPTIMIZED_SIBLING_CHECK](https://github.com/apache/polaris/blob/740993963cb41c2c1b4638be5e04dd00f1263c98/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java#L346) feature configuration, once the above SQL statements are run. As it may lead to incorrect behavior.
```
dimas-b: `once the above SQL statements are run` - Is there a path forward for users to enable `OPTIMIZED_SIBLING_CHECK` after an upgrade?

dimas-b (commented):
Changes LGTM in general, just trying to clarity the mentioned issues.

CHANGELOG.md
flyrain: The backfill for this column has to be there. Otherwise, the feature may yield the wrong result.  The tricky part of backfilling is that it could NOT be easily done by a SQL command.

dimas-b: This means that `OPTIMIZED_SIBLING_CHECK` must be `false` if the Polaris data in the database came from an older version. `Once` in the comment above makes the impression that it is somehow "caused" by running the `ALTER` statement, while in fact the real reason is that data is missing because older servers did not store it.

Could you clarity the note, please?

singhpk234: sure thing ! please let me know how the current version reads.

dimas-b: better :) thx !

PR #2680, opened 2025-09-25
Implement OpaPolarisAuthorizer

sungwy:
Resolves https://github.com/apache/polaris/issues/138

RFC: https://docs.google.com/document/d/1HadMFygjbuZathZZPanO6cFVorx0Ju0FopkICxX1tCE/edit?tab=t.0
Mailing list discussion: https://lists.apache.org/thread/54qdbsxs3j7wwhv3tsccqj6qng5lqgmz

Some followup items highlighted as a part of this PR review:
- introduce `OpaHttpClientFactory` that utilizes `PoolingHttpClientConnectionManager` to get opa-http-client to be more production ready
- publish user-facing docs on OPA authorization
- publish json schema for docs and opa server checks
- introduce per-realm configuration support

PR #2697, opened 2025-09-26
Implement Finer Grained Operations and Privileges For Update Table

travis-bowen:
This implements finer grained operations and privileges for update table in a backwards compatible way as discussed on the mailing list. 

(Seems the mailing list split the thread in two but it's [here](https://lists.apache.org/thread/19tjcs8j9b7mzk8l64p2xxv358t8vtq3) and [here](https://lists.apache.org/thread/vk0mtx0jjnr42szrqnk5b53f9l43dn87))

The design doc is [here](https://docs.google.com/document/d/1LRB0LDevc6EQa0NWiLHi-jKmrwNA_XXWiHKfskZ6VwU/edit?usp=sharing)

The idea is that all the existing privileges and operations will work and continue to work even after this change. (i.e. TABLE_WRITE_PROPERTIES will still ensure update table is authorized even after these changes). 

However, because Polaris will now be able to identify each operation within an UpdateTable request and has a privilege model with inheritance that maps to each operation, users will now have the option of restricting permissions at a finer level if desired.


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -843,9 +843,74 @@ private UpdateTableRequest applyUpdateFilters(UpdateTableRequest request) {
...
+                          PolarisAuthorizableOperation.SET_TABLE_STATISTICS;
+                      case MetadataUpdate.RemoveStatistics removeStatistics ->
+                          PolarisAuthorizableOperation.REMOVE_TABLE_STATISTICS;
+                      case MetadataUpdate.RemovePartitionSpecs removePartitionSpecs ->
+                          PolarisAuthorizableOperation.REMOVE_TABLE_PARTITION_SPECS;
+                      default ->
+                          PolarisAuthorizableOperation
+                              .UPDATE_TABLE; // Fallback for unknown update types
+                    };
+                actions.add(operation);
```
HonahX: We may consider move this to a helper. Just want to prevent the method length from increasing when iceberg side add more and more update types :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -843,9 +843,74 @@ private UpdateTableRequest applyUpdateFilters(UpdateTableRequest request) {
...
+      // If there are no MetadataUpdates, then default to the UPDATE_TABLE operation.
+      if (actions.isEmpty()) {
+        actions.add(PolarisAuthorizableOperation.UPDATE_TABLE);
+      }
+
+      // Authorize all collected operations
+      for (PolarisAuthorizableOperation action : actions) {
+        authorizeBasicTableLikeOperationOrThrow(
+            action, PolarisEntitySubType.ICEBERG_TABLE, tableIdentifier);
+      }
```
HonahX: I think we want the `EnumSet``authorizeBasicTableLikeOperationOrThrow` here instead of doing a for-loop?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -259,12 +267,15 @@ protected void authorizeBasicTableLikeOperationOrThrow(
...
-        null /* secondary */);
+
+    for (PolarisAuthorizableOperation op : ops) {
+      authorizer.authorizeOrThrow(
+          polarisPrincipal,
+          resolutionManifest.getAllActivatedCatalogRoleAndPrincipalRoles(),
+          op,
+          target,
+          null /* secondary */);
+    }
```
HonahX, line 286: WDYT about having an `authorizeOrThrow` that could take multiple authorizableOperations? That way we could reduce the call to authorizer and potentially save some time from not repeating common logic inside that method.

HonahX (commented):
Thanks for working on this! @travis-bowen 

I am +1 on the idea of more fine grained commit actions. Left some initial comments. Let me know what you think!

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
sfc-gh-tbowen: Not opposed to it, if others are open to another interface method on the authorizer!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sfc-gh-tbowen: Done!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sfc-gh-tbowen: Done!

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
HonahX: Maybe this can be a follow-up discussion since it's about adding an interface method and does not need to block this PR.

But I think ultimately we need the new api to authorize a batch of operations. There is a recent effort to introduce `OpaAuthorizer`(https://github.com/apache/polaris/pull/2680) that need to make rest call for each `authorizeOrThrow` to contact remote privilege/policy agent. It'd better to avoid multiple REST calls given that one iceberg commit usually contains many update types. (especially when CTAS)

collado-mike: I think it would be good to decouple these discussions. I think the prospect of finer grained privileges is easy to accept, but interface changes have broader implications

regtests/t_spark_sql/src/spark_sql_fine_grained_authz.sh
```diff
@@ -0,0 +1,223 @@
...
+SPARK_BEARER_TOKEN="${REGTEST_ROOT_BEARER_TOKEN}"
+CATALOG_NAME="spark_sql_fine_grained_catalog"
+PRINCIPAL_NAME="fine_grained_user"
+PRINCIPAL_ROLE_NAME="fine_grained_principal_role"
+CATALOG_ROLE_NAME="fine_grained_catalog_role"
+
+echo "Creating catalog with fine-grained authorization enabled..."
+
+# Create catalog with fine-grained authorization feature flag enabled
+curl -i -X POST -H "Authorization: Bearer ${SPARK_BEARER_TOKEN}" -H 'Accept: application/json' -H 'Content-Type: application/json' \
```
collado-mike: I hate to ask, but can we make this _not_ a shell script? I hate the shell script reg tests - the python tests are so much easier to debug

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
```diff
@@ -155,6 +155,91 @@ public enum PolarisPrivilege {
       PolarisEntityType.POLICY,
       PolarisEntitySubType.NULL_SUBTYPE,
       PolarisEntityType.CATALOG_ROLE),
+  TABLE_ASSIGN_UUID(
```
collado-mike, line 158: Can we add another privilege, like `TABLE_MANAGE_STRUCTURE` or something similar, that is a super privilege for basically everything except `TABLE_ADD_SNAPSHOT` and `TABLE_SET_SNAPSHOT_REF`? A lot of users will want to grant table-management privileges all at once, rather than having to grant `TABLE_SET_PROPERTIES` and `TABLE_REMOVE_PROPERTIES` separately. Similarly, I think `TABLE_ADD_SNAPSHOT` and `TABLE_SET_SNAPSHOT_REF` could fall under a common parent

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1113,6 +1116,72 @@ public void renameView(RenameTableRequest request) {
     }
   }
 
+  private EnumSet<PolarisAuthorizableOperation> getUpdateTableAuthorizableOperations(
+      UpdateTableRequest request) {
+    boolean useFineGrainedOperations =
+        realmConfig.getConfig(FeatureConfiguration.ENABLE_FINE_GRAINED_UPDATE_TABLE_PRIVILEGES);
```
dimas-b: Since we're already in the catalog context it would be nice to resolve the config against the catalog entity (second `getCatalog()` parameter). This way users would be able to set it differently in different catalogs.

dimas-b: Cf. https://github.com/dimas-b/polaris/blob/68ccb0ee1957faee5d08fd2632d39571ed3bd621/runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java#L109-L112

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
travis-bowen: I'm not sure how much value that might add for users.

My main reasons are - it's hard for me to imagine that someone would grant everything except adding snapshots to a role. At that point, I'd think someone would just grant TABLE_WRITE_PROPERTIES which already encompasses everything. 

The two snapshot only as a group seem a little more useful to me though it's also only two at this point.

I'd also hesitate if there's a differing opinions on how to group the privileges which does get a little trickier around which snapshot ones belong together.

What do you think?

travis-bowen: It’s not that I’m not open to the idea of more super privilege groupings in general but wonder if they might require some more thought and a future PR

collado-mike: To me, it's not hard to imagine a table maintenance service that has the ability to expire snapshots or change data retention policies without the ability to read or write table data

spec/polaris-management-service.yml
```diff
@@ -1552,6 +1552,24 @@ components:
...
+        - TABLE_SET_SNAPSHOT_REF
+        - TABLE_REMOVE_SNAPSHOTS
+        - TABLE_REMOVE_SNAPSHOT_REF
+        - TABLE_SET_LOCATION
+        - TABLE_SET_PROPERTIES
+        - TABLE_REMOVE_PROPERTIES
+        - TABLE_SET_STATISTICS
+        - TABLE_REMOVE_STATISTICS
+        - TABLE_REMOVE_PARTITION_SPECS
+        - TABLE_MANAGE_STRUCTURE
```
HonahX, line 1608: Could you please add these to `NamespacePrivilege` and `Tableprivilege` above too? We probably need to refactor this a little bit but currently we need to add these there to make sure those can be granted to namespace and table

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -397,4 +397,13 @@ public static void enforceFeatureEnabledOrThrow(
...
               .defaultValue(true)
               .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Boolean> ENABLE_FINE_GRAINED_UPDATE_TABLE_PRIVILEGES =
+      PolarisConfiguration.<Boolean>builder()
+          .key("ENABLE_FINE_GRAINED_UPDATE_TABLE_PRIVILEGES")
+          .catalogConfig("polaris.config.enable-fine-grained-update-table-privileges")
+          .description(
+              "When true, enables finer grained update table privileges which are passed to the authorizer for update table operations")
+          .defaultValue(false)
```
HonahX: Shall we turn on this by default? It should not break any existing privilege model

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerFineGrainedAuthzTest.java
```diff
@@ -0,0 +1,276 @@
...
+ * <ul>
+ *   <li>All tests from the base class (validating they still work with the fine-grained authz
+ *       enabled)
+ *   <li>Additional fine-grained authz specific tests
+ *   <li>Disables the "feature disabled" test since it's not applicable here
+ * </ul>
+ */
+@QuarkusTest
+@TestProfile(IcebergCatalogHandlerFineGrainedAuthzTest.Profile.class)
+public class IcebergCatalogHandlerFineGrainedAuthzTest extends IcebergCatalogHandlerAuthzTest {
```
HonahX: I think this implies most of the tests in `IcebergCatalogHandlerAuthzTest` will be ran twice.

Could we extract common helpers to a util class and let both test inherit the base directly?

HonahX:
Overall LGTM! Thanks for adding this and updating! I have some small comments about the spec and test, please let me know WDYT?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
travis-bowen: I'm okay with that.

regtests/t_pyspark/src/test_spark_sql_fine_grained_authz.py
```diff
@@ -0,0 +1,700 @@
...
+        root_client.assign_catalog_role_to_principal_role(
+            principal_role_name=test_principal_role.name,
+            catalog_name=catalog_name,
+            grant_catalog_role_request=GrantCatalogRoleRequest(catalog_role=test_catalog_role)
+        )
+        
+        # Grant only basic privileges to test role
+        test_basic_privileges = [
+            CatalogPrivilege.TABLE_READ_PROPERTIES,
+            CatalogPrivilege.TABLE_READ_DATA,  # Needed to load table for operations
```
collado-mike: Isn't this only true if we specify `X-Iceberg-Access-Delegation=vended-credentials`? Otherwise, `TABLE_READ_PROPERTIES` is sufficient

regtests/t_pyspark/src/test_spark_sql_fine_grained_authz.py
```diff
@@ -0,0 +1,700 @@
...
+        # Test fine-grained operations using the test principal
+        with IcebergSparkSession(
+            credentials=f'{test_principal.principal.client_id}:{test_principal.credentials.client_secret.get_secret_value()}',
+            catalog_name=catalog_name,
+            polaris_url=polaris_catalog_url
+        ) as spark:
+            spark.sql(f'USE {catalog_name}')
+            
+            # SET operation should work
+            spark.sql("ALTER TABLE db1.test_table SET TBLPROPERTIES ('prop1' = 'value1')")
```
collado-mike: Isn't this the same test as on line 297?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -238,29 +241,52 @@ protected void authorizeCreateTableLikeUnderNamespaceOperationOrThrow(
...
+      // The underlying Catalog is also allowed to fetch "fresh" versions of the target entity.
+      resolutionManifest.addPassthroughPath(
+          new ResolverPath(
+              PolarisCatalogHelpers.tableIdentifierToList(identifier),
+              PolarisEntityType.TABLE_LIKE,
+              true /* optional */),
+          identifier);
+      resolutionManifest.resolveAll();
+    }
+  }
```
collado-mike, line 260: We should verify this with @dennishuo . This is really only so we can control the feature flag at the catalog-level, right? If we control it at a system level, we don't need this? Personally, I think it's fine to control at the realm level and if it means we don't have to muck with the entity resolution for authz, I think that would be better.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1113,6 +1120,74 @@ public void renameView(RenameTableRequest request) {
...
+            FeatureConfiguration.ENABLE_FINE_GRAINED_UPDATE_TABLE_PRIVILEGES,
+            getResolvedCatalogEntity());
+
+    if (useFineGrainedOperations) {
+      EnumSet<PolarisAuthorizableOperation> actions =
+          EnumSet.noneOf(PolarisAuthorizableOperation.class);
+      request
+          .updates()
+          .forEach(
+              update -> {
```
collado-mike: Could do this as 
```java
request
    .updates()
    .map(update -> ...)
    .collect(() -> EnumSet.noneOf(PolarisAuthorizableOperation.class), EnumSet::add, EnumSet::addAll);
```

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerFineGrainedAuthzTest.java
collado-mike: I'd just add the new test cases to the existing class. I don't see any point in keeping them separate. Eventually, I think we'll remove the feature flag and leave this on by default anyway

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
travis-bowen: So yes - it is only to be able to control at the catalog level. It was a suggestion from Dmitri here - https://github.com/apache/polaris/pull/2697#discussion_r2389842728

I think it's okay - if not better performant to do it this way. Before we would have created and resolved the manifest for each operation, which is already required. Now, we just resolve it once here and then get the benefit of being able to have this param be controllable at the catalog level (although it's hard to know why someone would want this controllable at the catalog level).

But happy to review this with @dennishuo as well as I already was going to get him to review.

regtests/t_pyspark/src/test_spark_sql_fine_grained_authz.py
travis-bowen: Yes - idk how I managed to do that.

regtests/t_pyspark/src/test_spark_sql_fine_grained_authz.py
travis-bowen: Removed it from the test! The catch was that when I turned off vended credentials I needed to explicitly remove the access delegation header config since the default was set to true in the regtest setup script.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
HonahX: How about we just creating another helper in `icebergCatalogHandler`, like 
```java
authorizeUpdateTableOperationOrThrow(
UpdateTableRequest request,
PolarisEntitySubType subType,
TableIdentifier identifier,
) {
// add path, resolve like normal

// get catalog entity and verify the config

getUpdateTableAuthorizableOperations(request);

for (PolarisAuthorizableOperation op : ops) {
      authorizer.authorizeOrThrow(
          polarisPrincipal,
          resolutionManifest.getAllActivatedCatalogRoleAndPrincipalRoles(),
          op,
          target,
          null /* secondary */);
    }
}
```
So we do not have to resolve twice. It will indeed create some duplicated code, but that could be solved in future refactoring and is IMHO preferred than resolve path twice (assuming in some case EntityCache is not available)?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerAuthzTest.java
```diff
@@ -1069,6 +1087,29 @@ public void testUpdateTableForStagedCreateInsufficientPermissions() {
         () -> newWrapper().updateTableForStagedCreate(TABLE_NS1A_2, new UpdateTableRequest()));
   }
 
+  @Test
+  public void testUpdateTableFallbackToCoarseGrainedWhenFeatureDisabled() {
+    // Test that when fine-grained authorization is disabled, it falls back to
+    // TABLE_WRITE_PROPERTIES
+    // This test validates that the feature flag works correctly by testing the negative case
```
dennishuo, line 1097: This seems to contradict the class-level javadoc comment that says that the feature is default true: `ENABLE_FINE_GRAINED_UPDATE_TABLE_PRIVILEGES (currently true).`

In this case since it's explicitly testing the "disabled" case maybe we should set the property directly in this test method?

dennishuo (approved):
Overall LGTM, just a comment on a test case where the comment didn't seem to match the intent, but I could've been reading it wrong if it's indeed working as intended.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
travis-bowen: So, went ahead and kept things as they are currently.
There shouldn't be any risk of resolving the path twice because I updated even the CatalogHandler to use ensureResolutionManifestForTable which only resolves if it's null. 

Let me know if I might have missed something though.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerAuthzTest.java
travis-bowen: Good catch!

dennishuo (approved):

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
dennishuo: Forgot to respond here. Yeah the control flow extracting the ResolutionManifest initialization looks good to me, and I agree it looks like it won't do any extra persistence resolution calls.

I guess conceptually the generation of the ResolutionManifest doesn't need to be strictly inlined in the authorize helper, even though it's intentionally coupled a bit because the intent is that at least the filled-in ResolutionManifest shouldn't "escape" to be used for any real actions before authz has basically "approved" the manifest.

At some point we might want to attach some counters to our ResolutionManifests and/or persistence impls so that in tests we can easily add regression tests to ensure we don't accidentally add extraneous persistence calls.

HonahX:
Seems all comments have been addressed and there's also no objection in the dev mailing list. Merging

Thanks @dimas-b @collado-mike @dennishuo for reviewing!

singhpk234:
Thank you for the change @travis-bowen ! would be really nice to add this in the CHANGELOG.md file too, to help users try this out :) 
Good news this will be part of 1.2 ! was just about to do rc cut today

sungwy:
I was just thinking about this feature today - thank you so much for working on this PR @travis-bowen ! This is such an awesome enhancement to the list of operation types that can be used for authorization decisions on Polaris.

PR #2698, opened 2025-09-27
Client: add credential reset option

MonkeyCanCode:
Add credential reset option for https://github.com/apache/polaris/pull/2197

### DIFF
--- a/client/python/cli/command/__init__.py
+++ b/client/python/cli/command/__init__.py
@@ -99,6 +99,8 @@ class Command(ABC):
                 remove_properties=[]
                 if remove_properties is None
                 else remove_properties,
+                new_client_id=options_get(Arguments.NEW_CLIENT_ID),
+                new_client_secret=options_get(Arguments.NEW_CLIENT_SECRET),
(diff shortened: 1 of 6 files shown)


jbonofre (approved):
Lgtm thanks !

client/python/integration_tests/test_catalog_apis.py
```diff
@@ -14,6 +15,7 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
+#
```
HonahX, line 18: Just curious, was this related to some auto-formatting rule?

HonahX (approved):
LGTM!

client/python/integration_tests/test_catalog_apis.py
MonkeyCanCode: Was trying to match the license header to the other files we have (where we have single '#' before/after the actual header.

singhpk234 (approved):

Issue #1867, opened 2025-06-11
Add Policy Support to Polaris CLI


HonahX:
### Is your feature request related to a problem? Please describe.

It will be good if we can use Polaris CLI for policy CRUD/attachment

MonkeyCanCode:
Hi @HonahX, are you currently working on this? If not, I’d love to give it a try. Thanks!

MonkeyCanCode:
> Hi [@HonahX](https://github.com/HonahX), are you currently working on this? If not, I’d love to give it a try. Thanks!

Hi @HonahX, I’m not sure if you’re still actively working on this. Since there hasn’t been any update for over two weeks, I’ll proceed with taking this issue. I have most of the changes ready and will open a PR shortly.

MonkeyCanCode:
Here is the PR: https://github.com/apache/polaris/pull/2701

PR #2701, opened 2025-09-28
Client: add support for policy management

MonkeyCanCode:
Implementation for policy management via Polaris CLI (https://github.com/apache/polaris/issues/1867). 

Here are the subcommands to API mapping:
```
attach
 - PUT /polaris/v1/{prefix}/namespaces/{namespace}/policies/{policy-name}/mappings
create
 - POST /polaris/v1/{prefix}/namespaces/{namespace}/policies/{policy-name}/mappings
delete
 - DELETE /polaris/v1/{prefix}/namespaces/{namespace}/policies/{policy-name}
detach
 - POST /polaris/v1/{prefix}/namespaces/{namespace}/policies/{policy-name}/mappings
get
 - GET /polaris/v1/{prefix}/namespaces/{namespace}/policies/{policy-name}
list
 - GET /polaris/v1/{prefix}/namespaces/{namespace}/policies 
   - This is default for `list` operation
 - GET /polaris/v1/{prefix}/applicable-policies
   - This is when we have `--applicable` option provided
update
 - PUT /polaris/v1/{prefix}/namespaces/{namespace}/policies/{policy-name}
```


client/python/cli/command/policies.py
```diff
@@ -0,0 +1,207 @@
...
+                update_policy_request=UpdatePolicyRequest(
+                    description=self.policy_description,
+                    content=json.dumps(policy_document),
+                    current_policy_version=current_policy_version
+                )
+            )
+        elif self.policies_subcommand == Subcommands.ATTACH:
+            target_parts = self.attach_target.split(":")
+            if len(target_parts) != 2:
+                raise Exception("Invalid attach target format. Expected 'type:path'")
```
HonahX: This may not work if the namespace or table name has `:` in it. This is an edge case and similar to the delimiter problem we've discussed before for the single string nested namespace in our endpoint path.

Shall we make this a separate argument to avoid introducing another delimiter?

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -1214,6 +1217,190 @@ options:
...
+The `attach` subcommand is used to create a mapping between a policy and a resource entity.
+
+```
+input: polaris policies attach --help
+options:
+  attach
+    Named arguments:
+      --catalog  The name of an existing catalog
+      --namespace  A period-delimited namespace
+      --attach-target  The target to attach the policy to, e.g. 'namespace:ns1' or 'table:ns1.tb1'
```
HonahX: Shall we use `table-like` instead of `table` since that is the option in the spec?

MonkeyCanCode: Oh yeah, that is missed on my end as `table` is not an valid payload but `table-like` is. I was using `table` but later removed it. But forgot to fix it here. Let me fix it along with other comments.

client/python/cli/command/policies.py
MonkeyCanCode: Didn't know table name can contain `:` in iceberg. But if yes, then yes, let me update it shortly with another arguent to avoid those edge cases.

MonkeyCanCode:
> @MonkeyCanCode Thank you so much for working on this! It looks great!

Thanks for the quick review. New code push for splitting `ATTACH_TARGET` into `ATTACHMENT_TYPE` and `ATTACHMENT_PATH`.

HonahX (dismissed):
LGTM!

MonkeyCanCode:
Thanks for review. I will merge this one later today.

> LGTM!

MonkeyCanCode:
> Thanks for review. I will merge this one later today.
> 
> > LGTM!

@HonahX do u mind approve again? I had to fix a merge conflict.

HonahX (approved):

HonahX:
Thanks for working on this @MonkeyCanCode !

Issue #2491, opened 2025-09-03
Support MySQL as Metastore


Y-Wakuta:
### Is your feature request related to a problem? Please describe.

Currently, Polaris only supports PostgreSQL and H2 databases for its relational JDBC persistence layer. Many organizations have existing MySQL infrastructure and expertise, making MySQL support a valuable addition for wider adoption.

### Describe the solution you'd like

Add MySQL support to the relational-jdbc persistence module by:

1. Add MySQL to DatabaseType enum: Extend DatabaseType.java to include MySQL as a supported database type.
2. Create MySQL-specific schema files: Add mysql/ directory under persistence/relational-jdbc/src/main/resources/ with MySQL-compatible schema-v2.sql (v1 is not needed for new database support).
3. Handle MySQL-specific data types in schema:
  - Use JSON instead of PostgreSQL's JSONB
  - Use TEXT instead of large VARCHAR for column size limitations
4. Update build dependencies: Add MySQL JDBC driver to persistence/relational-jdbc/build.gradle.kts: implementation(libs.mysql.connector.java)
5. Modify QueryGenerator for schema handling: Update getFullyQualifiedTableName() method to handle MySQL's database naming (without schema prefix):
6. Extend existing database type handling in Model classes: The JSON handling logic already exists in ModelEntity.toMap() and ModelPolicyMappingRecord.toMap() - just need to add MySQL case to use plain string instead of PGobject.

### Additional context

MySQL was previously supported through the EclipseLink persistence implementation (see issue https://github.com/apache/polaris/issues/241 and PR https://github.com/apache/polaris/pull/247), but this support is now deprecated since EclipseLink was deprecated in favor of the relational-jdbc implementation. The relational-jdbc module, which is now the recommended production persistence backend, does not currently include MySQL support.

dimas-b:
EclipseLink + MySQL always required a custom build to pull in the JDBC driver.

Current JDBC support provides the driver(s) with Apache Polaris binaries. With that in mind MySQL driver license may not be compatible with Apache. @jbonofre : WDYT?

That said, it should be possible to support MySQL via the MariaDB driver.

jbonofre:
Due to license constraint (MySQL JDBC driver is using GPL, https://downloads.mysql.com/docs/licenses/connector-j-9.4-gpl-en.pdf) it's not possible to ship and distribute MySQL JDBC driver as part of Polaris distribution.
[...]

PR #2704, opened 2025-09-29
Add support for MySQL Metastore

Jayden-Chiu:
As discussed in #2491 , there is currently no support for MySQL as a metastore. many organizations (mine included) rely on existing MySQL infrastructure and this PR adds MySQL support to the `relational-jdbc` module with tests. This PR mostly extends existing code for Postgres support, adding MySQL to `DatabaseType` enum, and changing existing methods to handle MySQL `DatabaseType` and adding MySQL JDBC driver to `persistence/relational-jdbc/build.gradle.kts` and `runtime/server/build.gradle.kts`

### DIFF
--- a/persistence/relational-jdbc/src/main/resources/mysql/schema-v1.sql
+++ b/persistence/relational-jdbc/src/main/resources/mysql/schema-v1.sql
@@ -1,0 +1,104 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file--
+--  distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"). You may not use this file except in compliance
+-- with the License.  You may obtain a copy of the License at
+--
+--  http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- Changes from v2:
+--  * Added `events` table
+
+CREATE DATABASE IF NOT EXISTS POLARIS_SCHEMA;
+USE POLARIS_SCHEMA;
+
+CREATE TABLE IF NOT EXISTS version (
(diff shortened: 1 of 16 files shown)


runtime/server/build.gradle.kts
```diff
@@ -44,6 +44,8 @@ dependencies {
   runtimeOnly("org.postgresql:postgresql")
   runtimeOnly(project(":polaris-relational-jdbc"))
   runtimeOnly("io.quarkus:quarkus-jdbc-postgresql")
+  implementation(libs.mysql.connector.j)
```
adutra, line 47: We cannot ship this dependency it seems.

See this comment:

https://github.com/apache/polaris/issues/2491#issuecomment-3253157608

\cc @jbonofre

dimas-b: Yes, the connector license is not compatible with ASF, AFAIK.

Jayden-Chiu: Sorry, missed this conversation, maybe I will create documentation later on which references this PR on how to get MySQL support if needed.

dimas-b: Everything _except_ the driver is acceptable, I think.

So, @Jayden-Chiu : you may want to keep the java code changes, but remove the driver dependency. That would be mergeable from my POV (pending reviews from other maintainers, of course). Then, instructions on bundling MySQL Connector in a local build could be much simpler.

If the instructions for downstream builds with MySQL were added to a README or docs, that would make a well-rounded PR, from my POV.

persistence/relational-jdbc/src/main/resources/mysql/schema-v1.sql
dimas-b, line 1: Why add earlier schemas? MySQL was not supported when v1 was current 🤔

eric-maynard: Since this is currently only [checked](https://github.com/apache/polaris/blob/10f4bc50a7f6e24343072ee4033d295e718b45fe/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java#L764) within a persistence impl, these version numbers are also per persistence impl. There's no requirement that MySQL vN align with postgres vN.

eric-maynard: So rather than just starting with mysql v3, you could just start with mysql v1 with all the tables that postgres v3 has.

PR #1515, opened 2025-05-02
[JDBC] : Deprecate EclipseLink

singhpk234:
Deperecate Eclipse-link and make relational-jdbc as default

dev-list : https://lists.apache.org/thread/b2qx02hlvxl6rncb7l0xnnx0w22t3z9j

PR #2706, opened 2025-09-29
Updating metastore documentation with Aurora postgres example

fabio-rizzo-01:


site/content/in-dev/unreleased/metastores.md
```diff
@@ -29,14 +29,45 @@ deprecated EclipseLink persistence backends.
...
 QUARKUS_DATASOURCE_JDBC_URL=<jdbc-url-of-postgres>
 ```
+using properties file:
+
+```
+polaris.persistence.type=relational-jdbc
+quarkus.datasource.jdbc.username=<your-username>
+quarkus.datasource.jdbc.password=<your-password>
+quarkus.datasource.jdbc.jdbc-url=<jdbc-url-of-postgres>
+```
```
singhpk234, line 51: please check this feedback from @dimas-b 
https://github.com/apache/polaris/pull/1515/files#r2074243975

IMHO we should be consistent, as far as application.propeties is concerned we reference the quarkus conf page above so that audience can choose what ever way they prefer.

dimas-b: I'd prefer to avoid options like env. vs. file in main docs. All of that is essentially a config entry. There are multiple sources for the config (env., system properties, properties file) and there are differences with how you can specify config names.

Also, lower case names with dots do work in the environment (e.g. in docker/k8s) but do _not_ work in `bash` (and probably other shells).

I think we may need to overhaul the docs WRT user-facing config, but let's keep the existing patterns for this PR (upper case env. variables).

dimas-b: Actually, the RDS config may not be easy to convert to upper case. So let's use lower case (with dots and dashes) in all cases. 

Users have references to Quarkus docs and https://polaris.apache.org/in-dev/unreleased/configuration/

I hope that should be enough input for them to figure out specifics in their deployment env.

dimas-b (commented):
Thanks, @fabio-rizzo-01 ! The doc changes LGTM in general 👍

site/content/in-dev/unreleased/metastores.md
fabio-rizzo-01: so no changes needed from my side right?

dimas-b: I guess that's right 😅 I think the PR is good to merge :)

dimas-b (approved):

dimas-b:
@fabio-rizzo-01 : please rebase to fix CI

fabio-rizzo-01:
> @fabio-rizzo-01 : please rebase to fix CI

done

PR #2711, opened 2025-09-29
Enforce that S3 credentials are vended when requested

dimas-b:
This is a follow-up change to #2672 striving to improve user-facing error reporting for S3 storage systems without STS.

* Add property to `AccessConfig` to indicate whether the backing storage integration can produce credentials.

* Add a check to `IcebergCatalogHandler` (leading to 400) that storage credentials are vended when requested and the backend is capable of vending credentials in principle.

* Update `PolarisStorageIntegrationProviderImpl` to indicate that FILE storage does not support credential vending (requesitng redential vending with FILE storage does not produce any credentials and does not flag an error, which matches current Polaris behaviour).

* Only those S3 systems where STS is not available (or disabled / not permitted) are affected.

* Other storage integrations are not affected by this PR.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
@@ -40,7 +40,7 @@ public abstract class PolarisStorageIntegration<T extends PolarisStorageConfigur
     this.integrationIdentifierOrId = identifierOrId;
   }
 
-  public T config() {
+  protected T config() {
     return config;
   }
 


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -804,7 +804,14 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
           credentialDelegation.getAccessConfig(
               tableIdentifier, tableMetadata, actions, refreshCredentialsEndpoint);
       Map<String, String> credentialConfig = accessConfig.credentials();
-      if (!credentialConfig.isEmpty() && delegationModes.contains(VENDED_CREDENTIALS)) {
+      if (delegationModes.contains(VENDED_CREDENTIALS)) {
```
flyrain, line 807: Thanks @dimas-b for the change. But I think this is not the right place to check. `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION` is mainly for POC or test env, where we allow Polaris to skip credential vending no matter what clients ask. 

I believe the right place to check is in `AwsCredentialsStorageIntegration::getSubscopedCreds()`. Clients ask for credential vending, but this storage doesn't have the capability to do so.

dimas-b: `AwsCredentialsStorageIntegration.getSubscopedCreds()` may be called when STS is disabled (so no credentials will be vended). It's a valid use case for non-AWS S3 storage. Clients still need non-credential config from the "storage integration" code (e.g. endpoints).

flyrain: We should put the check within ` if (shouldUseSts(storageConfig)) { ... }`, sorry, the method `shouldUseSts()` itself.

dimas-b: @flyrain : could you propose pseudo-code (or java example) for this check?

client/python/integration_tests/conftest.py
```diff
@@ -276,7 +282,6 @@ def test_pyiceberg_catalog(
             "warehouse": test_catalog.name,
             "credential": f"{test_principal.principal.client_id}:{test_principal.credentials.client_secret.get_secret_value()}",
             "scope": "PRINCIPAL_ROLE:ALL",
-            "header.X-Iceberg-Access-Delegation": "vended-credentials",
```
dimas-b, line 279: Cf. https://github.com/apache/iceberg-python/blob/pyiceberg-0.10.0/pyiceberg/catalog/rest/__init__.py#L487

CHANGELOG.md
```diff
@@ -34,6 +34,11 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Amazon RDS plugin enabled, this allows polaris to connect to AWS Aurora PostgreSQL using IAM authentication.
 - The EclipseLink Persistence implementation has been deprecated since 1.0.0 and will be completely removed
   in 1.3.0 or in 2.0.0 (whichever happens earlier).
+- Client requests for vended credentials will fail by default if no credentials are available (they used to return
+  no credentials). This can happen with FILE and S3 storage that does not have STS. Clients should normally be updated
```
singhpk234:
```suggestion
  no credentials). This can happen with FILE and S3 compatible storage that does not have STS. Clients should normally be updated
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogFileIntegrationTest.java
```diff
@@ -51,4 +53,18 @@ private static String stripTrailingSlash(String uri) {
         ? uri.substring(0, uri.length() - 1)
         : uri;
   }
+
+  @Test
+  @Override
+  @Disabled // Polaris does not provide credentials for FILE storage
+  public void testLoadCredentials() {
+    super.testLoadCredentials();
+  }
```
singhpk234: should we run these test with the config to allow empty creds on ?

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
```diff
@@ -301,6 +302,62 @@ public void testInternalEndpoints() throws IOException {
     }
   }
 
+  @Test
+  public void testCreateTableFailureWithCredentialVendingWithoutSts() throws IOException {
```
singhpk234, line 306: [optional] is there a way to run these test without relying on the defaults ? just incase we plan defaults this would break.

singhpk234 (commented):
Thank you @dimas-b !

CHANGELOG.md
dimas-b: AWS S3 may be used without STS (admin's choice).

How about `This can happen with FILE and S3 storage if STS is not available or disabled (not permitted).`?

singhpk234: SG ! for my understanding, when is the case when its actual S3 but we don't credential vending ? is it for case where the client already has credentials to work on storage, but then would they be sending the header ? or we just want admins to have this knob

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
dimas-b: It's not about config defaults in this case, it's about MinIO being the only backend available in OSS now that can operate without STS so we can only test vended credential enforcement on MinIO.

The `ALLOW_EMPTY_VENDED_CREDENTIALS` setting is really for users to restore old Polaris behaviour if the enforcement is problematic for them.

If we change the `ALLOW_EMPTY_VENDED_CREDENTIALS` default, this test will have to override that config locally.

dimas-b: I'll set this in the test explicitly to make it independent of defaults.

dimas-b:
It looks like the ripple effects of validating that vended credentials are available at the REST API levels are too wide.

I'll reset this PR and take a different approach (as @flyrain suggested).

CHANGELOG.md
dimas-b: Admins sometimes put restrictions on various policies / permissions even in AWS where STS endpoints are available.

Clients should not request vended credentials when they are known to be unavailable. The extra check is basically to make this kind of error more visible, I guess.

dimas-b:
Reworked this PR completely. Please review again.

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
dimas-b: `ALLOW_EMPTY_VENDED_CREDENTIALS` is gone.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogFileIntegrationTest.java
dimas-b: This is no longer disabled.

CHANGELOG.md
dimas-b: A changelog entry is not needed for the reworked changeset... WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: Updated. PTAL.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -69,6 +73,7 @@ static StorageCredentialCacheKey of(
         allowedListAction,
         allowedReadLocations,
         allowedWriteLocations,
-        refreshCredentialsEndpoint);
+        refreshCredentialsEndpoint,
+        credentialsRequired);
```
HonahX: [non-blocker] I think this implies some extra cache miss for azure and gcp where `credentialsRequired` does not make any difference. 

Do we have some general use-case of the `credentialsRequired` on gcp and azure, such that we could unify the credential existence check at `storageCredentialCache`, such that the persistence/credentialVendor just do its best effort to generate an access config, and let the cache api to decide whether the result is valid?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -70,19 +71,25 @@ public AwsCredentialsStorageIntegration(
...
+      boolean credentialsRequired) {
     int storageCredentialDurationSeconds =
         realmConfig.getConfig(STORAGE_CREDENTIAL_DURATION_SECONDS);
     AwsStorageConfigurationInfo storageConfig = config();
     String region = storageConfig.getRegion();
     AccessConfig.Builder accessConfig = AccessConfig.builder();
 
-    if (shouldUseSts(storageConfig)) {
+    boolean shouldUseSts = shouldUseSts(storageConfig);
+    Preconditions.checkArgument(
```
HonahX: I think this error will be wrapped in to an error SubScopedResult and finally be thrown as `UnprocessableEntityException` in the `StorageCredentialCache` line 155. Just check if this is intended

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
dimas-b: Good point about cache :thinking: Thx!

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: On the java client side it shows up as `org.apache.iceberg.exceptions.RESTException: Unable to process: Failed to get subscoped credentials: Credential vending was requested, but STS is not available`. Is that reasonable?

dimas-b: In the REST response it is indeed a 422 with type = `UnprocessableEntityException`

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
dimas-b: > Do we have some general use-case of the credentialsRequired on gcp and azure

My first attempt in this PR was do enforce credential presence close to the REST level, but it proved to have too wide a ripple effect because many use cases (including PyIceberg) request credential vending by default and that breaks on FILE storage.

I'll see if I can refactor the code to maintain cache validity, but still delegate config validation to storage integrations.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
HonahX: Sorry to clarify, I think this is ok since 422 implies the same error will occur again if client retries, which is the current case. But in the PR description

>  Add checks to AwsCredentialsStorageIntegration (leading to 400) that S3 storage credentials are vended when requested.

says this check will lead to 400. So just point out in case it is intended to be 400 : )

dimas-b: Good catch - thx! I'm reworking the impl. again... it will be 400 / `IllegalArgumentException` at the REST level.

dimas-b: Re-implemented. PTAL.

dimas-b:
I revamped this PR again :sweat_smile: Since there were too many "undo" changes, I squashed and force-pushed. Sorry about losing change history, but I think this revision is has the smallest "diff" so far :wink: Please review again.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
```diff
@@ -108,6 +108,14 @@ public abstract AccessConfig getSubscopedCreds(
           @Nonnull Set<PolarisStorageActions> actions,
           @Nonnull Set<String> locations);
 
+  /**
+   * @param credentialsRequired if {@code true} indicates that the caller requires the returned
+   *     {@link AccessConfig} to have storage credentials; if {@code false} the returned {@link
+   *     AccessConfig} may or may not contain credentials.
+   */
+  public void validateCredentials(
+      @Nonnull AccessConfig accessConfig, boolean credentialsRequired) {}
```
HonahX: Seems this is not used?

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
```diff
@@ -2093,6 +2093,7 @@ private PolarisEntityResolver resolveSecurableToRoleGrant(
     PolarisStorageIntegration<PolarisStorageConfigurationInfo> storageIntegration =
         ms.loadPolarisStorageIntegrationInCurrentTxn(callCtx, reloadedEntity.getEntity());
 
+    storageIntegration.config();
```
HonahX: I think we could remove this?

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
```diff
@@ -40,7 +40,7 @@ public PolarisStorageIntegration(T config, String identifierOrId) {
     this.integrationIdentifierOrId = identifierOrId;
   }
 
-  protected T config() {
```
HonahX, line 43: Seems to be unrelated change?

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
dimas-b: Thanks for catching this and sorry about leftovers from previous code revisions :sweat: ... removed.

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TransactionalMetaStoreManagerImpl.java
dimas-b: removed

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
dimas-b: removed

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -804,13 +804,22 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
...
+                  .prefix(tableMetadata.location())
+                  .config(credentialConfig)
+                  .build());
+        } else {
+          Boolean skipCredIndirection =
+              realmConfig.getConfig(FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
+          Preconditions.checkArgument(
+              !accessConfig.supportsCredentialVending() || skipCredIndirection,
+              "Credential vending was requested for table %s, but no credentials are available",
+              tableIdentifier);
```
HonahX, line 821: WDYT of moving this check to `getAccessConfig` /`FileIOUtil.refreshAccessConfig` in the later refactoring PR: https://github.com/apache/polaris/pull/2736, and we probably do not need the additional `supportsCredentialVending` field in AccessConfig, just adding an additional arg to let caller decide whether to require the credential to be vended.

But I think that should be in a follow-up and after the 1.2.0 cut to avoid additional noise on the release. If I understand correctly, we do not offer backward compatibility of these inner interface. So we could do this if needed in the next release

dimas-b: Sounds reasonable to me.

TBH, I do not like `AccessConfig.supportsCredentialVending()`, but it seemed to be the simplest way to make the check without breaking other things :sweat_smile:

PR #2713, opened 2025-09-30
Fix `delegationModes` parameter propagation in `createTableStaged()`

dimas-b:
This is follow-up bugfix for #2589

The bugfix part of #2711 is extracted here since #2711 proved to be non-trivial and may require extra time.

* Use the `delegationModes` method parameter as intended (as opposed to a local constant).


singhpk234:
Adding my late +1, thank you @dimas-b !

Issue #2675, opened 2025-09-24
Run JDBCPersistence test against all schema version (not just the latest)


singhpk234:

evindj:
@singhpk234  should this issue be closed since it is already addressed in PR #2714 ?

enesyesil:
Hey, is this still open?

dimas-b:
#2714 made some progress on this issue. From my POV it would be nice to enhance tests further and make them find and process all schema versions automatically.

Cf. versioned tests in https://github.com/singhpk234/polaris/tree/main/persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc

Issue #2671, opened 2025-09-24
Provide way of DB schema migration between releases


bacek:
### Describe the bug

Upgrading Polaris from 1.0.1 to 1.1.0 is virtually impossible without manually check difference between DB schemas.

Admin tool `boostrap` command is failing with:
```
java.lang.RuntimeException: Error executing sql script: Failed due to 'ERROR: column "location_without_scheme" does not exist                
  Position: 119' (error code 0, sql-state '42703'), after 1 attempts and 5000 milliseconds                                                   
    at org.apache.polaris.persistence.relational.jdbc.JdbcMetaStoreManagerFactory.bootstrapRealms(JdbcMetaStoreManagerFactory.java:158)      
    at org.apache.polaris.persistence.relational.jdbc.JdbcMetaStoreManagerFactory_ClientProxy.bootstrapRealms(Unknown Source)                
    at org.apache.polaris.admintool.BootstrapCommand.call(BootstrapCommand.java:149)                                                         
    at org.apache.polaris.admintool.BootstrapCommand.call(BootstrapCommand.java:32)                                                          
    at picocli.CommandLine.executeUserObject(CommandLine.java:2031)                                                                          
    at picocli.CommandLine.access$1500(CommandLine.java:148)                                                                                 
    at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2469)                                    
    at picocli.CommandLine$RunLast.handle(CommandLine.java:2461)                                                                             
    at picocli.CommandLine$RunLast.handle(CommandLine.java:2423)                                                                             
    at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)                                                         
    at picocli.CommandLine$RunLast.execute(CommandLine.java:2425)                                                                            
    at io.quarkus.picocli.runtime.PicocliRunner$EventExecutionStrategy.execute(PicocliRunner.java:26)                                        
    at picocli.CommandLine.execute(CommandLine.java:2174)                                                                                    
    at io.quarkus.picocli.runtime.PicocliRunner.run(PicocliRunner.java:40)                                                                   
    at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:141)                                              
    at io.quarkus.runtime.Quarkus.run(Quarkus.java:80)                                                                                       
    at io.quarkus.runtime.Quarkus.run(Quarkus.java:51)                                                                                       
[...]

PR #2714, opened 2025-09-30
JDBC: Handle schema evolution

singhpk234:
### About the change 

1. This PR introduces a new static helper which based on the schema can give back the columns for the Entity Table 
2. Handle the cases where `location_without_scheme` doesn't exist, v1 tables 
3. Enables running the metastore manager suite with all version of schema to detect v1, v2, v3 schema changes faster 
4. Adds a prod readiness check to not enable OPTIMIZE sibling check becuase it may bet that we migrated from v1 to v2 schema and the column `location_without_scheme` has missing data 

solves 
- https://github.com/apache/polaris/issues/2671
- https://github.com/apache/polaris/issues/2675


flyrain:
Thanks @singhpk234 for working on it. This is critical for a smooth Polaris upgrade! A must-have for 1.2. I will take a look soon.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/Converter.java
```diff
@@ -35,10 +35,10 @@ public interface Converter<T> {
   T fromResultSet(ResultSet rs) throws SQLException;
 
   /**
-   * Convert a model into a Map with keys as snake case names, where as values as values of member
-   * of model obj.
+   * Convert a model into a Map with keys as snake case names, and values as values of member of
+   * model obj.
    */
-  Map<String, Object> toMap(DatabaseType databaseType);
+  Map<String, Object> toMap(DatabaseType databaseType, int schemaVersion);
```
flyrain: We don't have to change this interface if we allow the `ModelEntity` carry a new field for schema version.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelEntity.java
```diff
@@ -176,6 +202,14 @@ public static Builder builder() {
 
   @Override
   public PolarisBaseEntity fromResultSet(ResultSet r) throws SQLException {
+    // location_without_scheme column was added in schema version 2
+    String locationWithoutScheme = null;
+    try {
+      locationWithoutScheme = r.getString("location_without_scheme");
+    } catch (SQLException e) {
+      // Column does not exist, handle gracefully
+    }
```
flyrain: I think this part of logic would be simpler if we allow the object to have a new filed for schema version.
```
if (schemaVersion >= 2) {
     locationWithoutScheme = r.getString("location_without_scheme");
}
```

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/AtomicMetastoreManagerWithJdbcBasePersistenceImplV1SchemaTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.
+ */
+
+package org.apache.polaris.persistence.relational.jdbc;
+
+public class AtomicMetastoreManagerWithJdbcBasePersistenceImplV1SchemaTest
+    extends AtomicMetastoreManagerWithJdbcBasePersistenceImplTest {
+  @Override
+  public int schemaVersion() {
+    return 1;
```
flyrain, line 26: I might be a paranoid. Should we add a test for schema version 0?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -302,4 +302,36 @@ public ProductionReadinessCheck checkInsecureStorageSettings(
         ? ProductionReadinessCheck.OK
         : ProductionReadinessCheck.of(errors.toArray(new Error[0]));
   }
+
+  @Produces
+  public ProductionReadinessCheck checkOverlappingSiblingCheckSettings(
+      FeaturesConfiguration featureConfiguration) {
+    var optimizedSiblingCheck = FeatureConfiguration.OPTIMIZED_SIBLING_CHECK;
+    var errors = new ArrayList<Error>();
+    if (Boolean.parseBoolean(featureConfiguration.defaults().get(optimizedSiblingCheck.key()))) {
```
flyrain: It's fine to enable this when schema version is larger or equals to 2. In that case, we might need an input for schema version, which I'm not sure if it's possible.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/Converter.java
singhpk234: i agree, though this would mean carrying a derived column in the model considering this i added this to the signature of this interface, please let me know if you feel strongly about it.

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/AtomicMetastoreManagerWithJdbcBasePersistenceImplV1SchemaTest.java
singhpk234: I totally understand, but for this suite, the test init (bootstrap) itself will fail for 0 as it will try to find `h2.v0.sql` file which doesn't exist.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
singhpk234: > It's fine to enable this when schema version is larger or equals to 2.

my understanding is even when schema version is 2 we don't know if the table was migrated from v1 to v2 ? 

> which I'm not sure if it's possible

would not have the handle of schema version here

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/AtomicMetastoreManagerWithJdbcBasePersistenceImplV1SchemaTest.java
singhpk234: discussed offline to understand more about v0, added the test

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -302,4 +302,36 @@ public ProductionReadinessCheck checkInsecureStorageSettings(
...
+      errors.add(
+          Error.ofSevere(
+              "This setting is not recommended for production environments as it may lead to incorrect behavior, due to missing data for location_without_scheme column in case of migrating from older Polaris versions."
```
dimas-b: Can we detect upgrades and only flag this error in upgraded environments?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: Any module can produce a `ProductionReadinessCheck`, I believe. If we did it in the JDBC persistence, would it be easier to access schema details?

I'd suggest checking for the some values in the new columns and assuming the values are correct if present. It should catch most common upgrade scenarios and also allow backfilling the values manually. WDYT?

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
dimas-b, line 1: Why add a v0 schema? Who would want to run this DDL script?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelEntity.java
```diff
@@ -195,7 +236,7 @@ public PolarisBaseEntity fromResultSet(ResultSet r) throws SQLException {
             // JSONB: use getString(), not getObject().
             .internalProperties(r.getString("internal_properties"))
             .grantRecordsVersion(r.getObject("grant_records_version", Integer.class))
-            .locationWithoutScheme(r.getString("location_without_scheme"))
+            .locationWithoutScheme(this.schemaVersion >= 2 ? locationWithoutScheme : null)
```
dimas-b: Why are we not getting `locationWithoutScheme` from `r`?

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/AtomicMetastoreManagerWithJdbcBasePersistenceImplV1SchemaTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.persistence.relational.jdbc;
+
+public class AtomicMetastoreManagerWithJdbcBasePersistenceImplV1SchemaTest
+    extends AtomicMetastoreManagerWithJdbcBasePersistenceImplTest {
+  @Override
+  public int schemaVersion() {
```
dimas-b, line 25: (optional) this may be more convenient to implement as `@Nested` tests under the JUnit5 framework (less top-level classes).

dimas-b (commented):
Thanks for taking care of schema issues, @singhpk234 ! I had only a brief look. Will review more tomorrow. Overall LGTM 👍

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelEntity.java
singhpk234: This was typo, sorry for that

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
singhpk234: This is to simulate case where a realm is already bootstraped and the metastore manager is recreated, since in 1.0 there was no schema table, schema we will be infering  be 0, now with schema-id just making sure everything works, to simulate a user who is trying to migrate from 1.0 (already bootstrapped) to 1.2

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
singhpk234: I was thinking about it, but didn't find a way to detect the upgrade here as it may happen on bootstrapped with v1 schema and then upgraded to v2 schema by adding column, in this case the schema will contain v2 only but its a migrated case, hence disabling it.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
singhpk234: > If we did it in the JDBC persistence, would it be easier to access schema details

ACK 

> I'd suggest checking for the some values in the new columns and assuming the values are correct if present

This would, still not guarantee, that backfill was done correctly for example a user went to v2 from v1 added the column to the entities table, now its create a table for that table we populated the `location_without_schema` but now for the subsequent table thats created under the namespace, we still doesn't have enough info if all the tables in the namespace have populated the value ? hence cutting this config entirely, if its helpful i can add a config to bypass this readiness check ?

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
```diff
@@ -0,0 +1,109 @@
...
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+
+-- TEST ONLY, TO SIMULATE already having v1 in place
```
flyrain: Minor: Would it be more descriptive if we say this?
```
TO SIMULATE the schema in Polaris 1.0.x, which doesn't have the version table, Polaris interpret its schema version as 0, even the schema file named "schema-v1.sql".
```

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
flyrain: I'v added a bit more description here(https://github.com/apache/polaris/pull/2714/files#r2395306809) to clarify. Hopefully it helps.

flyrain: @dimas-b, trying to provide more context, one of main point of this effort is to be compatible with Polaris 1.0 at this moment. I was thinking that adding a test to simulate the schema in 1.0 benefits a lot. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
flyrain: > Any module can produce a ProductionReadinessCheck

That's cool! Yes, we still have to check the schema version, purely checking the column won't work. And we could make it simpler by just relying on the schema versions. In that case, schema upgrade should include backfilling data if needed. If users just add the column without update the schema version, sorry, we don't recommend this feature for them.

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
dimas-b: If it's test-only, can we put it under the `test` source tree?

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
dimas-b: Did you mean simulate Polaris 0.9? 1.0 has schema v1: https://github.com/apache/polaris/blob/apache-polaris-1.0.0-incubating/persistence/relational-jdbc/src/main/resources/postgres/schema-v1.sql

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
flyrain: Before we having a migration tool, I think it's OK to block the feature universally, as it is risky to rely on it.

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
flyrain: 1.0's schema v1 doesn't have a schema version table. Here is the detailed discussion, https://github.com/apache/polaris/pull/2653#discussion_r2370731169.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: Re: guarantees: Polaris cannot guarantee that user actions are correct in any case. If a user touches the database, the user has to assume responsibility for data correctness.

Polaris should strive to help users to detect and avoid common pitfalls (e.g. updating the schema without back-filling data). That we can do via the "production readiness check", I think.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: I think "severe" is too strong for this check in its current form. It will cause a startup failure and will require a global suppression flag, which will suppress all other possible severe issues.

I propose to make a JDBC-specific check in `RelationalJdbcProductionReadinessChecks`.

* On startup (when the check is called)
* For each realm where `OPTIMIZED_SIBLING_CHECK` is enabled:
* Run a `SELECT` for all namespaces, table-like and view-like entities with `location_without_scheme` being `NULL` (limit 1)
* If found, produce a severe error.

Since this SELECT is not covered by an index, it may be expensive. Add a feature flag to disable it (for users who know what's involved). WDYT?

dimas-b: I'd be ok with deferring the "readiness" check to another PR, by the way.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
singhpk234: > Polaris should strive to help users to detect and avoid common pitfalls (e.g. updating the schema without back-filling data). That we can do via the "production readiness check", I think.

For this its even tricky, lets say we bootstrap a new realm and we were at 1.0, we will automatically go to schema 2, we don't have realm specific keys here 

https://github.com/apache/polaris/blob/395f7bdc23edd55bc3c33095f258f3026a92478e/persistence/relational-jdbc/src/main/resources/h2/schema-v2.sql#L32C1-L34C27

hence this is a bit tricky IMHO, i think the check you suggested later might be best way out

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: It looks like refining this check in upgrade situations requires some more thinking.

I propose to keep the "severe" error, but add another dedicated option for users to acknowledge the risks involved in `OPTIMIZED_SIBLING_CHECK`... kind of similar to `ALLOW_INSECURE_STORAGE_TYPES`.

I still believe this check belongs with the JDBC module since the risks come from there (EclipseLink is deprecated already).

singhpk234: > check belongs with the JDBC module since the risks come from there 

I tried this but since [FeaturesConfiguration](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/config/FeaturesConfiguration.java) class is in service runtime and the JdbcProdReadiness is in persistence module, which service runtime module takes a dependency on, hence i couldn't add it there, am i missing something ?

dimas-b: I'd prefer to avoid using the term "production". Correctness issues apply to all environments.

How about `This setting should be used with care and only enabled in new realms. Enabling it in previously used realms and may lead to incorrect behavior, due to missing data for location_without_scheme column. Set the ALLOW_OPTIMIZED_SIBLING_CHECK flag to acknowledge this warning and enable Polaris to start.`.

Then, I think we could add another feature flag `ALLOW_OPTIMIZED_SIBLING_CHECK` as a user-level safety. If `ALLOW_OPTIMIZED_SIBLING_CHECK` is `true` we do not flag  this as an error.

WDYT?

dimas-b: good point. Le'ts deal with this later, though.

singhpk234: ACK ! sounds reasonable to me, let me push the changes :)

persistence/relational-jdbc/src/main/resources/h2/schema-v0.sql
```diff
@@ -0,0 +1,109 @@
...
+CREATE SCHEMA IF NOT EXISTS POLARIS_SCHEMA;
```
flyrain: I assume we don't need this file anymore, as we have it in the test resoruces.

dimas-b (approved):
LGTM 👍 Thanks for your work on this, @singhpk234 !

singhpk234:
Thank you so much for your reviews everyone ! really appreciate your inputs :) !!

PR #2718, opened 2025-09-30
fix(enhancement): add .idea, .vscode, .venv to top level .gitignore

binarycat0:
fix(enhancement): add .idea, .vscode, .venv to top level .gitignore


client/python/.gitignore
```diff
@@ -66,4 +66,13 @@ target/
...
+spec/
+
+# IntelliJ
+/.idea
+*.iml
+*.ipr
+*.iws
+
+# vscode
+.vscode/
```
dimas-b: Are top-level .gitignore entries not enough?

binarycat0: Unfortunately it doesn't. In my case, it doesn't ignore inner `.idea` catalogues. 
The reason could be the `/` char in the beginning.

dimas-b: I believe it is possible to define those excludes at the top level... https://git-scm.com/docs/gitignore

It would preferable, IMHO, because the `client/python` dir is not special WRT to IDE project files. If we exclude here, why not in `./regtests`, etc.?

binarycat0: For me it's totally ok to do it on the top level. So, the goal was to ignore these catalogs/files. 
But I noticed an extra gitignore in the client/python catalog and decided to add these exclude there also. 
Let's keep it on the top level only.

dimas-b (approved):

HonahX (approved):
Thanks for adding these!

MonkeyCanCode (approved):

HonahX:
Merging! Thanks @binarycat0 for working on this and thanks @dimas-b and @MonkeyCanCode for reviewing!

Issue #1345, opened 2025-04-09
Persistence blocking call occurs in event-loop thread in @PreMatching filter


sfc-gh-rliu:
### Describe the bug

`@PreMatching` filters run on Vertx event loop threads. In some cases, we make a blocking call to persistence in a filter, hence blocking the event loop thread, which is strongly discouraged.

This occurs in 
```
PolarisPrincipalAuthenticatorFilter 
PolarisPrincipalRolesProviderFilter
```
and potentially other filters.

adutra:
Unfortunately it's not possible to annotate pre-matching filters with `@Blocking`, so we'll have to re-design the filters.

A possible solution would be to transform the incriminated filters into methods annotated with `org.jboss.resteasy.reactive.server.ServerRequestFilter` – as such methods are allowed to return a future, or a `Uni`.

dimas-b:
`ServerRequestFilter` appears to be specific to Quarkus. If we go that way, why not use a proper Quarkus Authentication Mechanism plugin?

adutra:
FYI I am working on a fix for this.

adutra:
FYI, I think we should reopen this issue because there is one more problematic filter: `RealmContextFilter`. Since it invokes `RealmContextResolver.resolveRealmContext()`, the filter execution may block if the call to `resolveRealmContext()` blocks.

@dimas-b @sfc-gh-rliu wdyt?

dimas-b:
`DefaultRealmContextResolver` appears to be non-blocking, so I'm not sure reopening this issue is justified, but it probably makes sense to fix `RealmContextFilter` to assume potentially blocking behaviour of `RealmContextResolver`. We could do that in a plain PR without an issue, though, WDYT?

PR #2720, opened 2025-09-30
Extract interface for RequestIdGenerator

adutra:
Summary of changes:

1. Extracted an interface from `RequestIdGenerator`.
2. The `generateRequestId` method now returns a `Uni<String>` in case custom implementations need to perform I/O or other blocking calls during request ID generation.
3. Also addressed comments in #2602.


adutra:
cc @adnanhemani

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
```diff
@@ -18,33 +18,62 @@
...
+  public Uni<Response> assignRequestId(ContainerRequestContext rc) {
     var requestId = rc.getHeaderString(loggingConfiguration.requestIdHeaderName());
-    if (requestId == null) {
-      requestId = requestIdGenerator.generateRequestId();
+    return (requestId != null
+            ? Uni.createFrom().item(requestId)
+            : requestIdGenerator.generateRequestId(rc))
+        .onItem()
+        .invoke(id -> rc.setProperty(REQUEST_ID_KEY, id))
+        .invoke(id -> ContextLocals.put(REQUEST_ID_KEY, id))
```
adnanhemani: What is this for?

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdGenerator.java
```diff
@@ -19,38 +19,21 @@
...
+/**
+ * A generator for request IDs.
+ *
+ * @see RequestIdFilter
+ */
+public interface RequestIdGenerator {
+
+  /**
+   * Generates a new request ID. IDs must be fast to generate and unique. If the generation involves
+   * I/O (which is not recommended), it should be performed asynchronously.
```
flyrain: Can we remove these comments, if I/O involving is not recommended?
```suggestion
   * Generates a new request ID. IDs must be fast to generate and unique. 
```

flyrain (commented):
Thanks @adutra for the PR — it looks good to me overall!
Would you mind elaborating a bit on the motivation behind it?
In what situation would we need a custom request ID generator?

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
adutra: Same as we do in `RealmContextFilter`: this places the request ID in the current Vertx context for later retrieval by another event loop thread:

https://quarkus.io/guides/duplicated-context#context-local-data

adutra:
> Would you mind elaborating a bit on the motivation behind it? 

I'm trying to avoid what happened with the `RealmContextFilter` a while ago: #1345.

IOW, we want to make sure that whatever technique is used to generate a request ID, we won't see Quarkus complaining about it being blocking.

> In what situation would we need a custom request ID generator?

I could see e.g. users storing their request IDs in the database, or using Zookeeper's distributed atomic counter.

runtime/service/src/main/java/org/apache/polaris/service/tracing/DefaultRequestIdGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+  record RequestId(UUID uuid, long counter) {
+
+    RequestId() {
+      this(UUID.randomUUID(), 1);
+    }
+
+    @Override
+    @Nonnull
+    public String toString() {
+      return String.format("%s_%019d", uuid(), counter());
```
dimas-b, line 51: nit: if lexicographic ordering is desired (based on previous conversations), it might be worth using something like `<MILLIS_SINCE_EPOCH_PADDED>_<COUNTER_PADDED>_<UUID>` WDYT?

dimas-b (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/tracing/DefaultRequestIdGenerator.java
dimas-b: or `<MILLIS_PADDED>_<SMALL_COUNTER>_<UUID>` - `SMALL_COUNTER` meaning something in the range of 10K to avoid calling the system clock too often.

but this might be getting too much into the monotonic clock area :sweat_smile:

adutra: I'm not sure lexicographic sort order is desirable for anything else than making it easier for humans to visualize and compare request IDs – so any of your suggestions is fine 😄 
How about we get this in without changing the generation logic, and then, we look into ways of making the logic better?

dimas-b: Absolutely :+1:

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
flyrain: IIUC, `ContextLocals` seems a global container to hold anything could be used within one request context. Another option is to use a explicit `RequestScoped` bean to store these, like we did in the call context bean. In that case, it's a more typed solution.  How would you compare both approaches?

adnanhemani: In a slightly orthogonal direction: I don't see us using the variables we've already stored in ContextLocals anywhere. So while I see that we are putting the information in `RealmContextFilter`, what is our (proposed?) use of this?

Another thought (I'm not sure if this actually answers the question) - if we were to use the request ID later downstream in the application (for example, Event Listeners), would it be better to use `ContextLocals` or `ContainerRequestContext`?

runtime/service/src/main/java/org/apache/polaris/service/tracing/DefaultRequestIdGenerator.java
adnanhemani: Personally, I think it's fine if the requests are only lexicographically sorted across all requests made per node. IMO request ID sorting doesn't mean much for API clients as they also usually log timestamps.

I'm okay to change it, if you both feel there's something better - but just my two cents 😃

adnanhemani:
> > In what situation would we need a custom request ID generator?

> I could see e.g. users storing their request IDs in the database, or using Zookeeper's distributed atomic counter.

To be clear, I am mostly fine with the changes and asking mainly to learn here - isn't the expected workflow that, in the case users want to store their request IDs in a database prior to the API call, that they would store their request IDs on client-side and pass it into Polaris via the request headers? Can you please explain further in what use case the server-side would actually want to store (and then use) pre-computed request IDs?

I think ZK's distributed atomic counter would be a good use case where making request ID asynchronous would help, although I'm sure most developers nowadays would prefer not to use ZK for a use case like creating request IDs unless they have no other option :)

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
adutra: > How would you compare both approaches?

Indeed, a request-scoped bean is undoubtedly better.

> So while I see that we are putting the information in RealmContextFilter, what is our (proposed?) use of this?

This was done a while ago, as I thought at that time that we would be leveraging context propagation to "transfer" some information from the initial request scope to, say, the async tasks framework.

It seems though that, since that time, we've been taking the opposite direction: that of _avoiding_ context propagation (due to some issues with bean proxies, etc.)

The situation today is that those calls to `ContextLocals` are probably not necessary anymore. They can be removed.

Are we OK if I remove the call to `ContextLocals` in `RequestIdFilter` in this PR, then remove the other call in `RealmContextFilter` in a different PR?

adutra:
FYI. One thing that annoys me a bit but I didn't want to change it without prior discussion:

The classes `RequestIdFilter`, `RequestIdGenerator` and `RequestIdResponseFilter` are all in the `org.apache.polaris.service.tracing` package – but the configuration they use is declared in the `org.apache.polaris.service.logging` package and the configuration option is prefixed with `polaris.log`. 

I find this setup a little confusing, especially for developers.

I can provide a "fix" for that later if you all agree.

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
adutra: FYI: #2747 (removed `ContextLocals` and seized the opportunity for some minor cleanup).

adutra: **UPDATE**: I actually remember now 😄 

`ContextLocals` _is required_ in this specific place:

https://github.com/apache/polaris/blob/20febdaede19fb7c46e120652fdd1a262c2138e4/runtime/service/src/main/java/org/apache/polaris/service/metrics/RealmIdTagContributor.java#L41

I even remember asking the Quarkus devs for this feature:

https://github.com/quarkusio/quarkus/pull/47887

So, TLDR: we need to keep `ContextLocals` in `RealmContextFilter`. As for `RequestIdFilter` we can remove it, at least for now.

flyrain (approved):
Thanks @adutra !

sfc-gh-ahemani (approved):

adnanhemani (approved):

PR #2723, opened 2025-09-30
Fix RDS devservices config + adopt for `:polaris-admin:test`

snazy:
Changes:
* Disables devservices for `:polaris-admin` tests as well, which is necessary to _not_ spin up test containers.
* Use the explicit devservices-config as everywhere else.

The first bullet point can cause excessive memory usage, especially with more test classes, eventually killing the whole GH runner.


fabio-rizzo-01:
> /cc @fabio-rizzo-01

thx I missed it in admin

PR #2724, opened 2025-09-30
TestEventsListener - reduce memory pressure

snazy:
Only the last events are needed, so it's enough to keep only the last events by type. Also adds a `.clear()` function to clear the captured events for a deterministic (empty) state for each test case.

Also uses a thread-safe collection now.


dimas-b (approved):

adutra (approved):

singhpk234 (approved):

adnanhemani (approved):
I think it's possible that we may need more than just the last event in the future - but for now, I think this is fine and we can modify the Map as required later.

dimas-b:
Merging based on responses so far. We can certainly change the impl. as new use cases come up later.

PR #2729, opened 2025-09-30
Do not run `publishToMavenLocal` in `spark-plugin-regtest`

dimas-b:
`spark-plugin-regtest` only needs the local Docker image, not Maven artifacts


dimas-b:
This change broke CI, I guess I missed something :sweat_smile: closing.

gh-yzou:
@dimas-b we have test that tests the maven installation behavior, and it needs publishToMavenLocal to make the maven pacakge available for installation.

PR #1347, opened 2025-04-09
Create a wrapper script to generate python client; regenerate the python client

eric-maynard:
As noted in #755 and elsewhere, the generated types in `client/python` are currently out of date. This introduces a script to regenerate them and a gradle task to run that script. Once this is working reliably in CI, we can consider .gitignore'ing these files and just generating them at build time.

PR #2731, opened 2025-09-30
IRC SPEC v3 support: Change the IRC spec to use 1.10 yaml to support upto date request / response models 

singhpk234:
### About the change 

Seems like we have just been bumping the iceberg spec versions but not syncing the rest-api.yaml (it was last synced in 1.8.1) and it doesn't reflect the current state of v3 : 

1. check row-lineage (the spec has changed a lot now the req and response have different field) 
2. check encryption (this is entirely missing) 

I think we need the up to date spec to make sure the request / response model are upto date incase there is a client which is already v3 compliant. 

Adding this in the pr as the first starting step and will see how to move forward from there. I am new to this segment of code so unsure if Polaris generates its own model or reuse the model from iceberg library.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
@@ -181,7 +181,7 @@ public class IcebergExceptionMapper implements ExceptionMapper<RuntimeException>
       case ForbiddenException e -> Status.FORBIDDEN.getStatusCode();
       case jakarta.ws.rs.ForbiddenException e -> Status.FORBIDDEN.getStatusCode();
       case NotAuthorizedException e -> Status.UNAUTHORIZED.getStatusCode();
-      case NamespaceNotEmptyException e -> Status.BAD_REQUEST.getStatusCode();
+      case NamespaceNotEmptyException e -> Status.CONFLICT.getStatusCode();
       case ValidationException e -> Status.BAD_REQUEST.getStatusCode();
       case ServiceUnavailableException e -> Status.SERVICE_UNAVAILABLE.getStatusCode();
       case RuntimeIOException e -> Status.SERVICE_UNAVAILABLE.getStatusCode();
--- a/spec/iceberg-rest-catalog-open-api.yaml
+++ b/spec/iceberg-rest-catalog-open-api.yaml
@@ -19,7 +19,7 @@
 
 # CODE_COPIED_TO_POLARIS
 
-# Apache Iceberg Version: 1.8.1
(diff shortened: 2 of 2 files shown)


spec/iceberg-rest-catalog-open-api.yaml
```diff
@@ -260,7 +260,6 @@ paths:
       parameters:
         - $ref: '#/components/parameters/page-token'
         - $ref: '#/components/parameters/page-size'
-        - $ref: '#/components/parameters/prefix'
```
HonahX, line 263: I remembered we once manually added this because there is a bug in python api generator:
https://github.com/apache/polaris/pull/1347/files/4018286036aacd699cf7ed081bf98d19237486f5#r2079036380

Seems we do not need these anymore : )

spec/iceberg-rest-catalog-open-api.yaml
```diff
@@ -429,6 +428,15 @@ paths:
               examples:
                 NoSuchNamespaceExample:
                   $ref: '#/components/examples/NoSuchNamespaceError'
+        409:
```
HonahX, line 431: We will need to update our exception mapper to be consistent with this: https://github.com/apache/polaris/blob/ea50fe351f19b177460a713cfb524b96d94aba2b/runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java#L184

HonahX (approved):
LGTM! I do not think we need to generate any additional models for this update - we typically use models in iceberg sdk. There is one update needed but can be in a follow-up

spec/iceberg-rest-catalog-open-api.yaml
singhpk234: Nice catch !

HonahX (approved):

runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java
```diff
@@ -181,7 +181,7 @@ static int mapExceptionToResponseCode(RuntimeException rex) {
       case ForbiddenException e -> Status.FORBIDDEN.getStatusCode();
       case jakarta.ws.rs.ForbiddenException e -> Status.FORBIDDEN.getStatusCode();
       case NotAuthorizedException e -> Status.UNAUTHORIZED.getStatusCode();
-      case NamespaceNotEmptyException e -> Status.BAD_REQUEST.getStatusCode();
+      case NamespaceNotEmptyException e -> Status.CONFLICT.getStatusCode();
```
dimas-b, line 184: How well does it work with older Iceberg java clients? What's their expectation in this failure mode?

PR #2745, opened 2025-10-01
Increase javadoc visibility in `:polaris-async-vertx`

dimas-b:
This is to fix javadoc error: `No public or protected classes found to document`

### DIFF
--- a/persistence/nosql/async/vertx/build.gradle.kts
+++ b/persistence/nosql/async/vertx/build.gradle.kts
@@ -52,4 +52,7 @@ dependencies {
   testImplementation(testFixtures(project(":polaris-async-api")))
 }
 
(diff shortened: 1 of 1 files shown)


snazy (approved):
I don't mind this change per-se.
Generally I think it is  better to keep implementations as "protected" as possible.

`JavaPoolAsyncExec` is `public` only for testing purposes.

Can we update the javadoc to just not emit the warning? Or let javadoc emit docs for package-protected types (`(options as StandardJavadocDocletOptions).showFromPackage()`)

dimas-b:
> Can we update the javadoc to just not emit the warning [...]

Sadly, it's an "error" (shows up pretty prominently in IJ)... I'll see about addressing it without making the class `public`.

image

snazy (approved):

PR #2747, opened 2025-10-02
Cleanup RealmContextFilter

adutra:
Summary of changes:

1. <strike>Remove the call to `ContextLocals` (context: https://github.com/apache/polaris/pull/2720#discussion_r2392709683)</strike>.
2. Don't include the exception's message in the response as it can leak details about Polaris internals.
3. Add a small test for success and failure cases.

UPDATE: we cannot remove `ContextLocals`, it's used by `RealmIdTagContributor`.

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/context/RealmContextFilterTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/context/RealmContextFilterTest.java
@@ -69,6 +69,8 @@ class RealmContextFilterTest {
 
   @Test
   public void testNoRealmHeader() {
+    // The default realm is "realm1" so the second pair of secrets is not valid without
+    // an explicit header
     givenTokenRequest("client2", "secret2")
         .header("irrelevant-header", "fake-realm")
         .when()
@@ -79,6 +81,7 @@ class RealmContextFilterTest {
 
   @Test
   public void testDefaultRealm() {
+    // The default realm is "realm1", now credentials match
     givenTokenRequest("client1", "secret1")
         .header("irrelevant-header", "fake-realm")
         .when()


adutra:
cc @flyrain @adnanhemani @dimas-b

snazy (commented):
LGTM

just a few now failing test assertions

adutra:
> just a few now failing test assertions

Whoops there were tests for that already 😅 I didn't know.

I will merge the two test classes together.

dimas-b (dismissed):

snazy (approved):
LGTM

runtime/service/src/test/java/org/apache/polaris/service/context/RealmContextFilterTest.java
```diff
@@ -0,0 +1,108 @@
...
+        .header("irrelevant-header", "fake-realm")
+        .when()
+        .post()
+        .then()
+        .statusCode(Response.Status.UNAUTHORIZED.getStatusCode());
+  }
+
+  @Test
+  public void testDefaultRealm() {
+    givenTokenRequest("client1", "secret1")
```
adnanhemani, line 85: What's the difference between this test and the one above it? Per what I'm seeing, it's the exact same request for different users who both have correct credentials as per L51.

I don't see the `require-header` variable changing between the two so I'm confused whether this is testing the default header properly or not, as the test names suggest? And if the default header was triggering the `IllegalArgumentException`, wouldn't that be a `400 Bad Request` error code rather than `Unauthorized`?

adutra: The behavior is different because the default realm is `realm1`.

In the original test class there were comments explaining this, which I forgot to port over. I will add them now.

dimas-b (approved):

runtime/service/src/main/java/org/apache/polaris/service/context/RealmContextFilter.java
```diff
@@ -46,19 +50,20 @@ public Uni<Response> resolveRealmContext(ContainerRequestContext rc) {
                     rc.getHeaders()::getFirst))
         .onItem()
         .invoke(realmContext -> rc.setProperty(REALM_CONTEXT_KEY, realmContext))
+        // ContextLocals is used by RealmIdTagContributor to add the realm id to metrics
         .invoke(realmContext -> ContextLocals.put(REALM_CONTEXT_KEY, realmContext))
```
flyrain, line 54: Trying to understand why `RealmIdTagContributor` cannot read from the `ContainerRequestContext` instead of relying on `ContextLocals` to pass around any object needed for a `HttpServerMetricsTagsContributor`. Or would it be reasonable to copy the whole `rc` into a `ContextLocals` by default? So that any operation on the `ContextLocals` can use the `ContainerRequestContext`. Would you mind sharing your thought on it?

adnanhemani (approved):
LGTM - thanks @adutra!

flyrain:
Thanks a lot for working on it @adutra ! Thanks @snazy @dimas-b @adnanhemani for the review!

PR #2748, opened 2025-10-02
Refactor resolutionManifest handling in PolarisAdminService

XN137:
- remove mutable `resolutionManifest` field in favor of letting the
  "authorize" methods return their `PolarisResolutionManifest`
- replace `find` helpers with `get` helpers that have built-in error
  handling


runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -266,14 +277,15 @@ private void authorizeBasicTopLevelEntityOperationOrThrow(
           .atDebug()
           .addKeyValue("principalName", topLevelEntityName)
           .log("Allowing rotate own credentials");
```
dimas-b, line 279: tangential: I wonder if this check can be done by the `authorizer` 🤔

dimas-b (approved):
This refactoring LGTM 👍 Thanks, @XN137 !

PR #2757, opened 2025-10-03
Rename request ID to correlation ID

adutra:
Context: https://github.com/apache/polaris/pull/2720#issuecomment-3360293659

<strike>
I moved everything to the `polaris.tracing` config root because it was the easiest move, but we could move the configuration to some other root – just let me know.
</strike>

---

**EDIT**

After giving this some more thought I decided to give a better name to this functionality: after some googling it seems that "correlation ID" is a fairly common name, hence I renamed `RequestIdFilter` and `RequestIdGenerator` to `CorrelationIdFilter` and `CorrelationIdGenerator`.

The configuration is now under the root `polaris.correlation-id`. The helm chart exposes this configuration under `correlationId`.

---

The old config is still supported.


runtime/service/src/main/resources/META-INF/services/io.smallrye.config.ConfigSourceInterceptor
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+org.apache.polaris.service.config.ConfigRelocationInterceptor
```
singhpk234: nit : newline ?

runtime/defaults/src/main/resources/application.properties
```diff
@@ -146,6 +145,9 @@ polaris.metrics.tags.application=Polaris
 # polaris.metrics.tags.environment=prod
 # polaris.metrics.tags.region=us-west-2
 
+polaris.tracing.request-id-generator.type=default
+polaris.tracing.request-id-generator.header-name=Polaris-Request-Id
```
singhpk234: minor / optional : does the header name has to to do anything with the generator, wondering if we just reuse the name we had earlier, WDYT ?
```suggestion
polaris.tracing.request-id-header-name=Polaris-Request-Id
```

CHANGELOG.md
```diff
@@ -53,6 +53,9 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 ### Deprecations
 
 * The property `polaris.active-roles-provider.type` is deprecated and has no effect anymore.
+* The property `polaris.log.request-id-header-name` has been renamed to
+  `polaris.tracing.request-id-generator.header-name`; the old name is still supported for backwards
+  compatibility, but will generate a warning.
```
singhpk234: should we use word deperecated ?

adutra:
@singhpk234 thanks for the review, but due to the Helm chart test failures, I think we need to find a new root for the request ID configuration. Mixing it with `polaris.tracing` will not work well for the Helm chart.

Let me put this into draft state and think a bit more about this.

Sorry for that.

adutra:
@singhpk234 sorry for the late reply, I was OOO last week. I think I have a better solution now, although it requires renaming a few existing classes. LMK what you think. Thanks!

runtime/defaults/src/main/resources/application.properties
```diff
@@ -106,6 +107,9 @@ quarkus.fault-tolerance.global.timeout.enabled=false
 # quarkus.fault-tolerance.global.timeout.unit=minutes
 # quarkus.fault-tolerance.global.timeout.value=10
 
+polaris.correlation-id.header-name=Polaris-Request-Id
```
adutra, line 110: I didn't change the value of this setting as this would be a breaking change, but `Polaris-Correlation-Id` would have been a better default.

runtime/service/src/main/java/org/apache/polaris/service/correlation/CorrelationIdFilter.java
```diff
@@ -26,39 +26,38 @@
...
 import org.jboss.resteasy.reactive.server.ServerRequestFilter;
 import org.jboss.resteasy.reactive.server.ServerResponseFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class RequestIdFilter {
+public class CorrelationIdFilter {
 
-  public static final String REQUEST_ID_KEY = "requestId";
+  public static final String CORRELATION_ID_KEY = "requestId";
```
adutra, line 36: Same here, `correlationId` would have been more appropriate imho, but it's too late.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryBufferEventListener.java
```diff
@@ -94,7 +94,7 @@ protected ContextSpecificInformation getContextSpecificInformation() {
   @Nullable
   @Override
   protected String getRequestId() {
```
adutra, line 96: I also didn't change this method and didn't change `PolarisEvent.requestId` because it's persisted in the database under the column name `request_id`. I think it's not worth changing this.

helm/polaris/values.yaml
```diff
@@ -518,6 +516,16 @@ realmContext:
   realms:
     - POLARIS
 
+# -- Configuration for correlation IDs.
+correlationId:
```
dimas-b, line 520: Sorry for coming late to this PR, but I wonder if we may want to support standard OTel context propagation techniques instead 🤔 

https://opentelemetry.io/docs/concepts/context-propagation/

dimas-b (approved):
Code changes LGTM 👍

helm/polaris/values.yaml
adutra: We already support that :-) In Quarkus, W3C Trace Context propagation is enabled by default: 

https://quarkus.io/guides/opentelemetry-tracing#propagators

IOW, Polaris correlation ID can be considered a simplified alternative to OTel context propagation; but the former doesn't preclude the latter. You can perfectly combine both techniques together.

Polaris correlation ID has one small advantage though: it is included in Polaris events, while OTel context is not. So the only way to correlate a Polaris event with an OTel trace is to first locate the trace using the correlation ID from the event.

dimas-b: Interesting... I think we should propagate OTel context into events. OTel has a concept of "detached" links... This is certainly beyond this scope of this PR, though :wink:

dimas-b: For ref: I think Span kinds of Producer (Polaris side) and Consumer (event receiver) would fit this kind of context propagation.

https://opentelemetry.io/docs/concepts/signals/traces/#producer

adutra: Yes, but the choice to use a specific `SpanKind` depends on the event receiver and what it does with events.

A span with `SpanKind.PRODUCER` would be the right thing to do if the event is being sent to a message queue.

But if the event is being sent via HTTP or gRPC to a remote system, a span with `SpanKind.CLIENT` would be more appropriate.

runtime/service/src/main/java/org/apache/polaris/service/correlation/CorrelationIdConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * The name of the header that contains the correlation ID.
+   *
+   * <p>If a request does not contain this header, it will be assigned a new correlation ID
+   * generated using the configured {@link #generator()}.
+   *
+   * <p>All responses will include the correlation ID in this header.
+   */
+  String headerName();
```
snazy, line 34: Nit: add a `@WithDefault`?

runtime/service/src/main/java/org/apache/polaris/service/config/ConfigRelocationInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+
+package org.apache.polaris.service.config;
+
+import io.smallrye.config.RelocateConfigSourceInterceptor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConfigRelocationInterceptor extends RelocateConfigSourceInterceptor {
```
snazy, line 26: 👍 for adding this!

snazy (approved):

runtime/service/src/main/java/org/apache/polaris/service/correlation/CorrelationIdConfiguration.java
adutra: It seems we prefer in Polaris to have a default declared in `runtime/defaults/src/main/resources/application.properties`, rather than a default declared in the Java class.

Do you prefer the other way around? I don't have a strong preference here :-)

snazy: I don't mind either.

helm/polaris/values.yaml
dimas-b: Good point :+1: I was actually thinking of embedding an outgoing OTel context into the event itself. This way regardless of the event delivery technology, consumers may want to perform async tasks that actually related to the event data, so in that case they would have the option of linking a `CONSUMER` span to the Polaris Event's `PRODUCER` span.... just an idea.

adutra: Hmm I'm not sure I fully understand your example. I'd imagine that, if some event listener wants to send the event to a remote system and propagate OTel context, it would leverage the OTel SDK and do something like this:

```java
@Inject Tracer tracer;

void processEvent(PolarisEvent event) {
  var span = tracer.spanBuilder("polaris event delivery").setSpanKind(SpanKind.PRODUCER).startSpan();
  try (Scope scope = span.makeCurrent()) {
    sendEventToMessageQueue(event);
  }
}
```

The created span would have the current request span as its parent, then the OTel magic would kick in when the appropriate propagator is invoked. 

We might not even have to do this btw, some message queue clients probably already add a PRODUCER span transparently whenever they produce a record to the message broker.

The `CONSUMER` span kind would have to be created on the receiving side (the message broker) for this to work, but I bet this is the case for most popular message queues as they have the required instrumentation already.

Would that example solve your problem?

dimas-b: The question how context propagation happens... `sendEventToMessageQueue` will probably use some integration code that takes the context and passes it to the message queue as a "header" of sorts.

If out intention is to correlate with Polaris events using only Polaris "specs", we could put the outgoing context into the event as an attribute.

adutra:
Putting this again back to draft 😅 – waiting on the ML discussion outcome:

https://lists.apache.org/thread/bb1qyxjt827t3tomv2xp0s1kovxjsp94

My apologies to those who already reviewed, but I think we are making good progress towards a thorough design.

adutra:
Given the outcome of the ML discussion, I'm going to close this PR and propose instead another PR that follows the action items that we agreed on.

Thanks to all the reviewers and sorry for back and forth on this topic.

PR #2523, opened 2025-09-08
SigV4 Auth Support for Catalog Federation - Part 3: Service Identity Info Injection

XJDKC:
## Milestones

This is Part 3 of the [[Splitting] Initial SigV4 Auth Support for Catalog Federation](https://github.com/apache/polaris/pull/1805). Upcoming parts will build on this system:
* ~~#1899~~
* #2190
* #2523
* SigV4 Auth Support for Catalog Federation - Part 4: Connection Credential Manager

## Introduction

This PR introduces service identity management for SigV4 Auth Support for Catalog Federation. Unlike user-supplied parameters, the service identity represents the identity of the Polaris service itself and should be managed by Polaris.

We introduce a new `ServiceIdentityProvider` interface responsible for three core operations:

### 1. Allocating a service identity to a catalog entity
* Invoked during external catalog creation with SigV4 authentication
* A vendor may use the same service identity across all entities in an account, or assign different identities per catalog
* The associated DPO stores only a reference (identifier) to the service identity, not the credentials themselves

### 2. Retrieving identity information for API responses
* When generating responses for APIs like `getCatalog`, Polaris uses `getServiceIdentityInfo()` to return identity metadata (e.g., AWS IAM ARN) without exposing credentials
* This method returns only display information, never sensitive credentials

### 3. Retrieving credentials for authentication
* When accessing a remote catalog, Polaris uses `getServiceIdentityCredential()` to obtain the full credential with AWS access keys
* After retrieving its IAM credentials, Polaris assumes the customer-provided IAM role to obtain temporary AWS credentials, then uses those to access the remote catalog via SigV4
* Credentials are resolved lazily only when actually needed

This PR also introduces a `DefaultServiceIdentityProvider`, which supports static configuration via Quarkus application properties. While it does not support dynamic runtime resolution (all identities must be specified before server startup), it provides:
* A simple, practical mechanism for testing SigV4 authentication
* A reference implementation that vendors can build upon

In the future, vendors may implement custom providers that integrate with their internal infrastructure for on-demand identity management or dynamic credential rotation.

## Design Overview

Each `ConnectionConfigInfoDpo` (used for remote catalog federation) now contains a `ServiceIdentityInfoDpo`, which holds a `SecretReference` that serves as a unique identifier for the service identity instance. This design allows:
* Polaris to store only a reference (identifier) to its service identity in the catalog entity
* The actual credentials to be looked up from configuration at runtime
* Separation of metadata (for API responses) from credentials (for authentication)
[...]

PR #2759, opened 2025-10-04
[DRAFT] SigV4 Auth Support for Catalog Federation - Part 4: Connection Credential Manager

XJDKC:
## Milestones

This is Part 3 of the [[Splitting] Initial SigV4 Auth Support for Catalog Federation](https://github.com/apache/polaris/pull/1805). Upcoming parts will build on this system:
* ~~#1899~~
* #2190
* #2523

## Introduction
This PR introduces a flexible connection credential management system for Polaris. Building on Part 3's service identity management, this system combines Polaris service identities with user-provided authentication parameters to generate credentials for remote catalog access.
The core of this PR is the new `ConnectionCredentialVendor` interface, which:
* Generates connection credentials by combining service identity with user auth parameters
* Supports different authentication types (SIGV4, OAuth) through CDI
* Provides on-demand credential generation
* Enables easy extension for new authentication types

## Key Components
1. **ConnectionCredentialVendor**
The central interface that defines how connection credentials are generated. It combines Polaris service identities with user-provided authentication parameters to produce the final credentials needed for remote catalog access. This interface serves as the foundation for all authentication type-specific implementations.

2. **DefaultPolarisCredentialManager**
The primary implementation that orchestrates credential generation. It uses CDI to automatically select the appropriate vendor based on authentication type, delegates the credential generation process, and provides consistent error handling. This class integrates with the ServiceIdentityProvider from Part 3 to resolve Polaris service identities.

3. **SigV4ConnectionCredentialVendor**
A reference implementation for AWS SigV4 authentication. It handles the complex process of resolving AWS IAM credentials from service identities, performing AWS STS AssumeRole operations, and generating temporary credentials for catalog access. This implementation demonstrates best practices for vendor development.

4. **@SupportsAuthType**
A CDI qualifier annotation that marks vendor implementations with their supported authentication types. This enables automatic vendor selection at runtime while maintaining type safety. The qualifier is essential for the CDI-based pluggable architecture.

##  Future Improvements
* Add credential caching support
* Migrate storage creds to this new framework


runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,92 @@
...
+  private final Instance<ConnectionCredentialVendor> credentialVendors;
+
+  public DefaultPolarisCredentialManager(
+      @Any Instance<ConnectionCredentialVendor> credentialVendors) {
+    this.credentialVendors = credentialVendors;
+  }
+
+  @Override
+  public @Nonnull ConnectionCredentials getConnectionCredentials(
+      @Nonnull ConnectionConfigInfoDpo connectionConfig) {
```
XJDKC, line 81: In the future, we may embed creds cache in the credential manager, if we can't find a valid entry in the cache, we will select the correct vendor and delegate the credential retrieval to the credential vendor (for both storage and connection).

In this way, we can deprecate the old flow (using metastore to get the subscoped creds).

Since we need to generate cache key and may need realm id and the catalog id, wandering if we should pass in the whole catalog entity here. But we can refactor this later once we add cache support for credential manager.

cc: @dimas-b @dennishuo @HonahX

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.iceberg.aws.AwsProperties;
+
+/**
+ * A subset of Iceberg catalog properties recognized by Polaris.
+ *
+ * <p>Most of these properties are meant to initialize Catalog objects for accessing the remote
+ * Catalog service.
+ */
+public enum ConnectionCredentialProperty {
+  AWS_ACCESS_KEY_ID(String.class, AwsProperties.REST_ACCESS_KEY_ID, "the aws access key id", true),
```
dimas-b, line 31: nit: WDYT about renaming to `CatalogAccessProperty` by analogy to `StorageAccessProperty`?

polaris-core/src/main/java/org/apache/polaris/core/credentials/PolarisCredentialManager.java
```diff
@@ -0,0 +1,52 @@
...
+ *   <li>The vendor uses the service identity credential together with user-provided authentication
+ *       parameters to obtain temporary access credentials (e.g., via AWS STS AssumeRole).
+ * </ol>
+ *
+ * <p>This design supports both SaaS and self-managed deployments, ensuring a clear separation
+ * between user-provided configuration and Polaris-managed identity. In the future, this interface
+ * will be extended to also manage storage credentials, providing a unified interface for all
+ * credential management needs.
+ */
+public interface PolarisCredentialManager extends ConnectionCredentialVendor {}
```
dimas-b, line 52: What's the value of having this empty interface? Why not merge with `ConnectionCredentialVendor`?

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
```diff
@@ -0,0 +1,134 @@
...
+ *   <li>Extracts customer's role ARN from {@link SigV4AuthenticationParametersDpo}
+ *   <li>Calls AWS STS AssumeRole to get temporary credentials
+ *   <li>Returns temporary access key, secret key, and session token
+ * </ol>
+ *
+ * <p>This is the default implementation with {@code @Priority(100)}. Custom implementations can
+ * override this by providing a higher priority value.
+ */
+@ApplicationScoped
+@SupportsAuthType(AuthenticationType.SIGV4)
```
dimas-b: How about just `@AuthType`?... The annotation being a `@Qualifier` makes it pretty clear that it is about "supporting" something, IMHO.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SupportsAuthTypes.java
```diff
@@ -0,0 +1,52 @@
...
+ * @SupportsAuthType(AuthenticationType.BEARER)
+ * public class NoOpConnectionCredentialVendor implements ConnectionCredentialVendor {
+ *   // Handles both IMPLICIT and BEARER authentication (no credential transformation)
+ * }
+ * }</pre>
+ */
+@Qualifier
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SupportsAuthTypes {
```
dimas-b: This annotation appears to be unused? 🤔

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManagerFactory.java
```diff
@@ -0,0 +1,54 @@
...
+  public DefaultPolarisCredentialManagerFactory(
+      @Any Instance<ConnectionCredentialVendor> credentialVendors) {
+    this.credentialVendors = credentialVendors;
+  }
+
+  @Override
+  public PolarisCredentialManager getOrCreatePolarisCredentialManager(RealmContext realmContext) {
+    return cachedCredentialManagers.computeIfAbsent(
+        realmContext.getRealmIdentifier(),
+        key -> new DefaultPolarisCredentialManager(credentialVendors));
```
dimas-b: `DefaultPolarisCredentialManager` does not have any state... What's the value in caching its instances per realm?

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,92 @@
...
+
+    // Select the appropriate vendor based on authentication type
+    AuthenticationType authType =
+        connectionConfig.getAuthenticationParameters().getAuthenticationType();
+    Instance<ConnectionCredentialVendor> selectedVendor =
+        credentialVendors.select(SupportsAuthType.Literal.of(authType));
```
dimas-b: This approach is fine, but it defers errors to runtime. I'd propose the `DefaultPolarisCredentialManager` to resolve all possible `ConnectionCredentialVendor` at build time. Note that defaults implementations are already present in the same build module (jar). With that approach we probably do not need `SupportsAuthType`.

Downstream builds that have non-default `ConnectionCredentialVendor` implementations will have to override the `PolarisCredentialManager` too, but I think that would be a reasonable tradeoff for achieving built-time cross-reference validation (downstream builds will be able to do build time validation too).

runtime/defaults/src/main/resources/application.properties
```diff
@@ -207,6 +210,7 @@ quarkus.arc.ignored-split-packages=\
   org.apache.polaris.service.auth.external.mapping,\
   org.apache.polaris.service.auth.external.tenant,\
   org.apache.polaris.service.auth.internal,\
+  org.apache.polaris.service.credentials,\
```
dimas-b: Can we avoid split packages in _new_ code?

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -242,6 +245,13 @@ public StsClientsPool stsClientsPool(
     return new StsClientsPool(config.effectiveClientsCacheMaxSize(), httpClient, meterRegistry);
   }
 
+  @Produces
+  public PolarisCredentialManagerFactory credentialManagerFactory(
+      PolarisCredentialManagerConfiguration config,
+      @Any Instance<PolarisCredentialManagerFactory> credentialManagerFactories) {
+    return credentialManagerFactories.select(Identifier.Literal.of(config.type())).get();
```
dimas-b: Do we really need a factory here?.. Why not produce `PolarisCredentialManager` directly?

polaris-core/src/main/java/org/apache/polaris/core/credentials/PolarisCredentialManager.java
XJDKC: Because in the future, it will implement the `StoargeCredentialVendor`.

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
XJDKC: Sounds good!

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
XJDKC: Sounds good!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManagerFactory.java
XJDKC: Makes sense, we can remove the factory maybe?

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
XJDKC: Correct, let me remove it!

runtime/defaults/src/main/resources/application.properties
XJDKC: I rebased my code from my old branch, forget to remove it

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SupportsAuthTypes.java
XJDKC: This interface is just a holder for `@SupportAuthType`, if one connection credential vendor supports two auth types, then we will use `@SupportAuthType` twice, then `@SupportsAuthTypes` will be used implicitly.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Agree that this approach would defer surfacing certain errors to runtime.

However, there are several benefits to using this CDI-based delegation approach:
1. **Cleaner dependency management**: `PolarisCredentialManager` would no longer need to hold objects specific to a particular `ConnectionCredentialVendor` that may never be used.
    * For example, in [PolarisStorageIntegrationProviderImpl.java](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java), we currently keep `stsClientProvider`, `stsCredentials`, and `gcpCredsProvider`. But if the storage happens to be GCP, the STS-related fields are unnecessary.
    * Moreover, the class uses a switch-case statement to instantiate the correct storage integration, and use new to construct the object. With the CDI-based approach, the implementation becomes cleaner and provides better isolation, also it's easier to extend.
    * The overall goal is to encapsulate all authentication- or storage-specific logic within their respective `ConnectionCredentialVendor` or `StorageCredentialVendor`. The `PolarisCredentialManager` would simply delegate credential retrieval to the appropriate vendor, acting primarily as a redirection and caching layer.
2. **Simpler extensibility**: if a Polaris vendor only wants to override the logic for a specific auth type, this structure makes that much easier.

As for handling errors, what do you think about adding a validation step in 
[ProductionReadinessChecks.java](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java)

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManagerFactory.java
dimas-b: Yes, +1 to removing the factory

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SupportsAuthTypes.java
dimas-b: How can it work in practice? I mean supporting two different auth types in one class? :thinking:

XJDKC: For example, vendor may have its own interface that unified the behavior of getting connection credentials. 
Then, vendor can just add one `ConnectionCredentialVendor` which supports all the auth methods.
Same as the storage creds vendor.

dimas-b: Does it work with CDI (just wondering)?

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: Fair enough. Supporting extensibility in the default manager sounds good to me. However, in that case I'd propose to separate the `PolarisCredentialManager` interface and `ConnectionCredentialVendor`.

The look up would be `PolarisCredentialManager.vendorFor(Class<X extends ConnectionConfigInfoDpo>)` -> `ConnectionCredentialVendor<X extends ConnectionConfigInfoDpo>` (using CDI for lookup), then `ConnectionCredentialVendor.connectionCredentials(X)`.

Implementations are in fact [bound](https://github.com/apache/polaris/blob/949923417b55a0931a8b4d274b47cad1cebdee97/runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java#L90) the the type of the config object (`X`), so I think it would make sense to make it explicit. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,92 @@
...
+
+    if (selectedVendor.isUnsatisfied()) {
+      LOGGER.warn("No credential vendor found for authentication type: {}", authType);
```
HonahX:
```suggestion
      LOGGER.warn("No connection credential vendor found for authentication type: {}", authType);
```
Also as we probably do not need connection credential vendor for OAUTH and BEARER in most cases, do you think it makes sense to provide noop implementation for them to silence this warning? Otherwise it may confuse user when they see this warning but they have configured everything as expected?

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,92 @@
...
+      return ConnectionCredentials.builder().build();
+    }
+
+    if (selectedVendor.isAmbiguous()) {
+      LOGGER.error(
+          "Multiple credential vendors found for authentication type: {}. "
```
HonahX:
```suggestion
          "Multiple connection credential vendors found for authentication type: {}. "
```
[nit] Shall we state this is connection credential vendor, given that we are planning to bring StorageCredentialVendor in the manager?

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,92 @@
...
+              + "Use @Priority to specify which vendor should be used. "
+              + "Higher priority values take precedence.",
+          authType);
+      throw new IllegalStateException(
+          String.format(
+              "Ambiguous credential vendor for authentication type: %s. "
```
HonahX:
```suggestion
              "Ambiguous connection credential vendor for authentication type: %s. "
```

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
XJDKC: Done!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManagerFactory.java
XJDKC: Done!

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
XJDKC: Done!

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
XJDKC: Done!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Good suggestions! Actually, I’m planning to add one for OAuth and Bearer as well.

The goal is to have PolarisCredentialManager handle all credential retrievals for accessing remote catalogs, including obtaining access tokens via OAuth.

To achieve this, we'll introduce an `OAuthConnectionCredentialVendor`.
* For now, it can simply return the OAuth credentials and let the Iceberg SDK fetch a new token for each API call to the remote catalog.
* In the future, we can enhance the vendor to handle token fetching and caching within `PolarisCredentialManager`, so we don't need to request a new token for every call.

Will open a follow up PR to add these two cuz this PR is already very large and I want to focus on SigV4 in this PR.

For Implicit Auth Type we need a `NoOpConnectionCredentialVendor`

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Make sense!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: I think it makes sense to add `<X extends ConnectionConfigInfoDp>`, but I also want the `PolarisCredentialManager` to serve as a caching layer.

It would still be ideal to have `PolarisCredentialManager` call `ConnectionCredentialVendor.connectionCredentials` internally and manage caching within the manager itself. This way, callers only need to interact with `PolarisCredentialManager`, they won’t need to care whether the credentials come from cache or are freshly fetched, and `PolarisCredentialManager` is the centra place to manage all the credentials.

The final state of PolarisCredentialManager would be:

```java
interface PolarisCredentialManager implements ConnectionCredentialVendor, StorageCredentialVendor {}

class DefaultPolarisCredentialManager implements PolarisCredentialManager {
  private AccessCredsCache accessCredsCache;
  private final Instance<ConnectionCredentialVendor> connectionCredentialVendors;
  private final Instance<StorageCredentialVendor> storageCredentialVendors;

  @Override
  public @Nonnull ConnectionCredentials getConnectionCredentials(
      @Nonnull ConnectionConfigInfoDpo connectionConfig) {
    // 1. generate cache key
    // 2. check if there is a valid connection credential in the cache
    // 3. if yes, get the credential from cache and return it directly
    // 4. If not, select the credential vendor based on Auth Type
    // 5. Delegate to the connection credential vendor to fetch the connection credentials
    // 6. Store the connection credentials in the cache
    // 7. return the connection credentials
  }

  @Override
  public @Nonnull StorageAccessConfig getSubscopedStorageCreds(
     @Nonnull RealmConfig realmConfig,
      boolean allowListOperation,
    // 1. generate cache key
    // 2. check if there is a valid storage credential in the cache
    // 3. if yes, get the credential from cache and return it directly
    // 4. If not, select the credential vendor based on Storage Type
    // 5. Delegate to the storage credential vendor to fetch the storage credentials
    // 6. Store the storage credentials in the cache
    // 7. return the storage credentials
  }
}
```

WDYT? If it makes sense to keep calling `ConnectionCredentialVendor.connectionCredentials(X)` inside `PolarisCredentialManager`, then adding `PolarisCredentialManager.vendorFor(Class<X extends ConnectionConfigInfoDpo>)` might not be necessary, but I can still add a private helper method in `PolarisCredentialManager`.

In that case, should I proceed to just add `<X extends ConnectionConfigInfoDpo>` instead.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SupportsAuthTypes.java
XJDKC: Hey Dmitri, thanks for bringing this up:
Dive deep into it: the native CDI framework doesn't support it, since the repeated `@AuthType` annotation will be wrapped in `@AuthTypes`, so we can use `Instance.select` to find the correct one directly. If we want to support this, we need to write an CDI extension, which is complicated. 

For now, let me just remove `@AuthTypes`, one vendor can only handle one classes.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Done!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Done!

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
```diff
@@ -0,0 +1,134 @@
...
+        (SigV4AuthenticationParametersDpo) connectionConfig.getAuthenticationParameters();
+
+    // Use Polaris's IAM identity to assume the customer's role
+    StsClient stsClient = getStsClient(sigv4Params);
+
+    // Build the AssumeRole request with Polaris's credentials
+    AssumeRoleRequest.Builder requestBuilder =
+        AssumeRoleRequest.builder()
+            .roleArn(sigv4Params.getRoleArn())
+            .roleSessionName(
```
dennishuo, line 114: Though this is probably fine for most use cases since the assumeRole session credential doesn't ever (normally) leave the Catalog server itself (in contrast to StorageIntegration ones where we need to vend it out), we should probably still find a way to provide a hook for adding a sessionPolicy here.

Maybe at least a TODO. A "simple" approach would be to allow configuring a whole service-level scoping policy string that would rely on wildcards to share across catalogs/realms.

More advanced would be to derive the policyString from a combination of the Type in the connectionConfig, and the service-level configuration.

If we ever do refactor to merge this for Catalog and Storage we at least would need to abstract out the sessionPolicy generation

XJDKC: Good point, since we don't vend / return this credentials to the client, so it should be fine.

For the `simple` approach, we need to use different service-level scoping policy for different services (based on the sigv4 signing name). e.g., if the service is 
* `glue` we need to get `glue:ListTables`, ``glue:ListDatabases`, e.t.c, 
    * We may also need some privileges on Lake Formation so that we can get the vended credentials, if we want to use the vended credentials to load table metadata rather than using the storage config.
* `execute-api` (api gateway), we need to use `execute-api:Invoke`
* `s3tables`: we need another set of privileges.

I will add a TODO for now since I need to spend some time to figure out the correct policy string for each service. We may need a refactor on the policy generation.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Hey @dimas-b, 

I considered making `ConnectionCredentialVendor` generic (i.e., `ConnectionCredentialVendor<T extends ConnectionConfigInfoDpo>`), but after thinking through it, I believe the current non-generic design is actually more appropriate. Here's why:
The type parameter doesn't represent what the vendor actually validates:
The vendor doesn't care whether it receives `IcebergRestConnectionConfigInfoDpo`, `HadoopConnectionConfigInfoDpo`, or `HiveConnectionConfigInfoDpo`, it only cares about the contents of the configuration:
* The service identity type (e.g., `AwsIamServiceIdentityInfoDpo`)
* The authentication parameters type (e.g., `SigV4AuthenticationParametersDpo`)

For example, `SigV4ConnectionCredentialVendor` works with any `ConnectionConfigInfoDpo` as long as it contains AWS IAM credentials + SigV4 parameters. The container type is irrelevant. I add some preconditions to validate that.

polaris-core/src/main/java/org/apache/polaris/core/credentials/PolarisCredentialManager.java
dimas-b: The future enhancements is fine, but as far as this PR is concerned (where `StoargeCredentialVendor` does not exist), IMHO, having an empty `PolarisCredentialManager` interface is confusing. All "real" methods are in `ConnectionCredentialVendor`.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: I agree. My generics proposal was not a good idea :sweat_smile:

dimas-b: Let's tackle caching in a separate PR.

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
```diff
@@ -0,0 +1,66 @@
...
+ * <p>Most of these properties are meant to initialize Catalog objects for accessing the remote
+ * Catalog service.
+ */
+public enum CatalogAccessProperty {
+  AWS_ACCESS_KEY_ID(String.class, AwsProperties.REST_ACCESS_KEY_ID, "the aws access key id", true),
+  AWS_SECRET_ACCESS_KEY(
+      String.class, AwsProperties.REST_SECRET_ACCESS_KEY, "the aws access key secret", true),
+  AWS_SESSION_TOKEN(
+      String.class, AwsProperties.REST_SESSION_TOKEN, "the aws scoped access token", true),
+  EXPIRATION_TIME(
```
dimas-b, line 36: Do we actually need this to be an enum value? I do not see any real usage of it 🤔

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: I'll open a new thread with more concrete suggestions just for the current PR.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,109 @@
...
+  public @Nonnull ConnectionCredentials getConnectionCredentials(
+      @Nonnull ConnectionConfigInfoDpo connectionConfig) {
+
+    // Select the appropriate vendor based on authentication type
+    AuthenticationType authType =
+        connectionConfig.getAuthenticationParameters().getAuthenticationType();
+    Instance<ConnectionCredentialVendor> selectedVendor =
+        credentialVendors.select(AuthType.Literal.of(authType));
+
+    if (selectedVendor.isUnsatisfied()) {
```
dimas-b: Calling `credentialVendors.select()` looks reasonable to me, but getting too much into CDI internals looks like an overkill. Is there material benefit in having this custom logic as opposed to letting the CDI framework throw an exception from `selectedVendor.get()`?

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Yeah, I will open a follow up PR to add the cache logic.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -0,0 +1,109 @@
...
+          authType);
+      throw new IllegalStateException(
+          String.format(
+              "Ambiguous connection credential vendor for authentication type: %s. "
+                  + "Multiple implementations found without @Priority annotation.",
+              authType));
+    }
+
+    // Delegate to the vendor to generate credentials
+    return selectedVendor.get().getConnectionCredentials(connectionConfig);
```
dimas-b: Since `DefaultPolarisCredentialManager` multiplexes over type-specific lower level implementations, having them under the same java interface (`ConnectionCredentialVendor`) does not sound right to me.

If a `ConnectionCredentialVendor` is impl. it type-specific, then `DefaultPolarisCredentialManager` should not be a sub-type of `ConnectionCredentialVendor` because it is not type-specific.

All calls are made via the `PolarisCredentialManager` interface.

WDYT about detaching `PolarisCredentialManager` from `ConnectionCredentialVendor` but keeping current `DefaultPolarisCredentialManager` code?

runtime/service/src/main/java/org/apache/polaris/service/credentials/PolarisCredentialManagerConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+@StaticInitSafe
+@ConfigMapping(prefix = "polaris.credential-manager")
+public interface PolarisCredentialManagerConfiguration {
+
+  /**
+   * The type identifier of the PolarisCredentialManager implementation to use. This corresponds to
+   * the {@code @Identifier} annotation value on the implementation (e.g., "default").
+   */
+  String type();
```
dimas-b, line 40: nit: Do we expect more config properties here? If not, it could be folded into the producer, as in: https://github.com/apache/polaris/blob/20febdaede19fb7c46e120652fdd1a262c2138e4/runtime/admin/src/main/java/org/apache/polaris/admintool/config/AdminToolProducers.java#L41

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
XJDKC: I just follow the `StorageAccessProperty`. If we want to remove `CatalogAccessProperty`, we should also deprecate `StorageAccessProperty` later. 

I think initially, the reason we add the `StorageAccessProperty` is because we want all the storage credentials related properties must be explicitly defined in the num class, otherwise, `AccessConfig` is just an open-end property map.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: I think I have explained here:

> However, there are several benefits to using this CDI-based delegation approach:
> * **Cleaner dependency management**: PolarisCredentialManager would no longer need to hold objects specific to a particular ConnectionCredentialVendor that may never be used.
>     * For example, in [PolarisStorageIntegrationProviderImpl.java](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java), we currently keep `stsClientProvider`, `stsCredentials`, and `gcpCredsProvider`. But if the storage happens to be GCP, the STS-related fields are unnecessary.
        * Moreover, the class uses a switch-case statement to instantiate the correct storage integration, and use new to construct the object. With the CDI-based approach, the implementation becomes cleaner and provides better isolation, also it's easier to extend.
        * The overall goal is to encapsulate all authentication- or storage-specific logic within their respective `ConnectionCredentialVendor` or `StorageCredentialVendor`. The `PolarisCredentialManager` would simply delegate credential retrieval to the appropriate vendor, acting primarily as a redirection and caching layer.
> * **Simpler extensibility**: if a Polaris vendor only wants to override the logic for a specific auth type, this structure makes that much easier.

> As for handling errors, what do you think about adding a validation step in
[ProductionReadinessChecks.java](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java)

If you feel strongly about it, I'm happy to go with the old approach similar to `PolarisStorageIntegrationProviderImpl`. I think both approaches have their pros and cons, I chose the current implementation mainly because I felt it was a bit clearer: `PolarisCredentialManager` doesn't need to understand the underlying logic of each `CredentialVendor`. The vendors can still reuse the common logic in `PolarisCredentialManager` (such as caching and multiplexing); they only need to override the parts specific to their own credential handling.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: I agree it might look a bit confusing right now since `PolarisCredentialManager` doesn't define additional methods and everything still lives under `ConnectionCredentialVendor`.

The main reason I introduced the `PolarisCredentialManager` interface early is to establish a clear separation of concerns before we add `StorageCredentialVendor` and potentially other vendor types. The goal is for `PolarisCredentialManager` to act as the central entry point for all credential management (including caching and multiplexing), while `ConnectionCredentialVendor` represents just one of several possible vendor implementations. i.e., The `PolarisCredentialManager` is also a Connection Credential Vendor.

So even though it looks slightly redundant in this PR, it helps set up a cleaner, more extensible structure for the upcoming changes and avoids unnecessary refactoring later. I'm happy to add a short comment or javadoc note to clarify that intent if it helps.

I'd prefer to keep `PolarisCredentialManager` implement `ConnectionCredentialVendor` for now. 
But if you feel strongly about separating it, I can also duplicate the method in `PolarisCredentialManager` and make it a standalone interface.

runtime/service/src/main/java/org/apache/polaris/service/credentials/PolarisCredentialManagerConfiguration.java
XJDKC: Explicitly defining a config property has some benefits: it gives us a dedicated top-level class to hold the property, rather than relying on conventions defined in the producer classes. WDYT? I'm okay to remove it.

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
dimas-b: `StorageAccessProperty.EXPIRATION_TIME` is actually referenced in code, while this one is not :thinking:

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: In this comment thread, I merely wonder if checking `selectedVendor.isUnsatisfied()` is worth the trouble. If it's not satisfied, `.get()` will throw anyway. I'm ok keeping current code, but I suspect it may lead to copy-paste replication in other areas and I'm not sure it is actually beneficial.

What would the user-visible errors be with an without this special `if` statement?

XJDKC: I see, I can give it a try, the reason I add this if statement is that I don't want to surface a confusing error message generated by CDI that it's hard to debug.

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
XJDKC: `StorageAccessProperty.EXPIRATION_TIME` is also not referenced in the callee, we just use it to add `expiration_time` property to the AccessConfig in `AzureCredentialsStorageIntegration` and `AwsCredentialStorageIntegration` (the equivalent of `ConnectionCredentialVendor`)

XJDKC: Another usage of this enum class is:
It's a bridge between polaris and iceberg sdk. Polaris can understand this enum class, but we need to convert it to the property that iceberg sdk can understands. 
Considering this, I prefer to keep it as it's.

i.e., Inside polaris, we will always use this enum class to refer the property, when we need to use iceberg sdk, like `RESTCatalog` and `FileIO`, we will convert the property to string-based property that iceberg sdk can understand.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: Are you suggesting that in the future `PolarisCredentialManager` will extend `ConnectionCredentialVendor` and some other `AbcCredentialVendor` and `DefCredentialVendor`?

dimas-b:
It feels like we're getting into somewhat hypothetical and a bit too abstract discussions here :slightly_smiling_face: I do not want to block progress on the Catalog Federation feature, so I'd be ok with merging this PR "as is" if other reviewers more familiar with the feature and intended final state of the code approve.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Yes, if we want to migrate the storage credential related to this new framework, we need to make `PolarisCredentialManager` extends `StorageCredentialVendor` as well.

```java
interface PolarisCredentialManager implements ConnectionCredentialVendor, StorageCredentialVendor {}

class DefaultPolarisCredentialManager implements PolarisCredentialManager {
  private AccessCredsCache accessCredsCache;
  private final Instance<ConnectionCredentialVendor> connectionCredentialVendors;
  private final Instance<StorageCredentialVendor> storageCredentialVendors;

  @Override
  public @Nonnull ConnectionCredentials getConnectionCredentials(
      @Nonnull ConnectionConfigInfoDpo connectionConfig) {
    // 1. generate cache key
    // 2. check if there is a valid connection credential in the cache
    // 3. if yes, get the credential from cache and return it directly
    // 4. If not, select the credential vendor based on Auth Type
    // 5. Delegate to the connection credential vendor to fetch the connection credentials
    // 6. Store the connection credentials in the cache
    // 7. return the connection credentials
  }

  @Override
  public @Nonnull StorageAccessConfig getSubscopedStorageCreds(
      boolean allowListOperation,
    // 1. generate cache key
    // 2. check if there is a valid storage credential in the cache
    // 3. if yes, get the credential from cache and return it directly
    // 4. If not, select the credential vendor based on Storage Type
    // 5. Delegate to the storage credential vendor to fetch the storage credentials
    // 6. Store the storage credentials in the cache
    // 7. return the storage credentials
  }
}
```

XJDKC:
> It feels like we're getting into somewhat hypothetical and a bit too abstract discussions here 🙂 I do not want to block progress on the Catalog Federation feature, so I'd be ok with merging this PR "as is" if other reviewers more familiar with the feature and intended final state of the code approve.


Thanks Dmitri, the runtime logic can always be evolved and refined later. There are still many things we need to do for catalog federation, so we can continue the discussion about the interface in future PRs!

cc: @dennishuo @HonahX

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: that approach is fine from my POV, except that I do not think `PolarisCredentialManager` is necessary as an interface... Why not implement all of those "leaf" interfaces in the `DefaultPolarisCredentialManager` _class_?

XJDKC: Because `PolarisCredentialManager` acts as a composition of different credential vendors, it's responsible for delegating calls to the appropriate underlying vendor. Each specific vendor, in turn, is just a subclass of one of the credential vendor types. e.g., `SigV4ConnectionCredentialVendor implements ConnectionCredentialVendor`, `S3StorageCredentialVendor implements StorageCredentialVendor`, e.t.c.

This pattern is also used by many classes:
```java
public interface ByteChannel extends ReadableByteChannel, WritableByteChannel {
// no methods
}
```

dimas-b: I get the composition part :sweat_smile: but I do not understand why it has to be expressed as an `interface` (`PolarisCredentialManager`) :sweat_smile: 

A class can implement multiple interfaces (e.g. `DefaultPolarisCredentialManager`), then CDI can inject its instance into locations that need some of the interfaces it implements. Usage sites do not need to be aware that `DefaultPolarisCredentialManager` implements multiple interfaces.

HonahX (dismissed):
Thanks @dimas-b for the thorough review on this! I think we've had a lot constructive and thoughtful discussions. The remaining points, as you mentioned, are more forward-looking points could be addressed when we actually add those features/refactorings.

With this and the fact that catalog federation is well-guarded by a param, I vote +1 on this to support federate to Glue IRC

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Hey @dimas-b, here is the response if I remove the if statements and remove the `@AuthType` annotation for `SigV4ConnectionCredentialVendor`.
```
{
  "error": {
    "message": "No bean found for required type [interface org.apache.polaris.core.credentials.connection.ConnectionCredentialVendor] and qualifiers [[@jakarta.enterprise.inject.Any(), @org.apache.polaris.service.credentials.connection.AuthType(value=SIGV4)]]",
    "type": "UnsatisfiedResolutionException",
    "code": 500
  }
}
```

It's actually quite informative, should I remove the `selectedVendor.isUnsatisfied()` and also `
selectedVendor.isAmbiguous()`? Will this expose more internal details?

dimas-b: Yeah, I'd prefer to remove `isUnsatisfied()` and `isAmbiguous()` and rely on the CDI infra to produce errors - thx!

That said, we should probably not expose internal messages in 500 responses, but log them, but that's another matter :sweat_smile:

dimas-b: If you prefer to catch `UnsatisfiedResolutionException` and re-throw something else, that would be fine from my POV.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: IIUC, you're suggesting something like:
```java
// Make DefaultPolarisCredentialManager implement both vendor interfaces directly
class DefaultPolarisCredentialManager implements ConnectionCredentialVendor, StorageCredentialVendor {}

@Inject ConnectionCredentialVendor credentialVendor; 
// CDI will create a DefaultPolarisCredentialManager for us
```

That makes sense at first glance, but there are two key issues with this approach:
1. Ambiguity in CDI injection
We'll have multiple `ConnectionCredentialVendor` implementations (e.g., SigV4ConnectionCredentialVendor, OAuthConnectionCredentialVendor, etc.), each responsible for a specific auth type.
If we rely on injecting `ConnectionCredentialVendor`, CDI won't know which implementation to provide, `DefaultPolarisCredentialManager` or one of the specific vendors.
Having an explicit `PolarisCredentialManager` abstraction ensures CDI can inject the central manager that multiplexes across vendors and delegates to the right one internally.
2. Centralized abstraction for all credential logic
`PolarisCredentialManager` is designed to be the central interface (and later, the main entry point) for all credential management, including caching, vendor selection, and storage-related logic.
It might look unnecessary right now since most logic still lives in `SigV4ConnectionCredentialVendor`, but defining the interface early helps keep a clean architecture and avoids major refactors once more vendor types and caching logic are introduced.
3. Enable vendor extension
By defining a `PolarisCredentialManager` interface, individual vendors (e.g., custom or third-party ones) can extend this interface if they want. This allows them to hook into their own credential management pipeline more easily.

dimas-b: > Ambiguity in CDI injection

I do not think so :) CDI "sees" through inheritance so `IMPL implements P` + `P extends A,B,C` is the same as `IMPL implements A,B,C` from CDI eligibility POV, AFAIK. (edit: fixed name confusion in the example)

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Got it! Let me revise my PR!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Yes, but if we have `AImpl implements A`, then on the callee side we no longer have an interface `P` to depend on, we only have the concrete class C.
That means the callee has to inject either the concrete class `C` or use the base interface `A` as the variable type.

If we use `C`, vendors can't provide their own custom implementation, since we're tied to that concrete class.
If we use `A`, we run into CDI ambiguity once multiple `A` implementations exist.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Done!
```json
{
  "error": {
    "message": "No credential vendor available for authentication type: SIGV4",
    "type": "IllegalStateException",
    "code": 500
  }
}
```

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: Why would callers depend on the "union" class/interface? If they need the functionality provided by `A`, then depend on `A`, and so on... Callers should preferably depend on the "minimal" interface they need :thinking:

CDI will inject objects as required. Even if the implementation class extends multiple interfaces, each caller gets the same instance, but the call site declaration only links to the specific interface (A, B or C in my example).

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
```diff
@@ -78,32 +78,24 @@ public RealmContext getRealmContext() {
...
-          authType);
+    // Use CDI to select the appropriate vendor based on the authentication type
+    ConnectionCredentialVendor selectedVendor;
+    try {
+      selectedVendor = credentialVendors.select(AuthType.Literal.of(authType)).get();
+    } catch (jakarta.enterprise.inject.UnsatisfiedResolutionException e) {
+      // No vendor registered for this authentication type
+      throw new IllegalStateException(
+          "No credential vendor available for authentication type: " + authType, e);
+    } catch (jakarta.enterprise.inject.AmbiguousResolutionException e) {
```
dimas-b: nit: I'd just catch the base exception. This is an internal error - we do not need to report specific messages to the client. We should log the CDI exception for trouble-shooting on the server side, but the client should receive as generic error like `Unable to obtain credentials required for executing this request`, IMHO.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: LGTM :+1: Thx!

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: You are right, but that's why there are some ambiguities: take a look into the code I posted above, if we remove `PolarisCredentialManager` and define `DefaultPolarisCredentialManager` like that, we have `SigV4ConnectionCredentialVendor` and  `DefaultPolarisCredentialManager` and both of them inherit `ConnectionCredentialVendor`. If the caller uses `ConnectionCredentialVendor` as the variable type, which object the caller will get. `DefaultPolarisCredentialManager` can get the correct one because it will select the candidate based on the authentication type, but if we just define `@Inject ConnectionCredentialVendor credentialVendor`, the CDI framework can't tell which one should be used.

`PolarisCredentialManager` is the multiplexing layer, the idea is that, for the caller, if you need a connection credential, or a storage credential, just use CDI to get a PolarisCredentialManager, pass in the needed info, then PolarisCredentialManager will handle all the things for you, e.g., forwarding the calls to the actual credential vendor, cache the credentials. So, we can't directly inject `ConnectionCredentialVendor`, we should use a separate interface, and `PolarisCredentialManager` is the one we are looking for.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Done, we haven't added credential vendor for other auth types, so I just swallow the `UnsatisfiedResolutionException` for now, will remove `catch (UnsatisfiedResolutionException e)` in a follow up PR.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: > If the caller uses ConnectionCredentialVendor as the variable type, which object the caller will get [...]

This can be resolved in CDI with the help of "qualifiers". Simple callers get the "default" impl., e.g. the cache / multiplexer, while the cache can further qualify the next level of implementation via `@AuthType`.

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
dimas-b: thx :+1:

runtime/service/src/main/java/org/apache/polaris/service/credentials/DefaultPolarisCredentialManager.java
XJDKC: Yes, this can't be resolved in CDI, but why do we make things very complicated.

With this approach, when we want to get the `DefaultPolarisCredentialManager`, we need to filter based on `Qualifier`.
When we want to get the specific vendor, we need to filter based on `@AuthType`.
```java
  @Produces
  @RequestScoped
  public ConnectionCredentialVendor connectionCredentialVendor(
      PolarisCredentialManagerConfiguration config,
      @Any Instance<ConnectionCredentialVendor> credentialManagers) {
    return (ConnectionCredentialVendor) credentialManagers.select(Identifier.Literal.of(config.type())).get();
  }

  @Produces
  @RequestScoped
  public StorageCredentialVendor storageCredentialVendor(
      PolarisCredentialManagerConfiguration config,
      @Any Instance<StorageCredentialVendor> credentialManagers) {
    return (StorageCredentialVendor) credentialManagers.select(Identifier.Literal.of(config.type())).get();
  }
```

Why not just add a new interface that implements all the credential vendors?
That way, whenever we need credentials, we can simply use `PolarisCredentialManager`. It's cleaner and doesn't introduce any downside.

If vendors want to provide their own `DefaultPolarisCredentialManager`, they don't need to understand the underlying mechanism and also follow this pattern. (i.e., providing two producers and return different vendor type).

And the callers also don't need to understand that the one they are using is not the specific `ConnectionCredentialVendor` but the `DefaultPolarisCredentialManager` which has the cache logic and the multiplexing logic.

dimas-b: let's revisit later :) I think current code is functional and clear enough for maintenance.

XJDKC: Sounds good, happy to discuss this further if needed!

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
dennishuo: Would we end up using the same `expiration-time` between different auth types in the future?

I recall `StorageAccessProperty.EXPIRATION_TIME` being more of the deprecated field, where we prefer `AWS_SESSION_TOKEN_EXPIRES_AT_MS` to be more specific about what it is. That one also has the `s3` prefix: `s3.session-token-expires-at-ms`.

Is there ever a situation where we might need to put an S3 credential into the same config map as a sigv4 catalog-connection credential?

In that case we'd need separate expiration times as well if they're not in separate tuples. Following the same convention of the `rest.` prefix used in `AwsProperties`, we'd then want this to be `rest.session-token-expires-at-ms`.

HonahX:
Thanks @XJDKC @dennishuo @dimas-b for reviewing! Given the size of the PR I propose to do non-urgent small changes in follow-up for easier reviewing process.

XJDKC:
> Generally LGTM, just a comment about the naming of the `expiration-time` property. We could fixup in a followup if needed

Good catch! I also looked into the Iceberg SDK, unfortunately, there isn't currently a property that indicates the expiration time of the token/credential for remote catalog access. So for now, I reused the `expiration-time` property, which is already used by `StorageAccessProperty`. 

I can modify this in a follow up PR!

PR #2762, opened 2025-10-06
JDBC: Fix Bootstrap with schema options

singhpk234:
### About the change 

1. Fix bootstrapping with schema option, presently its considered mutually exclusive so if give -v option with -r its fails saying they are exlusive, on not giving -r it say no relams configured 
2. schemaFile checks for empty string as not specifying the option means bootstrap can't use with -v either 
3. Adds a check that if the schema table exists then we can't downgrade as the schema is the global value and specially since realm is the part of schema such thing is not possible 

### TODO

while testing i found each QuarkusMainlauncher is launching new PG container, thinking of fix still to reuse it to test already bootstrapped schema.

Note: I don't think its a 1.2 blocker since we can't use this, but would love to know other community member takes.

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
@@ -23,6 +23,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Locale;
+import java.util.Objects;
 import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
 
 public enum DatabaseType {
@@ -53,9 +54,10 @@ public enum DatabaseType {
    * caller.
    */
   public InputStream openInitScriptResource(@Nonnull SchemaOptions schemaOptions) {
-    if (schemaOptions.schemaFile() != null) {
+    if (schemaOptions.schemaFile() != null
+        && !Objects.requireNonNull(schemaOptions.schemaFile()).isEmpty()) {
       try {
-        return new FileInputStream(schemaOptions.schemaFile());
+        return new FileInputStream(Objects.requireNonNull(schemaOptions.schemaFile()));
       } catch (IOException e) {
         throw new IllegalArgumentException("Unable to load file " + schemaOptions.schemaFile(), e);
       }
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
@@ -18,13 +18,8 @@
  */
 package org.apache.polaris.persistence.relational.jdbc;
 
-import jakarta.annotation.Nonnull;
-import java.io.FileInputStream;
-import java.io.IOException;
 import java.io.InputStream;
 import java.util.Locale;
-import java.util.Objects;
-import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
 
 public enum DatabaseType {
   POSTGRES("postgres"),
@@ -53,27 +48,15 @@ public enum DatabaseType {
    * Open an InputStream that contains data from an init script. This stream should be closed by the
    * caller.
    */
-  public InputStream openInitScriptResource(@Nonnull SchemaOptions schemaOptions) {
-    if (schemaOptions.schemaFile() != null
-        && !Objects.requireNonNull(schemaOptions.schemaFile()).isEmpty()) {
-      try {
-        return new FileInputStream(Objects.requireNonNull(schemaOptions.schemaFile()));
-      } catch (IOException e) {
-        throw new IllegalArgumentException("Unable to load file " + schemaOptions.schemaFile(), e);
-      }
-    } else {
-      final String schemaSuffix;
-      switch (schemaOptions.schemaVersion()) {
-        case null -> schemaSuffix = "schema-v3.sql";
-        case 1 -> schemaSuffix = "schema-v1.sql";
-        case 2 -> schemaSuffix = "schema-v2.sql";
-        case 3 -> schemaSuffix = "schema-v3.sql";
-        default ->
-            throw new IllegalArgumentException(
-                "Unknown schema version " + schemaOptions.schemaVersion());
-      }
-      ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
-      return classLoader.getResourceAsStream(this.getDisplayName() + "/" + schemaSuffix);
+  public InputStream openInitScriptResource(int schemaVersion) {
+    final String schemaSuffix;
+    switch (schemaVersion) {
+      case 1 -> schemaSuffix = "schema-v1.sql";
+      case 2 -> schemaSuffix = "schema-v2.sql";
+      case 3 -> schemaSuffix = "schema-v3.sql";
+      default -> throw new IllegalArgumentException("Unknown schema version " + schemaVersion);
     }
+    ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
+    return classLoader.getResourceAsStream(this.getDisplayName() + "/" + schemaSuffix);
   }
 }
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
@@ -49,14 +49,15 @@ public enum DatabaseType {
    * caller.
    */
   public InputStream openInitScriptResource(int schemaVersion) {
-    final String schemaSuffix;
-    switch (schemaVersion) {
-      case 1 -> schemaSuffix = "schema-v1.sql";
-      case 2 -> schemaSuffix = "schema-v2.sql";
-      case 3 -> schemaSuffix = "schema-v3.sql";
-      default -> throw new IllegalArgumentException("Unknown schema version " + schemaVersion);
+    // Preconditions check is simpler and more direct than a switch default
+    if (schemaVersion <= 0 || schemaVersion > 3) {
+      throw new IllegalArgumentException("Unknown or invalid schema version " + schemaVersion);
     }
+
+    final String resourceName =
+        String.format("%s/schema-v%d.sql", this.getDisplayName(), schemaVersion);
+
     ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
-    return classLoader.getResourceAsStream(this.getDisplayName() + "/" + schemaSuffix);
+    return classLoader.getResourceAsStream(resourceName);
   }
 }
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -400,8 +400,9 @@ public class DatasourceOperations {
   }
 
   public boolean isRelationDoesNotExist(SQLException e) {
-    return RELATION_DOES_NOT_EXIST.equals(e.getSQLState())
-        || H2_RELATION_DOES_NOT_EXIST.equals(e.getSQLState());
+    return (RELATION_DOES_NOT_EXIST.equals(e.getSQLState())
+            && databaseType == DatabaseType.POSTGRES)
+        || (H2_RELATION_DOES_NOT_EXIST.equals(e.getSQLState()) && databaseType == DatabaseType.H2);
   }
 
   private Connection borrowConnection() throws SQLException {
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
@@ -52,8 +52,12 @@ public class DatasourceOperations {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DatasourceOperations.class);
 
+  // PG STATUS CODES
   private static final String CONSTRAINT_VIOLATION_SQL_CODE = "23505";
   private static final String RELATION_DOES_NOT_EXIST = "42P01";
+
+  // H2 STATUS CODES
+  private static final String H2_RELATION_DOES_NOT_EXIST = "90079";
 
   // POSTGRES RETRYABLE EXCEPTIONS
   private static final String SERIALIZATION_FAILURE_SQL_CODE = "40001";
@@ -396,7 +400,8 @@ public class DatasourceOperations {
   }
 
   public boolean isRelationDoesNotExist(SQLException e) {
-    return RELATION_DOES_NOT_EXIST.equals(e.getSQLState());
+    return RELATION_DOES_NOT_EXIST.equals(e.getSQLState())
+        || H2_RELATION_DOES_NOT_EXIST.equals(e.getSQLState());
   }
 
   private Connection borrowConnection() throws SQLException {
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -748,10 +748,10 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
       }
       return schemaVersion.getFirst().getValue();
     } catch (SQLException e) {
-      LOGGER.error("Failed to load schema version due to {}", e.getMessage(), e);
       if (fallbackOnDoesNotExist && datasourceOperations.isRelationDoesNotExist(e)) {
         return SchemaVersion.MINIMUM.getValue();
       }
+      LOGGER.error("Failed to load schema version due to {}", e.getMessage(), e);
       throw new IllegalStateException("Failed to retrieve schema version", e);
     }
   }
@@ -761,10 +761,7 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
     try {
       List<PolarisBaseEntity> entities =
           datasourceOperations.executeSelect(query, new ModelEntity());
-      if (entities == null || entities.isEmpty()) {
-        throw new IllegalStateException("Failed to check if Entities table exist");
-      }
-      return true;
+      return entities != null && !entities.isEmpty();
     } catch (SQLException e) {
       if (datasourceOperations.isRelationDoesNotExist(e)) {
         return false;
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -756,6 +756,23 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
     }
   }
 
+  static boolean entityTableExists(DatasourceOperations datasourceOperations) {
+    PreparedQuery query = QueryGenerator.generateEntityTableExistQuery();
+    try {
+      List<PolarisBaseEntity> entities =
+          datasourceOperations.executeSelect(query, new ModelEntity());
+      if (entities == null || entities.isEmpty()) {
+        throw new IllegalStateException("Failed to check if Entities table exist");
+      }
+      return true;
+    } catch (SQLException e) {
+      if (datasourceOperations.isRelationDoesNotExist(e)) {
+        return false;
+      }
+      throw new IllegalStateException("Failed to check if Entities table exists", e);
+    }
+  }
+
   /** {@inheritDoc} */
   @Override
   public <T extends PolarisEntity & LocationBasedEntity>
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
@@ -20,15 +20,10 @@
 package org.apache.polaris.persistence.relational.jdbc;
 
 import java.util.Optional;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import org.apache.polaris.core.persistence.bootstrap.BootstrapOptions;
 import org.apache.polaris.core.persistence.bootstrap.SchemaOptions;
 
 public class JdbcBootstrapUtils {
-
-  // Define a pattern to find 'v' followed by one or more digits (\d+)
-  private static final Pattern pattern = Pattern.compile("(v\\d+)");
 
   private JdbcBootstrapUtils() {}
 
@@ -89,14 +84,6 @@ public class JdbcBootstrapUtils {
       Optional<Integer> version = schemaOptions.schemaVersion();
       if (version.isPresent()) {
         return version.get();
-      }
-      Optional<String> schemaFile = schemaOptions.schemaFile();
-      if (schemaFile.isPresent()) {
-        Matcher matcher = pattern.matcher(schemaFile.get());
-        if (matcher.find()) {
-          String versionStr = matcher.group(1); // "v3"
-          return Integer.parseInt(versionStr.substring(1)); // 3
-        }
       }
     }
     return -1;
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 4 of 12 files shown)


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -53,9 +54,10 @@ public static DatabaseType fromDisplayName(String displayName) {
    * caller.
    */
   public InputStream openInitScriptResource(@Nonnull SchemaOptions schemaOptions) {
-    if (schemaOptions.schemaFile() != null) {
+    if (schemaOptions.schemaFile() != null
+        && !Objects.requireNonNull(schemaOptions.schemaFile()).isEmpty()) {
```
dimas-b: why `requireNonNull` when we know it's not `null` by the previous condition?

polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/SchemaOptions.java
```diff
@@ -33,7 +34,8 @@ public interface SchemaOptions {
 
   @Value.Check
   default void validate() {
-    if (schemaVersion() != null && schemaFile() != null) {
+    if (schemaVersion() != null
+        && (schemaFile() != null && !Objects.requireNonNull(schemaFile()).isEmpty())) {
```
dimas-b: Why bother with empty `schemaFile()` here at all? Why not validate it separately and require that it be non-empty if set (if non-null)?

dimas-b: TBH, I'd personally convert these properties to `OptionalInt` and `Optional<String>`... WDYT? (this is optional 😉 )

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
```diff
@@ -35,20 +35,28 @@
...
-    FileInputOptions fileOptions;
+    ExclusiveOptions exclusiveOptions;
 
-    @CommandLine.ArgGroup(multiplicity = "1")
-    SchemaInputOptions schemaInputOptions;
+    // This @Mixin provides independent, optional schema flags.
+    @CommandLine.Mixin SchemaInputOptions schemaInputOptions = new SchemaInputOptions();
+
+    // This static inner class encapsulates the mutually exclusive choices.
+    static class ExclusiveOptions {
```
dimas-b: Suggestion: `RootCredentialOptions`

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
```diff
@@ -35,20 +35,28 @@
...
+
+      @CommandLine.ArgGroup(exclusive = false, heading = "Standard Input Options:%n")
+      StandardInputOptions stdinOptions;
 
+      @CommandLine.ArgGroup(exclusive = false, heading = "File Input Options:%n")
+      FileInputOptions fileOptions;
```
dimas-b, line 56: I'm still not sure we end up with the right delineation of options.

I think `--print-credentials` is conceptually applicable to all cases that involve root credentials (regardless of the source of credentials).

Also `--realm` is applicable to all cases. A use may want a sub-set of realms from a file or from `--credential` (which may come from a large env. var, etc.).

Rather than treating `--credential` and `--credentials-file` as mutually exclusive, I'd prefer the tool to merge `RootCredentialsSet` from all sources. 

In the end all options are applicable and we default `--realm` to all realms within the merged `RootCredentialsSet`.

I think that would be nicer from the end user POV. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/bootstrap/SchemaOptions.java
flyrain: Can we use something like `Strings.isNullOrEmpty()` from Guava to simplify a bit?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
singhpk234: Ack, it was prompted by IJ, let me think of something better to get past IJ

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -396,7 +400,8 @@ public boolean isConstraintViolation(SQLException e) {
   }
 
   public boolean isRelationDoesNotExist(SQLException e) {
-    return RELATION_DOES_NOT_EXIST.equals(e.getSQLState());
+    return RELATION_DOES_NOT_EXIST.equals(e.getSQLState())
+        || H2_RELATION_DOES_NOT_EXIST.equals(e.getSQLState());
```
flyrain: Is it possible that H2 status codes collide with postgres ones? In that case, it may misinterpret the status code. I believe we may need introduce a type somewhere, which is also needed for `mysql` and `CockRoach DB`.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
flyrain: --> `Strings.isNullOrEmpty()`?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -154,6 +159,19 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
       RealmContext realmContext = () -> realm;
       if (!metaStoreManagerMap.containsKey(realm)) {
         DatasourceOperations datasourceOperations = getDatasourceOperations();
+        int schemaVersion =
+            JdbcBasePersistenceImpl.loadSchemaVersion(
+                datasourceOperations,
+                configurationStore.getConfiguration(
+                    realmContext, BehaviorChangeConfiguration.SCHEMA_VERSION_FALL_BACK_ON_DNE));
+        // skip validation if no schema is specified.
```
flyrain: Should we try-catch line 162 - 166, as it may throw? What the expected behavior when it throws? Skipping the version validation or error out?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
dimas-b: If `SchemaOptions` validated (and did not permit) empty values, we'd not have to validate here (even better with `Optional`)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -154,6 +159,19 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
+        int requestedSchemaVersion = getSchemaVersion(bootstrapOptions);
+        Preconditions.checkState(
+            (requestedSchemaVersion == schemaVersion)
+                || (schemaVersion == 0 || requestedSchemaVersion == -1),
```
flyrain: In case schemaVersion is 0, I think we have to make sure requestedSchemaVersion is 1.

flyrain: The version we want to bootstrapped should be  `schemaVersion` when requestedSchemaVersion is -1. The current logic will always go to `v3`. Would this be problematic?
```
      switch (schemaOptions.schemaVersion()) {
        case null -> schemaSuffix = "schema-v3.sql";
        case 1 -> schemaSuffix = "schema-v1.sql";
        case 2 -> schemaSuffix = "schema-v2.sql";
        case 3 -> schemaSuffix = "schema-v3.sql";
```

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -281,4 +299,23 @@ private void checkPolarisServiceBootstrappedForRealm(RealmContext realmContext)
           "Realm is not bootstrapped, please run server in bootstrap mode.");
     }
   }
+
+  private int getSchemaVersion(BootstrapOptions bootstrapOptions) {
```
flyrain: Combining with the comment here(https://github.com/apache/polaris/pull/2762/files#r2408431752), maybe we could have a method taking both current schema and request schema, then emit the final decision.

flyrain: Would you mind adding a test for this? I think it's important to prevent regression. I assume the test would take a list of combination of current schema and request schema, and check the result based on the inputs like the following:
1. 0, -1 -> 1
2. 1, -1 -> 1
3. 0, 1 -> 1
4. equals -> the same version
5. 2, 3 -> mismatch, error out
6. No realm, -1 -> 3
7. No realm, 2 -> 2
...

flyrain (commented):
Hi @singhpk234, thanks a lot for fixing it! Left some comments. Keep in mind these are my personal options, I'm open for discussion for these comments.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
eric-maynard: This is getting really gnarly, but it may be wise to track different error code mappings like we track different schemas and then switch on them based on `QUARKUS_DATASOURCE_DB_KIND`

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
eric-maynard: If this throws I think we want to error out the bootstrapping

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
eric-maynard: Conceptually I think you're right, though merging a conflict across `--credential` and `--credentials-file` is nontrivial, so I think it's okay if we want to leave that edge case out.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
eric-maynard: Wait, what exactly are we trying to do here? If the metastore is already bootstrapped with _any_ schema version, shouldn't bootstrapping fail? We don't support "upgrades" (e.g. type changes).

singhpk234: > If the metastore is already bootstrapped with any schema version, shouldn't bootstrapping fail 

our sql scripts are create if not exists kind of sql so when we lets say one bootstrap a new realm when there are already bootstraped with lower version but now when its ran with like another version, it just bump the schema to upgraded version without doing anything like for example some realms in v1 and new realm to added in v2,  hence this handling, am i missing something ?

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
singhpk234: Agree, IMHO would also prefer to take this up in a follow-up,  specially realm is both the place would be tricky like what if for a realm --credential had just client id and the --credetial-file has the secret do we merge ? please let me know you think

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
singhpk234: Agree, let me refactor

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: > In case schemaVersion is 0, I think we have to make sure requestedSchemaVersion is 1.

schemaVersion 0 will be 2 cases, 
- 1.0 bootstraped realms they are at v1, so i guess they should stay at v1
- no bootstraped realms doesn't matter which version they go so in this case we can go to requestedSchemaVersion

requestSchema -1 : 
- already bootstraped realms schema should stay currentSchema
- no realms bootstrapped doesn't matter which version they go so in this case we can go to most recent

Nice catch on the cases i think we just need to know if we ever ran bootstrap if yes, then 
- schemaVersion 0 means v1
- requestSchema -1 means current schema

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: used AI ;) for tests

persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
```diff
@@ -30,7 +30,7 @@ CREATE TABLE IF NOT EXISTS version (
 
 MERGE INTO version (version_key, version_value)
     KEY (version_key)
-    VALUES ('version', 2);
+    VALUES ('version', 3);
```
singhpk234, line 33: spent last night chasing test failures due to this :'(

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
dimas-b: follow up is ok from my POV :+1:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -52,26 +48,15 @@ public static DatabaseType fromDisplayName(String displayName) {
...
-                "Unknown schema version " + schemaOptions.schemaVersion());
-      }
-      ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
-      return classLoader.getResourceAsStream(this.getDisplayName() + "/" + schemaSuffix);
+  public InputStream openInitScriptResource(int schemaVersion) {
+    final String schemaSuffix;
+    switch (schemaVersion) {
+      case 1 -> schemaSuffix = "schema-v1.sql";
+      case 2 -> schemaSuffix = "schema-v2.sql";
+      case 3 -> schemaSuffix = "schema-v3.sql";
```
dimas-b: nit: could be a string format call with a `Preconditions` check (simpler)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -52,26 +48,15 @@ public static DatabaseType fromDisplayName(String displayName) {
...
+      default -> throw new IllegalArgumentException("Unknown schema version " + schemaVersion);
     }
+    ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
+    return classLoader.getResourceAsStream(this.getDisplayName() + "/" + schemaSuffix);
```
dimas-b: nit: `getClass().getResource(shortName)` is preferable IMHO

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -396,7 +400,9 @@ public boolean isConstraintViolation(SQLException e) {
   }
 
   public boolean isRelationDoesNotExist(SQLException e) {
-    return RELATION_DOES_NOT_EXIST.equals(e.getSQLState());
+    return (RELATION_DOES_NOT_EXIST.equals(e.getSQLState())
+            && databaseType == DatabaseType.POSTGRES)
+        || (H2_RELATION_DOES_NOT_EXIST.equals(e.getSQLState()) && databaseType == DatabaseType.H2);
```
dimas-b, line 405: nit: this feels a bit out of place in this PR, but the change LGTM.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
```diff
@@ -0,0 +1,104 @@
...
+   * Determines the correct schema version to use for bootstrapping a realm.
+   *
+   * @param currentSchemaVersion The current version of the database schema.
+   * @param requiredSchemaVersion The requested schema version (-1 for auto-detection).
+   * @param hasAlreadyBootstrappedRealms Flag indicating if any realms already exist.
+   * @return The calculated bootstrap schema version.
+   * @throws IllegalStateException if the combination of parameters represents an invalid state.
+   */
+  public static int getRealmBootstrapSchemaVersion(
+      int currentSchemaVersion, int requiredSchemaVersion, boolean hasAlreadyBootstrappedRealms) {
```
dimas-b, line 40: Having the `hasAlreadyBootstrappedRealms` parameter make the logic in this method's body hard to follow as it depends on external factors... Can we fold `hasAlreadyBootstrappedRealms` into this method?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
```diff
@@ -0,0 +1,104 @@
...
+  public static int getRequestedSchemaVersion(BootstrapOptions bootstrapOptions) {
+    SchemaOptions schemaOptions = bootstrapOptions.schemaOptions();
+    if (schemaOptions != null) {
+      Optional<Integer> version = schemaOptions.schemaVersion();
+      if (version.isPresent()) {
+        return version.get();
+      }
+      Optional<String> schemaFile = schemaOptions.schemaFile();
+      if (schemaFile.isPresent()) {
+        Matcher matcher = pattern.matcher(schemaFile.get());
```
dimas-b: If we have to rely on `schemaFile()` to know the schema version, I propose to remove `schemaFile()` from `SchemaOptions` and keep only the (optional) number.

What is the use case for `schemaFile()`? From my POV is it rather cryptic for end users. I think allowing only built-in SQL scripts (resources) is preferable.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -154,12 +154,27 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
+            "Effective schema version: {} for bootstrapping realm: {}",
+            effectiveSchemaVersion,
+            realm);
         try {
           // Run the set-up script to create the tables.
           datasourceOperations.executeScript(
               datasourceOperations
                   .getDatabaseType()
-                  .openInitScriptResource(bootstrapOptions.schemaOptions()));
+                  .openInitScriptResource(effectiveSchemaVersion));
```
dimas-b, line 177: `openInitScriptResource` no longer supports plain files... so why keep files in `SchemaOptions`?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
singhpk234: IIUC, the feedback is to infer `hasAlreadyBootstrappedRealms` inside function, though this would mean passing datasourceOperation here and making the API call to infer the same, and the if else still remain same,
am i missing something ? 
The present one make testing easy and the same time keeps the logic to infer if the any of the realms bootstraps in JdbcMetastoreManagerFactory hence i wrote it this way, please let me know your thoughts considering above.

dimas-b: It's more about overall confusion :sweat_smile: 

We're trying to figure out what version of the DDL script to run when bootstrap is called.

We check the version table and the existence of some other table and the bootstrap options.

However, from my POV, the big question is whether to run the DDL at all.

If tables exist and already contain realm `A` in schema `X`, then someone bootstraps realm `B` in schema `Y`, why would we (automatically) run DDL for `Y` and affect realm `A`?

I'd think we should deduce the current schema version (`X`) and if `X != Y` error out (or require a new "upgrade" flag in `SchemaOptions`).

singhpk234: >If tables exist and already contain realm A in schema X, then someone bootstraps realm B in schema Y, why would we (automatically) run DDL for Y and affect realm A? 

This is because schema table is not realm specific :( meaning if i had a realm in version 1 and then i bootstrap with version 2 then i set the schema version globally to 2 
specially now we have schemas for example 1.0 which don't have version at all, so we deduce the value as 0 which would mean yes we did bootstrap but this 0 means 1

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -52,9 +52,13 @@ public class DatasourceOperations {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DatasourceOperations.class);
 
+  // PG STATUS CODES
   private static final String CONSTRAINT_VIOLATION_SQL_CODE = "23505";
```
flyrain, line 56: Not a blocker: do we need to handle a different code for H2? It would be nice to do so, but it's not related to this PR.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -748,14 +748,28 @@ static int loadSchemaVersion(
       }
       return schemaVersion.getFirst().getValue();
     } catch (SQLException e) {
-      LOGGER.error("Failed to load schema version due to {}", e.getMessage(), e);
       if (fallbackOnDoesNotExist && datasourceOperations.isRelationDoesNotExist(e)) {
         return SchemaVersion.MINIMUM.getValue();
       }
+      LOGGER.error("Failed to load schema version due to {}", e.getMessage(), e);
```
flyrain, line 754: Minor: Do we need this error message given that we throw right after it?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
flyrain: I'd prefer to have `hasAlreadyBootstrappedRealms` as an input other than embedded within this method. It make the tests much easier by passing a boolean.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
```diff
@@ -0,0 +1,91 @@
...
+    if (currentSchemaVersion == 0) {
+      if (hasAlreadyBootstrappedRealms) {
+        // System was bootstrapped with v1 before schema versioning was introduced.
+        if (requiredSchemaVersion == -1 || requiredSchemaVersion == 1) {
+          return 1;
+        }
+      } else {
+        // A truly fresh start. Default to v3 for auto-detection, otherwise use the specified
+        // version.
+        return requiredSchemaVersion == -1 ? 3 : requiredSchemaVersion;
```
flyrain, line 57: LGTM, we could improve it by having a variable like `latestSchemaVersion`, so that we don't have to change this method every time we update the schema version.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
dimas-b: Please consider my previous comments non-blocking

persistence/relational-jdbc/src/main/resources/h2/schema-v3.sql
flyrain: It seems a copy-paste error. I guess we could add a test to validate it. We could probably do that in another PR.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java
singhpk234: Ack, let me add an ENUM in a follow-up pr

singhpk234:
Thank you for the reviews @flyrain @dimas-b @eric-maynard ! 
Merged this for now, will follow-up with the feedbacks pending

PR #2763, opened 2025-10-06
Update team entry for  eric-maynard

eric-maynard:
I'm no longer affiliated with Snowflake, so we should update this page accordingly.


HonahX (approved):

adnanhemani (approved):
🥲

PR #2765, opened 2025-10-06
Add Arguments to Various Event Records

adnanhemani:
The following events are modified with the corresponding reasons:
* AfterLoadTableEvent, AfterRegisterTableEvent, AfterCreateViewEvent, AfterLoadViewEvent: The table/view names were not in the `Load(Table/View)Response` objects, so we did not have a way to get that information in the `After*` events - so I've added those as an explicit argument.
* AfterReplaceViewEvent/AfterUpdateTableEvent: The actual MetadataUpdate objects that were applied as a result of this request could not be found earlier, so I've added the request object to the `After*` event as well.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
@@ -275,7 +275,8 @@ public class IcebergRestCatalogEventServiceDelegator
             realmContext,
             securityContext);
     polarisEventListener.onAfterLoadTable(
-        new AfterLoadTableEvent(catalogName, namespaceObj, (LoadTableResponse) resp.getEntity()));
+        new AfterLoadTableEvent(
+            catalogName, namespaceObj, table, (LoadTableResponse) resp.getEntity()));
     return resp;
   }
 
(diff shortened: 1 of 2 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -275,7 +275,8 @@ public Response loadTable(
             realmContext,
             securityContext);
     polarisEventListener.onAfterLoadTable(
-        new AfterLoadTableEvent(catalogName, namespaceObj, (LoadTableResponse) resp.getEntity()));
+        new AfterLoadTableEvent(
+            catalogName, namespaceObj, table, (LoadTableResponse) resp.getEntity()));
```
eric-maynard, line 279: In general I feel this highlights a problem with the newly-refactored events API -- there seems to be an assumption that the event content will always be accessible through the response, which really may not be the case. 

Indeed, if all the event's information can be derived from the request/response pairs, why not just pass the request/response pairs in to the listener? Why have event types at all?

eric-maynard (approved):

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: Request POJOs don't always have all the fields unfortunately either. I get the point though and agree with it generally. 

My thought process has been to ensure that all information regarding what was queried or changed as part of the request should be available in one way or another in the After event using the following guidelines:
* If the information is regarding a Polaris entity, then the entity's hierarchy should be in distinct fields within the event.
* If the request is for a modification of sorts, then the response object should be placed as-is in the After event.
* All other relevant objects should be placed as similarly to the request object into the event.

I know this isn't being followed across the board - but does this make sense? Happy to make changes where this isn't the case or look into a refactor if you believe in different guidelines.

dimas-b (approved):

flyrain (approved):
LGTM. Thanks @adnanhemani !

singhpk234 (approved):

PR #2779, opened 2025-10-08
Update the LICENSE and NOTICE files in the runtime

jbonofre:


jbonofre:
@singhpk234 as discussed together 😄

jbonofre:
After 1.2 release I will "simplify" the license and notice files.

PR #2782, opened 2025-10-09
[Catalog Federation] Add Connection Credential Vendors for Other Auth Types

XJDKC:
Add Connection Credential Vendors for Other Auth Types

This change is a prerequisite for enabling connection credential caching.
By making `PolarisCredentialManager` the central entry point for obtaining connection credentials, we can introduce caching cleanly and manage all credential flows in a consistent way.

`PolarisCredentialManager::getConnectionCredentials` will delegate the call to the appropriate credential vendor based on the authentication type:
* `SigV4` -> `SigV4ConnectionCredentialVendor`
    * `ServiceIdentityProvider`: Uses `ServiceIdentityProvider` to retrieve the IAM user’s AWS credential that represents Polaris’s service identity.
    * `StsClient`: Then use `StsClient` to assume the user-provided iam role and obtain temporary AWS credentials.
* `OAuth` -> `OAuthConnectionCredentialVendor`
    * `UserSecretsManager`: Uses `UserSecretsManager` to retrieve the OAuth2 credential from the secret manager.
* `Bearer` -> `BearerConnectionCredentialVendor`
    * `UserSecretsManager`: Uses `UserSecretsManager` to retrieve the bearer token from the secret manager.
* `Implicit` -> `ImplicitConnectionCredentialVendor`
    * A no-op vendor used when no authentication credentials are required.

With this structure, we no longer need to use `UserSecretsManager` directly inside `IcebergCatalogPropertiesProvider::asIcebergCatalogProperties`.

This unifies credential retrieval across all authentication types and simplifies the overall credential management flow.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
@@ -374,7 +374,7 @@ public class ProductionReadinessChecks {
         var type = org.apache.polaris.core.connection.AuthenticationType.valueOf(name);
         if (credentialVendors.select(AuthType.Literal.of(type)).isUnsatisfied()) {
           errors.add(
-              Error.of(
+              Error.ofSevere(
                   format(
                       "Catalog federation is enabled but no ConnectionCredentialVendor found for "
                           + "authentication type '%s'. External catalog connections using this "
@@ -384,7 +384,7 @@ public class ProductionReadinessChecks {
         }
       } catch (IllegalArgumentException e) {
         errors.add(
-            Error.of(
+            Error.ofSevere(
                 format(
                     "Invalid authentication type '%s' in SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES "
                         + "configuration.",


polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
```diff
@@ -28,15 +29,23 @@
...
   AWS_ACCESS_KEY_ID(String.class, AwsProperties.REST_ACCESS_KEY_ID, "the aws access key id", true),
   AWS_SECRET_ACCESS_KEY(
-      String.class, AwsProperties.REST_SECRET_ACCESS_KEY, "the aws access key secret", true),
-  AWS_SESSION_TOKEN(
-      String.class, AwsProperties.REST_SESSION_TOKEN, "the aws scoped access token", true),
+      String.class, AwsProperties.REST_SECRET_ACCESS_KEY, "the aws secret access key", true),
+  AWS_SESSION_TOKEN(String.class, AwsProperties.REST_SESSION_TOKEN, "the aws session token", true),
+
+  // Metadata
   EXPIRATION_TIME(
```
dimas-b: This enum value does not appear to be used anywhere.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendor.java
```diff
@@ -0,0 +1,91 @@
...
+    String clientSecret = secretsManager.readSecret(oauthParams.getClientSecretReference());
+
+    // Format credential as "clientId:clientSecret"
+    String credential = COLON_JOINER.join(oauthParams.getClientId(), clientSecret);
+
+    // Return the OAuth credential
+    // OAuth credentials don't expire - set expiration to Instant.MAX to indicate infinite validity
+    // If the credential expires, users need to update the catalog entity to rotate the credential.
+    return ConnectionCredentials.builder()
+        .putCredential(CatalogAccessProperty.OAUTH2_CREDENTIAL.getPropertyName(), credential)
```
dimas-b: There is nothing in this class that actually relates to the OAuth2 spec as far as I can tell. It only relates to the Iceberg `OAuth2Properties.CREDENTIAL` constant, which again has very little to do with OAuth2 RFCs as far as connection configuration is concerned... It's just a property name plus format spec 🤷 

What the client does with this property (whether it's OAuth2 or something else) is beyond the scope of this credential vendor. At this level we do not know how the credential is used.

I'd propose to rename this to "Iceberg Credential Vendor".

runtime/service/src/test/java/org/apache/polaris/service/credentials/connection/BearerConnectionCredentialVendorTest.java
```diff
@@ -0,0 +1,108 @@
...
+        .hasSize(1)
+        .containsEntry(
+            CatalogAccessProperty.BEARER_TOKEN.getPropertyName(), "my-bearer-token-value");
+    Assertions.assertThat(credentials.expiresAt()).contains(Instant.MAX);
+  }
+
+  @Test
+  public void testGetConnectionCredentialsWithWrongAuthType() {
+    // Setup - use a mock with wrong authentication type
+    org.apache.polaris.core.connection.ConnectionConfigInfoDpo mockConfig =
```
dimas-b: import?

runtime/service/src/test/java/org/apache/polaris/service/credentials/connection/ImplicitConnectionCredentialVendorTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    // Verify - no credentials are provided, but expiration is set to infinite
+    Assertions.assertThat(credentials.credentials()).isEmpty();
+    Assertions.assertThat(credentials.expiresAt()).contains(Instant.MAX);
+  }
+
+  @Test
+  public void testGetConnectionCredentialsWithWrongAuthType() {
+    // Setup - use a mock with wrong authentication type
+    org.apache.polaris.core.connection.ConnectionConfigInfoDpo mockConfig =
```
dimas-b: import?

runtime/service/src/test/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendorTest.java
```diff
@@ -0,0 +1,115 @@
...
+        .containsEntry(
+            CatalogAccessProperty.OAUTH2_CREDENTIAL.getPropertyName(),
+            "my-client-id:my-client-secret");
+    Assertions.assertThat(credentials.expiresAt()).contains(Instant.MAX);
+  }
+
+  @Test
+  public void testGetConnectionCredentialsWithWrongAuthType() {
+    // Setup - use a mock with wrong authentication type
+    org.apache.polaris.core.connection.ConnectionConfigInfoDpo mockConfig =
```
dimas-b: import?

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
```diff
@@ -58,7 +58,7 @@
  * <p>This is the default implementation with {@code @Priority(100)}. Custom implementations can
  * override this by providing a higher priority value.
  */
-@ApplicationScoped
+@RequestScoped
```
dimas-b, line 62: Why do we need this change?

polaris-core/src/main/java/org/apache/polaris/core/connection/BearerAuthenticationParametersDpo.java
```diff
@@ -51,9 +49,10 @@ public BearerAuthenticationParametersDpo(
 
   @Override
   public @Nonnull Map<String, String> asIcebergCatalogProperties(
-      UserSecretsManager secretsManager, PolarisCredentialManager credentialManager) {
-    String bearerToken = secretsManager.readSecret(getBearerTokenReference());
-    return Map.of(OAuth2Properties.TOKEN, bearerToken);
+      PolarisCredentialManager credentialManager) {
+    // Return only metadata properties - credentials are handled by ConnectionCredentialVendor
+    // Bearer auth has no metadata properties
+    return Map.of();
```
HonahX: Could we make this a default implementation in abstract class `AuthenticationParametersDpo` or the interface?

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
XJDKC: Because I just realized that the `DefaultServiceIdentityRegistry` is a `RequestScoped` bean.

runtime/service/src/test/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendorTest.java
XJDKC: Done!

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/CatalogAccessProperty.java
XJDKC: Revised the `CatalogAccessProperty` a bit, follow the pattern in `StorageAccessCredential`.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendor.java
XJDKC: Yes, you're right, this class currently just retrieves the OAuth credential from `UserSecretsManager` and returns it. The `PolarisCredentialManager` will then cache it in `AccessCredsCache` (not yet implemented, I plan to add that in a follow-up PR) so that we don't need to fetch the credential from `UserSecretsManager` every time we make a call, especially if the secret manager doesn't have its own cache.

The main goal of this PR is to make `PolarisCredentialManager` the single entry point for credential retrieval so we can centralize the caching logic in one place.

Regarding your suggestion, I agree, it's still beneficial to include OAuth in the class name to make the purpose of this class clear at a glance.

Right now, the class only returns the stored OAuth credential, but in the future, we could extend it to **perform the client credentials flow to obtain a temporary token directly here**, rather than relying on the Iceberg SDK to fetch a new token for every request.

WDYT?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -335,4 +337,64 @@ public ProductionReadinessCheck checkOverlappingSiblingCheckSettings(
...
+      try {
+        var type = org.apache.polaris.core.connection.AuthenticationType.valueOf(name);
+        if (credentialVendors.select(AuthType.Literal.of(type)).isUnsatisfied()) {
+          errors.add(
+              Error.of(
+                  format(
+                      "Catalog federation is enabled but no ConnectionCredentialVendor found for "
+                          + "authentication type '%s'. External catalog connections using this "
+                          + "authentication type will fail.",
+                      type),
```
HonahX, line 382: Just curious, would this check be too strong that we only consider it to be production ready if all auth types need the vendor, or this is more like a warning despite it is called "Error"?

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/BearerConnectionCredentialVendor.java
```diff
@@ -0,0 +1,84 @@
...
+ * <p>The vendor provides only the bearer token. When connecting to the remote catalog, Iceberg SDK
+ * will use this token as-is in HTTP Authorization headers. Bearer tokens typically have a limited
+ * lifetime and should be refreshed by the user when they expire.
+ *
+ * <p>This is the default implementation with {@code @Priority(100)}. Custom implementations can
+ * override this by providing a higher priority value.
+ */
+@RequestScoped
+@AuthType(AuthenticationType.BEARER)
+@Priority(100)
```
HonahX: Could we extract the priority to some constants class, like `FilterPriorities`. It will be easier for custom implementation to use and managed the prioty.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
dimas-b: `DefaultServiceIdentityRegistry` does not exist in code? :thinking: Did you mean `DefaultServiceIdentityProvider`?

Referencing an application-scoped object from a request scope is usually ok :thinking:

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendor.java
dimas-b: I fully support implementing OAuth2 flows in the future.

However, in its current form this credential vendor provides a simple pair of values, but claims the type of `AuthenticationType.OAUTH`. I personally find it misleading :thinking: 

Even if Polaris supported OAuth2 flows to obtain a token for a catalog connection, this class populates Iceberg's `credential` property, which is not a token. This property cannot hold an OAuth2 token. Instead it instruct the (Iceberg) client to perform some activities internally in order to obtain the access token. From the Polaris POV OAuth is not apparent.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
HonahX: > Referencing an application-scoped object from a request scope is usually ok

I think in case of the SigV4ConenctionCredentialVendor is the reverse? That an application-scoped object referencing a request-scoped one. The request-scoped one will always be empty or contain wrong context I suppose.

dimas-b: > [...] an application-scoped object referencing a request-scoped one.

What exactly is the other object in this case? I could not figure it out from previous comments :sweat_smile:

dimas-b: So, this is about `DefaultServiceIdentityProvider` - all clear now :sweat_smile:

XJDKC: It's `DefaultServiceIdentityProvider`, I think the only related class is this

Specifically, a `ApplicationScoped` `SigV4ConnectionCredentialVendor` refers to a `RequestScoped` `DefaultServiceIdentityProvider`.

https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/identity/provider/DefaultServiceIdentityProvider.java

HonahX: Ah yeah, sorry I did not notice the previous confusion there. IDE just popped this class when I searched first few characters 😆

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendor.java
XJDKC: Correct, how about renaming it to `OAuthClientCredentialVendor` then add a real `OAuthConnectionCredentialVendor` to perform the oauth2 flow and get a token?

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java
XJDKC: `DefaultServiceIdentityRegistry` is the previous name, I mistyped it in my comment😆

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: Yeah, naming needs improvement here. An `error` production readiness check will not block startup. Only `severe` cases break startup.

dimas-b: I'm not sure the check is logically correct. A particular deployment only needs to provide `ConnectionCredentialVendor` implementations for auth types that are actually used by created catalogs.

I think the vendor availability is implicitly checked on catalog creation already.

On startup, I think we only need to check types for existing catalogs, not for all types listed in the enum. The latter would put an undue burden on custom deployments.

XJDKC: I see, I will switch to `severe`. 
> A particular deployment only needs to provide ConnectionCredentialVendor implementations for auth types that are actually used by created catalogs. I think we only need to check types for existing catalogs, not for all types listed in the enum

The logic is: if we enable the catalog federation then all the enabled auth types must have a corresponding `ConnectionCredentialVendor`. IIRC, we don't want to defer the check during the runtime based on the feedback in the previous PR.

If vendors only implement one / two auth types for catalog federation, they need to only enable them explicitly, right?

polaris-core/src/main/java/org/apache/polaris/core/connection/BearerAuthenticationParametersDpo.java
XJDKC: Sure!

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/BearerConnectionCredentialVendor.java
XJDKC: Makes sense!

XJDKC: Done!

polaris-core/src/main/java/org/apache/polaris/core/connection/BearerAuthenticationParametersDpo.java
XJDKC: Done!

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/OAuthConnectionCredentialVendor.java
XJDKC: Renamed `OAuthConnectionCredentialVendor` to `OAuthClientCredentialVendor`!

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: SGTM 👍

HonahX:
Seems we have all the comments resolved. Thanks @dimas-b for reviewing this!

XJDKC:
@HonahX @dimas-b Thanks for reviewing the PR!

PR #2795, opened 2025-10-12
Python client: remove Python 3.9 support

MonkeyCanCode:
Python 3.9 will be EOL by the end of the month (https://peps.python.org/pep-0596/). Raise this PR to drop it once we are ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,10 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 - Amazon RDS plugin enabled, this allows polaris to connect to AWS Aurora PostgreSQL using IAM authentication.
 
+### Breaking changes
+
+- Python3.9 support will be dropped due to EOL
+
 ### New Features
 
 - Added a finer grained authorization model for UpdateTable requests. Existing privileges continue to work for granting UpdateTable, such as `TABLE_WRITE_PROPERTIES`.


MonkeyCanCode:
> Thanks for the change @MonkeyCanCode, do you wanna maybe add a change log entry for this ?

Added.

flyrain:
Thanks a lot for removing 3.9, @MonkeyCanCode ! Thanks for the review @singhpk234 @HonahX !

dimas-b:
While I do not disagree with removing Python 3.9 support, I believe this is something that deserved a `dev` ML discussion :thinking:

MonkeyCanCode:
Should we submit one now and bring back 3.9?

> While I do not disagree with removing Python 3.9 support, I believe this is something that deserved a `dev` ML discussion :thinking:

dimas-b:
I think a notice on `dev` would be nice :slightly_smiling_face: Let's keep the removal, unless someone objects on `dev`.

dimas-b:
Maybe link to the Iceberg dev thread too? https://lists.apache.org/thread/z3dhxx2wc0dhvgq9js5ktbqf4ggsn4z8

HonahX:
+1 on a notice but keep the removal. The removal will have much less impactful on our side since we haven't publish the client to PyPi

MonkeyCanCode:
> Python 3.9 will be EOL by the end of the month (https://peps.python.org/pep-0596/). Raise this PR to drop it once we are ready.

@dimas-b @HonahX here is the ML: https://lists.apache.org/thread/pxkcthq6wnjgcfycq4x1svbt6l6of9kd

PR #2798, opened 2025-10-12
CLI: Add Hive federation option

flyrain:


client/python/cli/command/catalogs.py
```diff
@@ -268,6 +273,14 @@ def _build_connection_config_info(self):
                 service_identity=service_identity,
                 remote_catalog_name=self.iceberg_remote_catalog_name
             )
+        elif self.catalog_connection_type == CatalogConnectionType.HIVE.value:
+            config = HiveConnectionConfigInfo(
+                connection_type=self.catalog_connection_type.upper(),
+                uri=self.catalog_uri,
+                authentication_parameters=auth_params,
+                service_identity=service_identity,
+                warehouse=self.hadoop_warehouse
```
flyrain: It's weird to use `hadoop_warehouse`. We might use the name `warehouse`, so that it can be applied to both Hadoop and Hive federation.

flyrain: Thoughts, suggestions? @HonahX @eric-maynard @MonkeyCanCode

MonkeyCanCode: Sry for the late response. I was away for couple of days. I don't have a strong preference over this naming. Maybe ask @eric-maynard as he initially added it in https://github.com/apache/polaris/commit/eb6b6ad50f33d0ac73c92753063438dcf428c9db.

eric-maynard: The tentative convention I had in mind was to prefix each argument with the federation type it's specific to -- namely, `ICEBERG_REMOTE_CATALOG_NAME` is specific to `iceberg` federation type, and `HADOOP_WAREHOUSE` is specific to `HADOOP`. I thought it might be unclear what just `REMOTE_CATALOG_NAME` or `WAREHOUSE` meant.

eric-maynard: to be clear I'm supportive of re-using these flags across federations types (e.g. across Hive / Hadoop) and indeed I think if we ever flesh out federation in the way that I envisioned at this time we would need to. There would be a way to federate to a Hive catalog both for Iceberg and non-Iceberg tables, and these would surely share arguments.

My only hesitation would be that the CLI 's method of handling arguments is a bit brittle and if we re-use them we should just make sure the parsing and the `--help` display behave the way we expect.

eric-maynard (dismissed):

client/python/cli/command/catalogs.py
```diff
@@ -114,6 +114,11 @@ def validate(self):
                 if self.catalog_connection_type == CatalogConnectionType.HADOOP.value:
                     if not self.hadoop_warehouse or not self.catalog_uri:
                         raise Exception(f"Missing required argument for connection type 'HADOOP':"
+                                f" {Argument.to_flag_name(Arguments.HADOOP_WAREHOUSE)}"
+                                f" and {Argument.to_flag_name(Arguments.CATALOG_URI)}")
+                elif self.catalog_connection_type == CatalogConnectionType.HIVE.value:
+                    if not self.hadoop_warehouse or not self.catalog_uri:
+                        raise Exception(f"Missing required argument for connection type 'HIVE':"
```
eric-maynard: nit: `--help` was updated to list `iceberg-rest, hadoop, hive` so these logs (and the tests) should follow that same convention

MonkeyCanCode (approved):

Issue #1822, opened 2025-06-06
Python client is broken due to dependency conflict


HonahX:
### Describe the bug

The current python client in the main is unusable due to a dependency conflict. Specifically, when generating the client using `open-api-generator==7.12.0`, the client code requires `urllib3 >= 2.1.0, < 3.0.0`, ([ref](https://github.com/OpenAPITools/openapi-generator/issues/20826#issuecomment-2713210369)). However, our client also needs boto3, which is only compatible with `urllib3 2.x` for Python 3.10+

This leads to 2 ways of solving this
1. Generate the client using open-api-generator==7.11.0 to remove the requirement of urllib 2.x: https://github.com/OpenAPITools/openapi-generator/issues/20826#issuecomment-2710837978
2. Drop support for python 3.9

Since python 3.9's supports ends in October 2025, which means it is still active now, I think we should go with approach 1 to downgrade the generator version used

### To Reproduce

For reference, you can reproduce the client error by running spark getting-started example. You will get
```
File /opt/conda/lib/python3.11/site-packages/urllib3/poolmanager.py:303, in PoolManager.connection_from_host(self, host, port, scheme, pool_kwargs)
    300 request_context["port"] = port
    301 request_context["host"] = host
--> 303 return self.connection_from_context(request_context)

File /opt/conda/lib/python3.11/site-packages/urllib3/poolmanager.py:326, in PoolManager.connection_from_context(self, request_context)
    324 if not pool_key_constructor:
    325     raise URLSchemeUnknown(scheme)
--> 326 pool_key = pool_key_constructor(request_context)
    328 return self.connection_from_pool_key(pool_key, request_context=request_context)

File /opt/conda/lib/python3.11/site-packages/urllib3/poolmanager.py:147, in _default_key_normalizer(key_class, request_context)
    144 if context.get("key_blocksize") is None:
    145     context["key_blocksize"] = _DEFAULT_BLOCKSIZE
--> 147 return key_class(**context)

TypeError: PoolKey.__new__() got an unexpected keyword argument 'key_ca_cert_data'
```

PR #2800, opened 2025-10-13
Bump version for python deps

MonkeyCanCode:
This PR bumps the version for poetry and removed the fixed version for virtualenv. Also, it matches the version for openapi-generator-cli.

We used fixed version for poetry and virtualenv due to https://github.com/python-poetry/poetry/issues/10504#issuecomment-3176923981. This is fixed via https://github.com/pypa/virtualenv/issues/2944

We used fixed version of openapi-generator-cli due to https://github.com/apache/polaris/issues/1822#issue-3123175447. Even we dropped support for python 3.9 via https://github.com/apache/polaris/pull/2795, I noticed there are more issues with when using version of openapi-generator-cli. After some debugging, this is due to https://github.com/apache/polaris/pull/2801. Thus the latest working version with current code is `7.12.0`. Once https://github.com/apache/polaris/pull/2801 is merged and bump the openapi spec version from `3.0.3` to `3.1.1`, we can then bump up the version again.

### DIFF
--- a/regtests/requirements.txt
+++ b/regtests/requirements.txt
@@ -17,4 +17,4 @@
 # under the License.
 #
 
-poetry==2.1.4
+poetry==2.2.1


client/python/pyproject.toml
```diff
@@ -45,7 +45,7 @@ repository = "https://github.com/apache/polaris/"
 polaris = "cli.polaris_cli:main"
 
 [tool.poetry]
-requires-poetry = "==2.1.4"
+requires-poetry = ">=2.1.4"
```
HonahX: How about we still pin the poetry version but upgrade it to latest for example. Ideally this can help reduce confusion caused by tool version mismatch

client/python/pyproject.toml
```diff
@@ -61,19 +61,15 @@ types-python-dateutil = ">= 2.8.19.14"
 mypy = ">=1.18, <=1.18.2"
 pyiceberg = "==0.10.0"
 pre-commit = "==4.3.0"
-openapi-generator-cli = "==7.15.0"
+openapi-generator-cli = "==7.12.0"
```
HonahX, line 64: It will be good if we upgrade the generator version along with the one we used in java:
https://github.com/apache/polaris/blob/dbc21a3d289e140090f621838c664062ba0122e3/gradle/libs.versions.toml#L115

Let's keep it 7.12 for this PR : )

client/python/pyproject.toml
MonkeyCanCode: Good thinking. Let me do that.

client/python/pyproject.toml
MonkeyCanCode: Thanks for sharing the generator version in gradle (didn't saw that earlier), but yeah, it appear both are 7.12.0 (which the latest working version unless we will do 3.1.1)

PR #1424, opened 2025-04-22
Add KMS support for S3

fivetran-ashokborra:

PR #2802, opened 2025-10-13
adding support to use a kms key for s3 buckets data encryption (AWS only)

fabio-rizzo-01:
The addition of a kms keys used to encrypt data in the s3 bucket

For anyone wish to use data encryption.

Yes, the addition of 2 new params in the catalog creation : currentKmsKey,allowedKmsKeys

Tested in AWS EKS

### CHANGELOG.md
 Updated catalogs creation to include AWS current kms key and allowed kms key, as extra params in the storage config info, to be used for S3 data encryption

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - `client.region` is no longer considered a "credential" property (related to Iceberg REST Catalog API).
 - Relaxed the requirements for S3 storage's ARN to allow Polaris to connect to more non-AWS S3 storage appliances. 
 - Added checksum to helm deployment so that it will restart when the configmap has changed.
+- Added Windows support for Python client
 
 ### Deprecations
 
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -53,7 +53,6 @@ github:
         # Contexts are the names of checks that must pass. This is the value
         # of the job's `name` property if it's present.
         contexts:
-          - markdown-link-check
           - "Unit Tests"
           - "Quarkus Tests"
           - "Quarkus Integration Tests"
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -67,6 +67,7 @@ dependencies {
     api(project(":polaris-persistence-nosql-mongodb"))
 
     api(project(":polaris-persistence-nosql-maintenance-api"))
+    api(project(":polaris-persistence-nosql-maintenance-impl"))
     api(project(":polaris-persistence-nosql-maintenance-cel"))
     api(project(":polaris-persistence-nosql-maintenance-spi"))
 
--- a/build-logic/src/main/kotlin/copiedcode/CopiedCodeCheckerPlugin.kt
+++ b/build-logic/src/main/kotlin/copiedcode/CopiedCodeCheckerPlugin.kt
@@ -151,7 +151,7 @@ abstract class CheckCopiedCodeMentionsExistTask : DefaultTask() {
       logger.error(
         """
         The following {} files mentioned in {} do not exist, fix the {} file.
-        
+
         {}
         """
           .trimIndent(),
--- a/build-logic/src/main/kotlin/publishing/rootProject.kt
+++ b/build-logic/src/main/kotlin/publishing/rootProject.kt
@@ -46,19 +46,29 @@ internal fun configureOnRootProject(project: Project) =
       description =
         "Generate a source tarball for a release to be uploaded to dist.apache.org/repos/dist"
 
+      outputs.upToDateWhen { false }
+      outputs.cacheIf { false }
+
       val e = project.extensions.getByType(PublishingHelperExtension::class.java)
       doFirst { mkdir(e.distributionDir) }
 
-      executable = "git"
-      args(
-        "archive",
-        "--prefix=${e.baseName.get()}/",
-        "--format=tar.gz",
-        // use a fixed mtime for reproducible tarballs, using the same timestamp as jars do
-        "--mtime=1980-02-01 00:00:00",
-        "--output=${e.sourceTarball.get().asFile.relativeTo(projectDir)}",
-        "HEAD",
-      )
+      // Use a fixed mtime for reproducible tarballs, using the same timestamp as jars do.
+      // Also don't use the git-internal gzip as it's not stable, see
+      // https://reproducible-builds.org/docs/archives/.
+      commandLine =
+        listOf(
+          "bash",
+          "-c",
+          """
+        git \
+          archive \
+          --prefix="${e.baseName.get()}/" \
+          --format=tar \
+          --mtime="1980-02-01 00:00:00" \
+          HEAD | gzip -6 --no-name > "${e.sourceTarball.get().asFile.relativeTo(projectDir)}"
+          """
+            .trimIndent(),
+        )
       workingDir(project.projectDir)
 
       outputs.file(e.sourceTarball)
--- a/client/python/generate_clients.py
+++ b/client/python/generate_clients.py
@@ -36,12 +36,12 @@ import ast
 
 # Paths
 CLIENT_DIR = Path(__file__).parent
-HEADER_DIR = os.path.join(CLIENT_DIR, "templates")
-SPEC_DIR = os.path.join(CLIENT_DIR, "spec")
-POLARIS_MANAGEMENT_SPEC = os.path.join(SPEC_DIR, "polaris-management-service.yml")
-ICEBERG_CATALOG_SPEC = os.path.join(SPEC_DIR, "iceberg-rest-catalog-open-api.yaml")
-POLARIS_CATALOG_SPEC = os.path.join(SPEC_DIR, "polaris-catalog-service.yaml")
-OPEN_API_GENERATOR_IGNORE = os.path.join(CLIENT_DIR, ".openapi-generator-ignore")
+HEADER_DIR = CLIENT_DIR / "templates"
+SPEC_DIR = CLIENT_DIR / "spec"
+POLARIS_MANAGEMENT_SPEC = SPEC_DIR / "polaris-management-service.yml"
+ICEBERG_CATALOG_SPEC = SPEC_DIR / "iceberg-rest-catalog-open-api.yaml"
+POLARIS_CATALOG_SPEC = SPEC_DIR / "polaris-catalog-service.yaml"
+OPEN_API_GENERATOR_IGNORE = CLIENT_DIR / ".openapi-generator-ignore"
 
 # Open API Generator Configs
 PACKAGE_NAME_POLARIS_MANAGEMENT = (
@@ -58,20 +58,20 @@ KEEP_TEST_FILES = [
 ]
 EXCLUDE_PATHS = [
     Path(".gitignore"),
-    Path(".openapi-generator/"),
+    Path(".openapi-generator"),
     Path(".openapi-generator-ignore"),
-    Path(".pytest_cache/"),
+    Path(".pytest_cache"),
     Path("test/test_cli_parsing.py"),
-    Path("apache_polaris/__pycache__/"),
-    Path("apache_polaris/cli/"),
-    Path("apache_polaris/sdk/__pycache__/"),
-    Path("apache_polaris/sdk/catalog/__pycache__/"),
-    Path("apache_polaris/sdk/catalog/models/__pycache__/"),
-    Path("apache_polaris/sdk/catalog/api/__pycache__/"),
-    Path("apache_polaris/sdk/management/__pycache__/"),
-    Path("apache_polaris/sdk/management/models/__pycache__/"),
-    Path("apache_polaris/sdk/management/api/__pycache__/"),
-    Path("integration_tests/"),
+    Path("apache_polaris/__pycache__"),
+    Path("apache_polaris/cli"),
+    Path("apache_polaris/sdk/__pycache__"),
+    Path("apache_polaris/sdk/catalog/__pycache__"),
+    Path("apache_polaris/sdk/catalog/models/__pycache__"),
+    Path("apache_polaris/sdk/catalog/api/__pycache__"),
+    Path("apache_polaris/sdk/management/__pycache__"),
+    Path("apache_polaris/sdk/management/models/__pycache__"),
+    Path("apache_polaris/sdk/management/api/__pycache__"),
+    Path("integration_tests"),
     Path(".github/workflows/python.yml"),
     Path(".gitlab-ci.yml"),
     Path("pyproject.toml"),
@@ -86,9 +86,9 @@ EXCLUDE_PATHS = [
     Path("README.md"),
     Path("generate_clients.py"),
     Path(".venv"),
-    Path("dist/"),
-    Path("templates/"),
-    Path("spec/"),
+    Path("dist"),
+    Path("templates"),
+    Path("spec"),
     Path("PKG-INFO"),
 ]
 EXCLUDE_EXTENSIONS = [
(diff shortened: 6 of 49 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -242,6 +244,21 @@ private IamPolicy policyString(
...
+  private static void addKmsKeyPolicy(String kmsKeyArn, IamPolicy.Builder policyBuilder) {
+    if (kmsKeyArn != null) {
+      IamStatement.Builder allowKms =
+          IamStatement.builder()
+              .effect(IamEffect.ALLOW)
+              .addAction("kms:GenerateDataKeyWithoutPlaintext")
+              .addAction("kms:Encrypt")
+              .addAction("kms:DescribeKey")
+              .addAction("kms:Decrypt")
+              .addAction("kms:GenerateDataKey");
```
collado-mike: can we add the kms encryption context to this statement?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -242,6 +244,21 @@ private IamPolicy policyString(
...
 
+  private static void addKmsKeyPolicy(String kmsKeyArn, IamPolicy.Builder policyBuilder) {
+    if (kmsKeyArn != null) {
+      IamStatement.Builder allowKms =
+          IamStatement.builder()
+              .effect(IamEffect.ALLOW)
+              .addAction("kms:GenerateDataKeyWithoutPlaintext")
+              .addAction("kms:Encrypt")
+              .addAction("kms:DescribeKey")
+              .addAction("kms:Decrypt")
```
singhpk234: should we give encrypt or decrypt based on the priviledge ? for example TABLE_READ should get decrypt and TABLE_READ_AND_WRITE should both encrypt and decrypt ?

spec/polaris-management-service.yml
```diff
@@ -1103,6 +1103,10 @@ components:
               type: string
               description: the aws user arn used to assume the aws role
               example: "arn:aws:iam::123456789001:user/abc1-b-self1234"
+            kmsKeyArn:
+              type: string
+              description: the aws kms key arn used to encrypt s3 data
```
singhpk234, line 1108: by s3 data we mean data stored in s3 or iceberg data files ? 
if its just data stored in s3, a part of iceberg metadata is generated by polaris itself, do we need to plumb the encryption to the fileIO polaris uses ? 

if its just iceberg data files how are we making sure its not used against iceberg metadata files such as manifest lists ?

fabio-rizzo-01: the key is used by s3 to do encryption of any data in the bucket at rest, so as long as the IAM role has got the right entitlements on the key everything will work fine. I don't think anything needs to be changed on the polaris side.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
fabio-rizzo-01: are you suggesting to allow the user to specify a new param during catalog creation for the encryption context condition?

spec/polaris-management-service.yml
singhpk234: I am talking of setting these properties in the FileIO which polaris uses - https://iceberg.apache.org/docs/nightly/aws/#s3-server-side-encryption or are you relying on these catalog properties seeping automatically to the FileIO

dimas-b:
What's the relationship of this PR and #1424 ?

spec/polaris-management-service.yml
```diff
@@ -1103,6 +1103,10 @@ components:
               type: string
               description: the aws user arn used to assume the aws role
               example: "arn:aws:iam::123456789001:user/abc1-b-self1234"
+            kmsKeyArn:
```
dimas-b: By convention, adding REST API parameters needs to be discussed on the `dev` ML. Could you start a thread there and reference this PR?

fabio-rizzo-01:
> What's the relationship of this PR and #1424 ?

sorry I didn't see that one.

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### New Features
 
+- Updated catalogs creation to include AWS kms key ,as an extra param in the storage config info, to be used in S3 data encryption
```
dimas-b:
```suggestion
- Updated catalogs creation to include AWS kms key, as an extra param in the storage config info (`kmsKeyArn`), to be used in S3 data encryption.
```

spec/polaris-management-service.yml
fabio-rizzo-01: There is already a thread for this PR

dimas-b:
Re: #1424 - that PR appears to be dormant, so it's fine to offer an alternative. I was just trying to understand whether the feature is the same or not :)

spec/polaris-management-service.yml
dimas-b: My mistake... I did not see it :facepalm: 

For ref: https://lists.apache.org/thread/to7lbdw1k6dym5c6gnk1nm32vqdw24qk

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -178,7 +177,10 @@ private IamPolicy policyString(
     Map<String, IamStatement.Builder> bucketListStatementBuilder = new HashMap<>();
     Map<String, IamStatement.Builder> bucketGetLocationStatementBuilder = new HashMap<>();
 
-    String arnPrefix = arnPrefixForPartition(awsPartition);
+    String arnPrefix = arnPrefixForPartition(storageConfigurationInfo.getAwsPartition());
+    String kmsKeyArn = storageConfigurationInfo.getKmsKeyArn();
```
dimas-b: Can / will this work with MinIO KMS? Cf. https://github.com/minio/minio/blob/master/docs/kms/README.md

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
collado-mike: No, I mean for SSE, the encryption context is always the path to the file. In our IAM policy statement, we can specify that the key can be used only to en/decrypt files in the encrpytion context pattern we specify - e.g., see https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html and https://docs.aws.amazon.com/kms/latest/developerguide/encrypt_context.html#encryption-context-authorization . The user doesn't need to specify this parameter - it's automatic.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
fabio-rizzo-01: I will need to test that.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
fabio-rizzo-01: I don't think it is needed, the policy for the key and S3 bucket is restrictive enough and adding more restriction would cause the GenerateDataKey operation during the assume role to probably fail.

This is an example of a policy generated while creating a table:

 ```
IamPolicy(version=2012-10-17, statements=[

 IamStatement(effect=IamEffect(value=Allow), actions=[IamAction(value=s3:PutObject), IamAction(value=s3:DeleteObject)], resources=[IamResource(value=arn:aws:s3:::app-id-bucket/fns/t3/*)]), 

 IamStatement(effect=IamEffect(value=Allow), actions=[IamAction(value=kms:GenerateDataKeyWithoutPlaintext), IamAction(value=kms:DescribeKey), IamAction(value=kms:Decrypt), IamAction(value=kms:GenerateDataKey), IamAction(value=kms:Encrypt)], resources=[IamResource(value=arn:aws:kms:us-east-1:*********:key/id)]), 

 IamStatement(effect=IamEffect(value=Allow), actions=[IamAction(value=s3:ListBucket)], resources=[IamResource(value=arn:aws:s3:::app-id-bucket)], conditions=[IamCondition(operator=StringLike, key=s3:prefix, value=fns/t3/*)]), 

 IamStatement(effect=IamEffect(value=Allow), actions=[IamAction(value=s3:GetBucketLocation)], resources=[IamResource(value=arn:aws:s3:::app-id-bucket)]), 

 IamStatement(effect=IamEffect(value=Allow), actions=[IamAction(value=s3:GetObject), IamAction(value=s3:GetObjectVersion)], resources=[IamResource(value=arn:aws:s3:::app-id-bucket/fns/t3/*)])])


```
Also in https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html#encryption-context
it says that the bucket path is added by default in the encryption context.

Checking the encryption context in CloudTrail for the GenerateDataKey request I can see:

```
"encryptionContext": {
"aws:s3:arn": "arn:aws:s3:::app-id-*****"
}

```

File: polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java (hunk 6 above)
Comment (dimas-b) [reply polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java]: It's fine either way, but if it did not work with MinIO it would be good to mention AWS (only) in the PR description.

Also, I'm pretty sure we'll have to support KMS in MinIO at some point, so if current changes were reusable it would be great (but not required).

File: spec/polaris-management-service.yml (hunk 3 above)
Comment (fabio-rizzo-01) [reply spec/polaris-management-service.yml:1108]: I'm not sure I follow I might be missing something, how would you specify the kms key in Polaris without adding in the storage config, also we need to add the policy in there to be able to use right? thx

**Comment by fabio-rizzo-01** [comment]
Updated the PR with temporary changes, to see if this is the right direction. The build will fail because tests have been updated yet.

File: polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java [hunk 7]
```diff
@@ -136,7 +137,8 @@ public AccessConfig getOrGenerateSubScopeCreds(
                   k.allowedListAction(),
                   k.allowedReadLocations(),
                   k.allowedWriteLocations(),
-                  k.refreshCredentialsEndpoint());
+                  k.refreshCredentialsEndpoint(),
+                  props);
```
dimas-b: If `prop` is an argument to generating `AccessConfig` is should be part of the cache key.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -75,11 +80,15 @@ public AccessConfig getSubscopedCreds(
       boolean allowListOperation,
       @Nonnull Set<String> allowedReadLocations,
       @Nonnull Set<String> allowedWriteLocations,
-      Optional<String> refreshCredentialsEndpoint) {
+      Optional<String> refreshCredentialsEndpoint,
+      Map props) {
+    LOGGER.info("Getting subscoped creds props: {}", props);
+    String kmsKey = props.get("s3.sse.key") != null ? props.get("s3.sse.key").toString() : null;
```
dimas-b: I personally think that "storage integration" code should work independently of table properties. In other words Polaris should own storage integration and `AccessConfig`. Table properties are settable by the end users outside the context of a catalog (e.g. via registerTable). Mixing both sources of integration config is confusing IMHO.

Is the intention here to support different KMS config per table?

Polaris may need to be able to get `AccessConfig` in order to load table metadata (which has table properties)... Does this not create a chicken-and-egg problem?

That said, I'm open to a wider discussion about this on the `dev` ML.

fabio-rizzo-01: Yes that is to support the usage of the s3 properties defined in iceberg, we don't really need to pass down props or have that code in there as long as we can agree on the code https://github.com/apache/polaris/pull/2802/files#diff-d305f7a426a7690c576722c114257792b3fcee726624655d15893b71499827f8R274. 
If no kms key was defined in the polaris AWS storage then we allow the usage of any keys owned by that account.

dimas-b: Before we discuss `arnKeyAll` (and I do not have any objections there), I think we need to resolve the conceptual issue of using metadata properties for generating `AccessConfig`. Polaris may need `AccessConfig` in order to read metadata JSON files (which have table properties)... How can this work in principle?

dimas-b: I propose concentrating on per-catalog config for now (this PR) and addressing per-table config / concerns later.

ML: https://lists.apache.org/thread/2k5dqhz35fcshs2o083hv7pxyrkgrbk9

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### New Features
 
+- Updated catalogs creation to include AWS current kms key and allowed kms key,as extra params in the storage config info, to be used for S3 data encryption
```
dimas-b:
```suggestion
- Added KMS properties (optional) to catalog storage config to enable S3 data encryption.
```

spec/polaris-management-service.yml
```diff
@@ -1103,6 +1103,16 @@ components:
               type: string
               description: the aws user arn used to assume the aws role
               example: "arn:aws:iam::123456789001:user/abc1-b-self1234"
+            currentKmsKey:
+              type: string
+              description: the aws kms key arn used to encrypt s3 data
+              example: "arn:aws:kms::123456789001:key/01234578"
+            allowedKmsKeys:
+              type: array
+              description: the aws kms keys arn used to encrypt s3 data
```
dimas-b:
```suggestion
              description: The list of kms keys that this catalog and its clients are allow to use for reading s3 data
```

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
```diff
@@ -70,6 +72,8 @@ public void testJsonFormat() throws JsonProcessingException {
                 + "\"properties\":{\"default-base-location\":\"s3://test/\"},"
                 + "\"storageConfigInfo\":{"
                 + "\"roleArn\":\"arn:aws:iam::123456789012:role/test-role\","
+                + "\"currentKmsKey\":\"arn:aws:kms:us-east-1:012345678901:key/allowed-key-1\","
+                + "\"allowedKmsKeys\":[],"
```
dimas-b, line 103: Could you make a separate test case for this? This "minimal" config is intended to test that older clients do not get new optional properties.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -239,7 +254,86 @@ private IamPolicy policyString(
     bucketGetLocationStatementBuilder
         .values()
         .forEach(statementBuilder -> policyBuilder.addStatement(statementBuilder.build()));
-    return policyBuilder.addStatement(allowGetObjectStatementBuilder.build()).build();
+    var r = policyBuilder.addStatement(allowGetObjectStatementBuilder.build()).build();
+    LOGGER.info("Policies  {}", r);
```
dimas-b: do we really need to log this all the time? Maybe `.debug()`?

nit: double space char in the message.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -239,7 +254,86 @@ private IamPolicy policyString(
...
+    }
+
+    if (hasAllowedKeys) {
+      addAllowedKmsKeyResources(allowedKmsKeys, allowKms);
+    }
+
+    // Add KMS statement if we have any KMS key configuration
+    if (hasCurrentKey || hasAllowedKeys) {
+      policyBuilder.addStatement(allowKms.build());
+    } else if (!canEncrypt) {
```
dimas-b: the `canEncrypt` flag really means `canWrite`, does it not? 🤔

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -239,7 +254,86 @@ private IamPolicy policyString(
...
+    if (canEncrypt) {
+      allowKms.addAction("kms:Encrypt");
+    }
+
+    return allowKms;
+  }
+
+  private static void addKmsKeyResource(String kmsKeyArn, IamStatement.Builder allowKms) {
+    if (kmsKeyArn != null) {
+      LOGGER.info("Adding KMS key policy for key {}", kmsKeyArn);
```
dimas-b: too verbose? why not `.debug()` (in other places too)?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -63,6 +64,14 @@ public String getFileIoImplClassName() {
   @Nullable
   public abstract String getRoleARN();
 
+  /** KMS Key ARN for server-side encryption,used for writes, optional */
+  @Nullable
+  public abstract String currentKmsKey();
```
dimas-b: it might be preferable to follow the existing `get*` method name pattern for JSON attributes in this class.

polaris-core/src/test/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfoTest.java
```diff
@@ -70,7 +70,8 @@ public void testStsEndpoint() {
 
   private static ImmutableAwsStorageConfigurationInfo.Builder newBuilder() {
     return AwsStorageConfigurationInfo.builder()
-        .roleARN("arn:aws:iam::123456789012:role/polaris-test");
+        .roleARN("arn:aws:iam::123456789012:role/polaris-test")
+        .currentKmsKey("arn:aws:kms:us-east-1:012345678901:key/444343245");
```
dimas-b: Is KMS key required in this case?

dimas-b:
@fabio-rizzo-01 : Also, CI failed :shrug:

fabio-rizzo-01:
> @fabio-rizzo-01 : Also, CI failed 🤷

yes I know, I'm fixing it and I'll check your comments.

fabio-rizzo-01:
@dimas-b PR update and test have passed. The markdown test failed nothing to do with my changes it is getting 403s for some medium links that are in the readme files, can you rerun it please?

dimas-b:
@fabio-rizzo-01 : markdown link checks are inherently flaky, unfortunately, as they depend on external sites to be "up" :shrug:

dimas-b:
Re: markdown link checks : #3102

dimas-b:
@fabio-rizzo-01 : it looks like this PR has a lot of commits now... most of them unrelated to KMS :thinking: Please rebase with KMS commits only... in any case rebasing is also needed to get unblocked on the markdown check. Thx!

fabio-rizzo-01:
> @fabio-rizzo-01 : it looks like this PR has a lot of commits now... most of them unrelated to KMS 🤔 Please rebase with KMS commits only... in any case rebasing is also needed to get unblocked on the markdown check. Thx!

yeah let me fix it, not sure what happened

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
```diff
@@ -70,6 +71,36 @@ public void testJsonFormat() throws JsonProcessingException {
                 + "\"properties\":{\"default-base-location\":\"s3://test/\"},"
                 + "\"storageConfigInfo\":{"
                 + "\"roleArn\":\"arn:aws:iam::123456789012:role/test-role\","
+                + "\"allowedKmsKeys\":[],"
```
dimas-b, line 74: This is a bit of a concern, unfortunately. (Sorry, I did not notice it before because of the other test code changes). 

This means old clients will see the new allowedKmsKeys and other new properties (after a server upgrade) even if they were not set in their catalogs... IMHO, it's a potential backward compatibility issue.

However, our [history](https://github.com/apache/polaris/pull/2012#discussion_r2208655528) shows that the community might be ok with that kind of change.
From my POV, it would be preferable to avoid changes on the wire, if possible... Meaning that defaults on newly added properties should not be visible to old clients... WDYT?

If that is too cumbersome, I'd like some more reviewer to express their opinions whether exposing new properties by default is acceptable.

fabio-rizzo-01:
updated, it looks good now @dimas-b.

spec/polaris-management-service.yml
```diff
@@ -1103,6 +1103,16 @@ components:
               type: string
               description: the aws user arn used to assume the aws role
               example: "arn:aws:iam::123456789001:user/abc1-b-self1234"
+            currentKmsKey:
+              type: string
+              description: the aws kms key arn used to encrypt s3 data
+              example: "arn:aws:kms::123456789001:key/01234578"
+            allowedKmsKeys:
+              type: array
+              description: The list of kms keys that this catalog and its clients are allow to use for reading s3 data
```
dimas-b, line 1112: typo: `allow` -> `allowed`

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
dimas-b: I wonder, why does `allowedKmsKeys` show up in the default config JSON? I believe `null` values are not serialized... Is it not `null` by default? Could we make all new entries `null` by default to reduce impact to default data on the wire and handle those `null`s in java code?

fabio-rizzo-01: The problem is the immutable class that sets them to empty collection by default , this only happens with collections, and I didn't want to touch those setting because they might cause a lot more changes.
IMHO i don't think seeing the empty field returned is a massive issue but I'll leave the judgment to you guys.

dimas-b: I did a quick test with an old version of Polaris CLI, and it was able to create a non-KMS catalog without any errors reported. 

I'm personally ok with this change.

dimas-b:
Assuming lazy consensus from other reviewers... Merging on Nov 25. Cf. https://lists.apache.org/thread/bc7hm1wq1fj6tdyffsm36rwqknyprtgr

dimas-b:
Thanks for your contribution, @fabio-rizzo-01 !

spec/polaris-management-service.yml
dimas-b: #3172

PR #2810, opened 2025-10-14
Add findPrincipalById helper

XN137:
this simplifies frequent usage of the lower level `loadEntity` api (similar to the
existing `findPrincipalByName` helper)

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
@@ -1182,12 +1182,17 @@ public class PolarisAdminService {
               "Failed to %s secrets for principal '%s'",
               shouldReset ? "reset" : "rotate", principalName));
     }
-    PrincipalEntity newPrincipal =
-        metaStoreManager
-            .findPrincipalById(getCurrentPolarisContext(), currentPrincipalEntity.getId())
-            .orElseThrow();
+    Optional<PrincipalEntity> updatedPrincipalEntity =
+        metaStoreManager.findPrincipalById(
+            getCurrentPolarisContext(), currentPrincipalEntity.getId());
+    if (updatedPrincipalEntity.isEmpty()) {
+      throw new IllegalStateException(
+          String.format(
+              "Failed to reload principal '%s' by id: %s",
+              principalName, currentPrincipalEntity.getId()));
+    }
     return new PrincipalWithCredentials(
-        newPrincipal.asPrincipal(),
+        updatedPrincipalEntity.get().asPrincipal(),
         new PrincipalWithCredentialsCredentials(
             newSecrets.getPrincipalClientId(), newSecrets.getMainSecret()));
   }


polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -470,6 +470,20 @@ default Optional<PrincipalEntity> findRootPrincipal(PolarisCallContext polarisCa
     return findPrincipalByName(polarisCallContext, PolarisEntityConstants.getRootPrincipalName());
   }
 
+  default Optional<PrincipalEntity> findPrincipalById(
```
dimas-b, line 473: Having this helper method in principle SGTM. However, given that `PolarisMetaStoreManager` is one of the main plugin / integrations points, I'm not sure about expanding its API surface with one more method.

XN137: why do you think its expanding the API surface?
its implementation is just reusing existing methods but encodes more information about the proper way to use them and about the type of entity they will return.
also simplifying code around "exception" handling.

these defaults methods probably never need to be re-implemented by actual classes.

XN137:
> I think we need to resume work on untangling different API concerns currently mixed in PolarisMetaStoreManager

in what way is looking up a principal a different api concern in this case?

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
dimas-b: It gives callers two options: 1) use the old method 2) use the helper methods... In my mind that expands the API.

I'm thinking that adding methods like this is probably an indication that we may need to rework the interface as a whole. 

This is more of a point for thinking about, not an objection, really.

dimas-b:
> in what way is looking up a principal a different api concern in this case?

Principal is an Entity, but has dedicated access methods that are not applicable to other entities.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1182,15 +1182,12 @@ public void deletePrincipal(String name) {
...
-        PolarisEntity.of(
-            metaStoreManager.loadEntity(
-                getCurrentPolarisContext(),
-                0L,
-                currentPrincipalEntity.getId(),
-                currentPrincipalEntity.getType()));
+    PrincipalEntity newPrincipal =
+        metaStoreManager
+            .findPrincipalById(getCurrentPolarisContext(), currentPrincipalEntity.getId())
+            .orElseThrow();
```
flyrain: Are we supposed to throw here?  If yes, can we make the error message a bit more specific?
`No value present` --> `Failed to find the principal by id: {}`

XN137: ok i've added a commit to provide a more detailed error message.

however looking at the code in the beginning we do:
```
PrincipalEntity currentPrincipalEntity = getPrincipalByName(resolutionManifest, principalName);
```
which would fail hard if the principal did not exist.
and then in `rotatePrincipalSecrets` we are updating the principal behind `currentPrincipalEntity.getId()`.

so we can be quite sure that another call to `findPrincipalById` will return a non-empty result.
only if someone externally was dropping the principal with precise timing could we hit this scenario and the old code would have resulted in an NPE in this case.

on that note, seems like all callers of `rotatePrincipalSecrets` are looking up the updated principal afterwards... makes me wonder whether the method should just return it directly (it loads it already internally).

flyrain (approved):

PR #2828, opened 2025-10-16
Dependencies: use one testcontainers dependency version

snazy:
The localstack testcontainers dependency, "included" in the whole testcontainers project, is using a different (and older, 1.21.3 vs 2.0.0) version than the testcontainers-bom.

This change prefers to use the bom as the "only reference".


dimas-b (approved):

flyrain (approved):
+1. Thanks a lot for the change!

PR #2831, opened 2025-10-17
[Python] Add more tests cases for policy CLI

CodingBangboo:

The current policy tests do not include cases for attaching, getting applicable policies, and detaching policies for catalogs. In addition, the policy tests do not include a case to check the policy's inheritance behavior through the API. This pull request adds the above two scenarios' test cases.


singhpk234 (approved):
Thank you for the contribution @NuoyaJiang ! LGTM

Issue #2630, opened 2025-09-19
PropertyMapEventListener is incomplete / underspecified


adutra:
### Describe the bug

Currently, `PropertyMapEventListener` implements only one method from `PolarisEventListener` – all others are ignored. It is unclear if this is on purpose.

Furthermore:

- The API is not ideal: `transformAndSendEvent` takes a `HashMap` parameter, and it is unclear whether the requirement for a `HashMap` (instead of `Map`) is legit.
- The only implemented method – `onAfterRefreshTable` does not serialize all the event fields. Again, it is unclear if this is by design or an oversight.

And finally, the whole point of having an abstract class that transforms events into Maps, only to have the Maps transformed into Json (cf. `AwsCloudWatchEventListener`) seems unnecessary.

If the target serialization format is Json, it would be beneficial to provide Jackson serialization info in the event types themselves, e.g.:

```java
  public record AfterRefreshTableEvent(
      @JsonProperty("catalog_name")
      String catalogName,
      @JsonProperty("table_identifier")
      TableIdentifier tableIdentifier)
      implements PolarisEvent {}
```

Or alternatively:

```java
  @JsonNaming(SnakeCaseStrategy.class)
  public record AfterRefreshTableEvent(
      String catalogName,
      TableIdentifier tableIdentifier)
      implements PolarisEvent {}
```

The above would be naturally serialized as:

```json
{
  "catalog_name" : "test_catalog",
  "table_identifier" : {
    "namespace" : [ "test_namespace" ],
    "name" : "test_table"
  }
}
```

Without the need for `PropertyMapEventListener`.

`PropertyMapEventListener` will also become problematic for larger payloads e.g. `TableMetadata` as the serialization logic is not well-defined (which fields to include? which case? which order?)

vchag:
@adutra I'm happy to take this up.

adutra:
Thanks @vchag, the issue is assigned to you!

vchag:
Thanks @adutra ! ... a few thoughts; curious for your take. 

1. Currently `PropertyMapEventListener` the name makes it sound like a generic event listener but it only overrides `onAfterRefreshTable`, which does look odd. If the intent was to only care about `AfterRefreshTableEvent` then I think we should rename `PropertyMapEventListener` to `AfterRefreshTableEventListener`, any thoughts? 

2. Properties parameter `HashMap` -> `Map`:  This becomes moot if we agree on #3. 

[...]

PR #2836, opened 2025-10-17
Introducing AfterRefreshTableEventListener and Mixins

vchag:
These changes were created to address concerns raised in Issues #2630

1. Renaming `PropertyMapEventListener` for clarity: 
The class `PropertyMapEventListener` was misleadingly named, suggesting a generic event listening capability when its only overridden method was onAfterRefreshTable. To ensure the name accurately reflects its single responsibility, we have renamed it to `AfterRefreshTableEventListener`. This precise naming clearly communicates its purpose: to only process the AfterRefreshTableEvent.

2. Streamlining Event-to-JSON Serialization:
The previous architecture involved a redundant, two-step conversion process: events were transformed into intermediate Maps by an abstract class, and only then was the resulting map serialized into JSON (as seen in `AwsCloudWatchEventListener`).
Since Jackson JSON is our chosen serialization format, we've removed this unnecessary intermediate mapping step. We introduced Jackson Mixins to provide native JSON serialization support directly on the event objects. This refactoring simplifies the serialization pipeline from:
`Event` -> `Map` -> `JSON`
                  to
`Event` -> `JSON`

### DIFF
--- a/site/hugo.yaml
+++ b/site/hugo.yaml
@@ -65,6 +65,7 @@ params:
   # at the very right in the navbar.
   active_releases:
     # Mention all active releases here, in semver descending order
+    - "1.2.0"
     - "1.1.0"
     - "1.0.1"
     - "1.0.0"


runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/ObjectMapperFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.polaris.service.events.IcebergRestCatalogEvents;
+import org.apache.polaris.service.events.PolarisEvent;
+
+public class ObjectMapperFactory {
+  private ObjectMapperFactory() {}
+
+  public static ObjectMapper create() {
```
adutra: You should not create an ObjectMapper like this, but rather use CDI.

1. If possible, let's use the main ObjectMapper provided by Quarkus. You can customize it by adding your customizations in the `org.apache.polaris.service.config.PolarisIcebergObjectMapperCustomizer` bean.
2. If it's not possible to reuse the main ObjectMapper (for example, because of incompatible configuration), then you should produce a new ObjectMapper bean.

I realize that the old version of `AwsCloudWatchEventListener` was also doing it wrong, so let's seize the opportunity to fix this.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/ObjectMapperFactory.java
```diff
@@ -0,0 +1,52 @@
...
+    ObjectMapper mapper =
+        new ObjectMapper()
+            .registerModule(new Jdk8Module()) // If you never serialize Optional, you can remove the
+            // .registerModule(new Jdk8Module()) line.
+            .registerModule(new JavaTimeModule())
+            .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
```
adutra: Can you explain why we need to disable these features? I would prefer to customize the mapper as little as possible.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergThirdPartyMixins.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.polaris.service.events.json.mixins;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Mixins for Iceberg classes we don't control, to keep JSON concise. The @JsonValue marks
+ * toString() as the value to serialize.
+ */
+public class IcebergThirdPartyMixins {
```
adutra: nit: rename to `IcebergMixins` and make it final.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergMixins.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Mixins for Iceberg classes we don't control, to keep JSON concise. The @JsonValue marks
+ * toString() as the value to serialize.
+ */
+public class IcebergThirdPartyMixins {
+  private IcebergThirdPartyMixins() {}
+
+  public abstract static class NamespaceMixin {
```
adutra: I wonder if this is the best thing to do when serializing namespaces and table identifiers. Granted, it's very convenient, but it would not correctly handle dots in namespace segments.

For example the following namespaces would produce the same string:

```java
Namespace.of("one.two", "three.four", "ns");
Namespace.of("one", "two", "three", "four", "ns");
```

I think it's wiser to serialize these two classes as objects, not scalars.

runtime/service/src/main/java/org/apache/polaris/service/events/json/serde/TableIdentifierToStringSerializer.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.apache.polaris.service.events.json.serde;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import java.io.IOException;
+import org.apache.iceberg.catalog.TableIdentifier;
+
+public class TableIdentifierToStringSerializer extends JsonSerializer<TableIdentifier> {
```
adutra: Why do we need this serializer? We already have `IcebergThirdPartyMixins.TableIdentifierMixin`.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -151,28 +158,44 @@ void shutdown() {
     }
   }
 
+  @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+  public record CloudWatchEvent(
+      String principal,
+      String realmId,
+      Collection<String> activatedRoles,
+      String eventType,
+      @JsonUnwrapped PolarisEvent event // flatten
```
adutra: Nice 👍

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/AfterRefreshTableEventListener.java
```diff
@@ -30,14 +29,12 @@
  * {{@code @link#transformAndSendEvent(HashMap)}} method to define how the event data should be
  * transformed into a JSON string, transmitted, and/or stored.
  */
-public abstract class PropertyMapEventListener implements PolarisEventListener {
-  protected abstract void transformAndSendEvent(HashMap<String, Object> properties);
+public abstract class AfterRefreshTableEventListener implements PolarisEventListener {
```
adutra: I'm sorry but just renaming this class doesn't make it any more usable.

In its current form, this class does not qualify as a general-purpose component worthy of being distributed with Polaris OSS.

In my opinion this class _must_ be removed.

BTW, `AwsCloudWatchEventListener` is also extremely problematic, as it only handles after-table-refresh events, _and nothing else_. In theory, `AwsCloudWatchEventListener` should either equally handle all 150+ event types, or make the event types to handle configurable.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/AfterRefreshTableEventListener.java
```diff
@@ -19,7 +19,6 @@
 
 package org.apache.polaris.service.events.jsonEventListener;
```
adutra: We also need to modify this package name: `jsonEventListener` does not comply with java naming conventions:

https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/AfterRefreshTableEventMixin.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.polaris.service.events.json.mixins;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.polaris.service.events.json.serde.TableIdentifierToStringSerializer;
+
+@JsonTypeName("AfterRefreshTableEvent")
+public abstract class AfterRefreshTableEventMixin {
```
adutra: I don't think you need this mixin, we already have specified `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)` in `PolarisEventBaseMixin`. That should be enough.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -151,28 +158,44 @@ void shutdown() {
...
+      ) {}
+
   @Override
-  protected void transformAndSendEvent(HashMap<String, Object> properties) {
-    properties.put("realm_id", callContext.getRealmContext().getRealmIdentifier());
-    properties.put("principal", securityContext.getUserPrincipal().getName());
-    properties.put(
-        "activated_roles", ((PolarisPrincipal) securityContext.getUserPrincipal()).getRoles());
-    // TODO: Add request ID when it is available
+  protected void transformAndSendEvent(IcebergRestCatalogEvents.AfterRefreshTableEvent event) {
```
adutra: As said above, this class _must_ handle all event types, or determine the event types to handle via configuration.

For example, we could introduce the following configuration option:

```properties
polaris.event-listener.aws-cloudwatch.event-types=\
  org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterRefreshTableEvent,\
  org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterCommitTableEvent
```

I would be in favor of doing this change in this PR since it this class has the same shortcomings as `PropertyMapEventListener`.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergThirdPartyMixins.java
vchag: Will rename.

runtime/service/src/main/java/org/apache/polaris/service/events/json/serde/TableIdentifierToStringSerializer.java
vchag: This is redundant. I'll go ahead and remove it.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/AfterRefreshTableEventListener.java
vchag: How about we re-name it `listener`?

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/AfterRefreshTableEventMixin.java
vchag: Actually, Yes. We don't need this mixin.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/ObjectMapperFactory.java
vchag: I agree with the CDI based approach. I found option #2 to be better feasible. 

Let's seize the opportunity!

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/ObjectMapperFactory.java
vchag: This was an oversight on my part.... will remove them.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergMixins.java
vchag: Great catch!

How about  we aligned the serialization behavior with how the Iceberg REST API represents namespaces and table identifiers?

`Namespace.of("one", "two", "three", "four", "ns");`
would result in:
  `{"namespace": ["one", "two", "three", "four", "ns"]}`
  
  `TableIdentifier.of(Namespace.of("one", "two", "three", "four", "ns"), "table")`
  would result in:
  `{"namespace": ["one", "two", "three", "four", "ns"], "name": "table"}`

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
vchag: Oh right. Lets introduce a new config for `aws-cloudwatch.event-types`.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergMixins.java
adutra: Indeed I think that's preferable. It's slightly verbose for simple cases, but it's the only representation that is compatible with all the corner cases.

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergMixins.java
```diff
@@ -19,24 +19,39 @@
...
+  /** Serializes Namespace as an object like: "namespace": ["sales", "north.america"] */
   public abstract static class NamespaceMixin {
+    @JsonProperty("namespace")
+    public abstract String[] levels();
+
     @Override
-    @JsonValue
-    public abstract String toString(); // serializes "namespace" as "db.sales"
+    @JsonIgnore
+    public abstract String toString();
```
adutra: Do we still need to declare this method?

runtime/service/src/main/java/org/apache/polaris/service/events/json/mixins/IcebergMixins.java
```diff
@@ -19,24 +19,39 @@
...
+    public abstract Namespace namespace();
+
+    @JsonProperty("name")
+    public abstract String name();
+
     @Override
-    @JsonValue
-    public abstract String toString(); // serializes "table_identifier" as "db.sales.orders"
+    @JsonIgnore
+    public abstract String toString();
```
adutra: ditto

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/AllEventsForwardingListener.java
```diff
@@ -0,0 +1,855 @@
...
+import org.apache.polaris.service.events.CatalogsServiceEvents;
+import org.apache.polaris.service.events.IcebergRestCatalogEvents;
+import org.apache.polaris.service.events.PolarisEvent;
+import org.apache.polaris.service.events.PrincipalRolesServiceEvents;
+import org.apache.polaris.service.events.PrincipalsServiceEvents;
+import org.apache.polaris.service.events.listeners.PolarisEventListener;
+
+/**
+ * Base class for event listeners that with to generically forward all {@link PolarisEvent
+ * PolarisEvents} to an external sinks
```
adutra:
```suggestion
 * PolarisEvents} to an external sink.
```

runtime/service/src/main/java/org/apache/polaris/service/config/PolarisAWSCloudWatchObjectMapperProducer.java
```diff
@@ -0,0 +1,82 @@
...
+import jakarta.inject.Singleton;
+import org.apache.polaris.service.events.PolarisEvent;
+import org.apache.polaris.service.events.json.mixins.IcebergMixins;
+import org.apache.polaris.service.events.json.mixins.PolarisEventBaseMixin;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** This mapper is isolated and used exclusively for CloudWatch event serializations */
+public class PolarisAWSCloudWatchObjectMapperProducer {
```
adutra: Why do we need a separate mapper? This one seems fairly identical to the default one, except of course for the mixins. But you could add the mixins to the default mapper instead.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -77,14 +78,28 @@ public class AwsCloudWatchEventListener extends AfterRefreshTableEventListener {
...
+    this.objectMapper = mapper;
+    this.allowedEventTypes =
+        config.eventTypes().stream()
+            .map(String::trim)
+            .filter(s -> !s.isEmpty())
+            .collect(Collectors.toUnmodifiableSet());
+  }
+
+  @Override
+  protected boolean shouldHandle(Object event) {
```
adutra: Nice, this is exactly what I had in mind 👍

runtime/service/src/main/java/org/apache/polaris/service/quarkus/events/jsonEventListener/aws/cloudwatch/QuarkusAwsCloudWatchConfiguration.java
```diff
@@ -96,4 +97,21 @@ public interface QuarkusAwsCloudWatchConfiguration extends AwsCloudWatchConfigur
...
+   * org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterCommitTableEvent
+   *
+   * <p>Configuration property: {@code polaris.event-listener.aws-cloudwatch.event-types}
+   *
+   * @return a set of event types
+   */
+  @WithName("event-types")
+  @WithDefault("org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterRefreshTableEvent")
+  @Override
+  Set<String> eventTypes();
```
adutra: I think this could be `Set<Class<? extends PolarisEvent>>`.

runtime/service/src/main/java/org/apache/polaris/service/quarkus/events/jsonEventListener/aws/cloudwatch/QuarkusAwsCloudWatchConfiguration.java
```diff
@@ -96,4 +97,21 @@ public interface QuarkusAwsCloudWatchConfiguration extends AwsCloudWatchConfigur
...
+   * polaris.event-listener.aws-cloudwatch.event-types=\
+   * org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterRefreshTableEvent,\
+   * org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterCommitTableEvent
+   *
+   * <p>Configuration property: {@code polaris.event-listener.aws-cloudwatch.event-types}
+   *
+   * @return a set of event types
+   */
+  @WithName("event-types")
+  @WithDefault("org.apache.polaris.service.events.IcebergRestCatalogEvents.AfterRefreshTableEvent")
```
adutra: That's not a good default. I'd say the default should be an empty set, and we should treat an empty set as "process all events".

WDYT? 

Also, please add the corresponding property to `application.properties`, that helps users:

https://github.com/apache/polaris/blob/0acbb64692e6506967d3fee6a4e2f35266ae8138/runtime/defaults/src/main/resources/application.properties#L142-L147

runtime/service/src/main/java/org/apache/polaris/service/config/PolarisAWSCloudWatchObjectMapperProducer.java
```diff
@@ -0,0 +1,82 @@
...
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(PolarisAWSCloudWatchObjectMapperProducer.class);
+
+  private final long maxBodySize;
+
+  @Inject
+  public PolarisAWSCloudWatchObjectMapperProducer(
+      @ConfigProperty(name = "polaris.aws.cloudwatch.max-body-size", defaultValue = "16M")
```
adutra: Where is this config property coming from? It doesn't start with the prefix used for other CloudWatch properties, which is `polaris.event-listener.aws-cloudwatch.`.

site/content/in-dev/unreleased/configuration.md
```diff
@@ -81,52 +81,51 @@ read-only mode, as Polaris only reads the configuration file once, at startup.
...
+| `polaris.config.rollback.compaction.on-conflicts.enabled`                              | `false`                          | When set to true Polaris will apply the deconfliction by rollbacking those REPLACE operations snapshots which have the property of `polaris.internal.rollback.compaction.on-conflict` in their snapshot summary set to `rollback`, to resolve conflicts at the server end.                                                                                    |
+| `polaris.event-listener.type`                                                          | `no-op`                          | Define the Polaris event listener type. Supported values are `no-op`, `aws-cloudwatch`.                                                                                                                                                                                                                                                                       |
```
adutra: BTW `persistence-in-memory-buffer` is also supported.

site/content/in-dev/unreleased/configuration.md
```diff
@@ -81,52 +81,51 @@ read-only mode, as Polaris only reads the configuration file once, at startup.
...
-| `polaris.event-listener.aws-cloudwatch.region`                                         | `us-east-1`                        | Define the AWS region for the CloudWatch event listener.                                                                                                                                                                                                                                                                                                      |
-| `polaris.event-listener.aws-cloudwatch.synchronous-mode`                               | `false`                            | Define whether log events are sent to CloudWatch synchronously. When set to true, events are sent synchronously which may impact performance but ensures immediate delivery. When false (default), events are sent asynchronously for better performance.                                                                                                     |
-
```
adutra, line 129: Nit: Why remove this line?

vchag:
Closing in favor of a new PR from a feature branch.
My fork’s main fell behind upstream and a fork sync/force-push unintentionally rewrote the PR head, which auto-closed this PR.
I'll recreate the changes on a proper feature branch, rebased on the latest upstream/main, and addressed prior feedback.

Thanks for the reviews so far.....carrying over all context to the new PR.

adutra:
@vchag what's the status here? Are you still able to work in this issue?

vchag:
@adutra Yes, I'm on it. Will release it within the next 4 - 8 hrs.

vchag:
Here's the new PR https://github.com/apache/polaris/pull/2962.

Issue #342, opened 2024-10-03
[FEATURE REQUEST] support for returning client region in the loadTable response for S3


munendrasn:
### Is your feature request related to a problem? Please describe.

AWS s3 by default fails when making cross region calls. Latest version of AWS s3 sdk support crossRegionAccess but Iceberg doesn't yet support this [functionality](https://github.com/apache/iceberg/pull/9804). 
Any cross region calls are failing 

### Describe the solution you'd like

It would be great to specify the client.region as part of loadTableResponse in `config` block along with Credentials. This is property defined by RESTCatalog openAPI [spec](https://github.com/apache/iceberg/blob/b1d38b3cac326da296b770c8ab9219bc4f791666/open-api/rest-catalog-open-api.yaml#L3127)

### Describe alternatives you've considered

Using the crossRegionAccess property. Maybe once Iceberg add support, it would be great to add this property to the spec too

munendrasn:
FYI @dennishuo

PR #2838, opened 2025-10-17
Remove credential flag from `StorageAccessProperty.CLIENT_REGION`

dimas-b:
`CLIENT_REGION` is not a credential value, which is in line with Iceberg's `VendedCredentialsProvider` code.

Cf. https://github.com/apache/iceberg/pull/11389

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
@@ -29,9 +29,9 @@ import org.apache.iceberg.gcp.GCPProperties;
  * storage.
  */
 public enum StorageAccessProperty {
-  AWS_KEY_ID(String.class, "s3.access-key-id", "the aws access key id"),
-  AWS_SECRET_KEY(String.class, "s3.secret-access-key", "the aws access key secret"),
-  AWS_TOKEN(String.class, "s3.session-token", "the aws scoped access token"),
+  AWS_KEY_ID(String.class, "s3.access-key-id", "the aws access key id", true),
+  AWS_SECRET_KEY(String.class, "s3.secret-access-key", "the aws access key secret", true),
+  AWS_TOKEN(String.class, "s3.session-token", "the aws scoped access token", true),
   AWS_SESSION_TOKEN_EXPIRES_AT_MS(
       String.class,
       "s3.session-token-expires-at-ms",
@@ -42,7 +42,10 @@ public enum StorageAccessProperty {
   AWS_PATH_STYLE_ACCESS(
       Boolean.class, "s3.path-style-access", "whether to use S3 path style access", false),
   CLIENT_REGION(
-      String.class, "client.region", "region to configure client for making requests to AWS"),
+      String.class,
+      "client.region",
+      "region to configure client for making requests to AWS",
+      false),
   AWS_REFRESH_CREDENTIALS_ENDPOINT(
       String.class,
       AwsClientProperties.REFRESH_CREDENTIALS_ENDPOINT,
@@ -50,7 +53,7 @@ public enum StorageAccessProperty {
       false,
       false),
 
-  GCS_ACCESS_TOKEN(String.class, "gcs.oauth2.token", "the gcs scoped access token"),
+  GCS_ACCESS_TOKEN(String.class, "gcs.oauth2.token", "the gcs scoped access token", true),
   GCS_ACCESS_TOKEN_EXPIRES_AT(
       String.class,
       "gcs.oauth2.token-expires-at",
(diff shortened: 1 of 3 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
```diff
@@ -93,15 +96,6 @@ public enum StorageAccessProperty {
   private final boolean isCredential;
   private final boolean isExpirationTimestamp;
 
-  /*
-  s3.access-key-id`: id for for credentials that provide access to the data in S3
-           - `s3.secret-access-key`: secret for credentials that provide access to data in S3
-           - `s3.session-token
-   */
-  StorageAccessProperty(Class valueType, String propertyName, String description) {
```
HonahX, line 101: Is the idea here to make the enum declaration more explicit about the `isCredential` setting?

HonahX (commented):
Thanks for the PR @dimas-b. To confirm the context, the `s3.region` was included in credentials for https://github.com/apache/polaris/issues/342 because iceberg by that time does not support s3 cross-region access yet. That was later added in https://github.com/apache/iceberg/commit/3d9fc1dee1228e742e22234369498ee16b19f5a2 and released with iceberg 1.7.0.

Would this be a breaking change for users since the `s3.cross-region-access-enabled` is still default to be false? I assume users who are relying on this information  in vended credential to avoid cross-region call will encounter failure after this PR?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessProperty.java
dimas-b: Yes

dimas-b:
@HonahX : `s3.region` is still [included](https://github.com/apache/polaris/blob/f97c5eb50016489129575aab62d5efb3efb7552e/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java#L124) into `AccessConfig` and by extension into LoadTable responses, but as an [ordinary property](https://github.com/apache/polaris/blob/6b957ec14516ab764783833c4e5c40ba58fe3864/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L841).

Is that sufficient for Iceberg clients? What's the best way to test it?

For background: the reason for this PR is that the "nice error message" from #2711 does not happen if the region is set in storage config (but no credentials are vended).

HonahX:
@dimas-b Thanks for the pointer and context. Iceberg's `LoadTableResponse` have different fields holding normal config and credentials:
https://github.com/apache/iceberg/blob/03f2af8ab465d4a5238753be206bcf366130f9a0/open-api/rest-catalog-open-api.yaml#L3321-L3328



The RestCatalog client only use things in the `storage-credentials` fields to [initialize the FileIO](https://github.com/apache/iceberg/blob/03f2af8ab465d4a5238753be206bcf366130f9a0/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L457), where the vended credential is used. 

If `client.region` is in extra property, it will [not be added](builder: https://github.com/apache/iceberg/blob/03f2af8ab465d4a5238753be206bcf366130f9a0/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java#L117-L130) to credential field and thus the RestCatalog client cannot pick up that config when initializing the FileIO.

HonahX:
> For background: the reason for this PR is that the "nice error message" from https://github.com/apache/polaris/pull/2711 does not happen if the region is set in storage config (but no credentials are vended).

Would it be easier if consider the credential being "empty" if only CLIENT_REGION is there?

dimas-b:
@HonahX : I tested with Spark 3.5 + Iceberg 1.6.1 and AWS with the code in this PR... all seems to work well... WDYT?

Note: Spark did _not_ have any AWS FileIO config. It got everything from Polaris.

Here's a sample load table response for reference:
```
{
  "metadata-location": "s3://***/pol/ns/t1/metadata/00000-c8eeb778-d0ca-42ff-b48c-743f16751b7c.metadata.json",
  "metadata": {
    "format-version": 2,
    "table-uuid": "5403e77b-dbd4-425a-ba20-969bf448462b",
    "location": "s3://***/pol/ns/t1",
[...]
  },
  "config": {
    "s3.access-key-id": "***",
    "s3.secret-access-key": "****",
    "s3.session-token": "*****",
    "client.refresh-credentials-endpoint": "v1/polaris/namespaces/ns/tables/t1/credentials",
    "expiration-time": "1760987329000",
    "s3.session-token-expires-at-ms": "1760987329000",
    "client.region": "us-west-2"
  },
  "storage-credentials": [
    {
      "prefix": "s3://***/pol/ns/t1",
      "config": {
        "s3.access-key-id": "***",
        "s3.secret-access-key": "****",
        "s3.session-token": "*****",
        "expiration-time": "1760987329000",
        "s3.session-token-expires-at-ms": "1760987329000"
      }
    }
  ]
}
```

HonahX (approved):
@dimas-b Thanks for the test and examples. I also did one with Spark 3.5 + Iceberg 1.9.1. Apparently I missed some details when reading through the code. You are right! The `client.region` will still take effect in the `config` field. LGTM!

PR #2847, opened 2025-10-21
Add test for TracingFilter

adutra:
`TracingFilter` was lacking a proper test. 

After discussing with @dimas-b on [#2757](https://github.com/apache/polaris/pull/2757#discussion_r2446238692) I decided to create a test to illustrate two things:

- How the filter adds span attributes for the correlation ID and for the realm identifier;
- How the filter honors any parent trace propagated via W3C Trace Context propagation.

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/tracing/TracingFilterTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/tracing/TracingFilterTest.java
@@ -85,7 +85,7 @@ public class TracingFilterTest {
         // W3C headers
         .header("traceparent", traceparent)
         .header("tracestate", tracestate)
-        // Polaris correlation ID
+        // Polaris request ID
         .header("X-Request-ID", "12345")
         .when()
         .post()


runtime/service/src/test/java/org/apache/polaris/service/tracing/TracingFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+        .containsEntry(stringKey(TracingFilter.REQUEST_ID_ATTRIBUTE), "12345");
+
+    SpanContext parent = span.getParentSpanContext();
+    assertThat(parent.getTraceId()).isEqualTo(traceId);
+    assertThat(parent.getSpanId()).isEqualTo(spanId);
+    assertThat(parent.isRemote()).isTrue();
+    assertThat(parent.getTraceFlags().asByte()).isEqualTo((byte) 1);
+    assertThat(parent.getTraceState().asMap())
+        .containsEntry("rojo", "00f067aa0ba902b7")
+        .containsEntry("congo", "t61rcWkgMzE");
```
flyrain: Nit: using variable or constant here?  `t61rcWkgMzE` -> congoValue?
```suggestion
        .containsEntry("rojo", spanId)
        .containsEntry("congo", "t61rcWkgMzE");
```

flyrain (approved):
LGTM. Thanks @adutra !

runtime/service/src/test/java/org/apache/polaris/service/tracing/TracingFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+        await()
+            .atMost(Duration.ofSeconds(30))
+            .until(inMemorySpanExporter::getFinishedSpanItems, sp -> !sp.isEmpty());
+
+    SpanData span = spans.getFirst();
+
+    Map<AttributeKey<?>, Object> attributes = span.getAttributes().asMap();
+    assertThat(attributes)
+        .containsEntry(stringKey(TracingFilter.REALM_ID_ATTRIBUTE), "POLARIS")
+        .containsEntry(stringKey(TracingFilter.REQUEST_ID_ATTRIBUTE), "12345");
```
dimas-b, line 106: side note: my thinking was to reuse request ID from OTel (e.g. trace ID + span ID). Since we're moving towards "correlation", I think it makes sense to integrate with the existing OTel standard. In that case it should not be necessary to add `REQUEST_ID_ATTRIBUTE` to the span, but someone inspecting the response headers would be able to correlate them to the trace.

If OTel is present, what is the value of Polaris accepting / generating yet another request ID?

Are we considering the Polaris request ID as a first class feature of the protocol (REST API flowing into Events) as opposed to a "helper" or "observability" feature?

If OTel is not present, we can certainly generate a fresh request ID.

dimas-b (dismissed):
The test code LGTM. My comment is more for discussion... I wonder if it might be preferable to move that to the `dev` ML 🤔

runtime/service/src/test/java/org/apache/polaris/service/tracing/TracingFilterTest.java
adutra: OK, let's move this conversation to the dev ML:

https://lists.apache.org/thread/bb1qyxjt827t3tomv2xp0s1kovxjsp94

adutra:
@flyrain or @dimas-b can I get another review please?

dimas-b (approved):

PR #2850, opened 2025-10-21
Move TestPolarisEventListener to test fixtures

adutra:


dimas-b (approved):

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -222,18 +220,6 @@ public ProductionReadinessCheck checkRealmResolver(Config config, RealmContextRe
...
-  public ProductionReadinessCheck checkPolarisEventListener(
-      PolarisEventListener polarisEventListener) {
-    if (polarisEventListener instanceof TestPolarisEventListener) {
-      return ProductionReadinessCheck.of(
-          Error.of(
-              "TestPolarisEventListener is intended for tests only.",
-              "polaris.event-listener.type"));
-    }
-    return ProductionReadinessCheck.OK;
-  }
```
singhpk234, line 235: is this production readiness check not honoured any more ?

adutra: It's not necessary anymore since the class has been moved to a test folder and hence is not present anymore in Polaris production code. IOW: you cannot select this listener anymore when deploying Polaris, only when testing it.

PR #2855, opened 2025-10-21
Add --no-sts to CLI

dimas-b:
Following up on #2672, add new `--no-sts` option to CLI to allow configuring `stsUnavailable` in `AwsStorageConfigInfo`

### DIFF
--- a/client/python/cli/constants.py
+++ b/client/python/cli/constants.py
@@ -253,7 +253,7 @@ class Hints:
                 "(Only for S3) The STS endpoint to use when connecting to STS"
             )
             STS_UNAVAILABLE = (
-                "(Only for S3) Indicates that STS is not available"
+                "(Only for S3) Indicates that Polaris should not use STS (e.g. if STS is not available)"
             )
             PATH_STYLE_ACCESS = "(Only for S3) Whether to use path-style-access for S3"
 


snazy (approved):

client/python/cli/constants.py
```diff
@@ -251,6 +252,9 @@ class Create:
             STS_ENDPOINT = (
                 "(Only for S3) The STS endpoint to use when connecting to STS"
             )
+            STS_UNAVAILABLE = (
+                "(Only for S3) Indicates that STS is not available"
```
singhpk234:
```suggestion
                "(Only for S3 compatible object stores) Indicates that STS is not available"
```

singhpk234 (approved):

client/python/cli/constants.py
dimas-b: It could be used with AWS too. How about `(Only for S3) Indicates that Polaris should not use STS (e.g. if STS is not available)`?

dimas-b: Message updated. PTAL.

singhpk234: makes sense, thanks @dimas-b !

singhpk234 (approved):

PR #2863, opened 2025-10-22
CLI: Remove the trailing comma

flyrain:
The trailing comma makes the `parameter` a tuple, rather than a dict.
```
> cd client/python
> ../../polaris-venv/bin/python -m pytest test/test_cli_parsing.py::TestCliParsing::test_policies_attach_parameters_parsed_to_dict -q

    def test_policies_attach_parameters_parsed_to_dict(self):
        options = Parser.parse([
            'policies', 'attach', 'policy-name',
            '--catalog', 'cat',
            '--attachment-type', 'catalog',
            '--parameters', 'key=value',
        ])
>       self.assertIsInstance(command.parameters, dict)
E       AssertionError: ({'key': 'value'},) is not an instance of <class 'dict'>

test/test_cli_parsing.py:629: AssertionError

FAILED test/test_cli_parsing.py::TestCliParsing::test_policies_attach_parameters_parsed_to_dict - AssertionError: ({'key': 'value'},) is not an instance of <class 'dict'>
1 failed in 0.47s

```

### DIFF
--- a/client/python/cli/command/__init__.py
+++ b/client/python/cli/command/__init__.py
@@ -40,7 +40,7 @@ class Command(ABC):
         set_properties = Parser.parse_properties(options_get(Arguments.SET_PROPERTY))
         remove_properties = options_get(Arguments.REMOVE_PROPERTY)
         catalog_client_scopes = options_get(Arguments.CATALOG_CLIENT_SCOPE)
-        parameters = Parser.parse_properties(options_get(Arguments.PARAMETERS)),
+        parameters = Parser.parse_properties(options_get(Arguments.PARAMETERS))
 
         command = None
         if options.command == Commands.CATALOGS:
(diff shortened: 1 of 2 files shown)


client/python/cli/command/__init__.py
```diff
@@ -40,7 +40,7 @@ def options_get(key, f=lambda x: x):
         set_properties = Parser.parse_properties(options_get(Arguments.SET_PROPERTY))
         remove_properties = options_get(Arguments.REMOVE_PROPERTY)
         catalog_client_scopes = options_get(Arguments.CATALOG_CLIENT_SCOPE)
-        parameters = Parser.parse_properties(options_get(Arguments.PARAMETERS)),
+        parameters = Parser.parse_properties(options_get(Arguments.PARAMETERS))
```
HonahX, line 43: I am surprised that this is not caught by existing test and regtest. Thanks for adding the test and fix it!

singhpk234 (approved):

client/python/cli/command/__init__.py
flyrain: `parameters` may never be used.

flyrain:
Thanks @HonahX  and @singhpk234 for the review.

PR #2883, opened 2025-10-24
Rename AccessConfig and AccessConfigProvider for clarity

CodingBangboo:

As a follow up of https://github.com/apache/polaris/pull/2736#discussion_r2393425836, we would like to rename the `AccessConfig` and `AccessConfigProvider` to `StorageAccessConfig` and `StorageAccessConfigProvider` respectively, including renaming `AccessConfigProvider`'s method `getAccessConfig` to `getStorageAccessConfig`. This renaming PR adds clarity to differentiating between two types of external services: storage and catalog accesses.


HonahX (dismissed):
@CodingBangboo Thanks for the contribution! The rename LGTM as it would help avoid confusion when we consolidate and introduce CatalogAccessConfig in the future, per comment.

Would love to hear others' thoughts on this!

cc: @dimas-b @XJDKC

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/ScopedCredentialsResult.java
```diff
@@ -37,20 +37,20 @@ public class ScopedCredentialsResult extends BaseResult {
...
     super(errorCode, extraInformation);
-    this.accessConfig = null;
+    this.storageAccessConfig = null;
   }
 
   /**
    * Constructor for success
    *
-   * @param accessConfig credentials
+   * @param storageAccessConfig credentials
```
dimas-b: nit: this javadoc is no longer accurate. Since the class name is pretty much self-describing now, I do not see a reason to have javadoc for this parameter (can we put anything there that is not covered by the class?).

dimas-b (approved):
This rename looks reasonable to me 👍 Thanks for your contribution, @CodingBangboo !

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/ScopedCredentialsResult.java
CodingBangboo: Good point. I think we can keep it there for consistency with the javadoc in other `BaseResult` classes. I also added a tiny bit to indicate that it is the result from a successful `getSubscopedCredsForEntity()`.

HonahX (dismissed):

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/ScopedCredentialsResult.java
```diff
@@ -37,20 +37,20 @@ public class ScopedCredentialsResult extends BaseResult {
...
     super(errorCode, extraInformation);
-    this.accessConfig = null;
+    this.storageAccessConfig = null;
   }
 
   /**
    * Constructor for success
    *
-   * @param accessConfig credentials
+   * @param storageAccessConfig credentials generated by a successful getSubscopedCredsForEntity()
```
dimas-b: As far as this class is concerned, why does it matter that `storageAccessConfig` is `generated by a successful getSubscopedCredsForEntity()`?

XJDKC (approved):
LGTM!

polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/ScopedCredentialsResult.java
dimas-b: @CodingBangboo @HonahX : I know this is a minor point, but I'd prefer to avoid false requirements in javadoc... WDYT about this one?

To be clear: I propose to remove this javadoc line completely. I believe the java doc on the parameter type is sufficient in this case.

CodingBangboo: Removed.

HonahX: Sorry for being late here. I am ok with either approach

PR #2887, opened 2025-10-24
Added interface for reporting metrics

cccs-cat001:
I'm adding an interface to be able to code for the `/metrics` endpoint in the Iceberg REST API Spec. 

The metrics endpoint in Polaris is currently a no-op endpoint. In our organization we use it for auditing who is reading/writing tables, as this endpoint is called after every read/write with Iceberg. I added a default no-op implementation so that the behaviour remains the same, and I added a logging implementation to show how it could be used. 

it adds an optional configuration for telling the metrics endpoint how it should behave. the default behaviour is no change, however if the user sets the `polaris.metrics.reporting.type` to `logging` it will log the MetricsReport to INFO. There is also the ability to implement the interface for your own needs if necessary. 

Tests were added. it was also tested in a staging environment.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
@@ -20,12 +20,14 @@ package org.apache.polaris.service.reporting;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.smallrye.common.annotation.Identifier;
+import jakarta.enterprise.context.ApplicationScoped;
 import org.apache.commons.lang3.function.TriConsumer;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.metrics.MetricsReport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@ApplicationScoped
 @Identifier("default")
 public class DefaultMetricsReporter implements PolarisMetricsReporter {
   private static final Logger LOGGER = LoggerFactory.getLogger(DefaultMetricsReporter.class);


runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
```diff
@@ -0,0 +1,22 @@
...
+
+@RequestScoped
+@Identifier("logging")
+public class LoggingMetricsReporter implements MetricsReporter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMetricsReporter.class);
+
+  @Override
+  public void reportMetric(
+      String prefix, String namespace, String table, ReportMetricsRequest reportMetricsRequest) {
+    LOGGER.info(prefix);
```
adutra: It would be better to consolidate all information into a single message. This is because the four lines are not guaranteed to appear consecutively in the logs.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,19 @@
...
+
+@RequestScoped
+@Identifier("default")
+public class DefaultMetricsReporter implements MetricsReporter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultMetricsReporter.class);
+
+  @Override
+  public void reportMetric(
+      String prefix, String namespace, String table, ReportMetricsRequest reportMetricsRequest) {
+    // Do Nothing
```
adutra: 1. Would it make sense to define and maintain a set of Micrometer metrics and increment them here?
2. Would it make sense to fire a `PolarisEvent` here?

(I don't really know what an IRC server is supposed to do with these scan or commit metrics 😅)

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReporter.java
```diff
@@ -0,0 +1,8 @@
+package org.apache.polaris.service.reporting;
+
+import org.apache.iceberg.rest.requests.ReportMetricsRequest;
+
+public interface MetricsReporter {
```
adutra: Iceberg already has a `MetricsReporter` and a `LoggingMetricsReporter`. I suppose you created new ones because you wanted to pass the warehouse and namespace names as well? In this case would it be possible to _extend_ Iceberg's `MetricsReporter`?

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReporter.java
```diff
@@ -0,0 +1,8 @@
...
+  public void reportMetric(
+      String prefix, String namespace, String table, ReportMetricsRequest reportMetricsRequest);
```
adutra: 1. I would rather rename `prefix` to `warehouse` or `catalogName` since `prefix` is a REST-specific notion. 
    1. But you should to call `prefixParser.prefixToCatalogName(realmContext, prefix)` to get the catalog name from the prefix.
2. For a richer API, I would use `TableIdentifier` rather than 2 strings (namespace + table). It's very easy to get a `TableIdentifier`:
    ```java
    Namespace ns = decodeNamespace(namespace);
    ```

runtime/service/src/test/java/org/apache/polaris/service/reporting/MetricsReportingConfigurationTest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+public class MetricsReportingConfigurationTest {
+  @Inject MetricsReportingConfiguration metricsConfig;
+
+  @Test
+  @TestProfile(MetricsReportingConfiguration.DefaultProfile.class)
+  void testDefault() {
```
adutra: These tests are rather useless as they are basically testing SmallRye Config.

What would be more meaningful is a real test that invokes the IRC metrics endpoint and verifies that metrics were collected.

runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
```diff
@@ -0,0 +1,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RequestScoped
+@Identifier("logging")
+public class LoggingMetricsReporter implements MetricsReporter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMetricsReporter.class);
+
+  @Override
+  public void reportMetric(
```
adutra: I'm not sure we need a distinct logging reporter. The default reporter could log the metrics, then users could simply adjust their log levels and decide whether they want the metrics to appear in the logs or not (by default, they wouldn't be logged). WDYT?

adutra (commented):
Thanks a lot for this contribution @cccs-cat001! I left a few comments mostly around API design.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReporter.java
cccs-cat001: So that was my original plan! But the iceberg `MetricsReporter` is intended for client-side operations, so implementing it won't suffice for the `/metrics` endpoint reporting. 

It would be ideal to change the name of it so that it doesn't get confused, but I wasn't sure of a better name 😅

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
cccs-cat001: I'm not sure what those are 😅 my intentions were to have the default behaviour remain the same, and adding that may change the default behaviour

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -405,4 +407,11 @@ public PolarisCredentialManager polarisCredentialManager(
   public void closeTaskExecutor(@Disposes @Identifier("task-executor") ManagedExecutor executor) {
     executor.close();
   }
+
+  @Produces
+  @RequestScoped
+  public MetricsReporter metricsReporter(
+      MetricsReportingConfiguration config, @Any Instance<MetricsReporter> reporters) {
+    return reporters.select(Identifier.Literal.of(config.type())).get();
```
dimas-b, line 415: `config.type()` is fixed in runtime... Can this bean be `@ApplicationScoped`?

dimas-b (commented):
Thanks for your interest in Polaris and contributing, @cccs-cat001 !

I'm not sure how this PR relates to the `/metrics` endpoint in the Iceberg REST API, though 🤔  Could you clarify, please?

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
adutra: 😄 - no worries, we can look into that later.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReporter.java
adutra: I see. Would something like this suit your needs?

```java
public interface PolarisMetricsReporter extends MetricsReporter {
  void report(String catalog, TableIdentifier table, MetricsReport report);
}
```

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReporter.java
adutra: I forgot: it's probably better to pass `MetricsReport` here, not `ReportMetricsRequest`.

adutra:
> I'm not sure how this PR relates to the `/metrics` endpoint in the Iceberg REST API, though 🤔 Could you clarify, please?

If that can help: I was also puzzled initially, but we are actually talking about this IRC endpoint:

https://github.com/apache/polaris/blob/f7b5646308b0477f737e9e0615c4649b0a497805/spec/iceberg-rest-catalog-open-api.yaml#L1288

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReporter.java
cccs-cat001: the iceberg implementation is definitely not what we want to extend/implement here - https://iceberg.apache.org/docs/1.10.0/docs/metrics-reporting

it's a client side thing. but I can change the name of this interface to PolarisMetricsReporter to have that separation.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
singhpk234: My recommendation was to store them in persistence, so that later we can serve them later, I recommend the same to @pingtimeout https://lists.apache.org/thread/koogjvb73hbl9ospmm7owsdx8oh67wk1

runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
singhpk234: Agree, Also i would recommend adding table uuid in the logs

runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
cccs-cat001: As in use LOGGER.debug to print the logs? I'm not so sure, someone may want these logged without having debug logs enabled.

runtime/service/src/test/java/org/apache/polaris/service/reporting/MetricsReportingConfigurationTest.java
cccs-cat001: hmm is there a way to verify that something was logged?

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
adutra: I guess we would need to purge that table frequently. Not sure if it's a good default to store in persistence without providing the purge functionality. Could that be an alternate CDI bean?

runtime/service/src/test/java/org/apache/polaris/service/reporting/MetricsReportingConfigurationTest.java
adutra: Yes, but it's a bit convoluted unfortunately. Have a look at `ExceptionMapperTest`: https://github.com/apache/polaris/blob/c2e932b7064e7ae5551e192c8dd273de2da9fd0d/runtime/service/src/test/java/org/apache/polaris/service/exception/ExceptionMapperTest.java#L52.

runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
adutra: No I would recommend logging at INFO, but set this logger by default to OFF. Here is how to do it:

https://github.com/apache/polaris/blob/519e127c5f01352d3efa305535a594c734b941b0/runtime/defaults/src/main/resources/application.properties#L73

runtime/defaults/src/main/resources/application.properties
```diff
@@ -209,6 +209,9 @@ polaris.oidc.principal-roles-mapper.type=default
 # Polaris Credential Manager Config
 polaris.credential-manager.type=default
 
+# Configuration for the behaviour of the metrics endpoint
+polaris.metrics.reporting.type=default
```
adutra: Can we rename this to `polaris.iceberg-metrics.reporting.type`? I'm afraid users will be confuse this feature with "standard" server metrics (which are configured using `polaris.metrics.*`, cf line 153).

runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
```diff
@@ -0,0 +1,44 @@
...
+
+@RequestScoped
+@Identifier("logging")
+public class LoggingMetricsReporter implements PolarisMetricsReporter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMetricsReporter.class);
+
+  @Override
+  public void reportMetric(
+      String warehouse, TableIdentifier table, MetricsReport metricsReport) {
+    StringBuilder reportString = new StringBuilder();
```
adutra: 1. I don't think you need this. The `toString()` methods for `ScanReport` and `CommitReport` are generated by Java Immutables and contain all the information already. Same for `TableIdentifier` the `toString()` method is well defined. 
2. The log pattern is malformed, there is one dangling placeholder.

Just this should be enough:

```java
LOGGER.info("{}.{}: {}", warehouse, table, metricsReport);
```

runtime/service/src/test/java/org/apache/polaris/service/reporting/MetricsReportingConfigurationTest.java
adutra: Another way to test logger is to modify your reporter implementation like below:

```java
@RequestScoped
@Identifier("logging")
public class LoggingMetricsReporter implements PolarisMetricsReporter {
  private static final Logger LOGGER = LoggerFactory.getLogger(LoggingMetricsReporter.class);

  // A consumer that can be mocked in tests; production code uses the default consumer
  // which logs to the slf4j logger.
  private final TriConsumer<String, TableIdentifier, MetricsReport> reportConsumer;

  public LoggingMetricsReporter() {
    this(
        (warehouse, table, metricsReport) ->
            LOGGER.info("{}.{}: {}", warehouse, table, metricsReport));
  }

  @VisibleForTesting
  public LoggingMetricsReporter(
      TriConsumer<String, TableIdentifier, MetricsReport> reportConsumer) {
    this.reportConsumer = reportConsumer;
  }

  @Override
  public void reportMetric(String warehouse, TableIdentifier table, MetricsReport metricsReport) {
    reportConsumer.accept(warehouse, table, metricsReport);
  }
}
```

adutra: If that's not too much to ask, you could provide two tests:

1. A simple unit test that tests the reporter in isolation, testing that messages are generated in the logs, using any of the techniques above.
2. A `@QuarkusTest` using a mock reporter (you can inject one in your test using `@InjectMock`), testing that metrics received via the endpoint are correctly passed to the reporter.

runtime/service/src/main/java/org/apache/polaris/service/reporting/LoggingMetricsReporter.java
cccs-cat001: okay this is fine. I just made 2 implementations to ensure that I was going down the right path anyways :P 

I've dropped down to one implementation with the configurable logging.

runtime/service/src/test/java/org/apache/polaris/service/reporting/MetricsReportingConfigurationTest.java
cccs-cat001: hoping those tests I added are what you were looking for 😅

runtime/defaults/src/main/resources/application.properties
```diff
@@ -209,6 +209,11 @@ polaris.oidc.principal-roles-mapper.type=default
 # Polaris Credential Manager Config
 polaris.credential-manager.type=default
 
+# Configuration for the behaviour of the metrics endpoint
+polaris.iceberg-metrics.reporting.type=default
+# Set to INFO if you want to see iceberg metric reports printed to console
+quarkus.log.category."polaris.service.reporting".level=WARN
```
adutra:
```suggestion
quarkus.log.category."org.apache.polaris.service.reporting".level=OFF
```

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.polaris.service.reporting;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.smallrye.common.annotation.Identifier;
+import io.smallrye.mutiny.tuples.Functions.TriConsumer;
```
adutra:
```suggestion
import org.apache.commons.lang3.function.TriConsumer;
```

There is nothing wrong per se with Mutiny's `TriConsumer`, but the one from Commons Lang is more general-purpose.

runtime/service/src/test/java/org/apache/polaris/service/reporting/DefaultMetricsReporterTest.java
```diff
@@ -0,0 +1,42 @@
...
+import io.quarkus.test.junit.TestProfile;
+import io.smallrye.mutiny.tuples.Functions.TriConsumer;
+import java.util.Map;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.metrics.MetricsReport;
+import org.apache.polaris.service.catalog.Profiles;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@TestProfile(DefaultMetricsReporterTest.Profile.class)
```
adutra: This is just a unit test, you don't need to start the whole application.
```suggestion
```

runtime/service/src/test/java/org/apache/polaris/service/reporting/DefaultMetricsReporterTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public static class Profile extends Profiles.DefaultProfile {
+    @Override
+    public Map<String, String> getConfigOverrides() {
+      return ImmutableMap.<String, String>builder()
+          .putAll(super.getConfigOverrides())
+          .put("quarkus.log.category.\"polaris.service.reporting\".level", "INFO")
+          .build();
+    }
+  }
```
adutra:
```suggestion
```

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,35 @@
...
+  private final TriConsumer<String, TableIdentifier, MetricsReport> reportConsumer;
+
+  public DefaultMetricsReporter() {
+    this(
+        (warehouse, table, metricsReport) ->
+            LOGGER.info("{}.{}: {}", warehouse, table, metricsReport));
+  }
+
+  @VisibleForTesting
+  public DefaultMetricsReporter(
```
adutra:
```suggestion
 DefaultMetricsReporter(
```

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
```diff
@@ -882,6 +882,27 @@ public void testCreateAndLoadTableWithReturnedEtag() {
...
+                    + "\",\"timestamp\":"
+                    + System.currentTimeMillis()
+                    + ",\"filesScanned\":1,\"rowsScanned\":100,\"metrics\":{\"bytesRead\":1024}}")
+            .build();
+    Invocation.Builder metricEndpoint =
+        catalogApi.request(
+            "v1/{cat}/namespaces/ns1/tables/tbl1/metrics", Map.of("cat", currentCatalogName));
+    try (Response response = metricEndpoint.post(Entity.json(payload))) {
+      assertThat(response).returns(Response.Status.NO_CONTENT.getStatusCode(), Response::getStatus);
+    }
```
adutra:
```suggestion
    ScanReport scanReport =
        ImmutableScanReport.builder()
            .tableName("tbl1")
            .schemaId(4)
            .addProjectedFieldIds(1, 2, 3)
            .addProjectedFieldNames("c1", "c2", "c3")
            .snapshotId(23L)
            .filter(Expressions.alwaysTrue())
            .scanMetrics(ScanMetricsResult.fromScanMetrics(ScanMetrics.noop()))
            .build();
    Invocation.Builder metricEndpoint =
        catalogApi.request(
            "v1/{cat}/namespaces/ns1/tables/tbl1/metrics", Map.of("cat", currentCatalogName));
    try (Response response = metricEndpoint.post(Entity.json(ReportMetricsRequest.of(scanReport)))) {
      assertThat(response).returns(Response.Status.NO_CONTENT.getStatusCode(), Response::getStatus);
    }
```

cccs-cat001:
I'm staring at the logs of the two failed checks and I'm not really sure what the cause of the errors are. Do you have any advice? 😅

adutra:
> I'm staring at the logs of the two failed checks and I'm not really sure what the cause of the errors are. Do you have any advice? 😅

You have 4 files without license header:

```
runtime/service/src/test/java/org/apache/polaris/service/reporting/DefaultMetricsReporterTest.java
runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsReportingConfiguration.java
runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```

The other error occurred while pulling a Docker image. It's a transient error.

adutra:
@cccs-cat001 One last thing and we are good to go: do you mind adding this blurb to the CHANGELOG.md file?

```markdown
### Highlights

- Support for [Iceberg Metrics Reporting] has been introduced in Polaris. Out of the box, metrics can
  be printed to the logs by setting the `org.apache.polaris.service.reporting` logger level to `INFO` (it's
  set to `OFF` by default). Custom reporters can be implemented and configured to send metrics to
  external systems for further analysis and monitoring.

[Iceberg Metrics Reporting]: https://iceberg.apache.org/docs/latest/metrics-reporting/
```

adutra (dismissed):
Thank you @cccs-cat001 for your contribution! LGTM.

I will leave the PR open for a few more hours in case others want to review.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -209,6 +209,11 @@ polaris.oidc.principal-roles-mapper.type=default
 # Polaris Credential Manager Config
 polaris.credential-manager.type=default
 
+# Configuration for the behaviour of the metrics endpoint
+polaris.iceberg-metrics.reporting.type=default
+# Set to INFO if you want to see iceberg metric reports printed to console
```
dimas-b: nit: it's not necessary "console" - the level will affect all log targets, AFAIK.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -755,7 +760,16 @@ public Response reportMetrics(
...
       SecurityContext securityContext) {
-    return Response.status(Response.Status.NO_CONTENT).build();
+
+    Namespace ns = decodeNamespace(namespace);
+    TableIdentifier tableIdentifier = TableIdentifier.of(ns, RESTUtil.decodeString(table));
+    return withCatalog(
+        securityContext,
+        prefix,
+        catalog -> {
+          catalog.reportMetric(prefix, tableIdentifier, reportMetricsRequest);
```
dimas-b: Why not call `metricsReporter.reportMetric()` from this class?

cccs-cat001: I was following the pattern of the other functions where they passed the call along to the IcebergCatalogHandler

dimas-b: Yes, that pattern exists for "catalog" operations... however, in this case the "catalog" does not really handle anything in the metrics reports. I do not really see a need to for an extra indirection layer (i.e. `catalog.reportMetric(...)`). It seems quite possible to call the `metricsReporter` directly from `IcebergCatalogAdapter`. The benefit is avoiding non-functional changes in `catalog`. WDYT?

dimas-b: @adutra : what 's your take on this?

adutra: I agree it would be slightly better.

cccs-cat001: oh yeah I see, you're totally right! I'll clean that up

dimas-b: thx :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -755,6 +759,10 @@ public Response reportMetrics(
       ReportMetricsRequest reportMetricsRequest,
       RealmContext realmContext,
       SecurityContext securityContext) {
+
+    Namespace ns = decodeNamespace(namespace);
+    TableIdentifier tableIdentifier = TableIdentifier.of(ns, RESTUtil.decodeString(table));
+    metricsReporter.reportMetric(prefix, tableIdentifier, reportMetricsRequest.report());
```
dimas-b: Sorry for causing disruption on this PR 😅 We probably need to go through `prefixParser.prefixToCatalogName(realmContext, prefix)` here (cf. `withCatalog(...)`)... the previous version used the catalog name as `warehouse` in `metricsReporter.reportMetric()`, but `prefix` might be different.

nit: it might be worth clarifying the meaning of `warehouse` in `PolarisMetricsReporter` too. In Polaris the usual term is "catalog name".

adutra: That's right, I asked for it previously:

https://github.com/apache/polaris/pull/2887#discussion_r2460410478

And then I forgot about it 😆

dimas-b (dismissed):
Thanks for bearing with me @cccs-cat001 🙂

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
cccs-cat001: oops got lost in the noise 😅

dimas-b:
As @adutra noted, let's wait at least until tomorrow for other people to have a chance at reviewing.

Side note: we may want to add Authorization checks for metric reporting (in a follow-up PR, of course).

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.reporting;
+
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.metrics.MetricsReport;
+
+public interface PolarisMetricsReporter {
+  public void reportMetric(String catalogName, TableIdentifier table, MetricsReport metricsReport);
```
flyrain, line 25: nit: 
```suggestion
  void reportMetric(String catalogName, TableIdentifier table, MetricsReport metricsReport);
```

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
singhpk234: Agree, though It can be any persistence, not necessarily an OLTP imho an OLAP is a valid consideration too, so is dumping in object storge so analytical engines can make a table on top of from Polaris POV, it seems to me that we are settling on catalog name and tble identifiers in addition to report from interface POV, why is realm-id not part of it is my doubt, why is authenticate prinicipals not part of these metric ?

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
singhpk234, line 25: [non-blocking] don't we need stuff like realm-id / authenitcated polaris prinicipals etc in this ? It would be really helpful to see which table is read most by which service prinicipal

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
flyrain: This producer is annotated `@ApplicationScoped` but it returns a `PolarisMetricsReporter`, which is `@RequestScoped`. That’s problematic in CDI because an `@ApplicationScoped` bean lives for the entire application lifecycle, while a `@RequestScoped` bean only lives for a single request.
If the reporter (or a future implementation) ever injects request-scoped beans like `RealmContext` or `SecurityContext`, this setup would either fail at deployment or produce invalid references once the request ends.

Instead of producing a reporter directly in an app-scoped producer, introduce a small factory:

```java
@ApplicationScoped
public class PolarisMetricsReporterFactory {

  public PolarisMetricsReporter getOrCreate(String type) {

  }
}
```

Similar to `MetaStoreManagerFactory`,  which ensure the scope correct, while avoid creating a new object for every request. It is also more flexible when we introduce more reporter types.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
flyrain: Good point. I think prinicipals are super useful for audit use cases.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
jbonofre: A factory is `ApplicationScoped`(or `Singleton`), but a bean can be `ApplicationScoped` or `RequestScoped`:

* Should the bean exist only during the lifespan of a request? If so, use [@RequestScoped](http://docs.oracle.com/javaee/7/api/javax/enterprise/context/RequestScoped.html).
* Should the bean exist during the lifespan of the application and should the state of the bean be shared between all the requests? If so, use [@ApplicationScoped](http://docs.oracle.com/javaee/7/api/javax/enterprise/context/ApplicationScoped.html).

So, to use `ApplicationScoped` here, a factory is probably clearer.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,52 @@
...
+@Identifier("default")
+public class DefaultMetricsReporter implements PolarisMetricsReporter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultMetricsReporter.class);
+
+  private final TriConsumer<String, TableIdentifier, MetricsReport> reportConsumer;
+
+  public DefaultMetricsReporter() {
+    this(
+        (catalogName, table, metricsReport) ->
+            LOGGER.info("{}.{}: {}", catalogName, table, metricsReport));
```
snazy, line 40: This adds a lot of logging output. I'd prefer to have a no-op default implementation.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.iceberg.metrics.MetricsReport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RequestScoped
+@Identifier("default")
+public class DefaultMetricsReporter implements PolarisMetricsReporter {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultMetricsReporter.class);
+
+  private final TriConsumer<String, TableIdentifier, MetricsReport> reportConsumer;
```
snazy, line 35: Generally `TriConsumer` makes it difficult to reason about the argument types in lambda expressions.
However, as this is only used as an implementation detail, it's okay.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
adutra: I'm -1 on the factory approach, The factory approach in Polaris is generally used for creating _and caching_ realm-scoped beans. The producer method approach OTOH is used for selecting beans at runtime according to their `@Identifier`.

Here, it's clearly an application-scoped bean, selectable at runtime with its `@Identifier`.

The correct fix for me is just to replace `@RequestScoped` with `@ApplicationScoped` in the producer method:

```java
@ApplicationScoped
public MetricsReporter metricsReporter(...)
```

snazy: > The correct fix for me is just to replace `@RequestScoped` with `@ApplicationScoped` in the producer method:

+1

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
adutra: They can both be injected. I'm not sure we need to modify the interface.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
adutra: This logger is OFF by default.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,52 @@
...
+import io.smallrye.common.annotation.Identifier;
+import jakarta.enterprise.context.RequestScoped;
+import org.apache.commons.lang3.function.TriConsumer;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.metrics.MetricsReport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RequestScoped
+@Identifier("default")
```
pingtimeout, line 31: Would it be possible to use another name than `default` here?  This currently results in the configuration file saying "The default value for polaris.iceberg-metrics.reporting.type is `default`", which does not give any meaningful information.

This class could probably be named `LoggingMetricsReporter` and have a `logging` identifier.  Then as @snazy suggests, there could be a `no-op` reporter that does nothing.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
pingtimeout: @snazy note that by default the log level for this package is set to `OFF`.  So what this code will do is perform a boolean check every time and return.

That being said, I would also prefer to have a `no-op` implementation as well as this `logging` implementation and have the `polaris.iceberg-metrics.reporting.type` config parameter be very explicit on which one is picked, instead of relying on logging configuration to achieve the same goal.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
adutra: We already have a dozen CDI beans named `default`. Why is this one problematic?

adutra: For context, the author had initially 2 beans, a no-op and a logging one. 

I am the one that suggested that we could combine both together, since the logger level could be OFF by default. I still don't see why we would have 2, but fair enough.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
snazy: `OFF` SGTM as well.

pingtimeout: 👍

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
pingtimeout: > I am the one that suggested that we could combine both together, since the logger level could be OFF by default. I still don't see why we would have 2, but fair enough. 

I am ok with a single bean, that was just a link to Robert's other comment.

> We already have a dozen CDI beans named default. Why is this one problematic?

I was not aware of that :-).  I don't this this one is _more_ problematic than any other.  This is just an attempt at having clearer names.  Feel free to consider this as a nit.  If this name is consistent with the rest of the codebase, I guess `default` is fine.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
cccs-cat001: that's the reason why I made this interface :P I'm injecting them with my private implementation just fine.

dimas-b:
Are we ok to merge? I see a lot of recent comments resolved, but no fresh approvals :sweat_smile:

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
flyrain: I don't think my concern was addressed.

dimas-b: `PolarisMetricsReporter` is not request-scoped. Its CDI scope is not defined by this interfaces. The scope is defined by the producer method. `@ApplicationScoped` is correct given the current codebase. If the code ever changes to require per-request state, the producer will have to be adjusted... but such adjustments are not required now, IMHO.

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import io.smallrye.common.annotation.Identifier;
+import jakarta.enterprise.context.RequestScoped;
+import org.apache.commons.lang3.function.TriConsumer;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.metrics.MetricsReport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RequestScoped
```
flyrain: Can we remove this scope annotation if this is meant to be application scoped? More details are here, https://github.com/apache/polaris/pull/2887/files#r2487347077

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
cccs-cat001: So... am I meant to change something? `ServiceProducer::metricsReporter` is `@ApplicationScoped`, and the default implementation of `PolarisMetricsReporter`, `DefaultMetricsReporter` is `@RequestScoped`. Is that a problem?

adutra: @cccs-cat001 please change the scope of the `DefaultMetricsReporter` bean to also `@ApplicationScoped` – there is no reason for it to be request-scoped.

Once this change is done we should be good 🤞

runtime/service/src/main/java/org/apache/polaris/service/reporting/DefaultMetricsReporter.java
dimas-b: Makes sense! Good catch!

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
dimas-b: Actually, given this producer method, I do not think `DefaultMetricsReporter` needs a scope annotation at all. It is always created via this method.

cccs-cat001: Okay I see, I think you're right and we don't need the annotation. Other examples with @ApplicationScope in the ServiceProducer follow the same pattern.

adutra: > Actually, given this producer method, I do not think DefaultMetricsReporter needs a scope annotation at all. It is always created via this method.

Yes and no :-) It doesn't need the annotation but 1) all other beans in the same situation are annotated and 2) annotating the bean allows for direct injection, e.g. in a test:

```java
@Inject @Identifier("default") MetricsReporter defaultMetricsReporter;
```

So, I'm rather in favor of anotating both places, even if I understand that this can cause some confusion.

dimas-b: Annotating both places SGTM 👍 Using `@ApplicationScoped`, right?

adutra:
The current CI failure actually demonstrates that we _must_ annotate the CDI bean with something – otherwise, it's not a CDI bean and so it's not included in the alternatives passed to the producer method.

TLDR: `DefaultMetricsReport` must be annotated with `@ApplicationScoped`.

cc @cccs-cat001 @dimas-b

dimas-b:
TIL: re: bean annotations :sweat_smile: Like I commented on the specific thread (above) I'm ok with annotating both the bean and the producer with `@ApplicationScoped`

cccs-cat001:
Okay final answer (hopefully!!!)

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -0,0 +1,26 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.reporting;
+
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.metrics.MetricsReport;
+
+public interface PolarisMetricsReporter {
```
flyrain, line 24: Not a blocker: would the name `IcebergMetricsReporter` be more descriptive as this is only for iceberg tables, other tables managed by Polaris won't be able to report the metrics?

flyrain (approved):
LGTM. Thanks a lot for adding this, @cccs-cat001 !

adutra (approved):

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
adutra: Good point. But let's merge this PR (the author has been kind enough to address our collective feedback for two weeks now) and address this one in a follow-up task.

adutra:
Thanks you so much @cccs-cat001 for your contribution – and your patience 😅

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
cccs-cat001: there was another TODO mentioned here https://github.com/apache/polaris/pull/2887#issuecomment-3469144259 

If my time allows I'd be happy to help contribute that, and if we can get the requirements defined beforehand :)

PR #2890, opened 2025-10-24
Build: Capture jcstress output in a log file

snazy:
The jcstress output is pretty verbose and prints a lot to the console. This change captures the output in a log file. In case of a test failure, the output is logged to the console, but only in case of a failure.

### DIFF
--- a/persistence/nosql/idgen/impl/build.gradle.kts
+++ b/persistence/nosql/idgen/impl/build.gradle.kts
@@ -90,22 +90,24 @@ tasks.named<JcstressTask>("jcstress") {
   inputs.files(configurations.runtimeClasspath)
   outputs.dir(layout.buildDirectory.dir("reports/jcstress"))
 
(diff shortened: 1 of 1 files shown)


persistence/nosql/idgen/impl/build.gradle.kts
```diff
@@ -85,4 +89,23 @@ tasks.named("jcstress") {
...
+
+  actions.addFirst {
+    logStream = FileOutputStream(logFile)
+    standardOutput = logStream
+    errorOutput = logStream
+  }
+
+  actions.addLast {
+    logStream?.close()
+    if (state.failure != null) {
```
dimas-b: nit: maybe add a `-D` option to enable it in normal runs (for local poking)?

snazy: good idea, added

adutra (approved):

PR #2899, opened 2025-10-25
Use official spark image

MonkeyCanCode:
Currently we are using `jupyter/all-spark-notebook:spark-3.5.0` which is really old and no sign of support newer version from juputer. With this old image, we are then downloading newer spark binary which is a bit problematic if we want to support spark 4 (as there is no base image for that and downloading spark4 on top of this old base image may not work). To have more control over this, I suggest we should use official spark image instead.

### DIFF
--- a/plugins/spark/v3.5/getting-started/notebooks/Dockerfile
+++ b/plugins/spark/v3.5/getting-started/notebooks/Dockerfile
@@ -19,11 +19,8 @@
 
 FROM docker.io/apache/spark:3.5.6-java17
 
-ENV LANGUAGE='en_US:en'
-ENV SPARK_HOME=/opt/spark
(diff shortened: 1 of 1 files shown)


singhpk234 (commented):
LGTM, can you please double confirm you successfully ran the getting started post updating these ?

MonkeyCanCode:
Yes 

> LGTM, can you please double confirm you successfully ran the getting started post updating these ?

MonkeyCanCode:
> Yes
> 
> > LGTM, can you please double confirm you successfully ran the getting started post updating these ?

@singhpk234 let me know if there are additional concerns before I merge this one.

MonkeyCanCode:
Thanks for the review.

> LGTM, thanks @MonkeyCanCode !

PR #2907, opened 2025-10-27
Simplify digest generation

snazy:
Similarly to the change to simplify artifact signing, this change simplifies digest generation by introducing a function to digest the output files of any task. That function takes care of setting up the correct task dependencies and task execution.

Also removes an unnecessary double buffering during digest generation.

### DIFF
--- a/build-logic/src/main/kotlin/publishing/digest-task.kt
+++ b/build-logic/src/main/kotlin/publishing/digest-task.kt
@@ -19,40 +19,52 @@
 
 package publishing
 
+import java.io.File
 import java.security.MessageDigest
 import javax.inject.Inject
 import org.gradle.api.DefaultTask
+import org.gradle.api.Project
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.tasks.Input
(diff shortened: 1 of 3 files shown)


snazy:
Depends on #2906 (source conflicts)

build-logic/src/main/kotlin/publishing/digest-task.kt
```diff
@@ -19,40 +19,52 @@
...
-  val outputFile =
-    objectFactory.fileProperty().convention {
-      val input = file.get().asFile
-      val algo = algorithm.get()
-      input.parentFile.resolve("${input.name}.${algo.replace("-", "").lowercase()}")
-    }
+
+  @Suppress("unused", "UnstableApiUsage")
+  @get:OutputFiles
+  val outputFiles =
```
adutra, line 45: What is this val used for?

snazy: declaring the output artifacts of the task (https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/OutputFiles.html)

PR #2908, opened 2025-10-27
Build-logic: `GitInfo` refactor

snazy:
Allows use of `GitInfo` for other use cases than just Jar manifest attributes. SBOM generation will be another use case for Git information.

### DIFF
--- a/build-logic/src/main/kotlin/publishing/MemoizedJarInfo.kt
+++ b/build-logic/src/main/kotlin/publishing/MemoizedJarInfo.kt
@@ -24,46 +24,24 @@ import org.gradle.api.java.archives.Attributes
 import org.gradle.kotlin.dsl.extra
 
 /**
- * Container to memoize Git information retrieved via `git` command executions across all Gradle
- * projects. Jar release artifacts get some attributes added to the jar manifest, which can be quite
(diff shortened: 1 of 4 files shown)


build-logic/src/main/kotlin/publishing/MemoizedJarInfo.kt
```diff
@@ -24,46 +24,24 @@ import org.gradle.api.java.archives.Attributes
...
-        val gitHead = execProc(rootProject, "git", "rev-parse", "HEAD")
-        val gitDescribe =
-          if (isRelease)
-            try {
-              execProc(rootProject, "git", "describe", "--tags")
-            } catch (_: Exception) {
-              execProc(rootProject, "git", "describe", "--always", "--dirty")
-            }
-          else ""
+        val gi = GitInfo.memoized(rootProject)
```
adutra, line 44: nit: `gi` or `git`?

adutra (approved):

build-logic/src/main/kotlin/publishing/MemoizedJarInfo.kt
snazy: gi for git info ;)

PR #2909, opened 2025-10-27
Memoize ASF project information

snazy:
Information included in Polaris publications pulls some information about the project from ASF project metadata sources (Whimsey). This information is currently only used when generating Maven poms, but will also be needed in SBOMs.

This change adds a new, memoized `AsfProject` information object, which holds the project infromation from Whimsey.

### DIFF
--- a/build-logic/src/main/kotlin/asf/AsfProject.kt
+++ b/build-logic/src/main/kotlin/asf/AsfProject.kt
@@ -41,8 +41,7 @@ class AsfProject(
     fun memoized(project: Project, asfName: String): AsfProject {
       val rootProject = project.rootProject
       return if (rootProject.extra.has("asfProject")) {
-        @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/asf/AsfProject.kt
```diff
@@ -0,0 +1,188 @@
...
+  val bugDatabase: String,
+  val inceptionYear: Int,
+) {
+  companion object {
+
+    fun memoized(project: Project, asfName: String): AsfProject {
+      val rootProject = project.rootProject
+      return if (rootProject.extra.has("asfProject")) {
+        @Suppress("UNCHECKED_CAST")
+        rootProject.extra["asfProject"] as AsfProject
```
adutra: nit: call `unsafeCast`?

snazy: yep, nice catch!

PR #2910, opened 2025-10-27
Build: `GitInfo` function to build a raw github content URL

snazy:


snazy:
Depends on #2908

build-logic/src/main/kotlin/GitInfo.kt
```diff
@@ -56,7 +60,8 @@ class GitInfo(val gitHead: String, val gitDescribe: String) {
               execGit(rootProject, "describe", "--always", "--dirty")
             }
           else ""
-        val gitInfo = GitInfo(gitHead, gitDescribe)
+        val rawLinkRef = if (isRelease) gitDescribe else "HEAD"
```
dimas-b, line 63: nit: why not use `gitDescribe` for non-releases? `HEAD` in GH may be off with respect to the current state of the code during build 🤔

snazy: it potentially says 'abcdef-dirty' (not resolvable) - and the commit may not be in the apache/polaris repo

dimas-b: Does it matter that the commit is not in GH for not released code?

snazy: Yea - the link would be wrong and put in some relevant fields.

PR #2911, opened 2025-10-27
Root POM: link to project's LICENSE, add NOTICE+DISCLAIMER to comment

snazy:
This change replaces the project's license URL in the parent POM from the AL-2.0 template to the `LICENSE` file in the project's root directory.

Before (in `pom.xml`):
```
  <licenses>
    <license>
      <name>Apache-2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>;
    </license>
  </licenses>
```

With this change:
```
  <licenses>
    <license>
      <name>Apache-2.0</name>
      <url>https://raw.githubusercontent.com/apache/polaris/HEAD/LICENSE</url>
      <comments>NOTICE: https://raw.githubusercontent.com/apache/polaris/HEAD/NOTICE
DISCLAIMER: https://raw.githubusercontent.com/apache/polaris/HEAD/DISCLAIMER</comments>
    </license>
  </licenses>
```

### DIFF
--- a/build-logic/src/main/kotlin/publishing/configurePom.kt
+++ b/build-logic/src/main/kotlin/publishing/configurePom.kt
@@ -75,6 +75,7 @@ internal fun configurePom(project: Project, mavenPublication: MavenPublication, 
 
         task.doFirst {
           mavenPom.run {
+            val gitInfo = GitInfo.memoized(project)
             val asfProject = AsfProject.memoized(project, e.asfProjectId.get())
             val asfProjectId = asfProject.apacheId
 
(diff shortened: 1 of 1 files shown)


snazy:
Depends on #2910, #2909, #2908

adutra (approved):

build-logic/src/main/kotlin/publishing/configurePom.kt
```diff
@@ -85,7 +86,14 @@ internal fun configurePom(project: Project, mavenPublication: MavenPublication,
             licenses {
               license {
                 name.set("Apache-2.0") // SPDX identifier
-                url.set(asfProject.licenseUrl)
+                url.set(gitInfo.rawGithubLink("LICENSE"))
+                comments.set(
+                  """
+                  NOTICE: ${gitInfo.rawGithubLink("NOTICE")}
+                  DISCLAIMER: ${gitInfo.rawGithubLink("DISCLAIMER")}
```
dimas-b, line 93: Adding these links for release builds sounds valuable to me, but given current impl. of `GitInfo` for non-release builds (using `HEAD`) I do not think the links would be useful or even correct.

Could we identity whether `GetInfo` has a solid tag and avoid adding links if it does not?

snazy: The link would be correct (e.g. `https://raw.githubusercontent.com/apache/polaris/$rawLinkRef/$file` translates to https://raw.githubusercontent.com/apache/polaris/HEAD/LICENSE) in case it's not a release.
I don't see a better alternative than `HEAD` here for non-release builds, as the Git commit SHA could be from "anywhere".

However, the referenced files are probably not the right ones for the root pom.

Hm - maybe let's drop this PR. I'll leave it open for a bit, but it looks like it's incorrect.

snazy:
Let's abandon this PR.

PR #2914, opened 2025-10-27
Restore original Request ID functionality

### LINKED ISSUE #2913: JDBC persistence: storage/retrieval of principal name from/to events is broken
### Describe the bug

The column name for storing the principal name is `principal_name`:

https://github.com/apache/polaris/blob/e02bb7171ba7af1aeba224e4d5fd585ea6c88624/persistence/relational-jdbc/src/main/resources/postgres/schema-v3.sql#L131

But the code attempts to retrieve the column `actor` instead:

https://github.com/apache/polaris/blob/8e344c7d282bd27318be73e881bc15bc429c1fce/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelEvent.java#L86

### To Reproduce

_No response_

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Additional context

_No response_

### System information

_No response_

adutra:
This PR enacts the action items decided in the following ML thread:

https://lists.apache.org/thread/4gmlcn84o7n8d12qnpl7grfbm4zypk7b

The action items are:

1. Restore the original functionality for request IDs, change the default header name back to x-request-id
2. Remove RequestIdGenerator and related functionality.
3. Update PolarisEvent, expose request ID if available, expose OTel context if available.
4. Update events table SQL schema: insert request ID if available, insert OTel context if available.

Furthermore, this PR also fixes #2913 and adds a small integration test for the JDBC events sink: `InMemoryBufferEventListenerIntegrationTest`.


runtime/service/src/main/java/org/apache/polaris/service/events/listeners/inmemory/InMemoryBufferEventListener.java
```diff
@@ -97,6 +98,23 @@ protected String getRequestId() {
...
+    Span span = Span.current();
+    if (span.getSpanContext().isValid()) {
+      // construct a string from the span context according to W3C Trace Context proposal
+      // https://www.w3.org/TR/trace-context/
+      return "00-"
+          + span.getSpanContext().getTraceId()
+          + "-"
+          + span.getSpanContext().getSpanId()
+          + "-"
+          + span.getSpanContext().getTraceFlags().asHex();
```
dimas-b, line 113: Shall we embed some spec ID into this string? E.g. `traceparent:00-0af7651916cd43dd8448eb211c80319c-b9c7c989f97918e1-01`

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
```diff
@@ -18,60 +18,38 @@
...
   }
 
   @ServerResponseFilter
   public void addResponseHeader(
       ContainerRequestContext request, ContainerResponseContext response) {
     String requestId = (String) request.getProperty(REQUEST_ID_KEY);
-    if (requestId != null) { // can be null if request ID generation fails
-      response.getHeaders().add(loggingConfiguration.requestIdHeaderName(), requestId);
+    if (requestId != null) {
+      response.getHeaders().add(tracingConfiguration.requestId().headerName(), requestId);
```
dimas-b, line 57: Do we have any specific use cases / requests for response headers? My impression from the `dev` ML discussion was that this feature was not in demand 🤔

adutra: I did some archelogy prior to working on this, and my conclusion is that we _probably_ want a response header.

First, Dropwizard has a request ID fitler:

https://github.com/dropwizard/dropwizard/blob/3833f0e1d9fb8cae256fe09379733b8d651f8b87/dropwizard-jersey/src/main/java/io/dropwizard/jersey/filter/RequestIdFilter.java#L42-L49

Note how this filter creates random request IDs (!!!) _and_ sets a response header.

I don't know for sure if this filter is enabled by default. I think it is, because `dropwizard-core` depends on `dropwizard-jersey`. It exists since 2017.

OTOH, Polaris-Dropwizard only had this small line alluding to request IDs: 

https://github.com/apache/polaris/blob/4b18ec065ff16f74b11bc85fdc6ea9036eca7274/dropwizard/service/src/main/java/org/apache/polaris/service/dropwizard/PolarisApplication.java#L516

But you'll notice that the header is now `request_id` 🤷 . I wonder what the intent was here.

Furthermore the only other usage of request IDs in Polaris-Dropwizard seems to be in the log pattern:

https://github.com/apache/polaris/blob/4b18ec065ff16f74b11bc85fdc6ea9036eca7274/polaris-server.yml#L150

Notice the mysterious `{rid}` MDC context, that seems to mean "request ID". This is surprising. I would expect `request_id` instead, since that's the MDC variable name.

All of this to say:

1. I have no idea how this could be working 🤣 
2. I think it doesn't hurt to echo the request ID in the response header.

WDYT?

adutra: That said... I keep going back and forth with this, so I wouldn't mind removing the response header. What do others think?

persistence/relational-jdbc/src/main/resources/postgres/schema-v3.sql
```diff
@@ -131,6 +131,7 @@ CREATE TABLE IF NOT EXISTS events (
     principal_name TEXT,
     resource_type TEXT NOT NULL,
     resource_identifier TEXT NOT NULL,
+    otel_context TEXT,
```
flyrain, line 134: Is backfill needed for existing data?

runtime/service/src/main/java/org/apache/polaris/service/tracing/RequestIdFilter.java
dimas-b: From my POV it is easy to keep the response header, but it will be hard to remove it later, especially if the usage expectations are not clear :thinking: I'd prefer removing it now and re-adding if there's user demand.

@collado-mike : WDYT?

persistence/relational-jdbc/src/main/resources/postgres/schema-v3.sql
dimas-b: No, I do not think so. We mentioned this in 1.2.0 release notes: `The persistence schema is subject to change in future releases, and previously stored event data MAY become unreadable (i.e., dropped) after an upgrade.`

This data was not persisted before 1.2.0, AFAIK.

https://polaris.apache.org/downloads/#120

adutra: No, it's nullable.

CHANGELOG.md
```diff
@@ -31,8 +31,17 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
...
+- JDBC persistence: a new column has been added to the `events` table to store the OpenTelemetry 
+  context that was active when the event was emitted. This column is nullable. To update your schema
+  to the latest version, run the following SQL statement: 
+  ```sql 
+  ALTER TABLE POLARIS_SCHEMA.EVENTS ADD COLUMN IF NOT EXISTS otel_context TEXT; 
+  ```
+
 ### Breaking changes
 
+- The default request ID header name has changed from `Polaris-Request-Id` to `x-request-id`.
```
flyrain: Can we also change it in the doc page https://polaris.apache.org/in-dev/unreleased/telemetry?

adutra: Good catch! Will do.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelEvent.java
```diff
@@ -71,41 +102,47 @@ public interface ModelEvent extends Converter<PolarisEvent> {
...
   @Override
   default PolarisEvent fromResultSet(ResultSet rs) throws SQLException {
     var modelEvent =
         ImmutableModelEvent.builder()
-            .catalogId(rs.getString("catalog_id"))
-            .eventId(rs.getString("event_id"))
-            .requestId(rs.getString("request_id"))
-            .eventType(rs.getString("event_type"))
-            .timestampMs(rs.getLong("timestamp_ms"))
-            .principalName(rs.getString("actor"))
```
adutra, line 86: This change effectively fixes #2913.

adutra:
Closing this. I will open separate PRs for each thing that needs fixing according to the decisions taken on various ML threads.

PR #2925, opened 2025-10-28
Prefer PolarisPrincipal.getRoles in Resolver

XN137:
it should be sufficient to rely on `SecurityContext.getUserPrincipal` alone, we dont need to call `isUserInRole` explicitly.

note due to the `ResolverTest` testing with non-existent roles we have to add null-filtering to the `Resolver`.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
@@ -28,6 +28,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.polaris.core.PolarisCallContext;
@@ -69,7 +70,6 @@ public class Resolver {
 
   // the id of the principal making the call or 0 if unknown
   private final @Nonnull PolarisPrincipal polarisPrincipal;
-  private final @Nonnull SecurityContext securityContext;
 
   // reference catalog name for name resolution
   private final String referenceCatalogName;
@@ -137,7 +137,6 @@ public class Resolver {
     this.diagnostics = diagnostics;
     this.polarisMetaStoreManager = polarisMetaStoreManager;
     this.cache = cache;
-    this.securityContext = securityContext;
     this.referenceCatalogName = referenceCatalogName;
 
     // validate inputs
@@ -467,11 +466,11 @@ public class Resolver {
 
     // update all principal roles with latest
     if (!this.resolvedCallerPrincipalRoles.isEmpty()) {
-      List<ResolvedPolarisEntity> refreshedResolvedCallerPrincipalRoles =
-          new ArrayList<>(this.resolvedCallerPrincipalRoles.size());
-      this.resolvedCallerPrincipalRoles.forEach(
-          ce -> refreshedResolvedCallerPrincipalRoles.add(this.getFreshlyResolved(ce)));
-      this.resolvedCallerPrincipalRoles = refreshedResolvedCallerPrincipalRoles;
+      this.resolvedCallerPrincipalRoles =
+          resolvedCallerPrincipalRoles.stream()
+              .map(this::getFreshlyResolved)
+              .filter(Objects::nonNull)
+              .collect(Collectors.toList());
     }
 
     // update referenced catalog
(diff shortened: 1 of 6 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -791,22 +788,22 @@ private List<ResolvedPolarisEntity> resolveAllPrincipalRoles(
...
-   * @param roleNames
    * @return the filtered list of resolved principal roles
    */
   private List<ResolvedPolarisEntity> resolvePrincipalRolesByName(
       List<ResolvedPolarisEntity> toValidate, Set<String> roleNames) {
     return roleNames.stream()
-        .filter(securityContext::isUserInRole)
+        .filter(roleName -> polarisPrincipal.getRoles().contains(roleName))
         .map(roleName -> resolveByName(toValidate, PolarisEntityType.PRINCIPAL_ROLE, roleName))
+        .filter(Objects::nonNull)
```
XN137, line 806: https://github.com/apache/polaris/blob/2c77fbf530a7fa3e31b1001713bab47007c65a63/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/BaseResolverTest.java#L182-L183

this test was failing with NPE without null filtering... tbh i am not sure i understand it... it's creating a Principal with a non-existing role... shouldnt the Resolver be unsuccessful in that case?
currently it silently does not resolve that role.

dimas-b: It is generally the responsibility of the `Authenticator` to provide role names. However, given that it is a pluggable component and Polaris supports identity federation, core code should be resilient to roles being not present in the Polaris database.

I do not think the resolver should fail in this case. `PolarisAuthorizer` may deny the request if the unresolved role is critical for reaching the necessary permissions, though.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -791,22 +788,22 @@ private List<ResolvedPolarisEntity> resolveAllPrincipalRoles(
...
    *
-   * @param toValidate
-   * @param roleNames
    * @return the filtered list of resolved principal roles
    */
   private List<ResolvedPolarisEntity> resolvePrincipalRolesByName(
       List<ResolvedPolarisEntity> toValidate, Set<String> roleNames) {
     return roleNames.stream()
-        .filter(securityContext::isUserInRole)
+        .filter(roleName -> polarisPrincipal.getRoles().contains(roleName))
```
dimas-b, line 804: I agree that from the `polaris-core` perspective using `Principal` roles is preferable to using `securityContext::isUserInRole` as it reduces the scope of components that need to be in sync WRT auth data (note: `SecurityContext` is well-defined only for Web / REST applications).

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -791,22 +788,22 @@ private List<ResolvedPolarisEntity> resolveAllPrincipalRoles(
...
-   * Resolve the specified list of principal roles. The SecurityContext is used to determine whether
-   * the principal actually has the roles specified.
+   * Resolve the specified list of principal roles. The PolarisPrincipal is used to determine
+   * whether the principal actually has the roles specified.
    *
-   * @param toValidate
-   * @param roleNames
    * @return the filtered list of resolved principal roles
    */
   private List<ResolvedPolarisEntity> resolvePrincipalRolesByName(
```
adutra, line 801: I still think that there is something fishy with the fact that we re-resolve the principal entity and principal role entities here, at each pass.

The only authoritative source of truth for principal names and roles should be the `SecurityIdentity` produced during authentication. 

If the principal and/or principal roles are modified _after_ the authentication by some other process, we could get in the awkward situation where we would fetch different entities from the database, causing a mismatch with what `SecurityIdentity` exposes.

IMHO, we should stop resolving principals and principal roles altogether here. That's a prerogative of the `Authenticator`. This class should focus on resolving _catalog roles_ and _paths_.

adutra (commented):
Thanks! I was planning to tackle this myself :-) 

But I think we should go one step further, see my comment.

Nit: I think you could also remove the `getResolvedCallerPrincipal` method, it's unused.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
dimas-b: Good point, @adutra ! I agree that roles should be established during authentication and not re-resolved after that.

However, I may be a pretty big change. IIRC, in case of Polaris-manages Principals the idea is that the state of permission assignments should be consistent with catalog data during request execution. This probably means that the roles entities used by the authenticator should be made available to the `Resolver` in order to take their version numbers into account when traversing grant records.

I think the change to use Principal role names in this PR is consistent with current code. Let's merge this PR and refactor more in follow-up PRs. WDYT?

adutra: Sure that works for me. 

> This probably means that the roles entities used by the authenticator should be made available to the Resolver in order to take their version numbers into account when traversing grant records.

Maybe, but this needs more investigation imho. 

I still think it should possible to avoid that. Looking at the `Resolver#resolveByName()` method, we see that it ultimately calls `PolarisMetaStoreManager#loadResolvedEntityByName()` which returns the grant records for the entity. That creates a "snapshot" of [principal name + principal roles + catalog roles + grant records] that is consistent for a given pass, afaict (at least as consistent as it can be, given that the operation implies several reads that are not in the same transaction).

PR #2932, opened 2025-10-29
Prefer PolarisPrincipal over SecurityContext

XN137:
Follow-up to https://github.com/apache/polaris/pull/2925

The general idea is that `SecurityContext` comes from `jakarta.ws.rs` and
there is no reason for non-REST related classes to rely on those details.
Instead, once preprocessing of a REST-request has inferred the
`PolarisPrincipal` all inner/core code should rely on only that.

Note that this simplifies a bunch of tests that had to create their own
`SecurityContext` around the principal that they wanted to use, thus
having to decide how to implement `isUserInRole` and the other methods.

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
@@ -21,7 +21,6 @@ package org.apache.polaris.core.persistence.resolver;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import jakarta.annotation.Nullable;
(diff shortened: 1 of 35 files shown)


dimas-b (dismissed):
`SecurityContext` is an `rs-api` (Web / REST framework) class, indeed. Consolidating `polaris-core` dependencies to just `PolarisPrincipal` looks very reasonable to me 👍

XN137:
rebased after trivial merge conflict in:
```
runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogViewTest.java
```

snazy (approved):
Boring - I like it :)

adnanhemani (approved):
LGTM!

PR #2936, opened 2025-10-29
Fix drop with purge test in test_spark_sql_s3_with_privileges.py

travis-bowen:
This test hasn't been making the correct assertions for quite a while. It was trying to assert that delete with purge deleted the table data and metadata, but it was verifying against a storage location that was not the actual base location of the table.

Validated with docker compose command.
t_pyspark/src/test_spark_sql_s3_with_privileges.py ...............       [100%]
================== 15 passed, 31 warnings in 82.85s (0:01:22) ==================
Test SUCCEEDED: t_pyspark:test_spark_sql_s3_with_privileges.py

### DIFF
--- a/regtests/t_pyspark/src/test_spark_sql_s3_with_privileges.py
+++ b/regtests/t_pyspark/src/test_spark_sql_s3_with_privileges.py
@@ -495,15 +495,28 @@ def test_spark_credentials_can_delete_after_purge(root_client, snowflake_catalog
                       aws_secret_access_key=response.config['s3.secret-access-key'],
                       aws_session_token=response.config['s3.session-token'])
 
+    # Extract the table location from the metadata_location in the response
+    # metadata_location format: s3://bucket/path/to/table/metadata/v1.metadata.json
+    # We need to extract the base table path (everything before /metadata/)
+    metadata_location = response.metadata_location
+    assert metadata_location.startswith('s3://')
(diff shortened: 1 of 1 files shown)


regtests/t_pyspark/src/test_spark_sql_s3_with_privileges.py
```diff
@@ -495,15 +495,28 @@ def test_spark_credentials_can_delete_after_purge(root_client, snowflake_catalog
...
+    path_without_scheme = metadata_location[5:]  # Remove 's3://'
+    path_parts = path_without_scheme.split('/', 1)  # Split bucket and path
+    bucket_from_metadata = path_parts[0]
+    full_path = path_parts[1] if len(path_parts) > 1 else ''
+    # Extract table base path (everything before /metadata/)
+    table_base_path = full_path.rsplit('/metadata/', 1)[0] if '/metadata/' in full_path else ''
+
     objects = s3.list_objects(Bucket=test_bucket, Delimiter='/',
-                              Prefix=f'{aws_bucket_base_location_prefix}/snowflake_catalog/db1/schema/{table_name}/data/')
+                              Prefix=f'{table_base_path}/data/')
```
HonahX, line 512: How does the checks before drop with purge passed since we were checking the wrong locations. :scream: Seems we still somehow find some data files there.

collado-mike (approved):

Issue #2272, opened 2025-08-06
Helm chart provides incorrect defaults and documentation


bacek:
https://github.com/apache/polaris/blob/bae5444cf37ef86a2c472ded33c513dff3350053/helm/polaris/values.yaml#L567

`type` for both `authenticator` and `roleProvider` should be one of `internal`, `external`, or `hybrid`. `default` is not accepted by polaris itself

adutra:
Hi @bacek thanks for reporting this.

There are two different settings:

```yaml
authentication.type: internal
authentication.authenticator.type: default
```

* `authentication.type` can be: `internal`, `external`, or `mixed`.
* `authentication.authenticator.type` can only be `default` (for now).

Does that make sense?

bacek:
Hi @adutra 

I probably didn't put it clearly.

Problem is: existing helm chart provides defaults which can't be used. Generated configmap with defaults is
```
polaris.authentication.active-roles-provider.type=default
polaris.authentication.authenticator.type=default
polaris.authentication.token-broker.max-token-generation=PT1H
polaris.authentication.token-broker.type=rsa-key-pair
polaris.authentication.token-service.type=default
polaris.authentication.type=internal
polaris.file-io.type=default
polaris.log.request-id-header-name=Polaris-Request-Id
polaris.persistence.type=in-memory
polaris.rate-limiter.filter.type=no-op
polaris.realm-context.realms=POLARIS
polaris.realm-context.type=default
quarkus.http.port=8181
quarkus.log.category."org.apache.iceberg.rest".level=INFO
quarkus.log.category."org.apache.polaris".level=INFO
quarkus.log.console.enable=true 
quarkus.log.console.format=%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p [%c{3.}] [%X{requestId},%X{realmId}] [%X{traceId},%X{parentId},%X{spanId},%X{sampled}] (%t) %s%e%n quarkus.log.console.level=ALL
quarkus.log.file.enable=false
quarkus.log.level=INFO
quarkus.management.port=8182quarkus.otel.sdk.disabled=true                                                                                                                                                      
```

Pod is failing with
```
[...]

PR #2939, opened 2025-10-29
Docs: update Helm Chart page to show usage without cloning Polaris github repo

olsoloviov:
…

Minor documentation fix intended to let users know how to use Polaris Helm Charts 

There was a mention in https://github.com/apache/polaris/issues/2272 that it is not clear where to obtain released Polaris Helm charts, so it made sense to update the doc for users who want to use released distributives instead of working with Polaris code repo.

### DIFF
--- a/helm/polaris/README.md
+++ b/helm/polaris/README.md
@@ -42,18 +42,85 @@ A Helm chart for Apache Polaris (incubating).
 
 ## Installation
 
-### Running locally with a Minikube cluster
+The instructions below are for the local Minikube cluster. They assume Minikube and Helm are installed.
 
-The below instructions assume Minikube and Helm are installed.
+### Installing from the official repository (recommended)
 
-Start the Minikube cluster:
+Start the Minikube cluster, add the official Polaris Helm repository, create the target namespace, and install the chart:
+```bash
+minikube start
+helm repo add polaris https://downloads.apache.org/incubator/polaris/helm-chart
+helm repo update
+kubectl create namespace polaris
+helm install polaris polaris/polaris --namespace polaris --devel
+```
+
+> [!NOTE]
+> The --devel flag is required while Polaris is in the incubation phase.
+> Helm treats the -incubating suffix as a pre‑release by SemVer rules, and will skip charts that are not in a stable versioning scheme by default.
+
+Now the cluster should be up and running. You can run the built-in connection test to verify:
+```bash
+helm test polaris --namespace polaris
+```
+
+Default Polaris Helm chart values are for a non-persistent backend. To use a persistent backend, you will need to override default values and create required resources.
+Polaris Helm chart package includes a number of files that can be used as a starting point for such configuration. To use them, download the chart into a local directory:
+```bash
+helm pull polaris \
+  --repo https://downloads.apache.org/incubator/polaris/helm-chart \
+  --devel \
+  --untar \
+  --untardir ./helm
+```
+
+> [!WARNING]
+> The examples below use values files located in the `helm/polaris/ci` directory.
+> **These files are intended for testing purposes primarily, and may not be suitable for production use**.
+> For production deployments, create your own values files based on the provided examples.
+
+Create required resources:
+```bash
+kubectl apply --namespace polaris -f helm/polaris/ci/fixtures/
+kubectl wait --namespace polaris --for=condition=ready pod --selector=app.kubernetes.io/name=postgres --timeout=120s
+```
+
+> [!WARNING]
+> The Postgres deployment set up in the fixtures directory is intended for testing purposes only and is not suitable for production use. For production deployments, use a managed Postgres service or a properly configured and secured Postgres instance.
+
+Install the chart with a persistent backend:
+```bash
+helm upgrade --install --namespace polaris \
+  --values helm/polaris/ci/persistence-values.yaml \
+  polaris helm/polaris
+kubectl wait --namespace polaris --for=condition=ready pod --selector=app.kubernetes.io/name=polaris --timeout=120s
+```
+
+Run the catalog bootstrap using the Polaris admin tool to initialize the catalog with the required configuration.
+For example, to run it as a new pod in the Minikube cluster (the polaris-admin-tool image will be pulled automatically from Docker Hub):
+```bash
+kubectl run polaris-bootstrap \
+  -n polaris \
+  --image=apache/polaris-admin-tool:latest \
+  --restart=Never \
+  --rm -it \
+  --env="quarkus.datasource.username=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.username}' | base64 --decode)" \
+  --env="quarkus.datasource.password=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.password}' | base64 --decode)" \
+  --env="quarkus.datasource.jdbc.url=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.jdbcUrl}' | base64 --decode)" \
+  -- \
+  bootstrap -r POLARIS -c POLARIS,root,pass -p
+```
+
+### Build and install from source (advanced)
+
+Here it is assumed that you have cloned the Polaris Git repository and set up prerequisites to build the project (see [Quickstart Guide](https://polaris.apache.org/in-dev/unreleased/getting-started/install-dependencies/) for details).
+
+Start the Minikube cluster, then build and load the image into the Minikube cluster:
+
 ```bash
 minikube start
 eval $(minikube docker-env)
-```
 
-If you have cloned Polaris git repository, you may want to build and load image into the Minikube cluster:
-```bash
 ./gradlew \
   :polaris-server:assemble \
   :polaris-server:quarkusAppPartsBuild --rerun \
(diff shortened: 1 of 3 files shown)


helm/polaris/README.md
```diff
@@ -60,10 +62,24 @@ eval $(minikube docker-env)
...
+minikube image pull apache/polaris:latest
+minikube image pull apache/polaris-admin-tool:latest
+```
+
 ### Installing the chart locally
 
 The below instructions assume a local Kubernetes cluster is running and Helm is installed.
 
+> [!NOTE]
+> The exampled below assume that you have cloned Polaris git repository, which contains Helm chart in `helm/polaris` directory.
```
adutra:
```suggestion
> The examples below assume that you have cloned Polaris git repository, which contains the Polaris Helm chart in the `helm/polaris` directory.
```

helm/polaris/README.md
```diff
@@ -60,10 +62,24 @@ eval $(minikube docker-env)
...
+> If you do not want to clone the repository or want to use official release of Helm chart, you can download it from [Apache Polaris Helm Chart repository](https://downloads.apache.org/incubator/polaris/helm-chart/):
```
adutra:
```suggestion
> If you do not want to clone the repository or want to use an official Helm chart release, you can download it from the [Apache Polaris Helm Chart repository](https://downloads.apache.org/incubator/polaris/helm-chart/):
```

helm/polaris/README.md
```diff
@@ -167,8 +190,12 @@ the Polaris repo root:
 helm unittest helm/polaris
 ```
 
-You can also lint the chart using the Chart Testing tool, with the following command:
+You can also lint the chart using the Chart Testing tool. You will need to install the [yamllint](https://github.com/adrienverge/yamllint) tool:
+```bash
+brew install yamllint
```
adutra: Shouldn't we show how to install `ct` as well?

olsoloviov: We do - earlier in Prerequisites section. I can move yamlint instructions there as well

olsoloviov: done in the latest commit

helm/polaris/README.md
```diff
@@ -142,11 +165,13 @@ The following tools are required to run the tests:
 
 * [Helm Unit Test](https://github.com/helm-unittest/helm-unittest)
 * [Chart Testing](https://github.com/helm/chart-testing)
+* [yamllint](https://github.com/adrienverge/yamllint)
 
 Quick installation instructions for these tools:
 ```bash
 helm plugin install https://github.com/helm-unittest/helm-unittest.git
 brew install chart-testing
+brew install yamllint
```
MonkeyCanCode, line 218: @olsoloviov thanks for the change. Please update https://github.com/apache/polaris/blob/main/Makefile#L27 to include this new dependency.

site/content/in-dev/unreleased/helm.md
```diff
@@ -22,36 +22,22 @@ type: docs
 weight: 675
```
MonkeyCanCode, line 22: For this file, you can use `make helm-doc-generate` to generate instead of manual edit.

helm/polaris/README.md
olsoloviov: @MonkeyCanCode, thank you for reviewing. Fixed with a new commit

site/content/in-dev/unreleased/helm.md
olsoloviov: Thanks @MonkeyCanCode, I did not know it was automatically copied. But I see that there are already merged changes that diverge helm.md from helm/polaris/readme.md - markup for blockquotes was changed in https://github.com/apache/polaris/pull/2656.
Also, there is a problem with rendering badges in the original page header when hosted on the apache.org:
image
Apparently apache.org content security policy does not allow embedding images from external hosts:
```
Refused to load the image 'https://img.shields.io/badge/Version-1.2.0--incubating--SNAPSHOT-informational?style=flat-square' because it violates the following Content Security Policy directive: "default-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
```
It was my motivation to remove the header from page version for the site.

As a sidenote, we have other pages on the site that are copies from getting-started docs (e.g. Keycloak setup  page, Telemetry setup page), but they are not copied automatically. It would be good to have some unified approach for such cases.

MonkeyCanCode: Yes. That is correct. If u run local, it will all show up. Either we will need to update the policy or have a better way to handle these. Manually edit those can be fairly tedious. @flyrain

helm/polaris/README.md
```diff
@@ -60,10 +62,24 @@ eval $(minikube docker-env)
   -Dquarkus.container-image.build=true
 ```
 
+Otherwise load ready images from Docker Hub:
+```bash
+minikube image pull apache/polaris:latest
+minikube image pull apache/polaris-admin-tool:latest
+```
```
flyrain: Explicitly pulling isn't needed, Helm will pull images automatically if `imagePullPolicy`: `IfNotPresent` or `Always` is configured in the chart (which it is by default for most charts). All we need here is: 
```
Otherwise, Helm will automatically pull released images during installation.
```

site/content/in-dev/unreleased/helm.md
flyrain: Agreed with @MonkeyCanCode to rely on automation. Can we also add a post-process step to `make helm-doc-generate`, e.g., generate helm.md automatically with minor adjustments (strip badges, convert blockquotes)?

flyrain (commented):
Thanks a lot @olsoloviov for working on this! 
Since we are here, I’d suggest splitting this section into two parts:
- Recommended: Installing from the official Helm repository. This is the simplest and most common way to deploy Polaris, without cloning or building locally.
- Advanced: Building and running from source. For developers who want to test local changes or custom images.

This separation would make the “happy path” much clearer for most users, while still keeping the advanced instructions for contributors. It would also eliminate the need for commands like `minikube image pull` in the main flow since Helm automatically pulls images from the registry.

Something like the followng:

### Install from official Helm repository (recommended)
```
minikube start
helm repo add polaris https://downloads.apache.org/incubator/polaris/helm-chart
helm repo update
kubectl create namespace polaris
helm install polaris polaris/polaris --namespace polaris
```
### Build and install from source (advanced)
```
minikube start
eval $(minikube docker-env)
./gradlew :polaris-server:assemble :polaris-server:quarkusAppPartsBuild \
  :polaris-admin:assemble :polaris-admin:quarkusAppPartsBuild \
  -Dquarkus.container-image.build=true
helm upgrade --install polaris ./helm/polaris --namespace polaris
```

helm/polaris/README.md
adutra: This isn't for Helm, but for the kubelet inside Minikube. The kubelet can only download images if they are publicly available.

If the image is from a private registry, or only present locally in the host machine (e.g. because you built a custom `apache/polaris:latest` image from sources and you want to use it), then this instruction is necessary.

All in all, I think it's fine to leave the instructions.

site/content/in-dev/unreleased/helm.md
adutra: I would love to see this automated, but unfortunately the two pages are different, especially the warning box style.

I think we would need two different templates.

adutra: FWIW, I think this needs to be addressed separately, in a follow-up PR.

olsoloviov: I agree it is a good idea to converge helm.md back with helm chart readme and sync them automatically. 
I see that we can go back from Hugo shortcodes to github-style blockqoutes for alert blocks - I made it work locally, but it requires updating Hugo version. It may bring additional concerns (full site testing etc.), so I would make it a separate PR as @adutra suggested.

helm/polaris/README.md
flyrain: Thanks for the clarification, that makes sense. Good point about the kubelet behavior inside Minikube. I agree it’s fine to keep the `minikube image pull` instructions for users running with private or locally built images.

That said, I still think it’d be clearer to separate the doc into two sections,  one for the standard “Helm install from the official repo” path, and another “advanced / from source” section for developers. This way, most users can follow the simple flow without extra steps, while contributors still have the detailed Minikube setup if needed. More details are in this comment, https://github.com/apache/polaris/pull/2939#pullrequestreview-3409217448.

olsoloviov: Thanks @flyrain for your comments. I agree that clear separation of "user" and "dev" flows is a good thing - I remember there was even an initiative to clearly split all of Polaris documentation that way.
Regarding this particular document, there is a couple of things to consider:
- flows would be largely duplicating each other (we will need to download helm chart anyway to install persistent backend and create test setup for resources using helm/polaris/ci/fixtures)
- current helm.md document is generated from helm/chart/readme.md, which assumes that helm chart was downloaded or github repo was cloned, so it kinda implies being "advanced" by default. I was specifically asked to keep the documents synced: https://github.com/apache/polaris/pull/2939#discussion_r2483023146

So maybe it makes sense to create a separate document for a standard path and add a link to advanced path to it, wdyt?

flyrain: Thanks for the reply. I think it's more reasonable to put them in one place. Even helm chart was downloaded or github repo was cloned, user don't need or want to rebuild docker image, etc. In fact, helm chart can be downloaded separately without other Polaris source code, users won't be able to build the docker image in that case.

olsoloviov: I have updates the doc with two separate flows as you suggested, @flyrain. 
helm.md still has some manual changes after being generated from helm/polaris/readme.md.gotmpl due to blockquotes not properly rendered in Hugo - I will fix it in a follow up PR

site/content/in-dev/unreleased/helm.md
```diff
@@ -144,11 +214,13 @@ The following tools are required to run the tests:
 
 * [Helm Unit Test](https://github.com/helm-unittest/helm-unittest)
 * [Chart Testing](https://github.com/helm/chart-testing)
+* [yamllint](https://github.com/adrienverge/yamllint)
 
 Quick installation instructions for these tools:
 ```bash
 helm plugin install https://github.com/helm-unittest/helm-unittest.git
 brew install chart-testing
+brew install yamllint
```
dimas-b, line 223: nit: I do not mind installing `yamllint`, but I could not find references to it from the examples... Why does it need to be installed?

dimas-b (approved):

site/content/in-dev/unreleased/helm.md
olsoloviov: It is not used directly in our samples, but It is required by `ct lint`

dimas-b: thx!

dimas-b:
Thanks for your contribution, @olsoloviov !

Issue #2785, opened 2025-10-09
Add Server header to Polaris API responses to indicate Polaris version


fivetran-arunsuri:
**Description:**

Currently, Polaris API responses do not include a Server header. For example, when performing a POST /api/catalog request, the response headers include only:

Content-Type: application/json;charset=UTF-8
content-encoding: gzip
content-length: 187

<img width="893" height="369" alt="Image" src="https://github.com/user-attachments/assets/c0bbea9d-dbcc-42bc-bb8f-b4da31f5f962" />

There is no header identifying the Polaris server or its version.

**Enhancement Proposal:**
Add a standard Server header in all HTTP responses from Polaris services.
For example:

Server: Polaris/1.1.0

**Rationale:**

- Provides better observability and diagnostics during development and debugging.
- Allows users and clients to easily verify the server version they’re communicating with, especially across multi-environment or multi-region deployments.
- Simplifies support and compatibility checks when interacting with different Polaris API versions.
- Aligns with common REST API practices — many Apache and commercial services include Server or custom headers (e.g., X-Server-Version) for version traceability.

**Optional Consideration:**
If exposing the server version publicly is a concern for security reasons, a custom header like:

X-Polaris-Version: 1.1.0, could be used instead.

dimas-b:
@fivetran-arunsuri : could you provide specific examples of Apache projects that provide server or version headers in their REST API responses (just for reference)?

dimas-b:
From my POV observability and compatibility checks are very different use cases. The server/version header does make sense for observability, but I believe it might be preferable to use a different mechanism for compatibility checks. The latter is often not as simple as checking `version >= X`, so using capability flags may be more maintainable.

fivetran-arunsuri:
Thanks, @dimas-b.  agreed, capability flags would indeed be more robust for compatibility management.

Our primary use case is observability and very basic validation. We’d like the Polaris server to include a version identifier in the standard Server header (e.g., Server: Polaris/1.1.0). On the client side, we plan to send an X-Expected-Server header to validate that the response comes from the intended Polaris deployment or version.

fivetran-arunsuri:
> [@fivetran-arunsuri](https://github.com/fivetran-arunsuri) : could you provide specific examples of Apache projects that provide server or version headers in their REST API responses (just for reference)?

[...]

PR #2941, opened 2025-10-30
Add Server header to Polaris API responses to indicate Polaris version

fivetran-arunsuri:
Fixes https://github.com/apache/polaris/issues/2785

This change adds a Server response header to all HTTP responses, formatted as:

**Server: Polaris/1.1**

**Details**

PR to use the Quarkus http.header mechanism for enabling Polaris version

When enabled, the header will be included in all outgoing HTTP responses.

The primary purpose of this feature is observability — allowing clients or monitoring tools to identify the responding service version.

**Testing:**

- Added unit tests to verify:
  Header inclusion when the flag is enabled.
  Absence of the header when the flag is disable
- Verified using Postman

image

### DIFF
--- a/site/content/in-dev/unreleased/configuring-polaris-for-production.md
+++ b/site/content/in-dev/unreleased/configuring-polaris-for-production.md
@@ -124,6 +124,7 @@ polaris.realm-context.require-header=true
 
 This will cause Polaris to also return a `404 Not Found` response if the realm header is not present
 in the request.
+
 ### Metastore Configuration
 
 A metastore should be configured with an implementation that durably persists Polaris entities. By


runtime/service/src/main/java/org/apache/polaris/service/http/ServerHeaderFilter.java
```diff
@@ -0,0 +1,57 @@
...
+      @ConfigProperty(name = "polaris.http.version-header.enabled", defaultValue = "false")
+          boolean headerEnabled) {
+    this.headerEnabled = headerEnabled;
+  }
+
+  @Override
+  public void filter(
+      ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
+    if (headerEnabled) {
+      responseContext.getHeaders().putSingle(HttpHeaders.SERVER, SERVER_HEADER_VALUE);
```
fivetran-arunsuri: I can use `jakarta.ws.rs.core.HttpHeaders` but I will have to define a custom SERVER header for the same. Let me know if I have to change it

adutra:
Hi @fivetran-arunsuri Quarkus already exposes a similar feature:

https://quarkus.io/guides/http-reference#additional-http-headers

Couldn't you leverage that?

For example:

```properties
quarkus.http.header."Server".value=Polaris/${quarkus.application.version}
```

dimas-b:
The `quarkus.http.header` approach looks interesting and promising from my POV :slightly_smiling_face: 

It will also allow downstream builds to control whether the version header is exposed or not with per-deployment granularity (which is sufficient IMHO).

However, users of ASF Polaris binaries may not be able to "unset" the default value of this property :thinking: Those users will only be able to override it to something like `quarkus.http.header."Server".value=NotAvailable`. Alternatively, playing with the `path` and/or `method` sub-properties may be needed to exclude this header from responses... All in all, this might still be fine for end users, but we'll probably need to document it well in comments.

fivetran-arunsuri:
Thanks @adutra @dimas-b — I verified the Quarkus option works:

`quarkus.http.header."Server".value=Polaris/${quarkus.application.version}
`

I kept the ServerHeaderFilter behind polaris.http.version-header.enabled (default false) based on the previous dev email discussion to have a true off switch (fully removes the header, not just overrides). This addresses the “can’t unset” concern you mentioned. I’ll also document the Quarkus property as a downstream-friendly alternative for per-deployment control.

Unless there are objections, I’ll merge this as-is and add docs:

- how to enable via polaris.http.version-header.enabled=true
- how downstreams can set the Quarkus property instead if they prefer config-only.

If you’d rather we switch to config-only now, happy to do that, but I think this strikes a good balance (explicit toggle + simple override path).

dimas-b:
If we can avoid adding java code to support the new header, it would be preferable from my POV.

@fivetran-arunsuri : could you check whether adding `quarkus.http.header."Server".path=/irrelevant` actually removes the `Server` header from responses?

If that works, it should be sufficient. I do not expect suppressing this header to be a common use case and advanced uses can figure it out (provided we add comments in the default config).

fivetran-arunsuri:
Thanks @dimas-b, confirmed that adding a non-matching path (/__disabled__) prevents the header from being added, while keeping .value defined avoids the config-validation error.

simplified the PR to use the Quarkus http.header mechanism only, drop the custom filter, and document:

- how to enable via %prod.quarkus.http.header."Server".path=/*
- how to keep it disabled by default with path=/__disabled__.

dimas-b (approved):
Thanks, @fivetran-arunsuri !

snazy:
> If we can avoid adding java code to support the new header, it would be preferable from my POV.

I'd prefer this "no additional code" approach as well and document it.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -52,6 +52,14 @@ quarkus.http.cors.exposed-headers=*
 quarkus.http.cors.access-control-max-age=PT10M
 quarkus.http.cors.access-control-allow-credentials=true
 
+# --- Polaris Server header (Quarkus-managed) ---
+# Keep .value present (required by Quarkus), but make the path never match so it's effectively OFF by default.
+quarkus.http.header."Server".value=Polaris/${quarkus.application.version}
```
adutra: Do we need to put anything in this file? The property is not marked as a build-time property in the [configuration reference](https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus-http-header-header-value).

I think users could just add this setting in their own configuration. There should be no need to rebuild Polaris, just add it in one of the [supported configuration sources](https://quarkus.io/guides/config-reference#configuration-sources).

dimas-b: Good point. Doc changes alone should be sufficient for users.

flyrain: +1 on doc only, we don't have to make it default here. Thanks @adutra for the suggestion.

fivetran-arunsuri: Makes sense, thanks everyone for confirming. I’ll proceed with docs-only changes since users can add this configuration themselves.

site/content/in-dev/unreleased/configuring-polaris-for-production.md
```diff
@@ -125,6 +125,19 @@ polaris.realm-context.require-header=true
...
+Polaris can emit an informational `Server` HTTP response header using Quarkus' built-in header
+configuration. Add the following property to one of the supported configuration sources (for example,
+`application.properties`) to enable it with the Polaris version string:
+
+```properties
+quarkus.http.header."Server".value=Polaris/${quarkus.application.version}
+```
+
+If you prefer to scope the header to specific environments, only set the property for the desired
+profile (for example, `%prod`).
```
snazy:
```suggestion
```
I could imagine that "specific environment" is confusing for non-Quarkus affine users.

adutra (approved):
LGTM thanks @fivetran-arunsuri !

site/content/in-dev/unreleased/configuring-polaris-for-production.md
```diff
@@ -125,6 +125,19 @@ polaris.realm-context.require-header=true
 This will cause Polaris to also return a `404 Not Found` response if the realm header is not present
 in the request.
 
+### Polaris Server Header
```
snazy: Hm, I think it's better to move this section down to the end of the doc.
It's (at least IMHO) an advanced option for very specific use cases, while the other items mentioned here are all mandatory settings.

site/content/in-dev/unreleased/configuring-polaris-for-production.md
fivetran-arunsuri: Thanks for pointing that out, @snazy — that’s a fair observation. I do think the current phrasing (“specific environment”) works fine here, though, since the following example (%prod) already makes it clear that we’re referring to Quarkus profiles.
The intent is mainly to help readers understand where to apply the property, and the term “environment” keeps it a bit more approachable for those who might not be deeply familiar with Quarkus terminology. I’d prefer to keep it as is, but happy to revisit if others find it confusing.

site/content/in-dev/unreleased/configuring-polaris-for-production.md
fivetran-arunsuri: Fair point, moved to the end of doc

fivetran-arunsuri:
If everything looks good with the PR, would someone with write access be able to merge it? I don’t currently have the permissions to do so. Thank you

site/content/in-dev/unreleased/configuring-polaris-for-production.md
dimas-b: From my POV having this phrase in the doc is ok. People who go to this level of configuration / customization should be aware of the Quarkus runtime and its configuration features.

PR #2943, opened 2025-10-30
Change getting-start docker file to use official spark image from outdated jupyter image

MonkeyCanCode:
Missing changes for https://github.com/apache/polaris/pull/2899. I noticed `getting-started/spark/notebooks/Dockerfile` is using the same old jupyter spark image, thus created this PR to match to what we have in regtest. Couple additional changes made:
1. replace `source VENV` with `. VENV` (this remove the needed of change SHELL in dockerfile
2. move requirements.txt (the new one with jupyter) to regtest dir

### DIFF
--- a/regtests/notebook_requirements.txt
+++ b/regtests/notebook_requirements.txt
@@ -18,4 +18,4 @@
 #
 
 jupyterlab==4.4.10
-ipykernel==7.0.1
+ipykernel==7.1.0


getting-started/spark/notebooks/SparkPolaris.ipynb
```diff
@@ -792,7 +792,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.6"
+   "version": "3.10.12"
```
adutra, line 795: Why do we need to downgrade this?

plugins/spark/v3.5/getting-started/notebooks/Dockerfile
```diff
@@ -34,14 +34,13 @@ USER spark
...
-COPY --chown=spark plugins/spark/v3.5/getting-started/notebooks/requirements.txt /tmp/notebook_requirements.txt
 COPY --chown=spark regtests/requirements.txt /tmp
+COPY --chown=spark regtests/notebook_requirements.txt /tmp
 COPY --chown=spark plugins/spark/v3.5/spark/build/2.12/libs /home/spark/polaris_libs
 
-SHELL ["/bin/bash", "-c"]
 
 RUN python3 -m venv /home/spark/venv && \
-    source /home/spark/venv/bin/activate && \
+    . /home/spark/venv/bin/activate && \
```
dimas-b, line 43: just wondering: what is the reason behind this change?

getting-started/spark/notebooks/SparkPolaris.ipynb
MonkeyCanCode: Spark 3.5 (as well as latest 4.0) is using python 3.10 from official image. The jupyter image (very old and outdated) is using 3.11.6. As we are using official spark image, the jupyter is now installed on top of the new base.

MonkeyCanCode:
> Hey @MonkeyCanCode thanks for this PR! Do you mind changing the title to something more descriptive?

Done.

plugins/spark/v3.5/getting-started/notebooks/Dockerfile
MonkeyCanCode: So `source` is only available if u have `SHELL ["/bin/bash", "-c"]`...this is to simplify that (we do this in couple place already in current repo)

regtests/notebook_requirements.txt
dimas-b, line 1: Why are we moving this file? It does not look like there are any related changes in regtests 🤔

MonkeyCanCode: So I created this file in the previous PR mentioned in the current ticket. As I noticed there are two places which both needed this, thus moved it from plugin/spark to regtest/

dimas-b: .... but regtest does not use this file, does it?

MonkeyCanCode: That is correct. Currently only getting start example and plugin spark use it. Now both of them already have references to requirements.txt (which only contains poetry) under regtest. This, I moved it out from plugin spark into (added from previous PR) into here instead.

dimas-b: Thanks for the clarification! Since you maintain this code most of the time, I'm happy to accept what works best from your POV :slightly_smiling_face:

MonkeyCanCode: Sounds good. Yeah, i think the new placement does make sense for now. Thanks for the review.

singhpk234 (approved):
LGTM, thanks @MonkeyCanCode !

[request] Before merging can you please help double verify the getting started and regtest works with this change :) !

MonkeyCanCode:
Yes, I did that before PR. Thanks for the review.

> LGTM, thanks @MonkeyCanCode !
> 
> [request] Before merging can you please help double verify the getting started and regtest works with this change :) !

Issue #2843, opened 2025-10-20
List Polaris tools on the website


jbonofre:
You should list the tools and quickly describe the use case on the Polaris website.

Today, `polaris-tools` repository is not actually "visible" on the website, and the users don't know what they can find here:
- benchmark tool
- catalog migrator
- soon UI
- ...

adam-christian-software:
Howdy, @jbonofre. Can you assign to me? I'd love to take this on.

adam-christian-software:
This should be completed now.

dimas-b:
Closing as implemented by #2946. @jbonofre : please reopen if something is missing 🙂

jbonofre:
Good for me. Thanks !

jbonofre:
I reopen this one because we added the tools in the documentation, but we should provide visibility directly on the website, with the purpose of each tool.
For instance, I would expect a "Tools" entry in the menu listing all tools and what is the purpose of each.

adam-christian-software:
@jbonofre - Sounds good. I'll make sure to do this specific thing.

jbonofre:
We have a page on the website now: https://polaris.apache.org/tools/

PR #2946, opened 2025-10-30
docs(2843): Add documentation around Polaris-Tools

adam-christian-software:
# Motivation - Why Changes are Needed
This change starts referring to the Polaris-Tools repository within the Apache Polaris root ReadMe and Polaris website. This implements #2843 .

My goal here is to change the documentation based on use case rather than repository. Meaning, in the Polaris website, I put references to the Polaris-Tools in a Catalog Migration section rather than having a separate Polaris-Tools section. This makes it likely for users to easily discover the specific tool they need.

I did not reference the benchmarks directly since this is more related to development rather than end user usage. I did reference that there are tools in the README.

### Small Changes
- I did some small refactoring of the site README to help me better understand how to work with the site.

Yes, it's a docs change.

Locally by firing up the localhost.

### CHANGELOG.md
No need here. This is just a docs change.

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -103,9 +103,9 @@ the existing directory.
 cd site/
 
 cd content/releases/
-## Create a new empty branch 'versioned-docs'
+# Create a new empty branch 'versioned-docs'
 git worktree add --orphan -b versioned-docs .
-## (add some content to the content/releases/ directory - at least the "top-level" _index.adoc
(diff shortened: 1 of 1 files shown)


dimas-b (approved):
LGTM 👍 Thanks, @adam-christian-software !

dimas-b:
Re: #2843 - maybe reference it using the `#` notation in the commit message and/or PR description (to get an auto-generated link)?

PR #2854, opened 2025-10-21
Servcie: Remove a duplicated config

flyrain:

PR #2955, opened 2025-11-01
Docs: Fix regtest doc

MonkeyCanCode:
1. Change outdated docker command to `make client-regenerate`
2. Remove reference for python3.9 as that is removed from https://github.com/apache/polaris/pull/2795

### DIFF
--- a/regtests/README.md
+++ b/regtests/README.md
@@ -185,30 +185,10 @@ Other commands are available in the `regtests/t_spark_sql/src` directory.
 ## Python Tests
 
 Python tests are based on `pytest`. They rely on a python Polaris client, which is generated from the openapi spec.
-The client can be generated using two commands:
+The client can be generated using make command:
 
 ```shell
-# generate the management api client
-docker run --rm \
-  -v ${PWD}:/local openapitools/openapi-generator-cli generate \
-  -i /local/spec/polaris-management-service.yml \
-  -g python \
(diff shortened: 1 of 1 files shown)


MonkeyCanCode:
Will do. I haven't do for this one as it was only doc change. 

> LGTM Thanks @MonkeyCanCode !
> can you please double confirm running the regtests that they work post this

MonkeyCanCode:
> Will do. I haven't do for this one as it was only doc change.
> 
> > LGTM Thanks @MonkeyCanCode !
> > can you please double confirm running the regtests that they work post this

hi @singhpk234, so if u tried to run the test locally (either with/without docker), it will always failed if u don't have the latest image built locally. This is because the docker file is now pointing to `latest` from dockerhub (version 1.2.0) but there is one breaking change on their build and it got fixed by @flyrain last week via https://github.com/apache/polaris/pull/2854/files (in this one, `POST /v1/{prefix}/transactions/commit` got print twice from catalog config). However, if u run the change with the local image (latest main), you will be able to get completed build:
```
regtest-1 exited with code 0
```

This has nothing to do with the changes in this PR. let me know if you have any concern before the merge. 
cc. @flyrain 

Thanks,
Yong

ebyhr (approved):

singhpk234 (approved):
LGTM, thanks for checking and the code change @MonkeyCanCode

PR #2957, opened 2025-11-01
Refactor: improve and clean up Dockerfiles

MonkeyCanCode:
1. Combine multiple RUN layers into a single RUN layer to reduce image size
2. Combine ENV/ARG instructions for easily readability
3. Load polaris spark job automatically instead of hard-coded value and only copied needed jar instead of whole directory
4. Cleanup and simplify Dockerfiles 

General cleanup and simplify existed dockerfiles

No

Tested with various test suits we have around those docker files such as docker-compose instructions.

### CHANGELOG.md
N/A

### DIFF
--- a/plugins/spark/v3.5/getting-started/notebooks/Dockerfile
+++ b/plugins/spark/v3.5/getting-started/notebooks/Dockerfile
@@ -19,8 +19,6 @@
 
 FROM docker.io/apache/spark:3.5.6-java17
 
-ARG CURRENT_SCALA_VERSION=2.12
-
 ENV PYSPARK_PYTHON=/home/spark/venv/bin/python \
     PYTHONPATH="${SPARK_HOME}/python/:${SPARK_HOME}/python/lib/py4j-0.10.9.7-src.zip:/home/spark/venv/lib/python3.10/site-packages"
 
@@ -38,7 +36,7 @@ WORKDIR /home/spark
 COPY --chown=spark client /home/spark/client
 COPY --chown=spark regtests/requirements.txt /tmp
 COPY --chown=spark regtests/notebook_requirements.txt /tmp
-COPY --chown=spark plugins/spark/v3.5/spark/build/${CURRENT_SCALA_VERSION}/libs/*bundle.jar /opt/spark/jars/
+COPY --chown=spark plugins/spark/v3.5/spark/build/2.12/libs/*bundle.jar /opt/spark/jars/
 
 RUN python3 -m venv /home/spark/venv && \
     . /home/spark/venv/bin/activate && \
--- a/plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
+++ b/plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
@@ -265,6 +265,8 @@
     "from pyspark.sql import SparkSession\n",
     "\n",
     "spark = (SparkSession.builder\n",
+    "  # This jar is now automatically discovered, thus no longer needed\n",
+    "  #.config(\"spark.jars\", \"../polaris_libs/polaris-spark-3.5_2.12-1.2.0-incubating-SNAPSHOT-bundle.jar\")\n",
     "  .config(\"spark.jars.packages\", \"org.apache.iceberg:iceberg-aws-bundle:1.10.0,io.delta:delta-spark_2.12:3.2.1\")\n",
     "  .config(\"spark.sql.catalog.spark_catalog\", \"org.apache.spark.sql.delta.catalog.DeltaCatalog\")\n",
     "  .config('spark.sql.iceberg.vectorization.enabled', 'false')\n",
--- a/plugins/spark/v3.5/regtests/Dockerfile
+++ b/plugins/spark/v3.5/regtests/Dockerfile
@@ -38,7 +38,7 @@ WORKDIR /home/spark/polaris
 
 COPY --chown=spark:spark ./v3.5 /home/spark/polaris/v3.5
 
-# /home/spark/regtests might not be writable in all situations, see https://github.com/apache/polaris/pull/205
+# /home/spark/.../regtests might not be writable in all situations, see https://github.com/apache/polaris/pull/205
 RUN chmod -R 777 /home/spark/polaris/v3.5/regtests
 
 USER spark
--- a/site/docker/Dockerfile
+++ b/site/docker/Dockerfile
@@ -22,9 +22,9 @@ FROM ubuntu:24.04 AS hugo
 ENV LANGUAGE='en_US:en'
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends golang hugo asciidoctor npm curl git && \
-    npm install --global http-server && \
+    apt-get install --yes --no-install-recommends golang hugo asciidoctor npm curl git && \
     rm -rf /var/lib/apt/lists/* && \
+    npm install --global http-server && \
     mkdir -p /polaris/site/resources
 
 COPY _run_in_docker.sh /hugo/run


plugins/spark/v3.5/getting-started/notebooks/Dockerfile
```diff
@@ -19,8 +19,10 @@
 
 FROM docker.io/apache/spark:3.5.6-java17
 
-ENV PYTHONPATH="${SPARK_HOME}/python/:${SPARK_HOME}/python/lib/py4j-0.10.9.7-src.zip:/home/spark/venv/lib/python3.10/site-packages"
-ENV PYSPARK_PYTHON=/home/spark/venv/bin/python
+ARG CURRENT_SCALA_VERSION=2.12
```
snazy: What's the purpose of this argument? I only see 2.12 jars in the image.
Should this be a static ENV instead?

plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
```diff
@@ -265,7 +265,6 @@
     "from pyspark.sql import SparkSession\n",
     "\n",
     "spark = (SparkSession.builder\n",
-    "  .config(\"spark.jars\", \"../polaris_libs/polaris-spark-3.5_2.12-1.2.0-incubating-SNAPSHOT-bundle.jar\")  # TODO: add a way to automatically discover the Jar\n",
```
snazy, line 268: Huh? Isn't that jar needed?

plugins/spark/v3.5/regtests/Dockerfile
```diff
@@ -18,31 +18,29 @@
...
 
 FROM docker.io/apache/spark:3.5.6-java17
-ARG POLARIS_HOST=polaris
-ENV POLARIS_HOST=$POLARIS_HOST
-ENV SPARK_HOME=/opt/spark
-ENV CURRENT_SCALA_VERSION='2.12'
-ENV LANGUAGE='en_US:en'
+
+ARG POLARIS_HOST=polaris \
+    CURRENT_SCALA_VERSION=2.12
```
snazy, line 23: Similar here, only 2.12 jars in the image.

site/docker/Dockerfile
```diff
@@ -21,23 +21,14 @@ FROM ubuntu:24.04 AS hugo
...
-RUN npm install --global http-server
-
-# these dependencies are needed to build the static site
-#RUN npm install --global autoprefixer postcss postcss-cli http-server
-
-RUN mkdir /polaris
-RUN mkdir /polaris/site
-RUN mkdir /polaris/site/resources
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends golang hugo asciidoctor npm curl git && \
```
snazy: Why's `git` needed now?

snazy: Mind keeping the long-arg version `--yes`?

plugins/spark/v3.5/regtests/Dockerfile
```diff
@@ -18,31 +18,29 @@
...
+    mkdir -p /home/spark /tmp/polaris-regtests /opt/spark/conf && \
+    chown -R spark:spark /home/spark /tmp/polaris-regtests && \
+    chmod -R 777 /opt/spark/conf
 
 WORKDIR /home/spark/polaris
 
-COPY --chown=spark ./v3.5 /home/spark/polaris/v3.5
+COPY --chown=spark:spark ./v3.5 /home/spark/polaris/v3.5
 
 # /home/spark/regtests might not be writable in all situations, see https://github.com/apache/polaris/pull/205
```
snazy:
```suggestion
# /home/spark/.../regtests might not be writable in all situations, see https://github.com/apache/polaris/pull/205
```

site/docker/Dockerfile
```diff
@@ -21,23 +21,14 @@ FROM ubuntu:24.04 AS hugo
...
+    npm install --global http-server && \
+    rm -rf /var/lib/apt/lists/* && \
```
snazy, line 26: Super-super-super-nit: move the `rm` one line up near to the `apt install`.

plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
snazy: Current version on `main` is `1.3.0-incubating-SNAPSHOT`, so I'm not sure whether this removal breaks more ;)

MonkeyCanCode: This jar is needed but we don't need to manually specify it. This is because I am copying the jar into the spark libs directory. So it will get auto loaded. A lot cleaner imo for getting start notebook so we don't need to update this line every time we bump the release number.

site/docker/Dockerfile
MonkeyCanCode: Actually I was thinking about the same as rm is related to apt cache cleanup but npm is part of setup before cleanup, thus above layout. But I think it may be better for instructions with apt grouping then npm at the end.

site/docker/Dockerfile
MonkeyCanCode: So we have install-recommands on earlier which will install missing dependencies if any. As I tried to remove that, I got an error on the npm install line where the package needs git binary. Let me get the exact error and catch here later tonight.

plugins/spark/v3.5/regtests/Dockerfile
MonkeyCanCode: So we actually have both 2.12 and 2.13 plugin jar for spark 3.5 (supported by our CI as well in this dir for the plugin checks). This is used in the bash script under this path if we build 2.13. However, the existing getting start example is using 2.12 only. This ARG is needed as I will need to push the right jar  into the image (which matches the corresponding scala version used on the spark base image...also, the spark image we are using for 3.5 is using scala 2.12 as well). But I guess we can remove this and use 2.12 for now as that is what the spark image is using? For spark 4.x, we will need to change to 2.13 as support for scala 2.12 is dropped.

plugins/spark/v3.5/getting-started/notebooks/SparkPolaris.ipynb
snazy: Ah! Okay, didn't realize that.
Mind keeping that line just commented out for posterity for users - that they have to add that jar?

plugins/spark/v3.5/regtests/Dockerfile
snazy: > support for scala 2.12 is dropped

Finally ;)

plugins/spark/v3.5/getting-started/notebooks/Dockerfile
MonkeyCanCode: Discussed on the other comment. This will be removed for now.

plugins/spark/v3.5/regtests/Dockerfile
MonkeyCanCode: And for reference, `plugins/spark/v3.5/regtests/run.sh` use this. However, this is not needed for `plugins/spark/v3.5/getting-started/notebooks/Dockerfile`, thus, I removed it from there.

site/docker/Dockerfile
MonkeyCanCode: Here is the error I was referring to when `git` was not installed:
```
[+] Running 6/6
 ✔ polaris_site-hugo               Built                                                                                                                             0.0s
 ✔ Network polaris_site_default    Created                                                                                                                           0.0s
 ✔ Volume polaris_site_workspace   Created                                                                                                                           0.0s
 ✔ Volume polaris_site_resources   Created                                                                                                                           0.0s
 ✔ Volume polaris_site_hugo_cache  Created                                                                                                                           0.0s
 ✔ Container polaris_site-hugo-1   Created                                                                                                                           0.0s
Attaching to hugo-1
hugo-1  | go: github.com/google/docsy@v0.10.0: git init --bare in /hugo/cache/modules/filecache/modules/pkg/mod/cache/vcs/98f59a39bbbc2d4de9185a32a3d8790111161ae9778e3e5e32a90c95f026ac9b: exec: "git": executable file not found in $PATH
...
hugo-1  | Error: command error: failed to load modules: failed to download modules: failed to execute 'go [mod download -modcacherw]': failed to execute binary "go" with args [mod download -modcacherw]: go: github.com/google/docsy@v0.10.0: git init --bare in /hugo/cache/modules/filecache/modules/pkg/mod/cache/vcs/98f59a39bbbc2d4de9185a32a3d8790111161ae9778e3e5e32a90c95f026ac9b: exec: "git": executable file not found in $PATH
```

MonkeyCanCode:
@snazy all comments should be addressed now. Please take another look when you have a chance.

PR #2962, opened 2025-11-03
Introducing changes to enable forwarding of Polaris events to AWS Cloudwatch.

vchag:
These changes were created to address concerns raised in Issues https://github.com/apache/polaris/issues/2630

1. Streamlining Event-to-JSON Serialization:
The previous architecture involved a redundant, two-step conversion process: events were transformed into intermediate Maps by an abstract class, and only then was the resulting map serialized into JSON (as seen in AwsCloudWatchEventListener).
Since Jackson JSON is our chosen serialization format, we've removed this unnecessary intermediate mapping step. We introduced Jackson Mixins to provide native JSON serialization support directly on the event objects. This refactoring simplifies the serialization pipeline from:
Event -> Map -> JSON
to
Event -> JSON

2. Introduces `AllEventsForwardingListener`, a generic event listener designed to automatically receive and forward all Polaris events without requiring explicit event-type configuration.

3. Introduces `polaris.event-listener.aws-cloudwatch.event-types` property:  This property introduces a configurable event filter for the AWS CloudWatch event listener.
It defines which Polaris event types should be serialized and forwarded to CloudWatch Logs.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/AllEventsForwardingListener.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/AllEventsForwardingListener.java
@@ -43,7 +43,7 @@ public abstract class AllEventsForwardingListener implements PolarisEventListene
   protected abstract void handle(PolarisEvent event);
 
   /** Optional filter (config-based). Default: handle all. */
-  protected boolean shouldHandle(Object event) {
+  protected boolean shouldHandle(PolarisEvent event) {
     return true;
   }
 
--- a/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
@@ -89,15 +89,15 @@ public class AwsCloudWatchEventListener extends AllEventsForwardingListener {
   }
 
   @Override
-  protected boolean shouldHandle(Object event) {
-    if (!(event instanceof PolarisEvent polarisEvent)) {
+  protected boolean shouldHandle(PolarisEvent event) {
+    if (event == null) {
       return false;
     }
 
     if (this.listenToAllEvents) {
       return true;
     }
-    Class<? extends PolarisEvent> actualType = polarisEvent.getClass();
+    Class<? extends PolarisEvent> actualType = event.getClass();
     return allowedEventTypes.stream().anyMatch(cfg -> cfg.isAssignableFrom(actualType));
   }
 


runtime/defaults/src/main/resources/application.properties
```diff
@@ -145,6 +145,7 @@ polaris.event-listener.type=no-op
 # polaris.event-listener.aws-cloudwatch.log-stream=polaris-cloudwatch-default-stream
 # polaris.event-listener.aws-cloudwatch.region=us-east-1
 # polaris.event-listener.aws-cloudwatch.synchronous-mode=false
+# polaris.event-listener.aws-cloudwatch.event-types= // the absence of this property would result if processing all Polaris event types.
```
adutra:
```suggestion
# polaris.event-listener.aws-cloudwatch.event-types= // the absence of this property would result in processing all Polaris event types.
```

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -86,17 +92,22 @@ class AwsCloudWatchEventListenerTest {
...
   private AutoCloseable mockitoContext;
+  private ObjectMapper objectMapper;
 
   @BeforeEach
   void setUp() {
     mockitoContext = MockitoAnnotations.openMocks(this);
     executorService = Executors.newSingleThreadExecutor();
 
+    // Build the test mapper and apply the same customizations Quarkus would
+    objectMapper = new ObjectMapper();
```
adutra: nit: could be a final field.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -71,13 +78,38 @@ public class AwsCloudWatchEventListener extends PropertyMapEventListener {
   public AwsCloudWatchEventListener(
       AwsCloudWatchConfiguration config,
       Clock clock,
-      PolarisIcebergObjectMapperCustomizer customizer) {
+      PolarisIcebergObjectMapperCustomizer customizer,
```
adutra: You don't need to pass the customizer, the customizer will be invoked for you by Quarkus. Just inject the object mapper.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/AllEventsForwardingListener.java
```diff
@@ -0,0 +1,854 @@
...
+ * <p>This design follows the Template Method pattern, centralizing shared control flow in the base
+ * class while allowing subclasses to supply the event-specific behavior.
+ */
+public abstract class AllEventsForwardingListener implements PolarisEventListener {
+
+  /** Subclasses implement the actual logic once, generically. */
+  protected abstract void handle(PolarisEvent event);
+
+  /** Optional filter (config-based). Default: handle all. */
+  protected boolean shouldHandle(Object event) {
```
adutra:
```suggestion
  protected boolean shouldHandle(PolarisEvent event) {
```

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -71,13 +78,38 @@ public class AwsCloudWatchEventListener extends PropertyMapEventListener {
...
     this.region = Region.of(config.awsCloudWatchRegion());
     this.synchronousMode = config.synchronousMode();
     this.clock = clock;
+    this.objectMapper = mapper;
     customizer.customize(this.objectMapper);
+    this.listenToAllEvents =
+        config.eventTypes().isEmpty()
+            || config.eventTypes().map(Set::isEmpty).orElse(true)
+            || config.eventTypes().get().stream().anyMatch(e -> e == PolarisEvent.class);
+    this.allowedEventTypes = listenToAllEvents ? Set.of() : Set.copyOf(config.eventTypes().get());
```
adutra: nit: I think we can simplify a bit:

```java
    this.allowedEventTypes = config.eventTypes().orElse(Set.of());
    this.listenToAllEvents =
        allowedEventTypes.isEmpty()
            || allowedEventTypes.stream().anyMatch(c -> c == PolarisEvent.class);
```

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -71,13 +78,38 @@ public class AwsCloudWatchEventListener extends PropertyMapEventListener {
...
+  @Override
+  protected boolean shouldHandle(Object event) {
+    if (!(event instanceof PolarisEvent polarisEvent)) {
+      return false;
+    }
+
+    if (this.listenToAllEvents) {
+      return true;
+    }
+    Class<? extends PolarisEvent> actualType = polarisEvent.getClass();
```
adutra: Note: this is probably a bit inefficient under a high load. Switching to event type IDs should help because the lookup would be faster when using an EnumSet.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -86,6 +118,14 @@ void start() {
     ensureLogGroupAndStream();
   }
 
+  @PostConstruct
+  void verifyMapper() {
```
adutra: Unnecessary.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/AllEventsForwardingListener.java
```diff
@@ -0,0 +1,854 @@
...
+import org.apache.polaris.service.events.CatalogGenericTableServiceEvents;
+import org.apache.polaris.service.events.CatalogPolicyServiceEvents;
+import org.apache.polaris.service.events.CatalogsServiceEvents;
+import org.apache.polaris.service.events.IcebergRestCatalogEvents;
+import org.apache.polaris.service.events.PolarisEvent;
+import org.apache.polaris.service.events.PrincipalRolesServiceEvents;
+import org.apache.polaris.service.events.PrincipalsServiceEvents;
+
+/**
+ * Base class for event listeners that with to generically forward all {@link PolarisEvent
```
adutra:
```suggestion
 * Base class for event listeners that wish to generically forward all {@link PolarisEvent
```

runtime/defaults/src/main/resources/application.properties
vchag: My apologies for all the typos. I'll do better.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
vchag: Will address this.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
vchag: @adutra 
Fair point. 

1. In addition to using numeric event type IDs, I’m thinking of including a human-readable code string in each PolarisEventType. Any thoughts?
e.g.
```
public enum PolarisEventType {
    AFTER_REFRESH_TABLE(101, "after_refresh_table"),
    BEFORE_REFRESH_TABLE(102, "before_refresh_table"),
    UNKNOWN(0, "unknown"),
    ALL(-1, "all");
```
2. For polaris.event-listener.aws-cloudwatch.event-types property, should we accept IDs, codes, or keep supporting class names only?

runtime/defaults/src/main/resources/application.properties
vchag: Done.

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
vchag: Done.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
vchag: Done.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
vchag: Done.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
vchag: Removed.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/AllEventsForwardingListener.java
vchag: Done.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
adutra: Yes that's roughly the idea I have in mind, see this thread on the ML:

https://lists.apache.org/thread/yx7pkgczl6k7bt4k4yzqrrq9gn7gqk2p

But let's do this later, I think this PR is close to mergeable, so let's try to get it in as soon as possible. The current code is acceptable for now.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -51,33 +55,55 @@
...
+    this.objectMapper = mapper;
+    this.allowedEventTypes = config.eventTypes().orElse(Set.of());
+    this.listenToAllEvents =
+        allowedEventTypes.isEmpty()
+            || allowedEventTypes.stream().anyMatch(c -> c == PolarisEvent.class);
+  }
+
+  @Override
+  protected boolean shouldHandle(PolarisEvent event) {
+    if (event == null) {
```
adutra, line 93: nit: `event` will never be false here, this test is redundant.

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -86,17 +92,22 @@ class AwsCloudWatchEventListenerTest {
 
   private ExecutorService executorService;
   private AutoCloseable mockitoContext;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
```
adutra, line 95:
```suggestion
  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();

  static {
    new PolarisIcebergObjectMapperCustomizer(new MemorySize(BigInteger.valueOf(1024 * 1024)))
        .customize(OBJECT_MAPPER);
  }
```

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -309,17 +327,49 @@ void shouldSendEventInSynchronousMode() {
 
   @Test
   void ensureObjectMapperCustomizerIsApplied() {
```
adutra, line 329: Unnecessary test. You are testing a Quarkus feature.

runtime/service/src/main/java/org/apache/polaris/service/config/PolarisIcebergObjectMapperCustomizer.java
```diff
@@ -56,6 +61,10 @@ public void customize(ObjectMapper objectMapper) {
     objectMapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
     objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     objectMapper.setPropertyNamingStrategy(new PropertyNamingStrategies.KebabCaseStrategy());
+    objectMapper.addMixIn(PolarisEvent.class, PolarisEventBaseMixin.class);
+    objectMapper.addMixIn(TableIdentifier.class, IcebergMixins.TableIdentifierMixin.class);
```
adnanhemani, line 65: I thought these are taken care of by the `RESTSerializers` below?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchConfiguration.java
```diff
@@ -86,4 +89,8 @@ public interface AwsCloudWatchConfiguration {
   @WithName("synchronous-mode")
   @WithDefault("false")
   boolean synchronousMode();
+
+  @WithName("event-types")
+  Optional<Set<Class<? extends PolarisEvent>>>
```
adnanhemani, line 94: Is this possible to do? I don't see the ability to do this in the Quarkus public documentation?

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -86,17 +92,22 @@ class AwsCloudWatchEventListenerTest {
...
 
+    customizer.customize(objectMapper);
+
     // Configure the mocks
     when(config.awsCloudWatchLogGroup()).thenReturn(LOG_GROUP);
     when(config.awsCloudWatchLogStream()).thenReturn(LOG_STREAM);
     when(config.awsCloudWatchRegion()).thenReturn("us-east-1");
-    when(config.synchronousMode()).thenReturn(false); // Default to async mode
+    when(config.synchronousMode()).thenReturn(false); // default async
+    when(config.eventTypes()).thenReturn(java.util.Optional.empty()); // handle all events
```
adnanhemani, line 110: nit: 
```suggestion
    when(config.eventTypes()).thenReturn(Optional.empty()); // handle all events
```

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -236,12 +248,18 @@ void shouldSendEventToCloudWatch() {
           .satisfies(
               logEvent -> {
                 String message = logEvent.message();
+                JsonNode root = objectMapper.readTree(message);
+                JsonNode event = root.path("event").isMissingNode() ? root : root.path("event");
```
adnanhemani, line 252: nit: shouldn't the @JsonUnwrapped make sure that the event information is not under an "event" object?

runtime/service/src/test/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -309,17 +327,49 @@ void shouldSendEventInSynchronousMode() {
...
+            "test_catalog", TableIdentifier.of("db", "table"));
+
+    boolean shouldHandle = listener.shouldHandle(randomEvent);
+    assertThat(shouldHandle)
+        .as("Listener should handle all events when no eventTypes are configured")
+        .isTrue();
   }
 
   private void verifyLogGroupAndStreamExist(CloudWatchLogsAsyncClient client) {
-    // Verify log group exists
```
adnanhemani, line 322: Unnecessary change

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -309,17 +327,49 @@ void shouldSendEventInSynchronousMode() {
...
+
+    assertThat(objectMapper.findMixInClassFor(PolarisEvent.class))
+        .as("Namespace mixin should be registered")
+        .isEqualTo(PolarisEventBaseMixin.class);
+  }
+
+  @Test
+  void shouldListenToAllEventTypesWhenConfigNotProvided() {
+    // given: config.eventTypes() is empty → listen to all events
+    when(config.eventTypes()).thenReturn(java.util.Optional.empty());
```
adnanhemani, line 355: nit:

```suggestion
    when(config.eventTypes()).thenReturn(Optional.empty());
```

runtime/service/src/test/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchEventListenerTest.java
```diff
@@ -309,17 +327,49 @@ void shouldSendEventInSynchronousMode() {
...
+
+    AwsCloudWatchEventListener listener =
+        new AwsCloudWatchEventListener(config, clock, objectMapper);
+
+    // This is any random PolarisEvent — if the listener listens to all types,
+    // shouldHandle(event) should return true
```
adnanhemani, line 361: I understand the thought behind this test, but technically this is not the right way to test a "random" event. Maybe you create a new "RandomEvent" Java record that extends "PolarisEvent" (making it clear that it _definitely_ cannot be referred to outside the scope of this test class and show that `shouldHandle` still returns true.

runtime/service/src/main/java/org/apache/polaris/service/config/PolarisIcebergObjectMapperCustomizer.java
adutra: Good catch, `RESTSerializers` have serializers for `TableIdentifier` and `Namespace` already.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/aws/cloudwatch/AwsCloudWatchConfiguration.java
adutra: Yes this is possible. Here is an example:

```properties
polaris.event-listener.aws-cloudwatch.event-types=\
  org.apache.polaris.service.events.IcebergRestCatalogEvents$BeforeCreateNamespaceEvent,\
  org.apache.polaris.service.events.IcebergRestCatalogEvents$AfterCreateNamespaceEvent
```

See default converters here:

https://smallrye.io/smallrye-config/Main/config/getting-started/#converters

runtime/defaults/src/main/resources/application.properties
```diff
@@ -144,6 +144,7 @@ polaris.event-listener.type=no-op
 # polaris.event-listener.aws-cloudwatch.log-stream=polaris-cloudwatch-default-stream
 # polaris.event-listener.aws-cloudwatch.region=us-east-1
 # polaris.event-listener.aws-cloudwatch.synchronous-mode=false
+# polaris.event-listener.aws-cloudwatch.event-types= // the absence of this property would result in processing all Polaris event types.
```
adutra, line 147: You may want to add an example of how to actually specify events here:
```suggestion
# The absence of the property below would result in processing all Polaris event types (default).
# polaris.event-listener.aws-cloudwatch.event-types=\
#   org.apache.polaris.service.events.IcebergRestCatalogEvents$BeforeCreateNamespaceEvent,\
#   org.apache.polaris.service.events.IcebergRestCatalogEvents$AfterCreateNamespaceEvent
```

adutra:
> However, I don't think you fully considered all angles of why the `PropertyMapEventListener` was introduced. There are many event fields that likely should not ever be pushed to an external system - for security reasons (such as table credentials that could be exposed in a `LoadTableResponse`) or for functional reasons (such as when a `TableMetadata` object is larger than 1MB, the maximum size for a record in CloudWatch Logs). I'm sure there are many other reasons here that are not coming top of mind. `PropertyMapEventListener` is supposed to be the transformation layer where only relevant information can be parsed out of the event and then sent onwards.

I have some reservations regarding a common transformation layer for all data sinks. While the idea is sound in principle, I'm skeptical about reaching a consensus on the specific transformations to apply and information to redact.

We'll likely agree on redacting sensitive data. However, other transformations could lead to endless debate. For instance, CloudWatch rejects payloads over 1MB, but a Kafka cluster might accept up to 100MB. Similarly, if a use case doesn't require a specific field, another use case might. And finally, transformations should be vendor-agnostic, of course. 

This leads me to believe that a "transformation layer," despite good intentions, will never be generic enough to serve all purposes. While we could introduce parameters to influence transformations, this would be a significant and time-consuming effort.

Instead, I propose applying the least possible logic to events, ideally only security-related logic. Events should be passed as-is to the sinks, allowing each sink to perform further transformations and redactions on the event payload. I think we should absolutely avoid specific logic per event type, as this would imply having to special-case each one of the 150+ event types.

JSON mixins should also be as generic as possible. If a sink cannot use them due to them not producing the desired JSON, it should use a different object mapper. Alternatively, a sink could use mixins and the default object mapper to produce not a String, but a JsonNode instance, apply further transformations to it, and then send it to the target remote system.

Finally, I don't believe an intermediary Map representation offers much benefit.

Anyways, I think the transformation layer topic deserves a discussion on the ML.

> Maybe I could've done a better job to add a comment to the class stating this - but I can't approve this PR without all these considerations addressed. This is the exact reason why it has been taking me a while to complete the functionality through as well.
> I am still quite interesting in seeing if MixIns can make things better - but I'm not familiar with what other options MixIns/Jackson provide to do automated parsing and/or redacting of the sensitive/larger/etc objects. I am hoping you can help direct this conversation. I would also recommend you to post these findings not only here on the PR but on the Dev Mailing list, as it may help gain more eyes who can help with this issue.

I understand and share the security concerns, but we can't delay this PR indefinitely. Manually addressing each event for transformations is incredibly time-consuming, as you've noted.

Asking the original contributor to implement logic for sensitive or larger objects falls outside the PR's original scope and will take too long (we have 150+ events!). I propose merging this PR now and tackling the transformation layer as a follow-up.

Let's not forget: the events feature is still in beta for a good reason.

adnanhemani:
> I have some reservations regarding a common transformation layer for all data sinks. While the idea is sound in principle, I'm skeptical about reaching a consensus on the specific transformations to apply and information to redact.

I completely agree with this actually - a common transformation layer for all data sinks is likely not possible. But there can (and should) be groupings of Event Listener implementations that follow similar transformations. i.e. AWS CloudWatch Logs, Google Cloud Logging, and Azure Monitor Logs (I've not begun work on the latter two but are future enhancements I am planning for) all have similar maximum record sizes, and we should safely remove the `TableMetadata` objects wherever they exist in an event for these sorts of listeners as a result.

> Finally, I don't believe an intermediary Map representation offers much benefit.

I don't think having an intermediary Map representation is a requirement by any means, but it seemed like the most straight-forward way to grab only the relevant information from each event - even if it is quite verbose. If there is a way to make this much cleaner using MixIns solely, given the concerns that exist, I am all for it.

> I understand and share the security concerns, but we can't delay this PR indefinitely.

While I understand the thought here, I cannot advocate/approve for this PR if it does not resolve the security concerns, at a minimum. We should not be introducing security concerns into our code base (regardless of "beta" status of a feature) in the name of simply moving things forward. Polaris has not done so in the past, and we should continue to apply the same bar for the future. cc @snazy who has pointed out many security concerns in the past.

If you're still in support of merging the PR through without resolving the security concerns, let's start a thread on the ML to ensure the community is aware of this security gap, supports this decision, and acknowledges the precedence that this may inadvertently set for the future.

> Manually addressing each event for transformations is incredibly time-consuming, as you've noted.

That’s precisely why only one event was initially wired through in the original PRs. I’m not against adding more events — a few at a time — as long as they’re properly reviewed to ensure their payloads don’t introduce any issues. Even in that case, we should first outline a clear approach for handling events that may require additional filtering. Committing fully to a method without understanding how it will accommodate all scenarios wouldn’t be a prudent path forward.

adutra:
> I’m not against adding more events — a few at a time — as long as they’re properly reviewed to ensure their payloads don’t introduce any issues.

Let's assume we take 1 week of work per event. We have 150+ events. This effort would take almost 3 years to complete.

I don't think this is sustainable. We must take a step back and question the current design.

> While I understand the thought here, I cannot advocate/approve for this PR if it does not resolve the security concerns, at a minimum.

The author (@vchag) will have to spend weeks, maybe more, chasing all the possible security issues. What will happen is that they will lose motivation, close the PR, and move on to something else.

adnanhemani:
> Let's assume we take 1 week of work per event. We have 150+ events. This effort would take almost 3 years to complete.

I don't think a vast majority of events will require more than a few minutes of verification. The ones that will take a longer time can be temporarily not supported while the community works on them. Sure, it's more than a day's worth of effort to get this to work, but still worth it IMO in comparison to inadvertently introducing known security concerns.

Here's what I can suggest to continue moving things forward:
* A quick screening of payloads to verify there is no security concerns with the payloads being sent. If an event is flagged for potential issues, remove it from being emitted with a TODO for later fix.
* A known-to-work approach (approved by the community) for how do we redact (and/or strategically omit) information that may be nested within the Events payloads, as that will be necessary for the above point.

WDYT?

vchag:
Hey @adutra & @adnanhemani , 

Thanks for sharing your thoughts , both of you raise excellent points. I agree that maintaining strong security while keeping the design clean and generic is key. It feels like this might benefit from a broader discussion to reach a shared consensus.

Since reviewing every event payload could take a fair bit of time, what do you suggest as the best next step for this PR?

adnanhemani:
Hi @vchag - I believe this mailing list thread is particularly relevant to how we want to move forward in the medium- to long-term: https://lists.apache.org/thread/xonxwf9b38t9cxo841r0hn1b34plf7og. I'm still formulating my opinion on this and may require a few days to craft a response (sorry, am very busy the last few days!)

For the short-term, I think we may be at a standstill, unfortunately. I'm still not comfortable with merging this unless the points I listed in my last comment are addressed:
> * A quick screening of payloads to verify there is no security concerns with the payloads being sent. If an event is flagged for potential issues, remove it from being emitted with a TODO for later fix.
> * A known-to-work approach (approved by the community) for how do we redact (and/or strategically omit) information that may be nested within the Events payloads, as that will be necessary for the above point.

I understand it may take some time to review each payload. I know I am making this more difficult on a principled stance, so what I am willing to help with is to help review some of the payloads alongside you, if you are willing to coordinate it (feel free to reach out on the Polaris slack, if that's faster for you :) Apologies for the state we are in, I do definitely empathize with you (I have been in your exact position) - but unfortunately I, personally, cannot overlook the security concerns here.

adutra:
> * remove it from being emitted with a TODO for later fix.

But this PR _also_ introduces event filtering by event type. Granted, by default all events are processed. Wouldn't it be easier to change to default to "no events are processed"?

adnanhemani:
I'm not sure I understand the last comment, but even if we remove all events to merge this, then the following point is still not resolved:
> * A known-to-work approach (approved by the community) for how do we redact (and/or strategically omit) information that may be nested within the Events payloads, as that will be necessary for the above point.

adutra:
> I'm not sure I understand the last comment, but even if we remove all events to merge this, then the following point is still not resolved:
> 
> > * A known-to-work approach (approved by the community) for how do we redact (and/or strategically omit) information that may be nested within the Events payloads, as that will be necessary for the above point.

Sorry for the late reply. If we have all of the events NOT processed by default, then the question of how we redact sensitive data is temporarily solved. We could then merge this PR, and at the same time, we could iterate on the Dev ML about how to do filtering securely.

Wdyt?

adnanhemani:
No worries, I get what you're saying now. Yes, while the issue is temporarily solved, I don't see how this code will fit together in the final version of how we will solve the issue at hand. The question then comes down to: why are we merging code that we are unsure if it will be useful after all? That's why I am insisting on this:

 > A known-to-work approach (approved by the community) for how do we redact (and/or strategically omit) information that may be nested within the Events payloads, as that will be necessary for the above point.

If we have this piece (or be able to reasonably answer why we want to merge this into the codebase regardless of if this will be useful), then I will be satisfied that this contribution will be useful after all and we should proceed with merging it through.

adnanhemani:
As per #3217, I'm satisfied that there is a way forward for the code that would be introduced here. @vchag - can you please look into addressing the review comments and we can move forward with merging this through!

To recap, we should make all events not emitting by default. And then we can review (in new PRs) handful of events at a time to ensure whether a redaction mixin is required and if the current redaction mixins are sufficient for each new event we are emitting.

PR #2965, opened 2025-11-03
NoSQL: Persistence API

snazy:
Provides the persistence API parts for the NoSQL persistence work. Most of this PR are purely interfaces and annotations.

It consists of a low-level `Persistence` interface to read and write individual `Obj`ects and the corresponding pluggable `ObjType`s.

The API module also contains upstream SPIs for database specific implementations and downstream APIs for atomic commits and indexes. Also some CDI specific infrastructure helper annotations.

Unit tests cover the few pieces of actual executable code in this change.

This change also adds a README, which references functionality and modules that are not in this PR, but already provide an overview of the overall interactions.

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -48,6 +48,8 @@ dependencies {
     api(project(":polaris-nodes-impl"))
     api(project(":polaris-nodes-spi"))
 
+    api(project(":polaris-persistence-nosql-api"))
+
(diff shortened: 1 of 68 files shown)


dimas-b:
This PR does not affect any existing Polaris components or runtime. In the interest of making progress on the NoSQL Persistence, I'd like to merge now. If any comments / issues come up, we will address later (on `main`).

PR #2967, opened 2025-11-03
Build/nit: cache output of `generatedMarkdownDocs`

snazy:
`(Java)Exec` tasks are not cacheable by default, as annotated with `@DisableCachingByDefault`. Adding an `outputs.cacheIf { true }` enables caching on those tasks.

### DIFF
--- a/tools/config-docs/site/build.gradle.kts
+++ b/tools/config-docs/site/build.gradle.kts
@@ -53,6 +53,7 @@ val generatedMarkdownDocs = tasks.register<JavaExec>("generatedMarkdownDocs") {
 
   mainClass = "org.apache.polaris.docs.generator.ReferenceConfigDocsGenerator"
 
+  outputs.cacheIf { true }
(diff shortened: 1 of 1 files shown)


tools/config-docs/site/build.gradle.kts
```diff
@@ -53,6 +53,7 @@ val generatedMarkdownDocs = tasks.register<JavaExec>("generatedMarkdownDocs") {
 
   mainClass = "org.apache.polaris.docs.generator.ReferenceConfigDocsGenerator"
 
+  outputs.cacheIf { true }
```
dimas-b, line 56: Just for my education 😅 : even with `cacheIf { true }` the files are still subject to normal source/target modifications checks, right?

dimas-b (approved):

tools/config-docs/site/build.gradle.kts
snazy: correct

PR #2974, opened 2025-11-04
Make StorageAccessConfigProvider request-scoped

XN137:
- add `StorageCredentialsVendor` as request-scoped wrapper around `PolarisCredentialVendor`
- make `FileIOFactory` request-scoped
- make `TaskFileIOSupplier` request-scoped

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
@@ -20,17 +20,17 @@
 package org.apache.polaris.service.catalog.io;
 
 import jakarta.annotation.Nonnull;
-import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.RequestScoped;
 import jakarta.inject.Inject;
 import java.util.Optional;
 import java.util.Set;
 import org.apache.iceberg.catalog.TableIdentifier;
-import org.apache.polaris.core.context.CallContext;
+import org.apache.polaris.core.config.FeatureConfiguration;
 import org.apache.polaris.core.entity.PolarisEntity;
-import org.apache.polaris.core.persistence.MetaStoreManagerFactory;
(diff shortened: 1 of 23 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -91,14 +89,47 @@ public AccessConfig getAccessConfig(
...
+        storageActions.contains(PolarisStorageActions.LIST)
+            || storageActions.contains(PolarisStorageActions.ALL);
+    Set<String> writeLocations =
+        storageActions.contains(PolarisStorageActions.WRITE)
+                || storageActions.contains(PolarisStorageActions.DELETE)
+                || storageActions.contains(PolarisStorageActions.ALL)
+            ? tableLocations
+            : Set.of();
+    AccessConfig accessConfig =
+        storageCredentialCache.getOrGenerateSubScopeCreds(
```
dimas-b, line 116: WDYT about making `AccessConfigProvider` and interface and hiding `StorageCredentialCache` in one of the implementations?

`StorageCredentialCache` will remain application-scoped, of course?

dimas-b (commented):
This PR LGTM as it stands 👍 but I wonder if we could take it one step further (as commented below) 🤔 WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
XN137: i agree with splitting out an interface, however i think we should do it in a follow-up PR as its not directly related to the goal of this PR, wdyt?

dimas-b: Sure. A follow-up PR for this SGTM :+1:

XN137:
rebased after conflict with aa72157a8433c98f1de19b4237d31188cc6c3c83

PR #2977, opened 2025-11-04
NoSQL: database agnostic implementation + in-memory backend

snazy:
This change contains the database agnostic implementation plus the in-memory backend used for testing purposes, and a Junit extension.

These three modules are difficult to put into isolated PRs.

The "main implementation" contains the commit-logic, indexes-logic and the caching part. `PersistenceImplementation` is (more or less) a wrapper providing higher-level functionality backed by a database's `Backend` implementation. The latter provides the bare minimum functionality. Other implementations of the `Persistence` interface are just to transparently add caching and commit-attempt specific case. No call site needs to bother about the actual implementation and/or its layers.

Tests in the `polaris-persistence-nosql-impl` module use the in-memory backend via the Junit extension. Common tests for all backends, in-memory in this PR and MongoDB in a follow-up, are in the testFixtures of the `polaris-persistence-nosql-impl`.


dimas-b (approved):

dimas-b:
This PR is part of the previously discussed NoSQL Persistence impl. (#1189) and does not affect any existing Polaris code. I'm going to merge in a couple of hours. If comments / concerns are raised later we can fix-forward on `main`.

PR #2983, opened 2025-11-05
Fix helm chart error when value cannot be compared to nil

cccs-cat001:
I faced an issue deploying the helm chart with this line. I found that comparing with nil is unnecessary, because you can just do what's proposed. Sometimes an object/string/map is incompatible with nil and the helm chart will throw errors. 

## Checklist

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -63,7 +63,7 @@ flake8 = ">= 4.0.0"
 types-python-dateutil = ">= 2.8.19.14"
 mypy = ">=1.18, <=1.18.2"
 pyiceberg = "==0.10.0"
-pre-commit = "==4.3.0"
+pre-commit = "==4.4.0"
 openapi-generator-cli = "==7.17.0"
 pip-licenses-cli = "==3.0.1"
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -22,7 +22,7 @@ checkstyle = "10.25.0"
 hadoop = "3.4.2"
 hive = "3.1.3"
 iceberg = "1.10.0" # Ensure to update the iceberg version in regtests to keep regtests up-to-date
-quarkus = "3.29.1"
+quarkus = "3.29.2"
 immutables = "2.11.6"
 jmh = "1.37"
 picocli = "4.7.7"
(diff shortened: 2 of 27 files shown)


helm/polaris/templates/_helpers.tpl
```diff
@@ -142,7 +142,7 @@ line breaks, they will be escaped and a multi-line option will be printed.
 {{- $value := index . 1 -}}
 {{- $global := index . 2 -}}
 {{- $valAsString := "" -}}
-{{- if ne $value nil -}}
+{{- if $value -}}
```
adutra: We really want this to be `if ne $value nil`. This is unusual, I know, but here we need to distinguish other zero-values from nil. For example, `foo: false` or `foo: 0` should result in the config property `foo` being included in the ConfigMap, with value `false` or `0`.

If we change this as you suggested, these properties would not be included.

Taking a step back, can you show us a test case that is currently failing? This would help us understand your issue and fix it. Thanks!

cccs-cat001: I really have no idea what values it was failing on. Helm has been pretty useless in telling me where it went wrong. 
```
Error: template: catalog/charts/polaris/templates/configmap.yaml:216:7: executing "catalog/charts/polaris/templates/configmap.yaml" at <include "polaris.appendConfigOption" (list $k $v $global)>: error calling include: template: catalog/charts/polaris/templates/_helpers.tpl:145:7: executing "polaris.appendConfigOption" at <ne $value nil>: error calling ne: incompatible types for comparison
helm.go:81: [debug] template: catalog/charts/polaris/templates/configmap.yaml:216:7: executing "catalog/charts/polaris/templates/configmap.yaml" at <include "polaris.appendConfigOption" (list $k $v $global)>: error calling include: template: catalog/charts/polaris/templates/_helpers.tpl:145:7: executing "polaris.appendConfigOption" at <ne $value nil>: error calling ne: incompatible types for comparison
```

adutra: OK but what was your values file? So that I can reproduce and debug :-)

dimas-b:
@cccs-cat001 you pushed and closed this PR? Was it intentional or by mistake? :)

cccs-cat001:
I am very bad at git, it was definitely a mistake 😆 

Haven't had the time to create a minimal example for this sorry, I'll get to it next week!

cccs-cat001:
I can't seem to reproduce this again. It must've been some config I was using last week that has since changed, or possibly the outdated version of helm that I was using. 

I think we're safe to close this as a user error :P

adutra:
Thanks for confirming that everything is working on your end @cccs-cat001 ! Closing as per your request.

PR #2998, opened 2025-11-06
Event type IDs + event metadata incl. OTel context

adutra:
This PR implements the action items from the following discussion threads:

- https://lists.apache.org/thread/yx7pkgczl6k7bt4k4yzqrrq9gn7gqk2p
- https://lists.apache.org/thread/rl5cpcft16sn5n00mfkmx9ldn3gsqtfy
- https://lists.apache.org/thread/5dpyo0nn2jbnjtkgv0rm1dz8mpt132j9

Summary of changes:

- Introduced a `PolarisEventType` enum holding the 150+ event types.
- Introduced a `PolarisEventMetadata` interface as suggested by @adnanhemani, exposing: timestamp, realm ID, principal, request ID, and OTel context.
- Introduced a `PolarisEventMetadataFactory` to centralize the logic for gathering the various elements of an event metadata.
- Modified `PolarisEvent` to expose 3 new methods:
  - `UUID id()`
  - `PolarisEventType type()`
  - `PolarisEventMetadata metadata()`
- Persistence of OTel context is done in `additional_properties` as suggested by @flyrain.
- Added `InMemoryBufferEventListenerIntegrationTest` to verify that all contextual data is properly persisted.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
@@ -43,7 +43,6 @@ import org.apache.polaris.core.context.RealmContext;
 import org.apache.polaris.core.entity.PolarisPrivilege;
 import org.apache.polaris.service.admin.api.PolarisCatalogsApiService;
 import org.apache.polaris.service.events.CatalogsServiceEvents;
-import org.apache.polaris.service.events.PolarisEvent;
 import org.apache.polaris.service.events.PolarisEventMetadataFactory;
 import org.apache.polaris.service.events.listeners.PolarisEventListener;
 
@@ -60,15 +59,11 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
     polarisEventListener.onBeforeCreateCatalog(
         new CatalogsServiceEvents.BeforeCreateCatalogEvent(
-            PolarisEvent.createEventId(),
-            eventMetadataFactory.create(),
-            request.getCatalog().getName()));
+            eventMetadataFactory.create(), request.getCatalog().getName()));
     Response resp = delegate.createCatalog(request, realmContext, securityContext);
     polarisEventListener.onAfterCreateCatalog(
         new CatalogsServiceEvents.AfterCreateCatalogEvent(
-            PolarisEvent.createEventId(),
-            eventMetadataFactory.create(),
-            (Catalog) resp.getEntity()));
+            eventMetadataFactory.create(), (Catalog) resp.getEntity()));
     return resp;
   }
 
@@ -77,11 +72,11 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
       String catalogName, RealmContext realmContext, SecurityContext securityContext) {
     polarisEventListener.onBeforeDeleteCatalog(
         new CatalogsServiceEvents.BeforeDeleteCatalogEvent(
-            PolarisEvent.createEventId(), eventMetadataFactory.create(), catalogName));
+            eventMetadataFactory.create(), catalogName));
     Response resp = delegate.deleteCatalog(catalogName, realmContext, securityContext);
     polarisEventListener.onAfterDeleteCatalog(
         new CatalogsServiceEvents.AfterDeleteCatalogEvent(
-            PolarisEvent.createEventId(), eventMetadataFactory.create(), catalogName));
+            eventMetadataFactory.create(), catalogName));
     return resp;
   }
 
@@ -90,13 +85,11 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
       String catalogName, RealmContext realmContext, SecurityContext securityContext) {
     polarisEventListener.onBeforeGetCatalog(
         new CatalogsServiceEvents.BeforeGetCatalogEvent(
-            PolarisEvent.createEventId(), eventMetadataFactory.create(), catalogName));
+            eventMetadataFactory.create(), catalogName));
     Response resp = delegate.getCatalog(catalogName, realmContext, securityContext);
     polarisEventListener.onAfterGetCatalog(
         new CatalogsServiceEvents.AfterGetCatalogEvent(
-            PolarisEvent.createEventId(),
-            eventMetadataFactory.create(),
-            (Catalog) resp.getEntity()));
+            eventMetadataFactory.create(), (Catalog) resp.getEntity()));
     return resp;
   }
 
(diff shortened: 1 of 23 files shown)


adutra:
\cc @flyrain @adnanhemani @dimas-b

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -51,37 +53,50 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
...
   @Inject PolarisEventListener polarisEventListener;
+  @Inject PolarisEventMetadataFactory eventMetadataFactory;
 
   @Override
   public Response createCatalog(
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
     polarisEventListener.onBeforeCreateCatalog(
-        new CatalogsServiceEvents.BeforeCreateCatalogEvent(request.getCatalog().getName()));
+        new CatalogsServiceEvents.BeforeCreateCatalogEvent(
+            PolarisEvent.createEventId(),
```
adnanhemani: Wanted to see your thoughts on this: can we collapse Event ID into the EventMetadata as well? I think it makes sense as part of the "metadata" of the event and may help us save many lines of code uniformly across all events.

adutra: I thought about that too, but after some hesitation, I think an ID is a bit more than "metadata". The fact that this specific field must be unique makes it imho a bit special. Wdyt?

That said I don't have strong opinions on this, and I'd be OK to move it to metadata if there is a consensus to do so.

adnanhemani: I understand your point, it is definitely a bit of a "grey area." I think the amount of cleanliness that we gain from moving it into EventMetadata outweighs the ambiguities here, though. I'm also willing to sway my opinion with the will of the community, but I would vote for merging it into EventMetadata if others feel similarly :)

adnanhemani:
@adutra - I see you force-pushed the last commit so it's rewritten the commit history. I'm assuming it was for rebasing this PR onto main? Was there any other major change? Since it's a larger change, I fear I may have missed something.

adutra:
> @adutra - I see you force-pushed the last commit so it's rewritten the commit history. I'm assuming it was for rebasing this PR onto main? Was there any other major change? Since it's a larger change, I fear I may have missed something.

I rebased because of conflicts - no functional change.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adutra: The more I think about this the more I think we shouldn't move the id to the metadata. Here is what the `PolarisEventMetadata` would look like:

```java
@PolarisImmutable
public interface PolarisEventMetadata {
  UUID id();
  // other methods omitted for brevity
}
```

But that creates an ambiguity: is `id()` the identifier of the metadata object, or of the event object?

I could rename it to `eventId()` of course, but it still doesn't feel right to me. An identifier should be attached to the object it identifies.

dimas-b:
@adutra : there are more conflicts :shrug:

adutra:
> @adutra : there are more conflicts 🤷

re-rebased!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adnanhemani: I'd say it would be named `eventId()` in this model. IMO, the rest of the fields in `PolarisEventMetadata` are also vital parts of the event's identity - but I understand the argument philosophically. It's ultimately your call, but I still don't see the harm of doing this versus the benefits are the conciseness of the code.

adutra: Fair enough, I moved the id to `PolarisEventMetadata`. Please take another look at the latest commit [881a477](https://github.com/apache/polaris/pull/2998/commits/881a4778e44411960d78178a9e7ea500753b3269)

adutra:
FYI: re-rebased because of more conflicts.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadata.java
```diff
@@ -0,0 +1,59 @@
...
+@PolarisImmutable
+public interface PolarisEventMetadata {
+
+  static ImmutablePolarisEventMetadata.Builder builder() {
+    return ImmutablePolarisEventMetadata.builder();
+  }
+
+  /** The unique ID of the event. */
+  default UUID eventId() {
+    return UUID.randomUUID();
```
dimas-b, line 39: Same concern as for request IDs, if it is used for each event (likely multiple per request), it might be too slow / too much affected by the system source of randomness... but we can address this later if other people think it's a concern.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Extracts the realm ID from the current realm context.
+   *
+   * <p>In the unlikely case there is no realm context, e.g. in some tests, returns
+   * "UNKNOWN_REALM_ID".
+   */
+  private String getRealmId() {
+    return realmContext.isResolvable()
```
dimas-b, line 82: Why are we not sure that realm context is always present?

runtime/service/build.gradle.kts
```diff
@@ -116,6 +116,7 @@ dependencies {
   }
 
   testImplementation(project(":polaris-api-management-model"))
+  testImplementation(project(":polaris-relational-jdbc"))
```
dimas-b, line 119: I might have missing this... How does JDBC get into tests now ? 😅

dimas-b (approved):

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
adutra: In some tests, events are generated on the "client side", in which case, there is no realm context. (In production code there is always a ream context though.)

runtime/service/build.gradle.kts
adutra: To test the `persistence-in-memory-buffer` event listener with H2: see `InMemoryBufferEventListenerIntegrationTest`.

dimas-b: thx!

PR #3005, opened 2025-11-07
made the regex more generic

cccs-cat001:
We found a use case where the AWS ARN regex pattern was too specific and caused us issues connecting to our private S3 storage that's not on AWS.  This fixes that issue. 
## Checklist

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/entity/CatalogEntityTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/entity/CatalogEntityTest.java
@@ -255,8 +255,7 @@ public class CatalogEntityTest {
   }
 
   @ParameterizedTest
-  @ValueSource(
-      strings = {"", ":aws:iam:0123456:role/jdoe", "arn:aws:iam:0123456:role/jdoe", "aws-cn"})
+  @ValueSource(strings = {"", "arn:aws:iam:0123456:role/jdoe", "aws-cn"})
   public void testInvalidArn(String roleArn) {
     String basedLocation = "s3://externally-owned-bucket";
     AwsStorageConfigInfo awsStorageConfigModel =
@@ -289,9 +288,9 @@ public class CatalogEntityTest {
   @ParameterizedTest
   @ValueSource(
       strings = {
-        "",
         "arn:aws:iam::012345678911:role/rollerblade",
-        "test:test:iam:region:accountid:role/rollerblade"
+        "test:test:iam:region:accountid:role/rollerblade",
+        "a::iam:::role/rollerblade"
       })
   public void testValidArn(String roleArn) {
     String basedLocation = "s3://externally-owned-bucket";


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -46,7 +46,7 @@ public static ImmutableAwsStorageConfigurationInfo.Builder builder() {
 
   // Technically, it should be ^arn:(aws|aws-cn|aws-us-gov):iam::(\d{12}):role/.+$,
   @JsonIgnore
-  public static final String ROLE_ARN_PATTERN = "^arn:(aws|aws-us-gov):iam::(\\d{12}):role/.+$";
+  public static final String ROLE_ARN_PATTERN = "^.+:(.*):iam:.*:(.*):role/.+$";
```
dimas-b: nit: should we keep the `arn` prefix? In the end the pattern is called an "ARN" pattern 🤔

cccs-cat001: arn was part of the problem 😅  because it's not amazon, it doesn't start with `arn`

dimas-b: ah, fair point... maybe rename the constant to something like `ROLE_PATTERN` then?

dimas-b: actually, the constant name is probably ok as it eventually relates to AWS SDK terminology... but please add / update the comment to mention that it applies to non-AWS systems too.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
```diff
@@ -44,9 +44,9 @@ public static ImmutableAwsStorageConfigurationInfo.Builder builder() {
     return ImmutableAwsStorageConfigurationInfo.builder();
   }
 
-  // Technically, it should be ^arn:(aws|aws-cn|aws-us-gov):iam::(\d{12}):role/.+$,
-  @JsonIgnore
-  public static final String ROLE_ARN_PATTERN = "^arn:(aws|aws-us-gov):iam::(\\d{12}):role/.+$";
+  // Technically, it should be ^arn:(aws|aws-cn|aws-us-gov):iam::(\d{12}):role/.+$, but we've
+  // generalized it to support non-aws S3 implementations
+  @JsonIgnore public static final String ROLE_ARN_PATTERN = "^.+:(.*):iam:.*:(.*):role/.+$";
```
flyrain, line 49: Can we update the changelog.md as this is a behavior change?

runtime/service/src/test/java/org/apache/polaris/service/entity/CatalogEntityTest.java
flyrain: Do we still need it?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfo.java
dimas-b: yes, having a CHANGELOG entry under `Changes` will not hurt.

runtime/service/src/test/java/org/apache/polaris/service/entity/CatalogEntityTest.java
dimas-b: I'd prefer to add a test case that fails the regex-based check.

cccs-cat001: actually, I have a good test to replace it with. Just removing one of the colons should result in an invalid ARN

runtime/service/src/test/java/org/apache/polaris/service/entity/CatalogEntityTest.java
```diff
@@ -255,7 +255,7 @@ public void testValidAllowedLocationPrefix() {
   }
 
   @ParameterizedTest
-  @ValueSource(strings = {"", "arn:aws:iam::0123456:role/jdoe", "aws-cn"})
+  @ValueSource(strings = {"", "arn:aws:iam:0123456:role/jdoe", "aws-cn"})
```
flyrain: Can we have another test for a valid non-aws role string which can pass the validation? So that, any regression will be caught if the regex change happens.

flyrain (approved):

flyrain:
Thanks a lot for the change, @cccs-cat001 ! Thanks a lot for the review and dev mailing discussion, @dimas-b !

PR #3007, opened 2025-11-07
NoSQL: Realms handling

snazy:
Introduces handling for realms including realm-state management/transition.

The `RealmStore` implementation for NoSQL depends on CDI components, coming in a follo-up PR.

### DIFF
--- a/persistence/nosql/realms/impl/src/main/resources/META-INF/beans.xml
+++ b/persistence/nosql/realms/impl/src/main/resources/META-INF/beans.xml
@@ -1,0 +1,24 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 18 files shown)


persistence/nosql/realms/impl/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd">
+    <!-- File required by Weld (used for testing), not by Quarkus -->
```
dimas-b, line 23: sorry, I keep forgetting... why are these test resources in the `main` src tree? 😅

snazy: B/c a `beans.xml` needs to be in every classpath element (jar/folder) that shall be scanned for CDI stuff.
Quarkus doesn't require this, but Weld (CDI reference impl, used for faster testing purposes) does.

dimas-b: Ah, so the `beans.xml` file has to be located in the same artifact as the beans it references... I see.

dimas-b:
This is a well-isolated change. It does not affect any other Polaris code paths and it has been in review since Fri. I'll be merging today unless there are new comments.

Issue #3017, opened 2025-11-09
Can't add additional backend services to Helm chart ingress


MingzhouLin:
### Describe the bug

The current ingress only allows to expose the Polaris service, since the `backend.service.name` is "hardcoded" as the service name and port is set as the polaris service port. Seen the code snippet below
```
    - host: {{ .host | quote }}
      http:
        paths:
          {{- range .paths }}
          - path: {{ .path }}
            pathType: {{ .pathType }}
            backend:
              service:
                name: {{ $fullName }}
                port:
                  number: {{ $svcPort }}
```
This prevent user from exposing other backend services like the management service. We might want to allow user to expose additional services besides the core service.

### To Reproduce

I want to configure the value.yaml to expose both polaris service and management service
```
ingress:
  className: "somename"
  enabled: true
  hosts:
    - host: chart-example.local
      paths:
        - path: /polaris
           pathType: Prefix
           backend:
                service:
                    name: polaris-release
                    port: 8081
        - path: /polaris-mgmt
           pathType: Prefix
           backend:
                service:
                    name: polaris-release-mgmt
                    port: 8082
```

### Actual Behavior

It would result as
```
  - host: chart-example.local
    http:
      paths:
        - path: /polaris
          pathType: Prefix
          backend:
            service:
              name: polaris-release
              port:
                number: 8181
        - path: /polaris-mgmt
          pathType: Prefix
          backend:
            service:
              name: polaris-release
              port:
                number: 8181
```

### Expected Behavior

```
  - host: chart-example.local
    http:
      paths:
        - path: /polaris
          pathType: Prefix
          backend:
            service:
              name: polaris-release
              port:
                number: 8181
        - path: /polaris-mgmt
          pathType: Prefix
          backend:
            service:
              name: polaris-release-mgmt
              port:
                number: 8182
```

### System information

OS: MacOS
Polaris version: 1.2.0
Storage: S3

MingzhouLin:
Update the ingress as this to allow me expose extra services. Happy to raise a PR for this.
```
  rules:
    {{- range .Values.ingress.hosts }}
    - host: {{ .host | quote }}
      http:
        paths:
          {{- range .paths }}
          - path: {{ .path }}
            pathType: {{ .pathType }}
            backend:
              service:
                name: {{ default $fullName .backend.service.name }}
                port:
                  number: {{ default $svcPort .backend.service.port }}
          {{- end }}
    {{- end }}
  {{- end }}
```

[...]

PR #3018, opened 2025-11-09
fix(helm):Allow user to expose extra backend services in ingress

MingzhouLin:
Fixes: #3017

Background: I want to expose extra services(management service for example) through ingress, but there is no way to do so by only editing the configuration.

So, I add this PR to allow user add services through config file.

## Checklist
- N/A 💡 Added comments for complex logic
- N/A 🧾 Updated `CHANGELOG.md` (if needed)
- N/A 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

### DIFF
--- a/helm/polaris/templates/ingress.yaml
+++ b/helm/polaris/templates/ingress.yaml
@@ -50,14 +50,6 @@ spec:
     - host: {{ .host | quote }}
       http:
         paths:
-          {{- range .paths }}
-          - path: {{ .path }}
-            pathType: {{ .pathType }}
-            backend:
-              service:
-                name: {{ default $fullName .backend.service.name }}
-                port:
-                  number: {{ default $svcPort .backend.service.port }}
-          {{- end }}
+          {{ tpl (toYaml .paths) $ | nindent 10 }}
     {{- end }}
   {{- end }}
--- a/helm/polaris/tests/ingress_test.yaml
+++ b/helm/polaris/tests/ingress_test.yaml
@@ -163,13 +163,15 @@ tests:
               backend:
                 service:
                   name: polaris
-                  port: 8181
-            - path: /polaris-mgmt
+                  port:
+                    number: 8181
+            - path: /polaris-other-service
               pathType:  Prefix
               backend:
                 service:
-                  name: polaris-mgmt
-                  port: 8182
+                  name: polaris-other-service
+                  port:
+                    number: 8182
     asserts:
       - equal:
           path: spec.rules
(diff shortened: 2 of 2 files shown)


snazy:
See https://github.com/apache/polaris/issues/3017#issuecomment-3510730141

helm/polaris/templates/ingress.yaml
```diff
@@ -55,9 +55,9 @@ spec:
             pathType: {{ .pathType }}
             backend:
               service:
-                name: {{ $fullName }}
+                name: {{ default $fullName .backend.service.name }}
                 port:
-                  number: {{ $svcPort }}
+                  number: {{ default $svcPort .backend.service.port }}
           {{- end }}
```
adutra: While at it, shouldn't we just copy the entire `.paths` here?

```suggestion
          {{ tpl (toYaml .paths) $ | nindent 10 }}
```

adutra:
@MingzhouLin I think #3017 is quite debatable, but I wouldn't mind repurposing this PR to allow people to enter the whole `path` object rather than just portions of it.

helm/polaris/templates/ingress.yaml
MingzhouLin: Thanks! updated

helm/polaris/templates/ingress.yaml
```diff
@@ -50,14 +50,6 @@ spec:
...
-          {{- range .paths }}
-          - path: {{ .path }}
-            pathType: {{ .pathType }}
-            backend:
-              service:
-                name: {{ $fullName }}
-                port:
-                  number: {{ $svcPort }}
-          {{- end }}
+          {{ tpl (toYaml .paths) $ | nindent 10 }}
```
adutra, line 53: @MingzhouLin I think I didn't really think about the consequences of my suggestion, but looking at the result, it seems to me that it's not fair to ask users to input the whole `path` object, _if they don't need to_.

One of the issues is that now users must input the service name (`backend.service.name`) and port (`backend.service.port`) – but these fields have good default values.

Would you be OK for you if we do this instead:

```yaml
          {{- range .paths }}
          - path: {{ .path }}
            pathType: {{ .pathType }}
            {{- if .backend }}
            backend:
              {{- tpl (toYaml .backend) $ | nindent 14 }}
            {{- else }}
            backend:
              service:
                name: {{ $fullName }}
                port:
                  number: {{ $svcPort }}
            {{- end }}
          {{- end }}
```

In the above example, if you customize `backend` it will be used, otherwise a default `backend` is generated.

Sorry for the back and forth 😬

PR #3023, opened 2025-11-10
Helm chart: include configmap checksum in deployment annotations

cccs-cat001:
1.) I added a checksum to the deployment so that any time the configmap changes the deployment restarts (so it can pick up the new configmap changes)

* If I'm going to be making more PRs I realize I need to figure out how to not have git showing commits from other changes 😅 
## Checklist

### DIFF
--- a/helm/polaris/tests/deployment_test.yaml
+++ b/helm/polaris/tests/deployment_test.yaml
@@ -181,11 +181,11 @@ tests:
   - it: should only set checksum annotation by default
     template: deployment.yaml
     asserts:
-      - lengthEqual:
-          path: spec.template.metadata.annotations
-          count: 1
       - exists:
           path: spec.template.metadata.annotations.checksum/config
+      - matchRegex: 
+          path: spec.template.metadata.annotations.checksum/config
+          pattern: "^[a-f0-9]{64}$"
   - it: should set pod annotations
     template: deployment.yaml
     set:


cccs-cat001:
I think the `helm test` failures is because of the version... it deploys fine with helm 3.18.4 and the tests are using helm 3.16...

adutra:
Hi @cccs-cat001 I fully agree with the configmap checksum, but for the bootstrap job, we decided a while ago that it's not a good idea:

https://github.com/apache/polaris/pull/626#discussion_r1937641536

Do you mind reshaping this PR to focus on the checksum only? 

For the bootstrap job, if you really think you need one, let's discuss a bit about it before.

cccs-cat001:
okay! Maybe my understanding of the bootstrap is lacking, it seemed to me more like a "one and done" type thing, not a "this could destroy you" thing 😅 I'll reconfigure the PR and I'll start a discussion in slack for this? 

As for the `helm tests` is there any issue with updating the workflow to use a newer version of helm?

adutra:
> okay! Maybe my understanding of the bootstrap is lacking, it seemed to me more like a "one and done" type thing, not a "this could destroy you" thing 😅 I'll reconfigure the PR and I'll start a discussion in slack for this?

I think the problem is also that there is currently no option to ignore an already-bootstrapped realm. So your job will fail if the realm is already bootstrapped, which means that your chart installation would fail as well.

> As for the `helm tests` is there any issue with updating the workflow to use a newer version of helm?

You mean, the Helm version used by GitHub CI? No problems imho.

PR #3024, opened 2025-11-10
Embrace request-scoped TokenBroker

XN137:
`TokenBroker` and `CallContext` are both request-scoped, so instead of passing the former into the latter, we can do this via the `TokenBrokerFactory` and thus simplify the `TokenBroker` interface.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/TokenBrokerFactory.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/TokenBrokerFactory.java
@@ -19,7 +19,6 @@
 package org.apache.polaris.service.auth.internal.broker;
 
 import org.apache.polaris.core.PolarisCallContext;
-import org.apache.polaris.core.context.RealmContext;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 
 /**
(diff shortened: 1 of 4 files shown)


runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/TokenBrokerFactory.java
```diff
@@ -18,11 +18,17 @@
...
 /**
  * Factory that creates a {@link TokenBroker} for generating and parsing. The {@link TokenBroker} is
  * created based on the realm context.
  */
-public interface TokenBrokerFactory extends Function<RealmContext, TokenBroker> {}
+public interface TokenBrokerFactory {
+  TokenBroker create(
+      RealmContext realmContext,
+      PolarisMetaStoreManager metaStoreManager,
+      PolarisCallContext polarisCallContext);
```
adutra: `PolarisCallContext` already exposes `RealmContext`, why do we need both parameters? Could we pass `CallContext` instead?

XN137: you are right that it was redundant, added a commit to address it.

dimas-b (approved):
This looks like a nice refactoring to me 👍 Thanks, @XN137 !

PR #3027, opened 2025-11-11
NoSQL: Add correctness tests

snazy:
Verifies the correctness of concurrent commits, and big index handling.

These tests are intentionally _not_ part of the base-backend test suite for two reasons:
1. These tests do not run against the `Backend` interface but the `Persistence` interface, including commit and index logic.
2. These tests are intended to be runnable against a custom provisioned database cluster, not just tiny-ish test containers.


dimas-b:
Isolated change. I propose merging tomorrow.

PR #3028, opened 2025-11-11
NoSQL: Add maintenance API, SPI

snazy:
Maintenance operations include a bunch of tasks that are regularly executed against a backend database.

Types of maintenance operations include:
* Purging unreferenced database objects and references
* Purging whole realms that are marked to be purged

Implementation added in a follow-up PR.


dimas-b:
Isolated change. I propose merging tomorrow.

PR #3041, opened 2025-11-13
Test-fix: Cleanup OPA test container on stop

snazy:
Quarkus takes care of reusing a test-resource across tests. The current behavior leaves the container around.

Plus some nit-fixes (deprecation + local var)


singhpk234 (approved):
LGTM, thanks @snazy

adutra (approved):

PR #3042, opened 2025-11-13
OPA: Tackle deprecation warnings

snazy:
Instead of suppressing the deprecations, this change updates the code a little bit to remove the mocks (except to create a non-nullable parameter).

### DIFF
--- a/extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
+++ b/extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
@@ -21,6 +21,7 @@ package org.apache.polaris.extension.auth.opa;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.annotations.VisibleForTesting;
 import jakarta.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


snazy:
@sungwy mind taking a look?

adutra (dismissed):

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
```diff
@@ -174,27 +177,35 @@ private boolean queryOpa(
...
-        return respNode.path("result").path("allow").asBoolean(false);
-      }
-    } catch (IOException e) {
+      return httpClientExecute(httpPost, this::queryOpaCheckResponse);
+    } catch (HttpException | IOException e) {
       throw new RuntimeException("OPA query failed", e);
     }
   }
 
+  <T> T httpClientExecute(
```
sungwy, line 188:
```suggestion
  @VisibleForTesting
  <T> T httpClientExecute(
```

It looks like we are following the convention of marking it as being visible for testing, if that's our intent https://github.com/apache/polaris/blob/286e77d5e3b6c0a9de051928471d0f9e3ec2d453/runtime/service/src/main/java/org/apache/polaris/service/storage/aws/StsClientsPool.java#L59-L70

What are your thoughts on adding it here?

snazy: Yup, that one should be there.

PR #3056, opened 2025-11-14
Skip Hugo Site workflow on forks

XN137:
forks usually don't have the "versioned-docs" tag and thus PRs
against forks or rebasing the main branch on a fork currently
always causes workflow failures.


XN137:
note that the "Hugo Site" workflow in this PR of course still gets executed normally, as it targets the right repository

snazy (approved):

singhpk234 (approved):

PR #3064, opened 2025-11-17
Core: resolveAll() must be called before reading resolution results

flyrain:
To fail fast and explicitly when resolveAll() isn't called before reading the results

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -68,6 +68,14 @@ public class PolarisResolutionManifest implements PolarisResolutionManifestCatal
   // Set when resolveAll is called
   private ResolverStatus primaryResolverStatus = null;
 
+  private ResolverStatus requirePrimaryResolverStatus() {
```
singhpk234: minor : require gives an impression that its just doing assertion, but its additionally returning status how about this 

```suggestion
  private ResolverStatus safeGetPrimaryResolverStatus() {
```

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -277,6 +285,7 @@ public PolarisResolvedPathWrapper getResolvedRootContainerEntityAsPath() {
 
   public PolarisResolvedPathWrapper getResolvedReferenceCatalogEntity(
       boolean prependRootContainer) {
+    requirePrimaryResolverStatus();
```
singhpk234: [doubt] is just checking status not null sufficient ? or should we bake the assertion that that status != null and status = SUCCESS ?

singhpk234 (commented):
Thanks for fix @flyrain !

flyrain:
> @flyrain : could you explain why "resolveAll() _must_ be called before reading resolution results"? The PR description provides information about enforcing it, but the rationale for doing that is not very clear... at least to me 😅 Thx!

The result will be invalid, or null if it's not invoked. The concurrent code actually already enforce that. You can find the following code in each place reading the result, but NPE will be threw in case of `resovleAll()` not called. This PR is mainly to make it fail explicitly instead of a NPE.
```
if (primaryResolverStatus.getStatus() != ResolverStatus.StatusEnum.SUCCESS) {
   return null
}
```

dimas-b (dismissed):

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
flyrain: Refactored a bit. Thanks for the suggestion.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
flyrain: That's a good point. Removed this line to avoid any behavior change.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
dimas-b: Thx! The new code is much nicer :slightly_smiling_face:

singhpk234 (approved):
LGTM thanks @flyrain !

PR #3085, opened 2025-11-19
Publish/pom: don't include test-fixtures dependencies as `runtime` scope

snazy:
The list of dependencies in pom's includes the api/runtime elements of the test-fixtures, which is not what should be published, as it "pulls up" deps like junit, mockito and assertj as Maven runtime scope dependencies. This change fixes this.

### DIFF
--- a/build-logic/src/main/kotlin/publishing/PublishingHelperPlugin.kt
+++ b/build-logic/src/main/kotlin/publishing/PublishingHelperPlugin.kt
@@ -20,6 +20,7 @@ import com.github.jengelman.gradle.plugins.shadow.ShadowPlugin
 import javax.inject.Inject
 import org.gradle.api.Plugin
 import org.gradle.api.Project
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/publishing/PublishingHelperPlugin.kt
```diff
@@ -125,7 +126,17 @@ constructor(private val softwareComponentFactory: SoftwareComponentFactory) : Pl
...
+                      ->
+                      configurations.findByName(cfg)?.apply {
+                        component.addVariantsFromConfiguration(this) { skip() }
+                      }
+                    }
+                  }
+                  from(component)
                 }
 
                 suppressPomMetadataWarningsFor("testFixturesApiElements")
```
adutra, line 142: Do we still need to suppress the corresponding warnings?

Issue #3086, opened 2025-11-19
Releasy: make Helm package reproducible


snazy:
### Is your feature request related to a problem? Please describe.

`helm package` produces a non-reproducible tarball. The archive entries' timestamps are _always_ set to the current timestamp, aka the `helm package` invocation timestamp.

There is sadly no way to pass tar or gzip options to `helm package`.

For Polaris releases, we need a _signed_ Helm package, producing a `.prov`  file, which contains more information than "just" the cryptographic signature ([example contents here](https://github.com/snazy/helm-gpg/actions/runs/19240703985/job/55002364031#step:7:8)). Having said that, it's not sufficient to "just" replace `helm package`  with a manual `tar`+`gzip`+`gpg` command chain.

snazy:
FTR: This requires Helm 4 in GH workflow runners.

MonkeyCanCode:
This seems to be fixed now. Should we close this @snazy ?

snazy:
Not yet. Helm 4 isn't there yet.

MonkeyCanCode:
> Not yet. Helm 4 isn't there yet.

Seems to be released: https://helm.sh/blog/helm-4-released/

MonkeyCanCode:
@snazy should we close this one as the current GH action is pulling in v4.0.4: https://github.com/apache/polaris/actions/runs/20854821902/job/59920537649

snazy:
Let's check when the next RC is there. There was something with v4, I just don't recall what.

PR #3088, opened 2025-11-19
Releasy: prepare for Helm 4 (helm package repro)

snazy:
Part of #3086


snazy:
> In #3086 you added a link to [helm/helm#31323](https://github.com/helm/helm/pull/31323). IIUC that link says that the next Helm version will use file modification time as archive creation time. Which means that for a given Git SHA, building the chart multiple times will yield the same result. There should be no need to tweak the file modification time.

The file modification timestamps after a `git clone` are the timestamps at which the file was created.

PR #2871, opened 2025-10-23
Refactor Polaris cloudTests and adopt changes from PR #2405

tmater:
## Summary
This patch picks up changes from [apache/polaris#2405](https://github.com/apache/polaris/pull/2405).  
It includes only the refactoring portions, as there have been updates to the `createViewWithCustomMetadataLocationUsingPolaris` and `createViewWithCustomMetadataLocation` tests that require further investigation.

## Details
Integrated relevant refactors from upstream PR #2405, renaming classes, environment variables and making Base classes abstract.

## Known Issues
The following tests are still failing in `cloudTest`:
- `createViewWithCustomMetadataLocationUsingPolaris`
- `createViewWithCustomMetadataLocation`

These will be investigated and fixed in a follow-up PR.

## Testing
- Executed `cloudTests` locally to validate refactor integrity.

PR #3095, opened 2025-11-19
Disable cloud storage tests that would use @TempDir

tmater:
## Summary

Disables 5 view catalog tests for cloud storage providers (S3, ADLS, GCS) that use `@TempDir`. The `ViewCatalogTests` base class uses `Paths.get()` with the temp directory, which doesn't support cloud storage paths.

Follow-up to https://github.com/apache/polaris/pull/2871.

## Background

These tests inherit from Iceberg's `ViewCatalogTests`, which passes a `@TempDir` path to `Paths.get()`. Since `Paths.get()` only accepts local filesystem paths, it cannot handle cloud storage URIs. For example, when `@TempDir` provides `/tmp/junit123`, `Paths.get()` converts it to a URI like `file:///tmp/junit123`. But for cloud storage base paths like `s3://bucket/path`, it would incorrectly create `s3:/bucket/path` (single slash) instead of the valid `s3://bucket/path` format.

These tests are disabled to prevent false positives - they would pass by testing against local filesystem instead of actually validating cloud storage behavior.

## Changes

**Base test classes** - Added `@Disabled` overrides for 5 tests:
- `PolarisRestCatalogViewAdlsIntegrationTestBase.java`
- `PolarisRestCatalogViewGcsIntegrationTestBase.java`
- `PolarisRestCatalogViewS3IntegrationTestBase.java`

**Concrete test classes** - Removed reflection workaround:
- `RestCatalogViewAdlsIT.java`
- `RestCatalogViewGcsIT.java`
- `RestCatalogViewS3IT.java`

## Testing

Run cloud tests: `./gradlew cloudTest`

## Future Work

Will follow up with changes to Iceberg's `ViewCatalogTests` to avoid using `Paths.get()` with `@TempDir`. Once that's done, these tests can be re-enabled.

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogViewAdlsIntegrationTestBase.java
@@ -19,10 +19,13 @@
 package org.apache.polaris.service.it.test;
 
 import com.google.common.base.Strings;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.stream.Stream;
 import org.apache.polaris.core.admin.model.AzureStorageConfigInfo;
 import org.apache.polaris.core.admin.model.StorageConfigInfo;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.io.TempDir;
 
 /** Runs PolarisRestCatalogViewIntegrationTest on Azure. */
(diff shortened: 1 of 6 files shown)


tmater:
Thank you for the quick review, @dimas-b and @snazy!

@dimas-b, I’ve already started experimenting with some changes. I don’t expect them to be too disruptive, but we’ll also need to bump our Iceberg version for that. I figured it’s better to disable these for now.

dimas-b:
Yes, I'm sure an Iceberg change will be needed to make the test reusable in cloud environments.

PR #3102, opened 2025-11-20
PRs: Remove markdown-links-check from required checks

snazy:
The "Check Markdown links" workflow is known to produce false failures, leading to #3097.

This change is intended to unblock PRs due to these false failures.

PR #3103, opened 2025-11-20
PRs: Re-add markdown-links-check step

snazy:
The step was disabled in #3102 to pass CI and enable merging.


snazy:
> I don't think we should re-enable those checks. They are flaky. And usually, people develop a habit of ignoring those checks all the time. I would rather have them run manually once in a while (possibly integrated to the release workflows) but that's it.

Yea, I agree. Let's bring it up on the dev-ML.

PR #3113, opened 2025-11-21
Add timeout and retry logic to Azure token fetch

fivetran-rahulprakash:
### Problem
The `getAccessToken()` method in `AzureCredentialsStorageIntegration` used an unbounded blocking call which could hang indefinitely if Azure's token endpoint was slow or unresponsive. This could lead to:
- Thread pool exhaustion in high-concurrency scenarios
- Cascading failures when Azure AD experiences degraded performance
- Poor user experience with no visibility into token fetch failures

### Solution
This PR adds defensive timeout and retry mechanisms using Project Reactor's built-in capabilities:

- **15-second timeout** per individual token request attempt to prevent indefinite blocking
- **Exponential backoff retry** (3 attempts with delays: 2s, 4s, 8s) with 50% jitter to prevent thundering herd during mass failures
- **90-second overall timeout** as a safety net for the complete operation
- **Intelligent retry filtering** for known transient Azure AD errors:
  - `AADSTS50058` - Token endpoint timeout
  - `AADSTS50078` - Service temporarily unavailable  
  - `AADSTS700084` - Token refresh required
  - `503` - Service unavailable
  - `429` - Too many requests
- **Enhanced logging** for better observability (warnings on errors, info on retries)

### Testing
- Code leverages existing reactor dependencies (no new dependencies)
- Follows existing Polaris patterns for reactive error handling

### Benefits
- Improves system resilience to transient Azure service issues
- Prevents indefinite blocking that could cascade to request timeouts
- Provides better observability with structured logging
- Uses well-established retry patterns with exponential backoff and jitter

## Checklist
- [x] 🛡️ Don't disclose security issues! (Not applicable - this is a resilience improvement)
- [x] 🧪 Manually tested via compilation; reactive behavior follows reactor-core semantics
- [x] 💡 Added comprehensive Javadoc explaining the retry strategy
- [ ] 🧾 Updated `CHANGELOG.md` (awaiting maintainer guidance on format)
- [ ] 📚 Updated documentation (no user-facing config changes)

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -439,46 +439,42 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
           .defaultValue(true)
           .buildFeatureConfiguration();
 
-  public static final FeatureConfiguration<Integer> STORAGE_API_TIMEOUT_MILLIS =
+  public static final FeatureConfiguration<Integer> AZURE_TIMEOUT_MILLIS =
       PolarisConfiguration.<Integer>builder()
-          .key("STORAGE_API_TIMEOUT_MILLIS")
+          .key("AZURE_TIMEOUT_MILLIS")
           .description(
-              "Timeout in milliseconds for cloud storage provider API requests. "
-                  + "Prevents indefinite blocking when cloud provider endpoints are slow or unresponsive. "
-                  + "Used internally by storage integrations for credential vending and other cloud operations. "
-                  + "Currently only used by Azure storage integration. Designed to be applicable to AWS S3, GCP, and other cloud storage providers.")
+              "Timeout in milliseconds for Azure API requests. "
+                  + "Prevents indefinite blocking when Azure endpoints are slow or unresponsive. "
+                  + "Used internally by Azure storage integration for credential vending and other operations.")
           .defaultValue(15000)
           .buildFeatureConfiguration();
 
-  public static final FeatureConfiguration<Integer> STORAGE_API_RETRY_COUNT =
+  public static final FeatureConfiguration<Integer> AZURE_RETRY_COUNT =
       PolarisConfiguration.<Integer>builder()
-          .key("STORAGE_API_RETRY_COUNT")
+          .key("AZURE_RETRY_COUNT")
           .description(
-              "Number of retry attempts for cloud storage provider API requests. "
-                  + "Uses exponential backoff with jitter to handle transient failures. "
-                  + "Currently only used by Azure storage integration. Designed to be applicable to AWS S3, GCP, and other cloud storage providers.")
+              "Number of retry attempts for Azure API requests. "
+                  + "Uses exponential backoff with jitter to handle transient failures.")
           .defaultValue(3)
           .buildFeatureConfiguration();
 
-  public static final FeatureConfiguration<Integer> STORAGE_API_RETRY_DELAY_MILLIS =
+  public static final FeatureConfiguration<Integer> AZURE_RETRY_DELAY_MILLIS =
       PolarisConfiguration.<Integer>builder()
-          .key("STORAGE_API_RETRY_DELAY_MILLIS")
+          .key("AZURE_RETRY_DELAY_MILLIS")
           .description(
-              "Initial delay in milliseconds before first retry for cloud storage provider API requests. "
-                  + "Delay doubles with each retry (exponential backoff). "
-                  + "Currently only used by Azure storage integration. Designed to be applicable to AWS S3, GCP, and other cloud storage providers.")
+              "Initial delay in milliseconds before first retry for Azure API requests. "
+                  + "Delay doubles with each retry (exponential backoff).")
           .defaultValue(2000)
           .buildFeatureConfiguration();
 
-  public static final FeatureConfiguration<Double> STORAGE_API_RETRY_JITTER_FACTOR =
+  public static final FeatureConfiguration<Double> AZURE_RETRY_JITTER_FACTOR =
       PolarisConfiguration.<Double>builder()
-          .key("STORAGE_API_RETRY_JITTER_FACTOR")
+          .key("AZURE_RETRY_JITTER_FACTOR")
           .description(
-              "Jitter factor (0.0 to 1.0) applied to retry delays for cloud storage provider API requests. "
+              "Jitter factor (0.0 to 1.0) applied to retry delays for Azure API requests. "
                   + "The jitter is applied as a random percentage of the computed exponential backoff delay. "
                   + "For example, 0.5 means up to 50%% random jitter will be added to each retry delay. "
-                  + "Helps prevent thundering herd when multiple requests fail simultaneously. "
-                  + "Currently only used by Azure storage integration. Designed to be applicable to AWS S3, GCP, and other cloud storage providers.")
+                  + "Helps prevent thundering herd when multiple requests fail simultaneously.")
           .defaultValue(0.5)
           .buildFeatureConfiguration();
 }
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
@@ -18,10 +18,10 @@
  */
 package org.apache.polaris.core.storage.azure;
 
-import static org.apache.polaris.core.config.FeatureConfiguration.STORAGE_API_RETRY_COUNT;
-import static org.apache.polaris.core.config.FeatureConfiguration.STORAGE_API_RETRY_DELAY_MILLIS;
-import static org.apache.polaris.core.config.FeatureConfiguration.STORAGE_API_RETRY_JITTER_FACTOR;
-import static org.apache.polaris.core.config.FeatureConfiguration.STORAGE_API_TIMEOUT_MILLIS;
+import static org.apache.polaris.core.config.FeatureConfiguration.AZURE_RETRY_COUNT;
+import static org.apache.polaris.core.config.FeatureConfiguration.AZURE_RETRY_DELAY_MILLIS;
+import static org.apache.polaris.core.config.FeatureConfiguration.AZURE_RETRY_JITTER_FACTOR;
+import static org.apache.polaris.core.config.FeatureConfiguration.AZURE_TIMEOUT_MILLIS;
 import static org.apache.polaris.core.config.FeatureConfiguration.STORAGE_CREDENTIAL_DURATION_SECONDS;
 
 import com.azure.core.credential.AccessToken;
@@ -327,12 +327,11 @@ public class AzureCredentialsStorageIntegration
    * <p>This method implements a defensive strategy against slow or failing cloud provider requests:
    *
    * <ul>
-   *   <li>Per-attempt timeout (configurable via STORAGE_API_TIMEOUT_MILLIS, default 15000ms)
-   *   <li>Exponential backoff retry (configurable count and initial delay via
-   *       STORAGE_API_RETRY_COUNT and STORAGE_API_RETRY_DELAY_MILLIS, defaults: 3 attempts starting
-   *       at 2000ms)
-   *   <li>Jitter to prevent thundering herd (configurable via STORAGE_API_RETRY_JITTER_FACTOR,
-   *       default 0.5 = 50%%) default 0.5 = 50%%)
+   *   <li>Per-attempt timeout (configurable via AZURE_TIMEOUT_MILLIS, default 15000ms)
+   *   <li>Exponential backoff retry (configurable count and initial delay via AZURE_RETRY_COUNT and
+   *       AZURE_RETRY_DELAY_MILLIS, defaults: 3 attempts starting at 2000ms)
+   *   <li>Jitter to prevent thundering herd (configurable via AZURE_RETRY_JITTER_FACTOR, default
+   *       0.5 = 50%%)
    * </ul>
    *
    * @param realmConfig the realm configuration to get timeout and retry settings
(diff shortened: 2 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +314,83 @@ private void validateAccountAndContainer(
...
+   * @return the access token
+   * @throws RuntimeException if token fetch fails after all retries or times out
+   */
   private AccessToken getAccessToken(String tenantId) {
     String scope = "https://storage.azure.com/.default";
     AccessToken accessToken =
         defaultAzureCredential
             .getToken(new TokenRequestContext().addScopes(scope).setTenantId(tenantId))
-            .blockOptional()
+            .timeout(Duration.ofSeconds(15)) // Per-attempt timeout
```
dimas-b: We have `RealmConfig` here, could you add a general setting in `FeatureConfiguration` for this timeout. I suppose it could applicable to other integrations too (but, of course, in this PR we can concentrate on Azure only)

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +314,83 @@ private void validateAccountAndContainer(
...
+            .doOnError(
+                error ->
+                    LOGGER.warn(
+                        "Error fetching Azure access token for tenant {}: {}",
+                        tenantId,
+                        error.getMessage()))
+            .retryWhen(
+                Retry.backoff(3, Duration.ofSeconds(2)) // 3 retries: 2s, 4s, 8s
```
dimas-b: Having backoff settings configurable could also be helpful.

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +314,83 @@ private void validateAccountAndContainer(
...
+                    .jitter(0.5) // ±50% jitter to prevent thundering herd
+                    .filter(
+                        throwable ->
+                            throwable instanceof java.util.concurrent.TimeoutException
```
dimas-b: `TimeoutException` is already handled by `isRetriableAzureException`, right?

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +314,83 @@ private void validateAccountAndContainer(
...
+                                tenantId,
+                                retrySignal.totalRetries() + 1))
+                    .onRetryExhaustedThrow(
+                        (retryBackoffSpec, retrySignal) ->
+                            new RuntimeException(
+                                String.format(
+                                    "Azure token fetch exhausted after %d attempts for tenant %s",
+                                    retrySignal.totalRetries(), tenantId),
+                                retrySignal.failure())))
+            .blockOptional(Duration.ofSeconds(90)) // Maximum total wait time
```
dimas-b: Why do we need this on top of `.timeout()` (line 337)?

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
fivetran-rahulprakash: I've added four new configuration constants in FeatureConfiguration:
`CLOUD_API_TIMEOUT_SECONDS` (default: 15) - Per-attempt timeout
`CLOUD_API_RETRY_COUNT` (default: 3) - Number of retry attempts
`CLOUD_API_RETRY_DELAY_SECONDS` (default: 2) - Initial delay for exponential backoff
`CLOUD_API_RETRY_JITTER_MILLIS `(default: 500) - Maximum jitter to prevent thundering herd
These use generic naming (CLOUD_API_*) rather than Azure-specific names, making them reusable for future  implementations.

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
RahulPrakash96: Addressed

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
RahulPrakash96: Yes, you're absolutely right! I missed that. Removed the duplicate check now. Thank you for catching that!

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
RahulPrakash96: Good point! I initially added the overall timeout as a safety net to ensure we never block indefinitely, but you're right it's unnecessary. The combination of per-attempt timeout and .retryWhen() with exponential backoff already provides sufficient protection. Removed it now. Thanks for the feedback!

fivetran-rahulprakash:
Thank you @dimas-b for the thorough review and excellent suggestions! I've addressed all your comments

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -438,4 +438,46 @@ public static void enforceFeatureEnabledOrThrow(
...
+      PolarisConfiguration.<Integer>builder()
+          .key("CLOUD_API_RETRY_DELAY_SECONDS")
+          .description(
+              "Initial delay in seconds before first retry for cloud provider API requests. "
+                  + "Delay doubles with each retry (exponential backoff). "
+                  + "Currently only used by Azure storage integration (not yet implemented for AWS S3 or GCP).")
+          .defaultValue(2)
+          .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Integer> CLOUD_API_RETRY_JITTER_MILLIS =
```
fivetran-rahulprakash: I chose to use milliseconds instead of a 0-1 jitter factor for a few reasons:

**User clarity** - It's more intuitive for operators to specify "500 milliseconds of jitter" rather than understanding what "0.5 jitter factor" means (50% of the retry delay)
**Concrete vs relative** - Millis gives direct control over the maximum random delay added, while a factor requires understanding how it interacts with the exponential backoff delays
**Consistency** - All other time-based configs use concrete units (seconds/millis) rather than abstract factors
**Predictability** - With millis, the max jitter is always clear regardless of retry delay values

The small conversion cost (jitterMillis / 1000.0) is negligible compared to the benefits of making the config more operator friendly. Happy to change to 0-1 factor if you prefer that approach though!

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +318,103 @@ private void validateAccountAndContainer(
...
+   *
+   * @param realmConfig the realm configuration to get timeout and retry settings
+   * @param tenantId the Azure tenant ID
+   * @return the access token
+   * @throws RuntimeException if token fetch fails after all retries or times out
+   */
+  private AccessToken getAccessToken(RealmConfig realmConfig, String tenantId) {
+    int timeoutSeconds = realmConfig.getConfig(CLOUD_API_TIMEOUT_SECONDS);
+    int retryCount = realmConfig.getConfig(CLOUD_API_RETRY_COUNT);
+    int initialDelaySeconds = realmConfig.getConfig(CLOUD_API_RETRY_DELAY_SECONDS);
```
dimas-b: Would you mind using millis for `initialDelaySeconds`... in some cases even 1 sec may be too long. Let's delegate what the min delay should be to the admin user who configures it. 

Same for `timeoutSeconds`... I hope Azure SDK supports millis.

RahulPrakash96: Thanks for the suggestion!

Changed both to milliseconds:
CLOUD_API_TIMEOUT_MILLIS (default: 15000ms)
CLOUD_API_RETRY_DELAY_MILLIS (default: 2000ms)

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +318,103 @@ private void validateAccountAndContainer(
...
+   * @param tenantId the Azure tenant ID
+   * @return the access token
+   * @throws RuntimeException if token fetch fails after all retries or times out
+   */
+  private AccessToken getAccessToken(RealmConfig realmConfig, String tenantId) {
+    int timeoutMillis = realmConfig.getConfig(CLOUD_API_TIMEOUT_MILLIS);
+    int retryCount = realmConfig.getConfig(CLOUD_API_RETRY_COUNT);
+    int initialDelayMillis = realmConfig.getConfig(CLOUD_API_RETRY_DELAY_MILLIS);
+    int jitterMillis = realmConfig.getConfig(CLOUD_API_RETRY_JITTER_MILLIS);
+    double jitter = jitterMillis / 1000.0; // Convert millis to fraction for jitter factor
```
dimas-b: I'm not sure, I fully understand this logic... per javadoc of `reactor.util.retry.RetryBackoffSpec.jitter()` the factor applies to the "computed delay", which may not be 1000 ms 🤔 How can the user reason about what the `CLOUD_API_RETRY_JITTER_MILLIS` value of `750` (for example) means?

Would it not be simpler to use the 0.0-1.0 factor value in the config?

RahulPrakash96: Ah, I went with millis initially to keep it consistent with the other timeout/delay configs (which are all in milliseconds), thinking it'd be more straightforward to work with absolute values.

But you're right - that doesn't really make sense here since the jitter factor gets applied to the computed delay, which changes with each retry (2s → 4s → 8s). So 750 would mean something different on each attempt, which is pretty confusing.

Changed it to CLOUD_API_RETRY_JITTER_FACTOR using the 0.0-1.0 range

dimas-b (approved):
LGTM 👍 Thanks again @fivetran-rahulprakash !

I'll leave it open for some more time for other reviewers to comment (if there's interest)

jbonofre (dismissed):

fivetran-rahulprakash:
> LGTM 👍 Thanks again @fivetran-rahulprakash !
> 
> I'll leave it open for some more time for other reviewers to comment (if there's interest)

Thanks! Just following up - are we good to merge 🤔

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +318,103 @@ private void validateAccountAndContainer(
...
     AccessToken accessToken =
         defaultAzureCredential
             .getToken(new TokenRequestContext().addScopes(scope).setTenantId(tenantId))
+            .timeout(Duration.ofMillis(timeoutMillis))
+            .doOnError(
+                error ->
+                    LOGGER.warn(
+                        "Error fetching Azure access token for tenant {}: {}",
+                        tenantId,
+                        error.getMessage()))
```
flyrain: Can we log the full stack trace instead of `error.getMessage()` only for better debuggability?

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -312,16 +318,103 @@ private void validateAccountAndContainer(
...
+                Retry.backoff(retryCount, Duration.ofMillis(initialDelayMillis))
+                    .jitter(jitter) // Apply jitter factor to computed delay
+                    .filter(this::isRetriableAzureException)
+                    .doBeforeRetry(
+                        retrySignal ->
+                            LOGGER.info(
+                                "Retrying Azure token fetch for tenant {} (attempt {}/{})",
+                                tenantId,
+                                retrySignal.totalRetries() + 1,
+                                retryCount))
```
flyrain: Should we add 1 to retryCount? Sot it matches the word "attempt" here? The total attempt number would be 1 + retryCount:
```
int maxAttempts = retryCount + 1;
...

LOGGER.info("Retrying Azure token fetch for tenant {} (attempt {}/{})",
            tenantId,
            retrySignal.totalRetries() + 1,
            maxAttempts);
```

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -438,4 +438,47 @@ public static void enforceFeatureEnabledOrThrow(
               "If set to true (default), allow credential vending for external catalogs. Note this requires ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING to be true first.")
           .defaultValue(true)
           .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Integer> CLOUD_API_TIMEOUT_MILLIS =
```
flyrain: `CLOUD_API_TIMEOUT_MILLIS` may over-communicate its scope. Would a more precise name like `STORAGE_API_TIMEOUT_MILLIS` or even `AZURE_STORAGE_API_TIMEOUT_MILLIS` be clearer? 

If s3/GCS/MinIO doesn't have similar configuration or cannot be applied with the same timeout, I'd suggest to have a prefix `AZURE_` to make it clear.

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
RahulPrakash96: Done! Changed to `LOGGER.warn("Error fetching Azure access token for tenant {}", tenantId, error)` which will log the full exception with stack trace. 😄

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
RahulPrakash96: Modified

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
RahulPrakash96: Good call! I've changed it to STORAGE_API_* to be more specific.

While only Azure uses these configs right now, the timeout/retry pattern applies to other cloud storage providers (AWS S3, GCP) too - their SDKs have similar async patterns that would benefit from this. So keeping the naming generic makes it easier to reuse when we add resilience to those integrations later.

Updated the descriptions to clarify it's currently Azure only but designed for future reuse. Let me know if you'd prefer something different!

fivetran-rahulprakash:
> Thanks a lot for working on it, @fivetran-rahulprakash ! LGTM overall. Left some comments!

Thanks for the thorough review, @flyrain! I've addressed all your comments

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
flyrain: @fivetran-rahulprakash , thanks for the change! One small concern, different storage backends may eventually need different timeout/retry configs. We can always introduce a broader-scope config like `STORAGE_API_TIMEOUT_MILLIS` later if we find something truly common across all storage types, but evolving in the opposite direction is harder. Once a wide-scope config is adopted and used in different places, shrinking it down to a more specific scope becomes much more painful. WDYT?

RahulPrakash96: @flyrain I've renamed the configs to use AZURE_* prefix. Thanks for the feedback!

PR #3125, opened 2025-11-23
Fix LICENSE/NOTICE in the distributions

### LINKED ISSUE #3120: LICENSE/NOTICE in distribution and docker images are not correct
I checked the `LICENSE` and `NOTICE` files in the distribution archive, but also in the distribution docker image (admin-tool and server), and they are not correct.

I'm fixing that and I will create a separate PR to avoid this ni the future.

jbonofre:
This closes #3120

### DIFF
--- a/runtime/admin/distribution/LICENSE
+++ b/runtime/admin/distribution/LICENSE
@@ -235,46 +235,48 @@ License: BSD 2-Clause
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


jbonofre:
@pingtimeout it has to be merged for 1.3.0-incubating RC.

jbonofre:
NB: I will create another PR to avoid to do this in the future.

PR #3134, opened 2025-11-24
NoSQL: Prepare admin-tool

snazy:
No functional changes.

1. Refactor the configuration property to a configuration type.
2. Make `BaseCommand` suitable for non-meta-store-factory use cases.


dimas-b (approved):

singhpk234:
> Make BaseCommand suitable for non-meta-store-factory use cases.

can you please elaborate what are the use cases ?

snazy:
> > Make BaseCommand suitable for non-meta-store-factory use cases.
> 
> can you please elaborate what are the use cases ?

Commands where a "running" meta-store would cause conflicts.
Or even a command to setup the JDBC schema would be.

PR #3135, opened 2025-11-24
NoSQL: CDI / Quarkus

snazy:
Adds the Quarkus/CDI code handling Quarkus specifics.

### DIFF
--- a/persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/BackendProvider.java
+++ b/persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/BackendProvider.java
@@ -42,17 +42,17 @@ class BackendProvider {
 
     var backendName =
         backendConfiguration
-            .type()
+            .backend()
             .orElseThrow(
                 () ->
                     new IllegalStateException(
-                        "Mandatory configuration option polaris.persistence.backend.type is missing!"));
+                        "Mandatory configuration option polaris.persistence.nosql.backend is missing!"));
 
     var backendBuilder = backendBuilders.select(BackendType.Literal.of(backendName));
     if (!backendBuilder.isResolvable()) {
       throw new IllegalStateException(
           format(
-              "Backend '%s' provided in configuration polaris.persistence.backend.type is not available. Available backends: %s",
+              "Backend '%s' provided in configuration polaris.persistence.nosql.backend is not available. Available backends: %s",
               backendName,
               backendBuilders
                   .handlesStream()
@@ -70,7 +70,7 @@ class BackendProvider {
     if (backendBuilder.isAmbiguous()) {
       throw new IllegalStateException(
           format(
-              "Multiple implementations match the backend name '%s' provided in configuration polaris.persistence.backend.type is not available. All available backends: %s",
+              "Multiple implementations match the backend name '%s' provided in configuration polaris.persistence.nosql.backend is not available. All available backends: %s",
               backendName,
               backendBuilders
                   .handlesStream()


persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/MongoDbBackendBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+import jakarta.enterprise.context.Dependent;
+import jakarta.inject.Inject;
+import org.apache.polaris.persistence.nosql.api.backend.Backend;
+import org.apache.polaris.persistence.nosql.mongodb.MongoDbBackendConfig;
+import org.apache.polaris.persistence.nosql.mongodb.MongoDbBackendFactory;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@BackendType(MongoDbBackendFactory.NAME)
+@Dependent
+class MongoDbBackendBuilder implements BackendBuilder {
```
dimas-b, line 34: Would it be possible to (re-)structure CDI objects in a way that will not mix multiple backends in the same general-purpose module?

I imagine when new backends are added, it would be preferable to keep their code isolated in their respective modules as much as possible.

In other words, is it possible to build a server with classes from just one particular backend implementation present on the class path?

persistence/nosql/persistence/cdi/quarkus/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+  id("polaris-server")
+}
+
+description = "Polaris NoSQL persistence, providers for Quarkus."
+
+dependencies {
+  implementation(project(":polaris-persistence-nosql-cdi-common"))
+  implementation(project(":polaris-persistence-nosql-api"))
+  implementation(project(":polaris-persistence-nosql-inmemory"))
+  implementation(project(":polaris-persistence-nosql-mongodb"))
```
dimas-b, line 31: side note: this is related to my comment on `MongoDbBackendBuilder`

LICENSE
```diff
@@ -338,6 +338,9 @@ This product includes code from Project Nessie.
 * helm/polaris/templates/servicemonitor.yaml
 * helm/polaris/templates/storage.yaml
 
+Code underneath the components/persistence directory, especially pluggable object types, index related, cache,
```
dimas-b: `components/persistence` does not exist in the current `main` 🤔 Shall we list individual files (as above)?

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/MongoDbBackendBuilder.java
snazy: Sure, we can add quarkus-specific modules next to every backend. But that's way more work at this point.

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/BackendProvider.java
```diff
@@ -0,0 +1,105 @@
...
+  Backend backend(
+      BackendConfiguration backendConfiguration, @Any Instance<BackendBuilder> backendBuilders) {
+
+    var backendName =
+        backendConfiguration
+            .type()
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Mandatory configuration option polaris.persistence.backend.type is missing!"));
```
singhpk234: shouldn't the backend configuration have nosql in the prefix like we do in relational-jdbc
for example :
polaris.persistence.relational.jdbc 
https://polaris.apache.org/in-dev/unreleased/configuration/

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/ObservingBackend.java
```diff
@@ -0,0 +1,263 @@
...
+import org.apache.polaris.persistence.nosql.api.backend.PersistId;
+import org.apache.polaris.persistence.nosql.api.backend.WriteObj;
+import org.apache.polaris.persistence.nosql.api.obj.ObjRef;
+import org.apache.polaris.persistence.nosql.api.ref.Reference;
+
+/** Provides telemetry and tracing for all persistence backend operations. */
+@ApplicationScoped
+@Default
+public class ObservingBackend implements Backend {
+  public static final String TELEMETRY_PREFIX = "polaris.persistence";
```
singhpk234: same comment as above https://github.com/apache/polaris/pull/3135/files#r2558515915

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/MongoDbBackendBuilder.java
dimas-b: Fair enough.

LICENSE
dimas-b: @jbonofre : WDYT?

snazy: Hard to name "individual files", as it's only inspired. But I see this causes confusion. Removing this from the PR.

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/ObservingBackend.java
snazy: Not sure a change here is necessary. The service can have only one kind of persistence.
Telemetry keys are already quite long, so I'd like to avoid making it even longer.

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/BackendProvider.java
snazy: `backend` is already NoSql specific. I'm not sure whether making something that's already specific is worth changing it?

singhpk234: IMHO its NoSQL specific to folks who checks the code and sees backend is under NoSQL module, but not for user POV it can be confusing never the less we already have `polaris.persistence.relational.jdbc` in the confs where relational-jdbc is the module, IMHO we should keep it consistent.

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/ObservingBackend.java
singhpk234: > The service can have only one kind of persistence

JDBC will not be able to support this telemetry at the moment, having something like polaris.persistence and not adding nosql in the confs might confuse the user that this is available for jdbc too, I resonate with the keys becoming a little bit log on adding nosql, but from user experience POV it will helpful

dimas-b: From my POV it would be beneficial for the JDBC Persistence to provide a similar "observing" front end for producing OTel spans.

singhpk234: Agree it would be something that would be really helpful for the users ! though even in the scenario we would want jdbc and NoSQL will have their own prefix ?

dimas-b: In terms of actual prefix, indeed a given server can have only one kind of persistence. However, nothing guarantees that a given telemetry collector processes data only from servers running the same type of persistence.

Since the prefix is part of span names, having something like `polaris.persistence.nosql` here and `polaris.persistence.jdbc` in JDBC (later, hypothetically) would make sense to me.

It is not likely that span queries or alerts depend on the exact span name prefix, so appending a word to the prefix is probably a relatively safe change from the user's perspective.

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/BackendProvider.java
dimas-b: @singhpk234 : I agree with your point in general, but this particular config is already in use in our downstream projects (from where we're contributing NoSQL code). Changing the config name is possible, but very inconvenient at this point (recall the very long NoSQL persistence discussion process). If you do not mind, I'd prefer to merge with the existing config names. As @snazy pointed out, config namespaces do not overlap, so there's no technical ambiguity.

End users in most cases do not need to rationalize config names, IMHO, they'll mostly follow documentation, which will instruct them what exact config names to set depending on the use case.

flyrain: Given the feature is still WIP, I think we should address the naming inconsistency before shipping it.
For JDBC we follow a clear prefix (`polaris.persistence.relational.jdbc`), but the NoSQL config currently doesn’t follow the same pattern.

To keep the overall configuration model consistent and intuitive for users, I believe we should align the NoSQL prefix as well before the feature stabilizes.

dimas-b: @flyrain : as I mentioned above there are existing downstream dependencies on these config names. This aligns with the lenient approach to keeping `polaris-core` at java 11 as [discussed](https://github.com/apache/polaris/discussions/100#discussioncomment-10267097) in #100. WDYT?

That said, it is certainly possible to refactor the config names later, while adding backward compatibility, but for the initial NoSQL persistence contribution that would add undue hardship, IMHO.

flyrain: NonSql is a WIP feature. I'm confused by the "existing downstream dependencies". Can you elaborate it?
And also how does it related to polaris-core  and java 11?

dimas-b: `polaris-core` has downstream dependencies the rely on it being compiled targeting java 11. There is no other reason against upgrading the java language level in that module.

`polaris.persistence.backend.type` has downstream dependencies that rely on that config name.

dimas-b: @collado-mike : wondering about your opinion on this matter, if you do not mind.

flyrain: NoSql is a WIP feature. There is no dependency outside of Polaris. If you are talking about the current Polaris modules, can you point it out? We should fix that as well before shipping it.

dimas-b: @flyrain :
> There is no dependency outside of Polaris

How do you know that?

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/ObservingBackend.java
snazy: Updated the prefix to `polaris.persistence.nosql`

dimas-b (approved):
LGTM. I propose to merge with existing config names for reasons I highlighted on the specific comment thread in this PR. If people think this is a major concern, let's move that discussion to the `dev` ML.

flyrain:
> let's move that discussion to the `dev` ML.

+1, let move the discussion to ML

dimas-b:
`dev` ML: https://lists.apache.org/thread/vmdpb45j1nmmlhswz7fm87gxgoldmr28

persistence/nosql/persistence/correctness/README.md
```diff
@@ -29,8 +29,8 @@ The `correctnessManualTest` task however is meant to be run _manually_ against a
 providing the necessary backend configuration via system properties. For example:
 ```bash
 ./gradlew :polaris-persistence-nosql-correctness:correctnessManualTest \
-  -Dpolaris.persistence.backend=MongoDb \
-  -Dpolaris.persistence.backend.mongodb.uri=mongodb://localhost:27017/test
-  -Dpolaris.persistence.backend.mongodb.databaseName=polaris_mongo_test
+  -Dpolaris.persistence.nosql=MongoDb \
```
dimas-b, line 32: `polaris.persistence.nosql.backend=...`?

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/BackendProvider.java
dimas-b: @flyrain : please see the latest config name change. I believe they align with your suggestion and this comment thread is not a blocker for merging. WDYT?

flyrain: LGTM. Thanks @snazy and @dimas-b for the changes!  cc @singhpk234

persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/ObservingBackend.java
```diff
@@ -0,0 +1,263 @@
...
+import org.apache.polaris.persistence.nosql.api.backend.PersistId;
+import org.apache.polaris.persistence.nosql.api.backend.WriteObj;
+import org.apache.polaris.persistence.nosql.api.obj.ObjRef;
+import org.apache.polaris.persistence.nosql.api.ref.Reference;
+
+/** Provides telemetry and tracing for all persistence backend operations. */
+@ApplicationScoped
+@Default
+public class ObservingBackend implements Backend {
+  public static final String TELEMETRY_PREFIX = "polaris.persistence.nosql";
```
singhpk234, line 53: nit / optional : may be we can include `.` in prefix `polaris.persistence.nosql.`

singhpk234 (approved):
LGTM, thanks @snazy !

PR #3137, opened 2025-11-24
Testing: increase visibility + make PCC/PMSM accessible

snazy:
* `BasePolarisMetaStoreManagerTest`: make `PolarisCallContext` + `PolarisMetaStoreManager` + `PolarisTestMetaStoreManager` accessible by subclasses
* Make constants of `PolarisRestCatalogMinIOIT` accessible

### DIFF
--- a/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/BasePolarisMetaStoreManagerTest.java
+++ b/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/BasePolarisMetaStoreManagerTest.java
@@ -65,7 +65,7 @@ public abstract class BasePolarisMetaStoreManagerTest {
 
   protected final MutableClock clock = MutableClock.of(Instant.now(), ZoneOffset.UTC);
 
-  private PolarisTestMetaStoreManager polarisTestMetaStoreManager;
+  protected PolarisTestMetaStoreManager polarisTestMetaStoreManager;
 
   @BeforeEach
   public void setupPolarisMetaStoreManager() {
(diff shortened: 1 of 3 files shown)


polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/BasePolarisMetaStoreManagerTest.java
```diff
@@ -65,7 +65,7 @@ public abstract class BasePolarisMetaStoreManagerTest {
 
   protected final MutableClock clock = MutableClock.of(Instant.now(), ZoneOffset.UTC);
 
-  private PolarisTestMetaStoreManager polarisTestMetaStoreManager;
+  protected PolarisTestMetaStoreManager polarisTestMetaStoreManager;
```
dimas-b, line 68: why not add an accessible getter/setter instead?

snazy: is it an issue for test code?

dimas-b: Obviously this class is meant to be extended. Direct field access is generally harder to refactor later, IMHO.

dimas-b: However, the scope of this class is probably limited to Polaris' code inside this repo, so it's not a concern

PR #3145, opened 2025-11-24
Make pom.xml always reproducible

snazy:
It turned out in practice, that there's no guarantee that the `<parent>` element in `pom.xml` files always appear at the same place.

This change ensures that the `<parent>` elements always appears at a deterministic location at the top of `pom.xml` files.

### DIFF
--- a/build-logic/src/main/kotlin/publishing/configurePom.kt
+++ b/build-logic/src/main/kotlin/publishing/configurePom.kt
@@ -19,6 +19,7 @@
 
 package publishing
 
+import groovy.namespace.QName
 import groovy.util.Node
 import org.gradle.api.Project
 import org.gradle.api.Task
@@ -58,7 +59,23 @@ internal fun configurePom(project: Project, mavenPublication: MavenPublication, 
           val projectNode = asNode()
 
           val parentNode = projectNode.appendNode("parent")
+          // Guarantee that the <parent> element is at a deterministic location.
+          // This is important for reproducible builds!
+          projectNode.remove(parentNode)
+          for ((index, any) in projectNode.children().withIndex()) {
+            if (any is Node) {
+              val qName = any.name() as QName
+              if (qName.localPart == "groupId") {
+                // In theory, we could also replace the groupId element, as the group ID is
+                // currently the same.
+                // But this would break once another group ID is built.
+                projectNode.children().add(index, parentNode)
+                break
+              }
+            }
+          }
           val parent = project.parent!!
+          // Add GAV to <parent> element
           parentNode.appendNode("groupId", parent.group)
           parentNode.appendNode("artifactId", parent.name)
           parentNode.appendNode("version", parent.version)


snazy:
For posterity a pom.xml from 1.3.0-incubating-rc0's staging. Notice the XML elements at the end.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.apache.polaris</groupId>
  <artifactId>polaris-persistence-nosql-benchmark</artifactId>
  <version>1.3.0-incubating</version>
  <packaging>pom</packaging>
  <licenses>
    <license>
      <name>Apache-2.0</name>
    </license>
  </licenses>
  <parent>
    <groupId>org.apache.polaris</groupId>
    <artifactId>polaris</artifactId>
    <version>1.3.0-incubating</version>
  </parent>
  <dependencies/>
</project>
```

And locally built:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.apache.polaris</groupId>
  <artifactId>polaris-persistence-nosql-benchmark</artifactId>
  <version>1.3.0-incubating</version>
  <packaging>pom</packaging>
  <licenses>
    <license>
      <name>Apache-2.0</name>
    </license>
  </licenses>
  <dependencies/>
  <parent>
    <groupId>org.apache.polaris</groupId>
    <artifactId>polaris</artifactId>
    <version>1.3.0-incubating</version>
  </parent>
</project>
```

build-logic/src/main/kotlin/publishing/configurePom.kt
```diff
@@ -58,7 +59,23 @@ internal fun configurePom(project: Project, mavenPublication: MavenPublication,
           val projectNode = asNode()
 
           val parentNode = projectNode.appendNode("parent")
+          // Guarantee that the <parent> element is at a deterministic location.
+          // This is important for reproducible builds!
+          projectNode.remove(parentNode)
```
dimas-b, line 64: why append on line 61 and immediately remove here?

dimas-b (approved):

Issue #3038, opened 2025-11-13
[FEATURE REQUEST] - Support S3 STS AssumeRoleWithWebIdentity for on‑prem S3 providers


michal-kuzak:
### Is your feature request related to a problem? Please describe.

Polaris currently supports S3 access via STS AssumeRole (role ARN + optional external ID). Please add first‑class support for STS AssumeRoleWithWebIdentity (ARWI). Many enterprise S3‑compatible object stores deployed on‑prem rely on OIDC federation and expose an STS endpoint that accepts ARWI, rather than operating a native IAM service. Examples include Pure Storage FlashBlade, or Dell S3 storage among others. Adding ARWI unlocks secure, short‑lived, IdP‑federated access to S3‑compatible stores without long‑lived keys. 

On‑prem S3 providers often delegate authentication to external IdPs (Okta, Microsoft Entra ID, etc.) and expect clients to obtain credentials via AssumeRoleWithWebIdentity to their STS endpoint. Polaris today documents role‑arn/region/external‑id and custom S3/STS endpoints, but not ARWI. Supporting ARWI would let Polaris run in OIDC‑first environments (Kubernetes, corporate SSO) and still vend short‑lived credentials to engines.

### Describe the solution you'd like

Add a storage auth selector for S3 catalogs, e.g. stsAuthMethod with values:
 - assumeRole (current default/behavior)
-  webIdentity (new)

When stsAuthMethod=webIdentity:
- Use AWS SDK’s ARWI credentials provider (e.g., StsAssumeRoleWithWebIdentity… in AWS SDK v2) against the  configured sts-endpoint (already supported). ([polaris.incubator.apache.org](https://polaris.incubator.apache.org/in-dev/unreleased/command-line-interface/?utm_source=openai))
- Accept inputs via CLI and REST model:
  - role-arn (unchanged)
  - role-session-name (optional)
  - web-identity-token-file (path to OIDC/JWT file)
  - region (optional; also allow via env)
  - sts-endpoint (already documented)

### Additional context

Default remains stsAuthMethod=assumeRole; existing catalogs and clients are unaffected.
IMPLICIT mode continues to work; ARWI simply adds another supported path to obtain base credentials.

Ref:
https://support.purestorage.com/bundle/FlashBlade_Admin_Guide_4.6.4/resource/FlashBlade_Admin_Guide_4.6.4.pdf  (p.462)

[...]

PR #3170, opened 2025-11-26
Support Aws STS AssumeRoleWithWebIdentity

cccs-cat001:
Here is a preliminary pass at fixing https://github.com/apache/polaris/issues/3038, I've added in an option to have the sts use assumeRoleWithWebIdentity and pass along the user credentials. This will allow the STS to know who is asking for credentials instead of the request coming from some shared service credentials. 

We've found that using our on-prem S3 solution that the way Polaris accesses the STS results in connectivity errors, and the recommended way of accessing the STS is with a web identity token (A.K.A. access token), so adding this optional feature will result in Polaris being usable by more S3 appliances than just AWS :) 
## Checklist

### DIFF
--- a/polaris-core/src/test/java/org/apache/polaris/core/storage/InMemoryStorageIntegrationTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/storage/InMemoryStorageIntegrationTest.java
@@ -199,7 +199,8 @@ class InMemoryStorageIntegrationTest {
         boolean allowListOperation,
         @Nonnull Set<String> allowedReadLocations,
         @Nonnull Set<String> allowedWriteLocations,
-        Optional<String> refreshCredentialsEndpoint) {
+        Optional<String> refreshCredentialsEndpoint,
+        Optional<String> token) {
       return null;
     }
   }
--- a/polaris-core/src/test/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheTest.java
@@ -80,6 +80,7 @@ public class StorageCredentialCacheTest {
                 Mockito.anyBoolean(),
                 Mockito.anySet(),
                 Mockito.anySet(),
+                Mockito.any(),
                 Mockito.any()))
         .thenReturn(badResult);
     PolarisEntity polarisEntity =
@@ -94,6 +95,7 @@ public class StorageCredentialCacheTest {
                     true,
                     Set.of("s3://bucket1/path"),
                     Set.of("s3://bucket3/path"),
+                    Optional.empty(),
                     Optional.empty()))
         .isInstanceOf(UnprocessableEntityException.class)
         .hasMessage("Failed to get subscoped credentials: extra_error_info");
@@ -109,6 +111,7 @@ public class StorageCredentialCacheTest {
                 Mockito.anyBoolean(),
                 Mockito.anySet(),
                 Mockito.anySet(),
+                Mockito.any(),
                 Mockito.any()))
         .thenReturn(mockedScopedCreds.get(0))
         .thenReturn(mockedScopedCreds.get(1))
@@ -125,6 +128,7 @@ public class StorageCredentialCacheTest {
         true,
         Set.of("s3://bucket1/path", "s3://bucket2/path"),
         Set.of("s3://bucket3/path", "s3://bucket4/path"),
+        Optional.empty(),
         Optional.empty());
     Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(1);
 
@@ -135,6 +139,7 @@ public class StorageCredentialCacheTest {
         true,
         Set.of("s3://bucket1/path", "s3://bucket2/path"),
         Set.of("s3://bucket3/path", "s3://bucket4/path"),
+        Optional.empty(),
         Optional.empty());
     Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(1);
   }
@@ -149,6 +154,7 @@ public class StorageCredentialCacheTest {
                 Mockito.anyBoolean(),
                 Mockito.anySet(),
                 Mockito.anySet(),
+                Mockito.any(),
                 Mockito.any()))
         .thenReturn(mockedScopedCreds.get(0))
         .thenReturn(mockedScopedCreds.get(1))
@@ -173,6 +179,7 @@ public class StorageCredentialCacheTest {
         true,
         Set.of("s3://bucket1/path", "s3://bucket2/path"),
         Set.of("s3://bucket/path"),
+        Optional.empty(),
         Optional.empty());
     Assertions.assertThat(storageCredentialCache.getIfPresent(cacheKey)).isNull();
 
@@ -182,6 +189,7 @@ public class StorageCredentialCacheTest {
         true,
         Set.of("s3://bucket1/path", "s3://bucket2/path"),
         Set.of("s3://bucket/path"),
+        Optional.empty(),
         Optional.empty());
     Assertions.assertThat(storageCredentialCache.getIfPresent(cacheKey)).isNull();
 
@@ -191,6 +199,7 @@ public class StorageCredentialCacheTest {
         true,
         Set.of("s3://bucket1/path", "s3://bucket2/path"),
         Set.of("s3://bucket/path"),
+        Optional.empty(),
         Optional.empty());
     Assertions.assertThat(storageCredentialCache.getIfPresent(cacheKey)).isNull();
   }
@@ -205,6 +214,7 @@ public class StorageCredentialCacheTest {
                 Mockito.anyBoolean(),
                 Mockito.anySet(),
                 Mockito.anySet(),
+                Mockito.any(),
                 Mockito.any()))
         .thenReturn(mockedScopedCreds.get(0))
         .thenReturn(mockedScopedCreds.get(1))
@@ -219,6 +229,7 @@ public class StorageCredentialCacheTest {
           true,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://bucket/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(++cacheSize);
     }
@@ -236,6 +247,7 @@ public class StorageCredentialCacheTest {
           /* allowedListAction= */ true,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://bucket/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(++cacheSize);
     }
@@ -247,6 +259,7 @@ public class StorageCredentialCacheTest {
           /* allowedListAction= */ false,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://bucket/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(++cacheSize);
     }
@@ -258,6 +271,7 @@ public class StorageCredentialCacheTest {
           /* allowedListAction= */ false,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://differentbucket/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(++cacheSize);
     }
@@ -274,6 +288,7 @@ public class StorageCredentialCacheTest {
           /* allowedListAction= */ false,
           Set.of("s3://differentbucket/path", "s3://bucket2/path"),
           Set.of("s3://bucket/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(++cacheSize);
     }
@@ -290,6 +305,7 @@ public class StorageCredentialCacheTest {
                 Mockito.anyBoolean(),
                 Mockito.anySet(),
                 Mockito.anySet(),
+                Mockito.any(),
                 Mockito.any()))
         .thenReturn(mockedScopedCreds.get(0))
         .thenReturn(mockedScopedCreds.get(1))
@@ -302,6 +318,7 @@ public class StorageCredentialCacheTest {
           true,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://bucket3/path", "s3://bucket4/path"),
+          Optional.empty(),
           Optional.empty());
     }
     Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(entityList.size());
@@ -314,6 +331,7 @@ public class StorageCredentialCacheTest {
           true,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://bucket3/path", "s3://bucket4/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(entityList.size());
     }
@@ -326,6 +344,7 @@ public class StorageCredentialCacheTest {
           true,
           Set.of("s3://bucket1/path", "s3://bucket2/path"),
           Set.of("s3://bucket3/path", "s3://bucket4/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(entityList.size());
     }
@@ -337,6 +356,7 @@ public class StorageCredentialCacheTest {
           true,
           Set.of("s3://bucket2/path", "s3://bucket1/path"),
           Set.of("s3://bucket3/path", "s3://bucket4/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(entityList.size());
     }
@@ -349,6 +369,7 @@ public class StorageCredentialCacheTest {
           true,
           Set.of("s3://bucket2/path", "s3://bucket1/path"),
           Set.of("s3://bucket4/path", "s3://bucket3/path"),
+          Optional.empty(),
           Optional.empty());
       Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(entityList.size());
     }
@@ -428,6 +449,7 @@ public class StorageCredentialCacheTest {
                 Mockito.anyBoolean(),
                 Mockito.anySet(),
                 Mockito.anySet(),
+                Mockito.any(),
                 Mockito.any()))
         .thenReturn(properties);
     List<PolarisEntity> entityList = getPolarisEntities();
@@ -439,6 +461,7 @@ public class StorageCredentialCacheTest {
             true,
             Set.of("s3://bucket1/path", "s3://bucket2/path"),
             Set.of("s3://bucket3/path", "s3://bucket4/path"),
+            Optional.empty(),
             Optional.empty());
     Assertions.assertThat(config.credentials())
         .containsExactly(Map.entry("s3.secret-access-key", "super-secret-123"));
--- a/polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
@@ -321,6 +321,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
                     true,
                     Set.of(s3Path(bucket, firstPath), s3Path(bucket, secondPath)),
                     Set.of(s3Path(bucket, firstPath)),
+                    Optional.empty(),
                     Optional.empty());
         assertThat(storageAccessConfig.credentials())
             .isNotEmpty()
@@ -422,6 +423,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
                 false, /* allowList = false*/
                 Set.of(s3Path(bucket, firstPath), s3Path(bucket, secondPath)),
                 Set.of(s3Path(bucket, firstPath)),
+                Optional.empty(),
                 Optional.empty());
     assertThat(storageAccessConfig.credentials())
         .isNotEmpty()
@@ -537,6 +539,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
                 true, /* allowList = true */
                 Set.of(s3Path(bucket, firstPath), s3Path(bucket, secondPath)),
                 Set.of(),
+                Optional.empty(),
                 Optional.empty());
     assertThat(storageAccessConfig.credentials())
         .isNotEmpty()
@@ -624,6 +627,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
                 true, /* allowList = true */
                 Set.of(),
                 Set.of(),
+                Optional.empty(),
                 Optional.empty());
     assertThat(storageAccessConfig.credentials())
         .isNotEmpty()
@@ -667,6 +671,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
                     true, /* allowList = true */
                     Set.of(),
                     Set.of(),
+                    Optional.empty(),
                     Optional.empty());
         assertThat(storageAccessConfig.credentials())
             .containsEntry(StorageAccessProperty.AWS_TOKEN.getPropertyName(), "sess")
@@ -708,6 +713,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
                     true, /* allowList = true */
                     Set.of(),
                     Set.of(),
+                    Optional.empty(),
                     Optional.empty());
         assertThat(storageAccessConfig.credentials())
             .isNotEmpty()
@@ -791,6 +797,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
             true,
             Set.of(s3Path(bucket, warehouseKeyPrefix + "/table")),
             Set.of(s3Path(bucket, warehouseKeyPrefix + "/table")),
+            Optional.empty(),
             Optional.empty());
 
     // Test with allowed KMS keys and read-only permissions
@@ -836,6 +843,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
             true,
             Set.of(s3Path(bucket, warehouseKeyPrefix + "/table")),
             Set.of(),
+            Optional.empty(),
             Optional.empty());
 
     // Test with no KMS keys and read-only (should add wildcard KMS access)
@@ -872,6 +880,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
             true,
             Set.of(s3Path(bucket, warehouseKeyPrefix + "/table")),
             Set.of(),
+            Optional.empty(),
             Optional.empty());
 
     // Test with no KMS keys and write permissions (should not add KMS statement)
@@ -905,6 +914,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
             true,
             Set.of(s3Path(bucket, warehouseKeyPrefix + "/table")),
             Set.of(s3Path(bucket, warehouseKeyPrefix + "/table")),
+            Optional.empty(),
             Optional.empty());
   }
 
--- a/polaris-core/src/test/java/org/apache/polaris/service/storage/azure/AzureCredentialStorageIntegrationTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/service/storage/azure/AzureCredentialStorageIntegrationTest.java
@@ -354,6 +354,7 @@ public class AzureCredentialStorageIntegrationTest extends BaseStorageIntegratio
         allowListAction,
         new HashSet<>(allowedReadLoc),
         new HashSet<>(allowedWriteLoc),
+        Optional.empty(),
         Optional.empty());
   }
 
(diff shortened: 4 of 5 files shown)


cccs-cat001:
Alright dumb question; I need to rebuild the `api/management-model` project so it regenerates its classes to include the new `userTokenSts` option in the `AwsStorageConfigInfo`, but I'm not sure how those classes get generated 😅 I think getting those classes to regenerate will fix the tests. I added the new userTokenSts option to the `./spec/polaris-management-service.yaml` already...

client/python/apache_polaris/cli/command/__init__.py
```diff
@@ -54,6 +54,7 @@ def options_get(key, f=lambda x: x):
                 allowed_locations=options_get(Arguments.ALLOWED_LOCATION),
                 role_arn=options_get(Arguments.ROLE_ARN),
                 external_id=options_get(Arguments.EXTERNAL_ID),
+                user_token_sts=options_get(Arguments.USER_TOKEN_STS),
```
adnanhemani: I find "user_token_sts" not so clear. Given that these tokens must be WebIdentityTokens (as per STS spec), I'd suggest renaming to something more specific, like "use_sts_web_identity_token".

Or, I think this property could be useful for similar features in Azure/GCP in the future (e.g. Azure's Federated Credentials or GCP's Workload Identity Federation) - and maybe we keep it something a bit vague like "use_subject_token" so as to not make it specifically named for STS.

@dimas-b WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -90,35 +96,46 @@ public StorageAccessConfig getSubscopedCreds(
...
-      accessConfig.put(StorageAccessProperty.AWS_KEY_ID, response.credentials().accessKeyId());
-      accessConfig.put(
-          StorageAccessProperty.AWS_SECRET_KEY, response.credentials().secretAccessKey());
-      accessConfig.put(StorageAccessProperty.AWS_TOKEN, response.credentials().sessionToken());
-      Optional.ofNullable(response.credentials().expiration())
+      Credentials credentials;
+      if (storageConfig.getUserTokenSTS()) {
+        AssumeRoleWithWebIdentityRequest.Builder request =
+            AssumeRoleWithWebIdentityRequest.builder()
+                .webIdentityToken(
```
adnanhemani, line 104: I'm not sure I understand this part properly (I read through the email thread as well) - are we passing in the user's token directly to STS? If so, what type of token is this: a Polaris authN token or a WebIdentityToken?

If a Polaris AuthN token, how does STS translate between this token and an IAM trust policy enforcement - which should state which users are allowed to assume this role? And if such a translation does exist, where in Polaris is this logic?

If a WebIdentityToken, then what stops this token from contacting STS directly to assume the role and get credentials for S3?

I understand there may be some special rules you have in place for an on-prem solution that may limit the ability of a user to contact STS directly or to translate between a Polaris token and an identity that STS can understand - but this seems to be breaking the general architecture of Polaris where Polaris serves as a trusted credential broker and clients are assumed to not be able to bypass Polaris in order to get storage credentials.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -90,35 +96,46 @@ public StorageAccessConfig getSubscopedCreds(
...
+                    securityIdentity.getCredential(PolarisCredential.class).getToken())
+                .roleArn(storageConfig.getRoleARN())
+                .roleSessionName("PolarisAwsCredentialsStorageIntegration")
+                .durationSeconds(storageCredentialDurationSeconds);
```
adnanhemani, line 120: Don't see a `policy` block added here (like it is in the non-token code path). If this is intentional, please explain why the scoping-down of credentials isn't applicable here?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
adnanhemani: To be clear, I'm making this assumption that the token is a user's token based off of https://github.com/apache/polaris/pull/3170/files#diff-204e04f3365ea6ec898018ebfe943ca22277ae4643b58d1ca31462f70662c618R94. If the SecurityIdentity is populating with the server's token (not sure this makes sense to me, but feel free to correct me), then that may open a different set of questions.

runtime/service/src/test/java/org/apache/polaris/service/auth/external/OidcPolarisCredentialAugmentorTest.java
```diff
@@ -109,7 +109,7 @@ public void testAugmentOidcPrincipal() {
             .addRole("ROLE1")
             .addAttribute(TENANT_CONFIG_ATTRIBUTE, config)
             .build();
-
+    when(oidcPrincipal.getRawToken()).thenReturn("this_is_a_token");
```
adnanhemani, line 112: nit: let's separate "this_is_a_token" into a shared variable.

runtime/service/build.gradle.kts
```diff
@@ -30,7 +30,6 @@ dependencies {
   implementation(project(":polaris-api-management-service"))
   implementation(project(":polaris-api-iceberg-service"))
   implementation(project(":polaris-api-catalog-service"))
-
```
adnanhemani, line 33: Unnecessary change

runtime/service/build.gradle.kts
```diff
@@ -103,7 +102,7 @@ dependencies {
   implementation("com.fasterxml.jackson.core:jackson-annotations")
   implementation("com.fasterxml.jackson.core:jackson-core")
   implementation("com.fasterxml.jackson.core:jackson-databind")
-
```
adnanhemani, line 106: Unnecessary change

.gitignore
```diff
@@ -105,4 +105,5 @@ hs_err_pid*
 .vscode/
 
 # Python Virtual Env
-.venv
\ No newline at end of file
+.venv
+venv
```
adnanhemani: Unnecessary change

spec/polaris-management-service.yml
```diff
@@ -1136,6 +1136,11 @@ components:
                 for accessing data and metadata files within the related catalog. Setting this property to `true`
                 effectively disables vending storage credentials to clients. This setting is intended for configuring
                 catalogs with S3-compatible storage implementations that do not support STS.
+            userTokenSts:
```
dimas-b: suggestion: `propagateApiUserIdentity` - it is more directly connected to actual code behaviour... at least for me 😅 I'd welcome other naming suggestions too. Better have this discussion now than after the stuff is released :)

client/python/apache_polaris/cli/command/__init__.py
dimas-b: I made a comment about this name on the API yaml. Let's discuss actual name there.

Once that is decided, my personal opinion is that using the same name in Python CLI is quite appropriate.

client/python/apache_polaris/cli/command/catalogs.py
```diff
@@ -76,6 +76,7 @@ class CatalogsCommand(Command):
     sts_endpoint: str
     sts_unavailable: bool
     path_style_access: bool
+    user_token_sts: bool
```
dimas-b: nit: Combining CLI changes and backend changes in the same PR is ok, but these areas usually attract different sets of reviewers. As for me, I usually do CLI changes in a follow-up PR. Also, I usually do not have material opinions about CLI code changes :) all in all, it might be easier for reviewers to have CLI and java changes in different PRs... I guess.

integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java
```diff
@@ -159,6 +159,7 @@ private void setupCatalogs() {
         AwsStorageConfigInfo.builder()
             .setStorageType(StorageConfigInfo.StorageTypeEnum.S3)
             .setPathStyleAccess(true)
+            .setUserTokenSts(false)
```
dimas-b: Is this required? Could we handle defaults transparently to code areas not affected by the new feature?

runtime/service/src/main/java/org/apache/polaris/service/auth/PolarisCredential.java
```diff
@@ -31,14 +31,22 @@
 public interface PolarisCredential extends Credential {
 
   static PolarisCredential of(
-      @Nullable Long principalId, @Nullable String principalName, Set<String> principalRoles) {
```
dimas-b: Generally speaking, it is preferable to add new builder methods for new optional parameters (and chain old methods into them). This way existing code that does not use new optional parameter won't have to change.

dimas-b, line 34: Please add a new `of()` method for the new parameter (since it's optional). This way existing code that does not use the token, does not have to change.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -44,13 +46,15 @@
...
-import software.amazon.awssdk.services.sts.model.AssumeRoleResponse;
+import software.amazon.awssdk.services.sts.model.AssumeRoleWithWebIdentityRequest;
+import software.amazon.awssdk.services.sts.model.Credentials;
 
 /** Credential vendor that supports generating */
 public class AwsCredentialsStorageIntegration
     extends InMemoryStorageIntegration<AwsStorageConfigurationInfo> {
   private final StsClientProvider stsClientProvider;
   private final Optional<AwsCredentialsProvider> credentialsProvider;
+  private final SecurityIdentity securityIdentity;
```
dimas-b: Depending on Quarkus classes in core is really not desirable... AFAIK, sorry.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisCredential.java
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.polaris.service.auth;
+package org.apache.polaris.core.auth;
```
dimas-b: I'm not sure about moving this class to core 🤔 

Why not add the token as a well-known property to `PolarisPrincipal`?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: `PolarisPrincipal` should be injectable into `StorageAccessConfigProvider` and then it should be possible to propagate it to this class as method parameters... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisCredential.java
cccs-cat001: That's a fantastic question and I completely agree! Problem is I couldn't find where the PolarisPrincipal was getting set up and the classes you referenced for me were setting up a PolarisCredential so I ran with that lol

client/python/apache_polaris/cli/command/__init__.py
cccs-cat001: yeah I think `propagateApiUserIdentity` is good, it's generic enough that it can work with any other storage appliance, not just S3, if it's ever needed.

.gitignore
cccs-cat001: I disagree. Some people (including myself) use venv instead of .venv and it's a good QOL addition to not have to sift through the venv to find your changes. 

That being said, is there any security issues or genuine concerns that arise having this `venv` directory ignored?

spec/polaris-management-service.yml
cccs-cat001: Discussion happening in the other thread (https://github.com/apache/polaris/pull/3170#discussion_r2566597009), will close this comment

client/python/apache_polaris/cli/command/catalogs.py
cccs-cat001: I've split out the CLI changes and will post a new PR after this one. 

Or should I post it now too?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisCredential.java
dimas-b: I believe it happens here: https://github.com/apache/polaris/blob/3c5dbaf2bc93bb2861b5fb0407ffc1597e48bdd2/runtime/service/src/main/java/org/apache/polaris/service/auth/DefaultAuthenticator.java#L97

The effectiveness of the new STS request mode will depend on the `Authenticator` implementations supporting token propagation. It is a bit obscure, but I think it aligns well with the general authentication flows in Polaris.

In any case, there's only one `Authenticator` in OSS code ATM.

dimas-b: Note: for OIDC flows, the JWT in the request is already authenticated by Quarkus by the time `DefaultAuthenticator` is called... It may still need to be passed through `PolarisCredential`, but at least that part does not touch "core".

client/python/apache_polaris/cli/command/catalogs.py
dimas-b: Later is fine... we need to settle on the config names in API first :sweat_smile:

dimas-b: CLI does not _have to_ be always 1:1 with the REST API spec... We can evolve it separately... even across release boundaries... but I hope we can finish all for 1.4.0 :wink:

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
```diff
@@ -72,6 +72,7 @@ public void testJsonFormat() throws JsonProcessingException {
                 + "\"storageConfigInfo\":{"
                 + "\"roleArn\":\"arn:aws:iam::123456789012:role/test-role\","
                 + "\"allowedKmsKeys\":[],"
+                + "\"propagateApiUserIdentity\":false,"
```
dimas-b: nit: could you handle `propagateApiUserIdentity` similar to `stsUnavailable` so that old clients will not see this entry unless it's set explicitly?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
```diff
@@ -74,4 +106,8 @@ static PolarisPrincipal of(String name, Map<String, String> properties, Set<Stri
    * as permissions, preferences, or other metadata.
    */
   Map<String, String> getProperties();
+
+  /** Returns the access token of the current user. */
+  @Nullable
+  String getToken();
```
dimas-b, line 112: nit: WDYT about `Optional<String>` here?

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -88,7 +91,8 @@ public PolarisStorageIntegrationProviderImpl(
                 new AwsCredentialsStorageIntegration(
                     (AwsStorageConfigurationInfo) polarisStorageConfigurationInfo,
                     stsClientProvider,
-                    stsCredentials);
+                    stsCredentials,
+                    context);
```
dimas-b: This "provider" is `@ApplicationScoped`, but `SecurityContext` is conceptually request-scoped. `AwsCredentialsStorageIntegration` is accessed via `StorageCredentialCache`.

Current setup creates a logical inconsistency: the cache may return data from memory, while the request-scoped context contains a different token. The cache does not "see" this context.

I believe it would be preferably to carry the token through the method parameter so that the cache can take it into account when it computes its key. This will increase cache churn, of course, but it's a matter of correctness at this point, not optimization (I think).

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -151,6 +151,7 @@ options:
       --role-arn  (Only for AWS S3) A role ARN to use when connecting to S3
       --region  (Only for S3) The region to use when connecting to S3
       --external-id  (Only for S3) The external ID to use when connecting to S3
+      --propagate-api-user-identity  (Only for S3) Whether to use the requesting users identity when connecting to the STS
```
dimas-b: Does this belong in the CLI PR? 🤔

.gitignore
dimas-b: I agree with @cccs-cat001 on this topic, but I'd propose to make this change in a separate PR to avoid mixing QOL changes with feature changes 😉

runtime/service/build.gradle.kts
```diff
@@ -103,7 +102,7 @@ dependencies {
...
+  implementation(libs.apache.httpclient5)
```
dimas-b: Is this required for the new Assume Role With Web Identity call?

dimas-b:
@adutra : could you double check quarkus auth integration changes (they LGTM)

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
cccs-cat001: from what I gather with this flow, we don't need to provide a policy string. Seems that the STS reads the users token and provides you with all the credentials you have permission for.

I'm not an expert in S3 or STS so I could be way off and feel free to correct me if I am, but from testing with our appliance this is the way it seems to be.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
cccs-cat001: Yeah we pass the polaris token directly to the STS. Again, not an S3 or STS expert, but our STS and storage appliance seems happy to accept the polaris token. I don't have the means to test any other storage appliance with any other configurations unfortunately.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
cccs-cat001: Seems the same to me, what's bad about using `@Nullable`?

site/content/in-dev/unreleased/command-line-interface.md
cccs-cat001: sure does! my bad :)

runtime/service/build.gradle.kts
cccs-cat001: Nope!

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: Access down-scoping (policy-based restrictions) makes more sense in the case when Polaris has its own "service" credential because we do not want to delegate more rights to the user than what is necessary for the API request.

In this case the STS session is based in the unmodified API user identity. The user can gain the same access by talking to STS directly.

The only reason for restricting access might be to enforce catalog-level location settings. However, that would be mostly about preventing accidental mistakes on the client side rather than a true security control.

That said, for the sake of narrowing Polaris behaviour variations it might still be worth applying the same policy as in the "service" account case... just for ease of reasoning about how the system works.

Taking this one step further, it might be worth controlling the policy application with yet another config property, if there are use cases when the extra policy is not desirable.

.gitignore
cccs-cat001: Can't we just slip in a quick little gitignore change 🥺

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
cccs-cat001: I'm not sure what I did differently from `stsUnavailable` tbh. I'm gonna remove that change and see if this was just a "quirk" of my local setup

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
cccs-cat001: okay, I'll test out the policy string on our storage before committing it

.gitignore
dimas-b: I'm ok with this... just proposing a faster (IMHO) path :slightly_smiling_face:

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: from my POV, it would be ok to do the policy change in a follow-up PR. Current change is already under a dedicated config, so existing clients and servers will not be not affected even if we merge "as is".

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
dimas-b: It's a "nit" comment :) `Optional` is slightly preferable IMHO because it's an essential piece of information for making the `StorageAccessConfig`, but the content of the token is optional... really fine point :) I'm ok either way.

dimas-b:
@cccs-cat001 : please resolve conflicts... otherwise CI does not run :shrug:

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
dimas-b: @cccs-cat001 : just for clarity, from my POV this is the only issue that is really critical to resolve before merging.

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
dimas-b: The latest state of the PR LGTM. As long as the new property does not show up when unset, it's fine. Resolving this thread.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisPrincipal.java
dimas-b: You may want to hold this until you deal with `StorageCredentialCacheKey`, then use whatever is more convenient on the cache side.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
cccs-cat001: Yeah I'm working on it, I'm just dumb 😅 😆

adutra:
> @adutra : could you double check quarkus auth integration changes (they LGTM)

I find the changes in this PR a bit invasive tbh. I would rather explore Quarkus OIDC token propagation techniques, and more specifically, I would try to inject `io.quarkus.oidc.client.Tokens` wherever necessary:

https://quarkus.io/guides/security-openid-connect-client-reference#inject-tokens

But I don't want to delay this PR that is extremely useful. I think we can go with the approach taken here and then switch to token propagation in a follow-up PR.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
adutra: > I'm not sure I understand this part properly (I read through the email thread as well) - are we passing in the user's token directly to STS? If so, what type of token is this: a Polaris authN token or a WebIdentityToken?

It's the authenticated principal's OAuth2 access token, as provided by the authentication layer.

> how does STS translate between this token and an IAM trust policy enforcement - which should state which users are allowed to assume this role? And if such a translation does exist, where in Polaris is this logic?

The IAM role being assumed must have a pre-configured trust policy that explicitly trusts the IDP and potentially specifies conditions like the expected audience and subject of the token. The STS service is supposed to validate the signature and claims of the provided token against the registered IDP before issuing credentials.

> this seems to be breaking the general architecture of Polaris where Polaris serves as a trusted credential broker and clients are assumed to not be able to bypass Polaris in order to get storage credentials.

It's not that simple. The remote client may be in possession of valid OAuth2 credentials for WebIdentity, but you still need an STS client to issue the `AssumeRoleWithWebIdentityRequest`. The STS client requires its own credentials. If the remote client doesn't have those credentials, it cannot talk to STS.

adutra:
> By creating this approach, we are introducing a security loophole that is completely against Polaris' security modelling today.

I am completely failing to see what security loopholes we would be introducing by leveraging `AssumeRoleWithWebIdentity`. Propagating the user's access token to AWS STS using `AssumeRoleWithWebIdentity` is the standard pattern recommended by AWS itself for federated OIDC access.

> the client should be modified to talk to STS directly to gain credentials without Polaris' involvement. IMO Polaris should not be involved in abetting ANY security loopholes.

Are you serious about allowing clients to talk to STS directly? THAT, indeed, would be a giant security loophole.

> Or we can look towards [...] setting Polaris up as an intermediate token broker which the "STS" service trusts, a new credential can be minted for the client to access the storage layer.

IMHO this is unrealistic, and over-engineered. That would require a form of token exchange and would be extremely hard to implement for little added-value. And again: **let's please stop considering Polaris as an OAuth2 token broker. This is legacy behavior.**

> Trying to fit a simplistic solution to quick-solve a problem for a system we don't even fully support is a _very_ dangerous precedence to set.

Can you clarify what "simplistic solution" you are talking about and what is this "system [that] we don't [...] fully support"?

dimas-b:
I support @adutra 's point about Polaris _not_ acting as a token broker or authorization server. Polaris, in general, should be a resource server in OAuth2 terms.

cccs-cat001:
> I find the changes in this PR a bit invasive tbh. I would rather explore Quarkus OIDC token propagation techniques, and more specifically, I would try to inject `io.quarkus.oidc.client.Tokens` wherever necessary:
> 
> https://quarkus.io/guides/security-openid-connect-client-reference#inject-tokens
> 
> But I don't want to delay this PR that is extremely useful. I think we can go with the approach taken here and then switch to token propagation in a follow-up PR.

that looks pretty simple compared to all the changes I'm making tbh. Is everything set up so that I can just `@Inject Token token;`? 🤔

dimas-b:
> Is everything set up so that I can just @Inject Token token;? 🤔

Injecting in `runtime/service` classes should be ok, but `polaris-core` should remain CDI-neutral by convention (i.e. no CDI annotations in core).

That is to say, I think current `PolarisPrincipal` changes will probably have to remain.

adutra:
> that looks pretty simple compared to all the changes I'm making tbh. Is everything set up so that I can just `@Inject Token token;`? 🤔

You may need to add the `io.quarkus:quarkus-oidc-client` dependency, I don't remember if it's already there or not.

adutra:
> That is to say, I think current `PolarisPrincipal` changes will probably have to remain.

True, at some point we'd need to pass the token as a parameter. @cccs-cat001 please don't rush into using token propagation if you can't find a simple path forward – what you have currently is perfectly acceptable.

tokoko:
To be honest I'm very confused about the implementation in the PR as well. Not sure about other s3-compatible systems, but in the case of minio, (if I'm reading this correctly...) this option will only work if all catalog users have the ability to assume the role that's configured centrally on the catalog level and which also needs to have read/write privileges on all catalog locations. I fail to see how this can be useful to anyone. why would anyone hand out assume role privileges on what's basically a superuser?

If the problem that the PR is trying to solve is simply to support those systems that only support `AssumeRoleWithWebIdentity`, the more straightforward solution would be to enable the catalog to acquire required web identity instead of acquiring it from a user. The same way that the standard `AssumeRole` option relies on AWS environment variables to authenticate the call, the new `AssumeRoleWithWebIdentity` solution should read necessary oauth configs and have an internal background process that obtains and refreshes a token needed to authenticate AssumeRoleWithWebIdentity calls. I'm not sure why we would complicate this with some sort of token pass-through mechanism.

dimas-b:
@tokoko : AFAIK (and @cccs-cat001 may have a better answer) this feature is mostly intended for custom systems. Using the workflow it enables with AWS, for example, may be possible, yet whether it is the best approach with AWS is subject to specific user demands. I tend to agree that this may not be the best approach for MinIO or AWS S3. 

Nonetheless, I do not see why Polaris as an OSS project should not open this use case for deployments what may benefit from it. The feature is well isolated under a catalog property, there is no impact to other use cases.

For reference, here's the related dev ML thread: https://lists.apache.org/thread/tm76ntbgdqt31r6402dro8vb7m4pdzzq

adnanhemani:
> Propagating the user's access token to AWS STS using AssumeRoleWithWebIdentity is the standard pattern recommended by AWS itself for federated OIDC access.

This is only the standard pattern recommended by AWS for federated OIDC access - _if you do not have a governance system in between_. If you feel that's wrong, please feel free to drop links to the AWS documentation here that support your claim.

> Are you serious about allowing clients to talk to STS directly? THAT, indeed, would be a giant security loophole.

(Also mentioned in Laurent's email to the Dev ML regarding this approach) Isn't this possible today with how the code is written? If it isn't, what stops it? Is it a network policy? And as a result, are we betting the whole of Polaris' security posture on such a network policy? Is that truly wise?

Let's compare this to the `AssumeRole` path we have today. In our best practices, are IAM roles allowed to be assumed by the clients directly? That is clearly not the model that Polaris has today - so why are we breaking that security model for this feature?

> Nonetheless, I do not see why Polaris as an OSS project should not open this use case for deployments what may benefit from it. The feature is well isolated under a catalog property, there is no impact to other use cases.

Not sure I agree, we are enabling a use case here that does not align with our current security best practices. And to be clear, this is not as a result of there not being a better way of doing this, but rather because this is the fastest way of achieving this goal. Trampling security best practices for velocity is not something that we have a track record of within this project.

I agree that this use case is a valid one - and that we should be supporting it. But I highly disagree in the way that this implementation does so. Rather than bending the security model that we have today (which generally can be stated like: Polaris user who is unprivileged with regards to storage credentials, authenticates and authorizes with Polaris in order to gain storage credentials) to fit this use case, we should be tailoring the implementation of this use case to fit the security model.

api/management-model/src/test/java/org/apache/polaris/core/admin/model/CatalogSerializationTest.java
cccs-cat001: Okay so I've been banging my head against the wall with this for a little bit. The only difference I can find between this and the `stsUnavailable` is that we have `propagateApiUserIdentity` default to false, this was so that we didn't get errors in `AwsCredentialsStorageIntegration.java L101`, but I've since added the `Boolean.TRUE.equals` so I think it's fine to remove the default in the `polaris-management-service.yml`?

cccs-cat001:
Marking as draft as tests are failing for unknown reasons, they'll potentially be fixed with #3203, so we'll wait for that.

cccs-cat001:
Closing this PR in favour of #3224 and #3236

PR #3171, opened 2025-11-26
Make homepage image full-width

tmater:
## Summary

Makes the homepage hero image display at full viewport width without left/right margins.

## Problem

The homepage hero image was displaying with unwanted left and right spacing. This was caused by the `.td-content` container padding (controlled by Bootstrap's `--bs-gutter-x` variable) that wraps all home page content.

## Solution

The container padding comes from Hugo's Docsy theme layout and should not be modified as it provides comfortable reading margins for text content. Instead, added `.imageblock.full-width` CSS class that uses negative margins to pull specific images out to the viewport edges, compensating for the container's padding without affecting other content.

Applied the class to the homepage image using AsciiDoc role attribute `[.full-width]`.

This approach is targeted and reusable - any image can be made full-width by adding `[.full-width]` before the `image::` directive in AsciiDoc.

## Before/After

**Before:** 
before
**After:**
after

### DIFF
--- a/site/assets/scss/_styles_project.scss
+++ b/site/assets/scss/_styles_project.scss
@@ -24,3 +24,14 @@
 .td-sidebar-toc {
   padding-top: 1.75rem;
 }
+
+// Make image span full viewport width by canceling out
+// container padding
+.imageblock.full-width {
+  margin-inline: calc(0px - var(--bs-gutter-x));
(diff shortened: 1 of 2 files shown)


dimas-b (approved):

flyrain (approved):

flyrain:
Thanks for the change, @tmater ! Thanks for the review @binarycat0 @dimas-b !

PR #3199, opened 2025-12-03
added venv to the gitignore

cccs-cat001:
extra QOL change for those of use not using hidden venvs :) 

## Checklist


dimas-b (approved):
Both directories are supported locations in [venv](https://docs.python.org/3/library/venv.html), so I think it make sense to exclude both and leave the choice of which one to use to the developer.

dimas-b:
Merging with 3 approvals... If other developers have concerns, we can always revisit the excludes :slightly_smiling_face:

PR #2825, opened 2025-10-16
[Site] fix typo of catalog federation configuration in command-line-interface.md

CodingBangboo:
### What changes were proposed in this pull request?

Python CLI currently support configuring 2 types of connection for catalog federation
- `iceberg-rest` for Rest Catalog
- `hadoop` for Hadoop Catalog

However, the CLI page currently shows the wrong strings
```
--catalog-connection-type  The type of external catalog in [ICEBERG, HADOOP].
```

This PR fixes typo to avoid confusion when users try catalog federation through python CLI.

PR #2761, opened 2025-10-06
Site: Add docs for catalog federation

flyrain:
The PR adds docs for catalog federation. With the help of chatGPT codex.

PR #3215, opened 2025-12-05
[doc]: Doc fix for CLI usage

MonkeyCanCode:
Synced usage from latest CLI.

## Checklist

### DIFF
--- a/site/content/in-dev/unreleased/command-line-interface.md
+++ b/site/content/in-dev/unreleased/command-line-interface.md
@@ -158,7 +158,7 @@ options:
       --consent-url  (Only for Azure) A consent URL granting permissions for the Azure Storage location
       --service-account  (Only for GCS) The service account to use when connecting to GCS
       --property  A key/value pair such as: tag=value. Multiple can be provided by specifying this option more than once
-      --catalog-connection-type  The type of external catalog in [ICEBERG, HADOOP, HIVE].
+      --catalog-connection-type  The type of external catalog in [ICEBERG-REST, HADOOP, HIVE].
       --iceberg-remote-catalog-name  The remote catalog name when federating to an Iceberg REST catalog
       --hadoop-warehouse  The warehouse to use when federating to a HADOOP catalog
       --hive-warehouse  The warehouse to use when federating to a HIVE catalog
--- a/client/python/apache_polaris/cli/constants.py
+++ b/client/python/apache_polaris/cli/constants.py
@@ -274,7 +274,7 @@ class Hints:
 
         class External:
             CATALOG_CONNECTION_TYPE = (
-                "The type of external catalog in [ICEBERG, HADOOP, HIVE]."
+                "The type of external catalog in [ICEBERG-REST, HADOOP, HIVE]."
             )
             CATALOG_AUTHENTICATION_TYPE = (
                 "The type of authentication in [OAUTH, BEARER, SIGV4, IMPLICIT]"


site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -156,7 +158,7 @@ options:
       --consent-url  (Only for Azure) A consent URL granting permissions for the Azure Storage location
       --service-account  (Only for GCS) The service account to use when connecting to GCS
       --property  A key/value pair such as: tag=value. Multiple can be provided by specifying this option more than once
-      --catalog-connection-type  The type of external catalog in [iceberg-rest, hadoop, hive].
+      --catalog-connection-type  The type of external catalog in [ICEBERG, HADOOP, HIVE].
```
flyrain: Did we intentionally change it to lower case? cc @HonahX @CodingBangboo 

Reference: https://github.com/apache/polaris/pull/2825/files, https://github.com/apache/polaris/pull/2761#discussion_r2422126748

HonahX: Yes, I think it should be the lower cased version, since they are the actual value of those enum type:
https://github.com/apache/polaris/blob/be3c88b6d057009a05301a25ef99b775604955ea/client/python/apache_polaris/cli/constants.py#L51-L58

MonkeyCanCode: K. let me fix the tool doc.

MonkeyCanCode: so it doesn't appear it matters:
```
        Argument(Arguments.CATALOG_CONNECTION_TYPE, str,
                 Hints.Catalogs.External.CATALOG_CONNECTION_TYPE, lower=True,
                 choices=[ct.value for ct in CatalogConnectionType]),
```
as we are applying `lower=True` anyway. Also, in other places, we are refers to upper case ones as well:
```
# Allows both REST and HIVE connection type
polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST","HIVE"]
```

Also, as all other arguments with enum are pass in as upper case, should we keep them as upper case then. For example:
```
            CATALOG_AUTHENTICATION_TYPE = (
                "The type of authentication in [OAUTH, BEARER, SIGV4, IMPLICIT]"
            )
...
        Argument(Arguments.CATALOG_AUTHENTICATION_TYPE, str,
                 Hints.Catalogs.External.CATALOG_AUTHENTICATION_TYPE, lower=True,
                 choices=[at.value for at in AuthenticationType]),
```

Another thing is we have it as `ICEBERG` but the actual value is `iceberg-rest` and in the java server side, we have it as `ICEBERG_REST`, should we keep them consistent?

flyrain: I'm fine with `[ICEBERG-REST, HADOOP, HIVE]` in that case.

HonahX: Thanks for diving deep into this！
Upper case SGTM. As for the iceberg-rest to iceberg change that would be a breaking change for current users. It's worth to mention in the changelog if we modify it.

IMHO, ICEBERG-REST should be fine.

MonkeyCanCode: Thanks for review. As the current update is correct (the upper case ones). Should we proceed with this change first? I will create another PR for changing from `ICEBERG_REST` to `ICEBERG-REST` (have deprecation for `ICEBERG_REST` then we can remove it in the next release?

MonkeyCanCode:
@flyrain mind take another look? Then I do the needed change for `ICEBERG_REST` to `ICEBERG-REST` (be backward compatible)

site/content/in-dev/unreleased/command-line-interface.md
flyrain: I tested this against the `main` branch. Passing `--catalog-connection-type ICEBERG` currently fails with the following error:

```
yufei@Yufeis-MacBook-Pro-2 polaris % ./polaris --profile dev catalogs create --type external --catalog-connection-type ICEBERG --default-base-location file://temp/c1 --storage-type file --catalog-authentication-type BEARER --catalog-bearer-token my-token-xxx c1
usage: ...
polaris catalogs create: error: argument --catalog-connection-type: invalid choice: 'iceberg' (choose from 'hadoop', 'iceberg-rest', 'hive')
```

Given that we’ve already been using `ICEBERG_REST` on the server side, I think it makes sense to keep it as-is to avoid introducing a breaking change. Sorry for the confusion earlier, I didn’t mean to suggest changing it to `ICEBERG-REST`.

For the CLI, one option could be to standardize on uppercase enum-style values (e.g. `ICEBERG_REST`, `HADOOP`, `HIVE`) for consistency with other arguments.

What do you think? @MonkeyCanCode @HonahX

MonkeyCanCode: Got it. Thanks for testing. Let me look into this tomorrow and fix this.

MonkeyCanCode: This should now be resolved as the issue is in `client/python/apache_polaris/cli/constants.py`. On the code base, it is enum as what @HonahX was mentioning earlier but we only ref by value in the helper function. Thus, I updated the helper used by options parser to use the right value now.

flyrain (approved):
LGTM. Thanks @MonkeyCanCode !

PR #3210, opened 2025-12-05
Make each task attempt run in a dedicated CDI request context

dimas-b:
Currently, tasks inherit the CDI context from the requests that submitted them, but run asynchronously. Therefore, if the original request context ends, the task may not be able to use the expired beans for that context.

This change makes each task run in its own dedicated CDI request context with `RealmContext` explicitly propagated in `TaskExecutorImpl`.

Test-only error handlers are added to `TaskExecutorImpl` to facilitate detecting task errors during CI.

Fixes #3203

## Checklist
- [ ] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [ ] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [ ] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3224, opened 2025-12-06
feat: pass principal name as part of aws subscoped credentials session

### LINKED ISSUE #3196: Pass principal name as part of aws subscoped credentials session name

We are trying to monitor how different principals use the catalog. While Polaris listeners enable us to audit table reads on table level, we also need information regarding physical storage reads for each principal, for example to keep track of the s3 costs incurred by each. This seems to be impossible at the moment when using vended credentials as all subscoped credentials are produced by the same parent role and session name for each subscoped credential is a simple hardcoded value - "PolarisAwsCredentialsStorageIntegration".

It would be ideal to use principal name as part of the session name so that the final value would look something like this "Polaris_TestPrincipal" (probably best to keep the prefix short as session name length is capped at 64 chars). This would enable users to use aws monitoring tools like CloudTrail to get better information about physical s3 reads.

_No response_

### Additional context

_No response_

tokoko:
PR changes role session name of temporary credentials generated for s3 to contain principal name. The goal is to simplify audit of storage access with credentials generated by Polaris. PolarisPrincipal is injected in `StorageAccessConfigProvider`, used as part of a cache key and then value propagated through the call chain. Azure and Gcp integration classes also accept PolarisPrincipal, but the values are ignored for now.

This will probably also result in relatively increased amount of sts calls as credential requests for the same table by different principals will no longer hit the same cache.

Fixes #3196

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -450,4 +450,43 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
               "If set to true (default), allow credential vending for external catalogs. Note this requires ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING to be true first.")
           .defaultValue(true)
           .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Integer> AZURE_TIMEOUT_MILLIS =
+      PolarisConfiguration.<Integer>builder()
+          .key("AZURE_TIMEOUT_MILLIS")
+          .description(
+              "Timeout in milliseconds for Azure API requests. "
+                  + "Prevents indefinite blocking when Azure endpoints are slow or unresponsive. "
+                  + "Used internally by Azure storage integration for credential vending and other operations.")
+          .defaultValue(15000)
+          .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Integer> AZURE_RETRY_COUNT =
+      PolarisConfiguration.<Integer>builder()
+          .key("AZURE_RETRY_COUNT")
+          .description(
+              "Number of retry attempts for Azure API requests. "
+                  + "Uses exponential backoff with jitter to handle transient failures.")
+          .defaultValue(3)
+          .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Integer> AZURE_RETRY_DELAY_MILLIS =
+      PolarisConfiguration.<Integer>builder()
+          .key("AZURE_RETRY_DELAY_MILLIS")
+          .description(
+              "Initial delay in milliseconds before first retry for Azure API requests. "
+                  + "Delay doubles with each retry (exponential backoff).")
+          .defaultValue(2000)
+          .buildFeatureConfiguration();
+
+  public static final FeatureConfiguration<Double> AZURE_RETRY_JITTER_FACTOR =
+      PolarisConfiguration.<Double>builder()
+          .key("AZURE_RETRY_JITTER_FACTOR")
+          .description(
+              "Jitter factor (0.0 to 1.0) applied to retry delays for Azure API requests. "
+                  + "The jitter is applied as a random percentage of the computed exponential backoff delay. "
+                  + "For example, 0.5 means up to 50%% random jitter will be added to each retry delay. "
+                  + "Helps prevent thundering herd when multiple requests fail simultaneously.")
+          .defaultValue(0.5)
+          .buildFeatureConfiguration();
 }
--- a/.github/workflows/check-md-link.yml
+++ b/.github/workflows/check-md-link.yml
@@ -39,7 +39,7 @@ jobs:
     - uses: actions/checkout@master
     - name: Setup test environment
       uses: ./.github/actions/setup-test-env
-    - uses: gaurav-nelson/github-action-markdown-link-check@5c5dfc0ac2e225883c0e5f03a85311ec2830d368 # v1
+    - uses: tcort/github-action-markdown-link-check@e7c7a18363c842693fadde5d41a3bd3573a7a225
       with:
         use-quiet-mode: 'yes'
         config-file: '.github/workflows/check-md-link-config.json'
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,7 @@ helm-doc-generate: DEPENDENCIES := helm-docs
 helm-doc-generate: check-dependencies ## Generate Helm chart documentation
 	@echo "--- Generating Helm documentation ---"
 	@helm-docs --chart-search-root=helm
-	@cp helm/polaris/README.md site/content/in-dev/unreleased/helm.md
+	@python3 helm/polaris/tools/prepare_helm_readme.py helm/polaris/README.md site/content/in-dev/unreleased/helm.md
 	@echo "--- Helm documentation generated and copied ---"
 
 helm-unittest: DEPENDENCIES := helm
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -62,6 +62,7 @@ dependencies {
     api(project(":polaris-persistence-nosql-api"))
     api(project(":polaris-persistence-nosql-impl"))
     api(project(":polaris-persistence-nosql-benchmark"))
+    api(project(":polaris-persistence-nosql-metastore-types"))
     api(project(":polaris-persistence-nosql-correctness"))
     api(project(":polaris-persistence-nosql-cdi-common"))
     api(project(":polaris-persistence-nosql-cdi-quarkus"))
--- a/extensions/auth/opa/tests/build.gradle.kts
+++ b/extensions/auth/opa/tests/build.gradle.kts
@@ -43,7 +43,7 @@ dependencies {
 
   // Test container dependencies
   intTestImplementation(platform(libs.testcontainers.bom))
-  intTestImplementation("org.testcontainers:junit-jupiter")
+  intTestImplementation("org.testcontainers:testcontainers-junit-jupiter")
   intTestImplementation(project(":polaris-container-spec-helper"))
 }
 
--- a/getting-started/ceph/.env.example
+++ b/getting-started/ceph/.env.example
@@ -1,0 +1,32 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+LANG=en_US.utf8                                     # Default system locale used inside containers
+TZ=UTC                                              # Timezone used inside containers
+DASHBOARD_PORT=8443                                 # Port for Ceph Dashboard
+INTERNAL_DASHBOARD_PORT=8443                        # Internal port for Ceph Dashboard
+RGW_PORT=8080                                       # Port for Rados Gateway
+MON_IP=$(hostname -i)                               # IP address of the monitor
+RGW_ACCESS_KEY=POLARIS123ACCESS                     # Access key for Polaris S3 user
+RGW_SECRET_KEY=POLARIS456SECRET                     # Secret key for Polaris S3 user
+FSID=b2f59c4b-5f14-4f8c-a9b7-3b7998c76a0e           # Unique cluster identifier (use `uuidgen` to regenerate)
+OSD_UUID_1=80505106-0d32-4777-bac9-3dfc901b1273     # Unique OSD identifier (use `uuidgen` to regenerate)
+S3_ENDPOINT_URL=http://rgw1:7480                    # Internal endpoint for S3-compatible RGW service
+S3_REGION=us-east-1                                 # S3 region name
+S3_POLARIS_BUCKET=polaris-storage                   # Default S3 bucket name for Polaris storage
--- a/getting-started/ceph/README.md
+++ b/getting-started/ceph/README.md
@@ -1,0 +1,148 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+ 
+   http://www.apache.org/licenses/LICENSE-2.0
+ 
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Getting Started with Apache Polaris and Ceph
+
+## Overview
+
+This guide describes how to spin up a **single-node Ceph cluster** with **RADOS Gateway (RGW)** for S3-compatible storage and configure it for use by **Polaris**.
+
+This example cluster is configured for basic access key authentication only.
+It does not include STS (Security Token Service) or temporary credentials.
+All access to the Ceph RGW (RADOS Gateway) and Polaris integration uses static S3-style credentials (as configured via radosgw-admin user create).
+
+Spark is used as a query engine. This example assumes a local Spark installation.
+See the [Spark Notebooks Example](../spark/README.md) for a more advanced Spark setup.
+
+## Starting the Example
+
+Before starting the Ceph + Polaris stack, you’ll need to configure environment variables that define network settings, credentials, and cluster IDs.
+
+The services are started **in sequence**:
+1. Monitor + Manager
+2. OSD
+3. RGW
+4. Polaris
+
+Note: this example pulls the `apache/polaris:latest` image, but assumes the image is `1.2.0-incubating` or later. 
+
+### 1. Copy the example environment file
+```shell
+cp .env.example .env
+```
+
+### 2. Start monitor and manager
+```shell
+docker compose up -d mon1 mgr
+```
+
+### 3. Start OSD
+```shell
+docker compose up -d osd1
+```
+
+### 4. Start RGW
+```shell
+docker compose up -d rgw1
+```
+#### Check status
+```shell
+docker exec --interactive --tty ceph-mon1-1 ceph -s
+```
+You should see something like:
+```yaml
+cluster:
+  id:     b2f59c4b-5f14-4f8c-a9b7-3b7998c76a0e
+  health: HEALTH_WARN
+          mon is allowing insecure global_id reclaim
+          1 monitors have not enabled msgr2
+          6 pool(s) have no replicas configured
+
+services:
+  mon: 1 daemons, quorum mon1 (age 49m)
+  mgr: mgr(active, since 94m)
+  osd: 1 osds: 1 up (since 36m), 1 in (since 93m)
+  rgw: 1 daemon active (1 hosts, 1 zones)
+```
+
+### 5. Create bucket for Polaris storage
+```shell
+docker compose up -d setup_bucket
+```
+
+### 6. Run Polaris service
+```shell
+docker compose up -d polaris
+```
+
+### 7. Setup polaris catalog
+```shell
+docker compose up -d polaris-setup
+```
+
+## 8. Connecting From Spark
+
+```shell
+bin/spark-sql \
+    --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.9.0,org.apache.iceberg:iceberg-aws-bundle:1.9.0 \
+    --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions \
+    --conf spark.sql.catalog.polaris=org.apache.iceberg.spark.SparkCatalog \
+    --conf spark.sql.catalog.polaris.type=rest \
+    --conf spark.sql.catalog.polaris.io-impl=org.apache.iceberg.aws.s3.S3FileIO \
+    --conf spark.sql.catalog.polaris.uri=http://localhost:8181/api/catalog \
+    --conf spark.sql.catalog.polaris.token-refresh-enabled=true \
+    --conf spark.sql.catalog.polaris.warehouse=quickstart_catalog \
+    --conf spark.sql.catalog.polaris.scope=PRINCIPAL_ROLE:ALL \
+    --conf spark.sql.catalog.polaris.header.X-Iceberg-Access-Delegation="" \
+    --conf spark.sql.catalog.polaris.credential=root:s3cr3t \
+    --conf spark.sql.catalog.polaris.client.region=irrelevant \
+    --conf spark.sql.catalog.polaris.s3.access-key-id=POLARIS123ACCESS \
+    --conf spark.sql.catalog.polaris.s3.secret-access-key=POLARIS456SECRET
+```
+
+Note: `s3cr3t` is defined as the password for the `root` user in the `docker-compose.yml` file.
+
+Note: The `client.region` configuration is required for the AWS S3 client to work, but it is not used in this example
+since Ceph does not require a specific region.
+
+## 9. Running Queries
+
+Run inside the Spark SQL shell:
+
+```
+spark-sql (default)> use polaris;
+Time taken: 0.837 seconds
+
+spark-sql ()> create namespace ns;
+Time taken: 0.374 seconds
+
+spark-sql ()> create table ns.t1 as select 'abc';
+Time taken: 2.192 seconds
+
+spark-sql ()> select * from ns.t1;
+abc
+Time taken: 0.579 seconds, Fetched 1 row(s)
+```
+## Lack of Credential Vending
+
+Notice that the Spark configuration does not contain a `X-Iceberg-Access-Delegation` header.
+This is because example cluster does not include STS (Security Token Service) or temporary credentials. 
+
+The lack of STS API is represented in the Catalog storage configuration by the 
+`stsUnavailable=true` property.
--- a/getting-started/ceph/ceph-conf/ceph.conf
+++ b/getting-started/ceph/ceph-conf/ceph.conf
@@ -1,0 +1,41 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+[global]
+fsid = b2f59c4b-5f14-4f8c-a9b7-3b7998c76a0e
+mon_initial_members = mon1
+mon_host = mon1
+auth_cluster_required = cephx
+auth_service_required = cephx
+auth_client_required = cephx
+osd_pool_default_size = 1
+osd_pool_default_min_size = 1
+osd_pool_default_pg_num = 333
+osd_crush_chooseleaf_type = 1
+mon_allow_pool_size_one= true
+
+[mon.mon1]
+mon_data = /var/lib/ceph/mon/ceph-mon1
+mon_rocksdb_min_wal_logs = 1
+mon_rocksdb_max_total_wal_size = 64M
+mon_rocksdb_options = max_background_compactions=4;max_background_flushes=2
+
+[client.rgw1]
+host = ceph-rgw1
+rgw_frontends = civetweb port=7480
--- a/getting-started/ceph/docker-compose.yml
+++ b/getting-started/ceph/docker-compose.yml
@@ -1,0 +1,209 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+services:
+
+  mon1:
+    image: quay.io/ceph/ceph:v19.2.3
+    entrypoint: "/bin/sh"
+    command:
+      - "-c"
+      - >-
+        set -ex;
+        mkdir -p /var/lib/ceph/bootstrap-osd;
+        mkdir -p /var/lib/ceph/osd/ceph-0;
+        ceph-authtool --create-keyring /var/lib/ceph/tmp/ceph.mon.keyring --gen-key -n mon. --cap mon 'allow *';
+        ceph-authtool --create-keyring /etc/ceph/ceph.client.admin.keyring \
+          --gen-key -n client.admin \
+          --cap mon 'allow *' --cap osd 'allow *' --cap mgr 'allow *' --cap mds 'allow *';
+        ceph-authtool --create-keyring /var/lib/ceph/bootstrap-osd/ceph.keyring \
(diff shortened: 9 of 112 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -81,20 +82,24 @@ public StorageAccessConfig getSubscopedCreds(
...
       @Nonnull Set<String> allowedWriteLocations,
+      @Nonnull PolarisPrincipal polarisPrincipal,
       Optional<String> refreshCredentialsEndpoint) {
     int storageCredentialDurationSeconds =
         realmConfig.getConfig(STORAGE_CREDENTIAL_DURATION_SECONDS);
     AwsStorageConfigurationInfo storageConfig = config();
     String region = storageConfig.getRegion();
     String accountId = storageConfig.getAwsAccountId();
     StorageAccessConfig.Builder accessConfig = StorageAccessConfig.builder();
+    String roleSessionName = "polaris-" + polarisPrincipal.getName();
```
dimas-b: I'd propose to cover this with a feature flag (defaulting to enabled + CHANGELOG entry). I imagine in some cases Polaris administrators may not want to expose principal names to external systems, which STS may be.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -49,13 +50,16 @@ public class StorageAccessConfigProvider {
...
   private final StorageCredentialCache storageCredentialCache;
   private final StorageCredentialsVendor storageCredentialsVendor;
+  private final PolarisPrincipal polarisPrincipal;
 
   @Inject
   public StorageAccessConfigProvider(
       StorageCredentialCache storageCredentialCache,
-      StorageCredentialsVendor storageCredentialsVendor) {
+      StorageCredentialsVendor storageCredentialsVendor,
+      PolarisPrincipal polarisPrincipal) {
```
dimas-b, line 59: I wonder how it works for async tasks 🤔 Currently `PolarisPrincipal` is [derived](https://github.com/apache/polaris/blob/fa472299fdf86736822b5fb60c0cfe22633603ac/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java#L203) from `SecurityContext`. The latter is available only in the context on HTTP requests, but tasks run in independent contexts now (#3210).

I imagine `TaskExecutorImpl` may have to propagate the `PolarisPrincipal` in a similar way it propagates the Realm ID 🤔

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -123,6 +125,7 @@ public StorageAccessConfig getOrGenerateSubScopeCreds(
             allowListOperation,
             allowedReadLocations,
             allowedWriteLocations,
+            polarisPrincipal,
```
dimas-b: This naturally broadens the range of key values for _all_ storage provides, but only STS use cases benefit from the value. Other storage backends will see a reduction in cache efficiency without any change in credential vending behaviour.

If you agree to my proposal of a feature flag, then even for STS the cache efficiency may be reduced (when the flag is off) without any benefit.

WDYT?

I do not known of an easy solution for this, though. In my mind, ideally, storage integrations implementation should provide a converter method, condensing all available parameters to a sub-set of used parameters (based on type and config), which would later be used for the cache key... however, this will probably require a lot of refactoring.

dimas-b (commented):
Thanks for your contribution, @tokoko !

Overall the PR looks pretty good to me with some comments as noted below.

Given that it affects credential vending, I'd propose to also send a `dev` email about it for visibility (a custom for major changes in Polaris). Also, a CHANGELOG entry would be good to have for this.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
dimas-b: Indeed... CI results (a fragment):
```
:polaris-runtime-service:test org.apache.polaris.service.catalog.iceberg.IcebergCatalogRelationalNoEntityCacheTest

java.lang.IllegalStateException: No REST request in progress	
	at org.jboss.resteasy.reactive.server.injection.ContextProducers.getContext(ContextProducers.java:149)	
		•••
	at org.apache.polaris.service.config.ServiceProducers.polarisPrincipal(ServiceProducers.java:197)	
		•••
	at org.apache.polaris.core.storage.cache.StorageCredentialCacheKey.of(StorageCredentialCacheKey.java:77)	
	at org.apache.polaris.core.storage.cache.StorageCredentialCache.getOrGenerateSubScopeCreds(StorageCredentialCache.java:122)	
		•••
```
```
:polaris-runtime-service:test org.apache.polaris.service.it.RestCatalogFileIntegrationTest

java.lang.AssertionError: 	
Expecting code not to raise a throwable but caught	
  "java.util.concurrent.CompletionException: java.lang.IllegalStateException: No REST request in progress	
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)	
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)	
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1807)	
	at io.smallrye.context.impl.wrappers.SlowContextualRunnable.run(SlowContextualRunnable.java:19)	
	at org.jboss.threads.EnhancedViewExecutor$EnhancedViewExecutorRunnable.run(EnhancedViewExecutor.java:496)	
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$15.runWith(VertxCoreRecorder.java:645)	
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)	
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)	
	at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)	
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)	
	at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:11)	
	at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:11)	
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)	
	at java.base/java.lang.Thread.run(Thread.java:1583)	
Caused by: java.lang.IllegalStateException: No REST request in progress	
	at org.jboss.resteasy.reactive.server.injection.ContextProducers.getContext(ContextProducers.java:149)	
	at org.jboss.resteasy.reactive.server.injection.ContextProducers.securityContext(ContextProducers.java:143)	
	at org.jboss.resteasy.reactive.server.injection.ContextProducers_ProducerMethod_securityContext_Te5N_EppHq79q8NJD1SHmgfMtgk_Bean.create(Unknown Source)	
	at org.jboss.resteasy.reactive.server.injection.ContextProducers_ProducerMethod_securityContext_Te5N_EppHq79q8NJD1SHmgfMtgk_Bean.create(Unknown Source)	
	at io.quarkus.arc.impl.CurrentManagedContext$1.get(CurrentManagedContext.java:111)	
	at io.quarkus.arc.impl.CurrentManagedContext$1.get(CurrentManagedContext.java:107)	
	at io.quarkus.arc.generator.Default_jakarta_enterprise_context_RequestScoped_ContextInstances.ch44(Unknown Source)	
	at io.quarkus.arc.generator.Default_jakarta_enterprise_context_RequestScoped_ContextInstances.computeIfAbsent(Unknown Source)	
	at io.quarkus.arc.impl.CurrentManagedContext.getIfActive(CurrentManagedContext.java:107)	
	at io.quarkus.arc.impl.ClientProxies.getSingleContextDelegate(ClientProxies.java:30)	
	at jakarta.ws.rs.core.ContextProducers_ProducerMethod_securityContext_Te5N_EppHq79q8NJD1SHmgfMtgk_ClientProxy.arc$delegate(Unknown Source)	
	at jakarta.ws.rs.core.ContextProducers_ProducerMethod_securityContext_Te5N_EppHq79q8NJD1SHmgfMtgk_ClientProxy.getUserPrincipal(Unknown Source)	
	at org.apache.polaris.service.config.ServiceProducers.polarisPrincipal(ServiceProducers.java:197)	
	at org.apache.polaris.service.config.ServiceProducers_ProducerMethod_polarisPrincipal_zDmpifKRoFToLBmCZPPPdL3Hxns_Bean.create(Unknown Source)	
	at org.apache.polaris.service.config.ServiceProducers_ProducerMethod_polarisPrincipal_zDmpifKRoFToLBmCZPPPdL3Hxns_Bean.create(Unknown Source)	
	at io.quarkus.arc.impl.CurrentManagedContext$1.get(CurrentManagedContext.java:111)	
	at io.quarkus.arc.impl.CurrentManagedContext$1.get(CurrentManagedContext.java:107)	
	at io.quarkus.arc.generator.Default_jakarta_enterprise_context_RequestScoped_ContextInstances.ch8(Unknown Source)	
	at io.quarkus.arc.generator.Default_jakarta_enterprise_context_RequestScoped_ContextInstances.computeIfAbsent(Unknown Source)	
	at io.quarkus.arc.impl.CurrentManagedContext.getIfActive(CurrentManagedContext.java:107)	
	at io.quarkus.arc.impl.ClientProxies.getSingleContextDelegate(ClientProxies.java:30)	
	at org.apache.polaris.core.auth.ServiceProducers_ProducerMethod_polarisPrincipal_zDmpifKRoFToLBmCZPPPdL3Hxns_ClientProxy.arc$delegate(Unknown Source)	
	at org.apache.polaris.core.auth.ServiceProducers_ProducerMethod_polarisPrincipal_zDmpifKRoFToLBmCZPPPdL3Hxns_ClientProxy.getName(Unknown Source)	
	at org.apache.polaris.core.storage.cache.StorageCredentialCacheKey.of(StorageCredentialCacheKey.java:77)	
	at org.apache.polaris.core.storage.cache.StorageCredentialCache.getOrGenerateSubScopeCreds(StorageCredentialCache.java:122)	
	at org.apache.polaris.core.storage.cache.ServiceProducers_ProducerMethod_storageCredentialCache_hzAWPa00ffa2II6zBfUMmDXk9AQ_ClientProxy.getOrGenerateSubScopeCreds(Unknown Source)	
	at org.apache.polaris.service.catalog.io.StorageAccessConfigProvider.getStorageAccessConfig(StorageAccessConfigProvider.java:120)	
	at org.apache.polaris.service.catalog.io.StorageAccessConfigProvider_ClientProxy.getStorageAccessConfig(Unknown Source)	
	at org.apache.polaris.service.task.TaskFileIOSupplier.apply(TaskFileIOSupplier.java:64)	
	at org.apache.polaris.service.task.TaskFileIOSupplier_ClientProxy.apply(Unknown Source)	
	at org.apache.polaris.service.task.TableCleanupTaskHandler.handleTask(TableCleanupTaskHandler.java:96)	
	at org.apache.polaris.service.task.TaskExecutorImpl.handleTask(TaskExecutorImpl.java:209)	
	at org.apache.polaris.service.task.TaskExecutorImpl.handleTaskWithTracing(TaskExecutorImpl.java:256)	
	at org.apache.polaris.service.task.TaskExecutorImpl_Subclass.handleTaskWithTracing$$superforward(Unknown Source)	
	at org.apache.polaris.service.task.TaskExecutorImpl_Subclass$0.apply(Unknown Source)	
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)	
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:62)	
	at io.quarkus.arc.impl.ActivateRequestContextInterceptor.invoke(ActivateRequestContextInterceptor.java:124)	
	at io.quarkus.arc.impl.Activ
ateRequestContextInterceptor.aroundInvoke(ActivateRequestContextInterceptor.java:33)	
	at io.quarkus.arc.impl.ActivateRequestContextInterceptor_Bean.intercept(Unknown Source)	
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)	
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:30)	
	at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:27)	
	at org.apache.polaris.service.task.TaskExecutorImpl_Subclass.handleTaskWithTracing(Unknown Source)	
	at org.apache.polaris.service.task.TaskExecutorImpl.lambda$tryHandleTask$2(TaskExecutorImpl.java:165)	
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)	
	... 11 more	
"	
	at org.apache.polaris.service.task.TaskErrorHandler.assertNoTaskExceptions(TaskErrorHandler.java:65)	
		•••
```

tokoko: From what I managed to find out, async tasks that touch storage use `TaskFileIOSupplier` which is itself a RequestScoped bean and has `StorageAccessConfigProvider` injected.

Do you mean you can't rely on any RequestScoped bean for async tasks or is it just `SecurityContext`?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
tokoko: Makes sense, do you think it should be a storage access config or an application-level feature flag?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
tokoko: If the feature flag is application-level, then we could check it while constructing a cache key, but not really a fan of that option.



> storage integrations implementation should provide a converter method,

yeah, this would be a big refactor. currently cache lookup is done become storage integration is loaded, we'd have to switch the order first. Also they are too far apart from one another. The simpler solution might be to delegate cache key generation to integrations themselves.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
dimas-b: It's `SecurityContext` - it is only available for HTTP-based requests. Please see how Realm IDs are handled in `TaskExecutorImpl` (they also used to depend on `SecurityContext`, but now they do not).

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: From my POV it is not a functional concern, but an administrator preference, I'd use per-realm `FeatureConfiguration` (authentication that produces the principal is configured per-realm too).

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: We could use two flags: 1) whether to propagate the principal at all 2) whether to use the principal name in STS calls.

The first flag would be useful for avoiding excessive cache key proliferation. The second one - as I commented in another thread.

The first flag does add code complexity, but it could be an interim solution until the cache can be handled more holistically. I believe it is important to preserve cache efficiency in deployments that do not need to pass principals to storage.

... but of course, if you have an alternative solution, please share :slightly_smiling_face:

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
tokoko: After quite a bit of digging, looks like the core issue here is that a lot of tests are set up and executed in a way that http request context is not present during test execution. It's also not trivial to inject it in some way because we're using quarkus augmentors for auth and seems like SecurityContext injection doesn't really get along with that. In the end, I ended up switching PolarisPrincipal bean to acquire context from CurrentIdentityAssociation instead of SecurityContext. Added a testing Augmentor which injects root principal in CurrentIdentityAssociation. I still have a couple of integration test to pin down though (some use multiple principals in the span of a single test which is hard to replicate by injection).

@adutra I saw most of auth stuff was your doing.. can you take a look at this as well? If it seems like a worthwhile change, I will modify remaining failing tests as well.

tokoko: @dimas-b fixed async task behavior as well, similar to realm id example. One thing I changed was injecting `PolarisPrincipalHolder` as a field variable rather than a constructor arg. RequestScoped RealmContextHolder is injected into the constructor of ApplicationScoped TaskExecutorImpl. Doesn't that mean always a single realm context will be used?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
tokoko: so far I added a single feature flag called `INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` that is checked both by cache and aws integration. couldn't really justify to myself a second flag that would do essentially the same thing. The only downside I see is that as of now someone might cause key proliferation even if using azure or gcp, but the feature is by default disabled, so I don't think another feature flag to guard against that is necessary. Let me know what you think.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
dimas-b: Quarkus is smart enough to inject proxies when request/application scope boundaries are crossed. The proxies will switch to the right bean at execution time.

My personal preference would be to adhere to clean request/application scope boundaries, but it is not strictly required.

Update: What I wrote above is "in general". CDI stuff in this PR LGTM :+1: No need to change.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: Having only `INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` is ok for now, IMHO :+1:

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -79,6 +79,18 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL =
+      PolarisConfiguration.<Boolean>builder()
+          .key("INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL")
```
dimas-b, line 84: Please add a CHANGELOG entry for this (usual practice for feature flags).

Also, please open a `dev` ML thread for visibility, since this change is pretty fundamental. If you prefer I could email `dev` from my side.

tokoko: I did send an email, but I guess I wasn't fully subscribed to the list at the time and it got lost. I'll resend it now.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -89,12 +92,22 @@ public StorageAccessConfig getSubscopedCreds(
     String accountId = storageConfig.getAwsAccountId();
     StorageAccessConfig.Builder accessConfig = StorageAccessConfig.builder();
 
+    boolean includePrincipalNameInSubscopedCredential =
+        realmConfig.getConfig(FeatureConfiguration.INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL);
+
+    String roleSessionName =
+        includePrincipalNameInSubscopedCredential
+            ? "polaris-" + polarisPrincipal.getName()
+            : "polaris";
```
dimas-b: Please keep `PolarisAwsCredentialsStorageIntegration` as default for backward compatibility. I know, it's not a huge change, but we'd better not change old behaviours unless we have to.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -49,6 +50,9 @@ public interface StorageCredentialCacheKey {
   Set<String> allowedWriteLocations();
 
   @Value.Parameter(order = 7)
+  Optional<String> principalName();
```
dimas-b: Since this change is in `polaris-core`, would you mind keeping old indexes for old values. This is to simplify adjusting to the changed code in downstream projects.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
adutra: @tokoko indeed it's tricky to inject `SecurityIdentity` or `ContainerRequestContext` in some places. The former is tricky because we use [lazy authentication](https://quarkus.io/guides/security-proactive-authentication) so it's safer to use `CurrentIdentityAssociation`; the latter is tricky because it's not a normal CDI bean (it always appears resolvable, even when it's not).

I overcame both issues recently with some low-level trickery, feel free to take a similar approach if it's helpful:

https://github.com/apache/polaris/blob/64b13a9915274a04e82a12c9fdcd095d12797ca0/runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java

adutra: And you are absolutely right that some tests instantiate CDI beans manually – these are a pain to deal with.

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
```diff
@@ -97,6 +123,54 @@ public void testGetSubscopedCreds(String scheme) {
...
+
+    Mockito.when(stsClient.assumeRole(Mockito.isA(AssumeRoleRequest.class)))
+        .thenAnswer(
+            invocation -> {
+              assertThat(invocation.getArguments()[0])
+                  .isInstanceOf(AssumeRoleRequest.class)
+                  .asInstanceOf(InstanceOfAssertFactories.type(AssumeRoleRequest.class))
+                  .returns(externalId, AssumeRoleRequest::externalId)
+                  .returns(roleARN, AssumeRoleRequest::roleArn)
+                  .returns("polaris-principal", AssumeRoleRequest::roleSessionName);
```
dimas-b: nit using `polaris-test-principal` might be easier to follow. In current form is it not very obvious (without browsing through code quite a few lines away from here) whether the `principal` part is some sort of a code constant of comes from the test parameters

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -160,9 +165,17 @@ public void addTaskHandlerContext(long taskEntityId, CallContext callContext) {
       return CompletableFuture.failedFuture(e);
     }
     String realmId = callContext.getRealmContext().getRealmIdentifier();
+
+    PolarisPrincipal principalClone =
+        PolarisPrincipal.of(
```
dimas-b: Going through the builder would more future-proof: `ImmutablePolarisPrincipal.builder().from(polarisPrincipal)`... The `from` method could be promoted to the `PolarisPrincipal` interface too.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -70,6 +72,8 @@ public class TaskExecutorImpl implements TaskExecutor {
   private final MetaStoreManagerFactory metaStoreManagerFactory;
   private final TaskFileIOSupplier fileIOSupplier;
   private final RealmContextHolder realmContextHolder;
+  @Inject PolarisPrincipalHolder polarisPrincipalHolder;
+  @Inject PolarisPrincipal polarisPrincipal;
```
dimas-b: I suppose it works in practice, but how it works might be unclear to some readers 😅 Would you mind adding comments to describe how we propagate the principal into a the holder in a different CDI context?

nit: I'd prefer to stick to constructor-based injection for all args. I believe it will not have any functional effect, but will maintain code uniformity (and allow fields to be `final`, which simplifies reasoning about code behaviour).

runtime/service/src/test/java/org/apache/polaris/service/catalog/generic/PolarisGenericTableCatalogRelationalTest.java
```diff
@@ -23,6 +23,6 @@
 import org.apache.polaris.service.catalog.Profiles;
 
 @QuarkusTest
-@TestProfile(Profiles.DefaultProfile.class)
+@TestProfile(Profiles.AuthProfile.class)
```
dimas-b: Is this change necessary? What is broken without it (just trying to understand).

dimas-b:
CI appears to be stuck :shrug:

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: thx!

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
tokoko: nice, thanks. meant to come back to this, looked for something like that initially, but wrote a quick and dirty workaround when I was unable to find it.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
tokoko: thanks, I did find that snippet and used it to rewrite PolarisPrincipal bean to resolve principal from `CurrentIdentityAssociation`. I added testing [Augmentor](https://github.com/apache/polaris/pull/3224/files#diff-cf7a98222bd86bbc1284b28281baa8428f879c3648c49271ffe4e87efec88704) that does dummy authentication the same way actual auth is done.

runtime/service/src/test/java/org/apache/polaris/service/catalog/generic/PolarisGenericTableCatalogRelationalTest.java
tokoko: apparently it's not :smile:

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -201,6 +200,7 @@ public Map<String, String> getConfigOverrides() {
           .put("polaris.features.\"ALLOW_TABLE_LOCATION_OVERLAP\"", "true")
           .put("polaris.features.\"LIST_PAGINATION_ENABLED\"", "true")
           .put("polaris.behavior-changes.\"ALLOW_NAMESPACE_CUSTOM_LOCATION\"", "true")
+          .put("polaris.test.rootAugmentor.enabled", "true")
```
dimas-b, line 203: @adutra : do you have any comment on this? As for me, I'm ok with this approach, as I do not have a better alternative off the top of my head 😅

polaris-core/src/test/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheTest.java
```diff
@@ -135,10 +144,133 @@ public void testCacheHit() {
...
+        Set.of("s3://bucket1/path", "s3://bucket2/path"),
+        Set.of("s3://bucket3/path", "s3://bucket4/path"),
+        anotherPolarisPrincipal,
+        Optional.empty());
+    Assertions.assertThat(storageCredentialCache.getEstimatedSize()).isEqualTo(1);
+  }
+
+  @Test
+  public void testCacheHitForAnotherPrincipal() {
+    Mockito.when(storageCredentialsVendor.getRealmConfig())
```
adnanhemani, line 215: nit: Since this test case and the one above it have the same core logic, it would be best to extract a helper to make these tests more DRY. But not a big deal if there are no further changes being made to this PR, a simple refactor later can take care of this.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/PolarisPrincipalHolder.java
```diff
@@ -0,0 +1,65 @@
...
+    PolarisPrincipal setPrincipal = polarisPrincipal.get();
+
+    if (setPrincipal != null) {
+      return setPrincipal;
+    }
+
+    SecurityIdentity identity =
+        currentIdentityAssociation.getDeferredIdentity().subscribeAsCompletionStage().getNow(null);
+
+    Principal userPrincipal = identity.getPrincipal();
```
adutra, line 54: This will cause a NPE when there is no `SecurityIdentity` available.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
adutra: I think we can achieve the same by simply producing this bean from this profile class:

```java
public static class Profile extends Profiles.DefaultProfile {
  @Override public Map<String, String> getConfigOverrides() {...}
  
  @Produces @RequestScoped produceTestAugmentor() { return new RootPrincipalAugmentor(...); }
}
```

But the current approach is fine too.

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
```diff
@@ -97,6 +124,54 @@ public void testGetSubscopedCreds(String scheme) {
...
+    StorageAccessConfig storageAccessConfig =
+        new AwsCredentialsStorageIntegration(
+                AwsStorageConfigurationInfo.builder()
+                    .addAllowedLocation(warehouseDir)
+                    .roleARN(roleARN)
+                    .externalId(externalId)
+                    .build(),
+                stsClient)
+            .getSubscopedCreds(
+                PRINCIPAL_INCLUDER_REALM_CONFIG,
```
adnanhemani, line 172: nit: Similar comment as above. I also had to play "spot-the-difference" between this test and the one above it so a quick comment on this line would make that very obvious to a reader that we are overriding behavior here since it is not clear in the test name.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/PolarisPrincipalHolder.java
adnanhemani: Question: when can this happen? Mostly in test situations?

adutra: Or when the bean is produced before the authentication phase is complete.

tokoko: should I return null in that case or just add a `diagnostics.checkNotNull` check?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
tokoko: The idea was to change it later from a simple toggle to something that will hold more information. Something like `polaris.test.augmentor.identity`: `test_principal:test_principal_role`. I ended up not requiring it, but it could still come in handy later...

adutra: Sounds good 👍 

pinging @XN137 for awareness since I vaguely recall you were working on turning a bunch of manual setup code into proper Quarkus tests with CDI enabled.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/PolarisPrincipalHolder.java
adutra: Maybe throw `UnsatisfiedResolutionException`? That would be more meaningful imho than NPE.

tokoko: Added a null check beforehand.

polaris-core/src/test/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheTest.java
tokoko: done

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
tokoko: I did my best to make the purpose clearer

adnanhemani (approved):
LGTM! Thanks for this great contribution!!

dimas-b (approved):
The PR in its current state can be merged from my POV. The only thing I'd like to clarify (probably in the community sync meeting) is whether we may want to refactor the call paths leading to `StorageAccessConfig` to avoid having too many details in method parameters (i.e leverage CDI more).

adnanhemani:
Agreed with @dimas-b that this change may not be 110% ideal, but I think it is definitely good enough to merge and take care of the refactor in a later PR to at least unblock the use case!

PR #3226, opened 2025-12-06
Add location properties from TableMetadata into Table entity internalProps

singhpk234:
### About the change 

Presently since not all the location properties are stored in the entity one has to do complete loadTable table even for existing use cases such as credential endpoint which in present scenario since polaris doesn't store the complete metadata copy inside the persistence is kind of expensive. 

With this we will have a way to do cred vending, in future (to remote sign) without going to object store. 
Note if we take dependency on this we would have to think of backfill but for step 1 it seems reasonable step, would love to know what other folks think

https://github.com/apache/polaris/blob/8ba112eb09a76707ad346082e4f90436c7ba7c93/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java#L616-L625

co-author : @adutra 

## Checklist
- [X] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1657,7 +1657,18 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
   private static Map<String, String> buildTableMetadataPropertiesMap(TableMetadata metadata) {
     Map<String, String> storedProperties = new HashMap<>();
+    // Location specific properties
     storedProperties.put(IcebergTableLikeEntity.LOCATION, metadata.location());
+    if (metadata.properties().containsKey(TableProperties.WRITE_DATA_LOCATION)) {
+      storedProperties.put(
+          IcebergTableLikeEntity.USER_SPECIFIED_WRITE_DATA_LOCATION_KEY,
+          metadata.properties().get(TableProperties.WRITE_DATA_LOCATION));
+    }
+    if (metadata.properties().containsKey(TableProperties.WRITE_METADATA_LOCATION)) {
+      storedProperties.put(
+          IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY,
+          metadata.properties().get(TableProperties.WRITE_METADATA_LOCATION));
+    }
     storedProperties.put(
         IcebergTableLikeEntity.FORMAT_VERSION, String.valueOf(metadata.formatVersion()));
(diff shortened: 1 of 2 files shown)


adutra:
I might be wrong but it seems the locations are taken into account for overlap in `BasePolarisTableOperations#doCommit`:

https://github.com/apache/polaris/blob/fe57ea1b2a68a8e95b38705b05656431a164a744/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1465-L1472

The problem is that later on in that same method, the internal properties map is computed:

https://github.com/apache/polaris/blob/fe57ea1b2a68a8e95b38705b05656431a164a744/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1506

But the map currently doesn't include `write.data.path` or `write.metadata.path`:

https://github.com/apache/polaris/blob/fe57ea1b2a68a8e95b38705b05656431a164a744/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1658-L1702

So, the only way to check those locations is to load the table from the object store.

adutra:
> However, making use of the new entity properties can be tricky as Polaris code will have to deal with old entities (on upgrade), which may not have the new properties.

Correct – code would have to be lenient to the properties being absent. For request signing at least, the idea would be to load the table if the properties are absent, then update the entity, so that next time the properties are there.

singhpk234:
> So, the only way to check those locations is to load the table from the object store.

precisely, the idea is to have location props in sync with what is present in the object store, to avoid redundant call as presently the loadCreds is doing unneccessary call to load from objectstore.
I see this PR atleast (which is first step to what i wanna achieve) as an extension to Mike's recent PR my POV
- https://github.com/apache/polaris/pull/2735

when we plan to use it we need to handle backfill, Here is the discuss in Alex on the same 
- https://github.com/apache/polaris/pull/2280#discussion_r2505974770

Re: opening mail list thread, sure happy to facilitate the discussion over MT on how to handle back fill, me and Alex had some idea which Alex mentioned, on this context but defintely more inputs are certainly welcomed.

dimas-b (approved):
This change LGTM purely from the java code perspective and I do not mind merging it.

TBH, I lost context with any related `dev` discussions. @singhpk234 : Could you update on the status?

From my POV, I do not require a `dev` discussion, but it might be worth opening for the sake of awareness of other Polaris developers.

dimas-b:
Related `dev` discussion: https://lists.apache.org/thread/5obdjwhpk7y6kxkqst4m47pocr4k2fk3

flyrain (approved):

singhpk234:
Thanks for the review everyone, i merged this since there were no objection, i will followup with the credential endpoint optimisation !

PR #3231, opened 2025-12-07
Runtime/service: move getConfig() down to `IcebergCatalogHandler`

snazy:
All catalog specific functionality is implemented in `IcebergCatalogHandler`, whereas `IcebergCatalogAdapter` is meant to act as a "REST wrapper" to it.

This change moves the implementation of `getConfig` down to the handler, no functional changes.


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1265,4 +1314,37 @@ public void close() throws Exception {
       closeable.close();
     }
   }
+
+  public ConfigResponse getConfig() {
+    // 'catalogName' is taken from the REST request's 'warehouse' query parameter..
```
adutra:
```suggestion
    // 'catalogName' is taken from the REST request's 'warehouse' query parameter.
```

PR #3237, opened 2025-12-08
NoSQL: Metastore implementation

snazy:


persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
```diff
@@ -0,0 +1,168 @@
...
+
+  static ContentIdentifier identifierFor(TableIdentifier identifier) {
+    return identifier(identifier.namespace().levels(), identifier.name());
+  }
+
+  static ContentIdentifier identifierFor(Namespace namespace) {
+    return identifier(namespace.levels());
+  }
+
+  static ContentIdentifier identifierFromLocationString(String locationString) {
```
dennishuo, line 76: All the other factory methods are mostly self-explanatory, but this one could use some javadocs explaining what `locationString` is, expected format, etc.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
```diff
@@ -0,0 +1,168 @@
...
+    var builder = builder();
+    var len = locationString.length();
+    var off = -1;
+    for (var i = 0; i < len; i++) {
+      var c = locationString.charAt(i);
+      checkArgument(c >= ' ', "Control characters are forbidden in locations");
```
dennishuo: Is this really intended to support all chars with ascii codepoint > ' ' or is there a more intuitive regex? In particular, though I don't know all ascii codes offhand, a cursory search seems to indicate for example that `127` is the "Delete control character" even though it comes after all the basic symbols and letters.

Maybe `Character.isJavaIdentifierPart` will help provide a better standard constrained set while also making it easier to reason about when reading the code.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
```diff
@@ -0,0 +1,168 @@
...
+      if (c == '/' || c == '\\') {
```
dennishuo, line 83: Is this for linux vs windows path separators? Can we just use `java.nio.Path.of(...).iterator()`  instead?

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
```diff
@@ -0,0 +1,1347 @@
...
+          byName.put(nameKey, objRef(catalogObj));
+          byId.put(idKey, nameKey);
+
+          if (existing == null) {
+            // created
+            return new ChangeResult.CommitChange<>(
+                new CreateCatalogResult(catalog, catalogAdminRole));
+          }
+          // retry
+          return new ChangeResult.NoChange<>(new CreateCatalogResult(ENTITY_ALREADY_EXISTS, null));
```
dennishuo: I'm not sure this fallthrough should exist; if returning `ENTITY_ALREADY_EXISTS` it's interpreted by the upper stack to mean the same *name* is already occupied but that this isn't actually an idempotent retry attempt (e.g. it had a different id).

Looks like line 232 already captures that case.

If we reach this line does it mean we're accruing a bunch of updates into the updatable indexes that we'll just discard when it comes time to "commit"?

If we want consistent behavior with the other metastore impls we would basically need a way to return `NoChange` but with a *success* `CreateCatalogResult` maybe in an else branch on line 230 (`else if (catalogObj.stableId() == catalog.getId())`).

It's possible the idempotent-retry scenario that code is trying to catch isn't actually possible anymore though, not sure.

I *think* the `checkState(!byId.contains(idKey` on line 246 means that this line wouldn't actually represent stableId == id though, in which case reaching this line should maybe be an IllegalStateException indicating some kind of corrupted state? Unless I'm misunderstanding the state of the index.

Basically we'd reach here if the `byName` already exists, and `catalogObj.stableId() != catalog.getId()`, but somehow `catalogObj == null` so that line 230 doesn't already exit early.

Is it expected for the byName index to not be atomic with what's visible in `persistence.fetch`?

I think all these subtleties could use some code comments to clarify if this is working as intended.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationResults.java
```diff
@@ -0,0 +1,148 @@
...
+  void addPolicyIndexKeyToRemove(IndexKey indexKey) {
+    policyIndexKeysToRemove.add(indexKey);
+  }
+
+  void entityResult(PolarisBaseEntity entity) {
+    add(new EntityResult(entity));
+    anyChange = true;
+  }
+
+  void entityResultNoChange(PolarisBaseEntity entity) {
```
dennishuo: Could use some javadoc comments explaining the difference between `entityResultNoChange` and `unchangedEntityResult`

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/privs/GrantTriplet.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+package org.apache.polaris.persistence.nosql.metastore.privs;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import org.apache.polaris.core.entity.PolarisEntityCore;
+
+/**
+ * Represents the triplet of catalog-ID, entity-ID and type-code plus a reverse-or-key marker.
+ * String representations of this type are used as ACL names and "role" names.
```
dennishuo: Looks like there's an implicit encoding convention here regarding the `r/` and `d/` convention. Could use javadoc examples here or else a javadoc link to wherever else the convention is defined other than just in the code's parsing logic in here.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
snazy: yup, added

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
snazy: This `identifierFromLocationString` is used (only) for location-overlap handling, assuming that the storage locations are already validated. Adding assumptions here about the actual format of object-store specific conventions or valid file-path chars feels a bit too risky?

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
snazy: Similar to the above comment: we'd assume that it's a file-system path (NIO Path delegates to file-system implementations), but object-store conventions do not match the the restrictions/validations.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
snazy: Both (index and persistence state) are always consistent. This particular one is literally an impossible case, so I removed it.

This (and other) functions have to expect cases when call sites (management API handlers) pass already existing names and/or already existing entity IDs. For example, a call site can pass a non-existing catalog name with an already used entity ID. Would be easier if the IDs were generated exclusively by the implementations, not the call sites.

> I think the checkState(!byId.contains(idKey on line 246 means that this line wouldn't actually represent stableId == id though, in which case reaching this line should maybe be an IllegalStateException indicating some kind of corrupted state?
Actually not a corrupted state, but the API call sites passing an already used catalog ID into the "createCatalog" call.

When I wrote this function, I thought about how a retry could actually happen, given that the existing implementations assume entity-ID-equality (the reason for the TODO at the beginning of the function).

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationResults.java
snazy: Ah, two functions for the same thing. Removed one.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/privs/GrantTriplet.java
snazy: docs added

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
```diff
@@ -0,0 +1,593 @@
...
+    var renameOrMove =
+        entityParentId != originalObj.parentStableId()
+            || !entity.getName().equals(originalObj.name());
+
+    if (renameOrMove) {
+      if (entity.cannotBeDroppedOrRenamed()) {
+        mutationResults.entityResult(ENTITY_CANNOT_BE_RENAMED);
+        return;
+      }
+      if (!byName.remove(originalNameKey)) {
```
dennishuo: Does `byName` represent a fixed overall version for the duration of this method or is it a read-through to the "head" version of persistence?

My understanding is that it's a fixed/snapshot that then collects updates to apply at the end, is that correct? As in, the "immutable object index" from your design doc?

If so, would that mean that since we already validated `byName.get(originalNameKey)` on line 326 that the `remove` here *must* succeed, or else we have a code bug? In that case this should probably be an assert/IllegalStateException rather than returning `ENTITY_NOT_FOUND` because some assumptions must have gone wrong.

Or if `byName` is indeed subject to concurrent removal from somewhere else please add some comments around here indicating that it's possible and maybe additional comments around any other accesses in this method to the index to help explain how we safely reconcile the changing state.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
```diff
@@ -0,0 +1,593 @@
...
+    var entityIdKey = IndexKey.key(entity.getId());
+    var originalNameKey = byId.get(entityIdKey);
+
+    var entityParentId = entity.getParentId();
+
+    if (originalNameKey == null) {
+      mutationResults.entityResult(ENTITY_NOT_FOUND);
+      return;
+    }
+    var originalRef = byName.get(originalNameKey);
```
dennishuo, line 331: What are the scenarios that can lead to this `byName` not having this `originalNameKey` that we just got from `byId`? Would it indicate `byId` and `byName` getting out of sync somehow?

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
```diff
@@ -0,0 +1,593 @@
...
+    MemoizedIndexedAccess memoizedIndexedAccess) {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MutationAttempt.class);
+
+  public static ObjBase objForChangeComparison(
+      PolarisBaseEntity entity,
+      Optional<PolarisPrincipalSecrets> currentSecrets,
+      ObjBase originalObj) {
+    return mapToObj(entity, currentSecrets)
+        .updateTimestamp(originalObj.createTimestamp())
```
dennishuo: Is this intentional to use `originalObj.createTimestamp` as the `updateTimestamp` for the change-comparison object? If so, definitely needs some extensive comments explaining why.

If the intent is to canonicalize `updateTimestamp` to not be relevant for semantic comparison, it might be better to set it to 0L or something to indicate intentional/obvious missing values rather than having comparison objects floating around that have the `createTimestamp` written in.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
```diff
@@ -0,0 +1,593 @@
...
+    CommitterState<? extends ContainerObj, MutationResults> state,
+    UpdatableIndex<ObjRef> byName,
+    UpdatableIndex<IndexKey> byId,
+    UpdatableIndex<Change> changes,
+    UpdatableIndex<EntityIdSet> locations,
+    MemoizedIndexedAccess memoizedIndexedAccess) {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MutationAttempt.class);
+
+  public static ObjBase objForChangeComparison(
```
dennishuo, line 95: Please add javadoc comment for this one explaining intent and how it intends to canonicalize whichever fields for comparison.

Is this to detect cases where a *real* update actually happens to result in an entity whose full contents are unchanged, or is there a common concurrency situation that is expected to cause "no change" mutations (e.g. on some kind of retry?)

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
dennishuo: That raises an interesting point - we'd ideally be able to incorporate storage-type-specific canonicalization logic.

In this code for example, presumably the `\\` is specifically "windows local filesystem" use cases (which really should only be very limited test/dev scenarios where overlap-checks probably aren't relevant anyways; any "sensitive" production use cases would be really weird to run on a single-node Windows environment with engine and Polaris catalog in the same place, or else some NFS mount).

But for a nio Filesystem, it's actually *correct* to canonicalize with real filesystem syntax, such as collapsing "foo/./././././bar" into "foo/bar" or "foo/bar/.." becoming "foo" when it comes to "overlap detection" because overlap detection is precisely supposed to prevent security issues by sharing directories. So it becomes important to avoid creative syntax circumventing our check.

And yet, as mentioned before, the "local filesystem" use cases where paths have fancy syntactic interpretations also happen to be the cases where overlap-check probably doesn't actually matter.

I guess it's fine to leave as-is for now, but custom parsing of paths is probably not something we want to keep as ad-hoc inline logic here in the long-term. It might be worth adding a comment or filing a github issue to call out that this logic isn't necessarily authoritative for any future developers who might read it as gospel.

snazy: I think this is probably the wrong place to introduce a canonization. This piece of the code just deals with what's been configured. The _configured_ values should have already been canonicalized.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
snazy: Yea - checkState is better here.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
snazy: There's no thinkable scenario actually.

snazy: Ideally I'd prefer to not have the byId index at all and just refer to entities by name.
But that requires a bigger change.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
snazy: Oops, should have been `updateTimestamp`.
Sure, this could be optimized, but the call sites would become more complicated.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
snazy: It's to detect whether a update is actually changing something.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/ContentIdentifier.java
snazy: Overall, we should probably revisit the location-overlap-checks.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/mutation/MutationAttempt.java
dennishuo: Does this mean that if someone does craft an update request which happens to be a no-op, that we'll skip the database write, meaning:

1. The updateTimestamp won't change
2. the entityVersion won't change

but event and audit-logging features would still show a successful update at the new timestamp?  I can't remember if the current AtomicOperationMetaStore/TransactionalMetaStores do anything similar, or if this is divergent behavior.

dennishuo (approved):
Overall LGTM. I'm not sure about how the "skip no-op updates" optimization if it's divergent from the existing persistence impls and what it would mean for discrepancies between audit logs and `updatedTimestamp` in persistence, but I'm okay just revisiting that discussion later or in a followup

snazy:
Thanks for the reviews!

> not sure about how the "skip no-op updates" optimization

I think, it depends on the, say, "interpretation" of what's an update.

Issue #550, opened 2024-12-13
Support for GCP service account impersonation


collado-mike:
### Is your feature request related to a problem? Please describe.

GCP supports [service account impersonation](https://cloud.google.com/iam/docs/service-account-impersonation), so that given credentials for a service account, it's possible to impersonate a different service account, given that the first is granted privileges to do so. The `GcpStorageConfigurationInfo` catalog configuration [here](https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpStorageConfigurationInfo.java#L39-L40) actually has a `gcpServiceAccount` field that we never use when vending GCS storage credentials. We can use the code in https://cloud.google.com/iam/docs/create-short-lived-credentials-direct#create-access to assume the target service account, then generate a short-lived token that has the target service account's privileges subscoped to the table location during the credential vending process.

talatuyarer:
@collado-mike you can close this.

dimas-b:
@collado-mike : heads up: I removed this issue from the 1.4.0 milestone to avoid blocking the release on it (since it looks like it can be closed).

PR #3246, opened 2025-12-09
Add GCP service account impersonation for credentials.

talatuyarer:
This will close this bug: https://github.com/apache/polaris/issues/550

cc @flyrain

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
@@ -64,6 +64,9 @@ public class GcpCredentialsStorageIntegration
     extends InMemoryStorageIntegration<GcpStorageConfigurationInfo> {
   private static final Logger LOGGER =
       LoggerFactory.getLogger(GcpCredentialsStorageIntegration.class);
+  public static final String SERVICE_ACCOUNT_PREFIX = "projects/-/serviceAccounts/";
+  public static final String IMPERSONATION_SCOPE =
+      "https://www.googleapis.com/auth/devstorage.read_write";
 
   private final GoogleCredentials sourceCredentials;
   private final HttpTransportFactory transportFactory;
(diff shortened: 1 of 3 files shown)


polaris-core/src/test/java/org/apache/polaris/service/storage/gcp/GcpCredentialsStorageIntegrationTest.java
```diff
@@ -309,6 +315,63 @@ public void testRefreshCredentialsEndpointIsReturned() throws IOException {
...
+        true,
+        Set.of("gs://bucket/path"),
+        Set.of("gs://bucket/path"),
+        Optional.empty());
+
+    Mockito.verify(mockIamClient)
+        .generateAccessToken(
+            Mockito.argThat(
+                request ->
+                    request.getName().equals("projects/-/serviceAccounts/" + serviceAccount)
```
adnanhemani: nit: Would recommend moving these variables out to the main class as public variables and then referencing them here from there so that we don't have magic variables and it will be cleaner to maintain if there are any changes in the future!

adnanhemani (approved):
Looks great, @talatuyarer!! Thank you for this :) One small nit - not a requirement before merging (we can refactor after), but if you are making any changes, would be nice to take care of this in the same go.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -123,6 +134,46 @@ public StorageAccessConfig getSubscopedCreds(
     return accessConfig.build();
   }
 
+  /**
+   * Returns the credential to be used as the source for downscoping. If a specific service account
+   * is configured, it impersonates that account first.
+   */
+  private GoogleCredentials getBaseCredentials() {
+    if (config().getGcpServiceAccount() != null) {
```
dimas-b, line 145: Interesting that this config value was defined before, but not used 🤔 

Does anyone have insight into why it was this way?

Technically, now that `GcpServiceAccount` config values start affecting Polaris behaviour, this could be a breaking change in existing deployments that may have (possibly accidental and/or incorrect) values in that config property.

I do believe that such a situation is unlikely, so it should be fine to proceed with this PR. However, it probably deserves mentioning as a `Change` in CHANGELOG.md

flyrain: Looks like this is in the CLI doc for a while, https://polaris.apache.org/in-dev/unreleased/command-line-interface/#catalogs
```
      --service-account  (Only for GCS) The service account to use when connecting to GCS
```
Agreed with @dimas-b, not particularly a big concern, but +1 on adding this to the CHANGELOG.md.

flyrain:
Thanks a lot for adding this, @talatuyarer! Thanks @dimas-b @adnanhemani for the review!

PR #3250, opened 2025-12-10
Introduce `ExternalAuthenticator` and `isExternal` flag on `PolarisCredential`

sungwy:
Draft PR to introduce an `External` mode in the PolarisPrincipal. The goal of this PR is to introduce an authentication flow that skips principal resolution in the metastore for enhanced compatibility with authentication -> authorization flows that rely on external IDPs and PDPs.

This PR is motivated by @adutra 's [mailing list proposal](https://lists.apache.org/thread/xosrbtkvnt9d9nbrfbzvx1cr0d2jb57x)

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
@@ -41,6 +41,7 @@ import org.apache.polaris.core.entity.PolarisEntityId;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisGrantRecord;
 import org.apache.polaris.core.entity.PolarisPrivilege;
+import org.apache.polaris.core.entity.PrincipalEntity;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.ResolvedPolarisEntity;
 import org.apache.polaris.core.persistence.cache.EntityCache;
@@ -744,6 +745,20 @@ public class Resolver {
   private ResolverStatus resolveCallerPrincipalAndPrincipalRoles(
       List<ResolvedPolarisEntity> toValidate) {
 
+    if (isExternalPrincipal()) {
+      PrincipalEntity externalPrincipal = createExternalPrincipalEntity();
+      this.resolvedCallerPrincipal =
+          new ResolvedPolarisEntity(
+              diagnostics,
+              externalPrincipal,
+              List.of(),
+              externalPrincipal.getGrantRecordsVersion());
+      this.resolvedEntriesById.put(
+          this.resolvedCallerPrincipal.getEntity().getId(), this.resolvedCallerPrincipal);
+      this.resolvedCallerPrincipalRoles = List.of();
+      return new ResolverStatus(ResolverStatus.StatusEnum.SUCCESS);
+    }
+
     // resolve the principal, by name or id
     this.resolvedCallerPrincipal =
         this.resolveByName(toValidate, PolarisEntityType.PRINCIPAL, polarisPrincipal.getName());
@@ -751,6 +766,21 @@ public class Resolver {
     // if the principal was not found, we can end right there
     if (this.resolvedCallerPrincipal == null
         || this.resolvedCallerPrincipal.getEntity().isDropped()) {
+      if (isExternalPrincipal()) {
+        // For external principals we do not maintain principal entities in the metastore,
+        // so synthesize a placeholder entry and continue without resolving grants.
+        PrincipalEntity externalPrincipal = createExternalPrincipalEntity();
+        this.resolvedCallerPrincipal =
+            new ResolvedPolarisEntity(
+                diagnostics,
+                externalPrincipal,
+                List.of(),
+                externalPrincipal.getGrantRecordsVersion());
+        this.resolvedEntriesById.put(
+            this.resolvedCallerPrincipal.getEntity().getId(), this.resolvedCallerPrincipal);
+        this.resolvedCallerPrincipalRoles = List.of();
+        return new ResolverStatus(ResolverStatus.StatusEnum.SUCCESS);
+      }
       return new ResolverStatus(ResolverStatus.StatusEnum.CALLER_PRINCIPAL_DOES_NOT_EXIST);
     }
 
@@ -762,6 +792,17 @@ public class Resolver {
 
     // total success
     return new ResolverStatus(ResolverStatus.StatusEnum.SUCCESS);
+  }
+
+  private boolean isExternalPrincipal() {
+    return Boolean.parseBoolean(polarisPrincipal.getProperties().getOrDefault("external", "false"));
+  }
+
+  private PrincipalEntity createExternalPrincipalEntity() {
+    return new PrincipalEntity.Builder()
+        .setName(polarisPrincipal.getName())
+        .setProperties(polarisPrincipal.getProperties())
+        .build();
   }
 
   /**
--- a/runtime/service/src/main/java/org/apache/polaris/service/auth/ExternalAuthenticator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/auth/ExternalAuthenticator.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.auth;
+
+import io.smallrye.common.annotation.Identifier;
+import jakarta.enterprise.context.RequestScoped;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import org.apache.iceberg.exceptions.NotAuthorizedException;
+import org.apache.polaris.core.auth.PolarisPrincipal;
+
+/**
+ * Authenticator for principals supplied by an external identity provider.
+ *
+ * <p>Builds an in-memory {@link PolarisPrincipal} directly from credential claims without any
+ * persistence lookups.
+ */
+@RequestScoped
+@Identifier("external")
+public class ExternalAuthenticator implements Authenticator {
+
+  @Override
+  public PolarisPrincipal authenticate(PolarisCredential credentials) {
+    PolarisCredential externalCredentials = ensureExternal(credentials);
+
+    String principalName = externalCredentials.getPrincipalName();
+    Long principalId = externalCredentials.getPrincipalId();
+    if (principalName == null && principalId != null) {
+      principalName = String.valueOf(principalId);
+    }
+    if (principalName == null) {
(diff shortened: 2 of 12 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -744,13 +745,42 @@ private ResolverStatus resolvePaths(
...
+    }
+
     // resolve the principal, by name or id
     this.resolvedCallerPrincipal =
         this.resolveByName(toValidate, PolarisEntityType.PRINCIPAL, polarisPrincipal.getName());
 
     // if the principal was not found, we can end right there
     if (this.resolvedCallerPrincipal == null
         || this.resolvedCallerPrincipal.getEntity().isDropped()) {
+      if (isExternalPrincipal()) {
```
dimas-b, line 769: can this be `true` without returning on line 759?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -744,13 +745,42 @@ private ResolverStatus resolvePaths(
   private ResolverStatus resolveCallerPrincipalAndPrincipalRoles(
       List<ResolvedPolarisEntity> toValidate) {
 
+    if (isExternalPrincipal()) {
+      PrincipalEntity externalPrincipal = createExternalPrincipalEntity();
```
dimas-b, line 749: This is an entity without a (real) ID... Some it's a kind of ephemeral entity... Would it be possible to avoid creating it at all?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -764,6 +794,17 @@ private ResolverStatus resolveCallerPrincipalAndPrincipalRoles(
     return new ResolverStatus(ResolverStatus.StatusEnum.SUCCESS);
   }
 
+  private boolean isExternalPrincipal() {
+    return Boolean.parseBoolean(polarisPrincipal.getProperties().getOrDefault("external", "false"));
```
dimas-b, line 798: WDYT about making `isExternal()` a method of `PolarisPrincipal`?

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/NoneTokenBrokerFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        @Override
+        public TokenResponse generateFromClientSecrets(
+            String clientId,
+            String clientSecret,
+            String grantType,
+            String scope,
+            TokenType requestedTokenType) {
+          return TokenResponse.of(OAuthError.invalid_request);
```
dimas-b, line 54: I'm not sure about `invalid_request`... it may actually be well-formed... How about `unsupported_grant_type`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -744,13 +745,42 @@ private ResolverStatus resolvePaths(
   private ResolverStatus resolveCallerPrincipalAndPrincipalRoles(
       List<ResolvedPolarisEntity> toValidate) {
 
+    if (isExternalPrincipal()) {
```
dimas-b, line 748: Cf. #3228

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
sungwy: It cannot - this was before I moved the condition upto line 748 to avoid principal resolution altogether if it is external. I'll remove this check here to remove redundancy

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
sungwy: I think that's worth considering. I put it into `PolarisPrincipal` as a part of the property for now because it was already available, but I think adding it as an attribute or adding a class method that infers the property value is up for discussion

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/NoneTokenBrokerFactory.java
sungwy: That's a great suggestion - thank you!

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
sungwy: That's a good question - I'm a bit confused about the role of `PrincipalEntity`. We do have a public method for `getResolvedCallerPrincipal()` that returns `resolvedCallerPrincipal`. It's not being used anywhere in the codebase today, but I thought it'd be safe to populate it as it requires it to be `@Nonnull`: https://github.com/apache/polaris/blob/23ba2a05adc9c75f3e72aaf2ca370b4886964328/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java#L272C19-L280

dimas-b: Conceptually, IMHO, the principal _entity_ should only be required for AuthZ checks (which do not actually require it if it's external). So, any intermediate code that requires it may need to be refactored... but TBH, I do not remember that code very well :sweat_smile:

runtime/service/src/main/java/org/apache/polaris/service/auth/ExternalAuthenticator.java
```diff
@@ -0,0 +1,77 @@
...
+ * <p>Builds an in-memory {@link PolarisPrincipal} directly from credential claims without any
+ * persistence lookups.
+ */
+@RequestScoped
+@Identifier("external")
+public class ExternalAuthenticator implements Authenticator {
+
+  @Override
+  public PolarisPrincipal authenticate(PolarisCredential credentials) {
+    PolarisCredential externalCredentials = ensureExternal(credentials);
```
sungwy, line 42: instead of overriding the Principal to be external here, maybe we could validate that it is external and raise an exception if it isn't instead. Then we have a clearer separation of responsibility between:
1. detecting whether a `PolarisCredential`'s origin was internal or external
2. and determining if a `PolarisPrincipal` should be internal, federated or external.

dimas-b: @adutra : WDYT?

runtime/service/src/main/java/org/apache/polaris/service/auth/PolarisCredential.java
```diff
@@ -49,4 +59,10 @@ static PolarisCredential of(
 
   /** The principal roles, or empty if the principal has no roles. */
   Set<String> getPrincipalRoles();
+
+  /** True if the credential represents an external principal. */
+  @Value.Default
+  default boolean isExternal() {
```
adutra, line 65: Should this property have a default value at all? I think it's safer if the code building the credential explicitly decides if it's external or not.

adutra: Also thinking forward, maybe we should have an enum `PolarisPrincipalType` + method `getPrincipalType()` ?

If federated principals are implemented eventually, it would be cumbersome to have two methods `isExternal` and `isFederated` as they are mutually exclusive.

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/broker/NoneTokenBrokerFactory.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.polaris.core.PolarisCallContext;
+import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
+import org.apache.polaris.service.auth.PolarisCredential;
+import org.apache.polaris.service.auth.internal.service.OAuthError;
+import org.apache.polaris.service.types.TokenType;
+
+/** A no-op token broker factory used when authentication is delegated to an external IdP. */
+@ApplicationScoped
+@Identifier("none")
+public class NoneTokenBrokerFactory implements TokenBrokerFactory {
```
adutra, line 33: Hmm this token broker existed and was deleted a while ago by #2634. Why are we adding it back?

runtime/service/src/main/java/org/apache/polaris/service/auth/internal/InternalAuthenticationMechanism.java
```diff
@@ -118,7 +117,7 @@ public Uni<ChallengeData> getChallenge(RoutingContext context) {
 
   @Override
   public Set<Class<? extends AuthenticationRequest>> getCredentialTypes() {
-    return Collections.singleton(TokenAuthenticationRequest.class);
+    return Collections.singleton(InternalAuthenticationRequest.class);
```
adutra, line 120: Whoops this is an interesting bug that you fixed 😅 But I wonder how it worked before 🤔

runtime/service/src/main/java/org/apache/polaris/service/auth/ExternalAuthenticator.java
adutra: I think indeed it would be better if this authenticator rejected any non-external `PolarisCredential`.

However, the `DefaultAuthenticator` should not do the opposite, as it is valid to have persisted principals with an external IDP so it should accept any credential, internal or external.

adutra: In fact I'm wondering whether we need this second `Authenticator`.

We could instead introduce a `polaris.authentication.principal.type` setting (overridable per realm). Valid values would be the constants of the `PolarisPrincipalType` enum (e.g. `persisted`, `federated`, `external`).

Then the `DefaultAuthenticator` should be able to handle all kinds of principals. It would simply read this configuration setting to determine whether it should fetch the `PolarisEntity `or not. (It would also, of course, validate that if the principal type is `external` then the `PolarisCredential` must be external too.)

WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
adutra: Same remark: maybe we should expose a `getType` method in `PolarisPrincipal` that returns `PolarisPrincipalType`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
adutra: I think the ideal solution would be to not materialize this entity _at all_ – and very likely, pass an empty set of `activatedEntities` to the `PolarisAuthorizer`.

But I also understand that this may increase the blast radius of this PR, and also interfere with #3228.

So, I think having a temporary entity being created here can be a good compromise, as long as we don't forget to revisit this later.

runtime/service/src/main/java/org/apache/polaris/service/auth/PolarisCredential.java
adutra: Update: maybe not here... a credential can only be internal or external, there is no "federated" credential. Please disregard my comment above.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
adutra: In fact I think that #3427 and `resolvePathsOnly` could help here.

SBylemans:
Hello, I'm looking to use Polaris with a keycloak integration but would appreciate this feature. Any idea in which release this PR would be made available?

sungwy:
Hi @SBylemans - we are working on a refactor in the PolarisAuthorizer that will introduce this functionality.

I'm doing a pass through an RFC doc that will be shared in the dev mailing list, and once the proposal is accepted, we'll progress on with the implementation. I can cross post the RFC here as well when it is ready.

sungwy:
A broader Authorizer refactoring is being tracked through https://github.com/apache/polaris/issues/3779

PR #3252, opened 2025-12-11
Remove runtime-common and defaults from service deps

dimas-b:
The main idea is to avoid exposing RDS (added in #2650) as a primary dependency of runtime-service since it is conceptually only a dependency on leaf modules (like runtime-server, admin and tests).

This is expected to simplify dependency management in downstream projects since runtime-service has a reduced set of transitive dependencies now.

## Checklist

### DIFF
--- a/extensions/auth/opa/tests/build.gradle.kts
+++ b/extensions/auth/opa/tests/build.gradle.kts
@@ -32,6 +32,8 @@ dependencies {
   implementation(project(":polaris-extensions-auth-opa"))
 
   // Include all runtime-service dependencies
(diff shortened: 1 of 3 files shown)


snazy:
Guess it needs some refactoring for JDBC?

jbonofre:
I agree with this. You updated OPA, but I guess that other modules should be updated (including the distributions).

dimas-b:
Yes, sorry about the breakage, I thought it was that simple, but it's not :sweat_smile: I'll update the PR once I figure out the build issues.

PR #3261, opened 2025-12-13
Move integration tests to `polaris-server`

dimas-b:
`@QuarkusIntegrationTest` uses the regular server build with only non-test dependencies. Therefore, those tests naturally belong into the `server` module.

This change is as a prerequisite for #3252 to allow the `polaris-runtime-service` module to avoid having Polaris extensions as runtime dependencies. Extensions will be runtime dependencies only in the `polaris-server` module.

## Checklist

### DIFF
--- a/runtime/server/build.gradle.kts
+++ b/runtime/server/build.gradle.kts
@@ -54,27 +54,28 @@ dependencies {
   implementation(enforcedPlatform(libs.quarkus.bom))
   implementation("io.quarkus:quarkus-container-image-docker")
 
+  testImplementation(enforcedPlatform(libs.quarkus.bom))
   testImplementation("io.quarkus:quarkus-junit5")
 
-  testImplementation(testFixtures(project(":polaris-runtime-service")))
-  testImplementation(project(":polaris-runtime-test-common"))
-  testImplementation(project(":polaris-core"))
-  testImplementation(project(":polaris-api-management-model"))
-  testImplementation(project(":polaris-minio-testcontainer"))
+  intTestImplementation(testFixtures(project(":polaris-runtime-service")))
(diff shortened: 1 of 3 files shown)


runtime/server/build.gradle.kts
```diff
@@ -53,6 +53,32 @@ dependencies {
   // enforce the Quarkus _platform_ here, to get a consistent and validated set of dependencies
   implementation(enforcedPlatform(libs.quarkus.bom))
   implementation("io.quarkus:quarkus-container-image-docker")
+
+  testImplementation("io.quarkus:quarkus-junit5")
```
adutra, line 58: Two questions / doubts here: 
1. Why modify the `testImplementation` configuration here? I don't think it applies to integration tests.
2. Do we need to add `testImplementation(enforcedPlatform(libs.quarkus.bom))`?

dimas-b: oops, my mistake. Thx for catching it... will fix.

dimas-b: I had to make a build config change and move a non-QuarkusIntegrationTest back to `runtime/service`... but I hope the end result is better now :)

runtime/service/src/test/java/org/apache/polaris/service/it/ServiceProducersIT.java
adutra, line 1: Shouldn't we rename this test to something ending with `Test`? It is also oddly designed, the nested class seems unnecessary.

adutra: BTW: the nested test class is not being executed, it is lacking the `@Nested` annotation. All in all, I think this test class could be deleted.

PR #3265, opened 2025-12-14
[doc]: Add doc for helm prod deployment

MonkeyCanCode:
Adding a best practice doc for prod polaris deployment via Helm.

## Checklist

### DIFF
--- a/site/content/in-dev/unreleased/configuring-polaris-for-production/configuring-helm.md
+++ b/site/content/in-dev/unreleased/configuring-polaris-for-production/configuring-helm.md
@@ -53,11 +53,11 @@ Resource requests and limits can be set in the `values.yaml` file:
 ```yaml
 resources:
   requests:
-    memory: "1Gi"
-    cpu: "500m"
+    memory: "8Gi"
+    cpu: "4"
   limits:
-    memory: "2Gi"
-    cpu: "1"
+    memory: "8Gi"
+    cpu: "4"
 ```
 
 Adjust these values based on expected workload and available cluster resources.


site/content/in-dev/unreleased/configuring-polaris-for-production/configuring-helm.md
```diff
@@ -0,0 +1,132 @@
...
+  minReplicas: 2
+  maxReplicas: 5
+  targetCPUUtilizationPercentage: 80
+  targetMemoryUtilizationPercentage: 80
+```
+
+### Pod Topology Spreading
+
+For better fault tolerance, `topologySpreadConstraints` can be used to distribute pods across different nodes, racks, or availability zones. This helps prevent a single infrastructure failure from taking down all Polaris replicas.
+
```
flyrain, line 124: Should we add a warning callout here? Something like:
```
All pods across zones must point to the same persistent metastore (e.g., RDBMS). 
```

flyrain (dismissed):
This looks like a solid addition. Thanks @MonkeyCanCode !

site/content/in-dev/unreleased/configuring-polaris-for-production/configuring-helm.md
MonkeyCanCode: So that should be happen automatically (there is also no option for current chart to use a diff persistent metastore). Same as we increase replica counts where by default it depends on the underlying node scaler (e.g. karpenter). Often time, these node scaler will needed additional configuration to enforce the pods spreading. With this addition, we basically tell node scaler don't just schedule them blindly but auto respect the required constraints.

flyrain (dismissed):

site/content/in-dev/unreleased/configuring-polaris-for-production/configuring-helm.md
```diff
@@ -0,0 +1,132 @@
...
+## Resource Management
+
+For a production environment, it is crucial to define resource requests and limits for the Polaris pods. Resource requests ensure that pods are allocated enough resources to run, while limits prevent them from consuming too many resources on the node.
+
+Resource requests and limits can be set in the `values.yaml` file:
+
+```yaml
+resources:
+  requests:
+    memory: "1Gi"
```
adutra: These are low values. Of course the right values depend on the use case, but I would suggest something higher as an example of "good" production setup: 8Gi - 4 cpus for instance.

adutra (approved):

PR #3267, opened 2025-12-14
NoSQL: reduce heap pressure when running tests

snazy:
Some tests generate a lot of realms, likely one realm per test case. While the amount of data per realm is not much, it is nontheless nice to remove that data immediately (for tests).

The maintenance service, which purges data of eligible realms, cannot be run against the in-memory backend (different JVM).

This change adds a rather "test only" workaround to purge the realm data in the in-memory backend immediately.

### DIFF
--- a/persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStoreManagerFactory.java
+++ b/persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStoreManagerFactory.java
@@ -198,20 +198,28 @@ class NoSqlMetaStoreManagerFactory implements MetaStoreManagerFactory {
 
       var nextStatus =
           switch (existing.status()) {
-            case CREATED, LOADING, INITIALIZING, INACTIVE -> PURGING;
+            case CREATED, LOADING, INITIALIZING, INACTIVE, PURGING ->
+                // PURGING is the state telling the maintenance service to delete the realm data.
+                PURGING;
             case ACTIVE -> INACTIVE;
-            case PURGING ->
-                // TODO this state should really happen during maintenance!!
+            case PURGED ->
+                // The PURGED state is also handled by the maintenance service.
                 PURGED;
-            case PURGED -> PURGED;
           };
+
+      if (nextStatus == existing.status()) {
+        // No status change (PURGING/PURGED), stop here.
+        // Maintenance service will handle the actual deletion of the realm data.
+        break;
+      }
 
       var update = RealmDefinition.builder().from(existing).status(nextStatus).build();
 
       var updated = realmManagement.update(existing, update);
 
-      if (updated.status() == PURGED) {
-        realmManagement.delete(updated);
+      if (updated.status() == PURGING || updated.status() == PURGED) {
+        // Final status (PURGING/PURGED), stop here.
+        // Maintenance service will handle the actual deletion of the realm data.
         break;
       }
     }
--- a/persistence/nosql/realms/store-nosql/src/main/java/org/apache/polaris/persistence/nosql/realms/store/RealmStoreImpl.java
+++ b/persistence/nosql/realms/store-nosql/src/main/java/org/apache/polaris/persistence/nosql/realms/store/RealmStoreImpl.java
@@ -24,6 +24,7 @@ import static java.lang.String.format;
 import static org.apache.polaris.persistence.nosql.api.Realms.SYSTEM_REALM_ID;
 import static org.apache.polaris.persistence.nosql.api.obj.ObjRef.OBJ_REF_SERIALIZER;
 import static org.apache.polaris.persistence.nosql.api.obj.ObjRef.objRef;
+import static org.apache.polaris.persistence.nosql.realms.api.RealmDefinition.RealmStatus.PURGING;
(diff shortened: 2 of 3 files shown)


persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStoreManagerFactory.java
```diff
@@ -198,20 +198,28 @@ BaseResult purgeRealm(String realmId) {
...
+        break;
+      }
+
       var update = RealmDefinition.builder().from(existing).status(nextStatus).build();
 
       var updated = realmManagement.update(existing, update);
 
-      if (updated.status() == PURGED) {
-        realmManagement.delete(updated);
+      if (updated.status() == PURGING || updated.status() == PURGED) {
```
dimas-b, line 220: Do we really need the complex state transition sequence (via `INACTIVE`) with a loop now? Why not transition to the final state in one step?

persistence/nosql/realms/store-nosql/src/main/java/org/apache/polaris/persistence/nosql/realms/store/RealmStoreImpl.java
```diff
@@ -206,7 +211,18 @@ public RealmDefinition update(
...
-    return objToDefinition(realmId, realm.orElseThrow());
+    var result = realm.orElseThrow();
+    if (result.status() == PURGING && "InMemory".equals(backend.type())) {
+      // Handle the test/development case when using the in-memory backend:
+      // In this case there is no chance to run the maintenance service. To remove no longer
+      // necessary data from the Java heap, call the in-memory backend directly.
+      // This is "nice behavior" when running tests.
+      // The only "cost" is that the state PURGING may never be pushed to PURGED, but that is
+      // acceptable.
+      backend.deleteRealms(Set.of(realmId));
```
dimas-b, line 222: This looks awkward... Could we delegate that to `Backend` impl?... or maybe move the special case to `PolarisIntegrationTestFixture`?

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStoreManagerFactory.java
snazy: Realm-managment defines possible realm-states and allowed transitions between those.
The intent, in this "purge realm" case, is to later allow a workflow to first set a realm to `INACTIVE` (aka it's still there, but you cannot use it), allowing users to revert their decision within a couple of days, pretty much what people can do in other SaaS offerings. From `INACTIVE` it can transition to `PURGING`, which means that the management-service takes care of actually deleting all realm data and eventually delete the realm definition itself.

persistence/nosql/realms/store-nosql/src/main/java/org/apache/polaris/persistence/nosql/realms/store/RealmStoreImpl.java
snazy: It is a bit awkward, true. OTOH, `InMemory` is the only backend that doesn't persist anything, so we cannot use the management-service or even the admin-tool here.
`PolarisIntegrationTestFixture` can't be used in all tests, as it's only for Quarkus tests.

This change actually also applies when people try Polaris server w/ the `InMemory` backend.

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStoreManagerFactory.java
dimas-b: Yes, I understand the current state transitions :) however, since in this piece of code the intent is to transition from any non-purged state to `PURGING`, having state transition restrictions in `RealmManagementImpl` seems artificial. This code will transition to the state it wants anyway, so why impede this transition and make it go through a loop?

This is a tangential comment based on general observations... not a blocker for this PR :)

persistence/nosql/realms/store-nosql/src/main/java/org/apache/polaris/persistence/nosql/realms/store/RealmStoreImpl.java
dimas-b: but as far as I can tell `PolarisIntegrationTestFixture` is the only piece of code that calls purge in tests... no other tests or user-driven actions will get to `backend.deleteRealms()` on this line... or did I miss something :thinking:

snazy: You did ;)
There are tests in a lot of other modules, `polaris-nosql-nosql-metastore` and many other modules that do not "know" about those types.

dimas-b: ok, fair enough :)

snazy: Thanks :)

PR #3268, opened 2025-12-14
NoSQL: Metastore maintenance

snazy:
Implementation of the NoSQL meta-store maintenance implementation. It adds the meta-store specific handling to the existing NoSQL maintenance service to purge unreferenced and unneeded data from the database.

### DIFF
--- a/persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
+++ b/persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
@@ -26,8 +26,8 @@ import java.util.Optional;
 import org.apache.polaris.immutables.PolarisImmutable;
 
 /**
- * No SQL persistence implementation of Polaris stores a history of changes per kind of object (principals, principal roles, grants,
- * immediate tasks, catalog roles and catalog state).
+ * No SQL persistence implementation of Polaris stores a history of changes per kind of object
+ * (principals, principal roles, grants, immediate tasks, catalog roles and catalog state).
  *
  * <p>The rules are defined using a <a href="https://github.com/projectnessie/cel-java/">CEL
  * script</a>. The default rules for all kinds of objects are to retain the history for 3 days, for


persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogRetainedIdentifier.java
```diff
@@ -0,0 +1,334 @@
...
+                          if (now >= nextLog) {
+                            LOGGER.info(
+                                "... {} total index entries processed so far, at commit {}",
+                                total,
+                                commit);
+                            nextLog = now + PROGRESS_LOG_INTERVAL_MS;
+                          }
+                        }
+                      },
+                      x -> {});
```
singhpk234, line 254: can we have meaningful name what does `x` denotes ?

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
```diff
@@ -0,0 +1,103 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+import java.util.Optional;
+import org.apache.polaris.immutables.PolarisImmutable;
+
+/**
+ * Polaris stores a history of changes per kind of object (principals, principal roles, grants,
+ * immediate tasks, catalog roles and catalog state).
```
singhpk234: This is not true for all persistence in Apache Polaris.
```suggestion
 * No SQL persistence implementation of Polaris stores a history of changes per kind of object (principals, principal roles, grants,
 * immediate tasks, catalog roles and catalog state).
```

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * <p>The rules are defined using a <a href="https://github.com/projectnessie/cel-java/">CEL
```
singhpk234, line 32: why are we using this library to define the rules, have we considered alternatives ? what kind of rules is required ?

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>Example scripts
+ *
+ * <ul>
+ *   <li>{@code ageDays < 30 || commits <= 10} retains the reference history with at least 10
+ *       commits and commits that are younger than 30 days
+ *   <li>{@code true} retains the whole reference history
+ *   <li>{@code false} retains the most recent commit
+ * </ul>
+ */
+@ConfigMapping(prefix = "polaris.persistence.maintenance.catalog")
```
singhpk234:
```suggestion
@ConfigMapping(prefix = "polaris.persistence.nosql.maintenance.catalog")
```

this is not persistence implementation agnostic 

prev discussion : https://lists.apache.org/thread/vmdpb45j1nmmlhswz7fm87gxgoldmr28
similar callout: https://github.com/apache/polaris/pull/3135#discussion_r2558515915

singhpk234 (commented):
Thanks @snazy !

1. The maintenance configs are not scoped down to NoSql which is required as per concencus, dev list discussion : 
 Prev discussion : https://github.com/apache/polaris/pull/3135#discussion_r2558515915
 dev list discussion: https://lists.apache.org/thread/vmdpb45j1nmmlhswz7fm87gxgoldmr28

2. Unclear why maintenance module should be part of Polaris repo and not Polaris tools ? In Nosql presentation this module is not mentioned either 
https://docs.google.com/presentation/d/1lX2EdvM0SeyuOdO_u1idlWfmnlH3hFE16JEyWo45Bdo/edit?slide=id.p24#slide=id.p24 
can you please open dev list thread for the same

3. Using https://github.com/projectnessie/cel-java/ for rules, why such rules are required in the first place ? why not use https://github.com/google/cel-java directly which is actively maintained by google developers vs using https://github.com/projectnessie/cel-java/  which has just done dependency updates this year  https://github.com/projectnessie/cel-java/pulls?page=8&q=is%3Apr+is%3Aclosed

dimas-b:
> Unclear why maintenance module should be part of Polaris repo and not Polaris tools ?

The end-to-end maintenance workflow naturally includes some "service" code and "tool / CLI" code. This PR contains only the service portion, which has to align with the NoSQL metastore implementation, so it belongs in this repository, IMHO.

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
dennishuo: +1, I know there was some ongoing debate about config keys, but IIRC we did move ahead with the `polaris.persistence.nosql` key prefix in https://github.com/apache/polaris/pull/3135 so I'm assuming additional config keys will follow the same pattern.

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
```diff
@@ -0,0 +1,103 @@
...
+public interface CatalogsMaintenanceConfig {
+
+  String DEFAULT_PRINCIPALS_RETAIN = "false";
+  String DEFAULT_PRINCIPAL_ROLES_RETAIN = "false";
+  String DEFAULT_GRANTS_RETAIN = "false";
+  String DEFAULT_IMMEDIATE_TASKS_RETAIN = "false";
+  String DEFAULT_CATALOGS_HISTORY_RETAIN = "false";
+  String DEFAULT_CATALOG_ROLES_RETAIN = "false";
+  String DEFAULT_CATALOG_POLICIES_RETAIN = "ageDays < 30 || commits <= 1";
+  String DEFAULT_CATALOG_STATE_RETAIN = "ageDays < 30 || commits <= 1";
```
dennishuo, line 71: Is the `ageDays` condition the only thing to distinguish between "staged but not yet committed" states that are being assembled as part of a transaction and old states that were once committed/valid but are expired now?

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
dimas-b: Given the discussion in #3135, I think it is reasonable to use the `polaris.persistence.nosql` config prefix in this PR.

dimas-b:
> Seeing the end-to-end admin flow would also help contextualize the split of "server-side" and "client-side" code related to this maintenance, that @dimas-b talked about

Hypothetically (ATM), the Admin tool could involve the maintenance code to perform its job. 

Other implementations are possible too. It's an open-ended design where downstream projects are free to implement their own triggering mechanisms. I imagine k8s jobs is one of possible options.

snazy:
Thanks all for the helpful reviews!

Documentation of "everything" is definitely planned to be added to the web site once everything is in.

I have renamed the config mapping.

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
snazy: This has nothing to do with "not *yet* committed" data.
This maintenance cleans up stale + unreferenced as well as "unwanted" history. What "unwanted" means is defined by these defaults, which can optionally be [customized by users](https://github.com/apache/polaris/pull/3268/changes/BASE..04742bd59fedcbddaa6419a01914dd07cbf334eb#diff-f96135bd1e400c2ce4aa2de7c173e1e8a7578201818a1be967b014a589a52fb5R39).

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
dimas-b: `projectnessie/cel-java` is an implementation of the CEL spec in java. Unlike Google's impl. it also works with Jackson.

IMHO, CEL is simple and performant, while wielding a decent expressive power. I think it is a reasonable initial option for the rules language. I suggest merging this PR witl CEL for the sake of making progress. It is certainly open for discussion if alternatives are proposed later.

persistence/nosql/persistence/metastore-maintenance/src/main/java/org/apache/polaris/persistence/nosql/metastore/maintenance/CatalogsMaintenanceConfig.java
dimas-b: I do not think this is the "only thing". This rule is just a fairly safe common sense default.

singhpk234:
@snazy  can you please explain how were the feedbacks mentioned here addressed : https://github.com/apache/polaris/pull/3268#pullrequestreview-3576273215 specially request for raising this is the devlist since this module hasn't been discussed anywhere.

I am not sure until we have resolved the feedbacks how can we move forward, if the feedbacks is not addressed for 3 week hence the pr is opened for 3 weeks.

dimas-b:
@singhpk234 : The config concern was in fact addressed. Config names are like `polaris.persistence.nosql.maintenance`. If something was unintentionally missed in config names, let's fix-forward on `main`.

Re: the location of this code - I commented above. I do believe it explained the intention. A possible user-side tool is provided in #3395

Re: CEL, please see my comment [above](https://github.com/apache/polaris/pull/3268#discussion_r2670139185). This is why Nessie's CEL impl. is convenient. Please note that `projectnessie/cel-java` is a pure CEL impl. It has nothing to do with the Nessie Catalog. If this is still a concern, let's discuss (as I noted above) and possible use alternatives.

singhpk234:
I appreciate the additional context and the link to the user-side tool. However, my primary concern remains the introduction of new constructs—like retention expressions—that don’t currently exist in Polaris. This feels like a significant architectural shift that warrants a broader discussion on the dev list to reach project-wide consensus.

Regarding the cel-java dependency: my concern isn't its origin, but its long-term maintenance health. Given that we have diverging views on this merge (+1 vs -1), I believe the most collaborative path forward is to revert the PR for now. This would allow us to have that broader discussion on the dev list without the pressure of the code already being in the main branch.

dimas-b:
Let's discuss on `dev`

singhpk234:
About the concern or about the revert ?

flyrain:
> If this is still a concern, let's discuss (as I noted above) and possible use alternatives.

It is a real concern. Before the concern was resolved, I don't think the PR should be merged. It's not acceptable to do so in an ASF project.

dimas-b:
I mean any/all issues brought up here are probably best discussed on `dev`. I suppose an email thread per topic would be advisable on `dev`.

flyrain:
@dimas-b , would you mind sharing links about the discussion and conclusion of consensus of the following item?

>  Using https://github.com/projectnessie/cel-java/ for rules, why such rules are required in the first place ? why not use https://github.com/google/cel-java directly which is actively maintained by google developers vs using https://github.com/projectnessie/cel-java/ which has just done dependency updates this year https://github.com/projectnessie/cel-java/pulls?page=8&q=is%3Apr+is%3Aclosed

dimas-b:
@flyrain : my points about these items are available in this PR's comments trail. 

The need for rules is evident from the related java code.

If you prefer to go deeper into these aspects please bring your concerns to the `dev` ML and I'll be happy to discuss further.

dimas-b:
Related `dev` discussion: https://lists.apache.org/thread/znw87l7r1gh3l09pc0hxhj5gyl8vp28d

PR #3269, opened 2025-12-14
Migrate to Jackson mapper builder pattern

snazy:
Mappers and factories are fully immutable objects in Jackson 3. This change is rather a no-op, but migrates the code to use the builder-pattern.

This is only a little building-block for "real" Jackson 3 support, there's more to do and more that's required from other frameworks.

### DIFF
--- a/persistence/nosql/authz/impl/src/test/java/org/apache/polaris/persistence/nosql/authz/impl/TestAclEntryImpl.java
+++ b/persistence/nosql/authz/impl/src/test/java/org/apache/polaris/persistence/nosql/authz/impl/TestAclEntryImpl.java
@@ -18,7 +18,11 @@
  */
 package org.apache.polaris.persistence.nosql.authz.impl;
 
+import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
+import static com.fasterxml.jackson.databind.MapperFeature.DEFAULT_VIEW_INCLUSION;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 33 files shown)


persistence/nosql/authz/impl/src/test/java/org/apache/polaris/persistence/nosql/authz/impl/TestAclEntryImpl.java
```diff
@@ -39,7 +43,12 @@ public class TestAclEntryImpl {
 
   @BeforeAll
   static void setUp() {
-    mapper = new ObjectMapper().findAndRegisterModules();
+    mapper =
+        JsonMapper.builder()
+            .findAndAddModules()
+            .disable(FAIL_ON_UNKNOWN_PROPERTIES)
```
adutra, line 49: These features weren't present in the previous code, why do we need to add those now? There are many other occurrences of this in the PR, so I suppose there has to be some good reason to do so 😄

snazy: Ah, yes.
So in Jackson 2, FAIL_ON_UNKNOWN_PROPERTIES defaults to true, in Jackson 3 it defaults to false.
DEFAULT_VIEW_INCLUSION defaults to true in Jackson 2, but in Jackson 3 it defaults to false.
TL;DR a no-op-change for Jackson 2, but a required change for Jackson 3, to already account for these breaking changes.

dimas-b (approved):

PR #3276, opened 2025-12-15
Relax `client_id`, `client_secret` regex/pattern validation on reset endpoint call

rmannibucau:
client_id/client_secret patterns are validated when calling reset endpoint but the pattern is hardcoded which can be too rigid.
this PR just enables to configure it. 

## Checklist
- [X] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
@@ -136,18 +136,6 @@ public class PolarisServiceImpl
     return Response.status(Response.Status.CREATED).entity(newCatalog).build();
   }
 
-  private void validateClientId(String clientId) {
-    if (!clientId.matches("^[0-9a-f]{16}$")) {
-      throw new IllegalArgumentException("Invalid clientId format");
-    }
-  }
-
-  private void validateClientSecret(String clientSecret) {
-    if (!clientSecret.matches("^[0-9a-f]{32}$")) {
-      throw new IllegalArgumentException("Invalid clientSecret format");
-    }
-  }
-
   private void validateStorageConfig(StorageConfigInfo storageConfigInfo) {
     List<String> allowedStorageTypes =
         realmConfig.getConfig(FeatureConfiguration.SUPPORTED_CATALOG_STORAGE_TYPES);
@@ -304,12 +292,6 @@ public class PolarisServiceImpl
             ? resetPrincipalRequest
             : new ResetPrincipalRequest(null, null);
 
-    if (safeResetPrincipalRequest.getClientId() != null) {
-      validateClientId(safeResetPrincipalRequest.getClientId());
-    }
-    if (safeResetPrincipalRequest.getClientSecret() != null) {
-      validateClientSecret(safeResetPrincipalRequest.getClientSecret());
-    }
     return Response.ok(adminService.resetCredentials(principalName, safeResetPrincipalRequest))
         .build();
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Added `hierarchical` flag to `AzureStorageConfigInfo` to allow more precise SAS token down-scoping in ADLS when
   the [hierarchical namespace](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-namespace)
   feature is enabled in Azure.
+- Relaxed `client_id`, `client_secret` regex/pattern validation on reset endpoint call
 
 ### Changes
 


polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -34,9 +32,6 @@
  * @param <T> The type of the configuration
  */
 public abstract class PolarisConfiguration<T> {
-
-  private static final Logger LOGGER = LoggerFactory.getLogger(PolarisConfiguration.class);
```
dimas-b, line 38: nit: if you find "dead code", it's generally preferable to cleanup in a dedicated PR and avoid mixing cleanup with feature changes :)

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -137,13 +137,15 @@ public Response createCatalog(
   }
 
   private void validateClientId(String clientId) {
-    if (!clientId.matches("^[0-9a-f]{16}$")) {
+    if (!clientId.matches(
+        realmConfig.getConfig(FeatureConfiguration.CREDENTIAL_RESET_CLIENT_ID_PATTERN))) {
```
dimas-b: we should validate the value of `CREDENTIAL_RESET_CLIENT_ID_PATTERN` on startup to avoid deferred RegEx syntax errors (it's an admin mistake, not the API client's mistake)

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -137,13 +137,15 @@ public Response createCatalog(
...
       throw new IllegalArgumentException("Invalid clientId format");
     }
   }
 
   private void validateClientSecret(String clientSecret) {
-    if (!clientSecret.matches("^[0-9a-f]{32}$")) {
+    if (!clientSecret.matches(
+        realmConfig.getConfig(FeatureConfiguration.CREDENTIAL_RESET_CLIENT_SECRET_PATTERN))) {
```
dimas-b: If validating client ID/secret format is becoming a nuisance to users, from my POV this check can just be removed completely. I do not think Polaris code relies on these values following a particular format.

If backward compatibility is a concern, I'd rather add a simple boolean flag to _disable_ these checks (defaulting to enabled). WDYT?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
rmannibucau: Hmm, it will be per realm so no sure it can be done at startup properly, do you have a code pointer in mind? would bootstrap realm fulfill your expectation?

default is known valid - same as before, it is hardcoded

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
rmannibucau: I'm fine dropping the feature (the two validateClientX methods) too

adutra: +1 to @dimas-b suggestion.

dimas-b: Let's try removing the checks completely and get some more reviews. If people raise concerns we can add an on/off flag then.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
dimas-b: Cf. https://github.com/apache/polaris/blob/1aa95deb75b9f16c871fb8140576d7026471b2b1/runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java#L142

rmannibucau: hmm, not sure i'm a fan of this one, feature is quite specific and we would make the realm down for it? :thinking: 

let's drop the validation for now then

runtime/service/src/test/java/org/apache/polaris/service/admin/PolarisServiceImplTest.java
```diff
@@ -237,4 +240,27 @@ private void invokeValidateExternalCatalog(PolarisServiceImpl service, Catalog c
       }
     }
   }
+
+  private PolarisServiceImpl noAdminResetCredentialPolarisService() {
```
dimas-b: Is this method still used?

dimas-b (approved):

dimas-b:
@rmannibucau : sorry that this PR got forgotten. I think it still makes sense to merge it. Could you resolve conflicts, please?

rmannibucau:
@dimas-b sure, done

dimas-b (approved):

PR #3281, opened 2025-12-16
Move JDBC Quarkus configuration to the JDBC persistence module

flyrain:
Removes the `polaris-runtime-common` module and updates dependent modules accordingly.

Quarkus/JDBC configuration is moved to and owned by the JDBC persistence module, where it fits more naturally. No behavior changes expected.

## Checklist

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -208,7 +208,8 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
         metaStoreManagerMap.remove(realm);
       } catch (IllegalStateException e) {
         // Realm is not bootstrapped, return a failed result
-        results.put(realm, new BaseResult(BaseResult.ReturnStatus.ENTITY_NOT_FOUND, e.getMessage()));
+        results.put(
+            realm, new BaseResult(BaseResult.ReturnStatus.ENTITY_NOT_FOUND, e.getMessage()));
       }
     }
 


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QuarkusRelationalJdbcConfiguration.java
```diff
@@ -16,10 +16,9 @@
...
  * under the License.
  */
-package org.apache.polaris.quarkus.common.config.jdbc;
+package org.apache.polaris.persistence.relational.jdbc;
 
 import io.smallrye.config.ConfigMapping;
-import org.apache.polaris.persistence.relational.jdbc.RelationalJdbcConfiguration;
 
 @ConfigMapping(prefix = "polaris.persistence.relational.jdbc")
 public interface QuarkusRelationalJdbcConfiguration extends RelationalJdbcConfiguration {}
```
adutra: This subinterface is now useless. I suggest removing it and placing the `@ConfigMapping` annotation directly on `RelationalJdbcConfiguration`.

persistence/relational-jdbc/build.gradle.kts
```diff
@@ -33,8 +33,11 @@ dependencies {
   compileOnly(libs.jakarta.enterprise.cdi.api)
   compileOnly(libs.jakarta.inject.api)
 
+  compileOnly(libs.smallrye.config.core) // @ConfigMapping for Quarkus integration
   implementation(libs.smallrye.common.annotation) // @Identifier
   implementation(libs.postgresql)
+  implementation(platform(libs.quarkus.amazon.services.bom))
```
adutra: This new dependency is intriguing. This module doesn't need the RDS extension for compilation, not even for tests. If the intent is to create a Polaris server distribution that supports Amazon RDS, I wonder if a better approach wouldn't be to include the RDS Quarkus extension directly in `polaris-server`?

Conversely, if we decide to keep this dependency here: it is bringing `quarkus-core` transitively. This means that this module becomes de facto a Quarkus module. In this case we should declare the dependency `implementation(platform(libs.quarkus.bom))` and add the Quarkus plugin.

adutra: Oh I see now, the new dependency was added in #2650 to runtime-common, and @dimas-b is already taking a stab at moving this dependency to the appropriate place in #3252.

I suggest removing this dependency from here and use #3252 to find a better place for it. Wdyt?

flyrain: It isn't new dependency. As you said, it was there for a while. It looks like #3252 didn't do anything about it. Let me know if I missed something. 
At this moment, I think it'd be nice to keep as is until we find a better place to hold it. I'm a bit concern that if we removed it here, but didn't get the chance to place it anywhere before next release, it will break users. With that, we might just keep this PR as a pure refactor. WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QuarkusRelationalJdbcConfiguration.java
flyrain: Great idea! Changed per your suggestion.

singhpk234 (approved):
LGTM, thanks @flyrain

persistence/relational-jdbc/build.gradle.kts
adutra: I think it was introduced recently by #2650 – at least I don't see any traces of RDS in Polaris before that.

That said, yes, I agree to move the dependency here and keep thinking about a better home for it.

@dimas-b would you be OK with this?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcConfiguration.java
```diff
@@ -18,8 +18,10 @@
  */
 package org.apache.polaris.persistence.relational.jdbc;
 
+import io.smallrye.config.ConfigMapping;
 import java.util.Optional;
 
+@ConfigMapping(prefix = "polaris.persistence.relational.jdbc")
```
dimas-b, line 24: I do not oppose this change, but I'd like to highlight that this will impose the config naming convention defined in this annotation on downstream projects. It may be possible to provide a different config object for JDBC downstream or it may not be... I cannot say for sure.

Old code delegated runtime configuration to the server runtime, so this may be a behaviour change in downstream projects.

Given that NoSQL code follows a similar approach (not splitting interfaces and `@ConfigMapping`) this change looks reasonable to me.

persistence/relational-jdbc/build.gradle.kts
dimas-b: I do believe that runtime library dependencies should be concentrated in the `runtime/server` module.

So, I do _not_ think bringing RDS as an `implementation` dependency into JDBC persistence is justified. JDBC persistence code should not depend on specific drivers, only on JDBC interfaces.

dimas-b: I believe moving RDS deps to `runtime/server` is actually a small change :) hopefully not too much extra effort on this PR.

flyrain: Thanks for the feedback, @dimas-b! I can definitely do that. One question / concern though:

> I do believe that runtime library dependencies should be concentrated in the runtime/server module.

Can you share a bit more of the rationale behind that? My worry is that if we always push impl-specific deps into runtime/server, the server distribution may end up pulling in everything (JDBC, Spanner, MongoDB, etc.) even when users only need one backend.

Is that the intended direction (a “batteries-included” server), or should we aim for something more modular, e.g., keeping the server lean and enabling backends via optional/default modules/profiles so users can trim dependencies based on their needs?

dimas-b: JDBC Persistence and other similar modules compile against their respective interfaces.

The `runtime/server` module includes drivers (PostgreSQL, RDS, etc.) according to the Polaris project decisions - whatever is supposed to be included in the ASF releases. This does not have to be an all-encompassing list.

In this regard, the `runtime/server` module is essentially the "default" server "package". 

Downstream projects are not supposed to reuse `runtime/server` directly.

Downstream projects reuse JDBC Persistence, `runtime/service`, etc... as needed and build a custom `server`. Downstream projects also include libraries / drivers as needed according to the needs of the downstream project.

For example, PostgreSQL or RDS code may or may not be included - the decision should rest with the downstream project IMHO. A downstream project may even choose to use a completely different JDBC driver.

Conversely, with the current state of this PR, RDS is an impl. dependency of JDBC Persistence and NOT including it in downstream builds requires extra work at explicitly preventing RDS jars from leaking into custom builds... which is very inconvenient.

flyrain: Thanks for the explanation. 
The default dependencies, esp. drivers are not only used by `runtime/server`, but also the `admin` module. Instead of adding them to both, how about putting them in the `runtime/default` module, and let both `admin` and `server` depend on `runtime/default`? I think the downstream projects are also not supposed to directly reuse the `runtime/default` module. WDYT?

dimas-b: Good point about adding common server/admin dependencies to `runtime/default`... I kind of forgot about it :sweat_smile: but indeed I added it a long time ago precisely for this purpose :+1:

dimas-b: With that, I'm probably not going to need #3252 at all :) ... but I'll revisit after this PR is merged ;)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -196,15 +196,20 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
...
+
+        PolarisCallContext callContext = new PolarisCallContext(realmContext, session);
+        BaseResult result = metaStoreManager.purge(callContext);
+        results.put(realm, result);
+
+        sessionSupplierMap.remove(realm);
+        metaStoreManagerMap.remove(realm);
+      } catch (IllegalStateException e) {
+        // Realm is not bootstrapped, return a failed result
+        results.put(realm, new BaseResult(BaseResult.ReturnStatus.ENTITY_NOT_FOUND, e.getMessage()));
```
flyrain: To fix
```
RelationalJdbcPurgeCommandTest > testPurgeFailure(LaunchResult) FAILED
    org.assertj.core.error.AssertJMultipleFailuresError at PurgeCommandTestBase.java:48
```

This is needed as `runtime-defaults` has configuration file `application.properties` with Quarkus configuration like logging, console, or output-related settings, which interferes with how Quarkus Main tests capture stdout. When the exception is thrown, the output isn't captured in `LaunchResult.getOutput()`. It only goes to logs. This causes the test assertion to fail because `result.getOutput()` is empty. 

The fix ensures that when a realm isn't bootstrapped, the `PurgeCommand` receives a failed BaseResult and can properly write the error message to stdout, https://github.com/apache/polaris/blob/dc4e3439042be2d305b4ba945309653512f0dc3a/runtime/admin/src/main/java/org/apache/polaris/admintool/PurgeCommand.java#L53, instead of catching an exception and only writing to stderr (line 62).

persistence/relational-jdbc/build.gradle.kts
flyrain: Fixed it in the new commits. I think we should start to put the driver in the default module moving forward. We could move this to the runtime/default as well in a followup.
```
```

> With that, I'm probably not going to need https://github.com/apache/polaris/pull/3252 at all :)

You can buy me a drink :-)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
dimas-b: Hi @flyrain , the logic in your _message_ LGTM, but the logic in the _code_ is not specific enough, I'm afraid. 

Catching `IllegalStateException` does not necessarily mean that the failure is due to realm not having been bootstrapped.

Is it possible to make this change more specific to the bootstrap use case?

dimas-b: I'm fine with catching `IllegalStateException` in general, but why should the status be `ENTITY_NOT_FOUND` when we do not really know the root reason :thinking: ?

dimas-b: Actually, the matter might be related to mixing `runtime/admin/src/main/resources/application.properties` and `runtime/defaults/src/main/resources/application.properties` (or perhaps `application-test.properties`).

In my experience having more than one `application.properties` file in a Quarkus build leads to very obscure behaviours. So far I thought it was only happening when external jars (with `application.properties` were on the class path), but apparently the oddities happen in a multi-module build too :thinking:

dimas-b: I'm afraid the `runtime/defaults` module will have to be limited to holding only the `application.properties` for servers (shared with test modules).

We can probably keep the Quarkus config refactoring in this PR, but the common module may have to stay to hold dependencies shared between the server and admin tools. WDYT?

flyrain: > In my experience having more than one application.properties file in a Quarkus build leads to very obscure behaviours. 

Thanks for the explanation. This seems to be happening in the server module as well today. If this is a real concern, we should be consistent and remove the dependency from the server too. One concrete option would be to remove `application.properties` from the `defaults` module entirely, and keep it only in the `admin` and `server` modules. Also I don't think `application-it.properties` and `application-test.properties` will mess up with any other `application.properties`, the `server` module has been depended on it for a while. 

More broadly, the separate `admin` module has been a recurring source of friction. It effectively doubles license checks, Docker image publishing, and binary publishing, and it also forces additional shared modules like `common` and `test-common`. At this point, it may be worth stepping back and seriously considering whether Polaris would be simpler and more maintainable as a single module rather than continuing to split to two.

dimas-b: AFAIK, the use of `application.properties` on `main` is delineated well among gradle modules.

`runtime/server` imports `application.properties` from `runtime/defaults`. It's the only properties file the server has, AFAIK.

`runtime/admin` does _not_ import  `runtime/defaults`.

dimas-b: Re: eliminating the `admin` module, I personally do not think it's going to be net beneficial. The main purpose of that module is to hold the admin CLI, which is a different application type from `runtime/server`, so it cannot be produced by the Quarkus build action.

That said, I'm open to discussing ways to simplify the project layout and build... but that conversation is probably way beyond the scope of this PR :)

runtime/admin/build.gradle.kts
```diff
@@ -29,6 +29,7 @@ dependencies {
   implementation(project(":polaris-version"))
   implementation(project(":polaris-api-management-service"))
   implementation(project(":polaris-api-iceberg-service"))
+  implementation(project(":polaris-runtime-defaults"))
```
dimas-b, line 32: Let's not import `runtime/defaults` into the `admin` tool as it creates confusion about which `application.properties` file is effective.

dimas-b (commented):
Posting a fresh review with specific suggestions to unblock this PR. 

I'm open to further conversations about build refactoring after this PR is merged.

runtime/defaults/build.gradle.kts
```diff
@@ -37,4 +37,7 @@ dependencies {
   compileOnly("io.quarkus:quarkus-oidc")
   compileOnly("io.quarkus:quarkus-opentelemetry")
   compileOnly("io.quarkus:quarkus-smallrye-context-propagation")
+
+  implementation(platform(libs.quarkus.amazon.services.bom))
+  implementation("io.quarkiverse.amazonservices:quarkus-amazon-rds")
```
dimas-b, line 42: I'd like to propose moving these deps to `runtime/common` instead and keeping `runtime/defaults` unchanged for the purpose of this PR.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
flyrain: > runtime/server imports application.properties from runtime/defaults. It's the only properties file the server has, AFAIK.

The Sever has its own application.properties file, https://github.com/apache/polaris/blob/11a71cb3172364a40e8232d3c4ba9fe9bacb2967/runtime/server/src/main/resources/application.properties#L25. It will cause the same problem as you mentioned above, right?

 > That said, I'm open to discussing ways to simplify the project layout and build... but that conversation is probably way beyond the scope of this PR :)

Agreed, let me share the idea a bit later to dev ML.

dimas-b: Good find about `runtime/server/src/main/resources/application.properties`. This file is mostly empty, and probably does not affect anything... which is why I forgot about it :sweat_smile: 

In general, it can cause the same issue, indeed. I'll open a PR to remove it.

Quarkus will merge all `application.properties` files it finds. The exact effect of that merge can be hard to predict in my experience. In case of the server, the smaller file contains only built-time properties, which is probably it did not cause any weird issues.

dimas-b: #3341

PR #3293, opened 2025-12-18
Flatten events hierarchy

olsoloviov:
Flatten Polaris events structure to have a single PolarisEvent implementation with enum type and custom attributes.
This is an implementation of @adutra proposal: https://lists.apache.org/thread/xonxwf9b38t9cxo841r0hn1b34plf7og

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/RateLimiterFilter.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/RateLimiterFilter.java
@@ -37,6 +37,7 @@ import org.apache.polaris.service.events.listeners.PolarisEventListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/** Request filter that returns a 429 Too Many Requests if the rate limiter says so */
 @Provider
 @PreMatching
 @Priority(FilterPriorities.RATE_LIMITER_FILTER)
@@ -58,6 +59,7 @@ public class RateLimiterFilter implements ContainerRequestFilter {
     this.eventMetadataFactory = eventMetadataFactory;
   }
 
+  /** Returns a 429 if the rate limiter says so. Otherwise, forwards the request along. */
   @Override
   public void filter(ContainerRequestContext ctx) throws IOException {
     if (!rateLimiter.canProceed()) {
--- a/runtime/service/src/test/java/org/apache/polaris/service/events/AttributeKeyTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/events/AttributeKeyTest.java
@@ -29,7 +29,6 @@ import org.junit.jupiter.api.Test;
 class AttributeKeyTest {
 
   private static final String TEST_KEY = "test_key";
-  private static final String OTHER_KEY = "other_key";
 
   @Test
   void testCreateAndAccessors() {
@@ -37,19 +36,6 @@ class AttributeKeyTest {
 
     assertThat(key.name()).isEqualTo(TEST_KEY);
     assertThat(key.type()).isEqualTo(TypeToken.of(String.class));
-  }
-
-  @Test
-  void testEqualsAndHashCode() {
-    AttributeKey<String> key1 = new AttributeKey<>(TEST_KEY, String.class);
-    AttributeKey<String> key2 = new AttributeKey<>(TEST_KEY, String.class);
-    AttributeKey<String> key3 = new AttributeKey<>(OTHER_KEY, String.class);
-    AttributeKey<Integer> key4 = new AttributeKey<>(TEST_KEY, Integer.class);
-
-    assertThat(key1).isEqualTo(key2);
-    assertThat(key1.hashCode()).isEqualTo(key2.hashCode());
-    assertThat(key1).isNotEqualTo(key3);
-    assertThat(key1).isNotEqualTo(key4);
   }
 
   @Test
--- a/runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTest.java
@@ -19,7 +19,6 @@
 package org.apache.polaris.service.events;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import org.junit.jupiter.api.Test;
 
@@ -33,7 +32,7 @@ class PolarisEventTest {
       ImmutablePolarisEventMetadata.builder().realmId(TEST_REALM).build();
 
   @Test
-  void testBuilderCreatesEvent() {
+  void testConstructorWithAttributes() {
     PolarisEvent event =
         new PolarisEvent(
             PolarisEventType.BEFORE_CREATE_TABLE,
@@ -50,67 +49,11 @@ class PolarisEventTest {
   }
 
   @Test
-  void testAttributeReturnsEmptyForMissingKey() {
+  void testConstructorWithoutAttributes() {
     PolarisEvent event = new PolarisEvent(PolarisEventType.BEFORE_CREATE_TABLE, TEST_METADATA);
 
-    assertThat(event.attributes().get(EventAttributes.CATALOG_NAME)).isEmpty();
-  }
-
-  @Test
-  void testHasAttribute() {
-    PolarisEvent event =
-        new PolarisEvent(
-            PolarisEventType.BEFORE_CREATE_TABLE,
-            TEST_METADATA,
-            new AttributeMap().put(EventAttributes.CATALOG_NAME, TEST_CATALOG));
-
-    assertThat(event.attributes().contains(EventAttributes.CATALOG_NAME)).isTrue();
-    assertThat(event.attributes().contains(EventAttributes.TABLE_NAME)).isFalse();
-  }
-
-  @Test
-  void testAttributeMapSize() {
-    PolarisEvent event =
-        new PolarisEvent(
-            PolarisEventType.BEFORE_CREATE_TABLE,
-            TEST_METADATA,
-            new AttributeMap().put(EventAttributes.CATALOG_NAME, TEST_CATALOG));
-
-    assertThat(event.attributes().size()).isEqualTo(1);
-    assertThat(event.attributes().getRequired(EventAttributes.CATALOG_NAME))
-        .isEqualTo(TEST_CATALOG);
-  }
-
-  @Test
-  void testNullAttributeValueIsIgnored() {
-    PolarisEvent event =
-        new PolarisEvent(
-            PolarisEventType.BEFORE_CREATE_TABLE,
-            TEST_METADATA,
-            new AttributeMap().put(EventAttributes.CATALOG_NAME, null));
-
-    assertThat(event.attributes().contains(EventAttributes.CATALOG_NAME)).isFalse();
+    assertThat(event.type()).isEqualTo(PolarisEventType.BEFORE_CREATE_TABLE);
+    assertThat(event.metadata()).isEqualTo(TEST_METADATA);
     assertThat(event.attributes().isEmpty()).isTrue();
-  }
-
-  @Test
-  void testRequiredAttributeReturnsValue() {
-    PolarisEvent event =
-        new PolarisEvent(
-            PolarisEventType.BEFORE_CREATE_TABLE,
-            TEST_METADATA,
-            new AttributeMap().put(EventAttributes.CATALOG_NAME, TEST_CATALOG));
-
-    assertThat(event.attributes().getRequired(EventAttributes.CATALOG_NAME))
-        .isEqualTo(TEST_CATALOG);
-  }
-
-  @Test
-  void testRequiredAttributeThrowsForMissingKey() {
-    PolarisEvent event = new PolarisEvent(PolarisEventType.BEFORE_CREATE_TABLE, TEST_METADATA);
-
-    assertThatThrownBy(() -> event.attributes().getRequired(EventAttributes.CATALOG_NAME))
-        .isInstanceOf(IllegalStateException.class)
-        .hasMessage("Required attribute catalog_name not found");
   }
 }
--- a/runtime/service/src/test/java/org/apache/polaris/service/events/AttributeMapTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/events/AttributeMapTest.java
@@ -1,0 +1,125 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.events;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.jupiter.api.Test;
+
+class AttributeMapTest {
+
+  private static final AttributeKey<String> STRING_KEY =
+      new AttributeKey<>("string_key", String.class);
+  private static final AttributeKey<Integer> INT_KEY = new AttributeKey<>("int_key", Integer.class);
+
+  @Test
+  void testPutAndGet() {
+    AttributeMap map = new AttributeMap();
+    map.put(STRING_KEY, "value");
+
+    assertThat(map.get(STRING_KEY)).hasValue("value");
+  }
+
+  @Test
+  void testGetReturnsEmptyForMissingKey() {
+    AttributeMap map = new AttributeMap();
+
+    assertThat(map.get(STRING_KEY)).isEmpty();
+  }
+
+  @Test
+  void testGetRequired() {
+    AttributeMap map = new AttributeMap();
+    map.put(STRING_KEY, "value");
+
+    assertThat(map.getRequired(STRING_KEY)).isEqualTo("value");
+  }
+
+  @Test
+  void testGetRequiredThrowsForMissingKey() {
+    AttributeMap map = new AttributeMap();
+
+    assertThatThrownBy(() -> map.getRequired(STRING_KEY))
+        .isInstanceOf(IllegalStateException.class)
+        .hasMessage("Required attribute string_key not found");
+  }
+
+  @Test
+  void testContains() {
+    AttributeMap map = new AttributeMap();
+    map.put(STRING_KEY, "value");
+
+    assertThat(map.contains(STRING_KEY)).isTrue();
+    assertThat(map.contains(INT_KEY)).isFalse();
+  }
+
+  @Test
+  void testNullValueIsIgnored() {
+    AttributeMap map = new AttributeMap();
+    map.put(STRING_KEY, null);
+
+    assertThat(map.contains(STRING_KEY)).isFalse();
+    assertThat(map.isEmpty()).isTrue();
+  }
+
+  @Test
+  void testSize() {
+    AttributeMap map = new AttributeMap();
+    assertThat(map.size()).isEqualTo(0);
+
+    map.put(STRING_KEY, "value");
+    assertThat(map.size()).isEqualTo(1);
+
+    map.put(INT_KEY, 42);
+    assertThat(map.size()).isEqualTo(2);
+  }
+
+  @Test
+  void testIsEmpty() {
+    AttributeMap map = new AttributeMap();
+    assertThat(map.isEmpty()).isTrue();
+
+    map.put(STRING_KEY, "value");
+    assertThat(map.isEmpty()).isFalse();
+  }
+
+  @Test
+  void testCopyConstructor() {
+    AttributeMap original = new AttributeMap();
+    original.put(STRING_KEY, "value");
+
+    AttributeMap copy = new AttributeMap(original);
+
+    assertThat(copy.getRequired(STRING_KEY)).isEqualTo("value");
+
+    // Verify independence
+    original.put(INT_KEY, 42);
+    assertThat(copy.contains(INT_KEY)).isFalse();
+  }
+
+  @Test
+  void testPutReturnsThis() {
+    AttributeMap map = new AttributeMap();
+
+    AttributeMap result = map.put(STRING_KEY, "value");
+
+    assertThat(result).isSameAs(map);
+  }
+}


runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
```diff
@@ -0,0 +1,248 @@
...
+  public static final AttributeKey<String> CATALOG_NAME =
+      AttributeKey.of("catalog_name", String.class);
+  public static final AttributeKey<Catalog> CATALOG = AttributeKey.of("catalog", Catalog.class);
+  public static final AttributeKey<UpdateCatalogRequest> UPDATE_CATALOG_REQUEST =
+      AttributeKey.of("update_catalog_request", UpdateCatalogRequest.class);
+
+  // Namespace attributes
+  public static final AttributeKey<Namespace> NAMESPACE =
+      AttributeKey.of("namespace", Namespace.class);
+  public static final AttributeKey<String> NAMESPACE_STRING =
```
adutra: Isn't this the same as `NAMESPACE_NAME` below? Or is this rather the "fully-qualified" namespace name while the other is the simple name?

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
```diff
@@ -0,0 +1,248 @@
...
+              "update_namespace_properties_request", UpdateNamespacePropertiesRequest.class);
+  public static final AttributeKey<UpdateNamespacePropertiesResponse>
+      UPDATE_NAMESPACE_PROPERTIES_RESPONSE =
+          AttributeKey.of(
+              "update_namespace_properties_response", UpdateNamespacePropertiesResponse.class);
+
+  @SuppressWarnings("unchecked")
+  public static final AttributeKey<Map<String, String>> NAMESPACE_PROPERTIES =
+      (AttributeKey<Map<String, String>>)
+          (AttributeKey<?>) AttributeKey.of("namespace_properties", Map.class);
```
adutra: I'm wondering: in order to increase portability, shouldn't we restrict the types of attributes that we can put in this map? An attribute of type `Map<String, String>` is still OK, I guess, but what if the attribute doesn't have a clear serialized format, e.g. `Optional` or `Function`? Restricting to types that can be safely serialized to Json (and maybe gRPC) would imho make things easier for listener implementors.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/PropertyMapEventListener.java
```diff
@@ -37,10 +40,15 @@ protected abstract void transformAndSendEvent(
       HashMap<String, Object> properties, PolarisEventMetadata metadata);
 
   @Override
-  public void onAfterRefreshTable(IcebergRestCatalogEvents.AfterRefreshTableEvent event) {
-    HashMap<String, Object> properties = new HashMap<>();
-    properties.put("event_type", event.getClass().getSimpleName());
-    properties.put("table_identifier", event.tableIdentifier().toString());
-    transformAndSendEvent(properties, event.metadata());
+  public void onEvent(PolarisEvent event) {
```
adutra: With the flattened hierarchy, we need to ask ourselves if this class is still useful, as opposed to just passing the `PolarisEvent` as is to the CloudWatch sink and let it serialize it without the need for an intermediary representation in the form of a Map.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
```diff
@@ -55,24 +82,38 @@ public void onAfterCreateTable(IcebergRestCatalogEvents.AfterCreateTableEvent ev
...
+            catalog.getName());
     Map<String, String> openTelemetryContext = event.metadata().openTelemetryContext();
     if (!openTelemetryContext.isEmpty()) {
       polarisEvent.setAdditionalProperties(openTelemetryContext);
     }
     processEvent(event.metadata().realmId(), polarisEvent);
   }
 
-  protected abstract void processEvent(String realmId, PolarisEvent event);
+  private <T> T getRequiredAttribute(PolarisEvent event, AttributeKey<T> key) {
```
adutra: This could be declared in `PolarisEvent`, I think it's going to be useful for many listeners, not just this one.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
```diff
@@ -23,29 +23,56 @@
...
+    switch (event.type()) {
+      case AFTER_CREATE_TABLE -> handleAfterCreateTable(event);
+      case AFTER_CREATE_CATALOG -> handleAfterCreateCatalog(event);
+      default -> {
+        // Other events not handled by this listener
+      }
+    }
+  }
+
+  private void handleAfterCreateTable(PolarisEvent event) {
```
adutra, line 47: Again, with flattened events hierarchy there is not a huge difference between these two `handleXYZ` methods. It would certainly be possible to create a unified `handleEvent` method that is valid for all events, thus saving us the hassle of writing 150+ methods (just a thought for later, not for this PR though).

adutra (commented):
Very nice work @olsoloviov, that is very much what I had in mind. Left a few minor comments Thanks!

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
olsoloviov: Exactly, it a "namespace levels" string with full path. I guess the naming is bad here, I will rename to smth like namespace_id or namespace_levels

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/PropertyMapEventListener.java
olsoloviov: I agree. I thought it will be handled in the revived #2962, but the class does not make any sense now so no need to keep it.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adutra: Or `namespace_fqn`?

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
olsoloviov: Shouldn't it be handled during serialization phase? We will have pruning anyway. Also, not sure if it is a desirable scenario, but some non-serializable helper context could be passed with an event.
If you strongly prefer to restrict the types, we could whitelist the types during attributes creation, wdyt @adutra?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
olsoloviov: Moved to PolarisEvent, also switched to throw instead of logging

runtime/service/src/main/java/org/apache/polaris/service/events/AttributeKey.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.Objects;
+
+/**
+ * A type-safe key for event attributes. This allows for strongly-typed attribute access while
+ * maintaining flexibility for custom attributes.
+ *
+ * @param <T> the type of the attribute value
+ */
+public final class AttributeKey<T> {
```
adutra: Should we add some Jackson annotations here to customize Json serialization? I think this class could be serialized as a string scalar containing just the attribute name, wdyt?

runtime/service/src/main/java/org/apache/polaris/service/events/AttributeKey.java
```diff
@@ -0,0 +1,82 @@
...
+    return type;
+  }
+
+  @SuppressWarnings("unchecked")
+  public T cast(Object value) {
+    if (value == null) {
+      return null;
+    }
+    if (!type.isInstance(value)) {
+      throw new ClassCastException(
```
dimas-b: pls consider `type.cast(value)`

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
olsoloviov: Changed to namespace_fqn

runtime/service/src/main/java/org/apache/polaris/service/events/AttributeKey.java
olsoloviov: Done

runtime/service/src/main/java/org/apache/polaris/service/events/AttributeKey.java
olsoloviov: Makes sense, added

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adnanhemani: Admittedly I haven't thought through all the ramifications, but initially I agree with @adutra on this one. Maybe we just keep the bar that whatever the object is, it must be implementing `Serializable`?

Personally, I'd rather have this enforced at the time of the event generation than having to deal with it later in the Event lifecycle - but not a strong opinion.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -2360,13 +2362,15 @@ public void testEventsAreEmitted() {
     table.updateProperties().set(key, valOld).commit();
     table.updateProperties().set(key, valNew).commit();
 
-    var beforeRefreshEvent =
-        testPolarisEventListener.getLatest(IcebergRestCatalogEvents.BeforeRefreshTableEvent.class);
-    Assertions.assertThat(beforeRefreshEvent.tableIdentifier()).isEqualTo(TestData.TABLE);
+    PolarisEvent beforeRefreshEvent =
+        testPolarisEventListener.getLatest(PolarisEventType.BEFORE_REFRESH_TABLE);
+    Assertions.assertThat(beforeRefreshEvent.attribute(EventAttributes.TABLE_IDENTIFIER))
```
adnanhemani: nit: Let's test these using `requiredAttribute` rather than `attribute`?

runtime/service/src/test/java/org/apache/polaris/service/events/AttributeKeyTest.java
```diff
@@ -0,0 +1,107 @@
...
+  @Test
+  void testCreateAndAccessors() {
+    AttributeKey<String> key = AttributeKey.of("test_key", String.class);
+
+    assertThat(key.name()).isEqualTo("test_key");
+    assertThat(key.type()).isEqualTo(String.class);
+  }
+
+  @Test
+  void testTypeCastValidValue() {
```
adnanhemani: Not sure I see value in this test case. Maybe do something like making an AttributeKey of a Long value, then give it an int and see if it casted that properly?

IOW, we shouldn't really be testing `cast` if we didn't write it. We should be testing the code that we wrote.

runtime/service/src/test/java/org/apache/polaris/service/events/AttributeKeyTest.java
```diff
@@ -0,0 +1,107 @@
...
+  void testTypeCastNullValue() {
+    AttributeKey<String> key = AttributeKey.of("test_key", String.class);
+
+    String result = key.type().cast(null);
+
+    assertThat(result).isNull();
+  }
+
+  @Test
+  void testTypeCastInvalidType() {
```
adnanhemani: Similar to above, I would move this test into the test for `PolarisEvent` and test that sending an int to an AttributeKey<String> would throw when calling `key.attribute(..., 123)`.

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTest.java
```diff
@@ -0,0 +1,109 @@
...
+class PolarisEventTest {
+
+  private static final PolarisEventMetadata TEST_METADATA =
+      ImmutablePolarisEventMetadata.builder().realmId("test-realm").build();
+
+  @Test
+  void testBuilderCreatesEvent() {
+    PolarisEvent event =
+        PolarisEvent.builder(PolarisEventType.BEFORE_CREATE_TABLE, TEST_METADATA)
+            .attribute(EventAttributes.CATALOG_NAME, "my-catalog")
```
adnanhemani: nit: extract the magic variables for maintainability

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTest.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test
+  void testRequiredAttributeThrowsForMissingKey() {
+    PolarisEvent event =
+        PolarisEvent.builder(PolarisEventType.BEFORE_CREATE_TABLE, TEST_METADATA).build();
+
+    assertThatThrownBy(() -> event.requiredAttribute(EventAttributes.CATALOG_NAME))
+        .isInstanceOf(IllegalStateException.class)
+        .hasMessageContaining("catalog_name")
```
adnanhemani: nit: Can we test for a more descriptive message? Not looking for exact string matching but something similar to "Required attribute" and/or "not found in event".

adnanhemani (commented):
A very nice job done on this one, @olsoloviov - thank you so much! I left a handful of nit comments and a few minor ones as well, but I think this is very close to merging!

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
olsoloviov: I have added a white list for allowed attributes types

runtime/service/src/test/java/org/apache/polaris/service/events/AttributeKeyTest.java
olsoloviov: Thanks, @adnanhemani, for your comments, I agree that it does not make any sense to test Class.cast()

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
```diff
@@ -0,0 +1,134 @@
...
+  private AllowedAttributeTypes() {}
+
+  static final Set<Class<?>> ALLOWED_TYPES =
+      Set.of(
+          // Primitives
+          String.class,
+          Boolean.class,
+          Integer.class,
+          Long.class,
+          Double.class,
```
adutra: Missing Float? Also: we could probably authorize any type that extends `Number`, including `BigInteger` and `BigDecimal`.

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
```diff
@@ -0,0 +1,134 @@
...
+          // Collections
+          List.class,
```
adutra: This is probably too loose. We are not checking the collections element types here, so one could include a list or a set of any element type, effectively beating the purpose of this class in the first place. I would suggest to also check the element type and the map's key and value types: these should conform to the list specified here as well.

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
```diff
@@ -0,0 +1,134 @@
...
+          Map.class,
+          Set.class,
+          // Iceberg catalog types
+          Namespace.class,
+          TableIdentifier.class,
+          // Iceberg metadata types
+          TableMetadata.class,
+          ViewMetadata.class,
+          // Iceberg REST request types
+          CreateNamespaceRequest.class,
```
adutra: Maybe you could check if the type implements `RESTMessage`?

runtime/service/src/main/java/org/apache/polaris/service/events/AttributeKey.java
```diff
@@ -0,0 +1,79 @@
...
+ * maintaining flexibility for custom attributes.
+ *
+ * <p>Attribute types are validated at key creation time to ensure they can be serialized. See
+ * {@link AllowedAttributeTypes} for the list of allowed types.
+ *
+ * @param <T> the type of the attribute value
+ */
+public final class AttributeKey<T> {
+  private final String name;
+  private final Class<T> type;
```
adutra: I suggest that we use Guava's `com.google.common.reflect.TypeToken` here, as it handles generic types as well.

runtime/service/src/main/java/org/apache/polaris/service/events/AttributeKey.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * A type-safe key for event attributes. This allows for strongly-typed attribute access while
+ * maintaining flexibility for custom attributes.
+ *
+ * <p>Attribute types are validated at key creation time to ensure they can be serialized. See
+ * {@link AllowedAttributeTypes} for the list of allowed types.
+ *
+ * @param <T> the type of the attribute value
+ */
+public final class AttributeKey<T> {
```
adutra: This class could be a record. Then the constructor could be public and we wouldn't need the static `of` method.

The following should be enough (also introducing support for `TypeToken`):

```java
public record AttributeKey<T>(@JsonValue String name, TypeToken<T> type) {

  public AttributeKey(String name, Class<T> type) {
    this(name, TypeToken.of(type));
  }

  public AttributeKey(String name, TypeToken<T> type) {
    this.name = Objects.requireNonNull(name, "name");
    this.type = Objects.requireNonNull(type, "type");
    if (!AllowedAttributeTypes.isAllowed(type)) {
      throw new IllegalArgumentException("Type " + type + " is not allowed for event attributes");
    }
  }
}
```

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
```diff
@@ -0,0 +1,250 @@
...
+      AttributeKey.of("table_identifier", TableIdentifier.class);
+  public static final AttributeKey<CreateTableRequest> CREATE_TABLE_REQUEST =
+      AttributeKey.of("create_table_request", CreateTableRequest.class);
+  public static final AttributeKey<UpdateTableRequest> UPDATE_TABLE_REQUEST =
+      AttributeKey.of("update_table_request", UpdateTableRequest.class);
+  public static final AttributeKey<RegisterTableRequest> REGISTER_TABLE_REQUEST =
+      AttributeKey.of("register_table_request", RegisterTableRequest.class);
+  public static final AttributeKey<RenameTableRequest> RENAME_TABLE_REQUEST =
+      AttributeKey.of("rename_table_request", RenameTableRequest.class);
+  public static final AttributeKey<TableMetadata> TABLE_METADATA_BEFORE =
```
adutra: My IDE is flagging some attributes as unused, which is intriguing. This is one of them.

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
adutra: Here is a suggestion that leverages `TypeToken` in order to check element types of List, Set and Map:

```java
  static final Set<Class<?>> ALLOWED_TYPES =
      Set.of(
          String.class,
          Boolean.class,
          Number.class,
          RESTMessage.class,
          Namespace.class,
          TableIdentifier.class,
          TableMetadata.class,
          ViewMetadata.class,
          Catalog.class,
          Principal.class,
          PrincipalRole.class,
          PrincipalWithCredentials.class,
          CatalogRole.class,
          GrantResource.class,
          PolarisPrivilege.class,
          GenericTable.class);

  static boolean isAllowed(TypeToken<?> type) {
    Class<?> rawType = type.getRawType();
    if (rawType.equals(List.class)) {
      TypeToken<?> elementType = type.resolveType(List.class.getTypeParameters()[0]);
      return isSubtypeOfAllowedType(elementType.getRawType());
    } else if (rawType.equals(Set.class)) {
      TypeToken<?> elementType = type.resolveType(Set.class.getTypeParameters()[0]);
      return isSubtypeOfAllowedType(elementType.getRawType());
    } else if (rawType.equals(Map.class)) {
      TypeToken<?> keyType = type.resolveType(Map.class.getTypeParameters()[0]);
      TypeToken<?> valueType = type.resolveType(Map.class.getTypeParameters()[1]);
      return isSubtypeOfAllowedType(keyType.getRawType())
          && isSubtypeOfAllowedType(valueType.getRawType());
    } else {
      return isSubtypeOfAllowedType(rawType);
    }
  }

  private static boolean isSubtypeOfAllowedType(Class<?> rawType) {
    return ALLOWED_TYPES.stream().anyMatch(t -> t.isAssignableFrom(rawType));
  }
```

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
```diff
@@ -18,10 +18,66 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 
-  PolarisEventType type();
+/**
+ * Represents an event emitted by Polaris. Events have a type, metadata, and a map of typed
+ * attributes. Use {@link #builder(PolarisEventType, PolarisEventMetadata)} to create instances.
+ */
+public record PolarisEvent(
```
adutra, line 25: FYI we can leverage the Java Immutables library to create the builder class for us. Here is how to do it:

```java
@Value.Style(depluralize = true)
public record PolarisEvent(
    PolarisEventType type, PolarisEventMetadata metadata, Map<AttributeKey<?>, Object> attributes) {

  @Builder.Constructor
  public PolarisEvent { ...  }

}
```

Then we can create an event like this:

```java
PolarisEvent event =
  new PolarisEventBuilder()
      .type(PolarisEventType.BEFORE_LIMIT_REQUEST_RATE)
      .metadata(eventMetadataFactory.create())
      .putAttribute(EventAttributes.HTTP_METHOD, ctx.getMethod())
      .putAttribute(
          EventAttributes.REQUEST_URI, ctx.getUriInfo().getAbsolutePath().toString())
      .build();
```

Another option is to transform the record into an interface and annotate with `@PolarisImmutable`.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
```diff
@@ -18,10 +18,66 @@
...
+    PolarisEventType type, PolarisEventMetadata metadata, Map<AttributeKey<?>, Object> attributes) {
+
+  public PolarisEvent {
+    attributes = Collections.unmodifiableMap(new HashMap<>(attributes));
```
adutra:
```suggestion
    attributes = Map.copyOf(attributes);
```

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
```diff
@@ -18,10 +18,66 @@
...
+import java.util.Map;
+import java.util.Optional;
 
-  PolarisEventType type();
+/**
+ * Represents an event emitted by Polaris. Events have a type, metadata, and a map of typed
+ * attributes. Use {@link #builder(PolarisEventType, PolarisEventMetadata)} to create instances.
+ */
+public record PolarisEvent(
+    PolarisEventType type, PolarisEventMetadata metadata, Map<AttributeKey<?>, Object> attributes) {
```
adutra: Nit: exposing a `Map<AttributeKey<?>, Object>` parameter is a bit low-level. It would be better to introduce an `AttributeMap` class with type-safe operations for get and put. That's what Netty does, cf. `io.netty.util.AttributeMap`.

Very simple impl suggestion:

```java
public final class AttributeMap {

  private final Map<AttributeKey<?>, Object> attributes = new HashMap<>();

  @SuppressWarnings("unchecked")
  public <T> Optional<T> get(AttributeKey<T> key) {
    return Optional.ofNullable((T) attributes.get(key));
  }

  public <T> T getRequired(AttributeKey<T> key) {
    return get(key)
        .orElseThrow(() -> new IllegalStateException("Required attribute " + key + " not found"));
  }
}
```

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adutra: While widely adopted in the Hadoop ecosystem, `Serializable` is generally viewed as a legacy feature in Java. Leveraging this marker interface here imho wouldn't achieve our core objective, which is to ensure the attributes map can be easily serialized using common wire formats, particularly JSON. For example, many objects that we want to allow aren't `Serializable`, e.g. the request and response types.

All of this to say: if my suggestion proves too difficult to implement, I am fine if we don't enforce anything.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
olsoloviov: Thanks for catching, these are not needed after ...CommitTable events were removed

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
adnanhemani: I'm 50/50 on this being a requirement, but willing to go with @adutra's suggestion here :)

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
olsoloviov: Thanks for the notion. I see only one issue with such approach  - putAttribute() will not be type-safe, e.g.  something like `.putAttributes(EventAttributes.REQUEST_URI, 123)` will compile

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
olsoloviov: Thanks, that makes sense. It is not compatible with using Immutables on PolarisEvent though - putAttribute() will not be generated, so we will have to stick with manual builder.

olsoloviov:
Thanks for your comprehensive comments, @adutra! I have implemented most of your suggestions in the new commit. Immutables-generated builder does not work nice with AttributeMap though, so I stuck with manual builder.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
```diff
@@ -23,29 +23,45 @@
...
+    LoadTableResponse loadTableResponse =
+        event.requiredAttribute(EventAttributes.LOAD_TABLE_RESPONSE);
+    TableMetadata tableMetadata = loadTableResponse.tableMetadata();
+    String catalogName = event.requiredAttribute(EventAttributes.CATALOG_NAME);
+    Namespace namespace = event.requiredAttribute(EventAttributes.NAMESPACE);
+    String tableName = event.requiredAttribute(EventAttributes.TABLE_NAME);
+
+    org.apache.polaris.core.entity.PolarisEvent polarisEvent =
```
adnanhemani, line 55: nit: not sure why we had to write the full class hierarchy here. Was there a change?

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
```diff
@@ -0,0 +1,115 @@
...
+          return false;
+        }
+      }
+      return true;
+    }
+    return isSubtypeOfAllowedType(rawType);
+  }
+
+  private static boolean isSubtypeOfAllowedType(Class<?> rawType) {
+    return ALLOWED_TYPES.stream().anyMatch(t -> t.isAssignableFrom(rawType));
```
adnanhemani: Not sure if this is too nit-picky - but maybe something like this to ensure we don't waste a lot of cycles going through this check for the same handful of classes?

```
private static final ClassValue<Boolean> IS_ALLOWED =
    new ClassValue<>() {
      @Override
      protected Boolean computeValue(Class<?> type) {
        for (Class<?> allowed : ALLOWED_TYPES) {
          if (allowed.isAssignableFrom(type)) {
            return true;
          }
        }
        return false;
      }
    };

private static boolean isSubtypeOfAllowedType(Class<?> rawType) {
  return IS_ALLOWED.get(rawType);
}
```

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
adutra: That's right indeed. But now that we have `AttributeMap`, shouldn't we move those put methods to `AttributeMap`?

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
adutra: I've seen the updated version, that's much better, thanks! But I'm now concerned about the performance aspect. All this computation is done for _every event attribute_. I am worried that this is a bit too much. @adnanhemani and @olsoloviov, wdyt? Am I overthinking this?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
adutra: I implemented this idea, let me know what you all think:

https://github.com/adutra/polaris/commit/2286bfd68

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
adnanhemani: Commenting in case you forgot to make this change :)

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
adnanhemani: I think @adutra's idea is likely better. Let's go with it.

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
olsoloviov: Well, yeah. This is another PolarisEvent from polaris-core. It was imported before,  but now we had to import our org.apache.polaris.service.events.PolarisEvent, so now we have to fully qualify that one.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEvent.java
olsoloviov: Thanks, @adutra, I see what you mean, looks good. I'd like to use your commit if you are ok with it.

adutra: Of course, feel free to copy or cherry-pick.

runtime/service/src/main/java/org/apache/polaris/service/events/AllowedAttributeTypes.java
olsoloviov: Nice idea, thanks!

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/RateLimiterFilter.java
```diff
@@ -28,13 +28,15 @@
...
+import org.apache.polaris.service.events.AttributeMap;
+import org.apache.polaris.service.events.EventAttributes;
+import org.apache.polaris.service.events.PolarisEvent;
 import org.apache.polaris.service.events.PolarisEventMetadataFactory;
+import org.apache.polaris.service.events.PolarisEventType;
 import org.apache.polaris.service.events.listeners.PolarisEventListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** Request filter that returns a 429 Too Many Requests if the rate limiter says so */
```
adutra, line 37: Why remove the javadocs?

runtime/service/src/test/java/org/apache/polaris/service/events/AttributeKeyTest.java
```diff
@@ -0,0 +1,112 @@
...
+  @Test
+  void testCreateAndAccessors() {
+    AttributeKey<String> key = new AttributeKey<>(TEST_KEY, String.class);
+
+    assertThat(key.name()).isEqualTo(TEST_KEY);
+    assertThat(key.type()).isEqualTo(TypeToken.of(String.class));
+  }
+
+  @Test
+  void testEqualsAndHashCode() {
```
adutra: Some of the tests here are not useful anymore since the class is now a record.

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final String TEST_REALM = "test-realm";
+  private static final String TEST_CATALOG = "my-catalog";
+  private static final String TEST_TABLE = "my-table";
+
+  private static final PolarisEventMetadata TEST_METADATA =
+      ImmutablePolarisEventMetadata.builder().realmId(TEST_REALM).build();
+
+  @Test
+  void testBuilderCreatesEvent() {
```
adutra: Same here, this test is not very useful anymore since the class is a record.

Others below aren't useful either because they are testing `AttributeMap`: they should be moved to `AttributeMapTest`.

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/RateLimiterFilter.java
olsoloviov: Wow, this is weird, thanks for catching!

adnanhemani (approved):
I'm ready to go on these changes. Fantastic job with this, @olsoloviov! Thank you for bearing through this massive PR with us :)

adutra:
Thanks a lot for this awesome work @olsoloviov !

A minor improvement that comes to my mind right now: when there is no listener (e.g. if `polaris.event-listener.type=no-op`), or when the listener is not "interested" in all events (e.g. the CloudWatch listener), instantiating a `PolarisEvent` + `PolarisEventMetadata` + `AttributeMap` is useless. We should come up with a way to conditionally create events only when necessary. E.g.:

```java
  if (polarisEventListener.isEventEnabled(PolarisEventType.BEFORE_CREATE_CATALOG)) {
    polarisEventListener.onEvent(
        new PolarisEvent(
            PolarisEventType.BEFORE_CREATE_CATALOG,
            eventMetadataFactory.create(),
            new AttributeMap()
                .put(EventAttributes.CATALOG_NAME, request.getCatalog().getName())));
  }
  Response resp = delegate.createCatalog(request, realmContext, securityContext);
  if (polarisEventListener.isEventEnabled(PolarisEventType.AFTER_CREATE_CATALOG)) {
    polarisEventListener.onEvent(
        new PolarisEvent(
            PolarisEventType.AFTER_CREATE_CATALOG,
            eventMetadataFactory.create(),
            new AttributeMap().put(EventAttributes.CATALOG, (Catalog) resp.getEntity())));
  }
  return resp;
}
```


Wdyt?

olsoloviov:
> A minor improvement that comes to my mind right now: when there is no listener (e.g. if `polaris.event-listener.type=no-op`), or when the listener is not "interested" in all events (e.g. the CloudWatch listener), instantiating a `PolarisEvent` + `PolarisEventMetadata` + `AttributeMap` is useless. We should come up with a way to conditionally create events only when necessary. 

Thanks for the suggestion, @adutra, I think it's quite reasonable. Unless you prefer it to be fixed in this PR, I think I can do it in a follow up PR, since it is somewhat orthogonal to the current improvement.

adutra:
> Thanks for the suggestion, @adutra, I think it's quite reasonable. Unless you prefer it to be fixed in this PR, I think I can do it in a follow up PR, since it is somewhat orthogonal to the current improvement.

I meant this for a follow-up PR 👍

adutra:
I'm leaving a few more hours for review. If no objections come up, I'll merge tomorrow morning my time (CET).

PR #3295, opened 2025-12-18
Ensure release can only run from specific SHA

### LINKED ISSUE #3290: Release workflows - The 4th workflow does not verify it runs against the latest RC
### Describe the bug

The 4th release workflow, which publishes release artifacts, is run against a branch, as per Github Action UI.  It contains a check that verifies it is running against a `release/[major].[minor].x` branch, but it does not contain a check that verifies it is running against the latest RC tag of that version.

So it is technically possible to add commits to the release branch after the binaries have been packaged and the vote thread has started, and misuse the publication workflow.

The consequences would be that the binaries published to Nexus and Apache dist would be those of the tag, But the Docker image would be that of the branch HEAD.

The workflow should contain a check that verifies it is running against the tag corresponding to the last RC, to prevent this from happening.

### To Reproduce

_No response_

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Additional context

_No response_

### System information

_No response_

pingtimeout:
This PR includes multiple improvements to release workflows so that they are more robust and more user-friendly:

First, it contains an update the second release workflow so that it sets the project version to a final version only when RC=0.  This condition also determines it updates the changelog as it is unnecessary to call the gradle task for RC ≥ 1.

Second, it changes the run criteria for the third workflow that builds and stages artifacts.  Now that workflow **MUST** be run from an RC tag.  It cannot be run from a release branch anymore.  This change is required because of an unexpected scenario: when the same SHA is used to cut multiple RC.  Then the order of git tags is unknown and the command that was used to find _the current tag_ is non-deterministic.  With that change, the problem cannot happen anymore as the release manager must specify the exact RC tag that is used.

Finally, the fourth workflow that publishes artifacts after a vote passes contains a new safety check.  When the workflow is run, it verifies that it is running on a `release/[major].[minor].x` branch **and** that the head of that branch corresponds to the last RC tag.  That second condition ensures that no new commit has been added since the vote started for the last RC.  It is necessary because part of publishing involves rebuilding Polaris Docker images, so the source code must be identical to the RC.  This fixes #3290.

The PR includes an update to the release guide to reflect those changes.  It also contains a change just for consistency.  The e-mail templates used to reference the Polaris version as `x.y.z-rci` but the `x` element is also used in the release branch name as a letter (e.g. `release/1.3.x`) which is confusing.  All occurrences of `x.y.z` and `rci` have been replaced by `[major].[minor].[patch]` and `rc[N]`.

Testing has been done with dry-run mode and can be verified with those runs:
* When [the third release workflow is run from a release branch](https://github.com/pingtimeout/polaris/actions/runs/20373122420), it fails with the message `❌ Workflow must be run from a release candidate tag, not a branch.  Current ref: refs/heads/release/99.98.x.  Please select a release candidate tag (e.g., apache-polaris-1.0.0-incubating-rc0) from the 'Use workflow from' dropdown in the GitHub UI.`
* When [the fourth release workflow is run but commits have been pushed to the release branch after the RC was cut](https://github.com/pingtimeout/polaris/actions/runs/20374312784), it fails with the message: `❌ Current HEAD (4a46dbf0f5f37a42ac645344f2115a925d1a8468) does not match RC tag apache-polaris-99.98.0-incubating-rc1 (1ef2eda8720e8957e5df421401fa0db955a03bd8).`
* When [the fourth workflow is run normally](https://github.com/pingtimeout/polaris/actions/runs/20375021761), the check passes with the message `✅ Current HEAD matches RC tag apache-polaris-99.98.0-incubating-rc2`

### DIFF
--- a/site/content/community/release-guides/semi-automated-release-guide.md
+++ b/site/content/community/release-guides/semi-automated-release-guide.md
@@ -44,16 +44,17 @@ Polaris follows a schedule-driven release model. The first thing to do is to sen
 Note that the tentative date is only a suggestion. The actual date of the release will be determined by taking community feedback into account. For instance, if specific pull requests are about to be merged, the release may be delayed by a couple of days to include them.
 
 ```
-[DISCUSS] Apache Polaris x.y.z
+[DISCUSS] Apache Polaris [major].[minor].[patch]
 ```
 
 ```
 Hello everyone,
 
 The purpose of this e-mail is to collect feedback on the upcoming Apache
-Polaris x.y.z release. The tentative release date is YYYY-MM-DD. Please let
-me know if you have any concerns or comments, or if there are some specific
-pull requests that you would like to see included in the release.
+Polaris [major].[minor].[patch] release. The tentative release date is
+YYYY-MM-DD. Please let me know if you have any concerns or comments, or if
+there are some specific pull requests that you would like to see included in
+the release.
 
 Thanks,
 ```
@@ -121,7 +122,7 @@ The last step for a release candidate is to create a VOTE thread on the dev mail
 Recommended title subject:
 
 ```
-[VOTE] Release Apache Polaris x.y.z (rci)
+[VOTE] Release Apache Polaris [major].[minor].[patch] (rc[N])
 ```
 
 Recommended content:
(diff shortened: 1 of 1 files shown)


releasey/libs/_version.sh
```diff
@@ -146,19 +146,23 @@ function find_next_patch_number {
...
+  # that patch number (allowing for additional RCs like rc1, rc2, etc.).
   local major="$1"
   local minor="$2"
 
-  # Get all existing tags for this major.minor version
-  local tag_pattern="apache-polaris-${major}.${minor}.*-incubating-rc*"
-  local existing_tags
-  existing_tags=$(git tag -l "${tag_pattern}" | sort -V)
+  # Get all existing RC tags for this major.minor version
+  local rc_tag_pattern="apache-polaris-${major}.${minor}.*-incubating-rc*"
```
snazy, line 157:
```suggestion
  local rc_tag_pattern="apache-polaris-${major}\.${minor}.[0-9]+-incubating-rc[0-9]+"
```

releasey/libs/_version.sh
snazy, line 128:
```suggestion
  local tag_pattern="apache-polaris-${version_without_rc}-rc[0-9]+"
```

PR #3296, opened 2025-12-18
Site: Fix typos in release guide

snazy:


pingtimeout:
Woops, thanks for catching those, Robert !

PR #3314, opened 2025-12-22
Added HttpRoute and Gateway to Helm Chart

cccs-cat001:
We've been migrating to the HttpRoute/Gateway setup for network routing in our kubernetes clusters and figured this would be a useful addition to the community. 

## Checklist

### DIFF
--- a/helm/polaris/templates/_helpers.tpl
+++ b/helm/polaris/templates/_helpers.tpl
@@ -96,7 +96,7 @@ Validate that only one of ingress or httproute is enabled
 {{- if and .Values.ingress.enabled .Values.httproute.enabled }}
 {{- fail "Cannot enable both ingress and httproute. Please enable only one." }}
 {{- end }}
-{{- if and .Values.ingress.enabled (and (not .Values.httproute.enabled) .Values.gateway.enabled) }}
+{{- if and (not .Values.httproute.enabled) .Values.gateway.enabled }}
 {{- fail "In order to use the gateway please enable the httproute and disable the ingress."}}
 {{- end }}
 {{- end }}


helm/polaris/templates/_helpers.tpl
```diff
@@ -89,6 +89,15 @@ app.kubernetes.io/name: {{ include "polaris.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
+{{/*
+Validate that only one of ingress or httproute is enabled
+*/}}
+{{- define "polaris.validateRouting" -}}
+{{- if and .Values.ingress.enabled .Values.httproute.enabled }}
+{{- fail "Cannot enable both ingress and httproute. Please enable only one." }}
```
adutra, line 97: You could also validate that if `.Values.gateway.enabled` is true, then `.Values.httproute.enabled` must be true as well, wdyt?

helm/polaris/templates/httproute.yaml
```diff
@@ -0,0 +1,52 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "polaris.labels" . | nindent 4 }}
+  {{- with .Values.httproute.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  hostnames: {{ .Values.httproute.hosts }}
+  # This will be different depending on which Gateway implementation is in use and the configuration of the Gateway resource
```
adutra: Does this mean that a user could provide a Gateway configuration that would be incompatible with the HTTPRoute resource being generated? Just trying to understand what this comment implies.

helm/polaris/README.md
```diff
@@ -151,6 +151,17 @@ Below are two sample deployment models for installing the chart: one with a non-
 > **These files are intended for testing purposes primarily, and may not be suitable for production use**.
 > For production deployments, create your own values files based on the provided examples.
 
+### Using the Gateway API (Advanced) 
+
+You may want to use the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) for more advanced network routing. To do so, you may need to run the following commands:
```
adutra: I am not sure these commands make sense here, unless we are talking about testing the Helm chart (there is a constant fluctuation in this file between production vs testing guidelines). 

Assuming this section is about production guidelines:

I think the most important thing to mention is that in most cases, a Gateway will be already present in the cluster (since this resource is generally managed by an admin persona). 

The below commands would basically correspond to what that admin persona would have done. So imho they only make sense for testing purposes, not for deploying in production.

And finally we could also mention that a Gateway can be installed as part of the chart installation. This should be restricted to simpler setups, since it installs a Gateway in the same namespace as Polaris, and with the same privileges.

helm/polaris/README.md
```diff
@@ -151,6 +151,17 @@ Below are two sample deployment models for installing the chart: one with a non-
 > **These files are intended for testing purposes primarily, and may not be suitable for production use**.
 > For production deployments, create your own values files based on the provided examples.
 
+### Using the Gateway API (Advanced) 
+
+You may want to use the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) for more advanced network routing. To do so, you may need to run the following commands: 
+```bash
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/latest/download/standard-install.yaml
+helm install envoy-gateway oci://docker.io/envoyproxy/gateway-helm -n gateway --create-namespace
```
adutra: Maybe mention above or below that this is just an example of Gateway implementation by Envoy.

helm/polaris/README.md
```diff
@@ -151,6 +151,17 @@ Below are two sample deployment models for installing the chart: one with a non-
 > **These files are intended for testing purposes primarily, and may not be suitable for production use**.
 > For production deployments, create your own values files based on the provided examples.
 
+### Using the Gateway API (Advanced)
```
adutra: Please don't make changes directly to this file, but modify `README.md.gotmpl` instead then run `helm-docs --chart-search-root=helm`.

adutra: I think this section is not well placed, because the text right above makes references to the sections right below. Maybe move this down?

helm/polaris/templates/httproute.yaml
cccs-cat001: I don't really know what it means to be honest, I wrote that so long ago 😆  

Seems to me like it says "These are variables" 😆  I think it makes most sense just to remove this comment

helm/polaris/README.md
cccs-cat001: I realized this was dumb anyways, since I added the gateway to this helm chart.

helm/polaris/templates/_helpers.tpl
```diff
@@ -89,6 +89,18 @@ app.kubernetes.io/name: {{ include "polaris.name" . }}
...
+{{- end }}
+{{- if and .Values.ingress.enabled (and (not .Values.httproute.enabled) .Values.gateway.enabled) }}
```
adutra: I think this should just test `and (not .Values.httproute.enabled) .Values.gateway.enabled`, wdyt?

adutra (approved):
LGTM thanks @cccs-cat001 !

PR #3320, opened 2025-12-23
DO NOT MERGE: Polaris Core SPI Idea

adam-christian-software:
Based on discussions in the first community sprint, this is an idea of splitting Polaris into Polaris-Core-SPI & Polaris-Core-Impl.


polaris-core-spi/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+  compileOnly("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml")
+
+  compileOnly(project(":polaris-immutables"))
+  annotationProcessor(project(":polaris-immutables", configuration = "processor"))
+
+  compileOnly(project(":polaris-api-management-model"))
+
+  compileOnly(platform(libs.iceberg.bom))
+  compileOnly("org.apache.iceberg:iceberg-api")
+  compileOnly("org.apache.iceberg:iceberg-core")
```
dimas-b, line 39: Why is `iceberg-core` a dependency of Polaris SPI? 🤔

polaris-core-spi/src/main/java/org/apache/polaris/core/catalog/PolarisCatalogHelpers.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.collect.ImmutableList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.polaris.core.entity.PolarisEntity;
+
+/**
+ * Holds helper methods translating between persistence-layer structs and Iceberg objects shared by
```
dimas-b, line 30: Just from the javadoc it does not look like this class belongs in an SPI 🤔 It is not something to be implemented by plugins 🤔

polaris-core-spi/build.gradle.kts
adam-christian-software: So, there imports of at least:
org.apache.iceberg.rest.auth.OAuth2Properties
org.apache.iceberg.rest.auth.OAuth2Util
org.apache.iceberg.rest.auth.AuthProperties
org.apache.iceberg.CatalogProperties
org.apache.iceberg.TableMetadata
org.apache.iceberg.view.ViewMetadata
org.apache.iceberg.rest.RESTUtil
org.apache.iceberg.rest.RESTSerializers

I agree that it does not seem like this would be what we would want longer-term.

polaris-core-spi/src/main/java/org/apache/polaris/core/catalog/PolarisCatalogHelpers.java
adam-christian-software: It's used by NamespaceEntity.

Issue #3080, opened 2025-11-18
Releasy: Clean up staging repos for "failed RCs"


snazy:
### Is your feature request related to a problem? Please describe.

The release automation should at best automatically delete staged artifacts of a previous RC, including the Nexus repo.

PR #3321, opened 2025-12-23
Add cancel release candidate workflow (#3080)

pingtimeout:
Example run here : https://github.com/pingtimeout/polaris/actions/runs/20467542429

This PR contains (apart from 1 ninja commit) a new workflow that can be used after an RC vote fails.  The workflow deletes the staged artifacts from Apache dist dev SVN repository (if they exist) and drops the Nexus staging repository.  It also generates a template e-mail to be sent to the mailing list.

Marking this PR as draft as I still need to update the release guide accordingly.

### DIFF
--- a/.github/workflows/release-X-cancel-release-candidate.yml
+++ b/.github/workflows/release-X-cancel-release-candidate.yml
@@ -119,12 +119,6 @@ jobs:
           | Staging Repository ID | \`${STAGING_REPOSITORY_ID}\` |
           EOT
 
-      - name: Set up Java
-        uses: actions/setup-java@f2beeb24e141e01a676f977032f5a29d81c9e27e # v5
-        with:
-          distribution: 'temurin'
-          java-version: '21'
-
       - name: Drop Apache Nexus staging repository
         env:
           NEXUS_USERNAME: ${{ secrets.APACHE_USERNAME }}


.github/workflows/release-X-cancel-release-candidate.yml
```diff
@@ -0,0 +1,253 @@
...
+          cat <<EOT >> $GITHUB_STEP_SUMMARY
+          | Parameter | Value |
+          | --- | --- |
+          | Git tag | \`${git_tag}\` |
+          | Version | \`${version_without_rc}\` |
+          | RC number | \`${rc_number}\` |
+          | Staging Repository ID | \`${STAGING_REPOSITORY_ID}\` |
+          EOT
+
+      - name: Set up Java
```
dimas-b: Where is java used in this workflow? 🤔

pingtimeout: My bad, it is a leftover from the first attempt, which was using `./gradlew dropApacheStagingRepository`.  But unfortunately the gradle plugin does not offer the possibility to drop staging repositories so I had to revert to `curl` and forgot to remove the Java installation.  Fixing...

dimas-b (approved):

PR #3323, opened 2025-12-23
Deprecate untyped `RealmConfig.getConfig()`

dimas-b:
* `getConfig(String)` has a generic return type, but the call path that gets the value does not perform any type validation.

* Deprecate this method in favour of well-typed `getConfig(PolarisConfiguration)`

* Migrate the single use case in Polaris code to the well-typed method.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
@@ -31,7 +31,9 @@ public interface RealmConfig {
    * @param <T> the type of the configuration value
    * @param configName the name of the configuration key to check
    * @return the current value set for the configuration key, or null if not set
+   * @deprecated Use typed {@link #getConfig(PolarisConfiguration)} instead.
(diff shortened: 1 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfig.java
```diff
@@ -31,7 +31,9 @@ public interface RealmConfig {
    * @param <T> the type of the configuration value
    * @param configName the name of the configuration key to check
    * @return the current value set for the configuration key, or null if not set
+   * @deprecated Use typed {@link #getConfig(PolarisConfiguration)} instead.
```
singhpk234, line 34: do we have a version in mind, where were wanna remove it, if we do it would be nice to add it here !

dimas-b: I personally do not have a preference for a specific Polaris version where this can be removed, but open to suggestions :)

PR #3324, opened 2025-12-24
Refactor PolarisConfigurationStore to streamline value typing

dimas-b:
* Avoid generic return types when the value type is not certain

* Add new method to be overridden by implementation that returns `Object` to avoid arbitrary casts to `T`

* Type casts are performed only when the expected value type is known.

* This change is backward-compatible with older implementations of `PolarisConfigurationStore`

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
@@ -39,10 +39,11 @@ public interface PolarisConfigurationStore {
   /**
    * Retrieve the current value for a configuration key for a given realm. May be null if not set.
    *
-   * <p>This method is meant to be overridden by concrete configuration store implementations. This
-   * method is not meant to be calls by code that needs access to configuration values.
-   * Configuration consumers should call typed methods that take a {@link PolarisConfiguration}
-   * parameter instead.
+   * <p>This method is meant to be overridden by concrete configuration store implementations.
+   *
+   * <p>This method is <strong>not</strong> meant to be called by code that needs access to
+   * configuration values. Configuration consumers should call typed methods that take a {@link
+   * PolarisConfiguration} parameter instead.
    *
    * @param realmContext the realm context
    * @param configName the name of the configuration key to check


polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
```diff
@@ -39,11 +39,26 @@ public interface PolarisConfigurationStore {
   /**
    * Retrieve the current value for a configuration key for a given realm. May be null if not set.
    *
+   * <p>This method is meant to be overridden by concrete configuration store implementations. This
+   * method is not meant to be calls by code that needs access to configuration values.
+   * Configuration consumers should call typed methods that take a {@link PolarisConfiguration}
+   * parameter instead.
```
snazy:
```suggestion
   * <p>This method is meant to be overridden by concrete configuration store implementations.
   * Prefer the typed functions that take a {@link PolarisConfiguration} parameter instead of a {@code String}.
```

dimas-b: I reformatted the javadoc, hoping to emphasize the need to use typed methods.

TBH, I'm not sure whether the "prefer" suggestion is targeting callers or implementors 🤔 WDYT?

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfigurationStore.java
```diff
@@ -39,11 +39,27 @@ public interface PolarisConfigurationStore {
   /**
    * Retrieve the current value for a configuration key for a given realm. May be null if not set.
    *
+   * <p>This method is meant to be overridden by concrete configuration store implementations.
```
snazy, line 42: Actually, I wonder whether this function and the `getConfiguration` functions should have a default implementation at all. Wouldn't it be cleaner to move that to the actual implementation? Mean, this `interface` isn't really an interface but rather a not-really abstract class.

dimas-b: Good point. I've been thinking about splitting this into an API/SPI pair too, but I thought it might be too intrusive... Let me actually try that refactoring and see what's impacted.

dimas-b: Prerequisite refactoring: #3412

dimas-b:
Closing in favour of #3573

PR #3270, opened 2025-12-14
refactor: remove StorageCredentialsVendor

tokoko:
First of several PRs with the goal to simplify the code flow for fetching subscoped credentials. Removes `StorageCredentialsVendor` class which now sits between `StorageCredentialCache` and `PolarisCredentialVendor` but doesn't add any real value.

PR #3327, opened 2025-12-24
feat: Add AWS STS Session Tags support for credential vending

### LINKED ISSUE #3325: [Feature] Add AWS STS Session Tags to Credential Vending for CloudTrail Correlation

**Description:**

Add support for AWS STS Session Tags when vending S3 credentials via AssumeRole. This enables deterministic correlation between Polaris catalog operations and downstream S3 access events in AWS CloudTrail.

**Motivation:**

Currently, the only correlation mechanism between catalog credential vending and S3 access is the role session name. This provides principal-level attribution but lacks granularity for:

* Fine-grained audit trails (table → S3 reads)
* Cost allocation by catalog/namespace/table
* Security forensics
* Compliance reporting

**Proposed Solution:**

Add session tags (`polaris:catalog`, `polaris:namespace`, `polaris:table`, `polaris:principal`, `polaris:request-id`) to `AssumeRoleRequest`. Controlled by a new feature flag `INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL` (default: `false`).

**Acceptance Criteria:**

*  New feature flag `INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL` added
*  Session tags added to AssumeRoleRequest when feature enabled
* Tags marked as transitive for role chaining scenarios
* Values truncated to AWS limits (256 chars)
* Documentation updated with IAM policy requirements
* Unit and integration tests added

**1. Use sts:SourceIdentity**

Pros: Immutable, appears in CloudTrail
Cons: Single value only, cannot convey structured data

**2. Encode metadata in role session name**
Pros: No IAM policy changes needed
Cons: 64-char limit insufficient; parsing complexity; potential collisions

**3. External correlation via timestamp**
Current approach: Join audit logs and CloudTrail by time window
Cons: Non-deterministic; fails with concurrent requests; complex queries

### Additional context

**Dependencies:**

* Requires IAM role trust policy update to allow sts:TagSession

**Related:**

`INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` feature flag

obelix74:
Fixes #3325

## Summary
Adds support for AWS STS Session Tags when vending S3 credentials, enabling correlation between Polaris catalog operations and S3 access in AWS CloudTrail. This feature is controlled by a new feature flag `INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL` (default: `false`).
   
## Motivation
When Polaris vends temporary AWS credentials via STS `AssumeRole`, there's no way to correlate S3 access events in CloudTrail back to the originating Polaris catalog operation. This makes audit trails incomplete and security investigations difficult.

AWS STS Session Tags solve this by attaching metadata to the assumed role session that appears in CloudTrail events for all subsequent API calls made with those credentials.
   
### New Components
   - **`CredentialVendingContext`** - Immutable value class holding session tag context (catalog name, namespace, table name, request-id)
   - **`INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL`** - Feature flag in `FeatureConfiguration` (default: `false`)

### AWS Integration
   - **`AwsCredentialsStorageIntegration`** - Adds 5 session tags to `AssumeRoleRequest` when feature enabled:
     - `polaris:principal` - The authenticated principal name
     - `polaris:catalog` - The catalog being accessed
     - `polaris:namespace` - The namespace being accessed
     - `polaris:table` - The table being accessed
     - `polaris:request-id` - The originating request ID for correlation
   - Tags are marked as **transitive** for role chaining support
   - Tag values truncated to 256 characters (AWS limit)
   - Missing values use `"unknown"` placeholder for consistent CloudTrail presence

### Cache Key Updates
   - **`StorageCredentialCacheKey`** - Now includes `CredentialVendingContext` when session tags enabled
   - **`StorageCredentialCache`** - Includes principal in cache key when session tags OR principal-name-in-session features are
   enabled (prevents credential cross-contamination)
   - Added `toSanitizedLogString()` for safe debug logging without exposing sensitive context
   
### Interface Updates
   - Updated `PolarisStorageIntegration`, `PolarisCredentialVendor`, `StorageCredentialsVendor` to accept
   `CredentialVendingContext`
   - Updated Azure and GCP integrations for signature compatibility (pass-through only)
   
## Testing
   - **6 new unit tests** in `AwsCredentialsStorageIntegrationTest`:
     - `testSessionTagsIncludedWhenFeatureEnabled`
     - `testSessionTagsNotIncludedWhenFeatureDisabled`
     - `testSessionTagsWithPartialContext`
     - `testSessionTagsWithLongValues`
     - `testSessionTagsWithEmptyContext`
     - `testSessionTagsAccessDeniedGracefulHandling`
   - All existing tests updated for signature compatibility
   - Full test suites pass: `polaris-core:test`, `polaris-runtime-service:test`
   
## Configuration
   To enable session tags, set in your Polaris configuration:
```
   polaris:
     features:
       INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL: "true"
```
   
## IAM Requirements
   When enabled, the IAM role's trust policy must allow `sts:TagSession`:
```
   {
     "Effect": "Allow",
     "Principal": { "AWS": "arn:aws:iam::ACCOUNT:root" },
     "Action": ["sts:AssumeRole", "sts:TagSession"]
   }
```
   
## Breaking Changes
   None. Feature is off by default and all interface changes are backward compatible.

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #3325

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
@@ -32,10 +32,8 @@ import org.apache.polaris.immutables.PolarisImmutable;
  *   <li>{@code catalogName} - The name of the catalog vending credentials
  *   <li>{@code namespace} - The namespace/database being accessed (e.g., "db.schema")
  *   <li>{@code tableName} - The name of the table being accessed
+ *   <li>{@code activatedRoles} - Comma-separated list of activated principal roles
  * </ul>
- *
- * <p>Principal information (name and activated roles) is obtained directly from {@link
- * org.apache.polaris.core.auth.PolarisPrincipal} rather than being duplicated in this context.
  *
  * <p>These values appear in cloud provider audit logs (e.g., AWS CloudTrail), enabling correlation
  * between catalog operations and data access events.
@@ -63,6 +61,13 @@ public interface CredentialVendingContext {
   Optional<String> tableName();
 
   /**
+   * The activated roles for the principal, represented as a comma-separated sorted string. This is
+   * included in the context (rather than extracted from PolarisPrincipal) to ensure it is part of
+   * the cache key when session tags are enabled.
+   */
+  Optional<String> activatedRoles();
+
+  /**
    * Creates a new builder for CredentialVendingContext.
    *
    * @return a new builder instance
@@ -87,6 +92,8 @@ public interface CredentialVendingContext {
     Builder namespace(Optional<String> namespace);
 
     Builder tableName(Optional<String> tableName);
+
+    Builder activatedRoles(Optional<String> activatedRoles);
 
     CredentialVendingContext build();
   }
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
@@ -127,7 +127,8 @@ public class AwsCredentialsStorageIntegration
 
       // Add session tags when the feature is enabled
       if (includeSessionTags) {
-        List<Tag> sessionTags = buildSessionTags(polarisPrincipal, credentialVendingContext);
+        List<Tag> sessionTags =
+            buildSessionTags(polarisPrincipal.getName(), credentialVendingContext);
         if (!sessionTags.isEmpty()) {
           request.tags(sessionTags);
           // Mark all tags as transitive for role chaining support
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
@@ -20,9 +20,6 @@ package org.apache.polaris.core.storage.aws;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-import org.apache.polaris.core.auth.PolarisPrincipal;
 import org.apache.polaris.core.storage.CredentialVendingContext;
 import software.amazon.awssdk.services.sts.model.Tag;
 
@@ -43,24 +40,15 @@ public final class AwsSessionTagsBuilder {
   }
 
   /**
-   * Builds a list of AWS STS session tags from the principal and credential vending context. These
-   * tags will appear in CloudTrail events for correlation purposes.
+   * Builds a list of AWS STS session tags from the principal name and credential vending context.
+   * These tags will appear in CloudTrail events for correlation purposes.
    *
-   * @param principal the principal requesting credentials (provides name and activated roles)
-   * @param context the credential vending context containing catalog, namespace, and table
+   * @param principalName the name of the principal requesting credentials
+   * @param context the credential vending context containing catalog, namespace, table, and roles
    * @return a list of STS Tags to attach to the AssumeRole request
    */
-  public static List<Tag> buildSessionTags(
-      PolarisPrincipal principal, CredentialVendingContext context) {
+  public static List<Tag> buildSessionTags(String principalName, CredentialVendingContext context) {
     List<Tag> tags = new ArrayList<>();
-
-    // Extract principal name and roles
-    String principalName = principal.getName();
-    Set<String> roles = principal.getRoles();
-    String rolesString =
-        (roles != null && !roles.isEmpty())
-            ? roles.stream().sorted().collect(Collectors.joining(","))
-            : TAG_VALUE_UNKNOWN;
 
     // Always include all tags with "unknown" placeholder for missing values
     // This ensures consistent tag presence in CloudTrail for correlation
@@ -72,7 +60,7 @@ public final class AwsSessionTagsBuilder {
     tags.add(
         Tag.builder()
             .key(CredentialVendingContext.TAG_KEY_ROLES)
-            .value(truncateTagValue(rolesString))
+            .value(truncateTagValue(context.activatedRoles().orElse(TAG_VALUE_UNKNOWN)))
             .build());
     tags.add(
         Tag.builder()
--- a/polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
@@ -1014,16 +1014,14 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
         ArgumentCaptor.forClass(AssumeRoleRequest.class);
     Mockito.when(stsClient.assumeRole(requestCaptor.capture())).thenReturn(ASSUME_ROLE_RESPONSE);
 
+    // Roles are included in context (not extracted from principal) to be part of cache key
     CredentialVendingContext context =
         CredentialVendingContext.builder()
             .catalogName(Optional.of("test-catalog"))
             .namespace(Optional.of("db.schema"))
             .tableName(Optional.of("my_table"))
+            .activatedRoles(Optional.of("admin,reader"))
             .build();
-
-    // Create a principal with roles for this test
-    PolarisPrincipal principalWithRoles =
-        PolarisPrincipal.of("test-principal", Map.of(), Set.of("admin", "reader"));
 
     new AwsCredentialsStorageIntegration(
             AwsStorageConfigurationInfo.builder()
@@ -1037,7 +1035,7 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
             true,
             Set.of(s3Path(bucket, warehouseKeyPrefix)),
             Set.of(s3Path(bucket, warehouseKeyPrefix)),
-            principalWithRoles,
+            POLARIS_PRINCIPAL,
             Optional.empty(),
             context);
 
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
@@ -25,6 +25,7 @@ import jakarta.inject.Inject;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.polaris.core.auth.PolarisPrincipal;
@@ -151,12 +152,14 @@ public class StorageAccessConfigProvider {
    * Builds a credential vending context from the table identifier and resolved path. This context
    * is used to populate session tags in cloud provider credentials for audit/correlation purposes.
    *
-   * <p>Note: Principal information (name and activated roles) is obtained directly from {@link
-   * PolarisPrincipal} during session tag generation, not from this context.
+   * <p>The activated roles are included in this context (rather than extracted from
+   * PolarisPrincipal during session tag generation) to ensure they are part of the cache key when
+   * session tags are enabled. This prevents false positive cache hits when a principal's roles
+   * change.
    *
    * @param tableIdentifier the table identifier containing namespace and table name
    * @param resolvedPath the resolved entity path containing the catalog entity
-   * @return a credential vending context with catalog, namespace, and table
+   * @return a credential vending context with catalog, namespace, table, and activated roles
    */
   private CredentialVendingContext buildCredentialVendingContext(
       TableIdentifier tableIdentifier, PolarisResolvedPathWrapper resolvedPath) {
@@ -176,6 +179,13 @@ public class StorageAccessConfigProvider {
 
     // Extract table name from table identifier
     builder.tableName(Optional.of(tableIdentifier.name()));
+
+    // Extract activated roles from principal - included in context to be part of cache key
+    Set<String> roles = polarisPrincipal.getRoles();
+    if (roles != null && !roles.isEmpty()) {
+      String rolesString = roles.stream().sorted().collect(Collectors.joining(","));
+      builder.activatedRoles(Optional.of(rolesString));
+    }
 
     return builder.build();
   }


singhpk234:
This is a cool feature ! looking forward to the change :) ~

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -57,6 +60,16 @@ public class AwsCredentialsStorageIntegration
   private static final Logger LOGGER =
       LoggerFactory.getLogger(AwsCredentialsStorageIntegration.class);
 
+  // AWS STS Session Tag keys for CloudTrail correlation
+  static final String TAG_KEY_CATALOG = "polaris:catalog";
+  static final String TAG_KEY_NAMESPACE = "polaris:namespace";
+  static final String TAG_KEY_TABLE = "polaris:table";
+  static final String TAG_KEY_PRINCIPAL = "polaris:principal";
+  static final String TAG_KEY_REQUEST_ID = "polaris:request-id";
```
singhpk234: rather than having pre-defined tags how about taking tags as input in realm level confs ?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -130,10 +146,9 @@ public StorageAccessConfig getOrGenerateSubScopeCreds(
             allowedReadLocations,
             allowedWriteLocations,
             refreshCredentialsEndpoint,
-            includePrincipalNameInSubscopedCredential
-                ? Optional.of(polarisPrincipal)
-                : Optional.empty());
-    LOGGER.atDebug().addKeyValue("key", key).log("subscopedCredsCache");
+            includePrincipalInCacheKey ? Optional.of(polarisPrincipal) : Optional.empty(),
+            includeSessionTags ? Optional.of(credentialVendingContext) : Optional.empty());
+    LOGGER.atDebug().addKeyValue("key", key.toSanitizedLogString()).log("subscopedCredsCache");
```
dimas-b: Is this debug statement valuable at all? If we want to hide "potentially sensitive information", that reduces the usefulness of logging the key pretty much to zero, I guess 🤔 

To be clear, I support protecting "potentially sensitive information". I'm suggesting that this log statement may be removed now... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -144,7 +159,8 @@ public StorageAccessConfig getOrGenerateSubScopeCreds(
                   allowedReadLocations,
                   allowedWriteLocations,
                   polarisPrincipal,
-                  refreshCredentialsEndpoint);
+                  refreshCredentialsEndpoint,
+                  credentialVendingContext);
```
dimas-b: For the sake of correctness, we should use the value from the cache key here, because the key may be compared for returning previous entries, but `credentialVendingContext` here and in the key may have different values.

This applies to all values, but it's critical for `credentialVendingContext` and `polarisPrincipal`.

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
```diff
@@ -0,0 +1,99 @@
...
+ *   <li>{@code namespace} - The namespace/database being accessed (e.g., "db.schema")
+ *   <li>{@code tableName} - The name of the table being accessed
+ *   <li>{@code requestId} - A unique request identifier for correlation with catalog audit logs
+ * </ul>
+ *
+ * <p>These values appear in cloud provider audit logs (e.g., AWS CloudTrail), enabling
+ * deterministic correlation between catalog operations and data access events.
+ */
+@PolarisImmutable
+public interface CredentialVendingContext {
```
dimas-b, line 42: The code related to vended credential generation is in active development and refactoring now... Cf. #3270 ... So I guess this class may have to change too at some point. I hope you do not mind.

That said, I think introducing this context in the current codebase makes sense.

I'd like to propose moving the endpoint config and the principal name here too. Then, we could have a "reduce" method using `ReamConfig` to trim the information only to the required sub-set in order to have more efficient cache keys in environments where the new features are not required. WDYT?

e.g. (pseudo-code):

```
CredentialVendingContext reduce(RealmConfig config) {
  if (includePrincipalInCacheKey) {...}
}

File: runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java [hunk 5]
```diff
@@ -137,4 +150,56 @@ public StorageAccessConfig getStorageAccessConfig(
...
+   * Extracts the request ID from the current request context.
+   *
+   * <p>Note: we must avoid injecting {@link jakarta.ws.rs.container.ContainerRequestContext} here,
+   * because this may cause some tests to fail, e.g. when running with no active request scope.
+   *
+   * @return the request ID if available, empty otherwise
+   */
+  private Optional<String> getRequestId() {
+    // See org.jboss.resteasy.reactive.server.injection.ContextProducers
+    ResteasyReactiveRequestContext context = CurrentRequestManager.get();
```
dimas-b: Is it possible to avoid using `ResteasyReactiveRequestContext` here? Some access paths needing storage credentials may run outside of the REST API context (e.g. async tasks).

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -91,6 +91,28 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  /**
+   * When enabled, includes session tags (catalog, namespace, table, principal, request-id) in AWS
```
dimas-b: nit: the javadoc is mostly equivalent to the flag's descriptions... why bother duplicating that info?

dimas-b (commented):
Thanks for your contribution, @obelix74 !

The PR LGTM overall, my only real concern is relying on REST contexts.

Since this affects Polaris core, could you also send a discussion email to the `dev` ML for awareness?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
obelix74: Done. Removed the debug log statement since the sanitized key doesn't provide useful debugging information.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
obelix74: Good catch! Fixed - the loader now uses `k.credentialVendingContext().orElse(CredentialVendingContext.empty())` to ensure we use the context from the cache key rather than the local variable. This ensures consistent behavior when session tags are disabled.

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
obelix74: I agree this would be a cleaner design. 

However, this would touch multiple layers (StorageCredentialCache, StorageCredentialsVendor, PolarisCredentialVendor, and all storage integrations). Given that #3270 is also refactoring this area, would you prefer I:

1. Implement this in the current PR, or
2. Defer to a follow-up PR after #3270 merges to avoid conflicts?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
obelix74: Good point. Refactored to use `MDC.get(RequestIdFilter.REQUEST_ID_KEY)` instead. This works both within REST API request handlers and in async tasks where the MDC context is propagated, avoiding the dependency on Resteasy-specific classes.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
obelix74: Done. Removed the redundant javadoc comment since the description in the builder already documents the flag's behavior.

obelix74:
> Thanks for your contribution, @obelix74 !
> 
> The PR LGTM overall, my only real concern is relying on REST contexts.
> 
> Since this affects Polaris core, could you also send a discussion email to the `dev` ML for awareness?

Thanks. Sent an email to the ML.  I addressed most of your comments. I have a question on one of them.  Thanks again.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
singhpk234, line 100: i would recommend introducing a new function with this added arg signature.

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
```diff
@@ -0,0 +1,99 @@
...
+   * @return an empty context instance
+   */
+  static CredentialVendingContext empty() {
+    return ImmutableCredentialVendingContext.builder().build();
+  }
+
+  interface Builder {
+    Builder catalogName(String catalogName);
+
+    Builder catalogName(Optional<String> catalogName);
```
singhpk234: why not just have optional ? fill it as required ?

obelix74:
> Thanks for the change @obelix74 !
> 
> took an initial passs, i do agree with @dimas-b on sending rest context as part of tag, can you explain how do you wanna use this ?

I removed using `RestContext` and used `SLF4F#MDC` instead to get the request id in a commit that I pushed about 30 mins ago.  I am proposing using the request id (trace_id in the below picture) to tie in the request end to end from the client, all the way to the storage layer.  If the clients to the catalog are willing to share information about the SQLs executed, we can use the request id to tie it all together.

Do I make sense?

[
image
](url)

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
obelix74: Moved the tag key constants to `CredentialVendingContext` where they're more appropriately located alongside the context fields they describe. This makes them easier to find and customize.

For full realm-level configurability of tag keys, that would require adding new `FeatureConfiguration` entries and plumbing them through to the storage integrations. I can do it in this PR or as a follow up PR - please let me know. I am a bit concerned that this PR is already too big.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
obelix74: Good point on backward compatibility. Added a default method overload to `PolarisCredentialVendor` that delegates to the new signature with `CredentialVendingContext.empty()`. This maintains backward compatibility for existing callers. Applied the same pattern to `StorageCredentialsVendor`.

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
obelix74: Simplified the Builder interface to only expose `Optional<String>` overloads. Updated all callers to use `Optional.of()` when setting values. This removes the redundant String overloads and makes the API cleaner.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -118,6 +127,19 @@ public StorageAccessConfig getSubscopedCreds(
                           accountId)
                       .toJson())
               .durationSeconds(storageCredentialDurationSeconds);
+
+      // Add session tags when the feature is enabled
+      if (includeSessionTags) {
+        List<Tag> sessionTags =
+            buildSessionTags(polarisPrincipal.getName(), credentialVendingContext);
```
singhpk234: [optional] i would also add the activated roles

obelix74: Added.

obelix74:
@singhpk234 as discussed, removed `request_id` from this PR.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -366,4 +388,67 @@ private static String arnPrefixForPartition(String awsPartition) {
...
+    tags.add(
+        Tag.builder()
+            .key(CredentialVendingContext.TAG_KEY_TABLE)
+            .value(truncateTagValue(context.tableName().orElse(null)))
+            .build());
+    tags.add(
+        Tag.builder()
+            .key(CredentialVendingContext.TAG_KEY_ROLES)
+            .value(truncateTagValue(context.activatedRoles().orElse(null)))
+            .build());
```
singhpk234: minor: i would place this next to the prinicipal role for better readibility

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -366,4 +388,67 @@ private static String arnPrefixForPartition(String awsPartition) {
...
+
+  /**
+   * Builds a list of AWS STS session tags from the credential vending context and principal name.
+   * These tags will appear in CloudTrail events for correlation purposes.
+   *
+   * @param principalName the name of the principal requesting credentials
+   * @param context the credential vending context containing catalog, namespace, table, and roles
+   * @return a list of STS Tags to attach to the AssumeRole request
+   */
+  private List<Tag> buildSessionTags(String principalName, CredentialVendingContext context) {
```
singhpk234: should we move this to a dedicated utility class ?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -91,6 +91,19 @@ public static void enforceFeatureEnabledOrThrow(
...
+  public static final FeatureConfiguration<Boolean> INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL =
+      PolarisConfiguration.<Boolean>builder()
+          .key("INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL")
+          .description(
+              "If set to true, session tags (catalog, namespace, table, principal, roles) will be included\n"
+                  + "in AWS STS AssumeRole requests for credential vending. These tags appear in CloudTrail events,\n"
+                  + "enabling correlation between catalog operations and S3 data access.\n"
+                  + "Requires the IAM role trust policy to allow sts:TagSession action.\n"
+                  + "Note that enabling this feature leads to degradation in temporary credential caching as \n"
+                  + "catalog will no longer be able to reuse credentials for different tables/namespaces/roles.")
```
singhpk234:
```suggestion
                  + "Note that enabling this feature may lead to degradation in temporary credential caching as \n"
                  + "catalog will no longer be able to reuse credentials for different tables/namespaces/roles.")
```

i see allowedLocation is part of the table for general case we have overlapping check, having MAY is neutral in a sense it will not scare folks. 


though i am not sure why just principal is in the key and not the roles, @dimas-b you know why ?

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
dimas-b: Deferring this refactoring sound reasonable.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -366,4 +388,67 @@ private static String arnPrefixForPartition(String awsPartition) {
...
+    // This ensures consistent tag presence in CloudTrail for correlation
+    tags.add(
+        Tag.builder()
+            .key(CredentialVendingContext.TAG_KEY_PRINCIPAL)
+            .value(truncateTagValue(principalName))
+            .build());
+    tags.add(
+        Tag.builder()
+            .key(CredentialVendingContext.TAG_KEY_CATALOG)
+            .value(truncateTagValue(context.catalogName().orElse(null)))
```
dimas-b: Is having a null tag preferable to not having the tag at all? (just wondering)

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -130,21 +146,26 @@ public StorageAccessConfig getOrGenerateSubScopeCreds(
...
+            includeSessionTags ? Optional.of(credentialVendingContext) : Optional.empty());
     Function<StorageCredentialCacheKey, StorageCredentialCacheEntry> loader =
         k -> {
           LOGGER.atDebug().log("StorageCredentialCache::load");
+          // Use credentialVendingContext from the cache key for correctness.
+          // When session tags are disabled, the key stores Optional.empty(), and we should
+          // use an empty context to ensure consistent behavior regardless of what the
+          // local variable contains.
+          CredentialVendingContext contextFromKey =
+              k.credentialVendingContext().orElse(CredentialVendingContext.empty());
```
dimas-b: I believe this is still not ideal. The key might have an `Optional.empty()` but the storage integration code will use `CredentialVendingContext.empty()`... How can we be sure they signify the same thing? Why not just put `CredentialVendingContext.empty()` in the key and make this value non-optional in the key?

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisCredentialVendor.java
```diff
@@ -35,17 +35,64 @@ public interface PolarisCredentialVendor {
...
    * @param allowedWriteLocations a set of allowed to write locations
+   * @param polarisPrincipal the principal requesting credentials
    * @param refreshCredentialsEndpoint an optional endpoint to use for refreshing credentials. If
    *     supported by the storage type it will be returned to the client in the appropriate
    *     properties. The endpoint may be relative to the base URI and the client is responsible for
    *     handling the relative path
    * @return an enum map containing the scoped credentials
    */
   @Nonnull
+  default ScopedCredentialsResult getSubscopedCredsForEntity(
```
dimas-b, line 55: This method is not called in Polaris. Let's deprecate it for removal (in a later release).

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageCredentialsVendor.java
dimas-b, line 71: same here - this method in not used in Polaris. Let's deprecate.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: Principal was added as an small incremental change for a narrow use case, IIRC.

I think the whole flow for generating `StorageAccessConfig` needs some refactoring :sweat_smile: Cf. https://github.com/apache/polaris/pull/3327#discussion_r2651225156

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
obelix74: done.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
obelix74: Pros:

* Consistent tag presence in CloudTrail makes querying/filtering easier
* You can always filter by tag key without worrying about missing tags
* Makes it clear the tag was intentionally set vs. accidentally omitted

Cons:

* Slightly more data in CloudTrail events
* "unknown" values might be confusing

obelix74: Note that `truncateTagValue` sets the value to `unknown`.

dimas-b: Thanks for the explanation! I did miss the conversion to `unknown` in `truncateTagValue` :facepalm: but the new code is clear on that aspect :+1:

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -55,14 +56,47 @@ public interface StorageCredentialCacheKey {
...
+  CredentialVendingContext credentialVendingContext();
+
+  /**
+   * Returns a sanitized string representation of this cache key suitable for logging. This excludes
+   * potentially sensitive information like catalog names, namespaces, table names, and request IDs
+   * that may be present in the credential vending context.
+   *
+   * @return a sanitized string representation for logging
+   */
+  default String toSanitizedLogString() {
```
dimas-b: This method not used anymore... Would you mind removing?

dimas-b (commented):
LGTM with one remaining comment. Thanks for bearing with me, @obelix74  🙂

dimas-b (dismissed):
Let's wait a few days for other reviewers to have a chance at commenting since it's a holiday season 😉

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
```diff
@@ -965,4 +981,390 @@ public void testGetSubscopedCredsLongPrincipalName() {
...
+                return null;
+              }
+            },
+            () -> "realm");
+
+    // Simulate STS throwing AccessDeniedException when sts:TagSession is not allowed
+    // In AWS SDK v2, this is represented as StsException with error code "AccessDenied"
+    software.amazon.awssdk.services.sts.model.StsException accessDeniedException =
+        (software.amazon.awssdk.services.sts.model.StsException)
+            software.amazon.awssdk.services.sts.model.StsException.builder()
```
singhpk234: nit: can we remove inline imports, would be better to just import this normally and use ?

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -118,6 +124,19 @@ public StorageAccessConfig getSubscopedCreds(
                           accountId)
                       .toJson())
               .durationSeconds(storageCredentialDurationSeconds);
+
+      // Add session tags when the feature is enabled
+      if (includeSessionTags) {
+        List<Tag> sessionTags =
+            buildSessionTags(polarisPrincipal.getName(), credentialVendingContext);
```
singhpk234: we can send PolarisPrincipal as an arg here and can extract both principalName and `getRoles` from the prinicipal rather than making it part of credentialVendingContext ?

singhpk234 (dismissed):
LGTM, thanks @obelix74 for the all work !

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: I opened #3339 . If it's ok, let's merge it first, then rebase this PR.

obelix74: Done.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+   * @param context the credential vending context containing catalog, namespace, and table
+   * @return a list of STS Tags to attach to the AssumeRole request
+   */
+  public static List<Tag> buildSessionTags(
+      PolarisPrincipal principal, CredentialVendingContext context) {
+    List<Tag> tags = new ArrayList<>();
+
+    // Extract principal name and roles
+    String principalName = principal.getName();
+    Set<String> roles = principal.getRoles();
```
dimas-b: This makes `StorageAccessConfig` caching more obscure and potentially incorrect. As discussed in #3339 , the cache currently uses only the principal name as part of the key, so any changes in principal roles under the same name could result in false positive cache hits. False in the sense that the previous config with a credential session having old roles in the tags can be reused, while the principal's roles are different at access time.

All in all, this is a complex problem to fix in the current Polaris codebase. I believe it is best tackled in a dedicated PR.

For this PR, I'd propose to only use attributes that are covered by the storage credential's cache [key](https://github.com/apache/polaris/blob/27ae684304c6c62e5ee516404eee1419bc5d67f8/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java#L126).

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: Sorry, but I do not think the new code is correct. I added a specific comment here: https://github.com/apache/polaris/pull/3327#discussion_r2655536056

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
obelix74: I have reverted my changes and pushed a commit. Thanks @dimas-b .

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
obelix74: Reverted.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
singhpk234: GRANTs are done on roles and which is later used to infer the R/W location, nevertheless the CredentialVendingContext i think is added to the cache key right so roles becomes part of cache key ? i believe the existing cache will be cleared as soon as this changes are deployed since its in memory. I don't fully get the concern though.

Agree with @dimas-b we can have this discussion in a dedicated PR and work through it together, at this point fine with reverting :)

singhpk234 (approved):

obelix74:
I see that @jbonofre wants to review. I will wait for them.

dimas-b:
@jbonofre : I see you added yourself as a reviewer... Do you need more time or are you ok with merging?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
adnanhemani: +1 to everything that @singhpk234 said.

PR #3305, opened 2025-12-21
Enforce mypy for CLI

MonkeyCanCode:
This PR enforces mypy for CLI code base (https://github.com/apache/polaris/issues/3280). The current workflow was apply mypy to intergration_test only which doesn't make much sense. Also, due to the recent refactor of python cli, the lint was a bit off due to change of code base path. Thus, this got addressed in this PR as well.

## Checklist
- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [x] 💡 Added comments for complex logic
- [x] 🧾 Updated `CHANGELOG.md` (if needed)
- [x] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3330, opened 2025-12-24
Add KMS options to catalogs create CLI

dimas-b:
Following up on #2802

## Checklist

### DIFF
--- a/client/python/apache_polaris/cli/command/catalogs.py
+++ b/client/python/apache_polaris/cli/command/catalogs.py
@@ -183,7 +183,17 @@ class CatalogsCommand(Command):
                 )
 
     def _has_aws_storage_info(self):
-        return self.role_arn or self.external_id or self.user_arn or self.region or self.endpoint or self.endpoint_internal or self.sts_endpoint or self.path_style_access
+        return (self.role_arn
+                or self.external_id
+                or self.user_arn
+                or self.region
+                or self.endpoint
+                or self.endpoint_internal
+                or self.sts_endpoint
+                or self.current_kms_key
+                or self.allowed_kms_keys
+                or self.path_style_access
+                )
 
     def _has_azure_storage_info(self):
         return self.tenant_id or self.multi_tenant_app_name or self.consent_url


adnanhemani (approved):

client/python/apache_polaris/cli/command/__init__.py
```diff
@@ -72,6 +72,8 @@ def options_get(key, f=lambda x: x):
                 sts_endpoint=options_get(Arguments.STS_ENDPOINT),
                 sts_unavailable=options_get(Arguments.STS_UNAVAILABLE),
                 path_style_access=options_get(Arguments.PATH_STYLE_ACCESS),
+                current_kms_key=options_get(Arguments.KMS_KEY_CURRENT),
+                allowed_kms_keys=options_get(Arguments.KMS_KEY_ALLOWED),
```
singhpk234, line 76: does just setting `current_kms_key` automatically adds the `allowed_kms_keys` ?

client/python/apache_polaris/cli/constants.py
```diff
@@ -258,6 +260,12 @@ class Create:
                 "(Only for S3) Indicates that Polaris should not use STS (e.g. if STS is not available)"
             )
             PATH_STYLE_ACCESS = "(Only for S3) Whether to use path-style-access for S3"
+            KMS_KEY_CURRENT = (
+                "(Only for AWS S3) The AWS KMS key ARN to be used for encrypting new S3 data"
```
singhpk234, line 264: I am assuming this is required because we need to use this key to encrypt metadata.json ? as when we are vending creds we don't know which snapshot the client will be reading so we vend creds for all or we just give decrypt creds for allowed key and encrypt | decrypt creds for current keys ?

client/python/apache_polaris/cli/constants.py
```diff
@@ -258,6 +260,12 @@ class Create:
                 "(Only for S3) Indicates that Polaris should not use STS (e.g. if STS is not available)"
             )
             PATH_STYLE_ACCESS = "(Only for S3) Whether to use path-style-access for S3"
+            KMS_KEY_CURRENT = (
```
MonkeyCanCode, line 263: As these are optional and only for AWS, we may want to update `client/python/apache_polaris/cli/command/catalogs.py` as well for the function `_has_aws_storage_info()`. Here is a reference: https://github.com/apache/polaris/pull/3305/files#diff-a3e865c2a57514f7f505c706a3af70a5ac90b712f96656b513cdbfcee20c031eL181

client/python/apache_polaris/cli/command/__init__.py
dimas-b: yes

dimas-b: Currently "current" is the same as "allowed"... However, I believe only "current" needs write access... but that's in Polaris java code... it does not affect CLI.

client/python/apache_polaris/cli/constants.py
dimas-b: Polaris does not use KMS keys directly. It only generates AWS policies that allow those keys to be used on the AWS side when S3 requests are made. But, yes, the current key is used for writing new data. Zero or more additional keys are also allowed to be used because they might be required for dealing with old files.

client/python/apache_polaris/cli/constants.py
dimas-b: good point - updated

client/python/apache_polaris/cli/constants.py
singhpk234: > Polaris does not use KMS keys directly.

wouldn't we be needing this for encrypting / decrypting metadata.json ? 


> additional keys are also allowed to be used because they might be required for dealing with old files

I agree with additional keys but my question was why would Polaris vends creds for old kms keys for encrypting, files are immutable, so old keys should be vended for decrypt, similarly new key should have encrypt / decrypt. 
Do we vend creds for encryption and decryting for all key in our sts policy ?

dimas-b: > my question was why would Polaris vends creds for old kms keys for encrypting, [...]

Currently it does. However, this is beyond the scope of current PR (CLI). It's about the actual java code from #2802 :)

Normally, I'd think "additional" keys should get only decryption rights, but this may be tricky from the manual key rotation perspective.

MonkeyCanCode (approved):

client/python/apache_polaris/cli/constants.py
dimas-b: Filed #3338 for follow-up

PR #1517, opened 2025-05-03
[JDBC] Add retries with delay

singhpk234:
### About the change 

This change attempts to add retries in the JDBC persistence layer, these retries with jitter are tunable in the following ways : 
a. max_retries : total number of retries we expect the persistence to do before giving up. 
b. max_duaration_in_ms : time in ms since the first attempt this retries should be done. For ex on configured 500 ms the total time spent in retrying should not exceed 500ms (optimistically)
c. initial_delay_in_ms : intial delay before the first attempt 

--------

As a result of which we see the following improvements, with the setup of getting-started guide

Make 100% of the call pass 

```

========================================================================================================================
---- Global Information -------------------------------------------------------------|---Total---|-----OK----|----KO----
> request count                                                                      |    30,222 |    30,222 |         0
> min response time (ms)                                                             |         1 |         1 |         -
> max response time (ms)                                                             |       786 |       786 |         -
> mean response time (ms)                                                            |        17 |        17 |         -
> response time std deviation (ms)                                                   |        24 |        24 |         -
> response time 25th percentile (ms)                                                 |         5 |         5 |         -
> response time 50th percentile (ms)                                                 |        11 |        11 |         -
> response time 75th percentile (ms)                                                 |        23 |        23 |         -
> response time 99th percentile (ms)                                                 |        96 |        96 |         -
> mean throughput (rps)                                                              |     83.72 |     83.72 |         -
---- Response Time Distribution ----------------------------------------------------------------------------------------
> OK: t < 800 ms                                                                                         30,222   (100%)
> OK: 800 ms <= t < 1200 ms                                                                                   0     (0%)
> OK: t >= 1200 ms                                                                                            0     (0%)
> KO                                                                                                          0     (0%)
========================================================================================================================
```

Increased the TPS to 125 

```

========================================================================================================================
2025-05-05 21:01:49 GMT                                                                             360s elapsed
---- Requests -----------------------------------------------------------------------|---Total---|-----OK----|----KO----
> Global                                                                             |    37,305 |    37,305 |         0
> Authenticate                                                                       |         6 |         6 |         0
> Read / Check Namespace Exists                                                      |     2,059 |     2,059 |         0
> Read / Fetch all Views under parent namespace                                      |     2,095 |     2,095 |         0
> Write / Update Namespace Properties                                                |     6,199 |     6,199 |         0
> Read / Fetch all Tables under parent namespace                                     |     2,098 |     2,098 |         0
[...]

PR #3334, opened 2025-12-26
Relational JDBC: Reuse connection with execute select when within transaction

singhpk234:
### About the changes 

While reading code, i realized this particular check was not happening in the transaction i.e let say we observe a constrain violation because within a transaction we added some entries, we would not be able to find which entity caused the SQL constraint violation in the first place, we might get incorrect entity in the exception, though the exception remains the same. 

The proposed fix just uses the same conenction which is being used in the transaction and uses that to do the lookup to make sure the entity is found correctly. 

## Checklist

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
@@ -176,6 +176,21 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
       throws SQLException {
     ModelEntity modelEntity = ModelEntity.fromEntity(entity, schemaVersion);
     if (originalEntity == null) {
+      // Check if entity already exists before attempting INSERT to avoid constraint violations
+      // that would abort the transaction in PostgreSQL
+      PolarisBaseEntity existingEntity =
+          lookupEntityByName(
+              callCtx,
+              entity.getCatalogId(),
+              entity.getParentId(),
+              entity.getTypeCode(),
+              entity.getName(),
+              connection);
+
+      if (existingEntity != null) {
+        throw new EntityAlreadyExistsException(existingEntity);
+      }
+
       try {
         List<Object> values =
             modelEntity.toMap(datasourceOperations.getDatabaseType()).values().stream().toList();
@@ -188,22 +203,10 @@ public class JdbcBasePersistenceImpl implements BasePersistence, IntegrationPers
                 realmId));
       } catch (SQLException e) {
         if (datasourceOperations.isConstraintViolation(e)) {
-          PolarisBaseEntity existingEntity =
-              lookupEntityByName(
-                  callCtx,
-                  entity.getCatalogId(),
-                  entity.getParentId(),
-                  entity.getTypeCode(),
-                  entity.getName(),
-                  connection);
-          // This happens in two scenarios:
-          // 1. PRIMARY KEY violated
-          // 2. UNIQUE CONSTRAINT on (realm_id, catalog_id, parent_id, type_code, name) violated
-          // With SERIALIZABLE isolation, the conflicting entity may _not_ be visible and
-          // existingEntity can be null, which would cause an NPE in
-          // EntityAlreadyExistsException.message().
-          throw new EntityAlreadyExistsException(
-              existingEntity != null ? existingEntity : entity, e);
+          // Constraint violation despite our check above - this can happen due to race conditions
+          // where another transaction inserted the same entity between our SELECT and INSERT.
+          // We cannot lookup again here because PostgreSQL aborts the transaction after any error.
+          throw new EntityAlreadyExistsException(entity, e);
         }
         throw new RuntimeException(
             String.format("Failed to write entity due to %s", e.getMessage()), e);


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -431,15 +448,34 @@ public PolarisBaseEntity lookupEntityByName(
...
+   * @param connection optional connection to reuse; if null, a new connection will be obtained
+   * @return the entity if found, null otherwise
+   */
+  @Nullable
+  private PolarisBaseEntity getPolarisBaseEntity(
+      QueryGenerator.PreparedQuery query, @Nullable Connection connection) {
     try {
-      var results = datasourceOperations.executeSelect(query, new ModelEntity(schemaVersion));
+      List<PolarisBaseEntity> results =
+          connection != null
```
dimas-b: It looks like `connection` is `null` when called from `writeEntity`... Does that call path not have the same issue (reading in another transaction) that this PR is trying to fix?

singhpk234: writeEntity is an atomicOperation, its not wrapped within transaction, so if write failed due to constrain violation, it would have been because of an entity thats actually persisted, hence it would fine to do a lookup in new connection, unlike writeEntities which would conflict with the entities not committed yet. 
Let me know WDYT considering above, happy to wrap the writeEntity in a transaction too for connection reuse for consistency ?

dimas-b: I mean that `writeEntity` invokes this select as part of its execution, but the select will run in a new JDBC transaction (because `connection` is `null`). Basically, the change `persistEntity` (called by ` writeEntity`) will make and the select it may do on exception will run in different connections in runtime. The code of `persistEntity` declares a connection parameter, which makes the impression that _all_ the logic in that method runs over the same connection :thinking: 

I agree that there is no bug ATM, but the logic is very obscure now that connection may or may not be `null` (new).

Would it be possible to refactor the code such that `connection` parameters a non-null when present?

singhpk234: Attempted this in commit: https://github.com/apache/polaris/pull/3334/changes/192d7870869216fe24ba50a1a7fb36005c698d48, wdyt ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -142,6 +142,56 @@ public <T> List<T> executeSelect(
...
+   * Helper method to execute a SELECT query with a given connection and collect results.
+   *
+   * @param query : Query to execute
+   * @param converterInstance : Converter to transform ResultSet rows
+   * @param connection : Connection to use (must not be closed by this method)
+   * @return List of converted results
+   * @param <T> : Business entity class
+   * @throws SQLException : Exception during query execution
+   */
+  private <T> List<T> executeSelectWithConnection(
```
dimas-b, line 176: nit: this method is called from only one place... why not inline?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: LGTM - thx :+1:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -189,21 +203,10 @@ private void persistEntity(
...
-                  entity.getTypeCode(),
-                  entity.getName());
-          // This happens in two scenarios:
-          // 1. PRIMARY KEY violated
-          // 2. UNIQUE CONSTRAINT on (realm_id, catalog_id, parent_id, type_code, name) violated
-          // With SERIALIZABLE isolation, the conflicting entity may _not_ be visible and
-          // existingEntity can be null, which would cause an NPE in
-          // EntityAlreadyExistsException.message().
-          throw new EntityAlreadyExistsException(
-              existingEntity != null ? existingEntity : entity, e);
```
singhpk234, line 206: PG aborts the transaction when it encounters the error : 
```
If a statement within a transaction raises an exception, the entire transaction is 
  aborted and all statements are rolled back. The transaction remains in an aborted state, 
  and no further SQL commands will be executed until a ROLLBACK command is issued.
```

hence a lookup after write failure is not possible, the tests failed because of this. 
I changed to read before write attempt lets say we are working with SERIALIZABLE isolation would work now. I see and think more it was intentional :'( as the comment says, to do it this way worst case was we not able to get existing entity which would have been fine.

Now lookup before before write is gonna cause some perf issue (but only for certain cases) and wrapping writeEntity would elevate the scenario more, will need to run some benchmark before we check it in,

dimas-b: Thx for the research! That sounds like a good plan :+1:

dimas-b: Alternatively, we could explicitly start a new Tx for `lookupEntityByName` only on failure... WDYT?

singhpk234: precisely presently thats what we are doing in the current code, the connection is a separate connection than the one used in the transaction in cases of `writeEntities`  or a completely different connection in `writeEntity` i am also debating the same is it worth it, i will run benchmark for this 50:50 R/W to see how much we deviate from prev numbers as recorded here : https://github.com/apache/polaris/pull/1517#issue-3037027635

PR #3335, opened 2025-12-27
Documentation: Fix incorrect default metrics port

### LINKED ISSUE #3251: Documentation: Metrics endpoint information is incomplete and potentially incorrect
### Describe the bug

The current [telemetry docs](https://polaris.apache.org/in-dev/unreleased/telemetry/) has misleading information about the metrics endpoint that may cause confusion for users setting up Prometheus scraping.

  Issues Found

  1. Port number may be incorrect
  The documentation states:
  "they are available from Polaris's management interface (port 8282 by default)"

  However, in our deployment (and checking the default configuration), the management port is 8182, not 8282.

  2. Incomplete endpoint documentation
  The documentation only mentions /q/metrics:
  "under the path /q/metrics. For example, if the server is running on localhost, the metrics can be accessed via http://localhost:8282/q/metrics"

  However, Quarkus with Micrometer also exposes metrics at /metrics (the standard Prometheus endpoint), which is not documented.

  Testing Results

  When testing against a running Polaris 1.0 instance:

  # Both endpoints return identical metrics
  curl localhost:8182/metrics      # ✅ Works (standard Prometheus path)
  curl localhost:8182/q/metrics    # ✅ Works (Quarkus-specific path)

  Both endpoints return the exact same metric data.

  Why This Matters

  1. Standard Prometheus tooling expects /metrics - Most Prometheus configurations, ServiceMonitors, and scraping tools default to /metrics, not /q/metrics
  2. Port confusion - Users may waste time troubleshooting connectivity if the port number is wrong
  3. Real-world impact - In our case, our infrastructure was already scraping /metrics successfully, but the documentation made us question if we had configured it correctly

  Suggested Documentation Fix

  Update the telemetry documentation to:

  1. Correct the port number (or clarify it's configurable):
  Metrics are available from Polaris's management interface (port 8182 by default, configurable via quarkus.management.port)
  2. Document both endpoints:
  Metrics are exposed at two paths:
  - /metrics - Standard Prometheus endpoint (recommended for most use cases)
  - /q/metrics - Quarkus-specific management endpoint

  Both paths return identical metrics. For example, if the server is running on localhost:
  - http://localhost:8182/metrics
  - http://localhost:8182/q/metrics
  3. Add a note about Prometheus scraping:
  Note: Standard Prometheus configurations and Kubernetes ServiceMonitors typically use /metrics by default.

  Environment

  - Polaris Version: 1.2
  - Runtime: Quarkus 3.25.4
  - Deployment: Kubernetes

Happy to contribute the fix if everyone of us agree to this

rishii-19-works:
Fixes #3251

This PR corrects the default metrics port in the documentation
from 8282 to 8182, matching the actual running configuration.

### DIFF
--- a/site/content/in-dev/unreleased/telemetry.md
+++ b/site/content/in-dev/unreleased/telemetry.md
@@ -25,8 +25,8 @@ weight: 450
 ## Metrics
 
 Metrics are published using [Micrometer]; they are available from Polaris's management interface
-(port 8282 by default) under the path `/q/metrics`. For example, if the server is running on
-localhost, the metrics can be accessed via http://localhost:8282/q/metrics.
+(port 8182 by default) under the path `/q/metrics`. For example, if the server is running on
+localhost, the metrics can be accessed via http://localhost:8182/q/metrics.
 
 [Micrometer]: https://quarkus.io/guides/telemetry-micrometer
 


rishii-19-works:
Thanks for the review and merge!

PR #3339, opened 2025-12-30
Use most parameters from Credentials Cache key

dimas-b:
The Credentials Cache produces cache hits based on the cache key, but the cached data is generated from parameters that are not always reflected in the key (e.g. `PolaricPrincipal`). As a result, it is not evident that positive cache hits always represent exactly the same credentials that would be produced from scratch.

This change takes care of propagating all parameter through the cache key, except the entity, which is a bit complicated to handle and probably needs a separate PR.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
@@ -53,7 +53,7 @@ public interface StorageCredentialCacheKey {
   Optional<String> refreshCredentialsEndpoint();
 
   @Value.Parameter(order = 8)
-  Optional<String> principalName();
+  Optional<PolarisPrincipal> principal();
 
   static StorageCredentialCacheKey of(
       String realmId,
@@ -75,6 +75,6 @@ public interface StorageCredentialCacheKey {
         allowedReadLocations,
         allowedWriteLocations,
         refreshCredentialsEndpoint,
-        polarisPrincipal.map(PolarisPrincipal::getName));
+        polarisPrincipal);
   }
 }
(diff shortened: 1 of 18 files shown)


dimas-b:
CC: @tokoko

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -53,7 +53,7 @@ public interface StorageCredentialCacheKey {
   Optional<String> refreshCredentialsEndpoint();
 
   @Value.Parameter(order = 8)
-  Optional<String> principalName();
+  Optional<PolarisPrincipal> principal();
```
tokoko, line 56: I thought of doing something like this before. (using key to get the value feels right). Having said that, I convinced myself that it wasn't worth it because now you're using whole PolarisPrincipal object as part of the cache key which contains more than just the name. I doubt using roles and properties will cause cache misses all that much, but you're also relying on the fact that someone in the future won't go ahead and write something more dynamic in there (like a token which was discussed before iirc).

Or maybe I'm completely misunderstanding how this immutables library works...

dimas-b: right... the background for this is #3327 where, indeed, roles are used in STS calls

tokoko: okay, got it. I still think it's a ticking time bomb :laughing: especially that it will affect cache even if only `includePrincipalNameInSubscopedCredential` is activated.

I guess the only actual solution I see for these sorts of issues is to delegate key generation to Integration classes themselves:

- Manager always loads Integration class (might have to introduce some sort of a cache for Integration classes, maybe..)
- passes all the parameters and asks for a key
- checks the cache and returns if non-empty
- passes all the parameters again now asking for a value.

But that will need a major refactor that I'm still gonna get back to soon :laughing: Until then, I guess it's okay to merge this not to block the feature.

dimas-b: > I guess the only actual solution I see for these sorts of issues is to delegate key generation to Integration classes themselves

Yes, I actually thought about that exact approach too. I agree that it's ultimately the better solution.

I'm going to close this PR. I think there is no bug in current cache behaviour. Let's go for the proper solution.

PR #3340, opened 2025-12-30
Merge Polaris-Admin to Polaris-Server

flyrain:
Please check the motivation here, https://lists.apache.org/thread/fs7j81kv2kn0wsh05412yk6pp7w7r5nz

## Checklist

### DIFF
--- a/runtime/server/build.gradle.kts
+++ b/runtime/server/build.gradle.kts
@@ -38,18 +38,15 @@ val distributionElements by
   }
 
 dependencies {
+  implementation(project(":polaris-core"))
+  implementation(project(":polaris-version"))
+  implementation(project(":polaris-api-management-service"))
+  implementation(project(":polaris-api-iceberg-service"))
   implementation(project(":polaris-runtime-service"))
 
-  // Additional dependencies for CLI mode
-  implementation("io.quarkus:quarkus-picocli")
-  implementation(project(":polaris-version"))
-
-  // These are already provided transitively by polaris-runtime-service
-  // but we need them at compile time for CLI code
-  compileOnly(project(":polaris-core"))
-  compileOnly(project(":polaris-api-management-service"))
-  compileOnly(project(":polaris-api-iceberg-service"))
   compileOnly("com.fasterxml.jackson.core:jackson-annotations")
+
+  implementation("io.quarkus:quarkus-picocli")
 
   runtimeOnly("org.postgresql:postgresql")
   runtimeOnly(project(":polaris-relational-jdbc"))


dimas-b:
@flyrain : CI fails? :thinking:

runtime/server/src/main/java/org/apache/polaris/PolarisApplication.java
```diff
@@ -0,0 +1,57 @@
...
+import jakarta.inject.Inject;
+import org.apache.polaris.admintool.PolarisAdminTool;
+import picocli.CommandLine;
+
+/**
+ * Main entry point for Polaris application. Supports both HTTP server mode and CLI admin mode: - No
+ * arguments: Start HTTP server - With arguments: Run CLI commands (bootstrap, purge, etc.)
+ */
+@QuarkusMain
+public class PolarisApplication implements QuarkusApplication {
```
adutra, line 31: I think this could be simplified a bit:

```java
@QuarkusMain
public class PolarisApplication implements QuarkusApplication {

  /** Main method that detects CLI mode and activates the appropriate Quarkus profile. */
  public static void main(String... args) {
    if (args.length > 0) {
      System.setProperty("quarkus.profile", "cli");
      Quarkus.run(PicocliRunner.class, args);
    } else {
      Quarkus.run(PolarisApplication.class, args);
    }
  }

  @Override
  public int run(String... args) {
    Quarkus.waitForExit();
    return 0;
  }
}
```

I confirmed the above works, if you put back the `@TopCommand` annotation in `PolarisAdminTool`.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -293,3 +293,15 @@ quarkus.index-dependency.protobuf.artifact-id=protobuf-java
 # force the locale, just in case the system's using another default locale
 quarkus.default-locale=en_US
 
+# ---- CLI Mode Configuration (activated via -Dquarkus.profile=cli) ----
```
adutra: Rather than adding properties here, wdyt about introducing a profile-aware file `application-cli.properties`? I think this would more clearly demarcate the configuration for CLI from the one for server (which is also the default).

https://quarkus.io/guides/config-reference#profile-aware-files

adutra:
@flyrain the CI failure is an interesting one. After some investigation, it seems related to the colored output that we have by default when running server tests:

https://github.com/apache/polaris/blob/c4fc848bce5990ab44151980916a0c704696b79b/runtime/defaults/src/main/resources/application-test.properties#L28

Quarkus CLI tests seem unable to capture colored output properly.

This can be fixed in two ways: by removing the colored output completely for all tests (but that's frustrating) – or by implementing `getConfigProfile()` in `RelationalJdbcAdminProfile` and using another profile than `test`:

```java
public class RelationalJdbcAdminProfile extends RelationalJdbcProfile {
  @Override
  public Map<String, String> getConfigOverrides() {...}

  @Override
  public List<TestResourceEntry> testResources() {...}

  // ADD THIS
  @Override  public String getConfigProfile() { return "cli"; }
}
```

The above will make the test run under the `cli` profile and thus solves the issue. (We might end up needing a `cli-test` profile one day...)

runtime/defaults/src/main/resources/application.properties
```diff
@@ -293,3 +293,15 @@ quarkus.index-dependency.protobuf.artifact-id=protobuf-java
...
+# CLI-specific settings for admin tool commands
+%cli.quarkus.application.name=Apache Polaris Admin Tool (incubating)
+%cli.quarkus.banner.enabled=false
+%cli.quarkus.log.level=WARN
+%cli.quarkus.log.category."io.quarkus".level=ERROR
+%cli.quarkus.log.category."io.quarkus.agroal.deployment".level=ERROR
```
adutra: nit: unnecessary.

```suggestion
```

runtime/defaults/src/main/resources/application.properties
```diff
@@ -293,3 +293,15 @@ quarkus.index-dependency.protobuf.artifact-id=protobuf-java
...
+%cli.quarkus.log.category."org.hibernate.validator".level=ERROR
+%cli.quarkus.log.category."org.apache.polaris.service.config.ServiceProducers".level=ERROR
```
adutra: Rather than silencing the whole class, I'd suggest modifying the `maybeBootstrap` method as follows:

```java
  public void maybeBootstrap(
      @Observes Startup event,
      MetaStoreManagerFactory factory,
      PersistenceConfiguration config,
      RealmContextConfiguration realmContextConfiguration) {

    if (ConfigUtils.isProfileActive("cli")) {
      return; // never bootstrap realms in CLI profile
    }
...
```

I'd also suggest moving this method to a separate class, as it is becoming quite big and complex.

.github/workflows/gradle.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         run: |
           ./gradlew check sourceTarball distTar distZip publishToMavenLocal \
             -x :polaris-runtime-service:test \
-            -x :polaris-admin:test \
+            -x :polaris-server:test \
```
dimas-b, line 68: Why would we exclude server tests here? Should we just remove the old admin exclude?

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -295,7 +295,8 @@ jobs:
         run: |
           source "${LIBS_DIR}/_exec.sh"
 
-          exec_process ./gradlew :polaris-admin:assemble :polaris-admin:quarkusAppPartsBuild --rerun \
+          # polaris-admin merged into polaris-server - admin build handled by server build
+          exec_process ./gradlew :polaris-server:assemble :polaris-server:quarkusAppPartsBuild --rerun \
```
dimas-b: If Admin is merged into the server, it should probably not have a separate docker image? WDYT? 🤔

dimas-b: That said, I'd propose to tackle release workflow changes separately (for ease of review)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcConfiguration.java
```diff
@@ -18,8 +18,10 @@
  */
 package org.apache.polaris.persistence.relational.jdbc;
 
+import io.smallrye.config.ConfigMapping;
 import java.util.Optional;
 
+@ConfigMapping(prefix = "polaris.persistence.relational.jdbc")
```
dimas-b: Is this change required for the admin/server merge? If possible, let's separate different concerns in different PRs.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcProductionReadinessChecks.java
```diff
@@ -39,6 +40,11 @@ public ProductionReadinessCheck checkRelationalJdbc(
       return ProductionReadinessCheck.OK;
     }
 
+    // Skip check if datasource is not available (e.g., in CLI mode before configuration)
+    if (dataSource.isUnsatisfied() || dataSource.isAmbiguous()) {
+      return ProductionReadinessCheck.OK;
```
dimas-b: This is sub-optimal, IMHO. Could we not make readiness checks work normally for CLI too? It is fine to run different sets of checks for Admin CLI and Server runtime, but special cases like this in the main code look awkward to me.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcProductionReadinessChecks.java
```diff
@@ -48,6 +54,9 @@ public ProductionReadinessCheck checkRelationalJdbc(
                 "The current persistence (jdbc:h2) is intended for tests only.",
                 "quarkus.datasource.jdbc.url"));
       }
+    } catch (InactiveBeanException e) {
+      // Datasource is inactive (e.g., in CLI mode before configuration)
+      return ProductionReadinessCheck.OK;
```
dimas-b: (related to previous comment)... especially returning OK when something was not "active" 🤔

runtime/defaults/src/main/resources/application.properties
dimas-b: +1 to `application-cli.properties`

runtime/server/src/main/docker/Dockerfile.jvm
```diff
@@ -19,7 +19,7 @@
 FROM registry.access.redhat.com/ubi9/openjdk-21-runtime:1.24-2
 
 LABEL org.opencontainers.image.source=https://github.com/apache/polaris \
-      org.opencontainers.image.description="Apache Polaris (incubating)" \
+      org.opencontainers.image.description="Apache Polaris (incubating) - Server and Admin Tool" \
```
dimas-b: nit: I think the old description is sufficient

runtime/server/src/main/java/org/apache/polaris/admintool/PolarisAdminTool.java
```diff
@@ -36,8 +33,8 @@ public class PolarisAdminTool extends BaseMetaStoreCommand {
 
   @Override
   public Integer call() {
-    PrintWriter out = spec.commandLine().getOut();
-    spec.commandLine().usage(out);
+    // When invoked without a subcommand, show usage
```
dimas-b: How is it possible to invoke the Admin CLI _without_ a subcommand now? 🤔

runtime/server/build.gradle.kts
```diff
@@ -40,16 +40,36 @@ val distributionElements by
 dependencies {
   implementation(project(":polaris-runtime-service"))
 
+  // Dependencies from merged polaris-admin module
+  implementation(project(":polaris-core"))
```
dimas-b, line 41: Is this necessary? I suppose `runtime/server` included `polaris-core` before 🤔

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -75,6 +75,10 @@ public void warnOnFailedChecks(
       @Observes Startup event,
       Instance<ProductionReadinessCheck> checks,
       ReadinessConfiguration config) {
+    // Skip production readiness checks in CLI mode - they're only relevant for server deployments
+    if ("cli".equals(System.getProperty("quarkus.profile"))) {
```
dimas-b: Let's try to find a more natural way to perform readiness checks under the CLI env.

.asf.yaml
```diff
@@ -56,7 +56,7 @@ github:
           - "Unit Tests"
           - "Quarkus Tests"
           - "Quarkus Integration Tests"
-          - "Quarkus Admin Tests"
+          - "Quarkus Server Tests"
```
dimas-b: I believe we simply need to remove `Quarkus Admin Tests` here, not rename

runtime/server/src/main/java/org/apache/polaris/admintool/config/AdminToolProducers.java
```diff
@@ -33,6 +35,15 @@
...
 
+/**
+ * CDI producer for admin tool mode. These beans provide minimal implementations for CLI operations
+ * and take precedence over service beans when running admin commands.
+ *
+ * <p>Note: This class uses @Alternative with @Priority to override ServiceProducers beans that
+ * would conflict (Clock, PolarisDiagnostics, MetaStoreManagerFactory).
+ */
+@Alternative
+@Priority(100) // Higher priority than default to override ServiceProducers
```
dimas-b: I'm not sure how CDI actually works now that `AdminToolProducers` and `ServiceProducers` are part of the same class path 🤔 

@flyrain : do you have any insight? (I did not have time to debug personally 😅 )

runtime/server/src/main/java/org/apache/polaris/PolarisApplication.java
dimas-b: +1 to @adutra 's suggestion

.github/workflows/release-3-build-and-publish-artifacts.yml
flyrain: Sure, let's separate the release workflow to followup PRs

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcConfiguration.java
flyrain: It is not 100% necessary. We could finish this PR first, then rebase #3281 to tackle this issue.

runtime/server/src/main/java/org/apache/polaris/PolarisApplication.java
flyrain: Good idea, adapted in the new commit.

.github/workflows/release-3-build-and-publish-artifacts.yml
flyrain: Removed release related changes

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcConfiguration.java
flyrain: Removed this change.

.asf.yaml
flyrain: Removed, however, the CI is still asking for it as a pending task. Maybe we should remove it first in a separate PR?
Screenshot 2026-01-06 at 2 33 35 PM

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcProductionReadinessChecks.java
flyrain: Removed, I think these changes are not needed. Thanks for flagging it!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcProductionReadinessChecks.java
flyrain: Same here.

runtime/defaults/src/main/resources/application.properties
flyrain: Added application-cli.properties to host them. Thanks for the suggestion, @adutra !

runtime/server/build.gradle.kts
flyrain: Changed it to `compileOnly`, otherwise the compilation will fail.

runtime/server/src/main/docker/Dockerfile.jvm
flyrain: Removed

runtime/server/src/main/java/org/apache/polaris/admintool/config/AdminToolProducers.java
flyrain: Good Catch, Thanks @dimas-b ! Removed it. The `ServiceProducers` has covered all of them. No need to have this one.

.github/workflows/gradle.yml
flyrain: All tests now in `polaris-server` come from the removed module `polaris-admin`.  This is to keep the behavior unchanged.

runtime/server/src/main/java/org/apache/polaris/admintool/PolarisAdminTool.java
flyrain: Both work per my test. I have removed the comment to avoid confusion.

```
yufei@Yufeis-MacBook-Pro-2 polaris % java -jar runtime/server/build/quarkus-app/quarkus-run.jar -h
Usage: polaris-admin-tool.jar [-hV] [COMMAND]
Polaris administration & maintenance tool
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
Commands:
  help       Display help information about the specified command.
  bootstrap  Bootstraps realms and root principal credentials.
  purge      Purge realms and all associated entities.

yufei@Yufeis-MacBook-Pro-2 polaris % java -jar runtime/server/build/quarkus-app/quarkus-run.jar purge -h
Usage: polaris-admin-tool.jar purge [-hV] -r=<realm> [-r=<realm>]...
Purge realms and all associated entities.
  -h, --help            Show this help message and exit.
  -r, --realm=<realm>   The name of a realm to purge.
  -V, --version         Print version information and exit.

```

runtime/defaults/src/main/resources/application.properties
flyrain: Removed

runtime/defaults/src/main/resources/application.properties
flyrain: SGTM. Adopted the change. 

For moving it to a separate class, I'd suggest to tackle it in a followup PR given the current PR is big already.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
flyrain: Sure, I'm open for suggestions.

flyrain:
> ```java
> public class RelationalJdbcAdminProfile extends RelationalJdbcProfile {
>   @Override
>   public Map<String, String> getConfigOverrides() {...}
> 
>   @Override
>   public List<TestResourceEntry> testResources() {...}
> 
>   // ADD THIS
>   @Override  public String getConfigProfile() { return "cli"; }
> }
> ```
> 
> The above will make the test run under the `cli` profile and thus solves the issue. (We might end up needing a `cli-test` profile one day...)

Awesome! Thanks a lot for the suggestion, @adutra! All tests passed now!

.asf.yaml
flyrain: Filed https://github.com/apache/polaris/pull/3370 to unblock this PR.

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
adutra: It's safer to use the following idiom, as more than one profile can be specified with `quarkus.profile`:

```java
if (ConfigUtils.isProfileActive("cli")) {
```

flyrain: Fixed in the new commit. Thanks for the suggestion.

runtime/server/build.gradle.kts
dimas-b: Oh, I did not mean to imply that the dep. was wrong... I was just wondering :slightly_smiling_face: 

If core is require for complication, we should probably keep it as an `implementation` dep.

runtime/server/src/main/java/org/apache/polaris/admintool/config/AdminToolProducers.java
dimas-b: sweet :+1:

.github/workflows/gradle.yml
dimas-b: make sense. I was confused between `runtime/server` and `runtime/service` 🤦

runtime/defaults/src/main/resources/application-cli.properties
```diff
@@ -0,0 +1,42 @@
...
+# application during the build phase and cannot be changed at runtime.
+#
+# See: https://quarkus.io/guides/config-reference#build-time-configuration
+
+# Application identity
+quarkus.application.name=Apache Polaris Admin Tool (incubating)
+quarkus.banner.enabled=false
+
+# Logging configuration - suppress verbose output for CLI
+quarkus.log.level=WARN
```
dimas-b, line 40: Should we suppress CONSOLE log instead but keep default levels informative in case a user enabled FILE logging?

runtime/distribution/bin/admin
```diff
@@ -21,12 +21,16 @@ set -euo pipefail
 
 # Get the directory
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-cd "$SCRIPT_DIR/../admin"
+cd "$SCRIPT_DIR/../server"
```
dimas-b, line 24: for follow-up: with a single binary package, we may want to rename / restructure the binary dist for clarity.

runtime/distribution/bin/admin
```diff
@@ -21,12 +21,16 @@ set -euo pipefail
...
 # Get the Java command
 JAVA_CMD="${JAVA_HOME:+${JAVA_HOME%/}/bin/java}"
 JAVA_CMD="${JAVA_CMD:-$(command -v java)}"
 [ -x "$JAVA_CMD" ] || { echo "Java not found – set JAVA_HOME or add java to PATH." >&2; exit 1; }
 
-# Launch Quarkus
-exec "$JAVA_CMD" ${POLARIS_JAVA_OPTS:-} -jar quarkus-run.jar "$@"
+# Show the usage when no argument is provided
+if [ "$#" -eq 0 ]; then
+  set -- -h
```
dimas-b: nit: `--help`?

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
```diff
@@ -75,6 +76,10 @@ public void warnOnFailedChecks(
       @Observes Startup event,
       Instance<ProductionReadinessCheck> checks,
       ReadinessConfiguration config) {
+    // Skip production readiness checks in CLI mode - they're only relevant for server deployments
+    if (ConfigUtils.isProfileActive("cli")) {
```
dimas-b, line 80: This is probably ok in the interest of progress, but in general I hope we could leverage profile-specific config like `polaris.production.readiness.checks.enabled=true` (or `false` for CLI). WDYT?

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -274,6 +275,9 @@ public void maybeBootstrap(
       MetaStoreManagerFactory factory,
       PersistenceConfiguration config,
       RealmContextConfiguration realmContextConfiguration) {
+    if (ConfigUtils.isProfileActive("cli")) {
```
dimas-b, line 278: Same here: I'd prefer to use explicit config flags (set differently in each profile) vs. checking the name of the profile, please.

runtime/defaults/src/main/resources/application-cli.properties
flyrain: I'm open for that, but the `WARN` level logging was carried from the application.properties in the removed `admin` module, https://github.com/apache/polaris/blob/476301b9f8bc20f07ccb1ac9bbb83f88843920d9/runtime/admin/src/main/resources/application.properties#L67. Should we change it now?

runtime/distribution/bin/admin
flyrain: Definitely,  there are a few places needed to be changed, including doc. I will address them in followups.

runtime/defaults/src/main/resources/application-cli.properties
dimas-b: ah, sorry, I missed that. Current code is fine for now :+1:

runtime/distribution/bin/admin
flyrain: Fixed

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
flyrain: That's probably a good idea, esp. if we want to group a set of methods under the same configure key, instead of spreed `if (ConfigUtils.isProfileActive("cli")` in multiple places. We could improve in followups.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
flyrain: See https://github.com/apache/polaris/pull/3340#discussion_r2670309352

runtime/service/src/main/java/org/apache/polaris/service/config/ProductionReadinessChecks.java
dimas-b: I believe using dedicated config properties for this is a small change :) but I'm also ok with addressing this in a follow-up PR.

runtime/distribution/bin/admin
```diff
@@ -21,12 +21,16 @@ set -euo pipefail
...
 
 # Get the Java command
 JAVA_CMD="${JAVA_HOME:+${JAVA_HOME%/}/bin/java}"
 JAVA_CMD="${JAVA_CMD:-$(command -v java)}"
 [ -x "$JAVA_CMD" ] || { echo "Java not found – set JAVA_HOME or add java to PATH." >&2; exit 1; }
 
-# Launch Quarkus
-exec "$JAVA_CMD" ${POLARIS_JAVA_OPTS:-} -jar quarkus-run.jar "$@"
+# Show the usage when no argument is provided
+if [ "$#" -eq 0 ]; then
```
adutra, line 32: Nice 👍

runtime/server/build.gradle.kts
```diff
@@ -40,6 +40,17 @@ val distributionElements by
 dependencies {
   implementation(project(":polaris-runtime-service"))
 
+  // Additional dependencies for CLI mode
+  implementation("io.quarkus:quarkus-picocli")
+  implementation(project(":polaris-version"))
+
+  // These are already provided transitively by polaris-runtime-service
+  // but we need them at compile time for CLI code
+  compileOnly(project(":polaris-core"))
```
adutra: Agree with @dimas-b these should be `implementation` for most of them. I would reorder them as below:

```kotlin
    implementation(project(":polaris-core"))
    implementation(project(":polaris-version"))
    implementation(project(":polaris-runtime-service"))

    compileOnly("com.fasterxml.jackson.core:jackson-annotations")

    implementation("io.quarkus:quarkus-picocli")
```

runtime/server/src/main/java/org/apache/polaris/PolarisApplication.java
```diff
@@ -0,0 +1,72 @@
...
+   * ConfigUtils.getProfiles().add()) because Quarkus reads the active profile during early
+   * bootstrap, before ConfigUtils can affect the configuration.
+   *
+   * <p>Configuration Priority Order: Quarkus loads configuration from multiple sources (highest to
+   * lowest priority):
+   *
+   * <ol>
+   *   <li>(400) System properties: -D flags at JVM startup
+   *   <li>(300) Environment variables
+   *   <li>(260) External config/application-cli.properties: Profile-specific external config
```
adutra, line 46: As I mentioned on the ML, I'm unsure whether we should mention the possibility of adding this profile-aware external config file, as it has very little added value.

dimas-b: +1 Let's advise users to customize via the profile-neutral `application.properties` (and/or corresponding `-D` and env. vars).

flyrain: If a user runs both the server and the CLI with the same Polaris binary of the same location, a shared `application.properties` applies to both use cases. In that setup, having a separate `application-cli.properties` still seems useful to capture CLI specific configuration that should not affect the server behavior. That separation feels reasonable and keeps the shared defaults clean while allowing targeted overrides for the CLI. For example, CLI specific configs like the following will be overwritten by `application.properties`
```
quarkus.banner.enabled=false
quarkus.log.level=WARN
```

WDYT?

runtime/server/build.gradle.kts
flyrain: Fixed

runtime/server/src/main/java/org/apache/polaris/PolarisApplication.java
dimas-b: AFAIK, profile-specific config works in conjunction with the profile-neutral config. In other words, `application.properties` is _always_ loaded. This might be too much low-level details for end users to think about.

Even if all the code is bundled in the same Quarkus application, Server and Admin could still have distinct shell entry points and distinct `application.properties` files (each in its own directory). That would be my preference for the tar-based distribution.

Docker images can bundle anything that Polaris developers are ok with. In docker, users will override config via external mechanisms like env. variables. If a config map is used to define `application.properties` it will be specific to a particular runtime scenario anyway (Server or Admin, but not both).

flyrain: > profile-specific config works in conjunction with the profile-neutral config

That's true, and I can confirm that's how it work today.  

> This might be too much low-level details for end users to think about.

This isn't a user-face doc. This is a comment in the source code. I think it should be fine to be here to provide complete view of how it works.

> Even if all the code is bundled in the same Quarkus application, Server and Admin could still have distinct shell entry points and distinct application.properties files (each in its own directory). That would be my preference for the tar-based distribution.
> Docker images can bundle anything that Polaris developers are ok with. In docker, users will override config via external mechanisms like env. variables. If a config map is used to define application.properties it will be specific to a particular runtime scenario anyway (Server or Admin, but not both).

I understand that these particular use cases may not require profile specific config files today. That said, this does not mean the scenarios I mentioned will never occur. Providing clear and explicit messaging still has value in that context. Even if a user never actively uses `External config/application-cli.properties` as a profile specific external config, having a clear understanding of how it works can be important when diagnosing or debugging unexpected behavior.

runtime/server/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
dimas-b, line 1: Sorry for the last minute comment here, but I wonder if it might make sense to keep the `runtime/admin` module as a home for Admin code, but not build a Quarkus app from it. The `runtime/server` module will then import `runtime/admin` and we end up with one Quarkus app both for the Server and Admin Tool... WDYT?

My rationale for keeping the `runtime/admin` module is just better code isolation at compile time. The `runtime/server` module is mostly an assembly descriptor, ATM it does not have any feature code. It might be preferable to keep it this way to allow simpler reuse of feature code from their dedicated modules.

adutra: +1, that's actually a nice idea!

flyrain: I see the appeal of this approach, but I worry it may dismiss some of the key benefits of the current refactor(we will need a few common modules to enable a separate admin module), and make the dependency structure less clearer. Please note that, the CLI will need `ServiceProvider` to work now.
It also makes testing a bit awkward. Tests that are specific to the CLI behavior would have to depend on the `default` and `service` module just to pick up `application-cli` and `application` properties , which feels indirect and less explicit. For those reasons, I think the current direction still has advantages in terms of modularity, clarity, and test ergonomics.

dimas-b: I agree on "test ergonomics" as an advantage in the current diff of this PR, but I do not think "modularity" is better - that's exactly what I meant to highlight in my comment above :)

IMHO, better modularity is preferable to "test ergonomics".

dimas-b: Testing the admin tool could follow the same approach as testing OPA (which is not monolithic with the server, but a pure puglin)... WDYT?

flyrain: We will need a few common modules to enable a separate admin module. I believe this is one of areas we used to struggle with.

flyrain: In terms of modularity, splitting functionalities across multiple modules does not automatically improve modularity. In many cases it does the opposite. It makes the dependency graph harder to reason about, and forces consumers and tests to pull in modules that exist only to host a thin slice of logic. I think good modularity is about cohesion and clear ownership. A module should group functionality that changes together and represents a meaningful boundary. And I'm not a fan of how we layout OPA modules. 

In practice, over splitting often trades real modularity for superficial structure, while introducing costs in test ergonomics, dependency clarity, and mental overhead. Keeping related behavior together, even if it is small, is often the more modular choice. I have mentioned that we have to keep introducing new modules like `runtime/common`, `runtime/test-common`, `runtime/distribution`, they are the side effect of over splitting, which can be avoided in the beginning.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -292,4 +292,3 @@ quarkus.index-dependency.protobuf.artifact-id=protobuf-java
 
 # force the locale, just in case the system's using another default locale
 quarkus.default-locale=en_US
-
```
snazy, line 309: Unrelated change to this already big PR.

runtime/service/build.gradle.kts
```diff
@@ -33,8 +33,8 @@ dependencies {
 
   runtimeOnly(project(":polaris-relational-jdbc"))
 
-  implementation(project(":polaris-runtime-defaults"))
   implementation(project(":polaris-runtime-common"))
+  implementation(project(":polaris-runtime-defaults"))
```
snazy, line 37: Noop change in this big PR.

bom/build.gradle.kts
```diff
@@ -89,8 +89,6 @@ dependencies {
 
     api(project(":polaris-extensions-auth-opa"))
 
-    api(project(":polaris-admin"))
-    api(project(":polaris-runtime-common"))
```
snazy, line 95: This should not be removed from the bom

LICENSE
```diff
@@ -328,7 +328,6 @@ This product includes code from Project Nessie.
 * tools/minio-testcontainer/src/main/java/org/apache/polaris/test/minio/MinioAccess.java
 * tools/minio-testcontainer/src/main/java/org/apache/polaris/test/minio/MinioContainer.java
 * tools/minio-testcontainer/src/main/java/org/apache/polaris/test/minio/MinioExtension.java
-* runtime/admin/src/main/java/org/apache/polaris/admintool/PolarisAdminTool.java
```
snazy, line 335: This file has been moved, not removed.

runtime/admin/distribution/LICENSE
snazy, line 1: The LICENSE file contains entries that are not migrated.

snazy (changes requested):
This change effectively removed and required LICENSE entries.

This change will also "leak" all server parts into the admin-tool (+ vice versa) and cause (existing and future) CDI components to be initialized even if those must/should not run in the admin-tool. Everybody contributing code to Polaris server must now fully understand and consider this non-obvious behavior (e.g. observing the startup event). Working around the previous example would require conditional `if` statements scattered around the code base.

Users who would like to use configuration profiles for other more advanced purposes would have to change the production code. See these "if configuration-profile-**name** equals `cli`, which would not work for other profile names.

All changes and additions to the non-profile scoped configuration file "leak" into both the server and admin-tool, which can have unintended consequences, which are hard to understand and likely surprising, especially for new contributors.

I think we should be really careful with such an intrusive change.

If the main concern of this change is to address the admittedly long LICENSE + NOTICE file contents for server + admin-tool, I propose to split those into into one shared plus specific sets and join those during the build.

runtime/server/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
adutra: I don't think keeping the admin module around is "over-splitting". 

This PR, while interesting in that it achieves a single artifact/jar for both usages (server and CLI), does create an imbalanced layout where the top module includes code for the CLI, but the code for the server lives in a dependency:

```mermaid
flowchart TD
    A["`polaris-server<br/>(incl. admin tool)`"] --> B
    B[polaris-runtime-service]
```

Whereas @dimas-b proposal still keeps the single entrypoint but each usage has its own module:

```mermaid
flowchart TD
    A[polaris-server] --> B
    A --> C
    B[polaris-runtime-service] --> D
    C[polaris-admin] --> E
    B --> F
    C --> F
    D[specific deps]
    E[specific deps]
    F[common deps]
```

This has a net advantage imho of leaving `polaris-server` as a mere "assembly descriptor" as @dimas-b puts it.

> It makes the dependency graph harder to reason about, and forces consumers and tests to pull in modules that exist only to host a thin slice of logic

I don't buy much this argument: while a proliferation of modules can complicate the dependency graph, modularity offers many upsides: modern build tools efficiently manage even hundreds of dependencies; a DAG enforces clearer mental models about dependencies than monoliths that often create dependency cycles; smaller modules accelerate compilation; and downstream integrators benefit from leaner CDI contexts.

Polaris is currently in an intermediate, non-optimal state: it's neither a monolith, nor a well organized modular application. We should rethink the purpose of those "common" modules, but also consider separating disparate concerns in modules like `polaris-runtime-service`: what does telemetry have to do with authentication? And what does the secrets manager have to do with events processing?  I support moving towards a well-thought modular organization, and I think @dimas-b suggestion goes in that direction.

dimas-b: @flyrain : WDYT about this proposal?

LICENSE
dimas-b: Good point. We should just update the path (and possibly move this line if sort order matters).

PR #3350, opened 2026-01-04
Validate namespace depth during creation to prevent oversized AWS STS policies

### LINKED ISSUE #3243: Polaris fails to create a Table because of an STS policy size limit
### Describe the bug

While running benchmarks with much longer namespace names, I got the following error while creating a table:

```
Failed to get subscoped credentials: Packed policy consumes 118% of allotted space, please use smaller policy.
```

The path that caused this issue is composed by 14 nested namespaces, each with a 32 characters name.  The problem is located in https://github.com/apache/polaris/blob/31c7784630889d8fb6f0156610b8e59a4021f1c3/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java#L175.  It results in the following STS policy:

```

{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "s3:PutObject",
        "s3:DeleteObject"
      ],
      "Resource": "arn:aws:s3:::my-very-long-s3-bucket/catalog/f5a4d86558ed1f7fddec42ce11d8ee3a/e3fa7fd44796b949fce728cb334f3c15/f031eb9dc709150a3e1e9a76e9af550a/5598924229adee97260cda483d70674c/0857633d48470f538d1fc4cdc789c753/c4c06b9b673a710f7ce865690ff8797b/9594c7f9e8a1cd28054ff71b933fdc3b/0018c223e978d13aeec87488bc333c70/d4c484ca41745b69286372286f3f30aa/362f8d6579892af8a9c5f38b4e664b07/f28f3a8b846271c68bdbc09992d88d28/915107073e394e3471173ec4633137fa/5eac9dac6dd84afcb35f4805af9d8b34/e1a24630cbf9ced9dffcc123b70b2e43/ab4ffa55f688360e0c12aef543c18351/*"
    },
    {
      "Effect": "Allow",
      "Action": "s3:ListBucket",
      "Resource": "arn:aws:s3:::my-very-long-s3-bucket",
      "Condition": {
        "StringLike": {
          "s3:prefix": "catalog/f5a4d86558ed1f7fddec42ce11d8ee3a/e3fa7fd44796b949fce728cb334f3c15/f031eb9dc709150a3e1e9a76e9af550a/5598924229adee97260cda483d70674c/0857633d48470f538d1fc4cdc789c753/c4c06b9b673a710f7ce865690ff8797b/9594c7f9e8a1cd28054ff71b933fdc3b/0018c223e978d13aeec87488bc333c70/d4c484ca41745b69286372286f3f30aa/362f8d6579892af8a9c5f38b4e664b07/f28f3a8b846271c68bdbc09992d88d28/915107073e394e3471173ec4633137fa/5eac9dac6dd84afcb35f4805af9d8b34/e1a24630cbf9ced9dffcc123b70b2e43/ab4ffa55f688360e0c12aef543c18351/*"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": "s3:GetBucketLocation",
      "Resource": "arn:aws:s3:::my-very-long-s3-bucket"
    },
    {
      "Effect": "Allow",
      "Action": [
        "s3:GetObject",
        "s3:GetObjectVersion"
      ],
      "Resource": "arn:aws:s3:::my-very-long-s3-bucket/catalog/f5a4d86558ed1f7fddec42ce11d8ee3a/e3fa7fd44796b949fce728cb334f3c15/f031eb9dc709150a3e1e9a76e9af550a/5598924229adee97260cda483d70674c/0857633d48470f538d1fc4cdc789c753/c4c06b9b673a710f7ce865690ff8797b/9594c7f9e8a1cd28054ff71b933fdc3b/0018c223e978d13aeec87488bc333c70/d4c484ca41745b69286372286f3f30aa/362f8d6579892af8a9c5f38b4e664b07/f28f3a8b846271c68bdbc09992d88d28/915107073e394e3471173ec4633137fa/5eac9dac6dd84afcb35f4805af9d8b34/e1a24630cbf9ced9dffcc123b70b2e43/ab4ffa55f688360e0c12aef543c18351/*"
    }
  ]
}
```

I don't think we can eliminate the STS limitation at all.  But I think Polaris could be improved to better deal with this error.  For example:

* The returned HTTP code (400) is correct (Bad Request) but the associated error message should be clearer for the user.  The current one may not be understandable for everyone.
* During namespace creation, no check is performed on the total length of the path.  It might be worth adding so that the failure happens earlier.
* Parts of the STS policy could be merged together (AFAICT): the `s3:PutObject` and `s3:DeleteObject` statements could be merged with the `s3:GetObject` and `s3:GetObjectVersion` statements.  This would allow Polaris to support longer paths, while not removing the maximum path limitation entirely though.

### To Reproduce

* Create a deeply nested namespace structure with long length (e.g. `f5a4d86558ed1f7fddec42ce11d8ee3a/e3fa7fd44796b949fce728cb334f3c15/f031eb9dc709150a3e1e9a76e9af550a/5598924229adee97260cda483d70674c/0857633d48470f538d1fc4cdc789c753/c4c06b9b673a710f7ce865690ff8797b/9594c7f9e8a1cd28054ff71b933fdc3b/0018c223e978d13aeec87488bc333c70/d4c484ca41745b69286372286f3f30aa/362f8d6579892af8a9c5f38b4e664b07/f28f3a8b846271c68bdbc09992d88d28/915107073e394e3471173ec4633137fa/5eac9dac6dd84afcb35f4805af9d8b34/e1a24630cbf9ced9dffcc123b70b2e43/ab4ffa55f688360e0c12aef543c18351`)
* Create a table in the last namespace

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Additional context

_No response_

### System information

_No response_

rishii-19-works:
This PR adds early validation for excessively deep namespace hierarchies during namespace creation.

Without validation, deeply nested namespaces can generate oversized AWS STS policies that fail late with unclear AWS-level errors. This change fails fast with a clear Polaris-level message while ensuring existing namespaces remain unaffected.

Fixes #3243

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
@@ -91,8 +91,7 @@ public class NamespaceEntity extends PolarisEntity implements LocationBasedEntit
                 "Namespace depth %d exceeds supported limit of %d. "
                     + "Deeply nested namespaces can generate oversized AWS STS policies. "
                     + "Consider flattening the namespace hierarchy.",
-                depth,
-                MAX_NAMESPACE_DEPTH));
+                depth, MAX_NAMESPACE_DEPTH));
       }
 
       setType(PolarisEntityType.NAMESPACE);


polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
```diff
@@ -75,8 +77,23 @@ public String getBaseLocation() {
...
   public static class Builder extends PolarisEntity.BaseBuilder<NamespaceEntity, Builder> {
+
+    private static final int MAX_NAMESPACE_DEPTH = 10;
+
     public Builder(Namespace namespace) {
       super();
+      Preconditions.checkArgument(namespace != null, "Namespace must not be null");
+
+      int depth = namespace.length();
+      if (depth > MAX_NAMESPACE_DEPTH) {
```
dimas-b, line 88: I do not think namespace depth is the critical parameter for #3243. It's rather the total policy size, of which the table's full Polaris name is a factor (but not the only factor). I do not think restricting namespace depth will fix #3243.

Moreover, #3243 is specific to AWS. Any fix for #3243 should ideally not impact other storage systems.

dimas-b (commented):
Thanks for your contribution, @rishii-19-works !

Unfortunately, this PR does not look like an acceptable fix for #3243 from my POV. Specific comment below.

polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
rishii-19-works: Thanks for the detailed explanation, that makes sense.

I see your point — namespace depth alone is only one contributor to the overall policy size, and enforcing a hard depth limit in the core entity layer is neither sufficient nor storage-agnostic.

I agree that an acceptable fix for #3243 should focus on AWS-specific policy size handling rather than restricting namespace structure globally.

I’ll step back on this approach and either close this PR or rework it based on a more appropriate AWS-scoped solution. Thanks again for the guidance.

PR #3352, opened 2026-01-05
RELATIONAL-JDBC: Add support for cockroach DB

singhpk234:
### About the change

Add support of cockroach DB in relational-jdbc persistence 

There was pr long ago :
 - https://github.com/apache/polaris/pull/2485

seems like the author got busy, all pending was to testing it E2E with cockroach with all the integ tests along with making it work with schema evolution logic. also marking them as co-author for their contribution.
Happy to close it if they wanna resume their original pr 

co-author : @sathesuraj

## Checklist
- [ ] 🧾 Updated `CHANGELOG.md` (if needed) -- will add later
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed) - will add later


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -137,6 +139,14 @@ public void writeEntities(
       @Nonnull PolarisCallContext callCtx,
       @Nonnull List<PolarisBaseEntity> entities,
       List<PolarisBaseEntity> originalEntities) {
+    Set<Long> seenEntityIds = new HashSet<>();
+    for (PolarisBaseEntity entity : entities) {
+      if (!seenEntityIds.add(entity.getId())) {
+        throw new RetryOnConcurrencyException(
+            "Multiple updates to entity id '%s' in the same transaction", entity.getId());
+      }
+    }
```
singhpk234: Here's the difference and why its required:

  PostgreSQL SERIALIZABLE:
  - Uses Serializable Snapshot Isolation (SSI)
  - When it detects a write-write conflict within the same transaction, it typically throws SQLSTATE 40001 (serialization_failure) immediately
  - It's designed to fail fast rather than retry

  CockroachDB SERIALIZABLE:
  - Has automatic retry logic built-in at the database level
  - When it detects a conflict, it enters an internal retry loop hoping the conflict resolves
  - For conflicts within the same transaction (unresolvable), it can get stuck retrying

dimas-b:
> co-author : @sathesuraj

Please use the standard `Co-Authored-By` tag in commits :)

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

runtime/test-common/src/main/java/org/apache/polaris/test/commons/CockroachRelationalJdbcLifeCycleManagement.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Override
+  public void init(Map<String, String> initArgs) {
+    initScript = initArgs.get(INIT_SCRIPT);
+  }
+
+  @Override
+  public Map<String, String> start() {
+    // CockroachDB testcontainers uses PostgreSQL wire protocol compatibility
+    cockroach = new CockroachContainer("cockroachdb/cockroach:v24.3.0");
```
dimas-b: Please use `ContainerSpecHelper` to enable managing docker image versions via Renovate. 

Cf. https://github.com/apache/polaris/blob/5d77feb2e6699cacf229b56a45c615a247b5c43f/runtime/test-common/src/main/java/org/apache/polaris/test/commons/PostgresRelationalJdbcLifeCycleManagement.java#L46

persistence/relational-jdbc/src/main/resources/cockroachdb/schema-v4.sql
```diff
@@ -0,0 +1,146 @@
...
+    event_type TEXT NOT NULL,
+    timestamp_ms BIGINT NOT NULL,
+    principal_name TEXT,
+    resource_type TEXT NOT NULL,
+    resource_identifier TEXT NOT NULL,
+    additional_properties JSONB NOT NULL DEFAULT '{}'::JSONB,
+    PRIMARY KEY (event_id)
+);
+
+-- INT4 type used directly in table definitions for CockroachDB JDBC compatibility
```
dimas-b, line 305: It is probably preferable to move this comment to the top of this file for visibility. WDYT?

persistence/relational-jdbc/src/main/resources/cockroachdb/schema-v4.sql
dimas-b, line 1: Since CockroachDB uses mostly the same Polaris code a PostgreSQL, I believe the schema versions are best kept in sync. I do not think we have to support old schemas in CockroachDB, but the current one should probably be `v3` as our PostgreSQL schema... WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Why / when would this code be invoked with multiple entities having the same ID? 🤔

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -137,6 +139,14 @@ public void writeEntities(
       @Nonnull PolarisCallContext callCtx,
       @Nonnull List<PolarisBaseEntity> entities,
       List<PolarisBaseEntity> originalEntities) {
+    Set<Long> seenEntityIds = new HashSet<>();
+    for (PolarisBaseEntity entity : entities) {
+      if (!seenEntityIds.add(entity.getId())) {
+        throw new RetryOnConcurrencyException(
```
dimas-b: Why a "retry" exception as opposed to an `IllegalArgumentException`? I believe the behaviour of this method with multiple objects having the same ID is not well-defined and such situations should be avoided on the caller side.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -36,28 +45,128 @@ public String getDisplayName() {
...
 
+  /**
+   * Returns the latest schema version available for this database type. This is used as the default
+   * schema version for new installations.
+   */
+  public int getLatestSchemaVersion() {
+    return switch (this) {
+      case POSTGRES -> 3; // PostgreSQL has schemas v1, v2, v3
+      case COCKROACHDB -> 1; // CockroachDB currently has only schema v1
+      case H2 -> 3; // H2 uses same schemas as PostgreSQL
```
dimas-b: Related to my comment on the schema file: I believe all "latest" JDBC-based schemas should have the same version and evolve in sync. This is because from the caller side (Polaris JDBC persistence code), the "interface" of all JDBC schemas should be the same (column names and type compatibility).

Maintaining different column names and non-trivial type conversions separately for all JDBC backends is going to be too much overhead and will complicate Polaris code unnecessarily.

If the schemas can be assumed to be compatible from the client's (Polaris) perspective, using the same version number will only be natural. WDYT?

dimas-b (commented):
+1 to supporting CockroachDB and thanks for picking this work up, @singhpk234 !

Some comments below.

persistence/relational-jdbc/src/main/resources/cockroachdb/schema-v4.sql
singhpk234: I thought about the same though it make it a bit tricky in sense where we take schema as part of bootstrap, hence thought its better to keep it v1 thought the contents are of v3 sql (as it would just have v3 no v1 / v2). 

I additionally also thought how about we do this 

ALTER COLUMN FROM INT to INT4 to make it work with existing pg version scripts but it seems like we the feature to enable this 
```
Caused by: java.sql.SQLException: Failed due to 'ERROR: ALTER COLUMN TYPE from int to int4 is only supported experimentally

  Hint: See: https://go.crdb.dev/issue-v/49329/v24.3
  ``` 
  
  apparently 
  ```
  SET experimental_enable_alter_column_type_general = true;
ALTER TABLE your_table_name ALTER COLUMN your_column_name TYPE INT4;
```
setting this flag would be confusing. 
Considering we are above i resorted to v1 name, please let me know your thoughts considering above.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
singhpk234: precisely this test got stuck : https://github.com/apache/polaris/blob/main/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java#L1088

i think it more about the implementation of how transaction api  is implemented in Polaris 

https://github.com/apache/polaris/blob/4cf09ec77c77e01fc4b375c635b11a9e474b1281/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L1045

if you see we could collect all the changes for one table and then apply then one shot rather than making then two calls ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
singhpk234: Need to think this through more, i went ahead with the current one with the following in the mind : 
https://github.com/apache/polaris/pull/3352#discussion_r2662489115

persistence/relational-jdbc/src/main/resources/cockroachdb/schema-v4.sql
dimas-b: >  [...] a bit tricky in sense where we take schema as part of bootstrap,

Sorry, @singhpk234  , I do not quite understand this part :thinking: what exactly is "tricky" in case CockroachDB uses `v3` for the schema version?

singhpk234: No worries at all (apologies for not being clear), what i meant is what happens some one does bootstrap with v2 in this case, since we give schema version as arg for bootstrapping 
https://github.com/apache/polaris/blob/4cf09ec77c77e01fc4b375c635b11a9e474b1281/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBootstrapUtils.java#L31
may be we can abstract from the user in the handling ?

secondly can be confusing why there is no v1 / v2 unlike other persistence impl. 

I believe when we add mariadb / mysql support we would have to work with an assumption that the column name are same and types (for example JSONB vs JSON) are handled accordingly so i just thought of starting fresh with v1 for cockroach which is exactly same as v3 of postgres schema.

dimas-b: but what about this? https://github.com/apache/polaris/blob/b1142b56eeeaba61e469a18723a0800e9ca0f748/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelEntity.java#L75-L81

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: Interesting ... :thinking: 

Are you saying that our impl. of `commitTransaction()` will cause multiple `PolarisBaseEntity` objects to be written via the Persistence interface, while some of those objects share the same ID?

singhpk234: precisely :) atleast thats my understanding is, we should group by table-uuid collect all the changes and then call commit for individual tables

persistence/relational-jdbc/src/main/resources/cockroachdb/schema-v4.sql
singhpk234: Nice catch, i missed it, let me think this more

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: TBH, that looks like a bug in the `commitTransaction()` impl. :sweat_smile: 

+1 to grouping changes by table (ID) and writing in one batch.... However, it might be best to make that fix before introducing CockroachDB... just for the sake of clarity.

runtime/admin/src/test/java/org/apache/polaris/admintool/relational/jdbc/CockroachJdbcAdminProfile.java
```diff
@@ -0,0 +1,47 @@
...
+        "polaris.persistence.type",
+        "relational-jdbc",
+        "polaris.persistence.relational.jdbc.database-type",
+        "cockroachdb",
+        // These two options are required to "trigger" the enablement of JDBC data sources in
+        // admin-tool tests, but do not harm other tests.
+        "quarkus.datasource.active",
+        "true",
+        "quarkus.datasource.db-kind",
+        "postgresql");
```
flyrain, line 39: should this be cockroachdb? or `quarkus.datasource.db-kind` simplifies all  postgresql-like ones as postgresql. 
```suggestion
        "cockroachdb");
```

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
singhpk234: Here you go : https://github.com/apache/polaris/pull/3360

dimas-b: Sweet - let's merge #3360 , rebase and re-test whether this particular change is still necessary for CockroachDB, if you do not mind.

singhpk234: rebased this on top of the group of transaction fix, it happily passes now, let me check other fixes for now

flyrain:
@singhpk234, Thanks for adding more tests. There is a conflict and CI failures. Would you mind rebasing it?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -36,28 +45,128 @@ public String getDisplayName() {
     return displayName;
   }
 
+  /**
+   * Returns the latest schema version available for this database type. This is used as the default
+   * schema version for new installations.
+   */
+  public int getLatestSchemaVersion() {
+    return switch (this) {
+      case POSTGRES -> 3; // PostgreSQL has schemas v1, v2, v3
```
dimas-b: Is the latest schema v4 now?

persistence/relational-jdbc/src/main/resources/cockroachdb/schema-v4.sql
dimas-b, line 1: Should we go to v4 right away (to match the latest PostgreSQL schema)? Do we need to support older schemas for CockroachDB?

singhpk234: precisely, though its would be confusing but becuase of v2 handling where we added a column is handling, i was wondering if i can move this to ALTER TABLE add column and then do the same so far the version have been like ok schema version 2 support feature x and version 3 supports feature x + y 
but yes from this pov it makes most sense to start from v4

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -36,28 +45,128 @@ public String getDisplayName() {
...
+      }
+
+      // Couldn't infer and no configured type
+      throw new IllegalStateException(
+          "Cannot infer database type from product name: '"
+              + productName
+              + "'. "
+              + "Please set polaris.persistence.relational.jdbc.database-type explicitly.");
+
+    } catch (java.sql.SQLException e) {
```
dimas-b: nit: import?

runtime/test-common/src/main/java/org/apache/polaris/test/commons/CockroachRelationalJdbcLifeCycleManagement.java
dimas-b: ping :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1044,60 +1045,82 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
         new TransactionWorkspaceMetaStoreManager(diagnostics(), metaStoreManager());
     ((IcebergCatalog) baseCatalog).setMetaStoreManager(transactionMetaStoreManager);
 
+    // Group all changes by table identifier to handle them atomically.
+    // This prevents conflicts when multiple changes target the same table entity.
+    // LinkedHashMap preserves insertion order for deterministic processing.
+    Map<TableIdentifier, List<UpdateTableRequest>> changesByTable = new LinkedHashMap<>();
```
dimas-b, line 1051: hmmm.... is this related to CockroachDB ? IIRC, we have a separate PR for the multi-table commit 🤔

singhpk234:
This is based on https://github.com/apache/polaris/pull/3360 would be really to great this in first, if you all have time i added more tests there as well

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
singhpk234: precisely this is based on multi table commit because of an implementation bug in our transaction

dimas-b: Could we add CockroachDB support first and then rebase #3360 on top of that?

dimas-b: I'm fine with both the multi-table commit fix and CockroachDB, I would just prefer to have those changes in separate PR for the sake of clarity :)

dimas-b: Would it make sense to merge #3360 first?

singhpk234:
> We should probably mention this in the CHANGELOG, WYDT?

Totally agree, creating a pr for it right away !

Thank you so much for review @dimas-b @flyrain 

marked @sathesuraj as the primary author : https://github.com/apache/polaris/commit/8f9ff62a993e6fa990677d26a499676a01593291
thank you so much for your contribution

PR #3207, opened 2025-12-04
NoSQL: Add metastore types and mappings

snazy:
Add the NoSQL specific metastore persistence types including the mapping from and to `*Polaris*Entity`.

PR #3353, opened 2026-01-05
Require approval from 2 reviewers before merging

jbonofre:


singhpk234 (approved):
I think this is unfortunate but this is must required now specially 
after repeated occurence : 
1/ https://github.com/apache/polaris/pull/3207#issuecomment-3628655628
2/https://github.com/apache/polaris/pull/3268#issuecomment-3730739628

jbonofre:
@flyrain no objection from you about this one ? It seems we have already a consensus from several contributors.

flyrain:
To carry on my thoughts on the dev mailing discussion, I don't think enforcing 2 reviewers on PRs solves the problem, meanwhile it can slow down the review process, esp. on trivial PRs. But I will not block it if people want to give it a try.

jbonofre:
@flyrain I got your point. I'm not hard pushing for this one, but more as an attempt to involve more reviewers.

@adutra @MonkeyCanCode @singhpk234 @dimas-b @evindj @adam-christian-software what do you guys think ? Should we try this ?

I'm fine to close this PR if you think it won't help (I'm not very convinced either 😄 ).

snazy:
(Casting my objection on the dev-ML discussion here:)
Not really a fan of having a _hard technical_ requirement of two approvals for every little PR.

jbonofre:
OK, so, we don't have consensus. Let's close this one, we can revisit later if needed.

adutra:
For posterity I'm with @flyrain here: I don't really think 2 reviewers solves the original issue, but I wouldn't oppose the idea either.

PR #3360, opened 2026-01-06
 Fix: Group transaction changes by table

singhpk234:
### Problem

. Groups ALL changes by table- even if they appear randomly in the input like [A1, B1, A2, C1, A3] → groups to {A: 
  [A1, A2, A3], B: [B1], C: [C1]}
  2. For each table, processes changes sequentially :
    - Validate R1 against base metadata → Apply U1 → update currentMetadata
    - Validate R2 against updated metadata → Apply U2 → update currentMetadata
    - Validate R3 against updated metadata → Apply U3 → update currentMetadata
  3. Single commit per table prevents duplicate entity IDs in pendingUpdates

  This ensures:
  - Each change's requirements validate against the evolved state
  -  Updates are applied in order within each table
  -  Only one commit per table (solves the original problem)
  -  Conflicts detected early (e.g., second change expecting schema ID 0 fails when it's already 1)

related discussion: https://github.com/apache/polaris/pull/3352#discussion_r2663070012

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
@@ -1039,7 +1039,7 @@ public abstract class IcebergCatalogHandler extends CatalogHandler implements Au
     // LinkedHashMap preserves insertion order for deterministic processing.
     Map<TableIdentifier, List<UpdateTableRequest>> changesByTable = new LinkedHashMap<>();
     for (UpdateTableRequest change : commitTransactionRequest.tableChanges()) {
-      if (isCreate(change)) {
+      if (CatalogHandlerUtils.isCreate(change)) {
         throw new BadRequestException(
             "Unsupported operation: commitTranaction with updateForStagedCreate: %s", change);
       }


dimas-b:
nit: `Only one commit per table (solves the original problem)` - I'd simply say `Only one commit per table`. This will end up in the git commit log, where the "original problem" might be hard to understand later on... It's also fine to give a summary of the problem instead of the "original" reference if you prefer.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1042,57 +1043,71 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
...
-                        // Apply updates to builder
-                        singleUpdate.applyTo(metadataBuilder);
-                      });
-
-              // Commit into transaction workspace we swapped the baseCatalog to use
-              TableMetadata updatedMetadata = metadataBuilder.build();
-              if (!updatedMetadata.changes().isEmpty()) {
-                tableOps.commit(currentMetadata, updatedMetadata);
+    // Process each table's changes in order
+    changesByTable.forEach(
```
dimas-b, line 1058: The new code logic looks correct to me. I think it's a worthy change to merge in its own right.

However, as for the issue discussed in https://github.com/apache/polaris/pull/3352#discussion_r2663070012 , I think this fix is effective, but it's not apparent that it will work correctly.

The basic problem is that Persistence is called with multiple entity objects for the same ID. This means that the `updateEntitiesPropertiesIfNotChanged` call on line must contain at most one entity per ID. This may or may not be true depending on what the catalog forwards to `transactionMetaStoreManager` for each commit on line 1108.

I do believe that one `tableOps.commit()` will result in one entity update, so it may be sufficient to just add a comment about that around line 1113. WDYT?

dimas-b (dismissed):
Nice fix. Thanks, @singhpk234 ! 

Just a couple of minor comments :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1042,57 +1043,71 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
...
+    // Group all changes by table identifier to handle them atomically
+    // This prevents conflicts when multiple changes target the same table entity
+    // LinkedHashMap preserves insertion order for deterministic processing
+    Map<TableIdentifier, List<UpdateTableRequest>> changesByTable = new LinkedHashMap<>();
+    for (UpdateTableRequest change : commitTransactionRequest.tableChanges()) {
+      if (isCreate(change)) {
+        throw new BadRequestException(
+            "Unsupported operation: commitTranaction with updateForStagedCreate: %s", change);
+      }
+      changesByTable.computeIfAbsent(change.identifier(), k -> new ArrayList<>()).add(change);
```
flyrain, line 1046: Should we check and throw in case the table exists already? It'd also be nice to have a test case.

dimas-b: Does the Iceberg REST Catalog spec allow more than one table update in one `commitTransaction` operation?

flyrain: The spec doesn't mention the uniqueness of the table identifier, https://github.com/apache/polaris/blob/70ad92f32afd9f0ddb9a32343553acc5d586362c/spec/iceberg-rest-catalog-open-api.yaml#L3499. 

Within one `CommitTableRequest`, multiple `updates` are allowed, https://github.com/apache/polaris/blob/70ad92f32afd9f0ddb9a32343553acc5d586362c/spec/iceberg-rest-catalog-open-api.yaml#L3478, which should apply to the same snapshot. 

With that, I think it's undefined behavior whether the same table identifier could duplicate. We could explicitly disable it in Polaris though.

dimas-b: I'd rather not disable this in Polaris unless the spec explicitly flagged it as a disallowed use case (which it did not). 

In lieu of an explicit spec, applying multiple updates in sequence is a fairly straight-forward operation since each update is self-contained and will be validated against the current metadata.

flyrain: SGTM.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dennishuo: Yeah, logic also looks correct to me. +1 to adding a comment on the subtlety though that we're coalescing all the updates for a given table into a single Polaris entity update, which is a slightly different behavior than if the caller expected the various `UpdateTableRequests` in this `commitTransaction` to really behave as if they were each applied independently (but as if under a lock).

Note that this issue was a known limitation, and referenced in a TODO in `TransactionWorkspaceMetaStoreManager`:

https://github.com/apache/polaris/blob/8abf19a06f809c8d1d0764b44e613c2614ce2c67/polaris-core/src/main/java/org/apache/polaris/core/persistence/TransactionWorkspaceMetaStoreManager.java#L84

    // TODO: If we want to support the semantic of opening a transaction in which multiple
    // reads and writes occur on the same entities, where the reads are expected to see the writes
    // within the transaction workspace that haven't actually been committed, we can augment this
    // class by allowing these pendingUpdates to represent the latest state of the entity if we
    // also increment entityVersion. We'd need to store both a "latest view" of all updated entities
    // to serve reads within the same transaction while also storing the ordered list of
    // pendingUpdates that ultimately need to be applied in order within the real MetaStoreManager.

The alternative "fix" described there that is  more general but more complex and probably has pitfalls is to really queue up the sequential mutations per entity in that "uncommitted persistence layer".

The main implications would be that if we plug into the MetaStoreManager layer, we can intercept but inherit other relevant hooks, such as generating events, having entityVersion increments directly match actual update requests, etc.

But I'm in favor of this more targeted change-coalescing fix here for now. We could either update/remove the TODO in `TransactionWorkspaceMetaStoreManager` and/or leave a comment in the code here referencing the other approach so any future changes to the way we handle these can more easily sort through it.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1042,57 +1043,71 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
...
+                        FeatureConfiguration.ALLOW_NAMESPACE_LOCATION_OVERLAP)) {
+                  throw new BadRequestException(
+                      "Unsupported operation: commitTransaction containing SetLocation"
+                          + " for table '%s' and new location '%s'",
+                      change.identifier(), ((MetadataUpdate.SetLocation) singleUpdate).location());
+                }
               }
-            });
+
+              // Apply updates to builder
```
dennishuo, line 1098: Maybe add a `// TODO` to refactor this to better share/reconcile with the [update-application logic in CatalogHandlerUtils](https://github.com/apache/polaris/blob/8abf19a06f809c8d1d0764b44e613c2614ce2c67/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java#L477) (I understand this divergence was already latent and not introduced by this PR, but as the update-application logic grows in complexity here it's going to start getting a lot worse).

dennishuo (approved):

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
singhpk234: Ack, addressed this feedback !

singhpk234:
Apologies it took me a while to get back to it, was super swamped with internal stuff.
Addressed @dennishuo feedback to add detailed comment.
Addressed @flyrain's offline feedback to add more tests to iron this.

Please have another pass when you all get some time, really apprecaite your feedbacks here !

dimas-b:
@singhpk234 : thanks for pushing this forward :+1: Please fix fresh `IcebergCatalogHandler` conflicts :sweat_smile:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1033,60 +1034,82 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
...
+            TableMetadata.Builder metadataBuilder = TableMetadata.buildFrom(currentMetadata);
+            for (MetadataUpdate singleUpdate : change.updates()) {
+              // Note: If location-overlap checking is refactored to be atomic, we could
+              // support validation within a single multi-table transaction as well, but
+              // will need to update the TransactionWorkspaceMetaStoreManager to better
+              // expose the concept of being able to read uncommitted updates.
+              if (singleUpdate instanceof MetadataUpdate.SetLocation setLocation) {
+                if (!currentMetadata.location().equals(setLocation.location())
+                    && !realmConfig()
+                        .getConfig(FeatureConfiguration.ALLOW_NAMESPACE_LOCATION_OVERLAP)) {
```
flyrain, line 1090: This check allows the operation `setLocation` when ALLOW_NAMESPACE_LOCATION_OVERLAP is true.  I think the behavior isn't correct.  cc @dennishuo @collado-mike 
However, it isn't a blocker for me as this PR doesn't the logic.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
```diff
@@ -1189,6 +1189,101 @@ public void testMultipleConflictingCommitsToSingleTableInTransaction() {
...
+    // that the server correctly handles multiple update types within a single commit request.
+    Table table = catalog().buildTable(identifier, SCHEMA).create();
+    Transaction transaction = table.newTransaction();
+
+    UpdateSchema updateSchema =
+        transaction.updateSchema().addColumn("new_col", Types.LongType.get());
+    Schema expectedSchema = updateSchema.apply();
+    updateSchema.commit();
+
+    transaction.updateProperties().set("prop-key", "prop-val").commit();
```
flyrain, line 1271: Is behavior deterministic when we have two property updates like this? 
1.  transaction.updateProperties().set("prop-key", "prop-val1").commit();
2. transaction.updateProperties().set("prop-key", "prop-val2").commit();

If not, we may check with the Iceberg community to clarify the behavior. Not a blocker.

singhpk234: within a transaction (taking iceberg's client side transcation as an example) it will be consistent as if prop-key value would be prop-val2 as we create a new metadata locally and keep on apply update on top of last updated metadata, is my understanding.

flyrain: Are updateRequests ordered? If not, the final result could be either val2 or val1.

singhpk234: yes they are (atleast in java impl) I believe spec has an implicit assumption on this, we can iron the spec more bcz otherwise it will not be a true transaction if they are not ordered, is my take.

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L1320

The serializer and deserializer both respect the order :
https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/requests/CommitTransactionRequestParser.java#L42

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/requests/CommitTransactionRequestParser.java#L62

an E2E tests for this 
https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/requests/TestCommitTransactionRequestParser.java#L162

plz let me know how feel about it

singhpk234:
Thanks everyone for the review !

PR #3381, opened 2026-01-08
Setup gradle using Github action to avoid downloading from services.gradle.org

pingtimeout:
During workflow development, Github runners may be temporarily banned from services.gradle.org.  This PR ensures that Gradle and the Gradle wrapper are set up using the proper Github Action and therefore from Github mirrors so that the issue does not happen anymore.

Example of issue: https://github.com/pingtimeout/polaris/actions/runs/20488431923/job/58875553318

```
Error: Exception in thread "main" java.io.IOException: Downloading from https://services.gradle.org/distributions/gradle-9.2.1-bin.zip failed: timeout (10000ms)
```

### DIFF
--- a/.github/workflows/release-2-update-release-candidate.yml
+++ b/.github/workflows/release-2-update-release-candidate.yml
@@ -145,6 +145,9 @@ jobs:
           distribution: 'temurin'
           java-version: '21'
 
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5
+
       - name: Update project versions
         if: env.rc_number == '0'
         run: |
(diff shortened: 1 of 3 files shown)


snazy:
For posterity, Gradle's `setup-gradle` action provides functionality to cache Gradle distributions, see the [cache artifacts](https://github.com/apache/polaris/actions/caches) named `gradle-home-*`). The full name of such Gradle home cache artifacts also contains the strings `java-21` and `polaris-gradle[gradle]`. Those two pieces are configured via the `ci-incr-build-cache-prepare` action in the apache/polaris repo, via the environment variables `GRADLE_BUILD_ACTION_CACHE_KEY_ENVIRONMENT`, `GRADLE_BUILD_ACTION_CACHE_KEY_JOB` and `GRADLE_BUILD_ACTION_CACHE_KEY_JOB_INSTANCE`. Without setting those env vars, the release jobs would _not_ be able to reuse a cached Gradle-Home, so setting those is needed. Note that the total size for GitHub cache artifacts is limited to 10GB (LRU eviction policy).

The easiest option would be to just use `ci-incr-build-cache-prepare`.

Another thing I'd recommend is to make the cache read-only for releases (the default when using `ci-incr-build-cache-prepare`), to ensure that there is no way to leak sensitive data into cache artifacts.

PR #3386, opened 2026-01-08
Move idempotency_records schema to v4 and add H2 support

huaxingao:
Moves the idempotency_records table definition and indexes from schema-v3 to schema-v4 for Postgres, and adds the matching idempotency_records table and realm‑scoped expiry index to H2 schema-v4. This keeps both JDBC backends aligned on the v4 schema for idempotency persistence

## Checklist

### DIFF
--- a/persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
+++ b/persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
@@ -140,7 +140,7 @@ CREATE TABLE IF NOT EXISTS idempotency_records (
     realm_id TEXT NOT NULL,
     idempotency_key TEXT NOT NULL,
     operation_type TEXT NOT NULL,
-    resource_id TEXT NOT NULL,
+    resource_id TEXT NOT NULL, -- normalized request-derived resource identifier (not a generated entity id)
 
     -- Finalization/replay
     http_status INTEGER,       -- NULL while IN_PROGRESS; set only on finalized 2xx/terminal 4xx
--- a/persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
+++ b/persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
@@ -141,7 +141,7 @@ CREATE TABLE IF NOT EXISTS idempotency_records (
     realm_id TEXT NOT NULL,
     idempotency_key TEXT NOT NULL,
     operation_type TEXT NOT NULL,
-    resource_id TEXT NOT NULL,
+    resource_id TEXT NOT NULL,            -- normalized request-derived resource identifier (not a generated entity id)
 
     -- Finalization/replay
     http_status INTEGER,                 -- NULL while IN_PROGRESS; set only on finalized 2xx/terminal 4xx


persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
```diff
@@ -0,0 +1,162 @@
...
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+-- Changes from v2:
+--  * Added `events` table
```
singhpk234, line 21: we need to add idempotency key addition here in the comments.

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
```diff
@@ -0,0 +1,162 @@
...
+
+CREATE SCHEMA IF NOT EXISTS POLARIS_SCHEMA;
```
singhpk234, line 24: can we add test for v4 for the metastore manager.

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
huaxingao: Added. Thanks!

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
huaxingao: Test was added. Thanks!

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
```diff
@@ -0,0 +1,165 @@
...
+    additional_properties JSONB NOT NULL DEFAULT '{}'::JSONB,
+    PRIMARY KEY (event_id)
+);
+
+-- Idempotency records (key-only idempotency; durable replay)
+CREATE TABLE IF NOT EXISTS idempotency_records (
+    realm_id TEXT NOT NULL,
+    idempotency_key TEXT NOT NULL,
+    operation_type TEXT NOT NULL,
+    resource_id TEXT NOT NULL,
```
flyrain: Should we allow resource id nullable? For example, if we are using table entity id as resource id, the entity id will be generated only after the table creation succeeded.  When table creation failed(return 4xx), there may be no entity id generated.

huaxingao: Thanks for the comment! You’re right that if resource_id were a generated entity id, it could be null on 4xx failures (no entity created). In this implementation, resource_id is intentionally request-derived and stable (a normalized identifier like `tables/<ns>.<name>`), not an internal entity id. That means it’s always available even on failures. Keeping it NOT NULL ensures the idempotency key is always tied to one specific resource.

I added a comment to `resource_id TEXT NOT NULL` to make it clear
```
resource_id TEXT NOT NULL,            -- normalized request-derived resource identifier (not a generated entity id)
```

flyrain: That makes sense. I think the connection tables like `grant_records` and `policy_mapping_record` will need resource id to cover two normalized identifiers, e.g., a table identifier and a policy identifier, correct?

singhpk234:
Thanks @huaxingao for the change and @flyrain for the review

huaxingao:
Thanks @singhpk234 @flyrain

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
```diff
@@ -0,0 +1,163 @@
...
+SET SCHEMA POLARIS_SCHEMA;
+
+CREATE TABLE IF NOT EXISTS version (
+    version_key VARCHAR PRIMARY KEY,
+    version_value INTEGER NOT NULL
+);
+
+MERGE INTO version (version_key, version_value)
+    KEY (version_key)
+    VALUES ('version', 3);
```
obelix74, line 34: Should this version be set to `4`?  @singhpk234 @dimas-b

huaxingao: @obelix74 Thanks for catching this problem. I will fix this in next PR.

PR #3391, opened 2026-01-09
[DO NOT MERGE] POC for Using Intermediate Data as part of Events

adnanhemani:
This is a POC that shows an approach for #3209. Do not merge this code, we will likely need a pretty large refactor to use the new code to be introduced as part of #3293.

## Checklist
- [ ] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [ ] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [ ] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

Issue #3209, opened 2025-12-05
Implement sending POJOs not within the Response to Events


adnanhemani:
### Is your feature request related to a problem? Please describe.

Not a problem, but it would be nice to be able to add more information to events which are not currently returned to users. For example, in #3195, we are unable to send `LoadTableResponse`s in the `AfterTableUpdateEvent` (which make sense for table updates to tell what is the current state of the table) in the case of a CommitTransaction. Meanwhile, if a user calls the UpdateTable API directly, they will still get the `LoadTableResponse` as part of the `AfterTableUpdateEvent`.

This is solely due to the way we have implemented events through the delegator model where intermediate information, which is not surfaced to the user as part of the API response, cannot be used in the event emission. There are other potential uses cases, such as telling whether a credential generated was a read or read+write credential when loading a table - which would also be interesting for auditing purposes.

### Describe the solution you'd like

A clear mechanism is implemented where intermediary information can be resurfaced to the Events emission.

### Describe alternatives you've considered

1. [Not recommended] Overstuff the response body with all relevant information and then filter out only the IRC spec-required information before returning back to the user in the Delegator.
2. Using Container Request Context to store intermediate information during the call processing, then retrieve the information whenever event emission is happening.
    a. Unsure if this is wise, since most tests are not running with Container Request Contexts - and I'm unsure if inserting Quarkus mechanism within Iceberg business logic is a great idea. But maybe something to think about...

rishii-19-works:
Hi @adnanhemani  thanks for opening this issue. The use-case really makes sense
I’m interested in contributing here, so I wanted to clarify a few things before I start working on anything

From what I understood, the main idea is to make some intermediate data (like `LoadTableResponse` during a transaction commit) available to the event emitter even when it’s not actually part of the API response. I just had a couple of questions so I move in the right direction:-

1. Would you prefer using some kind of per-request context to store this metadata, or is there already any internal pattern in Polaris to pass such info between layers?
2. Are there any architectural constraints regarding Quarkus internals vs Polaris/Iceberg logic that I should keep in mind?
3. For the first PR, should I only focus on adding the mechanism for storing/passing this intermediate data, or also update a specific flow to show how it would work?

Once I get your guidance on the approach I can start drafting something or work on an initial implementation

Thanks!

rishii-19-works:
Hi @adnanhemani, I haven’t heard back yet, so I’ll go ahead and start exploring this on my end
[...]

PR #3392, opened 2026-01-09
Support intermediate request-scoped data in catalog generic table events

rishii-19-works:
This PR adds support for attaching and propagating intermediate, request-scoped data across catalog generic table events.

Following the approach proposed in #3391, the same event metadata instance is now reused between the corresponding before and after generic table events. This allows intermediate objects produced during request processing (for example, response payloads) to be shared with downstream event listeners without coupling listeners to API responses or re-deriving state.

A focused test is included to verify that intermediate data attached during the before event is correctly available in the corresponding after event.

This change addresses #3209 and keeps the scope intentionally minimal, limited to generic table load events, to establish a clear and extensible pattern for future event types.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
@@ -109,39 +109,45 @@ public class CatalogGenericTableEventServiceDelegator
   }
 
   @Override
-  public Response loadGenericTable(
-      String prefix,
-      String namespace,
-      String genericTable,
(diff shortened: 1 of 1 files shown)


rishii-19-works:
Hi, thanks for the earlier guidance.

I’ve followed the approach suggested in #3391 to propagate intermediate request-scoped data across before/after generic table events.

I’m currently running into CI compilation failures around how intermediate data should be attached (event wrapper vs metadata vs underlying PolarisEvent). I’ve tried a few variants, but it looks like I’m misunderstanding the intended API surface.

Could you please advise which object is expected to own intermediate data in this flow (the service event wrapper, the metadata, or the underlying PolarisEvent), and how it should be propagated to the after-event?

I’ll update the implementation accordingly. Thanks for your patience.

PR #3393, opened 2026-01-09
Release workflows should retry svn checkout in case of failure

pingtimeout:
During workflow development, I ran into issues like this one during Helm charts build job:

```
svn: E170013: Unable to connect to a repository at URL 'https://dist.apache.org/repos/dist/dev/incubator/polaris'
svn: E000110: Error running context: Connection timed out
Error: Process completed with exit code 1.
```

Example: https://github.com/pingtimeout/polaris/actions/runs/20792183122/job/59719038791

This is not expected as there is another job in the same workflow that executes the same `svn checkout` command and succeeds.  I suspect a rate limit on ASF Infra side.  This PR adds a sleep-then-retry mechanism to workflows so that they do not fail immediately.

Here is an example of a successful run.  Unfortunately, that time, the rate limit did not trigger, so it only shows that the workflow continue running identically with the code change: https://github.com/pingtimeout/polaris/actions/runs/20825202531/job/59827207724

### DIFF
--- a/.github/workflows/release-3-build-and-publish-artifacts.yml
+++ b/.github/workflows/release-3-build-and-publish-artifacts.yml
@@ -183,22 +183,7 @@ jobs:
           dist_dev_dir=${RELEASEY_DIR}/polaris-dist-dev
 
           # Retry logic for SVN checkout (Apache SVN can have transient connectivity issues)
-          max_attempts=5
-          attempt=1
-          while true; do
-            if exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"; then
-              break
-            fi
-            if [[ $attempt -ge $max_attempts ]]; then
-              echo "ERROR: SVN checkout failed after ${max_attempts} attempts"
-              exit 1
-            fi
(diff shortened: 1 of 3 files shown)


.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -181,7 +181,24 @@ jobs:
           source "${LIBS_DIR}/_exec.sh"
 
           dist_dev_dir=${RELEASEY_DIR}/polaris-dist-dev
-          exec_process svn checkout --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive "${APACHE_DIST_URL}${APACHE_DIST_PATH}" "${dist_dev_dir}"
+
+          # Retry logic for SVN checkout (Apache SVN can have transient connectivity issues)
```
snazy, line 185: Mind moving the retry-loop to a new function like `exec_process_with_retries` or so?

pingtimeout: Sure.  Implementation part is done, I just need to re-test.

pingtimeout: Confirmation that it still works : https://github.com/pingtimeout/polaris/actions/runs/20915713316/job/60088667452

Note that the workflow failed because of an unrelated issue: I am using a temporary GPG key that is not published to ubuntu keyserver.  So the gradle task that closes the staging repository fails and that in turn fails the release process.

PR #3395, opened 2026-01-09
NoSQL: Add maintenance service to Polaris admin tool

snazy:
This change wires the maintenance service up to the Polaris admin tool.

The new commands are grouped under the `nosql` parent command. This should provide enough isolation for end users who do not use NoSQL Persistence.

The initial NoSQL maintenance replies on CEL expressions for configuration. This will be reworked in follow-up PRs under #3847.

The Admin Tool remains a collection of utilities for low-level manipulation of the Polaris database, such as bootstrapping and maintenance. It targets admin users, so ordinary API / Catalog users are not expected to be affected by introducing NoSQL commands into the Admin CLI. Administrators are expected to be familiar with the multitude of Persistence backends and should not be confused by the new NoSQL-specific commands. 

Another area for follow-up (related to modularizing the Admin CLI for better dependency management) is covered by #3855.

### DIFF
--- a/runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
+++ b/runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
@@ -33,7 +33,7 @@ import picocli.CommandLine;
       NoSqlMaintenanceRunCommand.class,
     },
     mixinStandardHelpOptions = true,
-    description = "Polaris NoSQL persistence.")
+    description = "Sub-commands specific to NoSQL persistence.")
 public class NoSqlCommand extends BaseNoSqlCommand {
   @Inject protected Backend backend;
 
--- a/site/content/in-dev/unreleased/admin-tool.md
+++ b/site/content/in-dev/unreleased/admin-tool.md
@@ -62,7 +62,7 @@ Commands:
   help       Display help information about the specified command.
   bootstrap  Bootstraps realms and root principal credentials.
   purge      Purge realms and all associated entities.
-  nosql      Polaris NoSQL persistence.
+  nosql      Sub-commands specific to NoSQL persistence.
 ```
 
 ## Configuration
@@ -215,7 +215,7 @@ Again, the Polaris Admin Tool must be executed with appropriate configuration to
 The `nosql` admin tool command is used to perform NoSQL metastore specific operations.
 ```
 Usage: polaris-admin-tool.jar nosql [-hV] [COMMAND]
-Polaris NoSQL persistence.
+Sub-commands specific to NoSQL persistence.
   -h, --help      Show this help message and exit.
   -V, --version   Print version information and exit.
 Commands:


runtime/admin/src/main/java/org/apache/polaris/admintool/maintenance/NoSqlMaintenanceCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "maintenance",
+    subcommands = {
+      NoSqlMaintenanceLogCommand.class,
+      NoSqlMaintenanceRunCommand.class,
+    },
+    mixinStandardHelpOptions = true,
+    description = "Polaris persistence maintenance.")
```
dimas-b: Should we flag it as "beta" as discussed in the last community sync call?.. or perhaps "preview" since the actual persistent storage that can be maintained via this command has not been hooked up yet (IIRC).

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/maintenance/NoSqlMaintenanceRunCommand.java
```diff
@@ -0,0 +1,57 @@
...
+package org.apache.polaris.admintool.maintenance;
+
+import jakarta.inject.Inject;
+import org.apache.polaris.persistence.nosql.maintenance.api.MaintenanceService;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "run",
+    mixinStandardHelpOptions = true,
+    description = {"Run Polaris persistence maintenance."})
```
dimas-b, line 28: maybe `Run Polaris NoSQL persistence maintenance` for the sake of clarity?

runtime/admin/src/main/resources/application.properties
```diff
@@ -40,7 +40,24 @@ quarkus.rds.sync-client.type=apache
...
+# - relational-jdbc - JdbcMetaStoreManagerFactory
+# - nosql - NoSQL persistence backend, define the backend type via 'polaris.persistence.nosql.backend'
 polaris.persistence.type=relational-jdbc
+# Database backend for 'nosql' persistence-type
+# Available backends:
+# - InMemory - for testing purposes
+# - MongoDb - configure the via the Quarkus extension
+polaris.persistence.nosql.backend=InMemory
+
+## MongoDB version store specific configuration
```
dimas-b, line 58: `MongoDB version store specific configuration defaults`? I suppose the user can change any of these settings.

runtime/admin/src/main/resources/application.properties
```diff
@@ -40,7 +40,24 @@ quarkus.rds.sync-client.type=apache
...
+quarkus.mongodb.database=polaris
+quarkus.mongodb.metrics.enabled=true
+#quarkus.mongodb.connection-string=mongodb://localhost:27017
+quarkus.mongodb.devservices.enabled=false
```
dimas-b: Is it a runtime setting or build-time (the latter have a dedicated section)?

runtime/admin/src/main/resources/application.properties
```diff
@@ -50,10 +67,13 @@ quarkus.datasource.devservices.enabled=false
 
 quarkus.arc.ignored-split-packages=\
   org.apache.polaris.admintool.config,\
+  org.apache.polaris.admintool.maintenance,\
```
dimas-b: Could we refactor this code to avoid split packages upfront?

dimas-b (commented):
LGTM 👍 just a few minor comments 🙂

flyrain (dismissed):
I do not think the NoSQL maintenance tool belongs in the runtime admin module. NoSQL is just one persistence impl., and this maintenance logic is only applicable to that specific backend. Other persistence implementations do not need it, yet this change pulls NoSQL specific behavior into a shared admin layer. This creates unnecessary coupling, complicates the module boundaries, and makes the admin tooling harder to reason about and extend going forward.

cc @dennishuo @collado-mike @singhpk234 @jbonofre @HonahX @RussellSpitzer

snazy:
The admin-tool is meant for all persistence implementations, because it is a tool to perform _administrative_ tasks. We have discussed before that other persistence specific operations like schema setup for JDBC would go into the admin-tool.
Adding another tool is rather going to complicate user experience.

dimas-b:
@flyrain : 
> I do not think the NoSQL maintenance tool belongs in the runtime admin module. [...]

NoSQL persistence (by design) needs periodic maintenance by admin users. Not providing a tool for this will mean that no Apache Polaris user will be able to use NoSQL persistence efficiently.

As to the location of the tool, Admin CLI is a natural place as it hosts code for `bootstrap` and other administrative / setup tasks.

I believe this is the minimum the project community can do for OSS users.

dimas-b:
Related `dev` discussion: https://lists.apache.org/thread/znw87l7r1gh3l09pc0hxhj5gyl8vp28d

runtime/admin/src/main/resources/application.properties
RussellSpitzer: Same comments request here as on the service pr, These properties should probably be commented out by default since they aren't used

runtime/admin/src/main/java/org/apache/polaris/admintool/maintenance/NoSqlMaintenanceCommand.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  public Integer call() {
+    var out = spec.commandLine().getOut();
+
+    out.println("Polaris NoSql persistence maintenance has multiple subcommands,");
+    out.println("use the 'help maintenance' command.");
+    out.println();
+
+    checkInMemory();
```
RussellSpitzer: Shouldn't we be failing for everything that isn't NoSql not just InMemory?

runtime/admin/src/main/java/org/apache/polaris/admintool/maintenance/NoSqlMaintenanceCommand.java
RussellSpitzer: I think we also probably shouldn't call it persistence maintenance unless we plan on having maintenance for other backends?

runtime/admin/src/main/java/org/apache/polaris/admintool/maintenance/NoSqlMaintenanceCommand.java
```diff
@@ -0,0 +1,52 @@
...
+
+    out.println();
+    out.println("Information: selected NoSql persistence backend: " + backend.type());
```
RussellSpitzer: In the current formulation this isn't correct since you can call this with any backend right?

runtime/admin/src/main/java/org/apache/polaris/admintool/maintenance/BaseMaintenanceCommand.java
```diff
@@ -0,0 +1,190 @@
...
+import org.apache.polaris.persistence.nosql.maintenance.api.MaintenanceService;
+
+@SuppressWarnings("CdiInjectionPointsInspection")
+public abstract class BaseMaintenanceCommand extends BaseCommand {
+  @Inject protected MaintenanceService maintenanceService;
+  @Inject protected Backend backend;
+  @Inject protected MaintenanceConfig maintenanceConfig;
+
+  protected void checkInMemory() {
+    if ("InMemory".equals(backend.type())) {
```
RussellSpitzer: Do we need a check here for the persistence not being NoSQL?

RussellSpitzer:
As discussed on the mailing list, I think this is fine to add into the Admin tool but the current PR needs a bit of work to make it clear to end users that it is just for NoSQL.

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
```diff
@@ -0,0 +1,46 @@
...
+
+import jakarta.inject.Inject;
+import org.apache.polaris.admintool.nosql.maintenance.NoSqlMaintenanceInfoCommand;
+import org.apache.polaris.admintool.nosql.maintenance.NoSqlMaintenanceLogCommand;
+import org.apache.polaris.admintool.nosql.maintenance.NoSqlMaintenanceRunCommand;
+import org.apache.polaris.persistence.nosql.api.backend.Backend;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "nosql",
```
RussellSpitzer, line 29: +1

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/maintenance/NoSqlMaintenanceRunCommand.java
dimas-b: @snazy : WDYT?

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
```diff
@@ -0,0 +1,46 @@
...
+    subcommands = {
+      NoSqlMaintenanceInfoCommand.class,
+      NoSqlMaintenanceLogCommand.class,
+      NoSqlMaintenanceRunCommand.class,
+    },
+    mixinStandardHelpOptions = true,
+    description = "Polaris NoSQL persistence.")
```
dimas-b: nit: suggestion: `Sub-commands specific to NoSQL persistence.`

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/maintenance/NoSqlMaintenanceInfoCommand.java
```diff
@@ -0,0 +1,39 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.admintool.nosql.maintenance;
+
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "maintenance-info",
```
dimas-b, line 24: nit: Maybe add `maintenance` as another sub-command group?

Current: `java -jar ... nosql maintenance-info`
Proposed: `java -jar ... nosql maintenance info`

I hope the latter in nicer in terms of UX.

runtime/admin/src/main/resources/application.properties
dimas-b: Per https://quarkus.io/guides/mongodb-dev-services `quarkus.mongodb.devservices.enabled` is a build-time config. Please move to the `Build-time configuration` section (lines 20-38) for the sake of consistency.

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/maintenance/NoSqlMaintenanceRunCommand.java
dimas-b: ... and similar description changes in the other new commands?

runtime/admin/src/main/resources/application.properties
dimas-b: Note: this package no longer exists.

dimas-b: Package `org.apache.polaris.admintool.nosql.maintenance` only exists in one module, that is in `runtime/admin`... Is this config required anymore? 🤔

runtime/admin/src/main/resources/application.properties
```diff
@@ -50,10 +69,13 @@ quarkus.datasource.devservices.enabled=false
...
   org.apache.polaris.admintool
 
 ## Quarkus required setting for third party indexing
 # fixed at build-time
+quarkus.index-dependency.agrona.group-id=org.agrona
+quarkus.index-dependency.agrona.artifact-id=agrona
```
dimas-b: I tried building locally without these settings (with the [workaround](https://github.com/apache/polaris/issues/3467#issuecomment-3817989427) to get build warnings), but `agrona` was not flagged as needing an index... Is this setting required for the admin tool?

dimas-b:
Summarizing the [dev discussion](https://lists.apache.org/thread/znw87l7r1gh3l09pc0hxhj5gyl8vp28d).

- It looks like there is general consensus on adding NoSQL maintenance commands to the Admin Tool, scoped under a `nosql` top-level command group.

- CEL appears to be regarded as excessive in terms of the flexibility and code maintenance it brings for the use cases it enables. I propose merging this PR as is (since it does not touch CEL directly) and converting NoSQL maintenance rules from CEL to java code + config in a follow up PR. I'll open a GH issue for this if other people of ok with this approach.

- Further refactorings / renames are certainly possible and I invite ideas / proposals to be expressed on the `dev` thread (link above) or as GH issues.

dimas-b:
@flyrain : Would you please reconsider your -1 on this PR given the above discussion?

dimas-b:
@RussellSpitzer : just FYI: I re-requested your review my accident :sweat_smile: but fresh comments are welcome, of course :slightly_smiling_face:

dimas-b:
@snazy : could you rebase this PR? I suspect some of the config concerns are related to the other NoSQL modules merged to `main` but not preset under this PR's git history :thinking:

snazy:
Rebased + force-pushed

runtime/admin/src/main/resources/application.properties
dimas-b: I re-ran my local build after the recent rebase _without_ this property and I did not see any build warnings about Agrona being unindexed.

@snazy : WDYT about removing this property?

runtime/admin/src/main/resources/application.properties
dimas-b: Actually, I see different split package warnings in my build:

```
2026-02-03T00:51:08.714249614Z build-15 WARN Detected a split package usage which is considered a bad practice and should be avoided. Following packages were detected in multiple archives: 
- "org.apache.polaris.service.catalog.api" found in [org.apache.polaris:polaris-api-catalog-service:1.4.0-incubating-SNAPSHOT, org.apache.polaris:polaris-api-iceberg-service:1.4.0-incubating-SNAPSHOT]
- "org.apache.polaris.service.catalog.api.impl" found in [org.apache.polaris:polaris-api-catalog-service:1.4.0-incubating-SNAPSHOT, org.apache.polaris:polaris-api-iceberg-service:1.4.0-incubating-SNAPSHOT]
- "org.apache.polaris.service.types" found in [org.apache.polaris:polaris-api-catalog-service:1.4.0-incubating-SNAPSHOT, org.apache.polaris:polaris-api-iceberg-service:1.4.0-incubating-SNAPSHOT]
```

runtime/admin/src/main/resources/application.properties
```diff
@@ -50,6 +69,7 @@ quarkus.datasource.devservices.enabled=false
 
 quarkus.arc.ignored-split-packages=\
   org.apache.polaris.admintool.config,\
+  org.apache.polaris.admintool.maintenance,\
```
dimas-b:
```suggestion
```

runtime/admin/src/main/resources/application.properties
dimas-b: I've added a commit to remove this prop.

dimas-b: I'll deal with the other warnings on `main` later.

dimas-b:
@flyrain : Could you review again, please? There were code updates and fresh discussion points after your previous -1 review.

flyrain:
Sorry for the late reply, I have posted the concern in the dev ML https://lists.apache.org/thread/93nvjwwbvv77fz16rsdcwsm1m8nc5trt

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
dennishuo: +1 to this rewording

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
```diff
@@ -0,0 +1,46 @@
...
+public class NoSqlCommand extends BaseNoSqlCommand {
```
dennishuo, line 37: Just a question for my own understanding, since I'm not too familiar with how the command hierarchy works - how does this "sibling" command end up causing the nesting discussed on the dev list? It looks like the hierarchy is:

    NoSqlMaintenanceRunCommand extends BaseNoSqlMaintenanceCommand
    BaseNoSqlMaintenanceCommand extends BaseNoSqlCommand
    BaseNoSqlCommand extends BaseCommand

While 

    NoSqlCommand extends BaseNoSqlCommand

Thus appears to be a "sibling" of `BaseNoSqlMaintenanceCommand ` rather than being a "parent".

Some documentation on the expected full example commands would also be helpful to reason about this; could we add some to the PR description and/or readme/md docs?

dennishuo (commented):
(Belated) Thanks @dimas-b for the summary of the dev discussion in https://github.com/apache/polaris/pull/3395#issuecomment-3826916806 !

Based on that summary, I'm happy with the core content of the code in this PR to merge-and-iterate. To facilitate that, just a few things would be nice before merging to keep the most useful information intact for posterity:

1. Can we update the PR description with the main takeaways from the dev discussion and also link to the dev discussion for posterity so that future folks who track down this PR when figuring things out can easily see the main reference materials without sifting through the really long PR comment threads?
2. Can we add user-facing documentation along the same vein as https://github.com/apache/polaris/pull/3818/changes to actually see at a glance what the admin tool invocation will look like with the full nested subcommand and what it looks like to pass in arguments regarding retention time, etc? Normally I'd also expect the sample commands in the PR description of the PR that adds the new commands.
3. If the example commands include usage of CEL, to make a note in that documentation that the "overly" expressive CEL syntax might be re-evaluated in the near future so the user shouldn't depend on it if possible (or else they should also join the dev discussion about the CEL functionality).
4. Add a link to the aforementioned GH issue referred to in https://github.com/apache/polaris/pull/3395#issuecomment-3826916806 into the PR description or doc comments that explain the uncertain fate of CEL for easy navigation

Some of this request for docs follows from my comment in https://github.com/apache/polaris/pull/3268#pullrequestreview-3581419658 - I understand the response there was to add docs "later", but since this PR is the one that actually establishes the user interaction model in terms of admin-tool commands, it's nice to pair the addition of the command with instructions for how it works.

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
dennishuo: Reading through the code to answer my own question :)

I guess the `subcommands` tuple basically "intercepts" the other command classes rather than requiring those other command classes to know anything about being nested under a parent command?

dimas-b: AFAIK, command nesting is derived from annotations. Implementation classes are free to extend other classes just to share code.

dimas-b:
@dennishuo : Here's the issue for CEL follow-up: #3847

Re: Admin CLI docs, would you mind if they were added after this PR? The latest docs pages are still useful even if they do not make it to the same release as the binaries.

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
snazy: @dimas-b is right.
Java type hierarchy is orthogonal to picocli commands / annotations.

dimas-b:
@dennishuo : I updated the PR discription... Is it better? :)

snazy:
Added some docs for the web site.
The reference there to the "Configuration Reference" should be updated either in #3838 or here (whichever gets merged first.)

site/content/in-dev/unreleased/admin-tool.md
```diff
@@ -208,3 +209,85 @@ docker run --rm -it \
...
+If you encounter that the maintenance run operation puts too much load on the database,
+you can adjust scan rates to reduce the load.
+The option for the reference and object scan rates are
+`polaris.persistence.nosql.maintenance.object-scan-rate-limit-per-second` and
+`polaris.persistence.nosql.maintenance.reference-scan-rate-limit-per-second`, both default to 'unlimited.'
+It is safe to let the maintenance run for a long time, there is no runtime limit.
+Alternatively, you can run maintenance at high or the default unlimited scan rate during off-peak hours.
+
+{{< alert note >}}
+The `nosql maintenance-run` command may emit a warning that not all purgable database entried have been deleted.
```
adutra: typo: entried -> entries

dimas-b:
CI failed because of: https://lists.apache.org/thread/ln4smllk1j19on3qqq840qldf9yw19hq

runtime/admin/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
dimas-b: I took the liberty to push this update from my side. @snazy : I hope you do not mind.

dimas-b:
@flyrain : I opened #3867 to allow users flexibility over dependencies in the Admin CLI. I intend to work on that follow-up issue immediately after merging this PR.

Do you think we can go ahead with this plan? Thx!

PR #3396, opened 2026-01-09
NoSQL: Add to runtime-service

snazy:
This change adds the NoSQL persistence to polaris-runtime-service.

### DIFF
--- a/runtime/defaults/src/main/resources/application.properties
+++ b/runtime/defaults/src/main/resources/application.properties
@@ -41,7 +41,7 @@ quarkus.micrometer.enabled=true
 quarkus.micrometer.export.prometheus.enabled=true
 quarkus.oidc.enabled=true
 quarkus.otel.enabled=true
-quarkus.mongodb.metrics.enabled=true
+#quarkus.mongodb.metrics.enabled=true
 #quarkus.mongodb.connection-string=mongodb://localhost:27017
 
 # ---- Runtime Configuration ----


runtime/defaults/src/main/resources/application.properties
```diff
@@ -255,6 +263,11 @@ polaris.credential-manager.type=default
 polaris.iceberg-metrics.reporting.type=default
 # Set to INFO if you want to see iceberg metric reports logged
 quarkus.log.category."org.apache.polaris.service.reporting".level=OFF
+## MongoDB version store specific configuration
+quarkus.mongodb.database=polaris
+quarkus.mongodb.metrics.enabled=true
+#quarkus.mongodb.connection-string=mongodb://localhost:27017
+quarkus.mongodb.devservices.enabled=false
```
dimas-b: Quite a few Quarkus `devservices` config properties have to be set at build-time... Should we move this setting to the build-time section (above)?

runtime/service/src/main/java/org/apache/polaris/service/persistence/PersistenceConfiguration.java
```diff
@@ -31,7 +31,7 @@ public interface PersistenceConfiguration {
    */
   String type();
 
-  @WithDefault("in-memory")
+  @WithDefault("in-memory,nosql")
```
dimas-b: I'm not sure enabling auto-bootstrap for all `nosql` backends is correct... I believe it is preferable to direct users to bootstrapping via the Admin tool on real storage. 

Can we handle this in test code for now and refactor later to make it default for nosql+inmemory?

runtime/defaults/src/main/resources/application.properties
```diff
@@ -128,9 +128,17 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
 # realm overrides
 # polaris.features.realm-overrides."my-realm"."SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION"=true
 
-# polaris.persistence.type=in-memory-atomic
+# Available types:
+# - in-memory - InMemoryPolarisMetaStoreManagerFactory
+# - in-memory-atomic - InMemoryAtomicOperationMetaStoreManagerFactory
+# - nosql - NoSQL persistence backend, define the backend type via 'polaris.persistence.nosql.backend'
```
dimas-b: Shall we mark it as "beta" as discussed in the last community sync call?

runtime/service/src/main/java/org/apache/polaris/service/persistence/PersistenceConfiguration.java
flyrain: Agreed with @dimas-b. We shouldn't leak one persistence impl to this module.

flyrain:
I don’t think this PR is moving in the right direction. NoSQL is one of the persistence impl., but this change introduces NoSQL specific concerns into the business logic module. This blurs the separation of concerns and makes the business layer dependent on a concrete persistence backend, which will make the architecture harder to evolve and maintain.
cc @dennishuo @collado-mike @singhpk234 @RussellSpitzer

dimas-b:
@flyrain : 
> NoSQL persistence impl. should NOT leak to business module.

Could you, please clarify what you mean by "business module"?

dimas-b (approved):
The latest state of this PR LGTM 👍 

Changes in `runtime/default` properly flag the new NoSQL persistence as "beta". This should be sufficient for early adopters. Documentation for mainstream users TBD later.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -41,6 +41,9 @@ quarkus.micrometer.enabled=true
 quarkus.micrometer.export.prometheus.enabled=true
 quarkus.oidc.enabled=true
 quarkus.otel.enabled=true
+quarkus.mongodb.metrics.enabled=true
```
RussellSpitzer: Should these properties also be commented out since they aren't applicable?

runtime/defaults/src/main/resources/application.properties
```diff
@@ -128,16 +131,27 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
+# - in-memory-atomic - InMemoryAtomicOperationMetaStoreManagerFactory
+# - nosql (beta) - NoSQL persistence backend, define the backend type via 'polaris.persistence.nosql.backend'
+# - relational-jdbc
 polaris.persistence.type=in-memory
-# polaris.persistence.type=relational-jdbc
+# Database backend for 'nosql' persistence-type
+# Available backends:
+# - InMemory - for testing purposes
+# - MongoDb - configure the via the Quarkus extension
+polaris.persistence.nosql.backend=InMemory
```
RussellSpitzer: Same comment here, shouldn't this be commented out since it's not in use?

runtime/defaults/src/main/resources/application.properties
dimas-b: @RussellSpitzer : what do you mean by "aren't applicable"?

runtime/defaults/src/main/resources/application.properties
dimas-b: This is a reasonable default _if_ the user selects NoSQL persistence via `polaris.persistence.type`.

It has no effect otherwise.

runtime/defaults/src/main/resources/application.properties
RussellSpitzer: If I use this application.properties as is, this property won't do anything correct?

dimas-b: True. However, if NoSQL + MongoDB is enabled, getting metrics automatically would be handy, right?

I'm ok commenting this out too.

runtime/defaults/src/main/resources/application.properties
RussellSpitzer: That was basically my thought, I generally would keep properties unset unless they do something. So in this case we have a property that's set but doesn't actually do anything because other config stops it from being valid.


Like how we do in the cloudwatch config

```
# AWS CloudWatch event listener settings
# polaris.event-listener.type=aws-cloudwatch
# polaris.event-listener.aws-cloudwatch.log-group=polaris-cloudwatch-default-group
# polaris.event-listener.aws-cloudwatch.log-stream=polaris-cloudwatch-default-stream
# polaris.event-listener.aws-cloudwatch.region=us-east-1
# polaris.event-listener.aws-cloudwatch.synchronous-mode=false
```

dimas-b: Fair enough from my POV (but I do not consider it a blocker :slightly_smiling_face: )

runtime/defaults/src/main/resources/application.properties
RussellSpitzer: Definitely! I was mostly just thinking that we generally comment out properties that are available but don't do anything in the current config. 

It may make sense to just group nosql or mongo params together too

runtime/defaults/src/main/resources/application.properties
dimas-b: side note: it might be worth setting `polaris.persistence.type` automatically if `polaris.persistence.nosql.backend` is set (future).

RussellSpitzer:
> I don’t think this PR is moving in the right direction. NoSQL is one of the persistence impl., but this change introduces NoSQL specific concerns into the business logic module. This blurs the separation of concerns and makes the business layer dependent on a concrete persistence backend, which will make the architecture harder to evolve and maintain. cc @dennishuo @collado-mike @singhpk234 @RussellSpitzer

@flyrain Could you be more explicit in your concern here? 

I see the new config in the Application.conf which seems reasonable to me, and the test integration which also seems reasonable. This mirrors what we are doing with JDBC correct? Is there something else you are worried about?

runtime/defaults/src/main/resources/application.properties
RussellSpitzer: Just reviewing and leaving some feedback I hope would be helpful :)

runtime/defaults/src/main/resources/application.properties
dimas-b: Thinking some more about this: `runtime/defaults` is basically for the default Server image (or tar). However, users still need to make decisions about monitoring and backend database parameters. So it does make sense to comment out anything that is not "activated" by default.

Helm charts can enable extra config at deployment time (e.g. if MongoDB is selected).

Users of tar distributions should still set up extra infra for monitoring, so it's not really seamless even if it's enabled by default.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -128,16 +131,27 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
 
 polaris.secrets-manager.type=in-memory
 # if set to true it will try to start localstack at build and run time for the local environment
 # https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-rds.html#_configuration_reference for more details
 quarkus.rds.devservices.enabled=false
 quarkus.rds.sync-client.type=apache
 
+## MongoDB specific configuration
+quarkus.mongodb.database=polaris
```
flyrain: Do we need this config? I don't think we have to put it here as a default configs, given that NoSql is one of persistence options.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -283,6 +297,8 @@ quarkus.arc.ignored-split-packages=\
 
 ## Quarkus required setting for third party indexing
 # fixed at build-time
+quarkus.index-dependency.agrona.group-id=org.agrona
+quarkus.index-dependency.agrona.artifact-id=agrona
```
flyrain, line 302: Question, why do we need these two new configs?

flyrain:
> @flyrain Could you be more explicit in your concern here?

My concern is around https://github.com/apache/polaris/pull/3396#discussion_r2677073955(which seems to be resolved already), and NoSql specific configs in `application.properties`. I have put additional comments there. I hope it's helpful to clarify.

RussellSpitzer (approved):
Approved, I would recommend we comment out or group the parameters prior to merging though so it isn't confusing to users who might think that those parameters are always necessary or relevant.

runtime/defaults/src/main/resources/application.properties
dimas-b: The comment just above (lines 298-299) explains it, does it not?

dimas-b:
@flyrain : 
> My concern is around https://github.com/apache/polaris/pull/3396#discussion_r2677073955 seems to be resolved already) [...]

If your concern is resolved, would you mind lifting the "request changes" flag?

If you think that something in current `application.properties` is a blocker for merging, please specifically indicate what it is to allow for clarity in addressing these concerns.

flyrain:
> If you think that something in current `application.properties` is a blocker for merging, please specifically indicate what it is to allow for clarity in addressing these concerns.

Hi @dimas-b, the Polaris "Code Contribution Guidelines" clarifies what's a blocker comment or not, https://polaris.apache.org/community/contributing-guidelines/. 
"Consider open questions and concerns in all comments of your Pull Request, provide replies and resolve addressed comments, if those don’t serve reference purposes. If a comment doesn’t contain nit, minor, or not a blocker mention, please provide feedback to the comment before merging." 

PR authors should honor this, do they?

runtime/defaults/src/main/resources/application.properties
dimas-b: indexing jars is generally required for proper CDI

flyrain: CDI has been there since we introduced Quarkus to Polaris without indexing jars. Why do we need them now?

runtime/defaults/src/main/resources/application.properties
dennishuo: +1 to commenting out everything that's specific to enabling the particular persistence type. Someone reading through the config file trying to figure out could get confused whether a mongodb config setting is active or impacts runtime behaviors despite not turning on mongodb for anything.

Is there a way for the code-level default to already have these things enabled, and have a commented-out config basically reflect the code default (for these and the other similar mongodb configs discussed elsewhere).

I'm thinking of how things like `hadoop-env.sh` or `spark-env.sh` or `spark-defaults.conf` are basically just a big file of commented-out settings that mostly just reflect code-level defaults, but make it easy for a reader to mix in what they want to customize without getting overwhelmed about bare-minimum settings.

Are Quarkus application.properties "composeable" to easily define e.g. a `mongodb-mixins.properties` or `spanner-mixins.properties` to still make it easy to choose a persistence profile without ending up with a monolithic base config file?

If not, big commented-out sections still seems a reasonable way to scale this.

runtime/defaults/src/main/resources/application.properties
dennishuo: This config setting could definitely use a comment explaining what it is and what the options are. At first glance it's not clear what the config does. Are the options, for example, "polaris" vs "glue" or "unity" ;)

Is it just the database *name* in mongodb to use when mongodb is used as the persistence layer?

I'd also prefer to comment out the configs like this one that aren't "active" by default (actually, I would've preferred to comment out the `quarkus.rds` ones as well and have a nice config section for all the `quarkus.rds.*` settings that makes it clear which config setting enables the block of rds configs too).

In my experience, it's nice to pare down the actual active config strings in the application.properties file to the true bare minimum for correct default out-of-the-box. Even the configs that match code-level defaults would be present but commented out as documentation so it's easy to see what can be adjusted, but without having to worry about whether it's a crucial input when the deployment mode isn't applicable for the config.

runtime/defaults/src/main/resources/application.properties
dennishuo: +1 to commenting out

runtime/defaults/src/main/resources/application.properties
dennishuo: If I'm understanding correctly, *indexing* isn't the new thing, but `agrona` is the new library we need to index now in addition to e.g. avro, and guava, is that right?

I do agree a bit that agrona is less ubiquitous the others here, so could be worth a quick comment in here saying what agrona does (looks like it's a fancy utils library? https://github.com/aeron-io/agrona)

In my experience avro, guava, and protobuf are ubiquitous enough that no one thinks twice seeing it in deps, but agrona might be new to some folks (like me, or maybe I'm just old-fashioned :) )

I like to lean towards over-communicating rather than risk under-communicating in comments in these cases, since things like compliance audits, etc., will have folks looking through these types of configs and anything that can avoid unnecessary questions will help save time for everyone

dennishuo (commented):
Generally I agree with trying to leave as much non-default-specific config commented out in the `appliation.properties` file as possible to manage cognitive load on service runners, especially as we scale.

I could be convinced otherwise for individual config settings, but for anything that is active but isn't supposed to be applicable without another opt-in, I'd prefer at least a comment in the file explaining the config sufficiently to put the reader at ease.

Some things seem to follow this convention reasonably well, like all the `oidc` stuff. Others like `quarkus.rds` we might want to consider also fixing forward along this convention if others agree with my take.

For example, as a developer wanting to test Polaris in various environments, seeing the `rds` settings not commented out in the properties file makes me worry whether Polaris will try to be too clever and discover RDS whenever I happen to run in an EC2 VM that has  some environment variables enabled, for example, to talk to RDS via an instance role.

dennishuo:
> > If you think that something in current `application.properties` is a blocker for merging, please specifically indicate what it is to allow for clarity in addressing these concerns.
> 
> Hi @dimas-b, the Polaris "Code Contribution Guidelines" clarifies what's a blocker comment or not, https://polaris.apache.org/community/contributing-guidelines/. "Consider open questions and concerns in all comments of your Pull Request, provide replies and resolve addressed comments, if those don’t serve reference purposes. If a comment doesn’t contain nit, minor, or not a blocker mention, please provide feedback to the comment before merging."
> 
> PR authors should honor this, do they?

@dimas-b were you just asking for clarification on the preferred fix to the file as opposed to the nature of comments being blockers?

Looks like the remaining open comment was about `quarkus.mongodb.database=polaris` needing clarification.

@flyrain I also commented on that config key as something that raised eyebrows for me - would my suggestion of adding an explanatory comment and possibly also commenting it out by default also address your concern?

runtime/defaults/src/main/resources/application.properties
dimas-b: Actually, in my local env. I experimented with commenting out _all_ of these index properties and I do not see any effect :) 

I propose to open a GH issue for a more in-depth investigation and removal of them later, on `main`. I can take that on my TODO list. WDYT?

dimas-b:
> @dimas-b were you just asking for clarification on the preferred fix to the file as opposed to the nature of comments being blockers?

My question is to specifically flag what people consider as blockers on this PR. Some approvals exist. Comment threads are long... so it's not obvious for me what remains to be addressed.

Well, the concerns about `application.properties` entries are clear :) Anything else is a blocker?

runtime/defaults/src/main/resources/application.properties
flyrain: thanks @dimas-b , that works for me. We can always add them back if it turns out useful.

dimas-b: Here's the issue: #3467

RussellSpitzer (approved):
Now that the other properties are commented out, I'm a +1 

Although I notice
```quarkus.mongodb.metrics.enabled=true``` 

Was missed

Nit: The formatting is off a bit 
#property
Instead of 
# property

Also we should just add in the indexing dependencies in another PR unless there is a relationship to this PR that I don't see.

File: runtime/defaults/src/main/resources/application-test.properties [hunk 8]
```diff
@@ -22,6 +22,7 @@
 
 quarkus.datasource.devservices.enabled=false
 quarkus.keycloak.devservices.enabled=false
+quarkus.mongodb.devservices.enabled=false
```
sfc-gh-ygu, line 25: Do we need this config?

runtime/service/build.gradle.kts
```diff
@@ -39,6 +39,12 @@ dependencies {
   compileOnly(project(":polaris-immutables"))
   annotationProcessor(project(":polaris-immutables", configuration = "processor"))
 
+  runtimeOnly(project(":polaris-persistence-nosql-metastore"))
+  runtimeOnly(project(":polaris-persistence-nosql-cdi-quarkus"))
+  runtimeOnly(project(":polaris-persistence-nosql-cdi-quarkus-distcache"))
+  runtimeOnly(project(":polaris-persistence-nosql-maintenance-impl"))
+  runtimeOnly(project(":polaris-persistence-nosql-metastore-maintenance"))
```
HonahX, line 46: Since we already did this for JDBC, adding these makes sense.

That said, looking at the `build.gradle.kts`, we now have test dependencies for multiple persistence implementations in the service module. The main implementation already separates persistence backends from runtime/service, reflecting that the service layer logic is designed to be abstract and agnostic to specific persistence implementations. In the long term, it might be worth migrating the persistence-backend integration tests (both JDBC and NoSQL) to separate test modules to stay consistent with this separation.

Not a blocker

dimas-b: I support the idea of doing integration tests separately in each major plugin module (JDBC, NoSQL, etc.). OPA does that already.

It will probably add CI time due to extra Quarkus build activity, but improves isolation.

This is certainly for future discussion / follow-up PRs.

RussellSpitzer: This is imho, a great use of a followup issue. The follow up would make things arguably more clean but requires a larger refactoring than the scope of this pr which currently just follows the existing model from JDBC.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -128,16 +130,29 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
+# - relational-jdbc
 polaris.persistence.type=in-memory
-# polaris.persistence.type=relational-jdbc
+# Database backend for 'nosql' persistence-type
+# Available backends:
+# - InMemory - for testing purposes
+# - MongoDb - configure the via the Quarkus extension
+#     Configure the necessary MongoDB properties starting with 'quarkus.mongodb.' in this file.
+#     See https://quarkus.io/guides/mongodb#configuration-reference for details about these configurations.
+#polaris.persistence.nosql.backend=InMemory
```
RussellSpitzer, line 145: nit: whitespace

runtime/defaults/src/main/resources/application.properties
```diff
@@ -128,16 +130,29 @@ polaris.features."SUPPORTED_CATALOG_CONNECTION_TYPES"=["ICEBERG_REST"]
...
 
 polaris.secrets-manager.type=in-memory
 # if set to true it will try to start localstack at build and run time for the local environment
 # https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-rds.html#_configuration_reference for more details
 quarkus.rds.devservices.enabled=false
 quarkus.rds.sync-client.type=apache
 
+## MongoDB specific configuration
+#quarkus.mongodb.database=polaris
```
RussellSpitzer, line 154: nit: whitespace

dimas-b:
Re: "indexing things", I'm trying to do a holistic investigation for that in #3467 , but it turned out to be a tricky issue due to oddities with build-time logging. I propose to keep existing indexing properties in this PR and follow up with corrections / minimally required settings when the matter is clear.

Indexing things that are not required will not harm runtime in any way, even though it might slow CI down.

RussellSpitzer:
@dimas-b But how is it related to this pr? I don't like having a commit where we look at back and we can't figure out why a change was made. It's pretty simple to just make a quick 2 line PR that adds them in, and it would take less time than our discussion about those 2 lines

dimas-b:
> how is it related to this pr? [...]

I cannot say for sure, but if tests pass under this PR without the new indexing config, I suppose we can remove it from this PR and re-add later as an optimization. I did not personally test it though. @snazy WDYT?

dimas-b:
Update: I actually [did test](https://github.com/apache/polaris/pull/3396#discussion_r2699672628) running without the indexing properties, but forgot :facepalm: 

@snazy : would you mind removing them for now? I'll re-add under #3467 if build-time warnings still pop up.

runtime/defaults/src/main/resources/application.properties
dimas-b: I did more local debugging and this indexing property does appear to be advised by Quarkus for the `runtime/server` build.

Here's the warning message from Quarkus when running _without_ this property:
```
Unable to properly register the hierarchy of the following classes for reflection as they are not in the Jandex index:
	- org.agrona.collections.Int2ObjectHashMap (source: JacksonProcessor annotated with @com.fasterxml.jackson.databind.annotation.JsonDeserialize > org.apache.polaris.persistence.nosql.authz.api.ImmutableAclEntry$Json > org.apache.polaris.persistence.nosql.authz.api.PrivilegeSet > org.apache.polaris.persistence.nosql.authz.api.Privileges > org.agrona.collections.Int2ObjectHashMap)
	- org.agrona.collections.Object2ObjectHashMap (sources: JacksonProcessor annotated with @com.fasterxml.jackson.databind.annotation.JsonDeserialize > org.apache.polaris.persistence.nosql.authz.api.ImmutableAclEntry$Json > org.apache.polaris.persistence.nosql.authz.api.PrivilegeSet > org.apache.polaris.persistence.nosql.authz.api.Privileges > org.agrona.collections.Object2ObjectHashMap, JacksonProcessor annotated with @com.fasterxml.jackson.databind.annotation.JsonDeserialize > org.apache.polaris.persistence.nosql.authz.api.ImmutableAclEntry$Json > org.apache.polaris.persistence.nosql.authz.api.PrivilegeSet > org.apache.polaris.persistence.nosql.authz.api.Privileges > org.apache.polaris.persistence.nosql.authz.api.Acl$AclBuilder > org.agrona.collections.Object2ObjectHashMap, JacksonProcessor annotated with @com.fasterxml.jackson.databind.annotation.JsonDeserialize > org.apache.polaris.persistence.nosql.coretypes.acl.AclObj > org.apache.polaris.persistence.nosql.authz.api.Acl > org.agrona.collections.Object2ObjectHashMap)
Consider adding them to the index either by creating a Jandex index for your dependency via the Maven plugin, an empty META-INF/beans.xml or quarkus.index-dependency properties.
```

That said, I'm investigating the reason why this message does not show up during normal build (no debugger) under #3467.

I believe this resolves this comment thread. @flyrain : WDYT?

dimas-b:
Another update about indexing properties: I debugged the build and I can confirm that the property is advisable to have.

Please see my comment [above](https://github.com/apache/polaris/pull/3396#discussion_r2727978172) for details.

RussellSpitzer:
> Another update about indexing properties: I debugged the build and I can confirm that the property is advisable to have.
> 
> Please see my comment [above](https://github.com/apache/polaris/pull/3396#discussion_r2727978172) for details.

Is it necessitated by this PR? That's my only big question here, It's not obvious to me why the other changes in this PR caused this issue

dimas-b:
Re: indexing properties:

> Is it necessitated by this PR? That's my only big question here, It's not obvious to me why the other changes in this PR caused this issue

This PR integrates NoSQL code into the `runtime/server` module where Quarkus starts introspecting this code and issues the [warning](https://github.com/apache/polaris/pull/3396#discussion_r2727978172). Quarkus did not introspect this code until it was part of the server build.

I'd say we ought to fix the warning from Quarkus in order to be sure that its introspection yields a complete picture for CDI to function properly. As I noted above, not having this property does not lead to any visible breakage, however, I think it is much safer to set it in order to benefit from Quarkus build-time checks to the full extent.

RussellSpitzer:
> Re: indexing properties:
> 
> > Is it necessitated by this PR? That's my only big question here, It's not obvious to me why the other changes in this PR caused this issue
> 
> This PR integrates NoSQL code into the `runtime/server` module where Quarkus starts introspecting this code and issues the [warning](https://github.com/apache/polaris/pull/3396#discussion_r2727978172). Quarkus did not introspect this code until it was part of the server build.
> 
> I'd say we ought to fix the warning from Quarkus in order to be sure that its introspection yields a complete picture for CDI to function properly. As I noted above, not having this property does not lead to any visible breakage, however, I think it is much safer to set it in order to benefit from Quarkus build-time checks to the full extent.

We should definitely include it then. I just wasn't clear on whether this had been present before this PR or not, if the other changes pulls in this as a build warning then we should keep it in this PR.

runtime/defaults/src/main/resources/application.properties
flyrain: SGTM. Thanks @dimas-b !

runtime/service/build.gradle.kts
dimas-b: For consideration: #3569

flyrain:
Removed my "request changes" as my comments are resolved. Feel free to move forward with this PR. Thanks a lot for working on it.

dimas-b:
Thanks everyone for contributions (core and reviews)! Merging. As usual, follow-up improvements are possible on `main`

PR #3409, opened 2026-01-09
feat: storage-scoped aws credentials

### LINKED ISSUE #2970: Support Per-Catalog AWS Credentials in MinIO Deployments

Currently, Polaris requires AWS credentials to be provided via environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY) at the service startup level. This creates the following problems in MinIO deployments:

- Lack of Credential Isolation: All catalogs share a single set of AWS credentials, violating the principle of least privilege. Each catalog should have its own credentials with access restricted to its specific bucket.

- Missing MinIO Support: Unlike AWS (which supports STS AssumeRole for credential vending), MinIO does not provide an equivalent service. Currently, the only option for MinIO users is to use a single shared credential set, which is not production-ready.

Add support for inline AWS credentials (accessKeyId and secretAccessKey) in the storageConfigInfo object when creating a catalog. This would allow users to:

- Pass per-catalog credentials during catalog creation:
```
{
  "name": "catalog1",
  "storageConfigInfo": {
    "accessKeyId": "catalog1_access_key",
    "secretAccessKey": "catalog1_secret_key",
    "endpoint": "https://minio.example.com",
    "storageType": "S3",
    "pathStyleAccess": true,
    "region": "us-west-2",
    "allowedLocations": ["s3://bucket1/catalog1"]
  }
}
```
- Isolate credentials per catalog - each catalog uses only its own credentials for S3 operations

_No response_

### Additional context

_No response_

tokoko:
Introduces 2 new configurations:
- `polaris.storage.aws.<storage-name>.access-key`
- `polaris.storage.aws.<storage-name>.secret-key`

Users can set these to override aws credentials per catalog. Each catalog looks up it's keys by using host of the first uri in the list of AllowedLocations or it's own configured storageName if available.

Fixes #2970


runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
```diff
@@ -97,6 +113,18 @@ default AwsCredentialsProvider stsCredentials() {
     }
   }
 
+  default AwsCredentialsProvider stsCredentials(String storageName) {
+    if (storageName != null && storages().containsKey(storageName)) {
+      StorageConfig storageConfig = storages().get(storageName);
+      if (storageConfig.awsAccessKey().isPresent() && storageConfig.awsSecretKey().isPresent()) {
```
dimas-b: nit: `storageConfig.isPresent()`?

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
```diff
@@ -57,6 +58,21 @@ public interface StorageConfiguration extends S3AccessConfig {
   @WithName("aws.secret-key")
   Optional<String> awsSecretKey();
 
+  @WithName("aws")
+  Map<String, StorageConfig> storages();
```
dimas-b: `@WithDefaults` and/or `@WithParentName` might be helpful here... WDYT? e.g. https://github.com/apache/polaris/blob/20febdaede19fb7c46e120652fdd1a262c2138e4/runtime/service/src/main/java/org/apache/polaris/service/auth/AuthenticationConfiguration.java#L36

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -105,6 +107,17 @@ public String serialize() {
     }
   }
 
+  public Optional<String> resolveStorageName() {
+    if (getAllowedLocations().isEmpty()) {
+      return Optional.empty();
+    }
+    try {
+      return Optional.ofNullable(new URI(getAllowedLocations().get(0)).getHost());
```
snazy: `URI` will fail for certain storage-locations as there can be illegal characters.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -54,15 +54,16 @@ public class PolarisStorageIntegrationProviderImpl implements PolarisStorageInte
...
   @Inject
   public PolarisStorageIntegrationProviderImpl(
       StorageConfiguration storageConfiguration, StsClientProvider stsClientProvider, Clock clock) {
-    this(
-        stsClientProvider,
-        Optional.ofNullable(storageConfiguration.stsCredentials()),
-        storageConfiguration.gcpCredentialsSupplier(clock));
+    this.storageConfiguration = storageConfiguration;
+    this.stsClientProvider = stsClientProvider;
+    this.stsCredentials = Optional.empty(); // ofNullable(storageConfiguration.stsCredentials());
```
snazy:
```suggestion
    this.stsCredentials = Optional.empty();
```

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
snazy: [See](https://smallrye.io/smallrye-config/Main/config/mappings/#withparentname), replacing the `awsAccessKey()` and `awsSecretKey()` properties.
```java
  @WithParentName
  Optional<StorageConfig> defaultSecrets();
```

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
```diff
@@ -57,6 +58,21 @@ public interface StorageConfiguration extends S3AccessConfig {
...
+
+  interface StorageConfig {
+    @WithName("access-key")
+    Optional<String> awsAccessKey();
+
+    @WithName("secret-key")
+    Optional<String> awsSecretKey();
```
snazy: Don't think `Optional`'s needed here. Both values are needed (or none).

snazy (commented):
Can we safely assume that a bucket name is "globally unique" across all realms? Mean, this approach is probably good enough to serve most users' needs. Just wondering whether there might be situations in multi-realm setups that would require a more fine-grained approach. We can certainly defer that to when we have a "secrets provider" API.

tokoko:
@snazy from the discussion in the issue, I think the point is that it's pointless to try to come up with a way to infer a good storage name that will also be guaranteed to be unique. There might be more than one allowed buckets, the same bucket name could be referring to different buckets in s3 and minio and so on. So the PR tried to infer a "good enough" name from the first entry in allowed locations. Whenever that's impossible or incorrect, the user is free to set it's own name.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
dimas-b: Good point. Cf. https://github.com/projectnessie/nessie/issues/8328

adutra: We need to improve Polaris' `StorageLocation` + `StorageType`.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
```diff
@@ -43,19 +45,35 @@ public interface StorageConfiguration extends S3AccessConfig {
 
   Duration DEFAULT_TOKEN_LIFESPAN = Duration.ofHours(1);
 
-  /**
-   * The AWS access key to use for authentication. If not present, the default credentials provider
-   * chain will be used.
-   */
-  @WithName("aws.access-key")
-  Optional<String> awsAccessKey();
```
dimas-b, line 51: It might be worth keeping this old method name for backward compatibility (and redirect the the new lookup logic)... since it's a "core" class.

dimas-b (commented):
Thanks for the update, @tokoko !

It looks like we still need to make storage location parsing more robust, but otherwise this approach LGTM.

This PR enables Polaris deployments to be configured with a finite set of per-location credentials known ahead of time, which is a nice feature for Polaris Administrators, I guess.

There are on-going discussions about leveraging (developing) a Secrets Manager for loading credentials, so this part of the Polaris codebase may need to change soon, but I hope we can still preserver the basic feature of allowing per-bucket credentials.

All in all, I think this PR is worth merging as an intermediate solution since it helps some users while the bigger Secrets Manager is under development.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
tokoko: should i switch to some sort of substring magic instead? I was hoping to rely on the fact that 1) it's unlikely that root allowed locations will contain illegal characters and 2) users can set the name explicitly as a last resort

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
tokoko: done

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
dimas-b: 1) Illegal in what sense? AWS SDK accepts locations that `java.net.URI` does not accept :shrug: If those locations are generated by code before being used in Polaris, anything that the SDK accepts is possible.

dimas-b: Custom parsing may be sufficient in this particular case, but this issue will come up again, in a different context, I'm pretty sure.

I'd personally back-port [StorageUri](https://github.com/projectnessie/nessie/blob/main/tools/storage/uri/src/main/java/org/projectnessie/storage/uri/StorageUri.java) to Polaris as I obviously trust that code :sweat_smile: We could put it into `StorageLocation` or a new class. I wonder what other people think about this.

dimas-b: Re: 2) We'd have to expose the location name in Storage Config in the Management REST API. This is possible, but I'm not sure it's the best path forward now that the Secrets Manager discussion and per-table storage config discussion are active and it's not easy to predict how they are going to affect Storage Config definitions. 

https://lists.apache.org/thread/2y2vv8jrgps7s3sr241k6tgm440jzbp9

https://lists.apache.org/thread/boqdzdtmhhk1bncv2xr43sz4nsrhgwro

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
dimas-b: @tokoko : WDYT about `@WithParentName` for config handling? I hope it might be a bit clearer.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
dimas-b: I'm fine with using `URI` for now and migrating to `StorageUri` later (note: it's already in #3256 if perhaps in a reduced form).

dimas-b: Unparseable locations will get default credentials, which is find from my POV as far as this feature is concerned.

dimas-b (commented):
The only remaining (minor) concern from the my side is the use of `@WithParentName` in configuration... otherwise, I think we can merge.

@snazy , @adutra : WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
tokoko: tbh, I was thinking maybe it would be better to instead scrap the whole uri parsing here. My main reason for it is that I think it will be a headache to build some sort of authorization solution on top unless storage names are explicitly defined on the catalog config side. wdyt?

dimas-b: I'm fine with not parsing storage URIs here

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -83,6 +84,9 @@ public boolean validatePrefix() {
 
   public abstract List<String> getAllowedLocations();
 
+  @Nullable
+  public abstract String getStorageName();
```
dimas-b, line 88: Are you planning to hook it up to Management REST API later or keep it as an internal option for custom builds?

dimas-b (dismissed):
This revision LGTM 👍 

@adutra , @snazy : WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
tokoko: added storageName to the api as well

spec/polaris-management-service.yml
```diff
@@ -1075,6 +1075,9 @@ components:
           items:
             type: string
           example: "For AWS [s3://bucketname/prefix/], for AZURE [abfss://container@storageaccount.blob.core.windows.net/prefix/], for GCP [gs://bucketname/prefix/]"
+        storageName:
+          type: string
+          description: An optional name referencing a server-side storage configuration for credential vending
```
dimas-b: suggestion: `An optional name that could be referenced in a server-side configuration (for example for tuning credential vending)`

I suppose it may be used in other use cases laters.

dimas-b (dismissed):

spec/polaris-management-service.yml
tokoko: agreed, I'll drop credential vending part. but I think "referencing a server-side configuration" is technically more correct. in my mind, catalog config references a server config rather than the other way around 😄. do you agree?

dimas-b: sgmt - thx!

dimas-b:
As required for API and (major) config changes, I started this `dev` discussion: https://lists.apache.org/thread/33c50qx1bvoyp9x8wbv1ntkgpzcnpb41

Let's way a couple of days and see if other people have comments.

spec/polaris-management-service.yml
tokoko: done

dimas-b (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -54,15 +55,21 @@ public class PolarisStorageIntegrationProviderImpl implements PolarisStorageInte
...
+      RealmConfig realmConfig,
+      Clock clock) {
+    this.storageConfiguration = storageConfiguration;
+    this.stsClientProvider = stsClientProvider;
+    this.stsCredentials = Optional.empty();
+    this.gcpCredsProvider = storageConfiguration.gcpCredentialsSupplier(clock);
+    this.realmConfig = realmConfig;
   }
 
   public PolarisStorageIntegrationProviderImpl(
```
adutra, line 75: Some further cleanup (in a follow-up PR) would be nice here, as this constructor is now only used in 2 tests but it complicates the logic for resolving the creds quite a bit.

adutra (dismissed):
Thanks @tokoko for the PR, I think it has a lot of value!

General question, maybe naive: why not do the same, at least for GCS?

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
```diff
@@ -87,16 +121,25 @@ default Supplier<StsClient> stsClientSupplier(boolean withCredentials) {
...
     }
   }
 
+  default AwsCredentialsProvider stsCredentials(String storageName) {
+    if (storageName != null && aws().storages().containsKey(storageName)) {
+      StorageConfig storageConfig = aws().storages().get(storageName);
+      return StaticCredentialsProvider.create(
+          AwsBasicCredentials.create(storageConfig.accessKey(), storageConfig.secretKey()));
+    }
+    return stsCredentials();
```
collado-mike, line 144: This seems wrong - if the storage name was passed explicitly, why do we fall back on the default credentials? Seems we should throw here.

collado-mike (changes requested):
It would be nice if the credentials could be realm-scoped, but I suppose the original default credentials were never realm-scoped, so this isn't strictly worse. However, I think that falling back to the default credentials if a storage name is incorrectly specified is bad for security. Given the catalog admins ought to know the name of the storage credential set they want to use, I think falling back to default credentials allows for a potential security hole.

dimas-b:
@collado-mike : my understanding is that this feature is generally usable only in situations when the admin owns Polaris deployment (otherwise it is not possible to configure per-name credentials at all, and the system falls back to current behaviour).

Additionally, the new behaviour is controlled by a feature flag, which is off by default. So existing users are not affected even when they upgrade to this code.

In self-hosted environments, the admin user is, of course, expected to be aware of how has permissions for creating catalogs and should act accordingly when electing to use this feature.

That said, defaulting to the global credential or not is probably the admin user's choice.

Would covering that behaviour with another feature flag address your concern?

@tokoko : Please share your perspective too.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
tokoko: fixed. nice catch, thanks. missed this one on the latest change. now that storageName is always explicitly set, it no longer makes sense to fall back on the defaults.

tokoko:
@collado-mike I agree it should ideally be realm-scoped, but that probably also applies to gcp and azure configs as well and I would like to avoid blowing up the scope of this PR.

@adutra the same reason, I guess. we don't use gcs and didn't want to unnecessary blow up the scope. plus, I think (!) the issue is not as relevant with gcs as there's not a lot of gcs-compatible products out there. if this mechanism ends up being used for table-level configs, that will probably change, though.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
tokoko: yeah, that's for backwards-compat only

dimas-b (approved):

collado-mike:
> my understanding is that this feature is generally usable only in situations when the admin owns Polaris deployment (otherwise it is not possible to configure per-name credentials at all, and the system falls back to current behaviour).
> 
> Additionally, the new behaviour is controlled by a feature flag, which is off by default. So existing users are not affected even when they upgrade to this code.

My concern isn't for existing users. It's for people who try to use the feature and accidentally or maliciously take advantage of this fallback. E.g., if an admin configures a credential called `limited` but accidentally fat-fingers `likited`, a catalog admin can create a catalog that tries to load the `limited` credentials and accidentally falls back to the default (possibly less restricted) credentials unknowingly. This silent failure leaves the admins completely unaware that there's an error in the configuration and, in the worst case, allows someone access to credentials they could use maliciously.

tokoko:
@collado-mike that was already fixed in the last commit in case you missed it, just making sure.. :)

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java
collado-mike: 🎉

collado-mike:
> @collado-mike that was already fixed in the last commit in case you missed it, just making sure.. :) 

@tokoko , I had not seen the revision. Thanks for pointing that out

collado-mike (approved):

dimas-b:
If there are no fresh comments / concerns in GH or on the `dev` ML, I propose to merge on Feb 18.

flyingImer (approved):

sfc-gh-ahemani (approved):
Sorry, I'm late to the party :) But I think this is useful and we should merge this! Can see many future extensions including adding tags/etc. to the storage configurations and expanding this to other cloud providers potentially. Great addition!

PR #3410, opened 2026-01-09
(feat)Python CLI: Switch from Poetry to UV for python package management

MonkeyCanCode:
Per discussion in https://lists.apache.org/thread/2hlh3rvmgo7ol3qn08xmyf85grck2p35, team preferred `uv` over `poetry`. Couple major changes with this PR:
1. Switched from poetry to uv
2. Installed uv in virtualenv instead of system level to avoid potential conflict and keep system clean
3. Used hatch instead of poetry-core/setuptool for build management
4. Used hatch hook for openapi code generator

For the next PR, we can do more cleanup to remove unnecessary code/scripts around virtualenv management and let uv handles those natively if preferred. Currently, it is backward compatible as existed one in term of how to invoke the CLI and run docker-compose. 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,8 +20,6 @@
 regtests/derby.log
 regtests/metastore_db
 regtests/output/
-# This file, if checked in after running for example regtests, contains unmanaged dependencies that eventually
-# cause unnecessary "security alerts" like https://github.com/apache/polaris/pull/718.
 client/python/uv.lock
 
 # Python stuff


client/python/generate_clients.py
```diff
@@ -1,19 +1,11 @@
-#  Licensed to the Apache Software Foundation (ASF) under one
-#  or more contributor license agreements.  See the NOTICE file
-#  distributed with this work for additional information
-#  regarding copyright ownership.  The ASF licenses this file
-#  to you under the Apache License, Version 2.0 (the
-#  "License"); you may not use this file except in compliance
-#  with the License.  You may obtain a copy of the License at
 #
```
kevinjqliu, line 1: unintentional?

kevinjqliu (commented):
generally lgtm, i would double check the build artifacts to make sure that its the same as packaged previously using poetry. That tripped us up during the pyiceberg migration

MonkeyCanCode:
> generally lgtm, i would double check the build artifacts to make sure that its the same as packaged previously using poetry. That tripped us up during the pyiceberg migration

Yes, I did that earlier. Most of them are same except wheel one doesn't have the root level __init__.py (which is fine). One thing caught me earlier was spec as that got excluded from hatch build due to we have spec in gitignore.

client/python/generate_clients.py
dimas-b: looks like it just removes a duplicate header

MonkeyCanCode: Intensional. This file has slightly diff license header (extra space after hashtag) compared to the rest of the python files.

client/python/pyproject.toml
```diff
@@ -44,39 +46,48 @@ repository = "https://github.com/apache/polaris/"
...
+
+[tool.hatch.build.hooks.custom]
+path = "hatch_build.py"
 
 [build-system]
 # 7.12.0 is the latest version to use due to OpenAPI spec version we are using
-requires = ["poetry-core==2.2.1", "openapi-generator-cli==7.12.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["hatchling", "openapi-generator-cli==7.12.0"]
+build-backend = "hatchling.build"
```
HonahX, line 83: This solves the problem with `poetry.core.masonry.api` that we cannot build universal wheel if using custom build script. I checked that now with `make client-build`, we will get
```
Successfully built dist/apache_polaris-1.4.0.tar.gz
Successfully built dist/apache_polaris-1.4.0-py3-none-any.whl
```
This could significantly simplify the nightly build/release process that we do not need the additional wheel build workflow anymore: https://github.com/apache/polaris/pull/3036

.gitignore
```diff
@@ -20,12 +20,10 @@
 regtests/derby.log
 regtests/metastore_db
 regtests/output/
-# This file, if checked in after running for example regtests, contains unmanaged dependencies that eventually
-# cause unnecessary "security alerts" like https://github.com/apache/polaris/pull/718.
```
HonahX, line 24: [Question] Should we keep this comment since it explains why we exclude `*.lock" file from git? To be honest, I also wondered for a while why poetry.lock was excluded intially.

HonahX (commented):
Overall LGTM! Thanks for migrating this! I believe this could also make the release process easier in the future. 

One thing I noticed was that the spark getting-started example now miss the ipynb in jupyter lab:

Image

But this may not be relevant to this PR.

MonkeyCanCode:
> Overall LGTM! Thanks for migrating this! I believe this could also make the release process easier in the future.
> 
> One thing I noticed was that the spark getting-started example now miss the ipynb in jupyter lab:
> 
> Image
> But this may not be relevant to this PR.

So this may be due to change of base image that I done long time back where `ipynb` is not part of `jupyterlab` library. But do we need this dependency as the current getting started notebook is fully functional.

.gitignore
MonkeyCanCode: > [Question] Should we keep this comment since it explains why we exclude `*.lock" file from git? To be honest, I also wondered for a while why poetry.lock was excluded intially.

That is a very good call out. So earlier we have poetry.lock excluded due to project dependencies updated but not updated for the lock file. But it does appears renovate bot can handle that as well (https://docs.renovatebot.com/modules/manager/pep621/#additional-information). Let me add this line back in as well as the uv.lock file. But I am not sure if uv.lock will actually get update when it is in .gitignore file.

kevinjqliu (approved):
LGTM

.gitignore
HonahX: Should we also remove it from `.gitignore` if the comment there is no longer a problem? In the official doc: https://docs.astral.sh/uv/guides/projects/#uvlock, it states that the `uv.local` should be checked in the version control and it is also a common practice (e.g. pyiceberg)

MonkeyCanCode: Yeah, let me update it now.

kevinjqliu:
>  It turns out I am still on the old image....

ha i ran into this recently too. adding the `--build` flag helps auto rebuild https://github.com/apache/iceberg-python/pull/2885

MonkeyCanCode:
thanks for the review @kevinjqliu, @HonahX , and @dimas-b . @dimas-b any last concern before I merge this one?

PR #3411, opened 2026-01-10
Use new Request Context for each realm during implicit bootstrap

dimas-b:
The implicit (auto) bootstrap calls used to share Request Context for potentially many realms. That used to work by coincidence because `RealmConfig`, for example, is a `RequestScoped` bean.

With this change each realm will be bootstrapped in its own dedicated Request Context.

This change lays down a foundation for future refactoring related to `RealmConfig`.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
@@ -37,13 +37,20 @@ import org.apache.polaris.service.context.catalog.RealmContextHolder;
 /** Utility class for running per-realm bootstrap tasks each in a fresh Request Context. */
 @ApplicationScoped
 class Bootstrapper {
-  @Inject private RealmContextHolder realmContextHolder;
-  @Inject private MetaStoreManagerFactory factory;
+  private final ExecutorService executor;
+  private final RealmContextHolder realmContextHolder;
+  private final MetaStoreManagerFactory factory;
 
-  // Note: this executor is expected to NOT propagate CDI contexts to tasks.
   @Inject
-  @Identifier("task-executor")
-  private ExecutorService executor;
+  Bootstrapper(
+      // Note: this executor is expected to NOT propagate CDI contexts to tasks.
+      @Identifier("task-executor") ExecutorService executor,
+      RealmContextHolder realmContextHolder,
+      MetaStoreManagerFactory factory) {
+    this.executor = executor;
+    this.realmContextHolder = realmContextHolder;
+    this.factory = factory;
+  }
 
   Map<String, PrincipalSecretsResult> bootstrapRealms(
       Iterable<String> realmIds, RootCredentialsSet rootCredentialsSet) {


runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -285,7 +287,13 @@ public void maybeBootstrap(
...
-      var result = factory.bootstrapRealms(realmIds, rootCredentialsSet);
+      HashMap<String, PrincipalSecretsResult> result = new HashMap<>();
+      for (String realmId : realmIds) {
+        PrincipalSecretsResult r = bootstrapper.bootstrapRealm(realmId, rootCredentialsSet);
+        if (r != null) {
+          result.put(realmId, r);
+        }
+      }
 
       result.forEach(
```
snazy, line 290: Looks like this function doesn't really need `Map<> results` anymore, only to populate `unusedRealmSecrets` below.

dimas-b: Makes sense - will fix

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
```diff
@@ -0,0 +1,75 @@
...
+public class Bootstrapper {
+  @Inject private RealmContextHolder realmContextHolder;
+  @Inject private MetaStoreManagerFactory factory;
+
+  // Note: this executor is expected to NOT propagate CDI contexts to tasks.
+  @Inject
+  @Identifier("task-executor")
+  private ExecutorService executor;
+
+  PrincipalSecretsResult bootstrapRealm(String realmId, RootCredentialsSet rootCredentialsSet) {
```
adutra: Wouldn't it be better to return a `CompletionStage`, then combine them all and wait for the combined result?

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -285,7 +287,13 @@ public void maybeBootstrap(
           RootCredentialsSet.ENVIRONMENT_VARIABLE,
           RootCredentialsSet.SYSTEM_PROPERTY);
 
-      var result = factory.bootstrapRealms(realmIds, rootCredentialsSet);
+      HashMap<String, PrincipalSecretsResult> result = new HashMap<>();
```
adutra: I think this should be moved to `Bootstrapper` as well. The `Bootstrapper` should take a set of ream ids as input and return a `Map<String, PrincipalSecretsResult>`. Wdyt?

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
dimas-b: I wanted to maintain old sequential bootstrap logic... this is not a main call path in real deployments, so I guess efficiency is not critical :thinking:

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
dimas-b: Yes, I'll move the map to `Bootstrapper`, but I'll keep sequential execution for now, if you do not mind :)

dimas-b: updated

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
dimas-b: I went with @adutra 's suggestion... WDYT?

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    return result;
+  }
+
+  private record Task(
+      RealmContextHolder realmContextHolder,
+      String realmId,
+      RootCredentialsSet rootCredentialsSet,
+      MetaStoreManagerFactory factory)
```
snazy, line 77:
```suggestion
      String realmId,
      RootCredentialsSet rootCredentialsSet)
```

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
```diff
@@ -0,0 +1,82 @@
...
+  // Note: this executor is expected to NOT propagate CDI contexts to tasks.
+  @Inject
+  @Identifier("task-executor")
+  private ExecutorService executor;
+
+  Map<String, PrincipalSecretsResult> bootstrapRealms(
+      Iterable<String> realmIds, RootCredentialsSet rootCredentialsSet) {
+    HashMap<String, PrincipalSecretsResult> result = new HashMap<>();
+    for (String realmId : realmIds) {
+      Task t = new Task(realmContextHolder, realmId, rootCredentialsSet, factory);
```
snazy, line 59:
```suggestion
      Task t = new Task(realmId, rootCredentialsSet);
```

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.apache.polaris.core.persistence.MetaStoreManagerFactory;
+import org.apache.polaris.core.persistence.bootstrap.RootCredentialsSet;
+import org.apache.polaris.core.persistence.dao.entity.PrincipalSecretsResult;
+import org.apache.polaris.service.context.catalog.RealmContextHolder;
+
+/** Utility class for running per-realm bootstrap tasks each in a fresh Request Context. */
+@ApplicationScoped
+public class Bootstrapper {
```
snazy:
```suggestion
class Bootstrapper {
```
Don't think this type should be used from other packages.

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
```diff
@@ -0,0 +1,82 @@
...
+  @Inject private RealmContextHolder realmContextHolder;
+  @Inject private MetaStoreManagerFactory factory;
```
snazy:
```suggestion
  @Inject RealmContextHolder realmContextHolder;
  @Inject MetaStoreManagerFactory factory;
```
[Less work](https://quarkus.io/guides/cdi-reference#quarkus-arc_quarkus-arc-transform-private-injected-fields) for DI instrumentation in Quarkus, and potentially nicer for Weld.

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
```diff
@@ -0,0 +1,82 @@
...
+
+  // Note: this executor is expected to NOT propagate CDI contexts to tasks.
+  @Inject
+  @Identifier("task-executor")
+  private ExecutorService executor;
```
snazy:
```suggestion
  ExecutorService executor;
```

snazy:
```suggestion
  AsyncExec executor;
```
Would prevent a dependency from bootstrapping to the current tasks implementation.

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
dimas-b: Sorry, I do not understand this suggestion :thinking: What does it mean to achieve?

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
dimas-b: Do you mean `Task` to be a proper inner class (not static)?

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
snazy: Oops, looks the text didn't make it...

The record's code can use the outer fields.

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
snazy: ah, yea - it would need to be non-static.
Let's maybe discard my comment then.

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
dimas-b: changed to ctor-based injection

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
dimas-b: updated

runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java
dimas-b: Not sure I understand this, sorry... Could you explain is another way :thinking:

snazy: `task-executor` is IIRC used to run the tasks. Not a big deal, just thinking it could be nice to use a "generic" pool.

dimas-b: Generic pool would be nice, indeed, but I'm not sure about CDI implications there :thinking: The `task-executor` pool is known NOT to propagate CDI context to tasks (#3210)

dimas-b: changed to ctor-based injection

Issue #3337, opened 2025-12-30
(feat) Implement Iceberg REST Catalog Metrics API for Compute Audit Correlation


obelix74:
### Is your feature request related to a problem? Please describe.

Implement the Apache Iceberg REST Catalog Metrics API (`POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/metrics`) to enable compute engines (Spark, Trino, Flink) to report query-level metrics back to the catalog. This enables correlation between catalog operations and compute-side query execution for comprehensive audit trails.

**Motivation:**

Currently, the Polaris catalog provides audit logging for catalog-level operations (table metadata access, namespace management, credential vending). However, **critical query-level metrics are not visible at the catalog layer**:

| Data Point | Available at Catalog? | Why Not? |
|------------|:---------------------:|----------|
| Rows read/written | No| Data flows directly from compute to storage |
| Query text (SQL) | No | Queries are composed in compute engines |
| Bytes processed | No | File scanning happens in compute engines |
| Query duration | No | Catalog only handles metadata lookups |

This gap prevents:
* Fine-grained audit trails (table → actual data access)
* Cost allocation by catalog/namespace/table based on actual usage
* Security forensics for data access patterns
* Compliance reporting with complete access records

### Describe the solution you'd like

**Proposed Solution:**

Implement the Iceberg REST Catalog Metrics endpoint as defined in the [Apache Iceberg REST Catalog OpenAPI Specification](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml) (line 1316):

```
POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/metrics
```

The endpoint accepts `ReportMetricsRequest` payloads containing either:
- **ScanReport**: For read operations (snapshot-id, filter, projected fields, scan metrics)
- **CommitReport**: For write operations (snapshot-id, sequence-number, operation, commit metrics)

Correlation with catalog events is achieved via `trace-id` passed in the `metadata` field of the metrics report.

**Acceptance Criteria:**

- [ ] Implement `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/metrics` endpoint
- [ ] Accept `ReportMetricsRequest` schema (ScanReport and CommitReport)
- [ ] Validate request against Iceberg OpenAPI specification
- [ ] Return HTTP 204 No Content on success (per Iceberg spec)
[...]

PR #3414, opened 2026-01-11
feat: Add trace_id to AWS STS session tags for end-to-end correlation

obelix74:
## Checklist

Fixes a part of https://github.com/apache/polaris/issues/3337

This change enables deterministic correlation between:
- Catalog operations (Polaris events)
- Credential vending (AWS CloudTrail via STS session tags)
- Metrics reports from compute engines (Spark, Trino, etc.)

Changes:
1. Add traceId field to CredentialVendingContext
   - Marked with @Value.Auxiliary to exclude from cache key comparison
   - Every request has unique trace ID, so including it in equals/hashCode would prevent all cache hits
   - Trace ID is for correlation/audit only, not authorization

2. Extract OpenTelemetry trace ID in StorageAccessConfigProvider
   - getCurrentTraceId() extracts trace ID from current span context
   - Populates CredentialVendingContext.traceId for each request

3. Add trace_id to AWS STS session tags
   - AwsSessionTagsBuilder includes trace_id in session tags
   - Appears in CloudTrail logs for correlation with catalog operations
   - Uses 'unknown' placeholder when trace ID is not available

4. Update tests to verify trace_id is included in session tags

This enables operators to correlate:
- Which catalog operation triggered credential vending
- Which data access events in CloudTrail correspond to catalog operations
- Which metrics reports correspond to specific catalog operations


obelix74:
@dimas-b as suggested by you, I have created a separate PR for the AWS STS trace_id changes.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -187,6 +189,24 @@ private CredentialVendingContext buildCredentialVendingContext(
       builder.activatedRoles(Optional.of(rolesString));
     }
 
+    // Extract OpenTelemetry trace ID for end-to-end correlation
+    // This enables correlation between credential vending (CloudTrail), catalog operations
+    // (Polaris events), and metrics reports from compute engines
+    builder.traceId(getCurrentTraceId());
```
dimas-b: Since `CredentialVendingContext` is part of the credentials cache, adding trace ID will make the cache ineffective (100% misses), if I understand correctly.

While this use case may be meaningful in some environments, it may lead to undesirable behaviour in others.

I propose to cover this with a separate feature flag in this PR (to allow using STS tags without trace IDs). WDYT?

dimas-b:
side note: I think this PR will need a CHANGELOG entry (eventually)

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
```diff
@@ -67,6 +70,19 @@ public interface CredentialVendingContext {
...
+   * credential vending (CloudTrail), catalog operations (Polaris events), and metrics reports from
+   * compute engines.
+   *
+   * <p>This field is marked as {@link Value.Auxiliary} to exclude it from {@code equals()} and
+   * {@code hashCode()} methods. This is critical for cache key comparison - including trace ID
+   * would prevent cache hits since every request has a unique trace ID. The trace ID is for
+   * correlation/audit purposes only and should not affect credential caching behavior.
+   */
+  @Value.Auxiliary
+  Optional<String> traceId();
```
obelix74, line 84: @dimas-b please see this, does this not make sure `traceId` is omitted from the cache?  (Edit: I meant to say `traceId` is not used in the cache key, not omitted).

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
obelix74: @dimas-b I will add a separate feature flag in this PR. Please see my comment in `CredentialVendingContext` - I think that should address `traceId` and caching.

obelix74: @dimas-b also, given that the `traceId` will not be used in cache, do we still need a separate feature flag?  There exists `INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL` to include all session tags in totality that was merged as part of https://github.com/apache/polaris/pull/3327.

obelix74:
> side note: I think this PR will need a CHANGELOG entry (eventually)

Perhaps after https://github.com/apache/polaris/pull/3385 is merged, I will add a `CHANGELOG` entry summarized how end-to-end telemetry works? A TL/DR of the google doc that was shared?

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
snazy: `@Value.Auxiliary` doesn't help there; those properties `will be stored and will be accessible.`

obelix74: Hi @snazy , the properties should be stored and accessible, that is the goal so they can be correlated. The key is they are not used for caching. 

```
Attributes annotated with @Value.Auxiliary will be stored and will be accessible, but are excluded from equals, hashCode and toString method implementations."
```

https://github.com/immutables/immutables/blob/master/value-annotations/src/org/immutables/value/Value.java

The trace ID is still stored and accessible for propagation to AWS STS session tags, but that is the goal of this PR.

dimas-b: > [...] does this not make sure traceId is omitted from the cache?

Any parameters that affect the vended credential must be part of the cache key for the sake of correctness.

Please also consider @snazy 's comment.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
dimas-b: Any parameters that affect the vended credential must be part of the cache key for the sake of correctness.

A feature flag is required for users to turn off trace IDs in tags and have better cache efficiency if they wish so.

dimas-b: heads up: I have WIP work about Credential Cache efficiency (non-conflicting with this PR, but aiming to refactor code for the sake of clarity and convenience).

For now, I believe my previous statement about cache keys holds.

adutra:
Hi @obelix74 thanks for providing support for trace ids!

I have a question: should we also include the request ID?

This aligns with an earlier, extensive discussion (back in October) regarding which information to include in events (OTel context vs. request IDs). The general agreement was to include request IDs when available.

Reference link: https://lists.apache.org/thread/p9357rcy3d1j94w4yogtdwcf2kxzg3jr

As noted in that thread: 

> Request ID should remain the canonical identifier for every request handled by Polaris. [...] OTel context is OPTIONAL."

I think it would be good to apply the same decisions for AWS STS session tags, wdyt?

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
obelix74: I added the feature flag. I have updated the code for cache correctness vs. efficiency trade-off by including trace IDs in cache keys only when they affect the vended credentials.  I hope that addresses your questions. @snazy @dimas-b please let me know if I am on the right track.

obelix74:
> Hi @obelix74 thanks for providing support for trace ids!
> 
> I have a question: should we also include the request ID?
> 
> This aligns with an earlier, extensive discussion (back in October) regarding which information to include in events (OTel context vs. request IDs). The general agreement was to include request IDs when available.
> 
> Reference link: https://lists.apache.org/thread/p9357rcy3d1j94w4yogtdwcf2kxzg3jr
> 
> As noted in that thread:
> 
> > Request ID should remain the canonical identifier for every request handled by Polaris. [...] OTel context is OPTIONAL."
> 
> I think it would be good to apply the same decisions for AWS STS session tags, wdyt?

Hi Alex

In an earlier PR https://github.com/apache/polaris/pull/3327/changes/BASE..11f4c58ce6f24284553e73887bd8d0d2991244ff, I had included the `request_id`. I had run into a few issues and based on @singhpk234 's advise, removed it from that PR.

1. I couldn't figure out a reliable way to obtain the `request_id` - I was using the `RESTContext` but @dimas-b pointed out that won't work for API calls. I then replaced it with `SLF4J`'s `MDC` - but that felt hacky.
2. My understanding is that Polaris generates a new `request_id` per incoming call, whereas the `trace_id` is more "atomic" and ensures that if a single atomic operation results in multiple REST calls, all get the same ID.

Based on these two issues, we had removed `request_id` from that pull request. Perhaps, this is a good time to add it back.

What are your thoughts? Also, any guidance on a reliable way to obtain the `request_id` would be helpful.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
```diff
@@ -78,6 +99,17 @@ public static List<Tag> buildSessionTags(String principalName, CredentialVending
...
             .build());
 
+    // Only include trace ID if explicitly enabled
+    // When disabled, this allows credential caching to work effectively since trace IDs
+    // are unique per request and would otherwise prevent any cache hits
+    if (includeTraceId) {
+      tags.add(
+          Tag.builder()
+              .key(CredentialVendingContext.TAG_KEY_TRACE_ID)
+              .value(truncateTagValue(context.traceId().orElse(TAG_VALUE_UNKNOWN)))
```
dimas-b: Since `context.traceId()` is an option, it is probably sufficient to check `INCLUDE_TRACE_ID_IN_SESSION_TAGS` only at the cache level. WDYT?

Direct calls to `AwsSessionTagsBuilder` (if they exist) should just pass the `CredentialVendingContext` with the right data.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -60,10 +60,38 @@ public interface StorageCredentialCacheKey {
...
+   *   <li>When {@code INCLUDE_TRACE_ID_IN_SESSION_TAGS} is enabled, trace IDs DO affect credentials
+   *       (via session tags), so this should contain the trace ID for cache correctness
+   * </ul>
+   *
+   * <p>This explicit field ensures cache correctness: credentials with different trace IDs are
+   * correctly treated as different cache entries when (and only when) trace IDs affect the
+   * credentials.
+   */
+  @Value.Parameter(order = 10)
+  Optional<String> traceIdForCaching();
```
dimas-b: `CredentialVendingContext` is already a part of this key and contains `traceId()`... why do we need a separate trace ID property here? 🤔

polaris-core/src/main/java/org/apache/polaris/core/storage/CredentialVendingContext.java
```diff
@@ -67,6 +70,20 @@ public interface CredentialVendingContext {
...
+   * credential vending (CloudTrail), catalog operations (Polaris events), and metrics reports from
+   * compute engines.
+   *
+   * <p>This field is marked as {@link Value.Auxiliary} to exclude it from {@code equals()} and
+   * {@code hashCode()} methods by default. When the {@code INCLUDE_TRACE_ID_IN_SESSION_TAGS}
+   * feature is enabled, trace IDs DO affect the vended credentials (via session tags), so they must
+   * be explicitly included in the cache key. See {@link
+   * org.apache.polaris.core.storage.cache.StorageCredentialCacheKey} for how this is handled.
+   */
+  @Value.Auxiliary
```
dimas-b: I believe this should be a normal value to avoid confusion. It is used as input in STS requests, so it should be considered in the credential cache's key. IMHO, it is easier to achieve that by making this a normal property.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
obelix74: Good point — I refactored this so we don’t thread a boolean down into AwsSessionTagsBuilder anymore.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
obelix74: I implemented this by making CredentialVendingContext.traceId() a normal property (removed @Value.Auxiliary), so it naturally participates in equality/hashCode when present. With that, I removed the separate traceIdForCaching field from StorageCredentialCacheKey entirely.  It was related to your third observation.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -104,6 +104,20 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> INCLUDE_TRACE_ID_IN_SESSION_TAGS =
+      PolarisConfiguration.<Boolean>builder()
+          .key("INCLUDE_TRACE_ID_IN_SESSION_TAGS")
```
dimas-b, line 109: nit: since we're adding a feature flag, I'd prefer to mention it in CHANGELOG.md right away.

obelix74: I will add it shortly.  Can you please see @adutra 's comment about `request_id`? If you remember, in the previous PR about session tags, we had explicitly removed `request_id` because we couldn't find a reliable way to get the `request_id`.

dimas-b: From my personal POV, I do not know of real use cases for "request IDs", but I suppose some people use it.

I'd be ok leaving it out for now and adding later if concrete use cases for propagating it to STS arise.

@adutra : WDYT?

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Added `topologySpreadConstraints` support in Helm chart.
 - Added `priorityClassName` support in Helm chart.
 - Added support for including principal name in subscoped credentials. `INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` (default: false) can be used to toggle this feature. If enabled, cached credentials issued to one principal will no longer be available for others.
+- Added support for optionally including OpenTelemetry trace IDs in AWS STS session tags (when session tags are enabled). `INCLUDE_TRACE_ID_IN_SESSION_TAGS` (default: false) can be used to toggle this feature.
```
obelix74: @dimas-b FYI.

dimas-b: `INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL` is _also_ required for trace IDs, right? Should we mention that for clarity?

obelix74: Ah yes, let me add that.

obelix74: Pushed a commit clarifying this.

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Added `topologySpreadConstraints` support in Helm chart.
 - Added `priorityClassName` support in Helm chart.
 - Added support for including principal name in subscoped credentials. `INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` (default: false) can be used to toggle this feature. If enabled, cached credentials issued to one principal will no longer be available for others.
+- Added support for including OpenTelemetry trace IDs in AWS STS session tags. This requires session tags to be enabled via `INCLUDE_SESSION_TAGS_IN_SUBSCOPED_CREDENTIAL` and can be toggled with `INCLUDE_TRACE_ID_IN_SESSION_TAGS` (default: false). Note: enabling trace IDs disables credential caching (each request has a unique trace ID), which may increase STS calls and latency.
```
obelix74, line 64: @dimas-b FYI.

dimas-b (dismissed):

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
adutra: If you are able to compute things like the principal or the otel context, it's because there is an active request context when the tags are computed. So, I'm a bit surprised that it's not possible to get the request ID as well. Have you tried this approach:

https://github.com/apache/polaris/blob/64b13a9915274a04e82a12c9fdcd095d12797ca0/runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java#L111-L120

I'm not saying we absolutely must include the request id here, but I do remember that for events, not including the request ID was a friction point for some contributors.

obelix74: Thanks @adutra. I am on it.

obelix74: @adutra and @dimas-b does the aforementioned code snippet handle non-HTTP clients or programmatic clients as well?

obelix74: I have pushed a commit adding support for `request_id`, but my understanding is that the code that fetches the `request_id` will only work for `HTTP` requests - this was the primary reason we removed `request_id` from the previous PR that added `request_id`.

I request @dimas-b @adutra and @singhpk234 to review this and let me know if we should keep this as part of this PR or remove it. 

The `request_id` code follows the same design suggested by @dimas-b for the `trace_id`. There is one change. Since the `request_id` is controlled by the end-user and may contain non-STS friendly characters, I added code to sanitize it - if this doesn't happen, vended credentials may fail to be issued.

dimas-b: I did not review the latest changes yet, posting proactively: STS may be involved in non-HTTP requests, so if request ID is part of STS tags we need to find a way to access / generate it for async tasks too (cf. `TaskManagerImpl`)... I'll review actual code later :hourglass_flowing_sand:

obelix74: @dimas-b thank you. Currently the code handles the absence of request-id by omitting it (request-id is optional, just like trace-id is). I will wait for your review.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
```diff
@@ -78,24 +98,68 @@ public static List<Tag> buildSessionTags(String principalName, CredentialVending
...
+  /**
+   * Sanitizes a tag value by replacing characters that are not allowed in AWS STS session tag
+   * values. AWS STS allows: Unicode letters, whitespace, numbers, and specific special characters
+   * (_.:/=+\-@). Invalid characters are replaced with underscores.
+   *
+   * @param value the value to sanitize (must not be null)
+   * @return the sanitized value with invalid characters replaced by underscores
+   */
+  static String sanitizeTagValue(String value) {
+    return INVALID_TAG_VALUE_CHARS.matcher(value).replaceAll("_");
```
dimas-b: What's the point in including trace ID / request ID if their values are modified? They will become useless for automated processing / correlation, I think 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -187,6 +193,62 @@ private CredentialVendingContext buildCredentialVendingContext(
...
+
+  /**
+   * Extracts the current request ID from the active RESTEasy Reactive request context.
+   *
+   * <p>Note: avoid injecting ContainerRequestContext here because some tests may run with no active
+   * request scope.
+   */
+  private Optional<String> getCurrentRequestId() {
+    // See org.jboss.resteasy.reactive.server.injection.ContextProducers
+    ResteasyReactiveRequestContext context = CurrentRequestManager.get();
```
dimas-b: Sorry, but I do not think this is a totally correct way to get request IDs. IIRC, `CurrentRequestManager.get()` will fail in runtime if the this method is called outside an HTTP (REST) request context, which is possible in async tasks.

Please see how `TaskExecutorImpl` deals with realm IDs. A similar CDI pattern is probably necessary for request IDs, except that we may not have to produce a request ID for background tasks. I do not have an opinion on whether request IDs should be propagated from REST requests to related async tasks or not. However, CDI must be solid and not cause runtime exceptions even if the request ID is not propagated.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: commented on related CDI aspects separately.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
obelix74: I was trying to be defensive in that an invalid request_id can derail vended credential grant (note that this is not an issue with trace_ids), but I see your point.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
obelix74: Can I file another feature request to wire the request IDs through a CDI pattern similar to realm IDs and fix it cleanly in that PR?  For this PR, I can roll back this last commit and merge it with just the trace_id. Once we have a reliable way to inject request_ids, we can add it to the STS tags then.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
dimas-b: From my POV if the trace ID value is not compliant with STS API requirements, I'd prefer to drop this tag with a WARN log message.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
dimas-b: From my POV it's totally fine to leave request IDs out of this PR and add in a follow-up PR (if there's demand).

dimas-b: @obelix74 : if you feel like continuing working on request IDs in STS tags, I think it might be preferable to start with a `dev` email about that to gauge interest (not all people watch all PRs :wink: ).

obelix74: @dimas-b Raised https://github.com/apache/polaris/issues/3444. Reverted my last commit so this PR is back to what you approved. Once I merge this PR and the metrics persistence PR, I can work on the request ID PR, and will start a discussion with the ML.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsSessionTagsBuilder.java
obelix74: @dimas-b 

If `trace_id` is coming from OpenTelemetry (`Span.current().getSpanContext().getTraceId()`), then in practice it shouldn’t ever be “invalid” for AWS STS session tags.
     • **Format/charset**: OpenTelemetry trace IDs are represented as a 32‑character lowercase hex string (`[0-9a-f]{32}`), and AWS session tag values allow letters/numbers plus a small set of symbols. Hex is entirely within that allowed set.
     • **Length**: 32 characters is far below the STS tag value limit (256).
     • **Validity gating in Polaris**: In StorageAccessConfigProvider#getCurrentTraceId(), we only include it when `spanContext.isValid()` is true; if the context is invalid (including “all zeros” or malformed), we return `Optional.empty()` and the `trace_id` tag won’t be added at all. 

The code above was specifically added for `request_id`s since they can be arbitrary strings - it has been removed in the revert of the last commit.

obelix74:
> LGTM 👍 Thanks again, @obelix74 !

Hi @dimas-b can this PR be merged?

dimas-b:
Behaviour changes in this PR are covered by a feature flag (with default being "no change"). The PR has been in review for a substantial period of time. I'm merging.

If concerns are raised later, let's address in follow-up PRs.

Issue #2365, opened 2025-08-15
Purge table task implementation prone to OOMs


snazy:
### Describe the bug

The implemented logic to prune Iceberg tables puts a huge amount of pressure to the Java heap.

The logic roughly works like this:
* Input parameter: pointer to a table metadata
* Read table metadata
* Read manifest lists of all snapshots
* Read all manifest files of all manifest lists
* Create a new task entity for each manifest file

Each manifest file can become really big, many MB of binary data.

The [base64 encoded full binary Iceberg manifest files](https://github.com/apache/polaris/blob/c9efc6c1af202686945efe2e19125e8f116a0206/runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java#L194) are included in the task entities, which are [consumed here](https://github.com/apache/polaris/blob/c9efc6c1af202686945efe2e19125e8f116a0206/runtime/service/src/main/java/org/apache/polaris/service/task/ManifestFileCleanupTaskHandler.java#L67).

Although some Java stream handling is being used, all manifest-files of the table to purge are [materialized at once on the Java heap](https://github.com/apache/polaris/blob/c9efc6c1af202686945efe2e19125e8f116a0206/runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java#L111-L130).

Even worse, the base64 encoded data is added to a JSON serialized object, which is added to a properties bag, which is in turn JSON serialized. Since JSON re-serializations of that property bag are "normal", the total heap pressure is bigger than the sum of all base64 serialized manifest-files.

flyrain:
I am not sure the issue as currently described is actually valid.

The base64 encoded manifest objects[1] being discussed are not the manifest files themselves. They are objects representing manifest files, which can be reconstructed from the manifest entries stored in the ManifestList files. As a result, the in memory footprint should be roughly equivalent to the size of a single manifest list file per snapshot, plus some additional base64 encoding overhead. That overhead does not seem significant enough on its own to explain large heap pressure.

This pattern is also handled in practice today. For example, multiple Spark procedures/actions and Spark planning process these manifest representations within a single node, typically the driver, without materializing full manifest files in memory. One concrete example is here:
https://github.com/apache/iceberg/blob/bf1074ff373c929614a3f92dd4e46780028ac1ca/spark/v4.0/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteTablePathSparkAction.java#L290

[...]

PR #3415, opened 2026-01-11
Service: Use iterator to avoid high space complexity

flyrain:
Fix https://github.com/apache/polaris/issues/2365#issuecomment-3723184230

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
@@ -212,15 +212,11 @@ public class TableCleanupTaskHandler implements TaskHandler {
       PolarisMetaStoreManager metaStoreManager,
       PolarisCallContext polarisCallContext) {
     // read the manifest list for each snapshot. dedupe the manifest files and schedule a
-    // cleanupTask
-    // for each manifest file and its data files to be deleted
-    // Use a Set to track seen paths for deduplication without materializing all ManifestFile
-    // objects
+    // cleanupTask for each manifest file and its data files to be deleted
     Set<String> seenPaths = new HashSet<>();
     return tableMetadata.snapshots().stream()
         .flatMap(sn -> sn.allManifests(fileIO).stream())
         // distinct by manifest path, since multiple snapshots will contain the same manifest
-        // Use stateful filter to dedupe while streaming
         .filter(mf -> seenPaths.add(mf.path()))
         .filter(mf -> TaskUtils.exists(mf.path(), fileIO))
         .map(
@@ -265,7 +261,6 @@ public class TableCleanupTaskHandler implements TaskHandler {
     PolarisCallContext polarisCallContext = callContext.getPolarisCallContext();
     int batchSize = callContext.getRealmConfig().getConfig(TABLE_METADATA_CLEANUP_BATCH_SIZE);
 
-    // Stream all metadata files without materializing them all at once
     Iterator<String> metadataFiles =
         Stream.of(
                 tableMetadata.previousFiles().stream().map(TableMetadata.MetadataLogEntry::file),


runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
```diff
@@ -160,13 +213,14 @@ private Stream<TaskEntity> getManifestTaskStream(
...
         .flatMap(sn -> sn.allManifests(fileIO).stream())
-        // distinct by manifest path, since multiple snapshots will contain the same
-        // manifest
-        .collect(Collectors.toMap(ManifestFile::path, Function.identity(), (mf1, mf2) -> mf1))
-        .values()
-        .stream()
+        // distinct by manifest path, since multiple snapshots will contain the same manifest
+        // Use stateful filter to dedupe while streaming
+        .filter(mf -> seenPaths.add(mf.path()))
         .filter(mf -> TaskUtils.exists(mf.path(), fileIO))
```
singhpk234, line 221:
```
  Set<String> uniquePaths = tableMetadata.snapshots().stream()
      .flatMap(sn -> sn.allManifests(fileIO).stream())
      .map(ManifestFile::path)
      .collect(Collectors.toSet());

  return uniquePaths.parallelStream()  // Parallel here!
      .filter(mf -> TaskUtils.exists(mf.path(), fileIO))
      .map(mf -> createManifestTask(...));
```

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
```diff
@@ -150,6 +156,53 @@ public boolean handleTask(TaskEntity cleanupTask, CallContext callContext) {
...
+        batch.clear();
+      }
+    }
+
+    // Create remaining tasks
+    if (!batch.isEmpty()) {
+      createAndRegisterTasks(batch, metaStoreManager, polarisCallContext, tableEntity);
+      totalCount += batch.size();
+    }
+
```
singhpk234, line 184: can explicitly call batch.clear ?

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
flyrain: Once we call `.collect(Collectors.toSet()`, the stream is fully materialized, which will lose the benefit of lazy execution. Here we are trying lower the memory footprint based on lazy execution.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
flyrain: We could, but we don't have to, as this is the last batch.

runtime/service/src/test/java/org/apache/polaris/service/task/TableCleanupTaskHandlerTest.java
```diff
@@ -434,6 +435,272 @@ public void testTableCleanupMultipleSnapshots() throws IOException {
...
+                    .returns(
+                        ManifestFileCleanupTaskHandler.ManifestCleanupTask.buildFrom(
+                            tableIdentifier, uniqueManifest),
+                        entity ->
+                            entity.readData(
+                                ManifestFileCleanupTaskHandler.ManifestCleanupTask.class)));
+  }
+
+  @Test
+  public void testMetadataFileBatchingWithManyFiles() throws IOException {
```
pingtimeout, line 616: This test is named `testMetadataFileBatchingWithManyFiles` but only creates 24 files in total.  Unfortunately that does not prove that the code is better at handling large tables.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
```diff
@@ -160,13 +213,14 @@ private Stream<TaskEntity> getManifestTaskStream(
...
     return tableMetadata.snapshots().stream()
         .flatMap(sn -> sn.allManifests(fileIO).stream())
-        // distinct by manifest path, since multiple snapshots will contain the same
-        // manifest
-        .collect(Collectors.toMap(ManifestFile::path, Function.identity(), (mf1, mf2) -> mf1))
-        .values()
-        .stream()
+        // distinct by manifest path, since multiple snapshots will contain the same manifest
+        // Use stateful filter to dedupe while streaming
+        .filter(mf -> seenPaths.add(mf.path()))
```
pingtimeout, line 220: This line adds all unique manifest files across all snapshots to a set that is maintained in memory.  Even though the stream is lazy, all unique manifest paths are materialized on the heap.  This means that the space complexity does not change.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
```diff
@@ -209,7 +263,28 @@ private Stream<TaskEntity> getMetadataTaskStream(
       CallContext callContext) {
     PolarisCallContext polarisCallContext = callContext.getPolarisCallContext();
     int batchSize = callContext.getRealmConfig().getConfig(BATCH_SIZE_CONFIG_KEY, 10);
-    return getMetadataFileBatches(tableMetadata, batchSize).stream()
+
+    // Stream all metadata files without materializing them all at once
```
pingtimeout: The only thing that this change does it to postpone the call to the `.map(...)` methods, but afaict the memory consumption stays identical.

dimas-b:
@pingtimeout : what is your take on this PR?

pingtimeout:
@dimas-b This PR is very confusing to me as after review, I do not think it fixes anything at all...

pingtimeout (changes requested):
Thanks @flyrain for the attempt at fixing the high space complexity issue.  This is a good start, but I don't think we are quite there yet.

As far as I can tell, the space complexity of table cleanup was `O(UM + PM + S + ST + PST + T)`.  And with this change, it is `O(UM + PM + S + ST + PST + batchSize)` where:
* `PM` = number of previous metadata files
* `S` = number of snapshots
* `ST` = number of statistics files
* `PST` = number of partition statistics files
* `UM` = number of unique manifest files across all snapshots
* `T` = total number of created TaskEntities

You can see that by running the code with large number of files under constrained memory.  You will see that with the current code, there is always a number of files that results in an OOME, proving that the space complexity issue has not been solved by the change.  You may want to use realistic (longer) paths to surface the issue faster.

I want to emphasize one critical point that must be addressed before this PR is merged.  In #3256, you said the following:

> please take a look to see if that solves the problem. It'd be really nice to run this with the same setup we used to validate the current PR which is this PR fixed the issue

Which contradicts the box that you checked in the description of this PR: `Added/updated tests with good coverage, or manually tested (and explained how)`.  Were you able to reproduce the issue before attempting to write a fix?

To summarize: based on my review of the code, I am convinced that this does not solve the underlying issue.  And based on the lack of testing, I do not think this PR is ready.  I appreciate the desire to provide an alternative to #3256.  But I think #3256 is the best option we have, all things considered.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
flyrain: Thanks for the detailed analysis. I agree that the only remaining unbounded structure here is the in memory set used to dedup manifest paths. I do not think this is a practical concern.

To put concrete numbers on it, with an extreme case that 1 million file paths and an estimated 50 to 100 bytes per path including object and set overhead, the memory footprint would be roughly 40 MB to 95 MB, which is acceptable. That is already a very large table cleanup scenario. At that scale, the question becomes whether we even want the Polaris server itself to handle such a task synchronously in memory. A delegation service would fit better in that case.

runtime/service/src/test/java/org/apache/polaris/service/task/TableCleanupTaskHandlerTest.java
flyrain: The intent of this unit test is not to simulate a truly large table, but to validate the batching behavior and correctness when metadata files are processed incrementally. As is common practice, we avoid stress or scale tests in unit tests, since they would significantly slow down CI execution and are better suited for dedicated benchmark.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
flyrain: The main change is that `stream().toList()` has been removed to avoid fully materializing the results in memory. Instead, an iterator is used together with a configurable batch size (`taskPersistenceBatchSize`) to read and process items incrementally. This bounds memory usage, as shown in lines 169 to 175.

runtime/service/src/test/java/org/apache/polaris/service/task/TableCleanupTaskHandlerTest.java
pingtimeout: I understand that unit tests should be quick to avoid slowing down CI.  My main concern here is whether this code change has been tested at scale.  And if so, how?

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
pingtimeout: The parameters of the stream construction are eager, so I am afraid the only thing lazily evaluated here is the call to `.flatMap(Function.identity())`

runtime/service/src/test/java/org/apache/polaris/service/task/TableCleanupTaskHandlerTest.java
flyrain: Thanks for bringing it up. I think it's a good idea to have a benchmark, more details are here, https://github.com/apache/polaris/pull/3256#issuecomment-3775726356.

runtime/service/src/main/java/org/apache/polaris/service/task/TableCleanupTaskHandler.java
flyrain: The comment is misleading, removed. However, all file paths here is part of `metadata.json` file, we've loaded the `matadata.json` file as a table metadata to memory already. Applying lazy evaluation doesn't make sense here.

PR #3423, opened 2026-01-12
[Minor] [Site] fix scheduled meetings table

danicafine:
Super minor change to the site. Noticed that the "Scheduled Meetings" table was formatted incorrectly. 
Screenshot 2026-01-12 at 3 29 06 PM

Updated it to add a new column, aligned the time format to 24h, and moved the 01-08 row to the past meetings table to await recording and notes links.
Screenshot 2026-01-12 at 3 37 28 PM


dimas-b:
Thanks for the fix, @danicafine !

singhpk234 (approved):

PR #499, opened 2024-11-29
Rework PolarisAuthorizer to use self-contained manifest on input

dimas-b:
Previously authZ checks were done based on a complex set of parameters provided individually. This PR consolidates authZ inputs into the PolarisResolutionManifest with declarative "selector" objects depending on operation context.

Naturally, Principal entity resolution is moved from the request authentication phase to authorization phase.

Request authentication merely validates the provided principal ID and/or name, but does not immediately access related entities from storage.

An anonymous authenticator and corresponding "allow all" authorizer are added as simple examples the separation of authN and authZ duties.

Also, application config is updated to allow different authorizer implementations to be chosen in runtime.

### Has This Been Tested?

Unit / Integration tests

PR #3427, opened 2026-01-13
core: introduce `resolvePathsOnly()` in `Resolver` and `PolarisResolutionManifest`

sungwy:
Introduce path‑only resolution API in the resolver stack without changing handler or authorizer behavior. While working on a PR to move the principal, roles and entity resolution logic into the `PolarisAuthorizerImpl`, I learned that resolution is done for a few reasons:
1. Materializing the concrete entity/ID and hierarchy needed for execution (e.g., delete/rename/move, policy attach/detach, create under namespace).
2. Existence checks (return not‑found early, avoid operating on missing entities).
3. Providing resolved targets to the authorizer (entity IDs, types, and parent relationships) so auth decisions are based on authoritative state.

While (2) and (3) can be moved into the `PolarisAuthorizer` in line with the suggestion discussed in this [PR comment](https://github.com/apache/polaris/pull/3228#issuecomment-3661802316), (1) needs to still be executed in the Handlers after a successful authorization.

Hence, introducing this path-only resolution API will allow a follow‑up auth refactor to decouple execution‑time path resolution from (2) and (3) for authorizable actions that are only require entity or path resolution, without attempting to resolve the principal, or associated roles.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1180,7 +1180,7 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
         ns ->
             resolutionManifest.addPath(
                 new ResolverPath(Arrays.asList(ns.levels()), PolarisEntityType.NAMESPACE), ns));
-    ResolverStatus status = resolutionManifest.resolveAll();
+    ResolverStatus status = resolutionManifest.resolvePathsOnly();
     if (!status.getStatus().equals(ResolverStatus.StatusEnum.SUCCESS)) {
       String message =
           "Unable to resolve sibling entities to validate location - " + status.getStatus();


sungwy:
Hi @adutra and @dimas-b - it took a minute for me to wrap my head around the PolarisResolutionManifest. Here’s a small PR to inch this auth refactoring forward. Could I ask for your help in reviewing if you approve of this direction for refactoring auth?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -143,6 +143,16 @@ public ResolverStatus resolveAll() {
     return primaryResolverStatus;
   }
 
+  /**
+   * Resolve only reference catalog, entities, and paths without resolving caller principal or
+   * roles. Returns SUCCESS, PATH_COULD_NOT_BE_FULLY_RESOLVED, or ENTITY_COULD_NOT_BE_RESOLVED and
+   * never returns CALLER_PRINCIPAL_DOES_NOT_EXIST.
+   */
+  public ResolverStatus resolvePathsOnly() {
```
dimas-b, line 151: nit: the method name sounds a bit obscure to me, TBH 😅 How about `resolveNonIamEntities` or `resolveResourceEntities` or `resolveCatalogEntities`? I personally prefer the latter, but I suppose it may still be confusing wrt "Catalog Roles". WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -54,6 +54,7 @@
  * incoming rest request, Once resolved, the request can be authorized.
  */
 public class Resolver {
+  private static final int MAX_RESOLVE_PASSES = 1000;
```
dimas-b, line 57: +1000 :)

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -447,6 +484,57 @@ private ResolverStatus runResolvePass() {
     return validationSuccess ? status : null;
   }
 
+  /**
+   * Execute one resolve pass on all entities without resolving caller principal or roles.
+   *
+   * @return status of the resolve pass
+   */
+  private ResolverStatus runResolvePassPathsOnly() {
```
dimas-b, line 492: WDYT about making a `Resolver` sub-class for the new logic?

dimas-b:
@sungwy : how to you envision choosing which `Resolver` method to use in runtime?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
sungwy: Hi @dimas-b - by that are you suggesting that that the new `Resolver` sub-class should be chosen over the existing `Resolver` through a runtime configuration?

sungwy:
> @sungwy : how to you envision choosing which `Resolver` method to use in runtime?

Good question @dimas-b.

Today we effectively hard-code the resolver choice in handlers. There are two call sites:

* **Handlers** (`CatalogHandler`, `PolarisAdminService`)
  * resolve for **existence checks** before authorization
  * resolve again to **fetch entities for execution** after authorization
* **Authorizers** (`PolarisAuthorizerImpl`, `OpaPolarisAuthorizer`)
  * implicitly depend on fully resolved entities to make authorization decisions

If we introduce a new `PolarisAuthorizer` API that accepts *unresolved* `AuthorizationTarget`s and move **existence checks into the Authorizer**, then:
* the **Authorizer** decides whether resolution is needed at all for that callsite, and which entities need resolution
* unsupported actions can fail fast (e.g. `PrincipalRole` creation in `OpaPolarisAuthorizer`) without resolution
* non Polaris-RBAC dependent authorizers can skip RBAC-entity resolution entirely

**Before refactor** (handlers always resolve eagerly):

| Callsite | Existence check (Handler) | Execution fetch (Handler) |
|---------|----------------------------|----------------------------|
| PolarisAuthorizerImpl – RBAC | `resolveAll()` | `resolveAll()` |
| PolarisAuthorizerImpl – Catalog | `resolveAll()` | `resolveAll()` |
| OpaPolarisAuthorizer – RBAC | `resolveAll()` | `resolveAll()` |
| OpaPolarisAuthorizer – Catalog | `resolveAll()` | `resolveAll()` |

**After refactor** (authorizer controls resolution):

| Callsite | Existence check (Authorizer) | Execution fetch (Handler) |
|---------|-------------------------------|----------------------------|
| PolarisAuthorizerImpl – RBAC | `resolveAll()` | `resolveAll()` |
| PolarisAuthorizerImpl – Catalog | `resolvePathsOnly()` | `resolvePathsOnly()` |
| OpaPolarisAuthorizer – RBAC | **throw** (unsupported) | *skipped* |
| OpaPolarisAuthorizer – Catalog | `resolvePathsOnly()` | `resolvePathsOnly()` |


In summary: by moving existence checks into the Authorizer and standardizing catalog call sites on `resolvePathsOnly()`:
* `OpaPolarisAuthorizer` can remain truly non-RBAC-dependent
* unsupported actions fail early, before any metastore lookups
* handlers do not need to understand authorization-specific resolution semantics

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
sungwy: Hmmm I agree that I'm not a fan of the current name, nor these options :)

But naming is important! Let me give this a bit more thought while we continue the review

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
dimas-b: Yes, something like that... but it depends on how / where we actually change to the new call path... which is not obvious in this PR :thinking:

dimas-b:
@sungwy : thanks for the refactoring proposal... 

I need some more time to think about it :sweat_smile: Ideally, I'd like the resolver algorithm to be tunable per persistence impl. Some efficiencies can be found in the relatively new NoSQL impl. (e.g. the hierarchical ID-based lookup via the parents trail is probably not required, paths can be resolved directly).

If other people have comments, I'd be interested to know their opinions too.

sungwy:
> @sungwy : thanks for the refactoring proposal... 
> 
> 
> 
> I need some more time to think about it :sweat_smile: Ideally, I'd like the resolver algorithm to be tunable per persistence impl. Some efficiencies can be found in the relatively new NoSQL impl. (e.g. the hierarchical ID-based lookup via the parents trail is probably not required, paths can be resolved directly).
> 
> 
> 
> If other people have comments, I'd be interested to know their opinions too.

I tried to convey the idea as best as possible here, but I think a proper RFC would be best given the scope of the proposal. I'm working on one, and I'll send it out on the mailing list once I clarify the doc with mentions of specific call sites I'm thinking of updating

sungwy:
> Ideally, I'd like the resolver algorithm to be tunable per persistence impl. Some efficiencies can be found in the relatively new NoSQL impl. (e.g. the hierarchical ID-based lookup via the parents trail is probably not required, paths can be resolved directly).

This is not a requirement I have been thinking of so far... how do you envision the relationship of persistence impl <-> resolver impl <-> auth impl to be in an ideal world?

So far, I've only been thinking of a world where auth impl would affect the resolver's behavior

dimas-b:
> how do you envision the relationship of persistence impl <-> resolver impl <-> auth impl to be in an ideal world?

Rough sketch :)

* REST API provides initial "seeds" (generally catalog paths or names) for the Resolver
* Resolver "outputs" are Entities found from those "seeds".
* The resolution algorithm is specific to the Persistence impl. (this is where Resolver sub-classes may be needed as I commented above). The one for JDBC uses current parent ID-based loop. The one for NoSQL uses direct lookup by path/name (this one may be implemented later).
* One of the "seeds" is the Principal + Role Names (provided by existing AuthN code)
* Whether Principal + Roles needs to be resolved to Entities is indicated by the Authorizer impl. This part is common to all Resolvers and can probably be done in the common base class.
* Ideally there is only one resolution "action" - it produces the complete data set for the service endpoint to perform AuthZ and actual API impl. logic. Some early abandoned ideas can be found in #499, but we do not have to do it exactly that way.

WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -812,14 +900,10 @@ private List<ResolvedPolarisEntity> resolvePrincipalRolesByName(
    */
   private ResolverStatus resolveReferenceCatalog(
```
adutra, line 901: nit: I'd prefer to have this method renamed to `resolveReferenceCatalogAndRoles` and the one below to `resolveReferenceCatalog`.

dimas-b:
> Your work on external principals in https://github.com/apache/polaris/pull/3250 is imho relevant [...]

Good idea about using the Principal's `isExternal` / `isFederated` / `type` property to make the decision whether to resolve it to a Polaris entity (same for "active" roles).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1180,7 +1180,7 @@ private <T extends PolarisEntity & LocationBasedEntity> void validateNoLocationO
         ns ->
             resolutionManifest.addPath(
                 new ResolverPath(Arrays.asList(ns.levels()), PolarisEntityType.NAMESPACE), ns));
-    ResolverStatus status = resolutionManifest.resolveAll();
+    ResolverStatus status = resolutionManifest.resolvePathsOnly();
```
flyrain, line 1183: Do we plan to make the same changes for other classes, like `Cataloghanlder`, `PolicyCatalogHandler`?

sungwy:
Refactoring is currently in progress and is being tracked on: https://github.com/apache/polaris/issues/3779

PR #3433, opened 2026-01-13
Fix the silently dropped fractional token

flyrain:
The old implementation silently dropped fractional token accrual. That means any rate below 1000 tokens/sec could never refill the bucket: each call adds <1 token, cast to long becomes 0, so tokens stay at 0 forever. Switching to double preserves fractional accumulation so the bucket refills correctly over time.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
@@ -25,20 +25,29 @@ import java.time.InstantSource;
  * amount of tokens. Each successful "tryAcquire" costs 1 token.
  */
 public class TokenBucket {
-  private final double tokensPerMilli;
-  private final long maxTokens;
+  /** Conversion factor: 1 token = 1000 milli-tokens. */
+  private static final long MILLI_TOKENS_PER_TOKEN = 1000L;
+
+  private static final long MAX_TOKENS_LIMIT = Long.MAX_VALUE / MILLI_TOKENS_PER_TOKEN;
+
+  private final long tokensPerSecond;
+  private final long maxMilliTokens;
   private final InstantSource instantSource;
 
-  private double tokens;
-  private long lastTokenGenerationMillis;
+  private long milliTokens;
+  private long lastUpdateMillis;
 
   public TokenBucket(long tokensPerSecond, long maxTokens, InstantSource instantSource) {
-    this.tokensPerMilli = tokensPerSecond / 1000D;
-    this.maxTokens = maxTokens;
+    if (maxTokens > MAX_TOKENS_LIMIT) {
+      throw new IllegalArgumentException(
+          "maxTokens must be <= " + MAX_TOKENS_LIMIT + " to avoid overflow");
+    }
+    this.tokensPerSecond = tokensPerSecond;
+    this.maxMilliTokens = maxTokens * MILLI_TOKENS_PER_TOKEN;
     this.instantSource = instantSource;
 
-    tokens = maxTokens;
-    lastTokenGenerationMillis = instantSource.millis();
+    milliTokens = maxMilliTokens;
+    lastUpdateMillis = instantSource.millis();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
```diff
@@ -51,11 +51,11 @@ public synchronized boolean tryAcquire() {
     long t = instantSource.millis();
     long millisPassed = Math.subtractExact(t, lastTokenGenerationMillis);
     lastTokenGenerationMillis = t;
-    tokens = Math.min(maxTokens, tokens + ((long) (millisPassed * tokensPerMilli)));
+    tokens = Math.min((double) maxTokens, tokens + (millisPassed * tokensPerMilli));
```
dimas-b: I believe a more robust fix would be to count the elapsed time in longer periods rather than computing small fractional increments. Note that floating point addition may not increase the value if given sufficient difference in exponents.

```
shell> double x = 1e100;
x ==> 1.0E100

jshell> x + 0.0000001
$2 ==> 1.0E100
```

dimas-b (approved):
Thanks for the fix, @flyrain 👍 

I think this PR is an improvement, but an even more robust solution is probably possible.

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
flyrain: That's a valid point, but only happen in an extreme case. For IEEE 754 double precision, the relative epsilon is ~2.2e-16. The issue would occur when:  `tokens + increment == tokens`, which happens when increment < tokens * epsilon

For example:
  - If maxTokens = 1e15 and tokensPerMilli = 0.001, the increment would be lost
  - But for practical values like maxTokens = 10000, adding 0.001 will always register

The Java long can reach up to 1e13, which is at the edge of losing the increment. Mostly likely it won't happen, users may just disable the rate limiter in that case.

flyrain: I have changed it to use long only, please take another look.

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
```diff
@@ -25,20 +25,29 @@
...
+      throw new IllegalArgumentException(
+          "maxTokens must be <= " + MAX_TOKENS_LIMIT + " to avoid overflow");
+    }
+    this.tokensPerSecond = tokensPerSecond;
+    this.maxMilliTokens = maxTokens * MILLI_TOKENS_PER_TOKEN;
     this.instantSource = instantSource;
 
-    tokens = maxTokens;
-    lastTokenGenerationMillis = instantSource.millis();
+    milliTokens = maxMilliTokens;
```
dimas-b, line 49: nit: why not `this.***` as with the above assignments?

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
```diff
@@ -47,15 +56,22 @@ public TokenBucket(long tokensPerSecond, long maxTokens, InstantSource instantSo
...
   public synchronized boolean tryAcquire() {
-    // Grant tokens for the time that has passed since our last tryAcquire()
     long t = instantSource.millis();
-    long millisPassed = Math.subtractExact(t, lastTokenGenerationMillis);
-    lastTokenGenerationMillis = t;
-    tokens = Math.min(maxTokens, tokens + ((long) (millisPassed * tokensPerMilli)));
+    long millisPassed = Math.subtractExact(t, lastUpdateMillis);
+    lastUpdateMillis = t;
+
+    long grant = millisPassed * tokensPerSecond;
```
dimas-b: not sure I understand the logic here... why multiply "millis passed" by something "per second" (not per milli)?... Should `grant` be `milliGrant`?

runtime/service/src/test/java/org/apache/polaris/service/ratelimiter/TokenBucketTest.java
```diff
@@ -93,6 +93,46 @@ void testConcurrent() throws InterruptedException {
...
+
+    for (int i = 0; i < 9; i++) {
+      clock.add(Duration.ofMillis(100));
+      assertCannotAcquire(tokenBucket);
+    }
+
+    // After 900ms total, we should have 0.9 tokens - not enough yet
+    // Add 100ms more to reach exactly 1 token
+    clock.add(Duration.ofMillis(100));
+    Assertions.assertTrue(tokenBucket.tryAcquire());
```
dimas-b, line 112: nit: why not `assertCanAcquire(tokenBucket, 1)` as above?

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
```diff
@@ -47,15 +56,22 @@ public TokenBucket(long tokensPerSecond, long maxTokens, InstantSource instantSo
...
+    if (tokensPerSecond != 0 && grant / tokensPerSecond != millisPassed) {
+      // Overflow occurred. If this much time passed, the bucket would be full anyway.
+      milliTokens = maxMilliTokens;
```
dimas-b, line 66: This case is _not_ tested by `TokenBucketTest`. Is it necessary?

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucket.java
```diff
@@ -47,15 +56,22 @@ public TokenBucket(long tokensPerSecond, long maxTokens, InstantSource instantSo
    * @return whether a token was successfully acquired and spent
    */
   public synchronized boolean tryAcquire() {
-    // Grant tokens for the time that has passed since our last tryAcquire()
     long t = instantSource.millis();
-    long millisPassed = Math.subtractExact(t, lastTokenGenerationMillis);
-    lastTokenGenerationMillis = t;
-    tokens = Math.min(maxTokens, tokens + ((long) (millisPassed * tokensPerMilli)));
+    long millisPassed = Math.subtractExact(t, lastUpdateMillis);
+    lastUpdateMillis = t;
```
dimas-b, line 61: IMHO, it would be more robust (and easier to understand) to advance `lastUpdateMillis` only when `millisPassed` is large enough to grant at least one token. We'll move `lastUpdateMillis` only to the point that exactly matches the boundary that results in that exact grant. This way we can count exact tokens (not milli tokens), which increases the operational range of token values (without overflow).

dimas-b:
`TokenBucket` switched to Guava's `RateLimiter` in #3507

PR #3442, opened 2026-01-14
Create PolarisEvent only when event type is actually supported by listener

olsoloviov:
This PR implements @adutra's idea from his comment in #3293: https://github.com/apache/polaris/pull/3293#issuecomment-3737852429.
I used a bit different approach with Supplier instead of conditional calls to OnEvent(). Polling the listener before emitting the event does not align well with future plans to support multiple active event listeners. 

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisEventListener.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisEventListener.java
@@ -18,14 +18,16 @@
  */
 package org.apache.polaris.service.events.listeners;
 
+import java.util.function.Supplier;
 import org.apache.polaris.service.events.PolarisEvent;
+import org.apache.polaris.service.events.PolarisEventType;
 
 /**
- * Event listener that responds to notable moments during Polaris's execution. Implementations can
- * filter events by checking {@link PolarisEvent#type()} or by querying attributes with {@link
- * PolarisEvent#hasAttribute} and {@link PolarisEvent#attribute}.
+ * Event listener that responds to notable moments during Polaris's execution. Implementations
+ * should check the event type and only call {@code eventSupplier.get()} for types they handle,
+ * avoiding unnecessary object allocation.
  */
 public interface PolarisEventListener {
 
-  default void onEvent(PolarisEvent event) {}
+  default void onEvent(PolarisEventType type, Supplier<PolarisEvent> eventSupplier) {}
 }
--- a/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
@@ -21,6 +21,7 @@ package org.apache.polaris.service.events.listeners;
 
 import com.google.common.collect.ImmutableMap;
 import java.util.Map;
+import java.util.function.Supplier;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableMetadataParser;
 import org.apache.iceberg.catalog.Namespace;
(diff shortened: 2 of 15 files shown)


runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisPersistenceEventListener.java
```diff
@@ -30,17 +31,16 @@
...
-      case AFTER_CREATE_TABLE -> handleAfterCreateTable(event);
-      case AFTER_CREATE_CATALOG -> handleAfterCreateCatalog(event);
-      default -> {
-        // Other events not handled by this listener
-      }
+  public void onEvent(PolarisEventType type, Supplier<PolarisEvent> eventSupplier) {
+    switch (type) {
+      case AFTER_CREATE_TABLE -> handleAfterCreateTable(eventSupplier.get());
+      case AFTER_CREATE_CATALOG -> handleAfterCreateCatalog(eventSupplier.get());
+      default -> {}
```
adutra, line 43: Maybe keep the comment?

runtime/service/src/main/java/org/apache/polaris/service/events/listeners/PolarisEventListener.java
```diff
@@ -18,14 +18,16 @@
...
- * filter events by checking {@link PolarisEvent#type()} or by querying attributes with {@link
- * PolarisEvent#hasAttribute} and {@link PolarisEvent#attribute}.
+ * Event listener that responds to notable moments during Polaris's execution. Implementations
+ * should check the event type and only call {@code eventSupplier.get()} for types they handle,
+ * avoiding unnecessary object allocation.
  */
 public interface PolarisEventListener {
 
-  default void onEvent(PolarisEvent event) {}
+  default void onEvent(PolarisEventType type, Supplier<PolarisEvent> eventSupplier) {}
```
adutra, line 32: Unfortunately this is not binary-compatible. We would need to keep the old method around. e.g.

```java
default void onEvent(PolarisEventType type, Supplier<PolarisEvent> supplier) { onEvent(supplier.get()); }

@Deprecated
default void onEvent(PolarisEventType type, Supplier<PolarisEvent> eventSupplier) {}
```

olsoloviov: Ok, we can do that. But I supposed that we could change the SPI while the feature is still in the experimental phase

adutra: We *probably* can, but it's always better to remain binary-compatible when it's possible.

olsoloviov:
Thanks for reviewing, @adutra!
I was thinking about using Suppliers.memoize() in case of multiple listeners, but the approach you proposed looks more effective. 
That said, I think we were considering to use Quarkus event bus for switching to multiple listeners approach. It looks like both approaches will not be compatible with it. We will not be able to enumerate listeners (unless we add some additional mechanisms for that), and messages will be serialized, so Supplier is  not an options as well. WDYT?

adutra:
> We will not be able to enumerate listeners (unless we add some additional mechanisms for that)

The introduction of an event bus is not incompatible with the idea of only firing the event if there is one interested listener. My `PolarisEventDispatcher` above could be rewritten as follows:

```java
@ApplicationScoped
public class PolarisEventDispatcher {

  private final EventBus eventBus;
  private final EnumSet<PolarisEventType> supportedTypes;

  @Inject
  public PolarisEventDispatcher(EventBus eventBus, @Any Instance<PolarisEventListener> listeners) {
    this.eventBus = eventBus;
    this.supportedTypes = EnumSet.noneOf(PolarisEventType.class);
    for (PolarisEventListener listener : listeners) {
      this.supportedTypes.addAll(listener.supportedEventTypes());
    }
  }

  public boolean hasListeners(PolarisEventType type) {
    return supportedTypes.contains(type);
  }

  public void fireEvent(PolarisEvent event) {
    eventBus.publish(event.type().name(), event);
  }
}
```

> messages will be serialized, so Supplier is not an options as well.

That is partially true :-) In general, as soon as a bus is involved, you are right to ask yourself the question of how the elements in the bus are serialized. 

The trick here though is that this bus is local to the Polaris instance, so there is no serialization at all. That's why I didn't mention serialization in my review. (See https://quarkus.io/guides/reactive-event-bus#using-codecs for more).

adutra:
Removing the stale label for now. I think this change is worth having.

nandorKollar:
@olsoloviov may I ask you, are you still working on this enhancement? I think this would be a useful modification, I can help out with moving this approach further by introducing Quarkus event bus with a `PolarisEventDispatcher`, as recommended by @adutra

dimas-b:
@nandorKollar : if you feel like carrying this forward, I think it should be fine if you opened a fresh PR based on @olsoloviov 's work and mentioned @olsoloviov as a [Co-Author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)

adutra:
Are we OK to close this PR? I think the work is now being done in #4064.

Issue #3440, opened 2026-01-14
STS inline policy includes KMS actions even when KMS is not configured


netapp-acheng:
### Describe the bug

I asked the question in https://github.com/apache/polaris/discussions/3434 and Dmitri B suggested me to report it as a bug.

When creating a Polaris S3 catalog that uses STS/AssumeRole against a non‑AWS S3‑compatible platform, table creation fails with:
Invalid action: kms:GenerateDataKeyWithoutPlaintext

Debug logs show that Polaris generates an inline STS policy that:

Contains KMS actions (e.g., kms:DescribeKey, kms:Decrypt, kms:GenerateDataKey, kms:GenerateDataKeyWithoutPlaintext) even when KMS is not configured for the catalog (i.e., allowedKmsKeys is empty). This appears to be a bug: if KMS is not configured, the IAM policy should not include KMS permissions.

Polaris Version
Server: 1.4.0-incubating-SNAPSHOT (local build under test)
CLI: matching snapshot
Quarkus: 3.28.2

Environment
Catalog type: INTERNAL
Storage type: S3
Endpoint: https://sgdemo.example.com 
Region: us-east-1 (dummy value for S3‑compatible)
STS endpoint: https://sgdemo.example.com
Auth: STS/AssumeRole 

Note: Creating catalog and tables work with --no-sts option in this same environment

### To Reproduce

Start Polaris server and authenticate the CLI.
Create an S3 catalog with STS enabled:
./polaris --host localhost --port 8181 --realm POLARIS --client-id root --client-secret <password> catalogs create sts_catalog --type INTERNAL --storage-type s3 --default-base-location s3://sts-polaris --endpoint https://sgdemo.example.com --region us-east-1 --allowed-location s3://sts-polaris --sts-endpoint https://sgdemo.example.com \  --role-arn arn:aws:iam::<account-id>:role/assumerole

The catalog created:
{
"type": "INTERNAL",
"name": "sts_catalog",
"properties": {
"default-base-location": "s3://sts-polaris"
},
"createTimestamp": 1766095794668,
"lastUpdateTimestamp": 1766095794668,
"entityVersion": 1,
"storageConfigInfo": {
"roleArn": "arn:aws:iam::123456789101112:role/assumerole",
"allowedKmsKeys": [],
"region": "us-east-1",
"endpoint": "https://sgdemo.example.com/",
"stsEndpoint": "https://sgdemo.example.com/",
"stsUnavailable": false,
"pathStyleAccess": false,
[...]

PR #3445, opened 2026-01-15
(WIP) Remove KMS policies when KMS is not configured

MonkeyCanCode:
This PR addressed issues reported in https://github.com/apache/polaris/issues/3440 where when end-user is not using KMS encryption for S3, Polaris still enforces in-lines policies which contains KMS related policies. While fixing this issue, I noticed our read-only policy for kMS is a bit too wide where `GenerateDataKey` and `GenerateDataKeyWithoutPlaintext` should be belongs to write operation instead of read. Thus, this PR also addresses this issue.

## Checklist

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/io/FileIOFactoryTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/io/FileIOFactoryTest.java
@@ -177,7 +177,8 @@ public class FileIOFactoryTest {
         .isInstanceOf(InMemoryFileIO.class);
 
     // 1. BasePolarisCatalog:doCommit: for writing the table during the creation
-    // 2. BasePolarisCatalog:doRefresh: for reading the table during the drop (2 calls: 1 to read, 1 to reload)
+    // 2. BasePolarisCatalog:doRefresh: for reading the table during the drop (2 calls: 1 to read, 1
+    // to reload)
     // 3. TaskFileIOSupplier:apply: for clean up metadata files and merge files
     Mockito.verify(testServices.fileIOFactory(), Mockito.times(4))
         .loadFileIO(Mockito.any(), Mockito.any(), Mockito.any());


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -297,42 +297,42 @@ private static void addKmsKeyPolicy(
...
 
     if (hasAllowedKeys) {
       addAllowedKmsKeyResources(allowedKmsKeys, allowKms);
-    }
-
-    // Add KMS statement if we have any KMS key configuration
-    if (hasCurrentKey || hasAllowedKeys) {
-      policyBuilder.addStatement(allowKms.build());
-    } else if (!canWrite) {
+    } else if (!canWrite && region != null && accountId != null) {
```
dimas-b: It looks like this `else` condition will now be evaluated when `hasCurrentKey == true` and `hasAllowedKeys == false`, but it was not evaluated in that case before 🤔 WDYT?

MonkeyCanCode: Thanks for the quick review @dimas-b and that is a fair point which I missed. It appears this is really specific for handling s3 and s3-compatible and I am not sure if the above logic will break or not due to lack of infra to test. I had made the needed code changes to preserve this workflow. Please take another look when you get a chance.

dimas-b (approved):

dimas-b:
CC: @fabio-rizzo-01

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b: thx - LGTM :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1409,6 +1409,22 @@ public void doRefresh() {
                       Set.of(PolarisStorageActions.READ, PolarisStorageActions.LIST));
               return TableMetadataParser.read(fileIO, metadataLocation);
             });
+
+        // After a refresh, re-load the FileIO with the new table metadata properties to
+        // ensure the right permissions are present for subsequent file system interactions.
+        if (currentMetadata != null) {
+          tableFileIO =
+              loadFileIOForTableLike(
```
dimas-b, line 1417: How is that related to KMS policies?

dimas-b: If it's just a parallel bugfix, I'd prefer to make it in a separate PR for the sake of clarity 🤔 WDYT?

MonkeyCanCode: Appear to be a parallel bug. I am waiting for testing from reporter then I can split this into two PRs. Currently if we merged current PR, it will trigger diff error from reporter.

dimas-b: sounds good - thx!

MonkeyCanCode:
Close this one and opened three different PRs for easier review:
1. https://github.com/apache/polaris/pull/3493
2. https://github.com/apache/polaris/pull/3494
3. https://github.com/apache/polaris/pull/3496

PR #3456, opened 2026-01-16
[Issue 3209] Surface Intermediate Data as part of Polaris Events

### LINKED ISSUE #3209: Implement sending POJOs not within the Response to Events

adnanhemani:
Adds a RequestScoped annotation to `AttributeMap`, which allows us to inject this bean into intermediary business logic and be able to patch data from there back out to the `PolarisEvent`.

@dimas-b had a great idea that we may be able to retire the Delegator model eventually by setting a "closing" hook on the RequestScoped bean - but for simplicity, I am introducing this smaller change that will solve the GH issue linked first and will investigate that approach later.

Closes #3209.

## Checklist
- [X] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #3209 
- [N/A] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -783,7 +787,14 @@ public Response commitTransaction(
...
                 .put(EventAttributes.COMMIT_TRANSACTION_REQUEST, commitTransactionRequest)));
-    for (UpdateTableRequest req : commitTransactionRequest.tableChanges()) {
+    List<LoadTableResponse> loadTableResponses =
+        attributeMap.getRequired(EventAttributes.LOAD_TABLE_RESPONSES);
+    for (int i = 0; i < commitTransactionRequest.tableChanges().size(); i++) {
+      UpdateTableRequest req = commitTransactionRequest.tableChanges().get(i);
+      LoadTableResponse loadTableResponse =
+          loadTableResponses != null && i < loadTableResponses.size()
+              ? loadTableResponses.get(i)
+              : null;
```
evindj, line 801: How does this code ensures that  map between loadTableResponses[i] maps to tableChanges of i ?
Is there a scenario where they might not match?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/CommitTransactionEventTest.java
```diff
@@ -117,6 +117,33 @@ void testEventsForUnSuccessfulTransaction() {
...
+    String table1Name = "test-table-5";
+    String table2Name = "test-table-6";
+    executeTransactionTest(false, table1Name, table2Name, testServices);
+
+    TestPolarisEventListener testEventListener =
+        (TestPolarisEventListener) testServices.polarisEventListener();
+
+    // Verify that AfterUpdateTable events contain LoadTableResponse objects
+    PolarisEvent afterUpdateTableEvent =
+        testEventListener.getLatest(PolarisEventType.AFTER_UPDATE_TABLE);
```
evindj, line 135: No assertion here?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/CommitTransactionEventTest.java
```diff
@@ -117,6 +117,33 @@ void testEventsForUnSuccessfulTransaction() {
...
+
+    // Verify second table's LoadTableResponse
```
evindj, line 137: probably a n00b question but my understanding was that two events will be generated if that is the case, should you also test the first table?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -183,6 +185,7 @@ public class IcebergCatalogHandler extends CatalogHandler implements AutoCloseab
   private final ReservedProperties reservedProperties;
   private final CatalogHandlerUtils catalogHandlerUtils;
   private final StorageAccessConfigProvider storageAccessConfigProvider;
+  private final AttributeMap attributeMap;
```
dimas-b: nit: maybe `EventAttributeMap`? When viewed in a context where the `import` statement is not visible, it is not obvious what the "attributes" are about 🤔

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributeMap.java
```diff
@@ -18,11 +18,13 @@
  */
 package org.apache.polaris.service.events;
 
+import jakarta.enterprise.context.RequestScoped;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
 /** A type-safe container for event attributes. This class is mutable and not thread-safe! */
+@RequestScoped
```
dimas-b, line 27: just thinking: `@RequestScoped` does not necessarily mean that the bean is accessed only by one thread.

That said, current usage seems to be ok wrt concurrent access.

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - (Before/After)UpdateTableEvent is emitted for all table updates within a transaction.
 - Added KMS options to Polaris CLI
 - Changed from Poetry to UV for Python package management 
+- Flattened Events hierarchy and introduced EventAttributes for all events. Additionally, `AttributeMap` is now exposed as a RequestScoped bean to allow intermediate event attributes to be added to events.
```
dimas-b: nit: `AttributeMap` is a low-level impl. detail... I'm not sure end user need to worry about it when reading the change log 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
dimas-b: Good point :+1: 

Since we already have `AttributeMap` as a context object for the whole multi-table transaction, I believe putting complete data relevant to events into it at the place of change (`IcebergCatalogHandler`) might be preferable.

Essentially, why bother with the not-so-imformative `LOAD_TABLE_RESPONSES` property, when we can have a list of "table change" objects inside `AttributeMap`?

In that regard, `AttributeMap` may become more of an "Catalog Operation Data Accumulator" with a richer API for dealing with one or more tables per operation. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributeMap.java
adutra: This is a good remark. Thanks to context propagation, the request context "follows" the execution flow, even if the thread changes. You usually wouldn't have race conditions, because two threads aren't accessing the bean _simultaneously_, even if different threads access it _successively_.

But we would enter a danger zone if/when we spawn threads manually (e.g. with the task executor), and pass this bean to more than one thread concurrently.

I don't think this is happening today, but it could happen in the future.

I therefore think we should either make this bean thread-safe, or explicitly call out in the class javadocs that it is not, and that it should not be passed to the task executor framework.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: Yes, this bean was private to the events package before. I agree that it has now a broader usage, it should be renamed.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adutra: Shouldn't we change all these `new AttributeMap()` occurrences to use the injected attribute map instead?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adutra: The list of table responses is constructed in `IcebergCatalogHandler`:

https://github.com/apache/polaris/blob/b820bd09db3f67dae9623b9c9f0f19c9387e66d5/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L1104

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1092,6 +1098,10 @@ public void commitTransaction(CommitTransactionRequest commitTransactionRequest)
               if (!updatedMetadata.changes().isEmpty()) {
                 tableOps.commit(currentMetadata, updatedMetadata);
               }
+
+              LoadTableResponse loadTableResponse =
+                  LoadTableResponse.builder().withTableMetadata(updatedMetadata).build();
```
adutra: We're creating a fake `LoadTableResponse` just for events. Couldn't we just store the `TableMetadata` object?

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributeMap.java
dimas-b: The nature of this class is to accumulate attributes as the request is processed (potentially by many threads). In light of that, from my POV it is preferable to make it thread-safe (which is not a huge task in this particular case, I hope).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: From my POV event payload does not have to match REST API payload. I actually think it is preferable for Polaris to define a separate API for event payloads because those are generated by Polaris, and generally follow the evolution of Polaris internal code rather than the evolution of the Iceberg REST API. For example, moving to a new REST API revision does not have to mean that all event payloads get adjustments.

Similarly if Polaris were to support multiple REST API revisions at the same time, it would be nice to have one stream of events with the same payloads.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: As Alex pointed out, the table responses are generated in the same order to the table changes. Therefore, they must be in the same order. 

@dimas-b, to confirm, are you suggesting that we add the TableChange objects to the AttributeMap so that we zip those up to the LoadTableResponses within the catalog handler? I'm not 100% sure what I understood  what the proposed action item is from your comment, so asking for some clarification.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/CommitTransactionEventTest.java
adnanhemani: This will auto-assert. See https://github.com/apache/polaris/blob/main/runtime/service/src/testFixtures/java/org/apache/polaris/service/events/listeners/TestPolarisEventListener.java#L39-L46

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/CommitTransactionEventTest.java
adnanhemani: A while back, to reduce heap pressure during testing, we no longer keep all events that have occurred. We can only see the latest event of each type that was emitted.

See [TestPolarisEventListener.java](https://github.com/apache/polaris/blob/main/runtime/service/src/testFixtures/java/org/apache/polaris/service/events/listeners/TestPolarisEventListener.java)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adnanhemani: Sounds good, I've done it in the latest revision - but it has added quite a few lines of changes since it's widespread in the Delegator classes FYI.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributeMap.java
adnanhemani: Given that the Javadoc already states that the class is not thread-safe, I'm leaning towards leaving things the way they are - and not passing the `AttributeMap` to the task executor framework in this PR. We should revisit this in a future PR where we start support passing this into tasks.

Let me know if there's a reason why we need to solve this issue within this PR itself.

CHANGELOG.md
adnanhemani: I added it here as it may be of interest to users who may be interested in creating their own custom event listener implementations. Happy to remove it.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: I had a different thought - but would like to hear your thoughts on this as well.

I was using `AttributeMap` to inject into the deeper parts of the codebase - but then using that to populate the final `AttributeMap` that gets passed to the Event Listener. The chief concern was to ensure that the `After` event attributes don't accidentally become strictly a super-set of the `Before` events attributes - and that cleaner attribute maps going into the event listener may help with filtering concerns that we will need to tackle in order to enable all events going to their eventual sink.

WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adnanhemani: Given that we will still be emitting an `AFTER_UPDATE_TABLE` event and (I'm assuming) that we should keep the same set of required attributes for this event type, are you both suggesting that we change the event's required attributes to take in the `TableMetadata` object instead?

(This is fine to me as well, but looking for confirmation before I make these changes.)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
evindj: > As Alex pointed out, the table responses are generated in the same order to the table changes. Therefore, they must be in the same order.

My concern is that it is an implicit assumption if one ever changes,  the other place will also have to change, and I don't think unit tests will catch this case.

dimas-b: The point I was trying to make above is that from my POV it is odd to have both `LOAD_TABLE_RESPONSE` and `LOAD_TABLE_RESPONSES` (plural) in `EventAttributeMap`.

If events may have more than one related table change in principle, I think it would be nicer to model `EventAttributeMap` according to be most general case (i.e. multiple tables). Then, we can add convenience methods for callers to operate on one table if they know from the event type that only one table is involved.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
```diff
@@ -174,6 +175,8 @@ private EventAttributes() {}
   // Transaction attributes
   public static final AttributeKey<CommitTransactionRequest> COMMIT_TRANSACTION_REQUEST =
       new AttributeKey<>("commit_transaction_request", CommitTransactionRequest.class);
+  public static final AttributeKey<List<LoadTableResponse>> LOAD_TABLE_RESPONSES =
```
dimas-b: Please move this definition closer to `LOAD_TABLE_RESPONSE` for ease of reading this code later, or at least cross-reference in javadoc.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: > My concern is that it is an implicit assumption if one ever changes, the other place will also have to change, and I don't think unit tests will catch this case.

For this implicit assumption to change, `tableChanges` would have to become an unordered collection. Given that it's coming from the Iceberg REST Spec, I don't think that's very likely at all. Happy to hear if you have a non-intrusive way to make a change to future-proof this.

> The point I was trying to make above is that from my POV it is odd to have both LOAD_TABLE_RESPONSE and LOAD_TABLE_RESPONSES (plural) in EventAttributeMap.

I see your point now. I'm not sure that I agree or disagree with this recommendation 100% - but I will implement for now and we can revisit if it ends up being an issue in the future.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adnanhemani: Changed.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
dimas-b: `EventAttributeMap` is a proper CDI bean now. It should be ok to rely on CDI to create and pass around (inject) one instance of it per request.

If we want to distinguish before/after data, we could do that by code / methods inside `EventAttributeMap`. As I commented in another thread, I see value in evolving this class to become a rich and typed container of event-related data for each request (more than a simple key-value map).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: > are you both suggesting that we change the event's required attributes to take in the TableMetadata object instead? 

Yes I was thinking about that 👍 Not sure about @dimas-b though 😅

dimas-b: `TableMetadata` as the attribute value SGTM :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: @dimas-b, if you're okay with it, let's handle this in a separate PR then? I think we may need to iterate on this idea a few times before we get it right - but this PR is attempting to do something very related but slightly orthogonal and IMO enough for a first take at this functionality.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adnanhemani: Done in the latest revision. Thanks both!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
dimas-b: I do not mind incremental work here, but please sync with @adutra in terms of timing.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
```diff
@@ -108,7 +110,9 @@ private EventAttributes() {}
   public static final AttributeKey<RenameTableRequest> RENAME_TABLE_REQUEST =
       new AttributeKey<>("rename_table_request", RenameTableRequest.class);
   public static final AttributeKey<LoadTableResponse> LOAD_TABLE_RESPONSE =
-      new AttributeKey<>("load_table_response", LoadTableResponse.class);
+      new AttributeKey<>("load_table_responses", LoadTableResponse.class);
```
adutra: The attribute name is `load_table_responses` in the plural, yet the attribute seems to target a single response object.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
```diff
@@ -108,7 +110,9 @@ private EventAttributes() {}
...
+  public static final AttributeKey<List<TableMetadata>> TABLE_METADATA =
+      new AttributeKey<>("table_metadatas", new TypeToken<List<TableMetadata>>() {});
```
adutra: The field name should reflect the attribute name: `TABLE_METADATAS`.

adutra: There seems to be an inconsistency here: sometimes this attribute seems to be legitimately collecting a list of metadatas, cf. `IcebergCatalogHandler`. But sometimes, it's collecting a singleton list of a unique table metadata, cf. `IcebergCatalogEventServiceDelegator`.

I think it would be better to distinguish two attributes: `TABLE_METADATA` and `TABLE_METADATAS`. (I understand that the latter may become a sort of "internal" attribute that is never exposed to listeners – we could maybe declare those separately if we are concerned about that.)

adutra: nit: `new TypeToken<>(){}` is enough:

```suggestion
  public static final AttributeKey<List<TableMetadata>> TABLE_METADATA =
      new AttributeKey<>("table_metadatas", new TypeToken<>() {});
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -345,15 +350,22 @@ public Response loadTable(
             snapshots,
             realmContext,
             securityContext);
+
+    EventAttributeMap generatedEventAttributes =
```
adutra: This change doesn't seem necessary?

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adnanhemani: Good catch, changed in the next revision.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adnanhemani: Sure.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
adnanhemani: Good call out, was a hangover needed for a previous revision.

runtime/service/src/main/java/org/apache/polaris/service/events/EventAttributes.java
adnanhemani: Sure, I'm glad to do this. I was trying to make everything fit within `TABLE_METADATAS` due to the similar comment that Dmitri had pointed out above. But I will make a comment for this for now so we can move past and revisit if there is a better way to do this later.

adnanhemani:
@adutra @dimas-b - I had to fix the merge conflicts :( Can you both please re-approve and merge?

adutra (approved):

Issue #3292, opened 2025-12-17
One load table API request results in two STS AssumeRole requests


dimas-b:
### Describe the bug

First STS call:
```
java.lang.Throwable: test1
	at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:733)
	at com.intellij.rt.debugger.MethodInvoker.invokeInternal(MethodInvoker.java:223)
	at com.intellij.rt.debugger.MethodInvoker.invoke1(MethodInvoker.java:35)
	at org.apache.polaris.core.storage.aws.AwsCredentialsStorageIntegration.getSubscopedCreds(AwsCredentialsStorageIntegration.java:96)
	at org.apache.polaris.core.persistence.AtomicOperationMetaStoreManager.getSubscopedCredsForEntity(AtomicOperationMetaStoreManager.java:1641)
	at org.apache.polaris.core.persistence.ServiceProducers_ProducerMethod_polarisMetaStoreManager_2QgveElh41S2GrgpagbE0iTcmng_ClientProxy.getSubscopedCredsForEntity(Unknown Source)
	at org.apache.polaris.core.storage.StorageCredentialsVendor.getSubscopedCredsForEntity(StorageCredentialsVendor.java:73)
	at org.apache.polaris.core.storage.ServiceProducers_ProducerMethod_storageCredentialsVendor_chZ10WlGqG-XefHi-NQqFKWM7fY_ClientProxy.getSubscopedCredsForEntity(Unknown Source)
	at org.apache.polaris.core.storage.cache.StorageCredentialCache.lambda$getOrGenerateSubScopeCreds$2(StorageCredentialCache.java:141)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$0(BoundedLocalCache.java:2707)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1916)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2705)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2686)
	at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:112)
	at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:63)
[...]

PR #3459, opened 2026-01-16
ensure sts is cached regardless of the refresh endpoint

evindj:
### Summary 
This change addresses issue #3292 
One load table API request was resulting in two STS AssumeRole requests because refreshCredentialsEndpoint was part of the cache key in  StorageCredentialCacheKey.java. During a loadTable operation, credentials were fetched twice:                                                                                  
  1. First call with refreshCredentialsEndpoint=Optional.empty() (server-side metadata read)                                                                 
  2. Second call with refreshCredentialsEndpoint=Optional.of(url) (building client response) 
Since refreshCredentialsEndpoint was part of the cache key, these two calls resulted in cache misses, causing duplicate STS calls. 

So the fix is to remove refreshCredentialsEndpoint as cache key. I did not find any evidence that we wanted an STS per refresh endpoint. 
### Test
Added unit test to show previous behavior, the test passed after the change was updated. 
- I setup an aws backend to see that the sts request was not called twice.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
@@ -51,9 +51,6 @@ public interface StorageCredentialCacheKey {
   Set<String> allowedWriteLocations();
 
   @Value.Parameter(order = 7)
-  Optional<String> refreshCredentialsEndpoint();
-
-  @Value.Parameter(order = 8)
   Optional<String> principalName();
 
   /**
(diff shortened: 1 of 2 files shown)


evindj:
cc @dimas-b

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCacheKey.java
```diff
@@ -84,7 +81,6 @@ static StorageCredentialCacheKey of(
         allowedListAction,
         allowedReadLocations,
         allowedWriteLocations,
-        refreshCredentialsEndpoint,
```
dimas-b, line 87: This certainly fixes the multiple STS calls, however the cache returns a `StorageAccessConfig`, which may indirectly contain the `refreshCredentialsEndpoint`. This change may cause cache hits to return `StorageAccessConfig` objects without or without the `refreshCredentialsEndpoint` that the caller does not expect. Note that `refreshCredentialsEndpoint` is a parameter to `StorageCredentialCache.getOrGenerateSubScopeCreds()` ... WDYT?

That said, #3292 is actually a pretty tricky problem to fix... I believe it requires substantial code refactoring 😅

evindj:
will rethink this a bit before moving forward

Issue #3444, opened 2026-01-14
Provide a robust CDI way to inject request_ids


obelix74:
### Is your feature request related to a problem? Please describe.

Currently, there is no robust way to inject request_id using CDI.  Approaches such as the following has been used in code.

https://github.com/apache/polaris/blob/64b13a9915274a04e82a12c9fdcd095d12797ca0/runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java#L111-L120

However, they only work for REST calls.

### Describe the solution you'd like

As suggested by @dimas-b in https://github.com/apache/polaris/pull/3414/changes/BASE..4b6234774c01f3197f2aa78e79dc5d5cbe488ba7#r2692150954, 

> Please see how TaskExecutorImpl deals with realm IDs. A similar CDI pattern is probably necessary for request 
IDs, except that we may not have to produce a request ID for background tasks. I do not have an opinion on whether request IDs should be propagated from REST requests to related async tasks or not. However, CDI must be solid and not cause runtime exceptions even if the request ID is not propagated.

Create a CDI pattern to inject request_ids.

### Describe alternatives you've considered

Approaches tried so far include using REST context or SLF4J logging MDC, but both have their shortcomings.

https://github.com/apache/polaris/blob/64b13a9915274a04e82a12c9fdcd095d12797ca0/runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java#L111-L120

flyingImer:
Rather than adding request ID propagation as another one-off pattern in `TaskExecutorImpl` (alongside the existing ad-hoc realm and principal propagation), I am thinking of the problem into a pluggable `AsyncContextPropagator` SPI:

```java
public interface AsyncContextPropagator {
    Object capture();             // snapshot on request thread
    AutoCloseable restore(Object); // re-establish on task thread + cleanup
}
```
Each propagator is responsible only for its own lifecycle — capture its own piece of context, restore it, and clean up after itself. The propagator does not need to know about other propagators, about TaskExecutorImpl's retry logic, or about CDI scope activation. TaskExecutorImpl handles orchestration (discovery via CDI Instance, LIFO cleanup ordering, error isolation); each propagator handles its own domain.

See my pr here: https://github.com/apache/polaris/pull/4061

PR #3468, opened 2026-01-17
feat(metrics): Evolve PolarisMetricsReporter interface with timestamp parameter and comprehensive documentation

obelix74:
This PR enhances the PolarisMetricsReporter SPI interface by adding a timestamp parameter to the reportMetric() method, enabling accurate time-series metrics reporting to external systems.

## Checklist


obelix74:
@dimas-b Created this PR for the event emission and added a feature flag for backwards compatibility.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -805,8 +810,41 @@ public Response reportMetrics(
...
       RealmContext realmContext,
       SecurityContext securityContext) {
-    return delegate.reportMetrics(
-        prefix, namespace, table, reportMetricsRequest, realmContext, securityContext);
+    // Check if metrics event emission is enabled
+    boolean metricsEventEmissionEnabled =
+        realmConfig.getConfig(FeatureConfiguration.ENABLE_METRICS_EVENT_EMISSION);
+
+    // If metrics event emission is disabled, call delegate directly without emitting events
+    if (!metricsEventEmissionEnabled) {
```
adutra: This is probably going to conflict a bit with #3442. The idea there is that events would only be emitted if there is interest for them from the listener(s).

I don't think creating a feature configuration for each event type would scale out well.

adutra:
> @dimas-b Created this PR for the event emission and added a feature flag for backwards compatibility.

I am not sure backwards compatibility is important here as currently no listener bundled with Polaris actually processes the metrics-related events.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
dimas-b: +1 to emitting events based on listener demand

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -805,8 +810,41 @@ public Response reportMetrics(
...
+      return delegate.reportMetrics(
+          prefix, namespace, table, reportMetricsRequest, realmContext, securityContext);
+    }
+
+    // Emit events when feature is enabled
+    String catalogName = prefixParser.prefixToCatalogName(realmContext, prefix);
+    Namespace namespaceObj = decodeNamespace(namespace);
+    polarisEventListener.onEvent(
+        new PolarisEvent(
+            PolarisEventType.BEFORE_REPORT_METRICS,
```
dimas-b: Apologies if I missed this in the design doc(s).

Emitting events with the full `reportMetricsRequest` object feels more like supporting a plugin point for custom event processors.

WDYT about abstracting our a proper metrics processor interface? The default one would persist metrics to the Polaris database. We could support delegation via CDI to allow custom processors to be invoked before/after the default one.

My view of Polaris Events is that they are a good vehicle for communicating with custom code about Polaris code processes, which are essentially controlled by Polaris (like table changes, etc.). In this case, it looks like we're merely passing REST API payloads to custom code transparently in a 1-to-1 manner, so going through events seems to be an unnecessary design complication.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+ *   public void process(MetricsProcessingContext context) {
+ *     // implementation
+ *   }
+ * }
+ * </pre>
+ *
+ * @see MetricsProcessingContext
+ * @see MetricsProcessorConfiguration
+ */
+public interface MetricsProcessor {
```
adutra: Is the idea that `MetricsProcessor` will _replace_ `PolarisMetricsReporter`? I confess I'm not seeing the value of keeping both 🤔

.polaris-work-notes.md
```diff
@@ -0,0 +1,18 @@
+# Polaris Work Notes
```
adutra: Accidentally committed?

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessor.java
obelix74: @adutra yeah, I think the `MetricsProcess` can replace `PolarisMetricsReporter`. We can either do it as part of this PR or future work. I was trying to not break existing code. I am writing a summary.

.polaris-work-notes.md
obelix74: Removed

obelix74:
@adutra and @dimas-b thanks for all the feedback.  Indeed, @singhpk234 also did not like the event based approach. I have made the following changes to address your feedback.

* Introduces a new MetricsProcessor interface with CDI support. This establishes the foundation for a simpler, more direct metrics processing architecture.  I have intentionally left out `PolarisMetricsReporter`.

* MetricsProcessingContext - Rich context object providing:
        • Catalog and table information
        • Realm ID and principal name
        • Request ID for correlation
        • OpenTelemetry trace ID and span ID
        • Timestamp
     • MetricsProcessorConfiguration - Type-safe configuration using Quarkus @ConfigMapping
     • CDI producer in ServiceProducers for MetricsProcessor selection via @Identifier annotations

The existing PolarisMetricsReporter interface and polaris.iceberg-metrics.reporting.type configuration remain unchanged and functional. 

**What's Not Included**

This PR intentionally excludes implementations to keep the scope focused:
     • No processor implementations (noop, logging, persistence)
     • No integration with IcebergCatalogAdapter
     • No backward compatibility adapters

   These will be added in subsequent PRs.

   Configuration

```
   polaris:
     metrics:
       processor:
         type: noop  # Default - implementations will be added in follow-up PRs
```
Next Steps

Follow-up PRs will add:
1. Built-in processor implementations (noop, logging, persistence)
2. Integration with the REST catalog adapter
3. Backward compatibility layer for existing PolarisMetricsReporter users

I am open to removing `PolarisMetricsReporter` - but need to figure out how to migrate existing code and users. In fact, I just upgraded my Polaris installation to `1.3.0` to use the metrics logging output while we sort out these.

obelix74:
Perhaps I should keep the `polaris.iceberg-metrics.reporting.type` property and not introduce a new property. That way we can delete any existing code and not break anyone else. I think I was trying to be too conservative in introducing a new property, but that adds more work.  @adutra what do you think?

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
```diff
@@ -0,0 +1,66 @@
...
+  String catalogName();
+
+  /** The table identifier */
+  TableIdentifier tableIdentifier();
+
+  /** The Iceberg metrics report (ScanReport or CommitReport) */
+  MetricsReport metricsReport();
+
+  /** The realm ID */
+  String realmId();
```
dimas-b: We already have `RealmContext` as a request-scoped CDI bean for conveying realm information. I'd prefer to avoid parallel contexts for the same data.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /** The catalog ID (internal entity ID) */
+  Optional<Long> catalogId();
+
+  /** The principal name who submitted the metrics */
+  Optional<String> principalName();
```
dimas-b: If processors require the principal, why would they not receive it via CDI? Cf. https://github.com/apache/polaris/blob/27ae684304c6c62e5ee516404eee1419bc5d67f8/runtime/service/src/main/java/org/apache/polaris/service/context/catalog/PolarisPrincipalHolder.java#L39

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /** The request ID for correlation */
+  Optional<String> requestId();
+
+  /** OpenTelemetry trace ID */
+  Optional<String> otelTraceId();
+
+  /** OpenTelemetry span ID */
+  Optional<String> otelSpanId();
```
dimas-b: I'd prefer to handle OTel via its natural context objects.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
```diff
@@ -0,0 +1,66 @@
...
+  String realmId();
+
+  /** The catalog ID (internal entity ID) */
+  Optional<Long> catalogId();
+
+  /** The principal name who submitted the metrics */
+  Optional<String> principalName();
+
+  /** The request ID for correlation */
+  Optional<String> requestId();
```
dimas-b: Requests IDs in Polaris are a bit special. I'm not sure a mainstream use case exists for them. In any case, I believe they should be handled via dedicated context objects similar to `PolarisPrincipalHolder`.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Process a metrics report with full context information.
+   *
+   * <p>Implementations should handle exceptions gracefully and not throw exceptions that would
+   * disrupt the metrics reporting flow. Errors should be logged and metrics about processing
+   * failures should be emitted.
+   *
+   * @param context the complete context for metrics processing
+   */
+  void process(MetricsProcessingContext context);
```
dimas-b: Why not evolve `PolarisMetricsReporter`? 

If we introduce a new SPI, we'll break existing use cases anyway. Changing the old SPI at least allows downstream builds to detect the breakage at compile time 🤔 

@cccs-cat001 : WDYT?

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
obelix74: • Agreed with avoiding duplicative “context” objects for realm data.
• I removed MetricsProcessingContext entirely (and the whole MetricsProcessor path), so we no
       longer create/maintain a parallel realm context (realmId() etc.). Any future metrics handling
       can use the existing request-scoped CDI beans (like RealmContext) instead.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
obelix74: • Agreed: principal should come from CDI request-scoped context rather than being copied into a
       new metrics-specific context object.
     • Since I removed MetricsProcessingContext + MetricsProcessor, we no longer pass principalName()
       through that path at all; implementations should use CDI (e.g., PolarisPrincipalHolder) if/when
        they need principal information.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
obelix74: • Agreed: propagating otelTraceId/otelSpanId as strings in a bespoke context object is not
       ideal.
     • By removing MetricsProcessingContext, we no longer encode OTEL IDs into a separate metrics
       context. Any OTel linkage should come from standard OTel context propagation mechanisms.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
obelix74: • Agreed: request IDs shouldn’t be copied into ad-hoc context DTOs unless there’s a clear need
       and established mechanism.
     • With MetricsProcessingContext removed, we’re no longer passing requestId() through this
       metrics path. If request correlation is needed, it should use Polaris’s existing
       request-scoped/context patterns.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessor.java
obelix74: I evolved `PolarisMetricsReporter`.

dimas-b: Thx! Let's see if @cccs-cat001 has any concerns :)

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -22,5 +22,6 @@
 import org.apache.iceberg.metrics.MetricsReport;
 
 public interface PolarisMetricsReporter {
-  public void reportMetric(String catalogName, TableIdentifier table, MetricsReport metricsReport);
+  void reportMetric(
+      String catalogName, TableIdentifier table, MetricsReport metricsReport, long timestampMs);
```
dimas-b: just wondering: is it critical to pass `timestampMs` from the outside? Is it envisioned to be used for some kind of correlation with other timestamps?

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -21,7 +21,67 @@
...
+ *   public void reportMetric(
+ *       String catalogName, TableIdentifier table, MetricsReport metricsReport, long timestampMs) {
+ *     // Send metrics to external system
+ *   }
+ * }
+ * }</pre>
+ *
+ * @see DefaultMetricsReporter
+ * @see MetricsReportingConfiguration
+ */
```
obelix74, line 39: @dimas-b my ignorance of this is what caused the prior implementation of a parallel context.

dimas-b: No worries

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: @dimas-b I was primarily thinking about external systems.  For instance, if someone does not want to use the built in metrics persistence and forwards it to some other system.  In the past (in different data oriented systems, I have run into problems where if someone wants to correlate an "event" based on time, there was an issue) - I understand if a `traceId` is present, this is a non-issue, perhaps I am being overly defensive in considering cases where the clients do not send a `traceId`.

Without the parameter:
• Custom implementations that want to forward metrics to external systems (CloudWatch, Prometheus, InfluxDB, Kafka) must capture their own timestamp
• That timestamp would be captured inside the reporter implementation, after the metrics have already passed through several layers of processing
• This creates timing drift and inconsistency across different implementations

With the parameter:
• The timestamp is captured at the point of receipt in IcebergCatalogAdapter.reportMetrics():
```     metricsReporter.reportMetric(
         catalogName, tableIdentifier, reportMetricsRequest.report(),
   System.currentTimeMillis());```
• All implementations receive the same, consistent timestamp representing when Polaris received
       the metrics

Important distinction: The MetricsReport from Iceberg contains duration data (how long a scan took), but not wall-clock time (when it happened). The timestampMs provides that missing piece.

Comment (dimas-b) [reply runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java]: That makes sense, however, different Polaris nodes may have different clocks. Nothing guarantees that those clocks are in sync... I'm not sure using a timestamp that is strict only inside one node helps that much... but maybe I'm missing something :thinking: 

I'm not against this change, just wondering if it is actually a benefit as opposed to false hope :thinking: WDYT?

Comment (obelix74) [reply runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java]: Valid point about clock skew @dimas-b. You're right that cross-node synchronization isn't guaranteed.

The main benefit is consistency within the metrics pipeline - capturing the timestamp at receipt in `IcebergCatalogAdapter` rather than letting each implementation capture it independently at different points in processing. It's also useful for correlating with other events on the same node during debugging.

That said, I'm not strongly attached to this parameter. If you think it adds more complexity than value, happy to remove it - the backward-compatible default method makes either approach work. Let me know your preference.

File: runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java [hunk 12]
```diff
@@ -32,23 +32,30 @@
...
+ *
+ * <p>Example implementation with OpenTelemetry correlation:
  *
  * <pre>{@code
  * @ApplicationScoped
  * @Identifier("custom")
  * public class CustomMetricsReporter implements PolarisMetricsReporter {
  *
  *   @Inject RealmContext realmContext;
+ *   @Inject SpanContext spanContext;
```
obelix74: This is the one CDI that is currently missing.  Without this, all users have to do `Span.current().getSpanContext()`.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -429,6 +431,46 @@ public void closeTaskExecutor(@Disposes @Identifier("task-executor") ManagedExec
...
+  @RequestScoped
+  public SpanContext currentSpanContext() {
+    Span currentSpan = Span.current();
+    if (currentSpan != null) {
+      SpanContext spanContext = currentSpan.getSpanContext();
+      if (spanContext != null) {
+        return spanContext;
+      }
+    }
+    return SpanContext.getInvalid();
```
obelix74: A simple producer that allows users to inject a `SpanContext` for OTel.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessor.java
adutra: I also agree that evolving is better than introducing a parallel SPI.

According to `evolution.md`, Java classes are expected to change at any time, so users are already aware of that. However, changes to any `polaris.*` configuration option is considered a major change. Are we honoring the existing `polaris.iceberg-metrics.reporting.type` option? I believe we are. So this evolution should be fine.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -64,18 +65,21 @@ public class IcebergRestCatalogEventServiceDelegator
   @Inject PolarisEventListener polarisEventListener;
   @Inject PolarisEventMetadataFactory eventMetadataFactory;
   @Inject CatalogPrefixParser prefixParser;
+  @Inject RealmConfig realmConfig;
```
adutra: Changes in this file seem unnecessary.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
adutra: Imho this bean is unnecessary.

In fact the static call to `Span.current()` is already "wired" by Quarkus to a machinery to retrieve the request-scoped `Span` associated with the current CDI context. Just use `Span.current()` whenever you need it, and it will give you the right `Span`.

I think that's what @dimas-b meant by "handle OTel via its natural context objects."

Alternatively, Quarkus already allows you to inject `Span` objects out of the box: https://quarkus.io/guides/opentelemetry-tracing#available-opentelemetry-cdi-injections.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
adutra: I also agree that this timestamp doesn't look super helpful, but why not. But I also find it a bit low-level. Why not pass a full `Instant` instead?

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
adutra: See my other comment. It's no bad practice to call `Span.current()`. Alternatively, injecting a `Span` works out of the box too (but not `SpanContext`).

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessor.java
adutra: Update: in fact even the SPI interface is backwards-compatible; this is nice 👍

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
adutra: I'm seeing many areas of Polaris trying to access common information such as the request ID, the realm ID, the Polaris principal and the OTel context.

First, the Events API introduced access to this data in `PolarisEventMetadataFactory` (you will notice that retrieving this info is less trivial than you'd think).

Then, `PolarisPrincipalHolder` was introduced, duplicating the logic in `PolarisEventMetadataFactory.getUser()`.

And now, it seems that the metrics reporting area will also make usage of that same info.

That's fine for now if each part of Polaris does the job locally; but I think some cleanup after this PR would be welcome. We should find a way to expose this data to the rest of the application in a consistent, centralized manner.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
obelix74: Removed, thanks.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: Updated to `Instant.`

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: Removed.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
obelix74: Removed.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
adutra: I see btw that you opened https://github.com/apache/polaris/issues/3444 – this would be a good thing to have!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -723,7 +724,7 @@ public Response reportMetrics(
     TableIdentifier tableIdentifier = TableIdentifier.of(ns, RESTUtil.decodeString(table));
 
     metricsReporter.reportMetric(
-        catalogName, tableIdentifier, reportMetricsRequest.report(), System.currentTimeMillis());
+        catalogName, tableIdentifier, reportMetricsRequest.report(), Instant.now());
```
adutra: You should instead inject a `Clock` instance then call `clock.instant()`.

runtime/service/src/main/java/org/apache/polaris/service/reporting/MetricsProcessingContext.java
obelix74: yeah, I will work on this once we are done with the metrics persistence.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
obelix74: Addressed it.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
dimas-b: Yes, +1 to `Span.current()`

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -18,9 +18,84 @@
...
+   * @param table the identifier of the table the metrics are for
+   * @param metricsReport the Iceberg metrics report
+   * @deprecated Use {@link #reportMetric(String, TableIdentifier, MetricsReport, Instant)} instead.
+   *     This method is provided for backward compatibility and will be removed in a future release.
+   */
+  @Deprecated
+  default void reportMetric(
+      String catalogName, TableIdentifier table, MetricsReport metricsReport) {
+    reportMetric(catalogName, table, metricsReport, Instant.now());
+  }
```
dimas-b: I'm not sure what this method achieves 🤔 Old implementation will have an override for it already, but will be missing an impl. for the new four-parameter `reportMetric()` method, so it's going to be a compilation error anyway (which is ok per our evolution guidelines).

Polaris will not call this method, either (unless I'm mistaken).

Third-party code is not supposed to call this SPI directly (only implement it to receive calls from Polaris).

...  so all in all this looks like dead code 🤔

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -18,9 +18,84 @@
...
+import java.time.Instant;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.metrics.MetricsReport;
 
+/**
+ * SPI interface for reporting Iceberg metrics received by Polaris.
+ *
+ * <p>Implementations can be used to send metrics to external systems for analysis and monitoring.
+ * Custom implementations must be annotated with {@link
+ * jakarta.enterprise.context.ApplicationScoped @ApplicationScoped} and {@link
```
dimas-b: Why do we have to force `ApplicationScoped` here? I believe it's more natural for `PolarisMetricsReporter` implementations to be request-scoped (in order to potentially benefit from other request-scoped data like `PolarisPrincipal`).

dimas-b: TBH, I'm not sure we need to go into too much CDI details in this javadoc (the example impl. below is fine)... otherwise, we'd have to do the same for may other CDI plugin points 😅 Developers integrating at this level can be expected to know CDI, I guess 😉 

Would it be sufficient to method the config-based bean resolution?

Downstream builds will be free to choose whatever CDI scope makes sense for their use cases.

CDI delegation (chaining) to other reported types would also be possible in downstream builds.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
obelix74: Removed already, @dimas-b. `Span.current()` should work.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRestCatalogEventServiceDelegator.java
```diff
@@ -805,6 +805,7 @@ public Response reportMetrics(
       ReportMetricsRequest reportMetricsRequest,
       RealmContext realmContext,
       SecurityContext securityContext) {
+    // Delegate directly to the underlying service.
```
dimas-b: nit: isn't it obvious from the code?

CHANGELOG.md
```diff
@@ -52,6 +52,10 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - The EclipseLink Persistence implementation has been completely removed.
 - The default request ID header name has changed from `Polaris-Request-Id` to `X-Request-ID`.
 - The (Before/After)CommitTableEvent has been removed.
+- The `PolarisMetricsReporter.reportMetric()` method signature has been extended to include a
+  `receivedTimestamp` parameter of type `java.time.Instant`. A backward-compatible default method
+  is provided for existing implementations, but custom implementations should migrate to the new
+  signature.
```
dimas-b: Please also see my comment on the actual SPI change. I do not think we provide backward compatibility in this case (nor do we have to).

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
dimas-b: Fair enough. Let's go with `Instant`

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -18,9 +18,40 @@
...
 
+import java.time.Instant;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.metrics.MetricsReport;
 
+/**
+ * SPI interface for reporting Iceberg metrics received by Polaris.
+ *
+ * <p>Implementations can be used to send metrics to external systems for analysis and monitoring.
+ * Custom implementations must be annotated with {@link
```
dimas-b: Sorry, I still think "must" is an overreach here 😅  Since Polaris will work with `@RequestScoped` impl. (I assume), we should not dictate how downstream projects scope their CDI beans.

dimas-b: Application-scoped beans can be used instead of request-scoped beans without any conflict, IIRC.

runtime/service/src/testFixtures/java/org/apache/polaris/service/events/listeners/TestPolarisEventListener.java
```diff
@@ -43,4 +43,17 @@ public PolarisEvent getLatest(PolarisEventType type) {
     }
     return latest;
   }
+
+  /**
+   * Returns the latest event of the specified type, or null if no such event has been recorded.
+   * This is useful for tests that need to verify no event was emitted.
+   */
+  public PolarisEvent getLatestOrNull(PolarisEventType type) {
```
dimas-b: nit: is it still used?

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: Removed it.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: Updated the documentation.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: Reworded the JavaDoc, please see if it makes sense.

runtime/service/src/testFixtures/java/org/apache/polaris/service/events/listeners/TestPolarisEventListener.java
obelix74: Removed unused methods.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
```diff
@@ -18,9 +18,39 @@
...
+ * <p>The implementation to use is selected via the {@code polaris.iceberg-metrics.reporting.type}
+ * configuration property, which defaults to {@code "default"}.
+ *
+ * <p>Implementations can inject other CDI beans for context.
+ *
+ * @see DefaultMetricsReporter
+ * @see MetricsReportingConfiguration
+ */
 public interface PolarisMetricsReporter {
-  public void reportMetric(String catalogName, TableIdentifier table, MetricsReport metricsReport);
```
singhpk234, line 25: we release this public interface in 1.3 right ? i wonder if we keep this api with default impl to the new method with null or something ? 
https://github.com/apache/polaris/commit/361b7e9241da4fdbb0af9f10cf57788fbb8f2dcb

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - The EclipseLink Persistence implementation has been completely removed.
 - The default request ID header name has changed from `Polaris-Request-Id` to `X-Request-ID`.
 - The (Before/After)CommitTableEvent has been removed.
+- The `PolarisMetricsReporter.reportMetric()` method signature has been extended to include a
```
singhpk234: i belive we need to freeze the changelog for 1.3 since its released and then add this in an unrelease section ?

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: @singhpk234 I had a deprecated backward compatible method and the consensus was to remove it since this SPI is not meant to be used by developers outside Polaris.

CHANGELOG.md
obelix74: I will move it to unreleased section.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
singhpk234: Didn't fully get it, wdym by developers outside Polaris ? can this not be used downstream projects, my recommendation is to make sure version upgrades are seemless so keep this and add more if we want and just update the default of this work under timestamp null assumption.

singhpk234: Let me also check previous discussions, can you please point me to that ?

CHANGELOG.md
obelix74: @singhpk234 I don't see a `1.3.0` section in the CHANGELOG - I synced my fork to upstream.

https://github.com/apache/polaris/blob/main/CHANGELOG.md

Am I missing something?

obelix74: Perhaps we need to rename `Unreleased` to `1.3.0-incubating` in CHANGELOG and move this line alone to `Unreleased`.

singhpk234: no you are not ideally there needs to a update in the main branch to freeze 1.3 section and create the unreleased section as soon as a version of Polaris is released, seems like we need to do it first ?

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
obelix74: @singhpk234 https://github.com/apache/polaris/pull/3468#discussion_r2709660554.  @dimas-b FYI.

obelix74: BTW, when I uptook `1.3.0`, my custom token broker had compilation errors due to signature changes. Looks like minor versions can change public signatures?

dimas-b: Per our standing [evolution guidelines](https://polaris.apache.org/in-dev/unreleased/evolution/) `public` classes / methods can change in any release. Unlike REST API changes, it is not considered a "major" change for versioning purposes. Essentially java methods are not part of the API surface in the [SemVer](https://semver.org/) sense.

We should and do try to make java API changes in a backward -compatible manner when practical.

Regarding this particular PR and java interfaces that are part of an SPI (defined and called by Polaris, implemented by 3rd party plugins), I do not see a practical way to evolve them in a backward-compatible manner without causing excessive maintenance burden in Polaris code.

In this case, Polaris would have to define a _new_ interface for the new method signature and perform `instanceof` checks at runtime in order to decide whether to call the old or the new method. I do believe it would be an overkill at the current stage of the project (still evolving actively). It is not too hard for downstream implementations to adjust code for the newly added parameter.

That said, I'd like to propose moving the SPI evolution discussion to the `dev` ML as it is a big and complex topic.

@singhpk234 : please clarify whether you consider this comment thread a blocker for merging (or not).

CHANGELOG.md
dimas-b: As far as the CHANGELOG "diff" in this PR is concerned, it looks correct to me.

I'd expect @pingtimeout (as the release manager for 1.3.0) to open a reconciliation PR for `main` and add a dedicated section for `1.3.0` soon :wink: I'm sure @pingtimeout will be able to deal with conflicts if this PR merges first :slightly_smiling_face:

dimas-b: @obelix74 : please do not add a 1.3.0 section to CHANGELOG in this PR :)

obelix74: @dimas-b Noted.

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
singhpk234: @dimas-b can we not add the api in the same interface and add the default impl of old api call this null clock, i would like to understand this more if this is possible. If its not possible i am supportive of the change !

> @singhpk234 : please clarify whether you consider this comment thread a blocker for merging (or not).

 thanks for asking the clarification i believe anything is thats not marked as 'nit' / 'not blocker' / 'not' is expected to be resolved before merging (per [here](https://polaris.apache.org/community/contributing-guidelines/#code-contribution-guidelines)) , i really appreciate you checking in !

CHANGELOG.md
singhpk234: SG !

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
dimas-b: > can we not add the api in the same interface and add the default impl of old api call this null clock [...]

I do not see a point in adding a `default` impl for the _old_ method alone. Existing implementations will have overrides for it already (javac will not let them miss that).

We could add a `default` impl. for the _new_ method and redirect to the old one (without the `Instant`). This will allow existing implementations to compile without changes. However, this creates uncertainty for the implementor regarding which method should do the real work. Javadoc helps, but adds cognitive load. We could add `default` to both methods and deprecate the old one, but this will add "cruft" to the interface definition, when, I believe, adapting to the new interface in downstream projects is very easy.

Please note that providing a custom implementation for this interface requires a downstream build. So all upgrades will go through local builds and CI (assuming normal software engineering practices), where the need for adjustments will be become apparent.

obelix74: When I uptook 1.3.0, I was glad that my compilation failed since if the old method was deprecated, I would be confused why I need to uptake the new approach and how to do it. 

Having said that, this PR started as adding events for metrics, we removed it. It only contains this extra attribute.  It is not a blocker for me.  If we don't agree on how to proceed, I would rather abandon this PR and focus on the simplified metrics persistence code here: https://github.com/apache/polaris/pull/3385.  That allows Polaris to persist table metrics to the database.

CHANGELOG.md
dimas-b: Cf. #3503

runtime/service/src/main/java/org/apache/polaris/service/reporting/PolarisMetricsReporter.java
singhpk234: Thats fair, though in this case we are forcing the upgrades to have this new interface implementation when having the old api was not hurting ? taking the case LOGGERs can be configured to implicitly log timestamp additionally and the way we have reporting wired if i wanna do custom one can just do timestamp internally rather than requesting from the signature ? 

```
   metricsReporter.reportMetric(
        catalogName, tableIdentifier, reportMetricsRequest.report(), clock.instant());
```

it not the time when the request hit the server but its the time when the reportMetric is called so does it matter of i do this vs 

 reportMetric(catalogName, tableIdentifier, reportMetricsRequest.report())) {
   timestamp = clock.instant() 
 }

with that being said i think its fine if we wanna move forward :), but above is my thought process. Please move forward @dimas-b trust your judgement here !

dimas-b:
It looks like we have consensus in all discussions and some approvals. If no new concerns are raised. I propose to merge on Jan 23.

obelix74:
Thanks all.

dimas-b:
@obelix74 : unfortunately, it looks like there's a conflict on `CHANGELOG.md` now... Could you resolve it?

obelix74:
> @obelix74 : unfortunately, it looks like there's a conflict on `CHANGELOG.md` now... Could you resolve it?

Resolved.

PR #3474, opened 2026-01-19
Use archive.apache.org for previous releases

jbonofre:
Only the last release should be on dist.apache.org, all previous releases are available on archive.apache.org.

I will clean up dist.apache.org once this PR is merged.


snazy (commented):
@pingtimeout should we update the release workflow to
1. "prematurely" push release artifacts to archive.a.o and
2. purge previous releases from downloads.a.o in the final workflow?

jbonofre:
No we should not push to archive: it's automatically sync with dist. 

However we can cleanup previous releases from dist.apache.org when a new release is promoted. We should also update the website to use archive.apache.org on previous releases.

pingtimeout:
@jbonofre If I understand correctly, there should only be one version on the dist release SVN server?

jbonofre:
@pingtimeout yes, dist.apache.org (release space e.g. https://dist.apache.org/repos/dist/release/incubator/polaris/) should contain only the latest release.

jbonofre:
@pingtimeout for instance https://dist.apache.org/repos/dist/release/karaf/ you can see only the latest releases on each active branch.

PR #3476, opened 2026-01-19
Describe the documentation steps of the release process

pingtimeout:
These steps have not been automated (and might not be automatable).  But more importantly, they were not clearly defined.  This PR is an attempt at sharing knowledge following the lessons learned during 1.3.0 release.

I have updated both the semi-automated release guide and manual release guide for completeness.

@jbonofre @flyrain @singhpk234 does this match your recollection from previous releases?

### DIFF
--- a/site/content/community/release-guides/manual-release-guide.md
+++ b/site/content/community/release-guides/manual-release-guide.md
@@ -466,22 +466,46 @@ You can now publish Docker images on DockerHub:
 ```
 
 ### Upload the documentation
-Now that the release artifacts have been published, the next step is to publish the associated documentation on the website.  Checkout the `versioned-docs` branch locally and copy the documentation associated to the release tag at the root of the repository.
+Now that the release artifacts have been published, the next step is to publish the associated documentation on the website.
+
+First, checkout the release tag:
 
 ```
-git checkout -b versioned-docs-[major].[minor].[patch] versioned-docs
+git checkout apache-polaris-[major].[minor].[patch]-incubating
+```
+
+Set up a worktree for the versioned docs and create a new branch:
+
+```
+site/bin/checkout-releases.sh
+cd site/content/releases
+git checkout -b versioned-docs-[major].[minor].[patch]
 mkdir [major].[minor].[patch]
-git archive apache-polaris-[major].[minor].[patch]-incubating site/content/in-dev/unreleased | tar -x --strip-components=4 -C [major].[minor].[patch]/
 ```
 
-Edit the file `[major].[minor].[patch]/_index.md` and perform the following modifications.
+Copy the documentation from the release tag:
+
+```
+cp -r ../../content/in-dev/unreleased/* [major].[minor].[patch]/
+```
+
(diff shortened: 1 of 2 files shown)


site/content/community/release-guides/manual-release-guide.md
```diff
@@ -465,11 +465,52 @@ You can now publish Docker images on DockerHub:
...
-### Publishing docs 
-1. Open a PR against branch [`versioned-docs`](https://github.com/apache/polaris/tree/versioned-docs) to publish the documentation
-2. Open a PR against the `main` branch to update website
-    - Add download links and release notes in [Download page](https://github.com/apache/polaris/blob/main/site/content/downloads/_index.md)
-    - Add the release in the [website menu](https://github.com/apache/polaris/blob/main/site/hugo.yaml)
+### Upload the documentation
+Now that the release artifacts have been published, the next step is to publish the associated documentation on the website.  Checkout the `versioned-docs` branch locally and copy the documentation associated to the release tag at the root of the repository.
+
+```
+git checkout -b versioned-docs-[major].[minor].[patch] versioned-docs
```
snazy: Noting there's the script `site/bin/checkout-releases.sh` do this (into `content/releases`), just via Git worktrees. `site/bin/remove-releases.sh` removes the Git worktree.

pingtimeout: TIL that a worktree can be added as a subdirectory of the current branch.  I wonder if it could be confusing that any `git commit` made inside that worktree will be for the `versioned-docs` branch even though the current repo is no another branch.  Also the worktree would need to be removed after the release to avoid confusion/mistakes.  But on the other hand, it would replace the `git archive` long command by a simpler recursive `cp`...

snazy: Nah, there's technically no issue w/ Git (it's smart ;) ).

From experience, I'd recommend to remove the "inner worktree" as soon as possible, because IDEs like IntelliJ recognize the "inner worktree" and for example fold that history in the Git view, which is then a bit confusing.

snazy:
@pingtimeout any updates on this one? I think we should have this for 1.4.0

site/content/community/release-guides/manual-release-guide.md
pingtimeout: It should be good now

site/content/community/release-guides/semi-automated-release-guide.md
```diff
@@ -301,7 +301,73 @@ This workflow can only be run from the `release/[major].[minor].x` branch for wh
 ## Publish docs
 These steps have not been automated yet.
 
-1. Open a PR against branch [`versioned-docs`](https://github.com/apache/polaris/tree/versioned-docs) to publish the documentation
-2. Open a PR against the `main` branch to update website
-    - Add download links and release notes in [Download page](https://github.com/apache/polaris/blob/main/site/content/downloads/_index.md)
-    - Add the release in the [website menu](https://github.com/apache/polaris/blob/main/site/hugo.yaml)
+### Upload the documentation
```
snazy, line 304: Created #3746 as a follow-up

PR #3484, opened 2026-01-20
NoSQL: Prepare for Quarkus 3.31.0

snazy:
Quarkus 3.31.0 comes with two changes to the MongoDB extension. Spoiler: this is a no-op change for older Quarkus versions.

The way `MongoClient` instances are produced by the Quarkus MongoDB extension has changed in a way that makes it impossible to get a _managed_ `MongoClient` from `MongoClients`. This requires a change to `MongoDbBackendBuilder`.

It also introduces a new configuration option `quarkus.mongodb.active` to enable explicitly enablement of the default or named MongoDB drivers. The default value of this configuration is `true`. The readiness and health management endpoints report the status of the MongoDB drivers. With the default `quarkus.mongodb.active=true`, the MongoDB driver would always "break" those checks, if MongoDB is not configured. This is handled by a custom configuration-source, which provides `quarkus.mongodb.active=true` if NoSQL is being used with Mongo, otherwise it provides `quarkus.mongodb.active=false`.

### DIFF
--- a/persistence/nosql/persistence/cdi/quarkus/src/main/META-INF/services/io.smallrye.config.ConfigSourceFactory
+++ b/persistence/nosql/persistence/cdi/quarkus/src/main/META-INF/services/io.smallrye.config.ConfigSourceFactory
@@ -1,0 +1,17 @@
+#
+# Copyright (C) 2026 Dremio
+#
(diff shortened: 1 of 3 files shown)


snazy:
Bummer - it would be a breaking change for Quarkus < 3.31

Issue #3039, opened 2025-11-13
Site: Keep documentation and downloads information at a reasonable size


snazy:
### Is your feature request related to a problem? Please describe.

The (current state of) https://polaris.apache.org/ web site has a couple of Polaris release related pieces of information:
* Versioned docs
  * documentation per release version
  * documentation of the latest state on the `main` branch
  * accessible via the "Documentation" drop-down nav-bar menu
* Releases (downloads) page, containing
  * Links to release artifacts
  * Changelog for the release
* Search
  * All pages, including the docs for old releases, are indexed for the search functionality

There are some elements that could or even should be improved.
* Search: There are a couple of options to limit the amount of search results:
  * Exclude all but the in-dev and latest release from the search index
  * Exclude all but the in-dev and ["active" releases]((https://github.com/apache/polaris/blob/74c58354c4182e76fa3e77d1e9faa07fbde71e52/site/hugo.yaml#L66)
* Releases (downloads) page.
  * Split the single page, as also [proposed here](https://github.com/apache/polaris/pull/2928#pullrequestreview-3457451706)
  * Either:
    * Have a separate page showing releases. Maybe split into active and older releases.
    * Move the individual download/changelog information to the versioned docs
* "Documentation" drop-down menu
  * Should it list all ever released versions?
  * Should it list all ["active" releases]((https://github.com/apache/polaris/blob/74c58354c4182e76fa3e77d1e9faa07fbde71e52/site/hugo.yaml#L66)
  * In either "yes" for the above: a separate page showing the links for the "older" releases would be nice.

Ideally, a job that's part of the automated-release-workflow should help by creating PRs do perform changes against the web site content and config on the `main` branch.
Some changes, depending on the outcome(s) of the above bullet-list options, like adding a front-matter option to exclude pages from the search index, affect a lot of files.

Issue #3037, opened 2025-11-13
Move the "Polaris Evolution" page out of versioned-docs


snazy:
### Is your feature request related to a problem? Please describe.

The ["Polaris Evolution"](https://polaris.apache.org/in-dev/unreleased/evolution/) page is currently in the directory structure targeting release-specific information, which means that it for example pops up multiple times, once for each release and "in-dev", as shown in the image below.

<img width="617" height="715" alt="Image" src="https://github.com/user-attachments/assets/219d9561-c26a-41be-989a-eb167f7580af" />

The page is probably better suited in the "non-release specific" part of the web site.

If we consider the URL to the page as a permalink, we can move the doc and add a configuration to redirect clients to the new URL - or use a different approach to keep the permalink alive.

eric-maynard:
Actually I feel like maybe the search function should just not show multiple versions of any doc? 

Like this is just bad UI/UX, it's not specific to any one doc

<img width="871" height="891" alt="Image" src="https://github.com/user-attachments/assets/b5ca0da0-8df1-4042-8185-6655e5ac49f5" />

snazy:
> Actually I feel like maybe the search function should just not show multiple versions of any doc?
> 
> Like this is just bad UI/UX, it's not specific to any one doc

I totally agree. It's rather confusing. Tackling this in a general way is why I opened #3039

PR #3485, opened 2026-01-20
Site: add tool to mark versioned-docs as "do not index"

snazy:
Adds a shell script to add the front-matter tags `robots: noindex` (HTML META tag) and `exclude_search: true` (local site search) for versioned docs for a specific version.

The rudimentary script handles only markdown files files, not asciidoc files, because there are currently only .md files in the versioned-docs.


snazy:
See also #3486

snazy:
See #3039 and #3037 for context.

PR #3486, opened 2026-01-20
Versioned-docs: do not index old versioned docs

snazy:
Adds the front-matter tags `robots: noindex` (HTML META tag) and `exclude_search: true` (local site search) for old released versions, aka all except 1.3.0. The change was created using the script from #3485.

### DIFF
--- a/1.0.0/getting-started/install-dependencies.md
+++ b/1.0.0/getting-started/install-dependencies.md
@@ -20,6 +20,8 @@
 Title: Installing Dependencies
 type: docs
 weight: 100
+robots: noindex
+exclude_search: true
 ---
 
 This guide serves as an introduction to several key entities that can be managed with Apache Polaris (Incubating), describes how to build and deploy Polaris locally, and finally includes examples of how to use Polaris with Apache Spark&trade;.
--- a/0.9.0/_index.md
+++ b/0.9.0/_index.md
@@ -29,6 +29,8 @@ cascade:
   params:
     show_page_toc: true
 # This file will NOT be copied into a new release's versioned docs folder.
+robots: noindex
+exclude_search: true
 ---
 
 Check out the [Quick Start]({{% ref "quickstart" %}}) page to get started.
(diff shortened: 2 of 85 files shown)


1.0.0/getting-started/install-dependencies.md
```diff
@@ -115,4 +117,4 @@ Ensure that `java --version` and `javac` both return non-zero responses.
 
 ## jq
 
-Most Polaris Quickstart scripts require `jq`. Follow the instructions from the [jq](https://jqlang.org/download/) website to download this tool.
\ No newline at end of file
+Most Polaris Quickstart scripts require `jq`. Follow the instructions from the [jq](https://jqlang.org/download/) website to download this tool.
```
dimas-b, line 120: what's the change here? 🤔

dimas-b: possibly something to be fixed in the tool? 🤔

dimas-b:
@snazy : what is the rationale for _not_ indexing old docs? thx!

snazy:
Rationale is:
a) the search on the web site yields duplicate results
b) search engines (e.g. Google) or AI tools consider outdated information

1.0.0/getting-started/install-dependencies.md
snazy: Newline

dimas-b:
From my POV the information is not necessarily duplicate / outdated as old docs are relevant to their respective project releases and the content can be different.

Although, I do not mind this change, I wonder what other people think about this :thinking:

snazy:
See #3039 and #3037 for context.

1.0.0/getting-started/install-dependencies.md
pingtimeout: Indeed, a newline at the end of the file

pingtimeout:
I think for now, given the pace of Polaris and the fact that we are not supporting older releases, it makes sense to only let search engines index the most recent release only.

@dimas-b IMO, the risk of indexing old versions is that there will be more http links pointing to those versions than on the most recent ones.  So people search for Polaris help would be redirected to old versions that are not necessary accurate anymore, missing either features or bugfixes.  Wdyt?

Issue #3061, opened 2025-11-15
loadView issues multiple storage requests for the same file


dimas-b:
### Describe the bug

* Execute `select * from myView` in Spark
* Observe multiple metadata load requests in Polaris.

Server-side stack traces:

```
java.lang.Exception: test
	at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:733)
	at com.intellij.rt.debugger.MethodInvoker.invokeInternal(MethodInvoker.java:223)
	at com.intellij.rt.debugger.MethodInvoker.invoke1(MethodInvoker.java:35)
	at org.apache.iceberg.aws.s3.S3InputFile.fromLocation(S3InputFile.java:128)
	at org.apache.iceberg.aws.s3.S3InputFile.fromLocation(S3InputFile.java:123)
	at org.apache.iceberg.aws.s3.S3FileIO.newInputFile(S3FileIO.java:180)
	at org.apache.polaris.service.catalog.io.ExceptionMappingFileIO.newInputFile(ExceptionMappingFileIO.java:55)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog$BasePolarisViewOperations.lambda$doRefresh$0(IcebergCatalog.java:1805)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog$PolarisOperationsBase.lambda$refreshFromMetadataLocation$0(IcebergCatalog.java:2033)
	at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413)
	at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:219)
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:203)
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:196)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog$PolarisOperationsBase.refreshFromMetadataLocation(IcebergCatalog.java:2033)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog$BasePolarisViewOperations.doRefresh(IcebergCatalog.java:1786)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog$BasePolarisViewOperations.refresh(IcebergCatalog.java:1715)
	at org.apache.polaris.service.catalog.iceberg.IcebergCatalog$BasePolarisViewOperations.current(IcebergCatalog.java:1705)
[...]

PR #3488, opened 2026-01-20
deduplicate storage requests when loading views.

evindj:
The PR addresses issue #3061 where loadview makes multiple requests for the same file.
This PR has two main changes to avoid making too many storage requests when loading tables and views. 
The first change is aimed at reusing ```fileIO``` in case ```refreshFromMetadataLocation``` is retried. this is done for both tables and views.

The second change is related to ```loadView``` only it overrides ```loadView``` in ```BaseMetastoreViewCatalog```  to reuse viewOps instead of creating them twice.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1838,7 +1838,7 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
                       resolvedEntities,
                       new HashMap<>(tableDefaultProperties),
                       Set.of(PolarisStorageActions.READ, PolarisStorageActions.LIST));
-                      
+
               return ViewMetadataParser.read(fileIO.newInputFile(metadataLocation));
             });
         polarisEventListener.onEvent(


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1390,25 +1410,23 @@ public void doRefresh() {
                 new AttributeMap()
                     .put(EventAttributes.CATALOG_NAME, catalogName)
                     .put(EventAttributes.TABLE_IDENTIFIER, tableIdentifier)));
+        // Create FileIO once outside the retry lambda to avoid redundant storage requests
```
dimas-b: How does this help with reducing storage requests? Creating a `FileIO` should not hit storage, AFAIK 🤔 

This is just about the comment. Code change LGTM.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -810,6 +812,24 @@ protected ViewOperations newViewOps(TableIdentifier identifier) {
     return new BasePolarisViewOperations(catalogFileIO, identifier);
   }
 
+  /**
+   * Override to fix a bug in {@link BaseMetastoreViewCatalog#loadView} where {@link #newViewOps} is
```
dimas-b, line 816: tangential: Could you also double check for something like this in tables?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogLoadViewOptimizationTest.java
```diff
@@ -0,0 +1,409 @@
...
+
+    // Act: Load the view
+    // View view = spyCatalog.loadView(VIEW_ID);
+
+    CatalogHandlerUtils catalogHandlerUtils = new CatalogHandlerUtils(2, false);
+    LoadViewResponse response = catalogHandlerUtils.loadView(spyCatalog, VIEW_ID);
+    ViewMetadata view = response.metadata();
+
+    // Assert: newViewOps() should be reused (called exactly once)
+    verify(spyCatalog, times(1)).newViewOps(VIEW_ID);
```
dimas-b: Running a `@QuarkusTest` (which builds a full server) to do some mock checks looks like an overkill to me.

Could we leverage `MeasuredFileIO` instead and move this test to an existing Quarkus test (to reduce the number of test servers)?

dimas-b: Also, if testing this particular bug turns out to be cumbersome, I think manual verification is sufficient. 

Existing CI should be sufficient to ensure the changed code works well.

evindj: Yeah it would be a bit challenging to test without turning up an environment because, the duplicate request happens while loading the view and while returning the response. I have verified manually so I am ok removing this expensive test.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
evindj: Prior to the change, if ```refreshFromMetadataLocation``` was retried, it will created another FileIO, I can update the comment.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1800,27 +1817,23 @@ public void doRefresh() {
                 new AttributeMap()
                     .put(EventAttributes.CATALOG_NAME, catalogName)
                     .put(EventAttributes.VIEW_IDENTIFIER, identifier)));
+        // Create FileIO once outside the retry lambda to avoid redundant storage requests
```
dimas-b: Same concern here - not re-creating FileIO is good, but I do not think it relates to duplicate storage requests...

To be completely clear wrt #3061, WDYT about making the FileIO optimization change in a separate PR?

dimas-b: Other than that this PR LGTM :+1: :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dimas-b: @evindj : any insight here? it's not required for current PR, but since you're involved in this issue already, I thought it might be easy for you to quickly double check that :wink:

evindj: yeah, it is my understanding that this would not happen for LoadTable, because loadtable directly uses the operations from the table to construct the response. See https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java#L387

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
evindj: I can move the FileIO optimization to a separate PR. let me remove then here.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1395,11 +1415,11 @@ public void doRefresh() {
             SHOULD_RETRY_REFRESH_PREDICATE,
             getMaxMetadataRefreshRetries(),
             metadataLocation -> {
-              String latestLocationDir =
-                  latestLocation.substring(0, latestLocation.lastIndexOf('/'));
               // TODO: Once we have the "current" table properties pulled into the resolvedEntity
               // then we should use the actual current table properties for IO refresh here
               // instead of the general tableDefaultProperties.
+              String latestLocationDir =
+                  latestLocation.substring(0, latestLocation.lastIndexOf('/'));
```
dimas-b: one last thing, please - would you mind removing these changes now that there are not necessary for the purpose of this PR?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1818,7 +1838,7 @@ public void doRefresh() {
                       resolvedEntities,
                       new HashMap<>(tableDefaultProperties),
                       Set.of(PolarisStorageActions.READ, PolarisStorageActions.LIST));
-
+
```
MonkeyCanCode: Please run `.//gradlew spotlessApply` or `make spotless-apply` to fix the lint then CI will move forward.

PR #3493, opened 2026-01-21
(fix) Remove KMS policies when KMS is not configured and improved default KMS permission for RO/RW

MonkeyCanCode:
This PR addressed one of the issues reported in https://github.com/apache/polaris/issues/3440 where when end-user is not using KMS encryption for S3, Polaris still enforces in-lines policies which contains KMS related policies. While fixing this issue, I noticed our read-only policy for kMS is a bit too wide where GenerateDataKey and GenerateDataKeyWithoutPlaintext should be belongs to write operation instead of read. Thus, this PR also addresses this issue.

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -299,9 +299,16 @@ private static void addKmsKeyPolicy(
       String region,
       String accountId) {
 
-    IamStatement.Builder allowKms = buildBaseKmsStatement(canWrite);
     boolean hasCurrentKey = kmsKeyArn != null;
     boolean hasAllowedKeys = hasAllowedKmsKeys(allowedKmsKeys);
+    boolean isAwsS3 = region != null && accountId != null;
```
dimas-b, line 304: For follow-up:

This logic for deciding "AWS or not" does not look robust to me.

I'd prefer to make this a property in the Storage Config (similar to `stsUnavailable`). We could choose the default value for it as `true` if no custom `endpoint` or `stsEndpoint` values are set (`false` if custom endpoints are set).

This should be fairly backward compatible... WDYT?

dimas-b: The change in this PR appears to be compatible to old KMS logic, so I think it's ok to merge "as is" and improve config later.

dimas-b (approved):

dimas-b:
FYI: @fabio-rizzo-01

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
MonkeyCanCode: That is correct. This is only for not include default RO KMS policy. For deciding "AWS or not", I raised a dev mailing list and a different PR. For sure using a new property will be most robust and less FP.

MonkeyCanCode:
@dimas-b @singhpk234 mind do another approve as I had to fix conflict on recent CHANGELOG.md issue.

PR #3494, opened 2026-01-21
(NOT_READY) Fixed IcebergCatalog stale FileIO after metadata refresh

MonkeyCanCode:
This PR addressed one of the issues reported in https://github.com/apache/polaris/issues/3440 where spark client is able to use STS temporary credential to create table but not using it for other S3 I/O such as insert for the following call when authed via AssumeRole. It doesn't appear we have any setup or test cases for actually validating this workflow thus the validation was done by the reporter.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1409,6 +1409,22 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
                       Set.of(PolarisStorageActions.READ, PolarisStorageActions.LIST));
               return TableMetadataParser.read(fileIO, metadataLocation);
             });
+
+        // After a refresh, re-load the FileIO with the new table metadata properties to
+        // ensure the right permissions are present for subsequent file system interactions.
+        if (currentMetadata != null) {
+          tableFileIO =
+              loadFileIOForTableLike(
+                  tableIdentifier,
+                  StorageUtil.getLocationsUsedByTable(currentMetadata),
+                  resolvedEntities,
+                  new HashMap<>(currentMetadata.properties()),
+                  Set.of(
+                      PolarisStorageActions.READ,
+                      PolarisStorageActions.WRITE,
+                      PolarisStorageActions.LIST));
+        }
+
         polarisEventListener.onEvent(
             new PolarisEvent(
                 PolarisEventType.AFTER_REFRESH_TABLE,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,8 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### Fixes
 
+- Fixed IcebergCatalog stale FileIO after metadata refresh, resolving AWS assumeRole permission errors.
+
 ### Commits
 
 ## [1.2.0-incubating]
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
@@ -114,6 +114,7 @@ import org.apache.polaris.core.exceptions.CommitConflictException;
 import org.apache.polaris.core.identity.provider.ServiceIdentityProvider;
 import org.apache.polaris.core.persistence.MetaStoreManagerFactory;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
+import org.apache.polaris.core.persistence.PolarisResolvedPathWrapper;
 import org.apache.polaris.core.persistence.cache.EntityCache;
 import org.apache.polaris.core.persistence.dao.entity.BaseResult;
 import org.apache.polaris.core.persistence.dao.entity.EntityResult;
@@ -124,6 +125,7 @@ import org.apache.polaris.core.persistence.resolver.Resolver;
 import org.apache.polaris.core.persistence.resolver.ResolverFactory;
 import org.apache.polaris.core.secrets.UserSecretsManager;
 import org.apache.polaris.core.storage.CredentialVendingContext;
+import org.apache.polaris.core.storage.PolarisStorageActions;
 import org.apache.polaris.core.storage.PolarisStorageIntegration;
 import org.apache.polaris.core.storage.PolarisStorageIntegrationProvider;
 import org.apache.polaris.core.storage.StorageAccessConfig;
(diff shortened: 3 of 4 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1409,6 +1409,22 @@ public void doRefresh() {
...
+          tableFileIO =
+              loadFileIOForTableLike(
+                  tableIdentifier,
+                  StorageUtil.getLocationsUsedByTable(currentMetadata),
+                  resolvedEntities,
+                  new HashMap<>(currentMetadata.properties()),
+                  Set.of(
+                      PolarisStorageActions.READ,
+                      PolarisStorageActions.WRITE,
+                      PolarisStorageActions.LIST));
```
dimas-b, line 1425: Why do we request R/W access for FileIO in a "refresh" method?

Having to reset `tableFileIO` in this context sounds like a design issue to me. I'd think access level (Read / Write / List) should be decided per use case (load vs. commit, etc.).

WDYT?

MonkeyCanCode: Yeah, resetting the permissions here does feel a bit odd. So we may need to do some refactor to fix this bug. Based on my understanding, the issue reporter faced was AWS assumeRole credentials became stale after table creation due to cached FileIO thus a quick "workaround" here is to reset it to full which then allowed the reporter to continue the test and validated the root cause. If we do want to keep the FileIO cached, we may need to do reader/writer FileIO separately to follow the least privileges pattern. Another alternative would be creating new FileIO per operation and right permissions for them. Any preferred route or better solutions?

dimas-b: I personally need to think more about this. I'm not totally familiar with this code.

If someone else is available for a review - please comment.

MonkeyCanCode: @fabio-rizzo-01 do you mind take a look?

dimas-b: @dennishuo : WDYT?

evindj: STS expiring is interesting as there are many ways to think about it:
- the catalog could take the stance that the client is in charge of making sure that the STS  is refreshed possibly after either some error code or some client side tracking as to when the STS will expire. This seems reasonable to me because from the catalog is just a broker for STS.
- When it comes with the interaction with the Catalog, we could raise a clear error code making it possible for the client to know that their STS has expire and thus they can trigger an STS refresh flow.
- We could also try to fix the way we are trying to do here, but I do think it should be part of some internal STS lifecycle tracking either based on feedback from the storage system. This would not address all the case of STS expiring as the client writing directly to the storage system will still have to addressed

dimas-b: Just a thought (while we're considering this change): I'd think Polaris may be used in a read-only mode where users just register existing tables, but only read then through Polaris... Would this `loadFileIOForTableLike` call that asks for R/W access prevent that use case? :thinking:

MonkeyCanCode: That is an interesting view as Polaris can be use as a RO catalog which it may ended up throwing error (as I don't have a setup that reporter has to fully test this). In that case, I would think we may want to pick one of the following routes instead of the solution above:
1. use separate reader/writer FileIO
2. create new FileIO per operation

I can also looking into to see if I can get a similar setup up (unfortunately, there is no minimal reproducible provided by reporter).

dimas-b: I'm leaning towards option 2. Do we know how often this FileIO instance is actually resued in runtime?

Sorry, I'm not very familiar with this area of the code myself, but I imagine every new REST request gets a new FileIO anyway, so reuse is limited to one request, I guess :thinking:

MonkeyCanCode: I think so. I am also a bit new to this specific code path but the cached FileIO appears to be the problem in this case. Let me update to use option 2 and avoid cache.

MonkeyCanCode: So I tried to implement the fix but it ends up being a major refactor and we may needed a delegate FileIO class. 

Based on my understanding, the issue here is for that specific setup (I also tried to see if I can get something similar setup on my local for validation but no luck so far as setup details are not being shared nor a minimal reproducible), after spark use its RO credential with assume role privilege to assume the role associated with catalog, it won't be able to use the STS token for insert thus falled back to client id/secret associated with spark client. Which in this case, it can only happened after the metadata refresh (as it will need to be call first before insert can happen based on my understanding). Then spark will use the returned FileIO from function `io()` which can lost write access (thus the initial fix worked). 

However, for the setup I have on AWS, the client role always doesn't have write access to the bucket directly and it uses assume role with the IAM role associated with the catalog for all operations. So I believed this may be very setup specific for this issue.

I am bit hesitated to do the major refactor for delegate FileIO (so permission can be determined based on context such as RO for `io().newInputFile()` and RW for `io().newOutputFile()`) as I don't have a setup which I can test tp validate as well as future refactors which can break this silently as well without some proper test cases/setups.

Maybe we should have someone who is more familiar with this code path to take another look in case I missed something here. cc @dimas-b @evindj @flyrain

dimas-b: @MonkeyCanCode : Thanks for the investigation!

From my side, I'm willing to look deeper into refactoring this code, but I do not really understand the problem (yet) :sweat_smile: 

Do you have repro instructions handy? Could you post them here or summarize in #3440 ?

MonkeyCanCode: > @MonkeyCanCode : Thanks for the investigation!
> 
> From my side, I'm willing to look deeper into refactoring this code, but I do not really understand the problem (yet) 😅
> 
> Do you have repro instructions handy? Could you post them here or summarize in #3440 ?

Unfortunately, we don't have a reproducible thus the hesitation for major refactor. Detailed added in #3440. Thanks for looking this this. I will set this as not ready for review.

PR #3503, opened 2026-01-22
(doc) Outdated changelog

MonkeyCanCode:
Currently the CHANGELOG.md is outdated on the main branch and people are adding new changes under UNRELEASED which also includes changes for 1.3.0-incubating (and 1.3.0 is missing-incubating).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### New Features
 
+- Added KMS properties (optional) to catalog storage config to enable S3 data encryption.
 - Added `topologySpreadConstraints` support in Helm chart.
 - Added `priorityClassName` support in Helm chart.
 - Added support for including principal name in subscoped credentials. `INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` (default: false) can be used to toggle this feature. If enabled, cached credentials issued to one principal will no longer be available for others.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -98,7 +109,6 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### New Features
 
-- Added KMS properties (optional) to catalog storage config to enable S3 data encryption.
```
pingtimeout, line 101: This change surprised me at first.  We should not be modifying the 1.2.0-incubating section as part of this PR, ideally.  But it turns out the be necessary.

But it turns out that the changelog was updated in the wrong section for fb9397f1a2.  The `Added KMS properties (optional) to catalog storage config to enable S3 data encryption.` bullet point should have been added to the `Unreleased` section for inclusion in the 1.3.0 release notes.  But it had been added to the `1.2.0-incubating` section instead, which was already released at the time the PR was merged.

Could you re-add this bullet point, but in the `1.3.0-incubatiing` section?

pingtimeout (commented):
Thanks @MonkeyCanCode !  There is just one thing that needs fixed, and then we should be good to go.

CHANGELOG.md
MonkeyCanCode: Yeah, while i was working on couple PRs for the last couple of days, I noticed there is no 1.3.0-incubating and looking at the released one, those are there...but on the unreleased section instead. This specific line was not in 1.3.0-incubating. But as you confirmed above, I will move it.

CHANGELOG.md
```diff
@@ -51,40 +82,21 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 - The EclipseLink Persistence implementation has been completely removed.
 - The default request ID header name has changed from `Polaris-Request-Id` to `X-Request-ID`.
-- The (Before/After)CommitTableEvent has been removed.
 
 ### New Features
 
+- Added KMS properties (optional) to catalog storage config to enable S3 data encryption.
```
dimas-b: Hmmm... It does not look like KMS is in 1.3.0 🤔 

https://github.com/apache/polaris/blob/apache-polaris-1.3.0-incubating/spec/polaris-management-service.yml

MonkeyCanCode: That is what i thought...So I had it correct earlier. Reverted back to unreleased. Nice catch on this.

pingtimeout: Oh my bad...

PR #3504, opened 2026-01-22
Add jitpack.io repository

snazy:
The Iceberg PR https://github.com/apache/iceberg/pull/14991 bumps the https://github.com/RoaringBitmap/RoaringBitmap version, which is not published to Maven Central. We need to declare in the jitpack.io repository to prepare for the next Iceberg release.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -118,7 +118,10 @@ dependencyResolutionManagement {
   repositoriesMode = RepositoriesMode.FAIL_ON_PROJECT_REPOS
   repositories {
     mavenCentral()
-    maven { url = uri("https://jitpack.io") }
+    maven {
+      url = uri("https://jitpack.io")
+      content { includeModule("com.github.RoaringBitmap.RoaringBitmap", "roaringbitmap") }
+    }
     gradlePluginPortal()
   }
 }


snazy:
I suspect, other downstream consumers of the dependencies might need to be updated as well to declare the jitpack repository?

ajantha-bhat:
There are other open source projects that uses roaringbitmap too. 

Apache Spark also added jitpack depenedency while upgrading their roaringbitmap version
https://issues.apache.org/jira/browse/SPARK-54192

But it is unreleased (Will be effective from Spark 4.2). 

Spark community also agree that downstream projects will be affected
https://github.com/apache/spark/pull/52892#pullrequestreview-3434565472

settings.gradle.kts
```diff
@@ -118,6 +118,7 @@ dependencyResolutionManagement {
   repositoriesMode = RepositoriesMode.FAIL_ON_PROJECT_REPOS
   repositories {
     mavenCentral()
+    maven { url = uri("https://jitpack.io") }
```
adutra: How about restricting the usage of Jitpack to just RoaringBitmap?

```suggestion
    maven {
      url = uri("https://jitpack.io")
      content {
        includeModule("com.github.RoaringBitmap.RoaringBitmap", "roaringbitmap")
      }
    }
```

snazy: Good idea, added this.

singhpk234:
[doubt] why does polaris needs bitmap ? do we need to some ops that requires this ?

snazy:
> [doubt] why does polaris needs bitmap ? do we need to some ops that requires this ?

Iceberg + Spark pull that dependency in, so it is already transitively present in Polaris.

singhpk234:
~why can't we exclude it ?~

nvm i played around with trying exlude it, it might be better the way we have it for now :), thanks for the change again !

kevinjqliu:
btw RoaringBitmap is back in maven 
https://github.com/RoaringBitmap/RoaringBitmap/issues/749

snazy:
> btw RoaringBitmap is back in maven [RoaringBitmap/RoaringBitmap#749](https://github.com/RoaringBitmap/RoaringBitmap/issues/749)

yup! #3848 is already up ;)

PR #3506, opened 2026-01-22
[POC] Replace custom token bucket with Quarkus Bucket4J extension

adutra:
This change replaces the custom rate limiter implementation with the Quarkus Bucket4J extension, which provides a more robust and well-tested rate limiting solution.

The `RateLimiter` interface is preserved (since it might be an "extension point"), and the default implementation (`RealmTokenBucketRateLimiter`) now delegates to Bucket4J via a `@RateLimited` annotated method with per-realm segmentation.

Key changes:

- Add quarkus-bucket4j dependency to gradle/libs.versions.toml
- Create `RealmIdentityResolver` for per-realm rate limit segmentation
- Rewrite `RealmTokenBucketRateLimiter` to use Bucket4J via self-injection
- Add 'enabled' flag to `RateLimiterFilterConfiguration`
- Update configuration to use Bucket4J properties:
  - `quarkus.rate-limiter.buckets.polaris.limits[0].permitted-uses`
  - `quarkus.rate-limiter.buckets.polaris.limits[0].period`
- Update Helm chart with simplified configuration

## Checklist


snazy:
I think it's worth to investigate this approach as well, either in addition to or a replacement of #3507, as it would work for a horizontally-scaled Polaris instance.

adutra:
> I think it's worth to investigate this approach as well, either in addition to or a replacement of #3507, as it would work for a horizontally-scaled Polaris instance.

Yes, I think this approach only makes sense if/when there is desire for a distributed token bucket.

snazy:
> Yes, I think this approach only makes sense if/when there is desire for a distributed token bucket.

Are you saying that we should see how users want to implement rate-limiting in the wild?

adutra:
Closing the PoC for now as we went with Guava's RateLimiter.

PR #3507, opened 2026-01-22
Replace custom token-bucket implementation with Guava's `RateLimiter`

snazy:
Addresses the issues discussed on the [dev mailing-list discussion](https://lists.apache.org/thread/gkyw7m4fcbjbzhcrlrp4kcq5lr05r0m4), opting to use Guava as the easiest replacement here.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucketConfiguration.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucketConfiguration.java
@@ -19,8 +19,8 @@
 package org.apache.polaris.service.ratelimiter;
 
 import io.smallrye.config.ConfigMapping;
-
 import java.time.Duration;
+import java.util.Optional;
 
 @ConfigMapping(prefix = "polaris.rate-limiter.token-bucket")
 public interface TokenBucketConfiguration {
@@ -32,7 +32,7 @@ public interface TokenBucketConfiguration {
 
   /** This setting is no longer used and will be removed in a future release. */
   @Deprecated(since = "1.3.0", forRemoval = true)
-  Duration window();
+  Optional<Duration> window();
 
   /**
    * The type of the token bucket factory. Must be a registered {@link


runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucketConfiguration.java
```diff
@@ -19,15 +19,15 @@
...
 
 @ConfigMapping(prefix = "polaris.rate-limiter.token-bucket")
 public interface TokenBucketConfiguration {
 
+  /**
+   * Number of allowed requests per second per realm. The value <em>must</em> be greater than zero.
+   */
   long requestsPerSecond();
 
-  Duration window();
```
dimas-b, line 29: This could be a breaking change if older deployments defined a config value for it in `application.properties` (which is introspected and validated by Quarkus, IIRC). A missing "receiver" for those configs could be a runtime error (and config is under strict guarantees in our standing [evolution guidelines](https://polaris.apache.org/in-dev/unreleased/evolution/)).

Could we keep a deprecated (in code and in CHANGELOG) receiver just to improve backward compatibility?

helm/polaris/README.md
```diff
@@ -416,7 +416,6 @@ ct install --namespace polaris --charts ./helm/polaris
 | rateLimiter.tokenBucket | object | `{"requestsPerSecond":9999,"type":"default","window":"PT10S"}` | The configuration for the default rate limiter, which uses the token bucket algorithm with one bucket per realm. |
 | rateLimiter.tokenBucket.requestsPerSecond | int | `9999` | The maximum number of requests per second allowed for each realm. |
 | rateLimiter.tokenBucket.type | string | `"default"` | The type of the token bucket rate limiter. Only the default type is supported out of the box. |
-| rateLimiter.tokenBucket.window | string | `"PT10S"` | The time window. |
```
MonkeyCanCode, line 419: As this is not a backward compatible change, should we update CHANGELOG.md as well?

helm/polaris/templates/configmap.yaml
```diff
@@ -64,7 +64,6 @@ data:
     {{- if ne .Values.rateLimiter.type "no-op" -}}
     {{- $_ = set $map "polaris.rate-limiter.token-bucket.type" .Values.rateLimiter.tokenBucket.type -}}
     {{- $_ = set $map "polaris.rate-limiter.token-bucket.requests-per-second" .Values.rateLimiter.tokenBucket.requestsPerSecond -}}
-    {{- $_ = set $map "polaris.rate-limiter.token-bucket.window" .Values.rateLimiter.tokenBucket.window -}}
```
MonkeyCanCode, line 67: We will also need to update https://github.com/apache/polaris/blob/main/helm/polaris/tests/configmap_test.yaml:
```
./helm/polaris/tests/configmap_test.yaml:      - matchRegex: { path: 'data["application.properties"]', pattern: "polaris.rate-limiter.token-bucket.window=PT10S" }
./helm/polaris/tests/configmap_test.yaml:      - matchRegex: { path: 'data["application.properties"]', pattern: "polaris.rate-limiter.token-bucket.window=PT5S" }
```

snazy: Those are already removed (see below)

runtime/service/src/main/java/org/apache/polaris/service/ratelimiter/TokenBucketConfiguration.java
snazy: Re-added the `window` option as "deprecated for removal" and added a note to the CHANGELOG.

helm/polaris/README.md
snazy: Added notes to the CHANGELOG

helm/polaris/templates/configmap.yaml
snazy: I missed to remove the `window` setting right above in configmap_test tho

snazy:
Rebased due to a conflict w/ #3503

CHANGELOG.md
```diff
@@ -31,9 +31,12 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### Upgrade notes
 
+- The custom token-bucket based rate limiter has been replaced with Guava's rate limiter implementation.
+
 ### Breaking changes
 
 - The (Before/After)CommitTableEvent has been removed.
+- The configuration option `polaris.rate-limiter.token-bucket.window` is no longer supported and should be removed.
```
dimas-b: Should this be under `Deprecations` for now? We've not removed the "receiver" for this config, only deprecated 🤔

snazy: I thought about that, too.
However, people should remove it now (or better: with the next release) to be safe for the release after that.
Sure, it's not really breaking, however, the parameter is no longer used - could be a interpreted as "a soft breaking change" (not sure there's a better term for that).

adutra: I think I agree with @dimas-b, the `Deprecations` section seems like a better fit for this announcement. 

When the property is fully removed, then we would put a note under this section (since that would be a hard breaking change).

I see that this is what we did, e.g. for the EclipseLink phase out.

snazy: okay, moved

snazy:
Had to rebase again

snazy:
Thanks for the reviews :)

PR #3514, opened 2026-01-23
Update Gradle to 9.3.0

snazy:


jbonofre:
I'm curious about the JDK version (unrelated). Maybe we should discuss to update Polaris to use JDK17 min. Just wondering as Gradle 9.x requires JDK 17.

snazy:
I see there could be some confusion here.
The Polaris Gradle build runs _with_ Java 21, but certain artifacts are built _for_ Java 11 (Via the `javac` `-release` option).
Setting Java 17 or even 21 as the baseline works for me tho.

jbonofre:
Yes, agree. As I said, this is an unrelated, just stupid wondering 😄

PR #3517, opened 2026-01-23
Let `site/bin/checkout-releases.sh` pull the latest state

snazy:
The script `site/bin/checkout-releases.sh` is a convenience to get the `versioned-docs` branch locally. It is missing a `git pull` to get the latest state of that branch though, which can be confusing.

### DIFF
--- a/site/bin/checkout-releases.sh
+++ b/site/bin/checkout-releases.sh
@@ -30,3 +30,5 @@ if [[ -d content/releases ]] ; then
 fi
 
 git worktree add content/releases versioned-docs
+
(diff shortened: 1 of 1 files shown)


site/bin/checkout-releases.sh
```diff
@@ -30,3 +30,5 @@ if [[ -d content/releases ]] ; then
 fi
 
 git worktree add content/releases versioned-docs
+
+(cd content/releases ; git pull)
```
MonkeyCanCode, line 34: nit: `git -C content/releases pull`

PR #3520, opened 2026-01-23
Site-workflow: add `workflow_call` trigger

### LINKED ISSUE #3516: Eagerly trigger website publication when pushing to the `versioned-docs` branch

The `site.yml` workflow is currently triggered in the following scenarios:
1. A push to the `main` branch, using the state of the site and the workflow as on `main`.
2. A push to a `release/*`branch, using the state of the site and the workflow as on that release branch.

Notice that workflows get the repo state (from the checkout actions) and the workflow state as its on that particular branch.
Put in other words: if we'd have a change to some old `release/1.0.x` branch, the web site would be updated as it is defined on that `release/1.0.x` branch, which is wrong and not the intended behavior.

Therefore, changes to the `versioned-docs` branch do not trigger the website publication, because the workflow that does this, the [`site` workflow](https://github.com/apache/polaris/blob/d36e88e6fcff2137eff26f5e3bf2c0632bef3d7b/.github/workflows/site.yml), is on the `main` branch.
Simply speaking: a push to `versioned-docs` doesn't trigger any workflow, because from that branch's perspective, there is just no workflow.

I'll come up with a PR to fix the triggers of the site workflow and in parallel work on a solution to get the website pushed when a push to the `versioned-docs` branch happens. The latter is a bit more complicated and security implications need to be considered.

_No response_

_No response_

### Additional context

_No response_

snazy:
This trigger allows the Hugo Site publication workflow to be called from the `versioned-docs` branch.
    
This change also introduces a concurrency group to prevent concurrent site publications.
    
Fixes #3516

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -31,7 +31,7 @@ on:
 # Ensure that no two site publications happen concurrently.
 # 1 job per reference can run concurrently, no cancellation
 concurrency:
-  group: ${{ github.ref }}
+  group: site-publishing
   cancel-in-progress: false
 
 jobs:


snazy:
Thanks for the approvals!
I still want to double-check the approach in a toy repo before I merge the PRs.

.github/workflows/site.yml
```diff
@@ -68,7 +72,7 @@ jobs:
         run: sudo cp .asf.yaml site/build/apache-site/
 
       - name: Deploy Static Site to GitHub
-        if: (github.event_name == 'push' && github.repository == 'apache/polaris') || github.event_name == 'workflow_call'
+        if: github.event_name == 'push' && github.repository == 'apache/polaris'
```
snazy, line 75: The `github` context is populated with the information of the calling workflow

.github/workflows/site.yml
```diff
@@ -23,15 +23,33 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+  workflow_call:
+    # The workflow is programmatically "protected" to be only callable from the 'apache/polaris' repo.
+    # Beware: this workflow will be called with with the `github` context from the 'versioned-docs' branch,
```
dimas-b:
```suggestion
    # Beware: this workflow will be called with the `github` context from the 'versioned-docs' branch,
```

.github/workflows/site.yml
```diff
@@ -23,15 +23,33 @@ on:
...
+    # which means, for example, that the `github.ref` would be 'refs/heads/versioned-docs'.
+
+# Ensure that no two site publications happen concurrently.
+# 1 job per reference can run concurrently, no cancellation
+concurrency:
+  group: ${{ github.ref }}
```
dimas-b: So, `github.ref` could be `refs/heads/versioned-docs` (when called programmatically) or `main`... and those jobs can still run concurrently? Is that expected?

snazy: Good point - updated this

PR #3523, opened 2026-01-24
(feat: persistence) Add schema-metrics-v1.sql for metrics tables (#3337)

obelix74:
Add new schema version 4 with tables for storing scan and commit metrics reports as first-class entities.

New tables:
- scan_metrics_report: Stores scan metrics with trace correlation
- scan_metrics_report_roles: Junction table for principal roles
- commit_metrics_report: Stores commit metrics with trace correlation
- commit_metrics_report_roles: Junction table for principal roles

Key design decisions:
- PRIMARY KEY (realm_id, report_id) for multi-tenancy
- Junction tables with CASCADE DELETE for roles
- Timestamp index for retention cleanup
- JSONB metadata column for extensibility (Postgres), TEXT for H2

## Checklist

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
@@ -60,4 +60,24 @@ public enum DatabaseType {
     ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
     return classLoader.getResourceAsStream(resourceName);
   }
+
+  /**
+   * Open an InputStream that contains data from the metrics schema init script. This stream should
+   * be closed by the caller.
+   *
+   * @param metricsSchemaVersion the metrics schema version (currently only 1 is supported)
+   * @return an InputStream for the metrics schema SQL file
+   */
+  public InputStream openMetricsSchemaResource(int metricsSchemaVersion) {
+    if (metricsSchemaVersion != 1) {
+      throw new IllegalArgumentException(
+          "Unknown or invalid metrics schema version " + metricsSchemaVersion);
+    }
+
+    final String resourceName =
+        String.format("%s/schema-metrics-v%d.sql", this.getDisplayName(), metricsSchemaVersion);
+
+    ClassLoader classLoader = DatasourceOperations.class.getClassLoader();
+    return classLoader.getResourceAsStream(resourceName);
+  }
 }
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -172,6 +172,13 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
               datasourceOperations
                   .getDatabaseType()
                   .openInitScriptResource(effectiveSchemaVersion));
+
+          // Run the metrics schema script if requested
+          if (JdbcBootstrapUtils.shouldIncludeMetrics(bootstrapOptions)) {
+            LOGGER.info("Including metrics schema for realm: {}", realm);
+            datasourceOperations.executeScript(
+                datasourceOperations.getDatabaseType().openMetricsSchemaResource(1));
+          }
         } catch (SQLException e) {
           throw new RuntimeException(
               String.format("Error executing sql script: %s", e.getMessage()), e);
--- a/persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
+++ b/persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
@@ -1,0 +1,166 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file--
+--  distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"). You may not use this file except in compliance
+-- with the License.  You may obtain a copy of the License at
+--
+--  http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+-- This schema is SEPARATE from the entity schema and can evolve independently.
+-- It contains tables for storing Iceberg metrics reports.
+--
+-- Tables:
+--   * `metrics_version` - Version tracking for the metrics schema
+--   * `scan_metrics_report` - Scan metrics reports
+--   * `commit_metrics_report` - Commit metrics reports
+-- ============================================================================
+
+CREATE SCHEMA IF NOT EXISTS POLARIS_SCHEMA;
+SET SCHEMA POLARIS_SCHEMA;
+
+-- Metrics schema version tracking (separate from entity schema version)
+CREATE TABLE IF NOT EXISTS metrics_version (
+    version_key VARCHAR PRIMARY KEY,
+    version_value INTEGER NOT NULL
+);
+
+MERGE INTO metrics_version (version_key, version_value)
+    KEY (version_key)
+    VALUES ('metrics_version', 1);
+
+COMMENT ON TABLE metrics_version IS 'the version of the metrics schema in use';
+
+-- ============================================================================
+-- SCAN METRICS REPORT TABLE
+-- ============================================================================
+
+CREATE TABLE IF NOT EXISTS scan_metrics_report (
+    report_id TEXT NOT NULL,
+    realm_id TEXT NOT NULL,
+    catalog_id BIGINT NOT NULL,
+    table_id BIGINT NOT NULL,
+
+    -- Report metadata
+    timestamp_ms BIGINT NOT NULL,
+    principal_name TEXT,
+    request_id TEXT,
+
+    -- Trace correlation
+    otel_trace_id TEXT,
+    otel_span_id TEXT,
+    report_trace_id TEXT,
+
+    -- Scan context
+    snapshot_id BIGINT,
+    schema_id INTEGER,
+    filter_expression TEXT,
+    projected_field_ids TEXT,
+    projected_field_names TEXT,
+
+    -- Scan metrics
+    result_data_files BIGINT DEFAULT 0,
+    result_delete_files BIGINT DEFAULT 0,
+    total_file_size_bytes BIGINT DEFAULT 0,
+    total_data_manifests BIGINT DEFAULT 0,
+    total_delete_manifests BIGINT DEFAULT 0,
+    scanned_data_manifests BIGINT DEFAULT 0,
+    scanned_delete_manifests BIGINT DEFAULT 0,
+    skipped_data_manifests BIGINT DEFAULT 0,
+    skipped_delete_manifests BIGINT DEFAULT 0,
+    skipped_data_files BIGINT DEFAULT 0,
+    skipped_delete_files BIGINT DEFAULT 0,
+    total_planning_duration_ms BIGINT DEFAULT 0,
+
+    -- Equality/positional delete metrics
+    equality_delete_files BIGINT DEFAULT 0,
+    positional_delete_files BIGINT DEFAULT 0,
+    indexed_delete_files BIGINT DEFAULT 0,
+    total_delete_file_size_bytes BIGINT DEFAULT 0,
+
+    -- Additional metadata (for extensibility)
+    metadata TEXT DEFAULT '{}',
+
+    PRIMARY KEY (realm_id, report_id)
+);
+
+COMMENT ON TABLE scan_metrics_report IS 'Scan metrics reports as first-class entities';
+
+-- Index for retention cleanup by timestamp
+CREATE INDEX IF NOT EXISTS idx_scan_report_timestamp ON scan_metrics_report(realm_id, timestamp_ms);
+
+-- ============================================================================
+-- COMMIT METRICS REPORT TABLE
+-- ============================================================================
+
+CREATE TABLE IF NOT EXISTS commit_metrics_report (
+    report_id TEXT NOT NULL,
+    realm_id TEXT NOT NULL,
+    catalog_id BIGINT NOT NULL,
+    table_id BIGINT NOT NULL,
+
+    -- Report metadata
+    timestamp_ms BIGINT NOT NULL,
+    principal_name TEXT,
+    request_id TEXT,
+
+    -- Trace correlation
+    otel_trace_id TEXT,
+    otel_span_id TEXT,
+    report_trace_id TEXT,
+
+    -- Commit context
+    snapshot_id BIGINT NOT NULL,
+    sequence_number BIGINT,
+    operation TEXT NOT NULL,
+
+    -- File metrics
+    added_data_files BIGINT DEFAULT 0,
+    removed_data_files BIGINT DEFAULT 0,
+    total_data_files BIGINT DEFAULT 0,
+    added_delete_files BIGINT DEFAULT 0,
+    removed_delete_files BIGINT DEFAULT 0,
+    total_delete_files BIGINT DEFAULT 0,
+
+    -- Equality delete files
+    added_equality_delete_files BIGINT DEFAULT 0,
+    removed_equality_delete_files BIGINT DEFAULT 0,
+
+    -- Positional delete files
+    added_positional_delete_files BIGINT DEFAULT 0,
+    removed_positional_delete_files BIGINT DEFAULT 0,
+
+    -- Record metrics
+    added_records BIGINT DEFAULT 0,
+    removed_records BIGINT DEFAULT 0,
+    total_records BIGINT DEFAULT 0,
+
+    -- Size metrics
+    added_file_size_bytes BIGINT DEFAULT 0,
+    removed_file_size_bytes BIGINT DEFAULT 0,
+    total_file_size_bytes BIGINT DEFAULT 0,
+
+    -- Duration and attempts
+    total_duration_ms BIGINT DEFAULT 0,
+    attempts INTEGER DEFAULT 1,
+
+    -- Additional metadata (for extensibility)
+    metadata TEXT DEFAULT '{}',
+
+    PRIMARY KEY (realm_id, report_id)
+);
+
+COMMENT ON TABLE commit_metrics_report IS 'Commit metrics reports as first-class entities';
+
+-- Index for retention cleanup by timestamp
+CREATE INDEX IF NOT EXISTS idx_commit_report_timestamp ON commit_metrics_report(realm_id, timestamp_ms);
--- a/persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
+++ b/persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
@@ -1,0 +1,177 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file--
+--  distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"). You may not use this file except in compliance
+-- with the License.  You may obtain a copy of the License at
+--
+--  http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- This schema is SEPARATE from the entity schema and can evolve independently.
+-- It contains tables for storing Iceberg metrics reports.
+--
+-- Tables:
+--   * `metrics_version` - Version tracking for the metrics schema
+--   * `scan_metrics_report` - Scan metrics reports
+--   * `commit_metrics_report` - Commit metrics reports
+-- ============================================================================
+
+CREATE SCHEMA IF NOT EXISTS POLARIS_SCHEMA;
+SET search_path TO POLARIS_SCHEMA;
+
+-- Metrics schema version tracking (separate from entity schema version)
+CREATE TABLE IF NOT EXISTS metrics_version (
+    version_key TEXT PRIMARY KEY,
+    version_value INTEGER NOT NULL
+);
+
+INSERT INTO metrics_version (version_key, version_value)
+VALUES ('metrics_version', 1)
+ON CONFLICT (version_key) DO UPDATE
+SET version_value = EXCLUDED.version_value;
+
+COMMENT ON TABLE metrics_version IS 'the version of the metrics schema in use';
+
+-- ============================================================================
+-- SCAN METRICS REPORT TABLE
+-- ============================================================================
+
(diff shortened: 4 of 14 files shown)


persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
```diff
@@ -0,0 +1,350 @@
...
+    ON commit_metrics_report(realm_id, timestamp_ms DESC);
+
+-- Junction table for commit metrics report roles
+CREATE TABLE IF NOT EXISTS commit_metrics_report_roles (
+    realm_id TEXT NOT NULL,
+    report_id TEXT NOT NULL,
+    role_name TEXT NOT NULL,
+    PRIMARY KEY (realm_id, report_id, role_name),
+    FOREIGN KEY (realm_id, report_id) REFERENCES commit_metrics_report(realm_id, report_id) ON DELETE CASCADE
+);
```
singhpk234: porting my comment from previous pr here, do we need a new table just for this ? @dimas-b thoughts

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
```diff
@@ -0,0 +1,350 @@
...
+    catalog_id TEXT NOT NULL,
+    event_id TEXT NOT NULL,
+    request_id TEXT,
+    event_type TEXT NOT NULL,
+    timestamp_ms BIGINT NOT NULL,
+    principal_name TEXT,
+    resource_type TEXT NOT NULL,
+    resource_identifier TEXT NOT NULL,
+    additional_properties JSONB NOT NULL DEFAULT '{}'::JSONB,
+    PRIMARY KEY (event_id)
```
singhpk234, line 136: [not related to pr] oh no :( we forgot `realm-id` in v3

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
obelix74: Something like this?

```
   CREATE TABLE scan_metrics_report (
       ...
       roles JSONB DEFAULT '[]'::JSONB,  -- or TEXT for H2
       ...
   );

   Pros:
     • No extra tables
     • Single INSERT per report
     • Works with both Postgres (JSONB) and H2 (TEXT)
     • Can query with JSON operators: WHERE roles @> '["analyst"]'

   Cons:
     • Harder to index efficiently
     • JSON parsing overhead
     • Less type-safe
```

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
obelix74: I see that we are already using JSONB for additional_properties here, so perhaps it is ok to do that for roles?

BTW, this `PRIMARY KEY` gap you identified - would that trigger a schema change?

singhpk234: This is unrelated for the change and more from the context of we didn't do this when we introduced events, was more from an FYI to the other reviewers pov

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
singhpk234: i believe the debate then is adding it in schema directly of the scan_metric report or normalizing the roles info into seperate i believe its fine to have role_name if we have principal name in the schema already, if we wanna be cautious we can just move the prinicipal_name and role name in the additional properties. 

H2 is just for test, for PG we can make indexes on the JSONB columns

obelix74: Sounds good @singhpk234. @dimas-b please let us know your thoughts.

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
```diff
@@ -0,0 +1,350 @@
...
+);
+
+-- Idempotency records (from v3)
```
dimas-b: This is not longer in v3 per #3386

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
dimas-b: TBH, I lost context on the RDBMS schema for metrics 😅 

Do we have a description (somewhere) of anticipated query patterns in Polaris and whether we expect external (intependent) queries against this schema?

dimas-b:
@obelix74 @singhpk234 : WDYT about starting an RFC doc + `dev` thread on this? I believe a structured overview of this feature would be good to set the stage for PRs :) (apologies if I missed it)

singhpk234:
@dimas-b there is a dev thread already please ref : https://lists.apache.org/thread/c83jnkvlwc2k3swm65cmvl4t0mt7p799
thanks @obelix74 for the the writing this up !

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
obelix74: I have merged my PR with the latest from `main`. My PR now only contains the metrics tables I added, and a potential fix for the correct version.

obelix74:
> @obelix74 @singhpk234 : WDYT about starting an RFC doc + `dev` thread on this? I believe a structured overview of this feature would be good to set the stage for PRs :) (apologies if I missed it)

I am trying to solve two sets of asks from my product folks with this.

1. Metrics - what tables were accessed by a client principal 
2. Auditing - which user accessed what data and why

From the metrics perspective, today, with `1.3.0`, I want to be able to report on table metrics based on:

Track table scan operations:

- by table
- by snapshot
- by time range
- by realm
- by user principal
- by engine

For commit report queries:

- by operation type
- data growth
- file churn
- storage analysis

Also many operational dashboards, and filtering by user, realm, engine name, version etc.

I have not thought about roles in this flow at all, perhaps it will be useful.  @singhpk234 recommended adding roles and I added them. I normalized the roles tables from a RDBMS perspective, but I didn't realize there are other similar fields stored as JSON already.

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
```diff
@@ -31,7 +36,7 @@ CREATE TABLE IF NOT EXISTS version (
 
 MERGE INTO version (version_key, version_value)
     KEY (version_key)
-    VALUES ('version', 3);
+    VALUES ('version', 4);
```
singhpk234, line 34: Thanks for catching this @obelix74 !

dimas-b:
@obelix74 : please rebase to fix CI

dimas-b:
Let's hold final review until #3616 is resolved... Intermediate comments are welcome, of course :)

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
```diff
@@ -0,0 +1,207 @@
...
+COMMENT ON COLUMN commit_metrics_report.realm_id IS 'Realm ID for multi-tenancy';
+COMMENT ON COLUMN commit_metrics_report.operation IS 'Commit operation type: append, overwrite, delete, replace';
+COMMENT ON COLUMN commit_metrics_report.otel_trace_id IS 'OpenTelemetry trace ID from HTTP headers';
+
+-- Index for retention cleanup by timestamp
+CREATE INDEX IF NOT EXISTS idx_commit_report_timestamp
+    ON commit_metrics_report(realm_id, timestamp_ms DESC);
+
+-- Junction table for commit metrics report roles
+CREATE TABLE IF NOT EXISTS commit_metrics_report_roles (
```
obelix74: @dimas-b there was an open question from @singhpk234 about this. He was asking us if we need a dedicated table for this or if we can store role ids as a JSON [] in commit_metrics_report.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -172,6 +172,13 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
               datasourceOperations
                   .getDatabaseType()
                   .openInitScriptResource(effectiveSchemaVersion));
+
+          // Run the metrics schema script if requested
+          if (JdbcBootstrapUtils.shouldIncludeMetrics(bootstrapOptions)) {
+            LOGGER.info("Including metrics schema for realm: {}", realm);
+            datasourceOperations.executeScript(
+                datasourceOperations.getDatabaseType().openMetricsSchemaResource(1));
```
obelix74, line 180: I think this is all we need to do to source the new schema file.

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
```diff
@@ -28,7 +28,7 @@ CREATE TABLE IF NOT EXISTS version (
     version_value INTEGER NOT NULL
 );
 INSERT INTO version (version_key, version_value)
-VALUES ('version', 3)
+VALUES ('version', 4)
```
dimas-b, line 31: Should we perhaps track metrics schema version separately?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
dimas-b: This is pretty close to the ideal from my POV 😅 Would it be too much trouble to apply the metrics DDL in a metrics persistence class (separate from the JDBC MetaStore, which deals with Entities)?

This may create a chicken-and-egg problem with respect to #3385 because we need the schema to run persistence, and we need persistence classes to create the schema 😅 

I think we may want to merge this "as is" and later move DDL to Metrics Persistence classes. `BaseMetaStoreCommand` will have to have a new bean injected to deal with metrics DDL.

WDYT?

dimas-b (approved):
LGTM. Let's give this PR a couple more days in review for other people to comment if they want.

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
dimas-b: I'm afraid, I do not understand this use case at all :sweat_smile: Why do we link roles to commit metrics?

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
obelix74: It is tracked separately here. https://github.com/apache/polaris/pull/3523/changes#diff-ecb303a29ddb64046d6eab1e6721f99a09d7af0986675934d8377c2c5be0695cR45

There is a bug in the current code checked into main, this version should be `4`.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
obelix74: Sounds good.  Let us merge this as is and move the DDL later.

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
obelix74: @singhpk234 suggested we track the roles as well with each table metrics, in addition to realm, principal and other information, but there can be more than 1 role, so this table. Perhaps just store this as a JSON array in the table itself?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatabaseType.java
```diff
@@ -50,7 +50,7 @@ public static DatabaseType fromDisplayName(String displayName) {
    */
   public InputStream openInitScriptResource(int schemaVersion) {
     // Preconditions check is simpler and more direct than a switch default
-    if (schemaVersion <= 0 || schemaVersion > 3) {
+    if (schemaVersion <= 0 || schemaVersion > 4) {
```
obelix74, line 53: This is another bug in the main branch.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
obelix74: Makes sense.

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
```diff
@@ -94,6 +92,9 @@ CREATE TABLE IF NOT EXISTS scan_metrics_report (
     indexed_delete_files BIGINT DEFAULT 0,
     total_delete_file_size_bytes BIGINT DEFAULT 0,
 
+    -- Principal roles (denormalized as JSON array)
+    principal_role_ids JSONB DEFAULT '[]'::JSONB,
```
obelix74: @dimas-b is this better?

persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
```diff
@@ -0,0 +1,177 @@
...
+    added_file_size_bytes BIGINT DEFAULT 0,
+    removed_file_size_bytes BIGINT DEFAULT 0,
+    total_file_size_bytes BIGINT DEFAULT 0,
+
+    -- Duration and attempts
+    total_duration_ms BIGINT DEFAULT 0,
+    attempts INTEGER DEFAULT 1,
+
+    -- Principal roles (denormalized as JSON array)
+    principal_role_ids TEXT DEFAULT '[]',
```
singhpk234: why ids and not names ?

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
```diff
@@ -93,6 +93,11 @@ static class SchemaInputOptions {
           paramLabel = "<schema version>",
           description = "The version of the schema to load in [1, 2, 3, LATEST].")
       Integer schemaVersion;
+
+      @CommandLine.Option(
+          names = {"--include-metrics"},
+          description = "Include metrics schema tables during bootstrap.")
+      boolean includeMetrics;
```
singhpk234, line 100: can we add tests for it ? 

Also i believe its fine to create a table as part of default bootstrap, we do the same for `events` which is by default off

singhpk234: I don;t have a strong opinion for vs against it

persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
obelix74: Can principal names change?  In the SPI we standardized on table ids and catalog ids because of this.

runtime/admin/src/main/java/org/apache/polaris/admintool/BootstrapCommand.java
obelix74: Added tests.

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
dimas-b: let's fix the version number bug in the main schema file in a separate PR then?

persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
dimas-b: Sorry, I missed out on this aspect... Why are we storing role IDs with the commit reports in the first place? 🤔

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
dimas-b: Let's continue in https://github.com/apache/polaris/pull/3523/changes#r2772661384

persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
obelix74: @dimas-b, Prashant suggested we track roles as useful information.

dimas-b: But what's the use case? Is it in the doc?

dimas-b: To expand on my concern: Polaris supports external IdP and PDP (e.g. Keycloak and OPA), whatever roles may be put here by Metrics Persistence code may not be aligned with AuthZ decisions in any way :thinking:

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
obelix74: @dimas-b I have raised https://github.com/apache/polaris/pull/3690 for the version number bugs. I will remove them from this PR.

dimas-b: thx!

persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
obelix74: @dimas-b no, it is not in the doc. I hadn't thought of it. The suggestion came from @singhpk234 and I thought it was a good idea. I had forgotten about the external IdP and OPA changes (that's the other area I am working on), great call on that one. Roles may not be enforced by Polaris at all.  I think we should remove them from here or make them optional. I am leaning towards removing them.  I will push out a commit with these removed. @singhpk234 let us know if you disagree.

obelix74: This is now removed in a new commit.

dimas-b: Thx! Let's proceed incrementally :) First add what is really required, then follow-up with more involved use cases :sweat_smile:

persistence/relational-jdbc/src/main/resources/h2/schema-metrics-v1.sql
```diff
@@ -52,7 +52,6 @@ CREATE TABLE IF NOT EXISTS scan_metrics_report (
     report_id TEXT NOT NULL,
     realm_id TEXT NOT NULL,
     catalog_id BIGINT NOT NULL,
-    namespace TEXT NOT NULL,
```
obelix74: • Removed namespace TEXT NOT NULL column from both H2 and Postgres schema files

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
```diff
@@ -0,0 +1,180 @@
...
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- ============================================================================
+-- POLARIS METRICS SCHEMA VERSION 1 (PostgreSQL)
```
dimas-b: nit: `VERSION 1` is implied by the file name and will have to be updated for subsequent versions 🤷

dimas-b (approved):
LGTM 👍 

@singhpk234 : WDYT?

persistence/relational-jdbc/src/main/resources/postgres/schema-metrics-v1.sql
obelix74: Addressed.

polaris-core/src/main/java/org/apache/polaris/core/persistence/MetaStoreManagerFactory.java
```diff
@@ -45,4 +45,18 @@ default Map<String, PrincipalSecretsResult> bootstrapRealms(BootstrapOptions boo
...
+   * Checks if metrics persistence has been bootstrapped. This validation should be called at
+   * startup when metrics persistence is enabled.
+   *
+   * <p>The default implementation does nothing. Persistence backends that support metrics should
+   * override this to check if the required metrics tables exist.
+   *
+   * @throws IllegalStateException if metrics persistence is enabled but the required tables have
+   *     not been bootstrapped
+   */
+  default void checkMetricsPersistenceBootstrapped() {
```
dimas-b: I'm very hesitant to make this addition to the metastore factory when the general direction on this feature is to have an independent persistence for metrics 🤔 

Please consider using a `ProductionReadinessCheck` instead. It should be possible to WARN or even fail on startup if feature flags are not aligned with schema.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcProductionReadinessChecks.java
```diff
@@ -55,4 +56,50 @@ public ProductionReadinessCheck checkRelationalJdbc(
...
+  @Produces
+  public ProductionReadinessCheck checkMetricsPersistenceBootstrapped(
+      MetaStoreManagerFactory metaStoreManagerFactory,
+      Instance<DataSource> dataSource,
+      RelationalJdbcConfiguration relationalJdbcConfiguration,
+      @ConfigProperty(name = "polaris.persistence.metrics.enabled", defaultValue = "false")
+          boolean metricsPersistenceEnabled) {
+    // This check should only be applicable when persistence uses RelationalJdbc.
+    if (!(metaStoreManagerFactory instanceof JdbcMetaStoreManagerFactory)) {
+      return ProductionReadinessCheck.OK;
```
dimas-b: This check does not look correct to me. We should test the Metrics Persistence impl.. I think.

... but it may not be available in this PR... Perhaps we should move this code to #3385? WDYT?

dimas-b: otherwise, the check code LGTM 👍

polaris-core/src/main/java/org/apache/polaris/core/persistence/MetaStoreManagerFactory.java
obelix74: Addressed these comments.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/RelationalJdbcProductionReadinessChecks.java
obelix74: I had to rework this in this commit. https://github.com/apache/polaris/pull/3523/changes/dfaa124161a72119c8ad840acb3950fb24794bb8.

obelix74: yeah, it has the same problem. I think it has to move to https://github.com/apache/polaris/pull/3385. On it.

obelix74: @dimas-b Moved it to https://github.com/apache/polaris/pull/3385. https://github.com/apache/polaris/pull/3385/changes/16abf1b187acd1bd8b2722e4c7eb3f92d7421544#diff-2d3092986da2282edf801ee580cdd68d707e31f78de1a654af7953d2bcd4bd8aR92.  Thanks again.

runtime/common/build.gradle.kts
```diff
@@ -24,6 +24,9 @@ plugins {
 
 dependencies {
   compileOnly(libs.smallrye.config.core)
+  compileOnly(libs.jakarta.enterprise.cdi.api)
+  compileOnly(libs.jakarta.inject.api)
+  implementation(project(":polaris-core"))
```
dimas-b: still needed?

obelix74: Removed.

PR #3525, opened 2026-01-24
Sanitize principal names in AWS STS role session names

yushesp:
Principal names containing invalid characters (spaces, parentheses, etc.) were causing AWS STS AssumeRole requests to fail with validation errors. AWS STS role session names must match the pattern [\w+=,.@-]*.

This change:
- Adds AwsRoleSessionNameSanitizer utility class to sanitize strings for use as AWS STS role session names
- Replaces invalid characters with underscores and truncates to 64 characters (AWS maximum)
- Updates AwsCredentialsStorageIntegration to sanitize principal names when INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL is enabled
- Adds tests to verify sanitization behavior and AWS pattern compliance

Fixes issue where principal names like `Joe (local)` would produce invalid role session names like `Polaris-Joe (local)` and cause AssumeRole to fail. Now sanitized to `polaris-Joe__local_`.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
@@ -103,17 +103,15 @@ public class AwsCredentialsStorageIntegration
 
     String roleSessionName =
         includePrincipalNameInSubscopedCredential
-            ? "polaris-" + polarisPrincipal.getName()
+            ? AwsRoleSessionNameSanitizer.sanitize("polaris-" + polarisPrincipal.getName())
             : "PolarisAwsCredentialsStorageIntegration";
-    String cappedRoleSessionName =
-        roleSessionName.substring(0, Math.min(roleSessionName.length(), 64));
 
     if (shouldUseSts(storageConfig)) {
       AssumeRoleRequest.Builder request =
           AssumeRoleRequest.builder()
               .externalId(storageConfig.getExternalId())
               .roleArn(storageConfig.getRoleARN())
-              .roleSessionName(cappedRoleSessionName)
+              .roleSessionName(roleSessionName)
               .policy(
                   policyString(
                           storageConfig,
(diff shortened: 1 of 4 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -103,17 +103,15 @@ public StorageAccessConfig getSubscopedCreds(
 
     String roleSessionName =
         includePrincipalNameInSubscopedCredential
-            ? "polaris-" + polarisPrincipal.getName()
+            ? AwsRoleSessionNameSanitizer.sanitize("polaris-" + polarisPrincipal.getName())
```
dimas-b, line 106: The problem this change fixes is specific to AWS, AFAIK, but the benefits of including the exact principal name may be applicable to other S3 systems, where the session name is less restricted. Cf. #3224

@tokoko : WDYT? Is having exact principal names critical for your use cases?

@yushesp : In your use cases, do you actually need some Principal info in the session name, or could you exclude it via the `INCLUDE_PRINCIPAL_NAME_IN_SUBSCOPED_CREDENTIAL` flag?

tokoko: exact names are preferable of course, but in case the name causes a fail, sanitizing it makes sense. we're effectively kind of already doing the same by capping the length of the role name. The principal name included could be different for really really long principal names.

tokoko: for our use case, we're using both s3 and minio, so erring on the side of s3 restrictions sounds good. for the general case, assuming s3 restrictions makes sense for me as well, otherwise we'd have to either 1) try a call w/o sanitization and fall back on sanitization or 2) try coding restrictions for each system. Neither seems like a good choice.

yushesp: @dimas-b We do need to inject the principal name into the session name. We enabled the feature flag and ran into an error because my principal name had a special character.

yushesp:
@dimas-b is this PR good to be merged?

dimas-b:
Thanks for the ping, @yushesp :wink:

PR #3553, opened 2026-01-26
CI for guides

snazy:
This change adds the ability to automatically test the code snippets in the guides on the web site.

A description of the functionality is included in this PR in the added `site/it/README.md`.

The added GitHub workflow is intentionally not (yet) tied into the existing `ci.yml` workflow to first gain some experience with the tests, also to prevent that the "weird" issue described in #3853 can prevent other PRs from being merged.

### DIFF
--- a/site/it/site_checks/docker.py
+++ b/site/it/site_checks/docker.py
@@ -37,31 +37,24 @@ def cleanup_docker(tee: Tee) -> float:
     try:
         tee.printf("::group::Cleanup Docker containers, networks, volumes")
         tee.printf("Purging Docker containers...")
-        try:
-            result = subprocess.run(
-                ["docker", "ps", "-a", "-q"],
-                capture_output=True,
-                text=True,
-                check=False,
-            )
-            container_ids = [line for line in result.stdout.splitlines() if line.strip()]
-            if container_ids:
-                tee.run(["docker", "container", "rm", "-f", *container_ids])
-        except FileNotFoundError:
-            pass
+        result = subprocess.run(
+            ["docker", "ps", "-a", "-q"],
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        container_ids = [line for line in result.stdout.splitlines() if line.strip()]
+        if container_ids:
+            tee.run(["docker", "container", "rm", "-f", *container_ids])
         tee.printf("Purging Docker networks...")
-        try:
-            tee.run(["docker", "network", "prune", "-f"])
-        except FileNotFoundError:
-            pass
+        tee.run(["docker", "network", "prune", "-f"])
         tee.printf("Purging Docker volumes...")
-        try:
-            tee.run(["docker", "volume", "prune", "-f"])
-        except FileNotFoundError:
-            pass
-        tee.printf("::endgroup::")
-    finally:
-        return time.monotonic() - start
+        tee.run(["docker", "volume", "prune", "-f"])
+    except Exception as e:
+        tee.printf(f"Failed to cleanup Docker containers, networks, volumes: {repr(e)}")
+        pass
+    tee.printf("::endgroup::")
+    return time.monotonic() - start
 
 
 def docker_compose_info(tee: Tee, md_base_name: str, md_dir_name: str) -> float:
--- a/site/it/site_checks/markdown_testing.py
+++ b/site/it/site_checks/markdown_testing.py
@@ -54,7 +54,7 @@ def run_test(
     build_tests_dir: Path,
     test_summary_file: Path,
     env: dict[str, str],
-    code_blocks_global: CodeBlocksGlobal
+    code_blocks_global: CodeBlocksGlobal,
 ) -> bool:
     """
     Exercises testing of a single Markdown file.
@@ -117,12 +117,14 @@ def run_test(
         docker_info_duration = docker_compose_info(tee, md_base_name, md_dir_name)
         cleanup_duration = cleanup_docker(tee)
 
-        tee.printf(dedent(f"""
-            Test steps durations for for {md_file} ({'❌ FAILED' if failed else '✅ passed'}):
+        tee.printf(
+            dedent(f"""
+            Test steps durations for for {md_file} ({"❌ FAILED" if failed else "✅ passed"}):
                 .. Test: {duration_str}
                 .. Docker Compose Info: {_format_duration(docker_info_duration)}
                 .. Docker Cleanup: {_format_duration(cleanup_duration)}
-            """))
+            """)
+        )
         tee.printf("*" * 96)
 
     return not failed
--- a/site/it/site_checks/code_block.py
+++ b/site/it/site_checks/code_block.py
@@ -84,7 +84,9 @@ class CodeBlockBase:
     line_no: int
     skip_all: bool
 
-    def __init__(self, line_no: int, start_marker: str, indent: int, code_type: Optional[str] = None, skip_all: bool = False):
+    def __init__(
+        self, line_no: int, start_marker: str, indent: int, code_type: Optional[str] = None, skip_all: bool = False
+    ):
         self.line_no = line_no
         self.start_marker = start_marker
         self.indent = indent
@@ -94,17 +96,21 @@ class CodeBlockBase:
         self.skip_all = skip_all
 
     def code_append(self, line: str) -> None:
-        self.code.append(line[self.indent:].rstrip())
+        self.code.append(line[self.indent :].rstrip())
 
     def __str__(self) -> str:
-        return (f"CodeBlock(line_no={self.line_no}, code_type={self.code_type!r}, "
-                f"skip_all={self.skip_all}, indent={self.indent}, lines={len(self.code)})")
+        return (
+            f"CodeBlock(line_no={self.line_no}, code_type={self.code_type!r}, "
+            f"skip_all={self.skip_all}, indent={self.indent}, lines={len(self.code)})"
+        )
 
 
 class CodeBlock(CodeBlockBase):
     nested_code_blocks: List[CodeBlockBase]
 
-    def __init__(self, line_no: int, start_marker: str, indent: int, code_type: Optional[str] = None, skip_all: bool = False):
+    def __init__(
+        self, line_no: int, start_marker: str, indent: int, code_type: Optional[str] = None, skip_all: bool = False
+    ):
         super().__init__(line_no, start_marker, indent, code_type, skip_all)
         self.nested_code_blocks = []
 
@@ -128,7 +134,9 @@ def emit_code_block(f: TextIOBase, type_name: str, code: List[str], line_no: int
     :param code: Code block lines.
     :param line_no: Starting line number of the code block in the Markdown file.
     """
-    f.write(f'echo "::group::Dump of {"skipped/repeated " if skipped else ""}{type_name} code block at line {line_no}"\n')
+    f.write(
+        f'echo "::group::Dump of {"skipped/repeated " if skipped else ""}{type_name} code block at line {line_no}"\n'
+    )
     # Write the code block to be `echo`d so it is shown in the CI job log.
     f.write("cat <<CODE_SNIPPET_DUMP\n")
     for line in code:
--- a/site/it/site_checks/gen_test_script.py
+++ b/site/it/site_checks/gen_test_script.py
@@ -58,8 +58,14 @@ from pathlib import Path
 from textwrap import dedent
 from typing import List, Optional
 
-from .code_block import CodeBlock, CodeBlocksGlobal, \
-    emit_code_block, tweak_curl, tweak_docker_compose_up, tweak_spark_sql
+from .code_block import (
+    CodeBlock,
+    CodeBlocksGlobal,
+    emit_code_block,
+    tweak_curl,
+    tweak_docker_compose_up,
+    tweak_spark_sql,
+)
 
 open_pattern = re.compile("((\\s*)`{3,})([a-z0-9-_:]+)?\\s*(\\bskip_all\\b)?\\s*")
 close_pattern = re.compile("(\\s*`{3,})")
@@ -90,7 +96,7 @@ def generate_markdown_test_script(markdown_file: Path, output_file: Path, code_b
                         start_marker=open_match.group(1),
                         indent=len(open_match.group(2)),
                         code_type=open_match.group(3),
-                        skip_all=open_match.group(4) is not None
+                        skip_all=open_match.group(4) is not None,
                     )
             else:
                 close_match = re.match(close_pattern, line)
@@ -154,8 +160,13 @@ def generate_markdown_test_script(markdown_file: Path, output_file: Path, code_b
                 code = tweak_docker_compose_up(code_block.code)
                 code = tweak_curl(code)
                 code = tweak_spark_sql(code)
-                emit_code_block(f=f, type_name="shell", code=code, line_no=code_block.line_no,
-                                skipped=code_blocks_global.is_skipped(code_block))
+                emit_code_block(
+                    f=f,
+                    type_name="shell",
+                    code=code,
+                    line_no=code_block.line_no,
+                    skipped=code_blocks_global.is_skipped(code_block),
+                )
             elif code_block.code_type == "sqlshell":
                 sql_code: List[str] = ['cat > "${BUILD_TESTS_DIR}/.current.sql" <<SQL_SNIPPET_EOF']
                 for nested_code_block in code_block.nested_code_blocks:
--- a/site/it/test/test_docker.py
+++ b/site/it/test/test_docker.py
@@ -64,18 +64,6 @@ def test_cleanup_docker_removes_containers_and_prunes() -> None:
     ]
 
 
-def test_cleanup_docker_handles_missing_docker_binary() -> None:
-    tee = TeeRaiser()
-
-    with patch("site_checks.docker.subprocess.run", side_effect=FileNotFoundError):
-        cleanup_docker(tee)
-
-    assert tee.run_calls == [
-        (["docker", "network", "prune", "-f"], {}),
-        (["docker", "volume", "prune", "-f"], {}),
-    ]
-
-
 def test_docker_compose_info_logs_entries() -> None:
     tee = TeeRecorder()
     result = SimpleNamespace(stdout='[{"ConfigFiles":"compose-a.yml"},{"ConfigFiles":"compose-b.yml"}]')


snazy:
Hint for reviewers:
Look at the summary of the `test-site-guides` workflow.
The log of the `Guide testing` step of that workflow has a lot of details about the performed operations and container logs.
You may also want to "just run" the executable `./site/it/markdown-testing.py` from the checked out PR branch.

site/content/guides/telemetry/index.md
```diff
@@ -38,9 +38,9 @@ This example requires `jq` to be installed on your machine.
 
     ```shell
     ./gradlew \
-      :polaris-server:assemble \
-      :polaris-server:quarkusAppPartsBuild --rerun \
-      -Dquarkus.container-image.build=true
+       :polaris-server:assemble \
```
adutra, line 41: Nit: the new indentation is weirder than the old one (7 spaces vs 6).

site/content/guides/_index.md
```diff
@@ -50,3 +50,110 @@ Each guide has detailed instructions.
 
 - [Keycloak](keycloak): An example that uses Keycloak as an external identity provider (IDP) for
   authentication.
+
+## Authoring Guides
+
+Writing new or updating Guides, especially those that use Docker Compose, should follow these
```
adutra:
```suggestion
Writing new Guides or updating existing ones, especially those that use Docker Compose, should follow these
```

site/content/guides/_index.md
```diff
@@ -50,3 +50,110 @@ Each guide has detailed instructions.
...
+./markdown-testing.py content/guides/rustfs/index.md content/guides/telemetry/index.md
+```
+
+More information about guides-testing can be found in the `site/it/README.md` file in the
+[source repository](https://github.com/apache/polaris).
+
+### Constraints
+
+1. `docker compose` invocations must be on a single line in a `shell` code block.
+2. Spark SQL shell invocations must be the only statement `shell` code block in a guide.
```
adutra: That's not clear imo: is it impossible to have two code blocks with `spark-sql` in the same guide, or is it impossible to have a code block with `spark-sql` and some other statement?

site/it/site_checks/markdown_testing.py
```diff
@@ -0,0 +1,240 @@
...
+        return False
+
+    spark_tarball_url = SPARK_TGZ_URL
+    spark_download_requested = False
+
+    for opt, val in opts:
+        if opt in ("-h", "--help"):
+            usage()
+            return True
+        elif opt in ("-s", "--spark-spark_tarball"):
```
adutra:
```suggestion
        elif opt in ("-s", "--spark-tarball"):
```

.github/workflows/site-guides.yml
```diff
@@ -0,0 +1,78 @@
...
+# under the License.
+#
+
+name: Site Guides
+
+on:
+  push:
+    branches: [ "main" ]
+    paths:
+      - 'site/content/guides/**'
```
adutra, line 26:
```suggestion
      - 'site/content/guides/**'
      - 'site/it/**'
```

site/it/site_checks/spark.py
```diff
@@ -0,0 +1,63 @@
...
+        "--strip-components=1",
+    ]
+    with subprocess.Popen(curl_cmd, stdout=subprocess.PIPE) as curl_proc:
+        with subprocess.Popen(tar_cmd, stdin=curl_proc.stdout) as tar_proc:
+            assert curl_proc.stdout is not None
+            curl_proc.stdout.close()
+            tar_rc = tar_proc.wait()
+        curl_rc = curl_proc.wait()
+    if curl_rc != 0 or tar_rc != 0:
+        raise subprocess.CalledProcessError(curl_rc or tar_rc, curl_cmd)
```
adutra:
```suggestion
   if curl_rc != 0:
       raise subprocess.CalledProcessError(curl_rc, curl_cmd)
   if tar_rc != 0:
       raise subprocess.CalledProcessError(tar_rc, tar_cmd)
```

build.gradle.kts
```diff
@@ -115,6 +115,9 @@ tasks.named<RatTask>("rat").configure {
   excludes.add("node_modules/**")
 
   // Guides testing
+  excludes.add("site/content/guides/ceph/**/*.keyring")
+  excludes.add("site/it/.idea")
+  excludes.add("/venv")
```
adutra: This one is more about Python, see section below where we already have:

```
  excludes.add("**/.venv/**")
  excludes.add("**/polaris-venv/**")
```

Could you maybe keep using `/.venv` instead of `/venv` ?

build.gradle.kts
```diff
@@ -115,6 +115,9 @@ tasks.named<RatTask>("rat").configure {
   excludes.add("node_modules/**")
 
   // Guides testing
+  excludes.add("site/content/guides/ceph/**/*.keyring")
+  excludes.add("site/it/.idea")
```
adutra: We already have `excludes.add(".idea")` on line 145, maybe we can merge these two together?

site/content/guides/telemetry/index.md
snazy: Yea, I had to align all these to let the code skip the repeated container image builds via https://github.com/snazy/polaris/blob/guides-ci/site/content/guides/.it-setup/index.md?plain=1

site/content/guides/_index.md
snazy: Will clarify

build.gradle.kts
snazy: yea, moved the 2nd one

build.gradle.kts
snazy: Hm - don't recall why I added `/venv`, removed that one.

site/content/guides/telemetry/index.md
dimas-b: well, what if someone makes a change and this code block no longer matches the common one?

.github/workflows/site-guides.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          python-version: '3.14'
+      - name: Setup test environment
+        uses: ./.github/actions/setup-test-env
+
+      - name: Test site testing
+        working-directory: site/it
+        run: |
+          echo "::group::Setup"
+          python3 -m pip install --upgrade pip
```
MonkeyCanCode, line 55: Should we just call `make client-install-dependencies` instead?

site/it/site_checks/docker.py
```diff
@@ -0,0 +1,112 @@
...
+# under the License.
+#
+
+import json
+import subprocess
+import time
+
+from .tee import Tee
+
+DASH_LINE = "------------------------------------------------------------------------------------------------"
```
MonkeyCanCode: NIT: maybe use `DASH_LINE = "-" * 96` instead

site/it/site_checks/docker.py
```diff
@@ -0,0 +1,112 @@
...
+
+    :param tee: Log output.
+    :return: Elapsed time (in seconds) spent running this cleanup.
+    """
+    start = time.monotonic()
+    try:
+        tee.printf("::group::Cleanup Docker containers, networks, volumes")
+        tee.printf("Purging Docker containers...")
+        try:
+            result = subprocess.run(
```
MonkeyCanCode: nit:

it may be more read-able by using shlex such as following:
```
result = subprocess.run(
    shlex.split("docker ps -a -q"),
    capture_output=True,
    text=True,
    check=False,
)
```

site/it/site_checks/docker.py
```diff
@@ -0,0 +1,112 @@
...
+                ["docker", "ps", "-a", "-q"],
+                capture_output=True,
+                text=True,
+                check=False,
+            )
+            container_ids = [line for line in result.stdout.splitlines() if line.strip()]
+            if container_ids:
+                tee.run(["docker", "container", "rm", "-f", *container_ids])
+        except FileNotFoundError:
```
MonkeyCanCode: So the `FileNotFoundError` here is to check if `docker` binary is present? Instead of using `pass` to ignore silently, should we do warning instead?

Also, should we put all run in a try block instead of diff try blocks as if one is missing, the remaining one will fail as well.

site/it/site_checks/docker.py
```diff
@@ -0,0 +1,112 @@
...
+    start = time.monotonic()
+    try:
+        tee.printf("::group::Cleanup Docker containers, networks, volumes")
+        tee.printf("Purging Docker containers...")
+        try:
+            result = subprocess.run(
+                ["docker", "ps", "-a", "-q"],
+                capture_output=True,
+                text=True,
+                check=False,
```
MonkeyCanCode: NIT: default value for `check` is `False`:
```
subprocess.run(args, *, stdin=None, input=None, stdout=None, stderr=None, capture_output=False, shell=False, cwd=None, timeout=None, check=False, encoding=None, errors=None, text=None, env=None, universal_newlines=None, **other_popen_kwargs)
```

site/it/site_checks/docker.py
```diff
@@ -0,0 +1,112 @@
...
+        tee.printf("::endgroup::")
+        return time.monotonic() - start
+
+    try:
+        compose_entries = json.loads(result.stdout) if result.stdout.strip() else []
+    except json.JSONDecodeError:
+        compose_entries = []
+
+    for entry in compose_entries:
+        docker_compose_file = entry.get("ConfigFiles")
```
MonkeyCanCode, line 90: nit: currently we are all using one docker-compose file per getting-started example, will there ever be consideration to use a base docker-compose file later then (e.g. to avoid repetitive of common services such as bucket-setup and polaris service) instead of a standalone one which has everything? if yes, we will need to parse this as a list then use `-f` for each of them.

site/it/site_checks/markdown_testing.py
```diff
@@ -0,0 +1,240 @@
...
+from typing import List
+
+from .code_block import CodeBlocksGlobal
+from .docker import cleanup_docker, docker_compose_info
+from .gen_test_script import generate_markdown_test_script
+from .spark import SPARK_TGZ_URL, ensure_spark
+from .tee import Tee
+
+
+def _format_duration(seconds: float) -> str:
```
MonkeyCanCode, line 38: nits: maybe use `divmod` to splits seconds into minutes and seconds such as following:
```
def _format_duration(seconds: float) -> str:
    seconds = max(0.0, float(seconds))
    if seconds < 60:
        return f"{seconds:.2f}s"
    minutes, rem_seconds = divmod(int(round(seconds)), 60)
    return f"{minutes}m {rem_seconds:02d}s"
```

site/it/site_checks/markdown_testing.py
```diff
@@ -0,0 +1,240 @@
...
+            workspace_dir=workspace_dir,
+            build_tests_dir=build_tests_dir,
+            test_summary_file=test_summary_file,
+            env=env,
+            code_blocks_global=code_blocks_global,
+        )
+
+    print(">")
+
+    print("************************************************************************************************")
```
MonkeyCanCode, line 237: nit: `print('*' * 96)`

site/it/site_checks/spark.py
```diff
@@ -0,0 +1,65 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+import subprocess
+from pathlib import Path
+
+SPARK_TGZ_URL = "https://downloads.apache.org/spark/spark-3.5.8/spark-3.5.8-bin-hadoop3.tgz"
```
MonkeyCanCode: As suggested by [dongjoon-hyun](https://github.com/dongjoon-hyun) from https://github.com/apache/polaris/pull/2038, it is better to use `closer.lua` instead.


Also, as this is just for testing, should we consider just install pyspark via pip instead? We can add it as a test dependency in `pyproject.yaml`.

.github/workflows/site-guides.yml
snazy: Not sure, but worth the effort if the prod-code actually needs dependencies.

site/it/site_checks/docker.py
snazy: Yea - the FNF exceptions are nonsense, removed those.

site/it/site_checks/docker.py
snazy: Added a comment to `_index.md`

site/it/site_checks/spark.py
snazy: Makes sense, added this

.github/workflows/site-guides.yml
MonkeyCanCode: But the above target will setup vent and dependencies. Currently they are a bit repetitive.

MonkeyCanCode:
Sorry for the delay @snazy. I will review again tomorrow.

.github/workflows/site-guides.yml
snazy: This part here also runs the tests.
Not sure whether we should reuse Polaris client Python setup specific here for Site Guides ITs.

site/it/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+Executed code blocks that do not complete successfully, stop the execution of
+the generated test script and mark the test for that guide as failed. 
+
+Each generated temporary bash script, one per guide, is called with the
+current working directory set to the projects _workspace_ directory.
+
+## Running the tests
+
+Requirements: Python 3.14
```
MonkeyCanCode, line 45: As the code is not depends on other dependencies, any non-EOL python version should be supported? Maybe we should do `Python >= 3.10`?

site/it/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+# under the License.
+#
+
+[project]
+name = "site-checks"
+version = "0.0.0"
+description = "Apache Polaris - Internal Site Guide Testing"
+authors = [{ name = "Apache Software Foundation", email = "dev@polaris.apache.org" }]
+license = "Apache-2.0"
+requires-python = ">=3.14"
```
MonkeyCanCode, line 26: Same here. Maybe we should do `">=3.10,<4.0"` instead as the current main Ci is `>=3.10` but `<3.14` due to not all dependencies works on `3.14`. Seems to be odd for enforcing 3.14.

site/it/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+[project]
+name = "site-checks"
```
MonkeyCanCode, line 21: For naming consistency, maybe we should use `apache-polaris-site-checks`? We current are using `apache-polaris` and `apache-polaris-mcp`.

site/it/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+version = "0.0.0"
+description = "Apache Polaris - Internal Site Guide Testing"
+authors = [{ name = "Apache Software Foundation", email = "dev@polaris.apache.org" }]
+license = "Apache-2.0"
+requires-python = ">=3.14"
+dependencies = [
```
MonkeyCanCode, line 27: maybe also add `readme = "README.md"` here?

site/it/lint.sh
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+set -euo pipefail
+
```
MonkeyCanCode, line 22: So this script has `ruff` as dependency, for better US, should we do a conditional check here and provide user a hint on what to do next if we are not plan to setup venv up for the end-user such as following:
```
if ! python3 -m ruff --version >/dev/null 2>&1; then
  echo "Error: ruff is not installed."
  echo "Install it with: pip install ruff"
  exit 1
fi
```

site/it/README.md
```diff
@@ -0,0 +1,117 @@
...
+
```
MonkeyCanCode, line 46: Also, as the `pyproject.toml` is `uv` based, should we add uv as dependency here then setup env for end-user with `uv sync`, this will avoid the potential conditional check in `lint.sh` with the newly added comment.

site/it/README.md
```diff
@@ -0,0 +1,117 @@
...
+The executed generated bash scripts and the corresponding log files are uploaded as
+GitHub workflow artifacts.
+
+## Linting
+
+Run the local lint checks with:
+
+```bash
+./lint.sh
+```
```
MonkeyCanCode, line 117: Should we add a section for development? So if others need to make changes to these python code, they will know what to do before/after their changes?

site/it/site_checks/spark.py
```diff
@@ -0,0 +1,65 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+import subprocess
+from pathlib import Path
+
+SPARK_TGZ_URL = "https://www.apache.org/dyn/closer.lua/spark/spark-3.5.8/spark-3.5.8-bin-hadoop3.tgz?action=download"
```
MonkeyCanCode: Seems likes all of those guides are only doing basic spark stuff and not the spark spark plugin for generic table. If this is the case, worth this be easier just to add pyspark as a dependency in `pyproject.yaml` instead?

site/it/site_checks/markdown_testing.py
```diff
@@ -0,0 +1,242 @@
...
+
+    try:
+        md_rel_path = md_file.resolve().relative_to(site_dir.resolve())
+    except ValueError:
+        raise Exception(f"Error: {md_file} must be inside {site_dir}")
+
+    md_rel_name = md_rel_path.as_posix()
+    md_base_name = md_file.name
+    md_dir_name = str(md_rel_path.parent)
+    test_name = md_rel_name.replace("/", "_-_")
```
MonkeyCanCode, line 83: `md_rel_name` only got used once. maybe we should switch to `test_name = md_rel_path.as_posix().replace("/", "_-_")` instead?

site/it/site_checks/markdown_testing.py
```diff
@@ -0,0 +1,242 @@
...
+            with open(env["GITHUB_STEP_SUMMARY"], "a", encoding="utf-8") as summary_out:
+                summary_out.write(f"{summary}\n")
+        tee.printf(">")
+
+        docker_info_duration = docker_compose_info(tee, md_base_name, md_dir_name)
+        cleanup_duration = cleanup_docker(tee)
+
+        tee.printf(
+            dedent(f"""
+            Test steps durations for for {md_file} ({"❌ FAILED" if failed else "✅ passed"}):
```
MonkeyCanCode, line 122: duplicate `for` here.

site/it/site_checks/markdown_testing.py
```diff
@@ -0,0 +1,242 @@
...
+    md_files = sorted(p.absolute().relative_to(site_dir) for p in raw_files)
+
+    test_summary_file = build_tests_dir / "test-summary.txt"
+    if test_summary_file.exists():
+        test_summary_file.unlink()
+    test_summary_file.touch()
+
+    code_blocks_global = CodeBlocksGlobal()
+
+    os.chdir(site_dir)
```
MonkeyCanCode, line 222: This may not be necessary as we are pass-in `site_dir=site_dir` in the `run_test`?

MonkeyCanCode:
@snazy overall looks great and I tested the functionality earlier. The only concern if we will need to draft these getting started guides in a precise layout (down to spaces in this space). Left couple more nits/comments. 

@flyrain you may be interested to take a look as well.

site/it/README.md
snazy: I just have only tested it with 3.14. If it works with other Python versions, it's fine.

site/it/pyproject.toml
snazy: Yea, but this one hasn't anything to do with "main CI". It's just a runtime thing in a separate CI job.

site/it/pyproject.toml
snazy: Not sure we should go the "full release ready" route. This is just a tool that's run from source.

site/it/README.md
snazy: Should be caught by the CI workflow.

snazy:
@MonkeyCanCode thanks for the re-review. I'm not sure we need to polish esp. the pyproject.toml file up. It's just a tool that runs in CI, nothing of it will be released as a standalone tool. WDYT?

MonkeyCanCode:
> @MonkeyCanCode thanks for the re-review. I'm not sure we need to polish esp. the pyproject.toml file up. It's just a tool that runs in CI, nothing of it will be released as a standalone tool. WDYT?

That is a fair point. Approved.

PR #3441, opened 2026-01-14
cleaned up *CommitViewEvent

evindj:
This PR addresses issue #3418 after this [PR](https://github.com/apache/polaris/pull/3195/changes#top) recently did the same for tables. 
Given that the current implementation of commitTransaction works only for table (from my reading, please correct if I am wrong), I concluded only removing events would be enough especially since *ReplaceView  is already supported.

Tests:

Ran integration tests.

## Checklist
- [X] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [X] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [X] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [X] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

Issue #3418, opened 2026-01-12
Clean up inconsistency of table vs view events


snazy:
### Describe the bug

The PR #3195 changed events for tables, introducing an asymmetry to the same sets of events for view.
See [this comment](https://github.com/apache/polaris/pull/3195#issuecomment-3640868435) and [dev ML](https://lists.apache.org/thread/8kgcqrfhlrk0t907ogltk2dwxfwr9b2q).

PR #3554, opened 2026-01-26
removed references of BEFORE/AFTER_COMMIT_VIEW

evindj:
This PR addresses issue https://github.com/apache/polaris/issues/3418 after this [PR](https://github.com/apache/polaris/pull/3195/changes#top) recently did the same for tables.
Given that the current implementation of commitTransaction works only for table (from my reading, please correct if I am wrong), I concluded only removing events would be enough especially since *ReplaceView is already supported.

Tests:

Ran integration tests.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - The custom token-bucket based rate limiter has been replaced with Guava's rate limiter implementation.
 
 ### Breaking changes
+
 - The (Before/After)CommitViewEvent has been removed.
 - The (Before/After)CommitTableEvent has been removed.
 - The `PolarisMetricsReporter.reportMetric()` method signature has been extended to include a `receivedTimestamp` parameter of type `java.time.Instant`.
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1852,7 +1852,6 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
     }
 
     public void doCommit(ViewMetadata base, ViewMetadata metadata) {
-      
       // TODO: Maybe avoid writing metadata if there's definitely a transaction conflict
       LOGGER.debug(
           "doCommit for view {} with metadataBefore {}, metadataAfter {}",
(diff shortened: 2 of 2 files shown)


evindj:
@snazy please have a look. I closed https://github.com/apache/polaris/pull/3441 in favor of this after mixing up the rebase.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
```diff
@@ -145,8 +145,8 @@ public enum PolarisEventType {
   AFTER_RENAME_VIEW,
   BEFORE_REPLACE_VIEW,
   AFTER_REPLACE_VIEW,
-  BEFORE_COMMIT_VIEW,
-  AFTER_COMMIT_VIEW,
+  BEFORE_COMMIT_VIEW, // REMOVED FROM SOURCE CODE
```
adutra, line 148: I'm catching up here and saw this related conversation:

https://github.com/apache/polaris/pull/3195/changes#r2593440033

I'm not sure I agree with keeping these constants. Granted the ordinals will change if we remove them. But the feature is in beta status right now. It feels odd to see deprecated/unused constants in an API that is not even "official" yet.

@adnanhemani @dimas-b what do you think?

dimas-b: I do not see any particular issue with changing ordinals. AFAIK, Polaris does not use java serialization for this enum and Jackson deals with names.

evindj: If we are to remove the constants, will be ok doing in a follow up PR?cc @adutra I can take ownership of that issue as well.

snazy: I think doing it in a separate PR makes sense, so the two other similar enum values can be removed as well.

snazy: @evindj can you take care of the enum above in a separate PR?

evindj: sure @snazy

snazy (approved):

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
adutra: +1 for tackling this in a separate PR. Thanks @evindj !

PR #3584, opened 2026-01-27
relational-jdbc: add DB-agnostic idempotency store + model

huaxingao:
https://github.com/apache/polaris/pull/3205 followup:
- Introduce JDBC model for idempotency records via `ModelIdempotencyRecord` 
- Add database-agnostic implementation `RelationalJdbcIdempotencyStore` implementing `IdempotencyStore`.
- Standardize core domain type location by moving `IdempotencyRecord` to `polaris-core` under `org.apache.polaris.core.entity` and updating `IdempotencyStore `imports accordingly.
- Update tests by replacing `PostgresIdempotencyStoreIT` with `RelationalJdbcIdempotencyStorePostgresIT` 
## Checklist

### DIFF
--- a/persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/QueryGeneratorTest.java
+++ b/persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/QueryGeneratorTest.java
@@ -253,14 +253,21 @@ public class QueryGeneratorTest {
     setClause.put("error_subtype", null);
     setClause.put("http_status", 200);
 
+    // Use ordered maps so WHERE clause order is deterministic.
+    Map<String, Object> whereEquals = new LinkedHashMap<>();
+    whereEquals.put("realm_id", "r1");
+    whereEquals.put("idempotency_key", "k1");
+    Map<String, Object> whereLess = new LinkedHashMap<>();
+    whereLess.put("http_status", 500);
+
     QueryGenerator.PreparedQuery q =
         QueryGenerator.generateUpdateQuery(
             List.of("error_subtype", "http_status", "realm_id", "idempotency_key", "executor_id"),
             "idempotency_records",
             setClause,
-            Map.of("realm_id", "r1", "idempotency_key", "k1"),
+            whereEquals,
             Map.of(),
-            Map.of("http_status", 500),
+            whereLess,
             Set.of("executor_id"),
             Set.of());
 


huaxingao:
cc @singhpk234 @flyrain Could you please take a look when you have a moment?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+    if (record.getExecutorId() == null || !record.getExecutorId().equals(executorId)) {
+      return HeartbeatResult.LOST_OWNERSHIP;
+    }
+
+    String sql =
+        "UPDATE "
+            + QueryGenerator.fullyQualifiedTableName(ModelIdempotencyRecord.TABLE_NAME)
+            + " SET heartbeat_at = ?, updated_at = ?"
+            + " WHERE realm_id = ? AND idempotency_key = ? AND http_status IS NULL AND executor_id = ?";
```
singhpk234: is it possible to model these in QueryGenerator ?

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStorePostgresIT.java
```diff
@@ -37,14 +37,14 @@
 import org.testcontainers.junit.jupiter.Testcontainers;
 
 @Testcontainers
-public class PostgresIdempotencyStoreIT {
+public class RelationalJdbcIdempotencyStorePostgresIT {
```
singhpk234, line 42: is it possible to have PG profile ? we can swap it with cocraoch later ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,213 @@
...
+import javax.sql.DataSource;
+import org.apache.polaris.core.entity.IdempotencyRecord;
+import org.apache.polaris.core.persistence.IdempotencyStore;
+import org.apache.polaris.persistence.relational.jdbc.DatasourceOperations;
+import org.apache.polaris.persistence.relational.jdbc.QueryGenerator;
+import org.apache.polaris.persistence.relational.jdbc.RelationalJdbcConfiguration;
+import org.apache.polaris.persistence.relational.jdbc.models.ImmutableModelIdempotencyRecord;
+import org.apache.polaris.persistence.relational.jdbc.models.ModelIdempotencyRecord;
+
+/** {@link IdempotencyStore} backed by the relational-jdbc schema. */
```
singhpk234: is this required ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,213 @@
...
+public class RelationalJdbcIdempotencyStore implements IdempotencyStore {
+
+  private final DatasourceOperations datasourceOperations;
+
+  public RelationalJdbcIdempotencyStore(
```
singhpk234, line 44: may be later i would recommend having a dedicated named Quarkus datasource here, its fine for now to have it this way may be during the Quarkus producer we inject the DS we want.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,213 @@
...
+
+    String sql =
+        "UPDATE "
+            + QueryGenerator.fullyQualifiedTableName(ModelIdempotencyRecord.TABLE_NAME)
+            + " SET heartbeat_at = ?, updated_at = ?"
+            + " WHERE realm_id = ? AND idempotency_key = ? AND http_status IS NULL AND executor_id = ?";
+    QueryGenerator.PreparedQuery update =
+        new QueryGenerator.PreparedQuery(
+            sql,
+            List.of(Timestamp.from(now), Timestamp.from(now), realmId, idempotencyKey, executorId));
```
singhpk234: see insteadt of using Timestamp.now() we can use the injected Clock ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: removed this line

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: Sounds good.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: Yes. `QueryGenerator` didn’t previously model `IS NULL` predicates, `<` comparisons, or partial-column `UPDATE/DELETE` shapes, so the initial version used handwritten SQL. I’ve added minimal support for these in `QueryGenerator` (partial `UPDATE/DELETE` + `<` + `IS NULL/IS NOT NULL`) and refactored the code to use the generator.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: We’re not using `Timestamp.now()` here — the store takes Instant `now` as a parameter and only converts via `Timestamp.from(now)`

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStorePostgresIT.java
huaxingao: Added Postgres IT wiring via containerSpecHelper/PostgresRelationalJdbcLifeCycleManagement so we can swap/add CockroachDB later with minimal changes.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,233 @@
...
+              ModelIdempotencyRecord.TABLE_NAME,
+              values,
+              realmId);
+      datasourceOperations.executeUpdate(insert);
+      return new ReserveResult(ReserveResultType.OWNED, Optional.empty());
+    } catch (SQLException e) {
+      if (datasourceOperations.isConstraintViolation(e)) {
+        return new ReserveResult(ReserveResultType.DUPLICATE, load(realmId, idempotencyKey));
+      }
+      throw new RuntimeException("Failed to reserve idempotency key", e);
```
singhpk234: can we have a defined exception ? what error code we wanna throw in case for the request ?

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
```diff
@@ -79,6 +86,14 @@ public String getOperationType() {
     return operationType;
   }
 
+  /**
+   * Normalized identifier of the resource affected by the operation.
+   *
+   * <p>This should be derived from the request (for example, a canonicalized path like {@code
+   * "tables/ns.tbl"}), not from a generated internal entity id. This ensures the binding is
```
singhpk234: how do we protect against drop + create and renames ?

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStorePostgresIT.java
```diff
@@ -37,14 +39,20 @@
...
 
   @Container
   private static final PostgreSQLContainer<?> POSTGRES =
-      new PostgreSQLContainer<>("postgres:17.5-alpine");
+      new PostgreSQLContainer<>(
+              containerSpecHelper("postgres", PostgresRelationalJdbcLifeCycleManagement.class)
+                  .dockerImageName(null)
```
singhpk234: why null ? can we do lastest tag ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -50,6 +50,15 @@ public record PreparedBatchQuery(String sql, List<List<Object>> parametersList)
...
   record QueryFragment(String sql, List<Object> parameters) {}
 
+  /**
+   * Returns the fully-qualified table name used by relational-jdbc queries.
+   *
+   * @param tableName Target table name.
+   */
+  public static String fullyQualifiedTableName(@Nonnull String tableName) {
+    return getFullyQualifiedTableName(tableName);
+  }
```
singhpk234: I am confused, why create a new public function ? can we not have private function itself made public ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -83,6 +83,11 @@ DatabaseType getDatabaseType() {
     return databaseType;
   }
 
+  /** Returns the detected database type for this datasource. */
+  public DatabaseType databaseType() {
+    return databaseType;
+  }
```
singhpk234: let make the getDatabaseType public ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
```diff
@@ -0,0 +1,211 @@
...
+  ModelIdempotencyRecord CONVERTER =
+      ImmutableModelIdempotencyRecord.builder()
+          .realmId("")
+          .idempotencyKey("")
+          .operationType("")
+          .resourceId("")
+          .createdAt(Instant.EPOCH)
+          .updatedAt(Instant.EPOCH)
+          .expiresAt(Instant.EPOCH)
+          .build();
```
singhpk234: I didn't fully get the rationale of this
orthognally if its absolutely required lets make it static ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,233 @@
...
+          datasourceOperations.executeSelect(query, ModelIdempotencyRecord.CONVERTER);
+      if (results.isEmpty()) {
+        return Optional.empty();
+      }
+      if (results.size() > 1) {
+        throw new IllegalStateException(
+            "More than one idempotency record found for realm/key: "
+                + realmId
+                + "/"
+                + idempotencyKey);
```
singhpk234, line 128: what status code we expect to throw 500 ?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -209,6 +277,32 @@ public static PreparedQuery generateDeleteQuery(
         "DELETE FROM " + getFullyQualifiedTableName(tableName) + where.sql(), where.parameters());
   }
 
+  /**
+   * Builds a DELETE query that supports richer WHERE predicates (equality, greater-than, less-than,
+   * IS NULL, IS NOT NULL).
+   */
+  public static PreparedQuery generateDeleteQuery(
```
singhpk234, line 270: we should add UTs for these in QueryGenerator

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,233 @@
...
+import javax.sql.DataSource;
+import org.apache.polaris.core.entity.IdempotencyRecord;
+import org.apache.polaris.core.persistence.IdempotencyStore;
+import org.apache.polaris.persistence.relational.jdbc.DatasourceOperations;
+import org.apache.polaris.persistence.relational.jdbc.QueryGenerator;
+import org.apache.polaris.persistence.relational.jdbc.RelationalJdbcConfiguration;
+import org.apache.polaris.persistence.relational.jdbc.models.ImmutableModelIdempotencyRecord;
+import org.apache.polaris.persistence.relational.jdbc.models.ModelIdempotencyRecord;
+
+public class RelationalJdbcIdempotencyStore implements IdempotencyStore {
```
singhpk234, line 40: How is this obj created ? How is core gonna use it

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
```diff
@@ -0,0 +1,211 @@
...
+  String ERROR_SUBTYPE = "error_subtype";
+  String RESPONSE_SUMMARY = "response_summary";
+  String RESPONSE_HEADERS = "response_headers";
+  String FINALIZED_AT = "finalized_at";
+
+  String CREATED_AT = "created_at";
+  String UPDATED_AT = "updated_at";
+  String HEARTBEAT_AT = "heartbeat_at";
+  String EXECUTOR_ID = "executor_id";
+  String EXPIRES_AT = "expires_at";
```
singhpk234, line 74: we should add code comments for each of the fields like we do in the rest of the model

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -0,0 +1,233 @@
...
+    // Raced with finalize/ownership loss; re-check to return a meaningful result.
+    Optional<IdempotencyRecord> after = load(realmId, idempotencyKey);
+    if (after.isEmpty()) {
+      return HeartbeatResult.NOT_FOUND;
+    }
+    if (after.get().getHttpStatus() != null) {
+      return HeartbeatResult.FINALIZED;
+    }
+    return HeartbeatResult.LOST_OWNERSHIP;
+  }
```
singhpk234, line 191: would it not need to be in a transaction ?

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStorePostgresIT.java
```diff
@@ -37,14 +39,20 @@
...
+                  .asCompatibleSubstituteFor("postgres"))
+          .withDatabaseName("polaris_db")
```
singhpk234: we set the foloowing in setup too, why we need here again ?

singhpk234:
These were final set of comments :), can't think of anymore, thanks @huaxingao for working on it and driving the whole idempotency effort !

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: Added `IdempotencyPersistenceException`

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: This path is a safety check: (realmId, idempotencyKey) should be unique. Expected behavior is HTTP 500.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: The actual heartbeat change is done via a single atomic UPDATE ... WHERE realm/key/executor AND http_status IS NULL, so it doesn’t require an explicit transaction for correctness.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
huaxingao: Added.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
huaxingao: I’ve refactored this to avoid the dummy instance by introducing a static fromRow(ResultSet) and a stateless CONVERTER singleton that delegates to it.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: Right, we don't need this. removed.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: Added. Thanks

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStorePostgresIT.java
huaxingao: null is intentional here: containerSpecHelper(...).dockerImageName(null) means “use the pinned default from Dockerfile-postgres-version and allow overrides via system props/env.

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStorePostgresIT.java
huaxingao: Removed

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
huaxingao: `RelationalJdbcIdempotencyStore` is just the relational-jdbc backend for the IdempotencyStore interface. It’s constructed with a DataSource + RelationalJdbcConfiguration and then the request/idempotency layer calls reserve/load/updateHeartbeat/finalize/purgeExpired through the IdempotencyStore SPI.

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
flyrain: Do we define `canonicalized path` somewhere?  Should we avoid duplication for a canonicalized path? A path like this `tables/ns.tbl` is very likely causing conflicts, as different catalogs could have the same namespace and table names.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
```diff
@@ -0,0 +1,235 @@
...
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.apache.polaris.persistence.relational.jdbc.DatabaseType;
+
+/**
+ * JDBC model for {@link IdempotencyRecord} mirroring the {@code idempotency_records} table.
+ *
+ * <p>This follows the same pattern as {@link ModelEvent}, separating the storage representation
+ * from the core domain model while still providing {@link Converter} helpers.
+ */
+@PolarisImmutable
```
flyrain: Do we need it to be PolarisImmutable? We didn't do that for `ModelEntity`, and a few others.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
```diff
@@ -83,6 +83,11 @@ DatabaseType getDatabaseType() {
     return databaseType;
   }
 
+  /** Returns the detected database type for this datasource. */
+  public DatabaseType databaseType() {
```
flyrain: Do we need this new method given that we've got a `getDatabaseType()` already(line 82)? We could change the scope to `public` if needed.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
```diff
@@ -0,0 +1,235 @@
...
+public interface ModelIdempotencyRecord extends Converter<IdempotencyRecord> {
+
+  String TABLE_NAME = "idempotency_records";
+
+  // Logical tenant / realm identifier.
+  String REALM_ID = "realm_id";
+  // Client-provided idempotency key.
```
flyrain, line 47: Nit: We got so many nice comments here while there is no comment on the class `IdempotencyRecord`.  I'd love to have these comments on the class `IdempotencyRecord`.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -192,6 +192,65 @@ public static PreparedQuery generateUpdateQuery(
     return new PreparedQuery(sql, bindingParams);
   }
 
+  /**
+   * Builds an UPDATE query that updates only the specified columns and supports richer WHERE
+   * predicates (equality, greater-than, less-than, IS NULL, IS NOT NULL).
+   *
+   * <p>Callers should prefer passing an ordered map (e.g. {@link java.util.LinkedHashMap}) for the
+   * set clause so generated SQL and parameter order are stable.
```
flyrain: nit: feel a bit easier to understand to use `consistent`, or `matched`. Just my personal preference. Feel free to ignore it.
```suggestion
   * <p>Callers should prefer passing an ordered map (e.g. {@link java.util.LinkedHashMap}) for the
   * set clause so generated SQL and parameter order are consistent.
```

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -209,6 +268,32 @@ public static PreparedQuery generateDeleteQuery(
...
+      @Nonnull Map<String, Object> whereGreater,
+      @Nonnull Map<String, Object> whereLess,
+      @Nonnull Set<String> whereIsNull,
+      @Nonnull Set<String> whereIsNotNull) {
+    Set<String> columns = new HashSet<>(tableColumns);
+    validateColumns(columns, whereEquals.keySet());
+    validateColumns(columns, whereGreater.keySet());
+    validateColumns(columns, whereLess.keySet());
+    validateColumns(columns, whereIsNull);
+    validateColumns(columns, whereIsNotNull);
```
flyrain: Do we need to validate them here as the method `generateWhereClauseExtended()` will validates them anyways?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -192,6 +192,65 @@ public static PreparedQuery generateUpdateQuery(
...
+      throw new IllegalArgumentException("Empty setClause");
+    }
+
+    Set<String> columns = new HashSet<>(tableColumns);
+    validateColumns(columns, setClause.keySet());
+    validateColumns(columns, whereEquals.keySet());
+    validateColumns(columns, whereGreater.keySet());
+    validateColumns(columns, whereLess.keySet());
+    validateColumns(columns, whereIsNull);
+    validateColumns(columns, whereIsNotNull);
```
flyrain: Same here

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -231,22 +316,55 @@ static QueryFragment generateWhereClause(
...
+  @VisibleForTesting
+  static QueryFragment generateWhereClauseExtended(
+      @Nonnull Set<String> tableColumns,
+      @Nonnull Map<String, Object> whereEquals,
+      @Nonnull Map<String, Object> whereGreater,
+      @Nonnull Map<String, Object> whereLess,
+      @Nonnull Set<String> whereIsNull,
+      @Nonnull Set<String> whereIsNotNull) {
+    // Preserve the original behavior of rejecting unknown columns. This is used by SELECT query
+    // generation too, not only by callers of the extended UPDATE/DELETE helpers.
```
flyrain: Nit: I guess this comment is more like a dev log. Do we need them as a comment?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
```diff
@@ -231,22 +316,55 @@ static QueryFragment generateWhereClause(
...
       @Nonnull Map<String, Object> whereGreater) {
+    return generateWhereClauseExtended(
+        tableColumns, whereEquals, whereGreater, Map.of(), Set.of(), Set.of());
+  }
+
+  private static void validateColumns(
+      @Nonnull Set<String> tableColumns, @Nonnull Set<String> columns) {
+    for (String column : columns) {
+      if (!tableColumns.contains(column) && !column.equals("realm_id")) {
+        throw new IllegalArgumentException("Invalid query column: " + column);
```
flyrain, line 316: `realm_id` is treated as a special implicit column for some models but explicitly included in `SELECT_COLUMNS` for `ModelIdempotencyRecord`. It might be cleaner to do something similar to other model, which keep `realm_id` out by only using `ALL_COLUMNS` instead of having `SELECT_COLUMNS`. WDYT?

singhpk234 (commented):
+1 LGTM overall

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
huaxingao: I removed `@PolarisImmutable` as suggested. That required a small follow-up refactor in `RelationalJdbcIdempotencyStore.reserve()`: we no longer build an `ImmutableModelIdempotencyRecord` just to get the insert bindings; instead we construct the insert map/values directly (same columns/values, still uses `QueryGenerator.generateInsertQuery`). This keeps behavior the same while avoiding the Immutables-generated type.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java
huaxingao: Removed. Thanks!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
huaxingao: Added more comments. Thanks!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: Fixed. Thanks for the suggestion!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: You are right. Removed.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: Removed

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: Removed.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/QueryGenerator.java
huaxingao: Makes sense. Fixed.

huaxingao:
Thanks a lot @flyrain @singhpk234

PR #3586, opened 2026-01-28
Change quick start example from MinIO to RustFS backend

MonkeyCanCode:
Per discussion in https://lists.apache.org/thread/8o31ly7cd8ov70opjbtg630qlhrfl5yh, we should separate getting started example and integration testings in the codebase (there are known issues we discovered as well as reported by community when using RustFS, however, these won't impact our getting start examples as they are not using testcontainers). With this change, we will be able to remove the disclaimer banner on the first couple pages users will be seeing (IMO, better UX and make users more confident about the project as we are not using vulnerable images).

Two small NITs for this PR:
1. Change auth from minio specific to more generic ones
2. use dash instead of underscore for service name (be consistent with the other one...so changed from `setup_bucket` to `bucket-setup`)

One thing kinds bugged me a bit but not changed in this PR is to change to vended credentials for spark client (somehow the current one is not using that, if not concern with this, I would like to change that too).

## Checklist

### DIFF
--- a/getting-started/quickstart/docker-compose.yml
+++ b/getting-started/quickstart/docker-compose.yml
@@ -19,25 +19,26 @@
 
 services:
 
-  minio:
-    image: quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
+  rustfs:
+    image: rustfs/rustfs:1.0.0-alpha.81
     ports:
       # API port
       - "9000:9000"
       # UI port
       - "9001:9001"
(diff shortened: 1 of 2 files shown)


dimas-b (approved):
LGTM 👍 Thanks, @MonkeyCanCode !

MonkeyCanCode:
Thanks for the review @jbonofre and @dimas-b .

PR #3587, opened 2026-01-28
Simplify getting started example for Ceph

MonkeyCanCode:
So currently users can just do `docker compose up -d` on ceph example due to the non-deterministic bucket creation service (as it is simply sleep for 5 seconds without proper service check). Also, the current README.md is a bit length and ask users to run various command to start services in-order (which is a bit un-necessary as the right ordering are already been defined via `depends_on`.

This PR simplify the doc a bit to allow user to run a single command to start the services. Also, a small nit fixing for using dash instead of underscore for service name (`setup_bucket` to `bucket-setup`). Beside this, I also added a gitignore to exclude files created by this docker compose.

Here is error users can get if run `docker compose up -d` (which is failing silently):
```
+ echo '>>> Waiting for RGW to become ready...'
+ sleep 5
>>> Waiting for RGW to become ready...
+ echo '>>> Create bucket if not exist...'
+ aws s3 mb s3://polaris-storage
>>> Create bucket if not exist...
make_bucket failed: s3://polaris-storage Could not connect to the endpoint URL: "http://rgw1:7480/polaris-storage"
+ true
+ tail -f /dev/null
```

And here is the output with current PR (matching output layout to the other docker-compose files for this specific service and ensure service):
```
Creating Ceph bucket...
make_bucket: polaris-storage
2026-01-28 01:50:49 polaris-storage
Bucket setup complete.
```

## Checklist

### DIFF
--- a/getting-started/ceph/README.md
+++ b/getting-started/ceph/README.md
@@ -52,7 +52,27 @@ cp .env.example .env
 docker compose up -d
 ```
 
-## 3. Connecting From Spark
+#### Check status
+```shell
+docker exec ceph-mon1-1 ceph -s
+```
+You should see something like:
+```yaml
+cluster:
+  id:     b2f59c4b-5f14-4f8c-a9b7-3b7998c76a0e
+  health: HEALTH_WARN
+          mon is allowing insecure global_id reclaim
+          1 monitors have not enabled msgr2
+          6 pool(s) have no replicas configured
+
+services:
+  mon: 1 daemons, quorum mon1 (age 49m)
+  mgr: mgr(active, since 94m)
+  osd: 1 osds: 1 up (since 36m), 1 in (since 93m)
+  rgw: 1 daemon active (1 hosts, 1 zones)
+```
+
+### 3. Connecting From Spark
 
 ```shell
 bin/spark-sql \
@@ -77,7 +97,7 @@ Note: `s3cr3t` is defined as the password for the `root` user in the `docker-com
 Note: The `client.region` configuration is required for the AWS S3 client to work, but it is not used in this example
 since Ceph does not require a specific region.
 
-## 4. Running Queries
+### 4. Running Queries
 
 Run inside the Spark SQL shell:
 


getting-started/ceph/docker-compose.yml
```diff
@@ -126,11 +126,18 @@ services:
...
       - osd1
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:7480"]
+      interval: 2s
+      timeout: 10s
+      retries: 10
+      start_period: 10s
 
-  setup_bucket:
+  bucket-setup:
```
snazy, line 136: Not sure this rename is necessary. The other guides use the same name.

getting-started/ceph/README.md
```diff
@@ -47,56 +47,12 @@ Note: this example pulls the `apache/polaris:latest` image, but assumes the imag
...
-docker compose up -d osd1
-```
-
-### 4. Start RGW
-```shell
-docker compose up -d rgw1
-```
-#### Check status
-```shell
-docker exec --interactive --tty ceph-mon1-1 ceph -s
```
snazy, line 66: I think this should stay, but the options `--interactive --tty ` removed.

getting-started/ceph/docker-compose.yml
```diff
@@ -126,11 +126,18 @@ services:
...
     image: amazon/aws-cli:2.33.8
     depends_on:
-      - rgw1
+      rgw1:
+        condition: service_healthy
```
snazy, line 140: That's a very good one!

getting-started/ceph/docker-compose.yml
MonkeyCanCode: So we have "polaris-setup" service down below while this one is "setup_bucket". Just NIT, for consistent, maybe we should do for all. What do you think?

getting-started/ceph/README.md
MonkeyCanCode: The content doesn't appear to be very useful but I will update it to be kept as a validation step before running the Spark example.

getting-started/ceph/docker-compose.yml
snazy: Let's keep for now - have a few PRs depending on it ;)

MonkeyCanCode: Sounds good. Will revert it later today for this change.

MonkeyCanCode:
@snazy thanks for the review and all comments had being addressed. Please take another look.

snazy (approved):

PR #3594, opened 2026-01-28
Minio guide: fix service dependencies

snazy:
Use the "long" options for compose service dependencies, and adds the explicit option `restart: no` option for setup tasks, which is necessary to let docker-compose correctly interpret the termination of such tasks.

### DIFF
--- a/getting-started/minio/docker-compose.yml
+++ b/getting-started/minio/docker-compose.yml
@@ -49,6 +49,8 @@ services:
     depends_on:
       minio:
         condition: service_healthy
+      setup_bucket:
+        condition: service_completed_successfully
     environment:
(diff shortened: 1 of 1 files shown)


dimas-b (approved):

getting-started/minio/docker-compose.yml
```diff
@@ -67,6 +69,7 @@ services:
 
   setup_bucket:
     image: quay.io/minio/mc:RELEASE.2025-08-13T08-35-41Z
+    restart: "no"
```
MonkeyCanCode: Any reason on why we need to set restart to no? I think that is default value is no.

PR #3597, opened 2026-01-28
Telemetry guide: fix setup-polaris service

snazy:
Fixes the polaris-setup service to use the "working" way/syntax, in alignment to other guides.

### DIFF
--- a/getting-started/telemetry/docker-compose.yml
+++ b/getting-started/telemetry/docker-compose.yml
@@ -69,7 +69,6 @@ services:
     command:
       - "-c"
       - >-
-        chmod +x /polaris/create-catalog.sh;
         /polaris/create-catalog.sh;
 
   prometheus:


getting-started/telemetry/docker-compose.yml
```diff
@@ -65,7 +65,12 @@ services:
       - CLIENT_SECRET=s3cr3t
     volumes:
       - ../assets/polaris/:/polaris
-    entrypoint: '/bin/sh -c "chmod +x /polaris/create-catalog.sh && /polaris/create-catalog.sh"'
+    entrypoint: "/bin/sh"
+    command:
+      - "-c"
+      - >-
+        chmod +x /polaris/create-catalog.sh;
```
MonkeyCanCode: will it be more cleaner if we just do `exec /polaris/create-catalog.sh;` as change file permission then run without implicitly shell is a bit unnecessary.

dimas-b: Actually, could we just fix UNIX permissions on `create-catalog.sh` in git?

MonkeyCanCode: Well, that is very good point. Upon checking, we can just remove chmod line completely as the file is 755 already and here is the mounted file permission:
```
/ # ls -l /polaris/create-catalog.sh
-rwxr-xr-x    1 root     root          2559 Dec 20 20:51 /polaris/create-catalog.sh
```

So we should just move chmod line.

snazy: good point, removed

PR #3608, opened 2026-01-28
Add PolarisEventType int codes and remove unused before/after commit view/table values.

evindj:
This PR is a continuation of https://github.com/apache/polaris/issues/3418 where we agreed we should remove associate enums values from Enum definition. 
The code also adds a constructor to help not having to rely on ordinals() to to have enum codes.

I incremented enum code assignments by 100 based on categories. I am happy to take feedback here.

Does not change logic and removes only enums that are no longer used so behavior does not change.

./gradlew clean assemble -x test

Is successful.

## Checklist

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTypeTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventTypeTest.java
@@ -27,9 +27,9 @@ class PolarisEventTypeTest {
 
   @Test
   void testAllCodesAreUnique() {
-    assertThat(Arrays.stream(PolarisEventType.values()).map(PolarisEventType::code).distinct().count())
+    assertThat(
+            Arrays.stream(PolarisEventType.values()).map(PolarisEventType::code).distinct().count())
         .as("All event type codes should be unique")
         .isEqualTo(PolarisEventType.values().length);
   }
-
 }


evindj:
@adutra @snazy @dimas-b  tagging you all for review. Since this is kind of breaking change since the ordinals will change, someone mentioned on a different thread that we should send a notice does anyone knows the process around that?

adnanhemani:
Not sure we need something this involved. I'd prefer instead to change the current name to `NO_LONGER_USED_xxx` with a comment that states what that enum value used to be prior.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
```diff
@@ -21,183 +21,187 @@
...
-  BEFORE_ATTEMPT_TASK,
-  AFTER_ATTEMPT_TASK,
+  BEFORE_ATTEMPT_TASK(1400),
+  AFTER_ATTEMPT_TASK(1401),
 
   // Rate Limiting Events
-  BEFORE_LIMIT_REQUEST_RATE,
+  BEFORE_LIMIT_REQUEST_RATE(1500);
+
+  private final int code;
```
snazy, line 198: Would you mind adding a test to ensure that no code is duplicated?
Something like this:
```java
@Test
void noDuplicateCodes() {
  assertThat(
    Arrays.stream(PolarisEventType.values())
      .map(PolarisEventType::code)
      .collect(Collectors.toSet()))
    .hasSize(PolarisEventType.values().length);
}
```

evindj: For sure, will add a test case shortly!

snazy:
Ping @evindj . @dimas-b I think we can also just go ahead with this PR?

evindj:
I don't see any risk associate with doing it in a single PR. I can see the angle of more clarity but quite frankly, splitting does not provide any addition safety guarantee beyond separating intent.

dimas-b:
I'm ok with merging this PR, but it would be nice to update the title to mention adding integer enum codes... just for the sake of clarity. The title shows up in the "brief" git log after merging.

dimas-b:
Thanks for your contribution, @evindj !

PR #3611, opened 2026-01-29
Guides: compose dependencies / long-option

snazy:
Docker compose's "short" `depends_on` syntax (just mentioning the dependent service) does not assert on the expected condition (`service_healthy` or `service_completed_successfully`). This can lead to services being started too early, although their dependencies are not in the expected state.

This change updates the affected parts to use the "long" syntax.

### DIFF
--- a/getting-started/telemetry/docker-compose.yml
+++ b/getting-started/telemetry/docker-compose.yml
@@ -78,6 +78,8 @@ services:
     depends_on:
       polaris:
         condition: service_healthy
+      polaris-setup:
+        condition: service_completed_successfully
     volumes:
       - ../assets/prometheus/:/etc/prometheus/
     command:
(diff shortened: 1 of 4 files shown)


MonkeyCanCode (approved):

getting-started/telemetry/docker-compose.yml
```diff
@@ -78,6 +78,8 @@ services:
     depends_on:
       polaris:
         condition: service_healthy
+      polaris-setup:
+        condition: service_completed_successfully
```
MonkeyCanCode, line 82: As we are changing to tail forever on the other PR, we will need to change this to service_healthy instead.

snazy: Good point!
I'll adjust the other PR.

snazy: Actually, it's [not strictly needed in the other PR for the telemetry guide](https://github.com/apache/polaris/pull/3612#discussion_r2749270997), but I've added it for consistency across the guides.

PR #3612, opened 2026-01-29
Guides: fix setup scripts to yield correct exit code

snazy:
The statements in the shell scripts for the setup services are often concatenated using `;`, which means that a previous' command exit code is _not_ propagated and the service, although it failed, is determined to be successful.

This change updates those scripts to use `&&` for the statement concatenation.

"Final" setup services (aka "polaris-setup") now have a final `sleep 120`. This is due to the behavior of `docker compose up --detach --wait`, which considers _any_ service (without dependants) that exits with exit code 0 as a failure, leading to that docker-compose command yielding an error code. That would break the guides testing code (#3553). That `sleep 120` in "polaris-setup" services does **not** cause a delay of the compose starting up - it is purely a "hack around" Docker Compose not having a notion of "setup services".

To avoid merge conflicts, this change also:
* updates affected `curl` invocations (as #3610)
* removes superfluous `restart: "no"`

### DIFF
--- a/getting-started/minio/docker-compose.yml
+++ b/getting-started/minio/docker-compose.yml
@@ -115,6 +115,7 @@ services:
           http://polaris:8181/api/management/v1/catalogs/quickstart_catalog/catalog-roles/catalog_admin/grants \
           -d '{"type":"catalog", "privilege":"CATALOG_MANAGE_CONTENT"}' &&
         echo Done. &&
+        touch /tmp/polaris-setup-done &&
         tail -f /dev/null
     healthcheck:
       interval: 1s
--- a/getting-started/ozone/docker-compose.yml
+++ b/getting-started/ozone/docker-compose.yml
@@ -166,6 +166,7 @@ services:
           http://polaris:8181/api/management/v1/catalogs/quickstart_catalog/catalog-roles/catalog_admin/grants \
           -d '{"type":"catalog", "privilege":"CATALOG_MANAGE_CONTENT"}' &&
         echo Done. &&
+        touch /tmp/polaris-setup-done &&
         tail -f /dev/null
     healthcheck:
       interval: 1s
--- a/getting-started/telemetry/docker-compose.yml
+++ b/getting-started/telemetry/docker-compose.yml
@@ -69,7 +69,14 @@ services:
     command:
       - "-c"
       - >-
-        /polaris/create-catalog.sh;
+        /polaris/create-catalog.sh &&
+        touch /tmp/polaris-setup-done &&
+        tail -f /dev/null
+    healthcheck:
+      interval: 1s
+      timeout: 10s
+      retries: 240
+      test: ["CMD", "test", "-f", "/tmp/polaris-setup-done"]
 
   prometheus:
     image: docker.io/prom/prometheus:v3.9.1
@@ -79,7 +86,7 @@ services:
       polaris:
         condition: service_healthy
       polaris-setup:
-        condition: service_completed_successfully
+        condition: service_healthy
     volumes:
       - ../assets/prometheus/:/etc/prometheus/
     command:


getting-started/keycloak/docker-compose.yml
```diff
@@ -71,12 +71,12 @@ services:
     command:
       - "-c"
       - >-
-        apk add --no-cache jq && 
-        chmod +x /polaris/create-catalog.sh && 
+        apk add --no-cache jq &&
         token=$$(curl http://keycloak:8080/realms/iceberg/protocol/openid-connect/token --user client1:s3cr3t -d 'grant_type=client_credentials' | jq -r .access_token) &&
```
MonkeyCanCode, line 75: should we do implicitly `--fail-with-body` here too for curl?

getting-started/ceph/docker-compose.yml
```diff
@@ -148,10 +148,10 @@ services:
...
       - "-c"
       - >-
-        echo Creating Ceph bucket...;
-        aws s3 mb s3://${S3_POLARIS_BUCKET};
-        aws s3 ls;
-        echo Bucket setup complete.;
+        echo Creating Ceph bucket... &&
+        aws s3 mb s3://${S3_POLARIS_BUCKET} &&
+        aws s3 ls &&
+        echo Bucket setup complete.
```
MonkeyCanCode, line 157: This will return 0 as well no?

getting-started/keycloak/docker-compose.yml
```diff
@@ -71,12 +71,12 @@ services:
...
       - >-
-        apk add --no-cache jq && 
-        chmod +x /polaris/create-catalog.sh && 
+        apk add --no-cache jq &&
         token=$$(curl http://keycloak:8080/realms/iceberg/protocol/openid-connect/token --user client1:s3cr3t -d 'grant_type=client_credentials' | jq -r .access_token) && 
         /polaris/create-catalog.sh realm-internal && 
         /polaris/create-catalog.sh realm-external $$token && 
-        /polaris/create-catalog.sh realm-mixed $$token
+        /polaris/create-catalog.sh realm-mixed $$token &&
+        sleep 120
```
MonkeyCanCode: So yeah, while I was doing a local prototyping, I ran into this issue as well where `--exit-code-from` is not happy with this. What I ended up doing is to use `tail /dev/null` to keep the service up then use health check on service ready instead of completed. Sleep 120 here will work but if for whatever reason these bash scripts took more than 120s (assuming they get very complex later on), this will bite us.

getting-started/keycloak/docker-compose.yml
snazy: My assumption is that the token would just be "invalid".
It's also that the pipe "eliminates" the exit code from curl, but `jq` would fail then.

getting-started/keycloak/docker-compose.yml
snazy: Man, docker-compose isn't a good friend at all :(

getting-started/ceph/docker-compose.yml
snazy: WDYM?

getting-started/keycloak/docker-compose.yml
snazy: Neither `tail /dev/null` nor `sleep 120` is particularly great.
I wanted to stay away from the added complexity of health-checks for the setup-scripts though.
Let me think a bit about this.

snazy: Probably best to let the polaris-setup services "tail forever" and have health-checks. Added that

getting-started/ceph/docker-compose.yml
MonkeyCanCode: So we have at most 2 setup services for these docker compose and you are adding `sleep 120` to pass the CI based on docker-compose (as exit of status code of zero will be treat as failed). Thus, this will also return status code of zero.

getting-started/keycloak/docker-compose.yml
MonkeyCanCode: Yes. That is what I ended up doing in my local prototyping with tail forever as status code of zero is not okay for compose service.

getting-started/keycloak/docker-compose.yml
MonkeyCanCode: That is fair enough. Yeah. Just thought to be consistent. But not a blocker.

getting-started/ceph/docker-compose.yml
snazy: Ah, I think I know what you're referring to.
Those setup-bucket services behave in an "interesting" way, which is different than how the "final" polaris-setup services behave (in terms of exit behavior).
An "intermediate" setup-service like `setup-bucket`, on which for example `polaris` depends using the `service_completed_successfully` condition, is fine to exit and does not "break" a `docker compose up --detach --wait`.
A "final" setup-service like some `polaris-setup`, a service that no other service depends on, lets `docker compose up --detach --wait` fail when it exits.

The "crux" with those "final" setup-services is that the behavior isn't immediately obvious and depends on the timing of the relevant services and when exactly `docker compose up --detach --wait` "thinks" the whole compose thing is ready.

But this ^^ is all from experiments. It's tricky to find good documentation on this behavior. However, the behavior of `depends_on` (... a setup service) with `service_completed_successfully` condition makes sense.

snazy:
@MonkeyCanCode mind taking another look?

PR #3616, opened 2026-01-29
feat(metrics): Add MetricsPersistence SPI for backend-agnostic metrics storage (#3337)

obelix74:
This commit introduces a Service Provider Interface (SPI) for persisting Iceberg metrics reports, addressing the extensibility concerns raised in the design review.

Key components:
- MetricsPersistence: Main SPI interface with write and query operations
- NoOpMetricsPersistence: Default do-nothing implementation for backends that don't support metrics persistence
- ScanMetricsRecord: Immutable interface for scan metrics data
- CommitMetricsRecord: Immutable interface for commit metrics data
- MetricsQueryCriteria: Query parameters with filtering and pagination
- MetricsContext: Context for conversion (realm, catalog, principal info)
- MetricsPersistenceFactory: Factory for realm-scoped instances
- MetricsRecordConverter: Converts Iceberg reports to SPI records

Relates to: #3337

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/metrics/iceberg/MetricsRecordConverter.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/metrics/iceberg/MetricsRecordConverter.java
@@ -20,7 +20,6 @@ package org.apache.polaris.core.metrics.iceberg;
 
 import java.time.Instant;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
@@ -45,7 +44,6 @@ import org.apache.polaris.core.persistence.metrics.ScanMetricsRecord;
  * ScanMetricsRecord record = MetricsRecordConverter.forScanReport(scanReport)
  *     .catalogId(catalog.getId())
  *     .tableId(tableEntity.getId())
- *     .namespace(namespace)
  *     .build();
  * }</pre>
  */
@@ -80,7 +78,6 @@ public final class MetricsRecordConverter {
     private final ScanReport scanReport;
     private long catalogId;
     private long tableId;
-    private List<String> namespace = Collections.emptyList();
     private Instant timestamp;
 
     private ScanReportBuilder(ScanReport scanReport) {
@@ -106,17 +103,6 @@ public final class MetricsRecordConverter {
     }
 
     /**
-     * Sets the namespace as a list of levels.
-     *
-     * @param namespace the namespace levels
-     * @return this builder
-     */
-    public ScanReportBuilder namespace(List<String> namespace) {
-      this.namespace = namespace != null ? namespace : Collections.emptyList();
-      return this;
-    }
-
-    /**
      * Sets the timestamp for the metrics record.
      *
      * <p>This should be the time the metrics report was received by the server, which may differ
@@ -138,7 +124,6 @@ public final class MetricsRecordConverter {
       return ScanMetricsRecord.builder()
           .reportId(UUID.randomUUID().toString())
           .catalogId(catalogId)
-          .namespace(namespace)
           .tableId(tableId)
           .timestamp(timestamp != null ? timestamp : Instant.now())
           .snapshotId(Optional.of(scanReport.snapshotId()))
@@ -181,7 +166,6 @@ public final class MetricsRecordConverter {
     private final CommitReport commitReport;
     private long catalogId;
     private long tableId;
-    private List<String> namespace = Collections.emptyList();
     private Instant timestamp;
 
     private CommitReportBuilder(CommitReport commitReport) {
@@ -207,17 +191,6 @@ public final class MetricsRecordConverter {
     }
 
     /**
-     * Sets the namespace as a list of levels.
-     *
-     * @param namespace the namespace levels
-     * @return this builder
-     */
-    public CommitReportBuilder namespace(List<String> namespace) {
-      this.namespace = namespace != null ? namespace : Collections.emptyList();
-      return this;
-    }
-
-    /**
      * Sets the timestamp for the metrics record.
      *
      * <p>This should be the time the metrics report was received by the server, which may differ
@@ -239,7 +212,6 @@ public final class MetricsRecordConverter {
       return CommitMetricsRecord.builder()
           .reportId(UUID.randomUUID().toString())
           .catalogId(catalogId)
-          .namespace(namespace)
           .tableId(tableId)
           .timestamp(timestamp != null ? timestamp : Instant.now())
           .snapshotId(commitReport.snapshotId())
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordIdentity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordIdentity.java
@@ -20,28 +20,24 @@ package org.apache.polaris.core.persistence.metrics;
 
 import com.google.common.annotations.Beta;
 import java.time.Instant;
-import java.util.List;
 import java.util.Map;
 
 /**
  * Base interface containing common identification fields shared by all metrics records.
  *
  * <p>This interface defines the common fields that identify the source of a metrics report,
- * including the report ID, catalog ID, table location, timestamp, and metadata.
+ * including the report ID, catalog ID, table ID, timestamp, and metadata.
  *
  * <p>Both {@link ScanMetricsRecord} and {@link CommitMetricsRecord} extend this interface to
  * inherit these common fields while adding their own specific metrics.
  *
  * <h3>Design Decisions</h3>
  *
- * <p><b>Entity IDs only (no names):</b> We store only catalog ID and table ID, not their names.
- * Names can change over time (via rename operations), which would make querying historical metrics
- * by name challenging and lead to correctness issues. Queries should resolve names to IDs using the
- * current catalog state.
- *
- * <p><b>Namespace as List&lt;String&gt;:</b> Namespaces are stored as a list of levels rather than
- * a dot-separated string to avoid ambiguity when namespace segments contain dots. The persistence
- * implementation handles the serialization format.
+ * <p><b>Entity IDs only (no names):</b> We store only catalog ID and table ID, not their names or
+ * namespace paths. Names can change over time (via rename operations), which would make querying
+ * historical metrics by name challenging and lead to correctness issues. Queries should resolve
+ * names to IDs using the current catalog state. The table ID uniquely identifies the table, and the
+ * namespace can be derived from the table entity if needed.
  *
  * <p><b>Realm ID:</b> Realm ID is intentionally not included in this interface. Multi-tenancy realm
  * context should be obtained from the CDI-injected {@code RealmContext} at persistence time. This
@@ -71,20 +67,12 @@ public interface MetricsRecordIdentity {
   long catalogId();
 
   /**
-   * Namespace path as a list of levels (e.g., ["db", "schema"]).
-   *
-   * <p>This is the namespace portion of the table identifier. Using a list avoids ambiguity when
-   * namespace segments contain dots. The persistence implementation handles the serialization
-   * format.
-   */
-  List<String> namespace();
-
-  /**
    * Internal table entity ID.
    *
    * <p>This matches the table entity ID in Polaris persistence, as defined by {@code
    * PolarisEntityCore#getId()}. The table name is not stored since it can change over time; queries
-   * should resolve names to IDs using the current catalog state.
+   * should resolve names to IDs using the current catalog state. The namespace can be derived from
+   * the table entity if needed.
    */
   long tableId();
 


obelix74:
@dimas-b This is the PR for the SPI, as you suggested. This contains the document feedback incorporated.

evindj (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
```diff
@@ -0,0 +1,162 @@
...
+@PolarisImmutable
+public interface ScanMetricsRecord {
+
+  // === Identification ===
+
+  /** Unique identifier for this report (UUID). */
+  String reportId();
+
+  /** Multi-tenancy realm identifier. */
+  String realmId();
```
dimas-b: Realm ID is generally part of CDI context. I believe it is preferable to keep it out of catalog-specific classes so that the code working inside a catalog won't have to worry about the realm.

Of course, persistence impl. will have to take realm into account.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
```diff
@@ -0,0 +1,162 @@
...
+   *
+   * <p>This is an optional identifier that the Iceberg client may include in the report's metadata
+   * map (typically under the key "trace-id"). It allows clients to correlate this metrics report
+   * with their own distributed tracing system or query execution context.
+   *
+   * <p>Note: Server-side tracing information (e.g., OpenTelemetry trace/span IDs) and principal
+   * information are not included in this record. The persistence implementation can obtain these
+   * from the ambient request context (OTel context, security context) at write time if needed.
+   */
+  Optional<String> reportTraceId();
```
dimas-b: I'd prefer to track it as a contextual piece of data like OTel trace info... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Returns whether this persistence backend supports metrics storage.
+   *
+   * <p>Backends that do not support metrics should return false. Service code should NOT use this
+   * to branch with instanceof checks - instead, call the interface methods directly and rely on the
+   * no-op behavior for unsupported backends.
+   *
+   * @return true if metrics persistence is supported, false otherwise
+   */
+  boolean isSupported();
```
dimas-b: Is this envisioned to be used as an optimization of some kind? Why not just call methods, which will not do anything (when not supported)?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ReportIdToken.java
```diff
@@ -0,0 +1,122 @@
...
+ *   <li>Stable (doesn't change over time)
+ * </ul>
+ *
+ * <p>Each backend implementation can use this cursor value to implement efficient pagination in
+ * whatever way is optimal for that storage system:
+ *
+ * <ul>
+ *   <li>RDBMS: {@code WHERE report_id > :lastReportId ORDER BY report_id}
+ *   <li>NoSQL: Use report ID as partition/sort key cursor
+ *   <li>Time-series: Combine with timestamp for efficient range scans
```
dimas-b, line 56: nit: This can probably work, but I believe it is preferable to allow the Persistence impl. to choose a token impl. that fits the backend's features.

It is ok to keep this class if you're planning to use it for JDBC... I just want to make sure that this token is not guaranteed to be used by all backends.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Client-provided trace ID to filter by (from report metadata).
+   *
+   * <p>This matches the {@code reportTraceId} field in the metrics records, which originates from
+   * the client's metadata map. Useful for correlating metrics with client-side query execution.
+   *
+   * <p>Note: This query pattern may require a custom index in deployment environments. The OSS
+   * codebase does not include an index for trace-based queries.
+   */
+  Optional<String> reportTraceId();
```
dimas-b: I'm not sure trace IDs are so well-defined in Polaris ATM to be used as a primary query parameter 🤔 Are you adding this just because trace IDs were mentioned elsewhere, or do you have a concrete use case in mind?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
```diff
@@ -0,0 +1,142 @@
...
+   * }
+   * }</pre>
+   *
+   * @param criteria the query criteria (filters)
+   * @param pageToken pagination parameters (page size and optional cursor)
+   * @return page of matching scan metrics records with continuation token if more results exist
+   */
+  @Nonnull
+  Page<ScanMetricsRecord> queryScanReports(
+      @Nonnull MetricsQueryCriteria criteria, @Nonnull PageToken pageToken);
```
dimas-b, line 139: `MetricsQueryCriteria` provides a lot of options, but it is not clear which permutations need to be supported. WDYT about adding a separate query method per query pattern (with individual parameters)?

dimas-b: If query patterns are the same between scan and commit reports, we could model that as a sub-interface, e.g. `scanReportQuery().execute(..., PageToken)` and `commitReportQuery().execute(..., PageToken)`

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordConverter.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.iceberg.metrics.ScanReport;
+import org.apache.iceberg.metrics.TimerResult;
+
+/**
+ * Utility class for converting Iceberg metrics reports to SPI record types.
+ *
+ * <p>This converter extracts all relevant metrics from Iceberg's {@link ScanReport} and {@link
+ * CommitReport} and combines them with context information to create persistence-ready records.
+ */
+public final class MetricsRecordConverter {
```
dimas-b: This converter class makes sense, but I guess it is technically not part of the "persistence" SPI... Do you think we could keep it in a different package? Then this SPI will be detached from Iceberg REST API details.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordConverter.java
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Converts an Iceberg ScanReport to a ScanMetricsRecord.
+   *
+   * @param scanReport the Iceberg scan report
+   * @param tableName the table name
+   * @param context the metrics context containing realm, catalog, and request information
+   * @return the scan metrics record ready for persistence
+   */
+  public static ScanMetricsRecord fromScanReport(
+      ScanReport scanReport, String tableName, MetricsContext context) {
```
dimas-b: `MetricsContext` feels like a group of parameters for convenience... WDYT about this pattern: `MetricsRecordConverter.forTable(catalogName, tableName, etc...).convert(ScanReport scanReport)`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
```diff
@@ -0,0 +1,162 @@
...
+  // === Identification ===
+
+  /** Unique identifier for this report (UUID). */
+  String reportId();
+
+  /** Multi-tenancy realm identifier. */
+  String realmId();
+
+  /** Internal catalog ID. */
+  String catalogId();
```
dimas-b: Why `String`? `PolarisEntityCore` defines it as a `long` 🤔

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /** Human-readable catalog name. */
+  String catalogName();
```
dimas-b: WDYT about using `CatalogEntity` instead of separate name / ID attributes?

... same for namespace / table.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/CommitMetricsRecord.java
```diff
@@ -0,0 +1,163 @@
...
+  String reportId();
+
+  /** Multi-tenancy realm identifier. */
+  String realmId();
+
+  /** Internal catalog ID. */
+  String catalogId();
+
+  /** Human-readable catalog name. */
+  String catalogName();
```
dimas-b: Would it be possible to move common data about the entity associated with the report (scan / commit) into a share (base?) interface?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistenceFactory.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Gets or creates a {@link MetricsPersistence} instance for the given realm.
+   *
+   * <p>Implementations may cache instances per realm. If the persistence backend does not support
+   * metrics persistence, this method should return {@link MetricsPersistence#NOOP}.
+   *
+   * @param realmContext the realm context
+   * @return a MetricsPersistence instance for the realm, or NOOP if not supported
+   */
+  MetricsPersistence getOrCreateMetricsPersistence(RealmContext realmContext);
```
dimas-b: Is a factory really required? Would `ServiceProducers` be able to produce an injectable bean without a factory? In any case, the factory feel more like a deployment / CDI concern than a core SPI concern. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
obelix74: Thank you @dimas-b for the thorough review. I've addressed all the feedback:

   1. Removed realmId from metrics records
     • Realm context is now obtained from CDI-injected RealmContext at persistence time
     • Added documentation in MetricsPersistence.java explaining this pattern

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/CommitMetricsRecord.java
obelix74: Created MetricsRecordIdentity base interface
New interface in org.apache.polaris.core.persistence.metrics with common fields:
        • reportId(), catalogId(), catalogName(), namespace(), tableName(), timestamp(), metadata()

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
obelix74: Changed catalogId from String to long. Now matches PolarisEntityCore#getId() as suggested

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
obelix74: Deleted.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistenceFactory.java
obelix74: Deleted MetricsPersistenceFactory. CDI producers in the deployment module handle implementation selection

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
obelix74: Removed reportTraceId field. Removed from ScanMetricsRecord, CommitMetricsRecord, and MetricsQueryCriteria. Added metadata() map to MetricsRecordIdentity for flexible key-value storage if needed

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordConverter.java
obelix74: Replaced it with a builder pattern and deleted MetricsContext.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ReportIdToken.java
obelix74: Updated ReportIdToken Javadoc. Clarified it's a reference implementation, not required by the SPI. Backends can implement their own Token subclass

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
obelix74: Deleted.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
obelix74: Thank you for the suggestion on query sub-interfaces. I agree this could provide better type safety for distinct query patterns. However, for the initial SPI, I'd like to keep the simpler MetricsQueryCriteria approach with optional fields. This allows flexibility without over-engineering the API upfront.

If we find that distinct query patterns emerge (e.g., trace-based queries vs. table-based queries with fundamentally different semantics), we can introduce sub-interfaces in a future iteration. The current design doesn't preclude this evolution.

Would you be okay with deferring this to a follow-up PR once we have more usage patterns to inform the design?

BTW, is this what you were suggesting?

```
   interface MetricsQuery {}
   interface TableMetricsQuery extends MetricsQuery {
       String catalogName();
       String namespace();
       String tableName();
       Optional<Instant> startTime();
       Optional<Instant> endTime();
   }
   interface TraceMetricsQuery extends MetricsQuery {
       String traceId();
   }
```

obelix74: About separate query methods per query pattern, is this what you had in mind?

```
       Page<ScanMetricsRecord> queryScanReportsByTable(
           String catalogName, String namespace, String tableName,
           Instant startTime, Instant endTime, PageToken pageToken);

       Page<ScanMetricsRecord> queryScanReportsByTraceId(
           String traceId, PageToken pageToken);
```

I considered separate query methods per pattern, but opted for the criteria-based approach for extensibility. With MetricsQueryCriteria, adding new filter options (e.g., by principal, by operation type) only requires adding optional fields to the criteria class, not changing the SPI interface.

Implementations can inspect which criteria fields are populated and optimize accordingly. For example, a JDBC implementation might use different SQL queries based on whether traceId or tableName is provided.

However, I am ok with separate query methods as well. Let me know what your preference is.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordConverter.java
obelix74: Moved it to `org.apache.polaris.core.metrics.iceberg` package. Does that work?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
obelix74: For table identification: Using Iceberg's TableIdentifier instead of separate namespace/tableName strings. This encapsulates both namespace path and table name, aligns with Iceberg patterns, and is consistent with `TableLikeEntity.getTableIdentifier()`.

For catalog identification: Keeping catalogId (long) and catalogName (String) as separate primitives rather than using CatalogEntity.

Rationale:

1. Heavyweight object: CatalogEntity extends PolarisEntity and includes storage config, properties, internal properties, timestamps, grant records version - none relevant for metrics identification

2. Immutables incompatibility: CatalogEntity is a class wrapping PolarisBaseEntity, not a @PolarisImmutable interface. The Immutables annotation processor cannot generate proper builders that include it - I can't figure out a solution for this.

3. Serialization complexity: CatalogEntity uses @JsonCreator/@JsonProperty annotations that don't integrate cleanly with Immutables-generated classes

4. Minimal data needed: For metrics records, we only need catalog ID (for foreign key relationships) and name (for display/query convenience)

obelix74:
@dimas-b Thank you again for the detailed review. I have addressed all your questions. There are a few open items.

* Did not use `CatalogEntity` - explained why.
* Added comments about query sub-interfaces and separate query methods. I will follow your lead when you respond.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
dimas-b: RDBMS is versatile in that sense. The whole paradigm implies the queries are not known ahead of time and you tune the database with indexes after query patterns are discovered... However, in NoSQL worlds the usual design approach is reversed - one often needs to know query patters upfront for an efficient storage design.

That said, it is certainly possible to have an efficient impl. if the query patterns are known even if they are expressed in a general-purpose interface.

All in all, I think we may want to flag this feature as "beta" in Polaris and go with the generic query interface for a start.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordConverter.java
dimas-b: yes. thx!

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
dimas-b: I'd prefer to avoid Iceberg classes like `TableIdentifier` in a Polaris SPI.

dimas-b: Good points about heavy objects in an "Immutable" class :thinking: However, catalog name can generally [change](https://github.com/apache/polaris/blob/d95bd6b1e3d77b5db7769e866c0474f08accad2c/polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java#L313) over time, right? Why would we persist it in a metrics record? Querying a time-series over a changeable attribute can be very challenging :thinking: 

Why not store (unique) IDs only inside Metrics Records? Granted, queries will have to resolve names to IDs first, but that can be done using the latest state of the catalog

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
obelix74: Sounds good.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
```diff
@@ -0,0 +1,146 @@
...
+ * @see org.apache.polaris.core.persistence.pagination.Page
+ * @see ReportIdToken
+ */
+@PolarisImmutable
+public interface MetricsQueryCriteria {
+
+  // === Table Identification (optional) ===
+
+  /** Catalog name to filter by. */
+  Optional<String> catalogName();
```
dimas-b: Catalog name is flagged as "convenience" in `MetricsRecordIdentity`, but may be used for queries here... It's a bit of a contradiction IMHO... I think it's ok to store it as a convenience attribute, if we do not have to query by it... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
```diff
@@ -0,0 +1,146 @@
...
+
+  /** Namespace to filter by (dot-separated). */
+  Optional<String> namespace();
```
dimas-b: dot-separated value can be tricky to query if name segments contain dots too... TBH, I'm not sure how well Polaris supports that now, but if escaping is in use it can be very non-obvious.

Why not use a `List<String>` here and let the impl. deal with persistent representation?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * <p>This enables filtering metrics by client-provided correlation data stored in the record's
+   * metadata map. For example, clients may include a trace ID in the metadata that can be queried
+   * later.
+   *
+   * <p>Note: Metadata filtering may require custom indexes depending on the persistence backend.
+   * The OSS codebase provides basic support, but performance optimizations may be needed for
+   * high-volume deployments.
+   */
+  java.util.Map<String, String> metadata();
```
dimas-b: nit: please use imports

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * @param catalogName the catalog name
+   * @param namespace the namespace (dot-separated)
+   * @param tableName the table name
+   * @param startTime the start time (inclusive)
+   * @param endTime the end time (exclusive)
+   * @return the query criteria
+   */
+  static MetricsQueryCriteria forTable(
+      String catalogName, String namespace, String tableName, Instant startTime, Instant endTime) {
```
dimas-b: nit: Would it be worth returning a Builder with just the table info and adding time ranges later (at the call site)?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
obelix74: Agreed. I've updated the SPI to use separate namespace (String) and tableName (String) primitives instead of TableIdentifier. This avoids Iceberg dependencies in the Polaris SPI.

The conversion from TableIdentifier to these primitives now happens in MetricsRecordConverter (in org.apache.polaris.core.metrics.iceberg package), where Iceberg dependencies are already present.

Good point about catalog renames. I've removed catalogName from the SPI records and now store only catalogId. The query criteria (MetricsQueryCriteria) now uses catalogId (OptionalLong) instead of catalogName.

For tables, I'm keeping namespace and tableName as strings rather than storing table entity IDs. While tables can also be renamed, users typically query metrics by table name rather than by internal entity ID. This is a pragmatic compromise - catalog renames are significant lifecycle events, while table renames within the same namespace are more routine and less likely to break historical queries.

Let me know if you'd prefer to also use table entity IDs instead.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
obelix74: This was replaced by `catalogId` based on feedback here. https://github.com/apache/polaris/pull/3616#discussion_r2766276818

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordIdentity.java
```diff
@@ -0,0 +1,97 @@
...
+   * represented with dots separating levels.
+   */
+  String namespace();
+
+  /**
+   * Table name.
+   *
+   * <p>This is the table name portion of the table identifier, without the namespace prefix.
+   */
+  String tableName();
```
dimas-b: Why not table ID (same as for the catalog ID) 🤔 Table names can change over time.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
dimas-b: I'm not completely sure about denormalizing table names into the metrics table. With renames, a different table may appear under the same name in the same namespace at different times... I tend to think it can easy become a correctness issue at query time.... IDs seem more resilient... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
obelix74: Updated both MetricsRecordIdentity and MetricsQueryCriteria to use List<String> for namespace. The persistence implementation will handle the serialization format

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
obelix74: Fixed. Now using proper imports for Map, List, Optional, and OptionalLong.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/ScanMetricsRecord.java
obelix74: ok. Updated to use tableId (long) instead of tableName (String) in both MetricsRecordIdentity and MetricsQueryCriteria. The caller now resolves table entity ID before creating records, similar to how we handle catalogId.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
obelix74: Updated forTable() to return a Builder pre-populated with catalogId and tableId, allowing the caller to add time ranges at the call site.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsRecordIdentity.java
obelix74: Fixed.

dimas-b:
@obelix74 : please rebase to fix CI

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsQueryCriteria.java
```diff
@@ -0,0 +1,164 @@
...
+   */
+  OptionalLong catalogId();
+
+  /**
+   * Namespace to filter by.
+   *
+   * <p>The namespace is represented as a list of levels to avoid ambiguity when segments contain
+   * dots. An empty list means no namespace filter is applied.
+   */
+  List<String> namespace();
```
dimas-b: Sorry to cycle back on this... Now that we have IDs for catalog and table, does it make sense to have name for namespace parts? 🤔 

Is the idea to query all records for all tables under a namespace? This can be confusing with table moves over time.

The namespace can still be a query parameter at the user level, but we could just resolve namespace to table IDs using the current catalog state and then query metrics persistence using table IDs. WDYT?

obelix74: Makes sense.    

I've removed namespace() from MetricsQueryCriteria. Queries are now by catalogId + tableId only (both required). If users want to query by namespace, the service layer should resolve namespace → table IDs using the current catalog state, then query by those IDs.

I've kept namespace in MetricsRecordIdentity for storage/display purposes - it's useful context when showing metrics to users, even though we don't query by it.

polaris-core/src/main/java/org/apache/polaris/core/persistence/metrics/MetricsPersistence.java
```diff
@@ -0,0 +1,147 @@
...
+ * </ul>
+ *
+ * <p>The {@link ReportIdToken} provides a reference cursor implementation based on report ID
+ * (UUID), but backends may use other cursor strategies internally.
+ *
+ * @see PageToken
+ * @see Page
+ * @see ReportIdToken
+ */
+public interface MetricsPersistence {
```
dimas-b, line 84: Given that this is a new SPI and not fully implemented in all Persistence backends yet, it would be nice to flag it as "experimental" for the sake of clarity... but we do not have a common patterns for this in Polaris 😅 

Perhaps we could try `com.google.common.annotations.Beta` in this particular case... as a POC 🤔 WDYT?

obelix74: Added @Beta annotation from Guava (com.google.common.annotations.Beta) to all public types in the Metrics Persistence SPI package

dimas-b:
This PR is one week old and it looks like it did not attract attention from many contributors :sweat_smile: (thanks for your review, @evindj !). So, unless fresh comments appear, I'm planning to merge tomorrow.

Since the new SPI is marked as `Beta`, if concerns are uncovered later, we'll address on `main`.

polaris-core/src/main/java/org/apache/polaris/core/metrics/iceberg/MetricsRecordConverter.java
```diff
@@ -115,6 +116,20 @@ public ScanReportBuilder namespace(List<String> namespace) {
...
+     * Sets the timestamp for the metrics record.
+     *
+     * <p>This should be the time the metrics report was received by the server, which may differ
+     * from the time it was recorded by the client.
+     *
+     * @param timestamp the timestamp
+     * @return this builder
+     */
+    public ScanReportBuilder timestamp(Instant timestamp) {
+      this.timestamp = timestamp;
```
obelix74, line 115: @dimas-b noticed this bug when I was rebasing https://github.com/apache/polaris/pull/3385 to this.

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/metrics/iceberg/MetricsRecordConverter.java
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * Sets the namespace as a list of levels.
+     *
+     * @param namespace the namespace levels
+     * @return this builder
+     */
+    public ScanReportBuilder namespace(List<String> namespace) {
+      this.namespace = namespace != null ? namespace : Collections.emptyList();
+      return this;
+    }
```
singhpk234: minor: why is namespace not an entity id ?

polaris-core/src/main/java/org/apache/polaris/core/metrics/iceberg/MetricsRecordConverter.java
```diff
@@ -0,0 +1,300 @@
...
+                  ? Optional.of(scanReport.filter().toString())
+                  : Optional.empty())
+          .projectedFieldIds(
+              scanReport.projectedFieldIds() != null
+                  ? scanReport.projectedFieldIds()
+                  : Collections.emptyList())
+          .projectedFieldNames(
+              scanReport.projectedFieldNames() != null
+                  ? scanReport.projectedFieldNames()
+                  : Collections.emptyList())
```
singhpk234, line 142: minor: can we not make it nullable instead ?

polaris-core/src/main/java/org/apache/polaris/core/metrics/iceberg/MetricsRecordConverter.java
obelix74: As discussed, removed namespace. 

  • Removed namespace() method from MetricsRecordIdentity.java
  • Removed namespace() calls from MetricsRecordConverter.java builders

dimas-b:
@singhpk234 : Are you ok to merge this PR?

Issue #2743, opened 2025-10-01
Support Dell ECS On-Premise S3 Object Store with STS support


rohangoli:
### Is your feature request related to a problem? Please describe.

Using Polaris with On-Premise S3 Storage like Dell ECS (in-built S3, IAM & STS services), code built on [1.2.0-incubating-SNAPSHOT](https://github.com/apache/polaris/tree/main) fails at following:
- Create Table fails with #2207 error
```
org.apache.iceberg.exceptions.RESTException: Unable to process: Failed to get subscoped credentials: (Service: Sts, Status Code: 400, Request ID: null) (SDK Attempt Count: 1)
```
- Fails to handle custom STS endpoints with suffix (ex: https://ecs-prod.example.com:4443/sts) [NO references with #1146 & #2672]
- Fails to parse ECS-Style STS XML response as Fallback when non-AWS Style response comes - [ECS Management API - IAM - STS Role](https://portal.ecstestdrive.com/Document/ECS/3.8/apidocs/StsService_AssumeRole_d41289fb2f39151a98d43ffd16dabaab_cfc300b9ce827245532e8894bd9ddbbb_detail.html) [NO references with #1146 & #2672]
    Or
  Instead use standardized JSON implementation
```
REQUEST:
POST https://ecs-prod.example.com:4443/sts?Action=AssumeRole&RoleSessionName=temp&RoleArn=urn:ecs:iam::s3:role/s3assumeRole&DurationSeconds=7200
Accept: */*
X-SDS-AUTH-TOKEN: <AUTH_TOKEN>

RESPONSE:
HTTP/1.1 200 OK
Content-Type: application/xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<ns2:AssumeRoleResponse xmlns:ns2="none">
  <ResponseMetadata>
    <RequestId>0af7b369:171a31ad5b8:12e3a:b</RequestId>
  </ResponseMetadata>
  <AssumeRoleResult>
    <AssumedRoleUser>
      <Arn>urn:ecs:sts::s3:assumed-role/s3assumeRole/temp</Arn>
      <AssumedRoleId>AROA48AEC87ADCFF133F:temp</AssumedRoleId>
    </AssumedRoleUser>
    <Credentials>
      <AccessKeyId>ASIA48DB3A69C19CFE56</AccessKeyId>
      <Expiration>2020-04-23T19:45:46Z</Expiration>
[...]

PR #3619, opened 2026-01-30
STS roleArn: enable 3rd party STS services

snazy:
Certain non-AWS STS service implementations use role ARNs that look quite different from AWS ones. This change shall enable those STS implementations.

Example role ARNs that currently fail:
* `urn:ecs:sts::s3:assumed-role/s3assumeRole/user1-105-temp`
* `urn:sgws:identity::12345:group/foo-bar-abcdef`

Related issue #2743

### DIFF
--- a/polaris-core/src/test/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfoTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfoTest.java
@@ -162,7 +162,10 @@ public class AwsStorageConfigurationInfoTest {
         Arguments.of("arn:aws:iam::012345678901:role/jdoe", "012345678901", "aws"),
         Arguments.of("arn:aws-us-gov:iam::012345678901:role/jdoe", "012345678901", "aws-us-gov"),
         Arguments.of("arn:aws-cn:iam::012345678901:role/jdoe", "012345678901", "aws-cn"),
-        Arguments.of("urn:ecs:sts::s3:assumed-role/s3assumeRole/user1-105-temp", "s3", "ecs"),
+        // Following are only there to have test coverage for non-AWS role-ARNs/URNs. ARN specific
+        // parts are irrelevant for those URNs.
+        Arguments.of(
+            "urn:ecs:sts::namespace:assumed-role/s3assumeRole/user1-105-temp", "namespace", "ecs"),
         Arguments.of("urn:sgws:identity::12345:group/foo-bar-abcdef", "12345", "sgws"));
   }
 }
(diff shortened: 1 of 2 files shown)


adutra (approved):

dimas-b:
The AWS [pattern](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) does not appear to require even the `:` separators... It might be worth relaxing the check even more.

snazy:
> The AWS [pattern](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) does not appear to require even the `:` separators... It might be worth relaxing the check even more.

Not strictly necessary in this change though. We have these partition + accountId extractions, which could yield `null` in odd cases.

MonkeyCanCode (approved):

MonkeyCanCode:
This will help us to do assume role on rustfs/minio as well. Wondering if we should add integration testing for this when using minio for the existed test cases?

CHANGELOG.md
```diff
@@ -68,6 +68,8 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### Fixes
 
+- Enable non-AWS STS role ARNs
```
singhpk234, line 71: minor : worth mentioning ECS as an example

polaris-core/src/test/java/org/apache/polaris/core/storage/aws/AwsStorageConfigurationInfoTest.java
```diff
@@ -148,6 +154,15 @@ public void testRoleArnParsing() {
...
-        .containsExactly("arn:aws:iam::012345678901:role/jdoe", "012345678901", "aws");
+        .containsExactly(roleArn, expectedAccountId, expectedPartition);
+  }
+
+  static Stream<Arguments> testRoleArnParsing() {
+    return Stream.of(
+        Arguments.of("arn:aws:iam::012345678901:role/jdoe", "012345678901", "aws"),
+        Arguments.of("arn:aws-us-gov:iam::012345678901:role/jdoe", "012345678901", "aws-us-gov"),
+        Arguments.of("arn:aws-cn:iam::012345678901:role/jdoe", "012345678901", "aws-cn"),
+        Arguments.of("urn:ecs:sts::s3:assumed-role/s3assumeRole/user1-105-temp", "s3", "ecs"),
```
singhpk234: I didn't fully understand this example, expected account id is `s3` ? can you please elaborate and add a code comment

adnanhemani: +1

snazy:
Add comment + moved AWS/KMS specific usage of 'account id' from an ARN to where it is used (only for KMS)

dimas-b:
@MonkeyCanCode : AssumeRole for MinIO is already covered by `RestCatalogMinIOSpecialIT` and `PolarisRestCatalogMinIOIT`

PR #3636, opened 2026-02-01
"Stale" job: restrict executions and adjust issue permissions

adutra:
The job wasn't capable of closing issues because of insufficient permissions.
This change also makes the job run only on the `apache/polaris` repo.

## Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,6 +20,7 @@
 ---
 jobs:
   stale:
+    if: github.repository == 'apache/polaris'
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/stale@d6f8a33132340b15a7006f552936e4b9b39c00ec
@@ -33,5 +34,5 @@ on:
   schedule:
     - cron: "30 1 * * *"
 permissions:
-  issues: read
+  issues: write
   pull-requests: write


jbonofre:
Good catch ! Thanks @adutra !

snazy (approved):

.github/workflows/stale.yml
```diff
@@ -33,5 +34,5 @@ on:
   schedule:
     - cron: "30 1 * * *"
 permissions:
-  issues: read
+  issues: write
```
dimas-b, line 37: Do we really want to close stale issues? I believe some issues are reported by users who do not have capacity for in-depth code changes in Polaris and thus allow the issues to just be there for other to review and propose solutions. Closing all issues without review seems risky in that regard... Would just flagging them as "stale" be sufficient? 🤔

snazy: You get a "warning"/"heads up" comment first, before the issue is being closed.

After the "heads up" you have two options:
* Comment on the issue, and the "grace time" starts from 0.
* Remove the "Stale" label and the issue won't be closed.

This is IMO better than having untackled issues lingering around forever?

dimas-b: That process has merit, I do not question that, and I'm ok with that if that's what the community agrees on.

My point is slightly different, though: if the original issue reporter disengages from the project and does not comment after the issue is flagged as stale, it does not mean that the issue is no longer relevant. Closing such issues automatically, might not be ideal.

adutra: While the practice of automatically closing issues is debatable, the immediate goal here is to re-enable this feature, as it was clearly intended to be operational from the start of Polaris but was non-functional due to a misconfiguration.

I suggest merging this change to correct the misconfiguration. We should then move the broader discussion about the practice itself to the ML, rather than silently relying on a misconfiguration to stop the feature from working as intended.

WDYT?

snazy: > I suggest merging this change  ... then move the broader discussion about the practice itself to the ML

+1

dimas-b: Fair enough :+1:

PR #3647, opened 2026-02-03
Releasey: adjust workflwo for Apache org level secrets

snazy:
See reference [INFRA-27430](https://issues.apache.org/jira/browse/INFRA-27430), requiring us to use `DOCKERHUB_USER` + `DOCKERHUB_TOKEN` instead of `DOCKERHUB_USERNAME` + `DOCKERHUB_TOKEN`.


snazy:
The "old"/"shadowing" secrets have been removed.

PR #3658, opened 2026-02-03
Fix FileIO usage in PolarisRestCatalogIntegrationBase

adutra:
Currently, all tests inheriting from `PolarisRestCatalogIntegrationBase` are forcibly using `InMemoryFileIO` for all `RESTCatalog` instances.

Furthermore, this class wasn't overriding `CatalogTests.baseTableLocation()`, so tests calling this method were actually operating on `file://` paths all the time.

Because of that, some child test classes, and in particular, the MiniIO-based ones, weren't actually testing anything useful.

This PR changes `PolarisRestCatalogIntegrationBase` to use a configurable `FileIO` impl, defaulting to `ResolvingFileIO`.

This change uncovered a few latent issues with current tests, and notably:

1) Tests using MinIO should request credential vending consistently, otherwise most tests would fail because the S3 client attempts to load credentials from the environment.

2) Tests calling `registerTable` need special handling since this endpoint does NOT support access delegation at all. For these tests, the `RESTCatalog` clients need to be given direct credentials for S3 – since the server won't give them anything – even with vended credentials requested.

Incidentally, this PR simplifies `IntegrationTestsHelper.mergeFromAnnotatedElements()` – but its functionality stays the same.

## Checklist

### DIFF
--- a/runtime/service/src/cloudTest/java/org/apache/polaris/service/it/RestCatalogAdlsCredentialVendingIT.java
+++ b/runtime/service/src/cloudTest/java/org/apache/polaris/service/it/RestCatalogAdlsCredentialVendingIT.java
@@ -20,10 +20,12 @@ package org.apache.polaris.service.it;
 
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 import io.quarkus.test.junit.TestProfile;
+import org.apache.polaris.service.it.env.RestCatalogConfig;
 import org.apache.polaris.service.it.test.PolarisRestCatalogAdlsIntegrationTestBase;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 
 @QuarkusIntegrationTest
 @TestProfile(CredentialVendingProfile.class)
 @EnabledIfEnvironmentVariable(named = "INTEGRATION_TEST_AZURE_PATH", matches = ".+")
+@RestCatalogConfig({"header.X-Iceberg-Access-Delegation", "vended-credentials"})
 public class RestCatalogAdlsCredentialVendingIT extends PolarisRestCatalogAdlsIntegrationTestBase {}


adutra:
Heads up: this change *may* have implications for the following cloud tests:

* `RestCatalogS3IT`
* `RestCatalogGcsIT`
* `RestCatalogAdlsIT`
* `RestCatalogAdlsCredentialVendingIT`

However I cannot test them 🤷‍♂️ 

The first 3 are probably fine: credential vending is implicitly enabled, but the clients don't request any access delegation mode by default. This should work as long as the `RESTCatalog` clients have access to their own storage credentials via the environment (which seems to be the intent).

The last one, `RestCatalogAdlsCredentialVendingIT`, _explicitly_ enables credentials vending, but in fact, behaves exactly like `RestCatalogAdlsIT`, i.e., the `RESTCatalog` clients are not configured to request any delegation mode. Imho this test is lacking the annotation:

```java
@RestCatalogConfig({"header.X-Iceberg-Access-Delegation", "vended-credentials"})
```

runtime/service/src/intTest/java/org/apache/polaris/service/it/PolarisRestCatalogMinIOIT.java
```diff
@@ -88,4 +95,15 @@ protected StorageConfigInfo getStorageConfigInfo() {
 
     return storageConfig.build();
   }
+
+  @Override
+  protected Map<String, String> extraCatalogProperties(TestInfo testInfo) {
```
adutra, line 100: This is admittedly hacky, but I couldn't find a better way to tweak _just_ this test's `RESTCatalog` without introducing unwanted S3 credentials in all other tests. Unfortunately we can't annotate the test with `@RestCatalogConfig` because `endpoint` is computed dynamically.

dimas-b:
> The last one, RestCatalogAdlsCredentialVendingIT, explicitly enables credentials vending, but in fact, behaves exactly like RestCatalogAdlsIT, i.e., the RESTCatalog clients are not configured to request any delegation mode. Imho this test is lacking the annotation: [...] @RestCatalogConfig({"header.X-Iceberg-Access-Delegation", "vended-credentials"})

It might be good to add that annotation, yet `RestCatalogAdlsCredentialVendingIT` behaves differently from `RestCatalogAdlsIT`. I added the former specifically to test hierarchical SAS tokens in ADLS. The latter test does not exercise storage integration code at all, but the former does.

runtime/service/src/intTest/java/org/apache/polaris/service/it/PolarisRestCatalogMinIOIT.java
```diff
@@ -88,4 +95,15 @@ protected StorageConfigInfo getStorageConfigInfo() {
...
+    if (testInfo.getTestMethod().orElseThrow().getName().equals("testRegisterTable")) {
+      // This test registers a table – operation that doesn't support access delegation –
+      // then attempts to use the table's FileIO. This can only work if the client has its
+      // own S3 credentials.
```
dimas-b, line 104: Interesting... Is it possible that Polaris neglects to return the credentials in the registerTable response, or is it a problem in the IRC API spec?

adutra: It's a problem in the spec: the `registerTable` endpoint does not declare the `data-access` header parameter:

https://github.com/apache/iceberg/blob/d0654440a4cd00feaff73c53cd3ef9fb787b92dc/open-api/rest-catalog-open-api.yaml#L883-L898

I will reach out to the Iceberg ML

adutra:
> It might be good to add that annotation, yet `RestCatalogAdlsCredentialVendingIT` behaves differently from `RestCatalogAdlsIT`. I added the former specifically to test hierarchical SAS tokens in ADLS. The latter test does not exercise storage integration code at all, but the former does.

Just curious, what triggers the usage of a hierarchical SAS token in the `RestCatalogAdlsCredentialVendingIT` test? As far as I can tell, both tests are identical since `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION` is false by default.

In any case, let's do this change in a follow-up PR.

dimas-b:
> Just curious, what triggers the usage of a hierarchical SAS token in the RestCatalogAdlsCredentialVendingIT test?

This: 

https://github.com/apache/polaris/blob/e5b51225bba00531abc5dc6729d69eb9d27ac022/runtime/service/src/cloudTest/java/org/apache/polaris/service/it/CredentialVendingProfile.java#L31

Vended credentials may not be used by clients (as you noted), but the server-side call path when it interacts with storage is the same as for vended credentials (i.e. it vends credentials to itself).

runtime/service/src/intTest/java/org/apache/polaris/service/it/PolarisRestCatalogMinIOIT.java
adutra: https://lists.apache.org/thread/360z73c29zw2gm708g8l11cx4b4z4vrn

adutra:
> the server-side call path when it interacts with storage is the same as for vended credentials (i.e. it vends credentials to itself).

Sure, but in `RestCatalogAdlsIT` too the server-side credential vending is enabled, since `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION` is false by default. I'm not seeing any difference between these two tests 😅

dimas-b:
> SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION is false by default [...]

I guess it's set to `true` for integration tests here:

https://github.com/apache/polaris/blob/0c790e0ad48d69d97033bdde08d3bf4bb415210f/runtime/defaults/src/main/resources/application-it.properties#L49

It's a bit convoluted, indeed :sweat_smile:

adutra:
> It's a bit convoluted, indeed 😅

Aaah right, I missed that! Thank you for the explanation.

runtime/service/src/intTest/java/org/apache/polaris/service/it/PolarisRestCatalogMinIOIT.java
adutra: Heads up: with the new register table endpoint with access delegation coming in Iceberg 1.11, this trick will become obsolete.

adutra:
@dimas-b as a follow up to your explanation, I went ahead and added `@RestCatalogConfig({"header.X-Iceberg-Access-Delegation", "vended-credentials"})` to `RestCatalogAdlsCredentialVendingIT` as it most likely required for this test to pass.

Can you have another look please?

runtime/service/src/intTest/java/org/apache/polaris/service/it/PolarisRestCatalogMinIOIT.java
snazy: The `IcebergCatalogHandler.registerTable` implementation just behaves differently than `IcebergCatalogHandler.loadTable`. Both return a `LoadTableResponse`, but the former doesn't vend credentials. I think, `registerTable` could easily do that, because the table exists when it returns?
(A change for a different PR though)

PR #3660, opened 2026-02-03
Add the apache security webpage URL

jbonofre:


adnanhemani (approved):

singhpk234:
Thanks @jbonofre for the prompt fix ! 
https://whimsy.apache.org/pods/project/polaris

RussellSpitzer (commented):
I'm not sure this is correct for the fix, I think this may have gotten clobbered on top of another pr.

Also I recommend we also change in 

hugo.yaml

```
- name: "Security"
      parent: "asf"
      url: "/community/security-report"
      weight: 46
 ```

To link to our own internal Security reporting instructions

---

This is I believe what whimsy is currently complaining about 

See

```
https://github.com/apache/polaris/blob/main/SECURITY.md | URL expected to match regular expression: ^(https?://.*apache.org\|[^:]*)/.*[Ss]ecurity"Security" should link to either to a project-specific page [...], or to the main http://www.apache.org/security/ page.

```

https://github.com/apache/polaris/blob/605114dfdac9c8aaf84f6e6db25eeb3ded7bc170/site/hugo.yaml#L228-L232

File: SECURITY.md [hunk 1]
```diff
@@ -19,6 +19,6 @@
 
 # Security Policy
 
-Apache Polaris follows the process from [ASF security team](https://www.apache.org/security/).
+Apache Polaris follows the process from [ASF security team](https://www.apache.org/security/) (https://www.apache.org/security).
```
dimas-b, line 22: I guess to please Whimsy we need to remove the trailing slash from the current URL: `https://www.apache.org/security` ... that's what Whimsy's RegEx shows 🤔 Why bother duplicating the link?

flyrain: Agreed, we might just remove the trailing slash.

RussellSpitzer: As I noted in my comment above, I don't think this is what whimsy was signaling on, it could be but I think it's the Security section being added by hugo. I also think we should link to our security reporting instructions

RussellSpitzer:
Since JB is asleep I did a quick fix to what I think is the issue based on the whimsy source code. 
https://github.com/apache/polaris/pull/3662/files

Feel free to ignore but I believe this is what was setting whimsy off, a non-apache link out related to "security"

SECURITY.md
dimas-b: Whimsy pointed to `SECURITY.md` specifically :shrug:

PR #3669, opened 2026-02-04
Remove "giant" constructors from Handlers and Adapters

adutra:
This PR attempts to reduce the "giant constructor" symptoms in catalog adapters and handlers.

It combines two techniques:

* The injectable runtime dependencies were grouped together in a "runtime" object, inspired by the "Parameter Object" pattern. These objects are `@PolarisImmutable` objects, and thus can leverage the generated builder to create instances without the need for big constructors.
* The runtime object is then produced by a new factory, which is a CDI request-scoped bean with direct field injection to avoid constructor injection.

**UPDATE**

* A second commit removed the Parameter Object pattern and made `CatalogHandler` and subtypes directly annotated with `@PolarisImmutable`.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
@@ -697,6 +697,9 @@ public class IcebergCatalogAdapter
   @Override
   public Response getConfig(
       String warehouse, RealmContext realmContext, SecurityContext securityContext) {
+    if (warehouse == null) {
+      throw new BadRequestException("Please specify a warehouse");
+    }
     return withCatalogByName(
         securityContext, warehouse, catalog -> Response.ok(catalog.getConfig()).build());
   }
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
@@ -1315,15 +1315,6 @@ public abstract class IcebergCatalogHandler extends CatalogHandler implements Au
   }
 
   public ConfigResponse getConfig() {
-    // 'catalogName' is taken from the REST request's 'warehouse' query parameter.
-    // 'warehouse' as an output will be treated by the client as a default catalog
-    //   storage base location.
-    // 'prefix' as an output is the REST subpath that routes to the catalog
-    //   resource, which may be URL-escaped catalogName or potentially a different
-    //   unique identifier for the catalog being accessed.
-    if (catalogName() == null) {
-      throw new BadRequestException("Please specify a warehouse");
-    }
     Resolver resolver = resolverFactory().createResolver(polarisPrincipal(), catalogName());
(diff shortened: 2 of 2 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -59,43 +52,25 @@ public abstract class CatalogHandler {
   // Initialized in the authorize methods.
   protected PolarisResolutionManifest resolutionManifest = null;
 
-  protected final ResolutionManifestFactory resolutionManifestFactory;
```
dimas-b, line 62: Just an idea: If we convert fields to abstract getters and make implementations of this class `@Immutable`, the `CatalogHandlerRuntime` object will probably become unnecessary and impl. instances will have direct builders... WDYT?

adutra: I had this approach initially but changed for the Parameter Object pattern, for a few reasons:

1. `CatalogHandler` and subclasses are inherently mutable. Fields like `resolutionManifest` are lazily evaluated, but are hard to migrate to a `@Value.Lazy` equivalent because they may be evaluated more than once.
2. Getters in `CatalogHandler` would have to be public, because it's in a different package from its subclasses, and thus the generated builders only compile if the getters are public.

That said, I don't mind showing my initial version, maybe it's acceptable indeed. The advantage being, that we could look into removing the mutable state little by little until the class hierarchy becomes fully immutable.

I will add a commit on top of this branch to show the approach.

adutra: Done, cf. commit 63c3abd.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -56,48 +55,36 @@
...
+  public abstract PolarisAuthorizer authorizer();
+
   protected PolarisResolutionManifest newResolutionManifest() {
-    return resolutionManifestFactory.createResolutionManifest(polarisPrincipal, catalogName);
+    return resolutionManifestFactory().createResolutionManifest(polarisPrincipal(), catalogName());
   }
 
+  // Initialized in the authorize methods.
+  @SuppressWarnings("immutables:incompat")
+  protected PolarisResolutionManifest resolutionManifest = null;
```
dimas-b, line 86: This might be fix/addressed if we go ahead with what @sungwy proposed for AuthZ refactoring plus my additional proposal for handling the "manifest" per API call: https://lists.apache.org/thread/srqp2jtts5438drzcvky47mdt8zs80wt

dimas-b (dismissed):
Thanks for considering my proposal, @adutra !

I tend to like using direct builders (despite having to suppress build warnings from the Immutables library) because it looks less verbose than using Object Parameters.

However, I'm not attached to this approach 😉 I wonder what other maintainers think 🤔

snazy (dismissed):
Nice work!

The constructors became bigger and bigger, I hope that we'll eventually, in the future, be able not not change a lot of sources to add a reference.

snazy (approved):

PR #3663, opened 2026-02-04
[DO-NOT-MERGE] Change MinIO to RustFS for integration testing

MonkeyCanCode:
I am not sure if we want to take this PR as it is changing existed integration testing on S3-compatible storage from MinIO to RustFS instead of creating a new test-containers. But here are the high level changes from this PR:
1. Refactored MinIO test-containers to RustFS test-containers
2. Use fixed port for RustFS test-containers due to it needs <domain>:<service_port> when using virtual host style access instead of only domain (details in https://github.com/rustfs/rustfs/issues/1593)
3. Changed S3 exception message as RustFS returns it without period at the end (change in integration-tests/src/main/java/org/apache/polaris/service/it/test/CatalogFederationIntegrationTest.java)
4. Removed custom cmd used by MinIO as RustFS image has those configured already
5. Changed healthcheck port from 9000 (used by MinIO) to 9001 (used by RustFS)

ML: https://lists.apache.org/thread/8o31ly7cd8ov70opjbtg630qlhrfl5yh

## Checklist
- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [x] 💡 Added comments for complex logic
- [x] 🧾 Updated `CHANGELOG.md` (if needed)
- [x] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3679, opened 2026-02-05
Add support for rustfs test-containers

MonkeyCanCode:
This is follow up PR for https://github.com/apache/polaris/pull/3663. As concluded in the reference PR, we would want to have both minio test-containers and rustfs test-containers co-existed and migrated NoSQL and Federation tests to RustFS.

## Checklist

### DIFF
--- a/tools/rustfs-testcontainer/src/main/java/org/apache/polaris/test/rustfs/RustfsAccess.java
+++ b/tools/rustfs-testcontainer/src/main/java/org/apache/polaris/test/rustfs/RustfsAccess.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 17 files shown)


tools/rustfs-testcontainer/src/main/java/org/apache/polaris/test/rustfs/RustfsAccess.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Provides access to Rustfs via a preconfigured S3 client and providing the by default randomized
+ * bucket and access/secret keys.
+ *
+ * <p>Annotate JUnit test instance or static fields or method parameters of this type with {@link
+ * Rustfs}.
+ */
+// CODE_COPIED_TO_POLARIS from Project Nessie 0.104.2
+public interface RustfsAccess {
```
dimas-b, line 36: I wonder if we could rename `MinioAccess` to `S3Access` and reuse it for RustFS... and potentially for other non-AWS backends later... WDYT?

I'm fine with a copy too.

dimas-b (approved):
LGTM with a minor comment. Thanks, @MonkeyCanCode !

tools/rustfs-testcontainer/src/main/java/org/apache/polaris/test/rustfs/RustfsAccess.java
MonkeyCanCode: Using `S3Access` sounds great but I am not sure if they are truly re-usable (e.g. switch from minio to rustfs would required couple changes already such as environments variables used by them as well as the domain with port when using virtual host style access). 

My only concern is, with generic naming, will that be more confusing as we can't really promised switching to another s3 compatible will work (e.g. rustfs to seaweedfs)? I am assuming we would drop minio one day and maybe do the change around that time instead?

dimas-b: Good points! I'm ok to merge "as is" :slightly_smiling_face:

PR #3681, opened 2026-02-06
core: Authorization SPI Refactor (Draft)

sungwy:
This is a Draft PR I prepared to aid reviewing the new [Authorization SPI Refactoring RFC](https://docs.google.com/document/d/1OaiQG_C4-yUe0ihaDBxtw_mEcOOzUBnWPazzVbjQi5U/edit?usp=sharing)

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
@@ -24,88 +24,83 @@ import java.util.List;
 import java.util.Set;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisPrivilege;
+import org.apache.polaris.immutables.PolarisImmutable;
 
 /**
  * Authorization request inputs for pre-authorization and core authorization.
  *
  * <p>This wrapper keeps authorization inputs together while preserving legacy semantics.
  */
-public final class AuthorizationRequest {
-  private final PolarisPrincipal principal;
-  private final PolarisAuthorizableOperation operation;
-  private final List<PolarisSecurable> targets;
-  private final List<PolarisSecurable> secondaries;
-
-  public AuthorizationRequest(
+@PolarisImmutable
+public interface AuthorizationRequest {
+  static AuthorizationRequest of(
       @Nonnull PolarisPrincipal principal,
       @Nonnull PolarisAuthorizableOperation operation,
       @Nullable List<PolarisSecurable> targets,
       @Nullable List<PolarisSecurable> secondaries) {
-    this.principal = principal;
-    this.operation = operation;
-    this.targets = targets;
-    this.secondaries = secondaries;
+    return ImmutableAuthorizationRequest.builder()
+        .principal(principal)
+        .operation(operation)
+        .targets(targets)
+        .secondaries(secondaries)
+        .build();
   }
 
-  public @Nonnull PolarisPrincipal getPrincipal() {
-    return principal;
-  }
+  @Nonnull
+  PolarisPrincipal getPrincipal();
 
-  public @Nonnull PolarisAuthorizableOperation getOperation() {
-    return operation;
-  }
+  @Nonnull
+  PolarisAuthorizableOperation getOperation();
 
-  public @Nullable List<PolarisSecurable> getTargets() {
-    return targets;
-  }
+  @Nullable
+  List<PolarisSecurable> getTargets();
 
-  public @Nullable List<PolarisSecurable> getSecondaries() {
-    return secondaries;
-  }
+  @Nullable
+  List<PolarisSecurable> getSecondaries();
 
-  public boolean isInternalPrincipalScope() {
+  default boolean isInternalPrincipalScope() {
     return hasSecurableType(PolarisEntityType.PRINCIPAL)
         || hasPrivilegeSecurableType(PolarisEntityType.PRINCIPAL)
         || hasPrivilegeNameForPrincipalScope();
   }
 
-  public boolean isInternalPrincipalRoleScope() {
+  default boolean isInternalPrincipalRoleScope() {
     return hasSecurableType(PolarisEntityType.PRINCIPAL_ROLE)
         || hasPrivilegeSecurableType(PolarisEntityType.PRINCIPAL_ROLE)
         || hasPrivilegeNameForPrincipalRoleScope();
   }
 
-  public boolean isInternalCatalogRoleScope() {
+  default boolean isInternalCatalogRoleScope() {
     return hasSecurableType(PolarisEntityType.CATALOG_ROLE)
         || hasPrivilegeSecurableType(PolarisEntityType.CATALOG_ROLE)
         || hasPrivilegeNameForCatalogRoleScope();
   }
 
   private boolean hasSecurableType(PolarisEntityType... types) {
-    if (targets != null && containsType(targets, types)) {
+    if (getTargets() != null && containsType(getTargets(), types)) {
       return true;
     }
-    return secondaries != null && containsType(secondaries, types);
+    return getSecondaries() != null && containsType(getSecondaries(), types);
   }
 
   private boolean hasPrivilegeSecurableType(PolarisEntityType... types) {
-    return containsPrivilegeType(operation.getPrivilegesOnTarget(), types)
-        || containsPrivilegeType(operation.getPrivilegesOnSecondary(), types);
+    return containsPrivilegeType(getOperation().getPrivilegesOnTarget(), types)
+        || containsPrivilegeType(getOperation().getPrivilegesOnSecondary(), types);
   }
 
   private boolean hasPrivilegeNameForPrincipalScope() {
-    return containsPrivilegeNameForPrincipalScope(operation.getPrivilegesOnTarget())
-        || containsPrivilegeNameForPrincipalScope(operation.getPrivilegesOnSecondary());
+    return containsPrivilegeNameForPrincipalScope(getOperation().getPrivilegesOnTarget())
+        || containsPrivilegeNameForPrincipalScope(getOperation().getPrivilegesOnSecondary());
   }
 
   private boolean hasPrivilegeNameForPrincipalRoleScope() {
-    return containsPrivilegeNameForPrincipalRoleScope(operation.getPrivilegesOnTarget())
-        || containsPrivilegeNameForPrincipalRoleScope(operation.getPrivilegesOnSecondary());
+    return containsPrivilegeNameForPrincipalRoleScope(getOperation().getPrivilegesOnTarget())
+        || containsPrivilegeNameForPrincipalRoleScope(getOperation().getPrivilegesOnSecondary());
   }
 
   private boolean hasPrivilegeNameForCatalogRoleScope() {
-    return containsPrivilegeNameForCatalogRoleScope(operation.getPrivilegesOnTarget())
-        || containsPrivilegeNameForCatalogRoleScope(operation.getPrivilegesOnSecondary());
+    return containsPrivilegeNameForCatalogRoleScope(getOperation().getPrivilegesOnTarget())
+        || containsPrivilegeNameForCatalogRoleScope(getOperation().getPrivilegesOnSecondary());
   }
 
   private static boolean containsPrivilegeType(
@@ -138,43 +133,41 @@ public final class AuthorizationRequest {
   }
 
   private static boolean containsPrivilegeNameForPrincipalScope(Set<PolarisPrivilege> privileges) {
-    if (privileges == null || privileges.isEmpty()) {
-      return false;
-    }
-    for (PolarisPrivilege privilege : privileges) {
-      String name = privilege.name();
-      if (name.startsWith("PRINCIPAL_ROLE_") || name.startsWith("CATALOG_ROLE_")) {
-        continue;
-      }
-      if (name.startsWith("PRINCIPAL_")) {
-        return true;
-      }
-    }
-    return false;
+    return containsPrivilegeName(privileges, "PRINCIPAL_MANAGE_GRANTS")
+        || containsPrivilegeNameForPrefix(privileges, "PRINCIPAL_");
   }
 
   private static boolean containsPrivilegeNameForPrincipalRoleScope(
       Set<PolarisPrivilege> privileges) {
+    return containsPrivilegeName(privileges, "PRINCIPAL_ROLE_MANAGE_GRANTS")
+        || containsPrivilegeNameForPrefix(privileges, "PRINCIPAL_ROLE_");
+  }
+
+  private static boolean containsPrivilegeNameForCatalogRoleScope(
+      Set<PolarisPrivilege> privileges) {
+    return containsPrivilegeName(privileges, "CATALOG_ROLE_MANAGE_GRANTS")
+        || containsPrivilegeNameForPrefix(privileges, "CATALOG_ROLE_");
+  }
+
+  private static boolean containsPrivilegeNameForPrefix(
+      Set<PolarisPrivilege> privileges, String prefix) {
     if (privileges == null || privileges.isEmpty()) {
       return false;
     }
     for (PolarisPrivilege privilege : privileges) {
-      String name = privilege.name();
-      if (name.startsWith("PRINCIPAL_ROLE_")) {
+      if (privilege.name().startsWith(prefix)) {
         return true;
       }
     }
     return false;
   }
 
-  private static boolean containsPrivilegeNameForCatalogRoleScope(
-      Set<PolarisPrivilege> privileges) {
+  private static boolean containsPrivilegeName(Set<PolarisPrivilege> privileges, String name) {
     if (privileges == null || privileges.isEmpty()) {
       return false;
     }
     for (PolarisPrivilege privilege : privileges) {
-      String name = privilege.name();
-      if (name.startsWith("CATALOG_ROLE_")) {
+      if (privilege.name().equals(name)) {
         return true;
       }
     }
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
@@ -19,11 +19,6 @@
 package org.apache.polaris.core.auth;
 
 import jakarta.annotation.Nonnull;
-import jakarta.annotation.Nullable;
-import java.util.List;
-import java.util.Set;
-import org.apache.polaris.core.entity.PolarisBaseEntity;
-import org.apache.polaris.core.persistence.PolarisResolvedPathWrapper;
 
 /** Interface for invoking authorization checks. */
 public interface PolarisAuthorizer {
@@ -32,35 +27,9 @@ public interface PolarisAuthorizer {
    *
    * <p>Implementations may resolve or validate any inputs needed to make an authorization decision.
    */
-  void resolveAuthorizationInputs(@Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request);
+  void resolveAuthorizationInputs(
+      @Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request);
 
   /** Core authorization entry point for the new SPI. */
   void authorize(@Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request);
-
-  /**
-   * Backwards-compatible external API that throws on deny for legacy call sites.
-   *
-   * <p>Default implementation delegates to {@link #authorize(AuthorizationState,
-   * AuthorizationRequest)}.
-   */
-  default void authorizeOrThrow(
-      @Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request) {
-    authorize(ctx, request);
-  }
-
-  @Deprecated
-  void authorizeOrThrow(
-      @Nonnull PolarisPrincipal polarisPrincipal,
-      @Nonnull Set<PolarisBaseEntity> activatedEntities,
-      @Nonnull PolarisAuthorizableOperation authzOp,
-      @Nullable PolarisResolvedPathWrapper target,
-      @Nullable PolarisResolvedPathWrapper secondary);
-
-  @Deprecated
-  void authorizeOrThrow(
-      @Nonnull PolarisPrincipal polarisPrincipal,
-      @Nonnull Set<PolarisBaseEntity> activatedEntities,
-      @Nonnull PolarisAuthorizableOperation authzOp,
-      @Nullable List<PolarisResolvedPathWrapper> targets,
-      @Nullable List<PolarisResolvedPathWrapper> secondaries);
 }
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
@@ -138,6 +138,7 @@ import org.apache.polaris.core.config.RealmConfig;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
 import org.apache.polaris.core.entity.PolarisEntityConstants;
 import org.apache.polaris.core.entity.PolarisEntityCore;
+import org.apache.polaris.core.entity.PolarisEntitySubType;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisGrantRecord;
 import org.apache.polaris.core.entity.PolarisPrivilege;
@@ -755,23 +756,7 @@ public class PolarisAuthorizerImpl implements PolarisAuthorizer {
     return false;
   }
 
-  @Override
-  public void authorizeOrThrow(
-      @Nonnull PolarisPrincipal polarisPrincipal,
-      @Nonnull Set<PolarisBaseEntity> activatedEntities,
-      @Nonnull PolarisAuthorizableOperation authzOp,
-      @Nullable PolarisResolvedPathWrapper target,
-      @Nullable PolarisResolvedPathWrapper secondary) {
-    authorizeOrThrow(
-        polarisPrincipal,
-        activatedEntities,
-        authzOp,
-        target == null ? null : List.of(target),
-        secondary == null ? null : List.of(secondary));
-  }
-
-  @Override
-  public void authorizeOrThrow(
+  private void authorizeResolved(
       @Nonnull PolarisPrincipal polarisPrincipal,
       @Nonnull Set<PolarisBaseEntity> activatedEntities,
       @Nonnull PolarisAuthorizableOperation authzOp,
@@ -915,13 +900,15 @@ public class PolarisAuthorizerImpl implements PolarisAuthorizer {
   }
 
   @Override
-  public void resolveAuthorizationInputs(@Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request) {
+  public void resolveAuthorizationInputs(
+      @Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request) {
     PolarisResolutionManifest manifest = ctx.getResolutionManifest();
     if (manifest.hasResolution()) {
       return;
     }
 
-    // Phase 2: preserve existing RBAC ordering by resolving the manifest within resolveAuthorizationInputs.
+    // Phase 2: preserve existing RBAC ordering by resolving the manifest within
+    // resolveAuthorizationInputs.
     manifest.resolveSelections(
         EnumSet.of(
             Resolvable.CALLER_PRINCIPAL,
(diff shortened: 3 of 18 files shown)


runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -217,23 +255,22 @@ private static CatalogEntity getCatalogByName(
...
-    resolutionManifest.resolveAll();
-    PolarisResolvedPathWrapper rootContainerWrapper =
-        resolutionManifest.getResolvedRootContainerEntityAsPath();
-    authorizer.authorizeOrThrow(
-        polarisPrincipal,
-        resolutionManifest.getAllActivatedPrincipalRoleEntities(),
-        op,
-        rootContainerWrapper,
-        null /* secondary */);
+    AuthorizationCallContext authzContext = new AuthorizationCallContext(resolutionManifest);
```
dimas-b: Would it be possible to inject it from CDI? Ideally `resolutionManifest` should also be a request-scoped (auto-created) bean, but for a start we could add it to `AuthorizationCallContext` manually since `AuthorizationCallContext` is expected to be mutable anyway 🤔 WDYT?

sungwy: Hi @dimas-b - great suggestion.

Yes, I think that aligns well with the proposed strategy to define the Resolution semantics at request-scope instead, and build consistency guarantees on it. I think creating a request-scoped bean for `AuthorizationCallContext` would be safe, and we can evaluate if we can do the same for the resolution semantics after a dedicated design proposal on the Resolution/Persistence model

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
```diff
@@ -287,6 +308,10 @@ public enum PolarisAuthorizableOperation {
             : privilegesOnSecondary;
   }
 
+  public boolean isRbacAdminOperation() {
```
dimas-b: Could this be done in a separate PR?.. just for the same of clarity... introduce new SPI first, then adjust OPA for special treatment of the admin Ops?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -908,4 +913,75 @@ public boolean hasTransitivePrivilege(
...
+
+  @Override
+  public void preAuthorize(
+      @Nonnull AuthorizationCallContext ctx, @Nonnull AuthorizationRequest request) {
+    PolarisResolutionManifest manifest = ctx.getResolutionManifest();
+    if (manifest.hasResolution()) {
+      return;
+    }
+
+    // Phase 2: preserve existing RBAC ordering by resolving the manifest within preAuthorize.
```
dimas-b: nit: "existing" will be "old" after merge 😉

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -908,4 +913,75 @@ public boolean hasTransitivePrivilege(
...
     return false;
   }
+
+  @Override
+  public void preAuthorize(
+      @Nonnull AuthorizationCallContext ctx, @Nonnull AuthorizationRequest request) {
+    PolarisResolutionManifest manifest = ctx.getResolutionManifest();
+    if (manifest.hasResolution()) {
+      return;
+    }
```
dimas-b, line 908: Would it make sense to fold this into `resolveSelections()`? Is there a use case for re-resolving multiple times with different selections?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -27,14 +27,36 @@
...
+  @Deprecated
   void authorizeOrThrow(
       @Nonnull PolarisPrincipal polarisPrincipal,
       @Nonnull Set<PolarisBaseEntity> activatedEntities,
       @Nonnull PolarisAuthorizableOperation authzOp,
       @Nullable PolarisResolvedPathWrapper target,
       @Nullable PolarisResolvedPathWrapper secondary);
 
+  @Deprecated
   void authorizeOrThrow(
```
dimas-b: What's the benefit of keeping this uncalled method? I believe downstream implementation have to adjust anyway, so dropping dead methods would be a trivial part of that adjustment 🤔

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -27,14 +27,36 @@
 
 /** Interface for invoking authorization checks. */
 public interface PolarisAuthorizer {
+  /**
+   * Pre-authorization hook for resolving authorizer-specific inputs.
+   *
+   * <p>Implementations may resolve or validate any inputs needed to make an authorization decision.
+   */
+  void preAuthorize(@Nonnull AuthorizationCallContext ctx, @Nonnull AuthorizationRequest request);
```
dimas-b: I'm not sure I see the rationale for having these two distinct parameters? 🤔 why not have just one parameter with the union of data?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -27,14 +27,36 @@
...
 
+  /** Core authorization entry point for the new SPI. */
+  void authorize(@Nonnull AuthorizationCallContext ctx, @Nonnull AuthorizationRequest request);
```
dimas-b: Are `ctx` and `request` expected to be the same in calls to `preAuthorize` and `authorize`? If so, how about this pattern?

```
AuthorizationHandler h = PolarisAuthorizer.startRequest(AuthorizationCallContext ctx, ...)
h.preAuthoriz();
h.authorize();
```

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
sungwy: Yes, actually my intent with this PR is to just use it as a POC and an aid to reviewing the RFC.

I'd like to break up the review into multiple PRs as I've noted in the [Implementation Plan section of the RFC](https://docs.google.com/document/d/1OaiQG_C4-yUe0ihaDBxtw_mEcOOzUBnWPazzVbjQi5U/edit?tab=t.0#heading=h.k4t1nkd1so5g). I think it'll make it easier to review and introduce the changes

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: That sounds like a great idea @dimas-b

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
sungwy: > Would it make sense to fold this into resolveSelections()

I think that makes sense, and there are two possible implementation approaches.
1. Introduce an idempotent resolution mechanism for each item, allowing the caller to iteratively resolve additional or different selections, as you suggested.
2. If the selection set is unchanged, return cached resolution results without re-resolving; if a different selection set is requested, fail fast with an error.

Option (2) is safe and preserves semantics similar to the current behavior, where only a single resolveAll() is allowed.

In my view, option (1) is only safe if one of the following holds:
- resolving different selections across different snapshots of the persistence layer is acceptable, or
- the resolver or persistence layer already operates over a consistent snapshot.

I’d need a deeper understanding of the persistence model and its consistency guarantees before taking a strong position on option (1)...

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: I wanted to keep them separate before `AuthorizationRequest` is purely about the authorization intent that matches `subject`, `action` and `resource`, whereas `AuthorizationState` is a request scoped state that can be mutated throughout the handling of the request... but I'm definitely open for more discussion!

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: actually, on second thought - I wonder if this would be an over-optimization, before we know how the rest of the new methods would look like in the Authorizer SPI. There's a few follow up discussions around filtering, fetching obligations and storage configs which I think we should factor in before deciding if this handler based approach would make sense...

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Set;
+import org.apache.polaris.core.entity.PolarisEntityType;
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+/**
+ * Authorization request inputs for pre-authorization and core authorization.
+ *
+ * <p>This wrapper keeps authorization inputs together while preserving legacy semantics.
+ */
+public final class AuthorizationRequest {
```
adutra: nit: now that polaris-core is on Java 17, this class, as well as `PolarisSecurable`, could be records. Alternatively, they could be annotated with `@PolarisImmutable`.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -170,6 +176,7 @@ public PolarisAdminService(
     this.userSecretsManager = userSecretsManager;
     this.serviceIdentityProvider = serviceIdentityProvider;
     this.reservedProperties = reservedProperties;
+    this.authorizationState = authorizationState;
```
adutra, line 179: Wondering: do we need to make this object a CDI request-scoped bean? I think it could be instantiated by the handler manually, which would have the advantage that we wouldn't need to expose setters in `AuthorizationState`.

That is, instead of:

```java
    AuthorizationState authzContext = authorizationState();
    authzContext.setResolutionManifest(resolutionManifest);
```

The handler would do:

```java
    AuthorizationState authzContext = new RequestAuthorizationState(resolutionManifest);
```

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -27,14 +27,36 @@
 
 /** Interface for invoking authorization checks. */
 public interface PolarisAuthorizer {
+  /**
+   * Pre-authorization hook for resolving authorizer-specific inputs.
+   *
+   * <p>Implementations may resolve or validate any inputs needed to make an authorization decision.
+   */
+  void preAuthorize(@Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request);
```
adutra: Just to confirm: are the new methods still expected to throw on deny?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -908,4 +913,73 @@ public boolean hasTransitivePrivilege(
...
+            Resolvable.REFERENCE_CATALOG,
+            Resolvable.REQUESTED_PATHS,
+            Resolvable.TOP_LEVEL_ENTITIES));
+  }
+
+  @Override
+  public void authorize(@Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request) {
+    PolarisResolutionManifest manifest = ctx.getResolutionManifest();
+    Set<PolarisBaseEntity> activatedEntities =
+        manifest == null ? Set.of() : manifest.getAllActivatedCatalogRoleAndPrincipalRoles();
```
adutra: nit: `manifest` cannot be null here.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -27,14 +27,36 @@
...
+  /** Core authorization entry point for the new SPI. */
+  void authorize(@Nonnull AuthorizationState ctx, @Nonnull AuthorizationRequest request);
+
+  /**
+   * Backwards-compatible external API that throws on deny for legacy call sites.
+   *
+   * <p>Default implementation delegates to {@link #authorize(AuthorizationState,
+   * AuthorizationRequest)}.
+   */
+  default void authorizeOrThrow(
```
adutra: I'm confused: I don't see this method used anywhere, and (contrary to what the javadocs say) it's a new method, not a legacy one.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
adutra: I agree: this one seems used only in `OpaPolarisAuthorizerTest`.

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -143,6 +182,25 @@ public ResolverStatus resolveAll() {
     return primaryResolverStatus;
   }
 
+  /**
+   * Resolves explicitly requested components.
+   *
+   * <p>Phase 1 behavior delegates to {@link #resolveAll()} to preserve existing semantics until
+   * resolver selection is fully implemented.
+   */
+  public ResolverStatus resolveSelections(Set<Resolvable> selections) {
```
adutra, line 169: nit: would `resolvePartially` better capture the opposition to `resolveAll`?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -217,23 +258,23 @@ private static CatalogEntity getCatalogByName(
...
-    authorizer.authorizeOrThrow(
-        polarisPrincipal,
-        resolutionManifest.getAllActivatedPrincipalRoleEntities(),
-        op,
-        rootContainerWrapper,
-        null /* secondary */);
+    AuthorizationState authzContext = authorizationState;
+    authzContext.setResolutionManifest(resolutionManifest);
+    authorizer.preAuthorize(authzContext, newAuthorizationRequest(op));
+    PolarisSecurable rootSecurable =
```
adutra, line 272: nit: I think the root securable could be a public constant in `PolarisSecurable`:

```java
  public static final PolarisSecurable ROOT =
      new PolarisSecurable(
          PolarisEntityType.ROOT, List.of(PolarisEntityConstants.getRootContainerName()));
```

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -131,6 +147,29 @@ public void addPassthroughPath(ResolverPath path, Object key) {
     passthroughPaths.put(key, path);
   }
 
+  /** Adds an alias key for a previously added path key. */
+  public void addPathAlias(Object existingKey, Object aliasKey) {
```
adutra: This method looks a bit too low-level, could we avoid introducing it if we change `addPath` to take several keys? E.g.

```java
 public void addPath(ResolverPath path, Object... key) {
    primaryResolver.addPath(path);
    for (Object k : key) {
      pathLookup.put(k, currentPathIndex);
    }
    addedPaths.add(path);
    ++currentPathIndex;
  }
```

Then at call sites, instead of:

```java
    resolutionManifest.addPath(
        new ResolverPath(parentNamespaceSecurable.getNameParts(), PolarisEntityType.NAMESPACE),
        parentNamespaceSecurable);
    resolutionManifest.addPathAlias(parentNamespaceSecurable, parentNamespace);
```

We'd have:

```java
    resolutionManifest.addPath(
        new ResolverPath(parentNamespaceSecurable.getNameParts(), PolarisEntityType.NAMESPACE),
        parentNamespaceSecurable,
        parentNamespace);
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
```diff
@@ -45,13 +47,19 @@ public class CatalogUtils {
...
   public static PolarisResolvedPathWrapper findResolvedStorageEntity(
       PolarisResolutionManifestCatalogView resolvedEntityView, TableIdentifier tableIdentifier) {
+    PolarisSecurable tableSecurable =
+        new PolarisSecurable(
+            PolarisEntityType.TABLE_LIKE,
+            PolarisCatalogHelpers.tableIdentifierToList(tableIdentifier));
     PolarisResolvedPathWrapper resolvedTableEntities =
         resolvedEntityView.getResolvedPath(
-            tableIdentifier, PolarisEntityType.TABLE_LIKE, PolarisEntitySubType.ICEBERG_TABLE);
+            tableSecurable, PolarisEntityType.TABLE_LIKE, PolarisEntitySubType.ICEBERG_TABLE);
```
adutra: Shouldn't the path have been registered also under the alias key `tableIdentifier`? 

I'm wondering if this change (and many other similar ones where a path is looked up by a `PolarisSecurable` instead of some other key) is necessary. 

Or is it because you plan to deprecate/remove lookup keys that are not `PolarisSecurable` instances later on?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1315,12 +1318,30 @@ public void close() throws Exception {
...
-    if (!resolverStatus.getStatus().equals(ResolverStatus.StatusEnum.SUCCESS)) {
+    // 'catalogName' is taken from the REST request's 'warehouse' query parameter.
+    // 'warehouse' as an output will be treated by the client as a default catalog
+    //   storage base location.
+    // 'prefix' as an output is the REST subpath that routes to the catalog
+    //   resource, which may be URL-escaped catalogName or potentially a different
+    //   unique identifier for the catalog being accessed.
+    if (catalogName() == null) {
+      throw new BadRequestException("Please specify a warehouse");
+    }
```
adutra: This has been moved to `IcebergCatalogAdapter`. Here, `catalogName()` is guaranteed to never be null.
```suggestion
```

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
```diff
@@ -0,0 +1,29 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.persistence.resolver;
+
+/** Explicit resolution components for {@link PolarisResolutionManifest#resolveSelections}. */
+public enum Resolvable {
```
adutra, line 22: Would you mind adding some javadocs explaining the scope of each constant? It might be just me 😅 but I'm not immediately seeing what `TOP_LEVEL_ENTITIES` is referring to: according to our docs, only catalogs are top-level entities. Polaris code also considers principals and roles as top-level entities for authorization purposes. But these entities are all covered already by `CALLER_PRINCIPAL`,
`CALLER_PRINCIPAL_ROLES` and `REFERENCE_CATALOG`. Is `TOP_LEVEL_ENTITIES` reserved to "optional" top-level entities?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -179,12 +237,17 @@ public PolarisResolvedPathWrapper getResolvedPath(
...
-    diagnostics.check(
-        passthroughPaths.containsKey(key),
-        "invalid_key_for_passthrough_resolved_path",
-        "key={} passthroughPaths={}",
-        key,
-        passthroughPaths);
+    if (!passthroughPaths.containsKey(key)) {
+      if (pathLookup.containsKey(key)) {
+        return getResolvedPath(key);
+      }
```
adutra: This if block looks like a change in the logic, can you explain why it is needed?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
dimas-b: Sounds like a good plan! Sorry about the confusion on my part :sweat_smile: ... too many reviews :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
dimas-b: I'm fine with option 2.

Consistency guarantees in Persistence is a very hard topic because Polaris supports pluggable (custom) Persistence. I wish we could have a precise Persistence contract defined in code or docs, but I do not think we're completely strict and well-defined on that side in the current state of the repository :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
dimas-b: Good point - thx for the clarification :+1:

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
dimas-b: No worries. We can defer the non-essential / beatification changes til later. I hope people are ok with several "breaking" changes in this SPI :sweat_smile:

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
sungwy: That's a possibility, and we could make it immutable.

sungwy: Based on what we discussed during the Polaris Community Sprint, it looks like there's an idea to add some additional states are intended to be added to the `AuthorizationState`... it's unfortunately a bit unclear to me what those will be. 

cc @singhpk234

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: Hi @adutra - actually @dimas-b brought up this point as well. We could return a boolean instead, but that would leave it to all of the Handler call sites to raise an exception instead.

For `preAuthorize` (or `resolveAuthorizationInput`  which is a naming option we are discussing in [this RFC comment](https://docs.google.com/document/d/1OaiQG_C4-yUe0ihaDBxtw_mEcOOzUBnWPazzVbjQi5U/edit?disco=AAABzyyDy_U)), I think it makes sense to raise an exception if we aren't able to resolve the inputs ahead of time.

What do you think?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
sungwy: That's a great idea @adutra - thank you!

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
sungwy: I introduced this due to the complexity in introducing `PolarisSecurable` as a key to the `PolarisResolutionManifest` cache. I'm realizing that there's a lot of complexity in doing so, because unless all call sites are updated to use `PolarisSecurable`, we'd expect to have a cache-miss even if we are referring to the same entity (e.g. `TableIdentifier` vs `PolarisSecurable` wrapper on a `TableIdentifier`).

I'm thinking on my foot now, and I think keeping `PolarisSecurable` as a thin wrapper to enforce intent-based authorization input and just leaving it there, and then keeping the `PolarisResolutionManifest` as-is, and making use of the existing key types may be a better approach to avoid this hassle of migrating to the new APIs.

This fallback was here to try to mitigate these cache misses, but I think that would mean that `passthroughPaths` may yield static and stale result from the `pathLookup`, which is undesirable.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
sungwy: +10000

I agree @dimas-b . I think we need to define the persistence contract asap :)

Otherwise it's going to be hard to expect consistent behaviors from all of the call sites in the longer term

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: Sounds good 👍

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: This is an artifact an intermediate state - let me clean this up

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
sungwy: > is it because you plan to deprecate/remove lookup keys that are not PolarisSecurable instances later on?

Hi @adutra - yes that was my original intention. I thought it would be a good idea to make use of `PolarisSecurable` instead of the existing heterogeneous object types. However, based on the concerns you raised around introducing changes to compensate for the complexity in potentially introducing an alias that could result in a cache-miss, I think removing `PolarisSecurable` from `PolarisResolutionManifest` calls would be a better approach for now.

https://github.com/apache/polaris/pull/3681/changes#r2796290376

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
sungwy: Yes, I meant to differentiate this from the the caller/reference catalog resolution since this is to resolve the target top level entities that are added via `addTopLevelName`

sungwy: I'll name this `REQUESTED_TOP_LEVEL_ENTITIES` to mirror `REQUESTED_PATHS`, and add some doc comments here

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
dimas-b: My point about `boolean` really makes a difference only for filtering API outputs (lists), I think.

Throwing on failing authorization checks in the main use case (once per API call) is probably ok.

dimas-b: That said, how about this pattern of defining the SPI to return a `boolean` or some object with explicit failure into and adding convenience methods for simple call sites?

```
AuthDecision preAuthorize(...);

default void preAuthorizeOrThrow(...)  {
   AuthDecition d = preAuthorize(...);
    if (!d.isAllowed()) {
      throw new ForbiddenException(d.getMessage()...);
    }
}
```

This way the exception type and API-level message format is controlled by Polaris (core) code, while AuthZ implementations only need to provide a yes/no output and (optionally) an explanation.

adutra: I fell like we'll need `AuthDecision` sooner or later.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
adutra: If it needs to stay mutable, that's another reason to consider if it needs to be a CDI bean.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
adutra: I actually think that having look up keys be strictly `PolarisSecurable` instances is a good thing.

sungwy: :) How about we focus the refactoring on first migrating to the new SPI, and then follow up with a PR to replace  resolution keys with `PolarisSecurable` instead?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
dimas-b: +1 to CDI beans for auth context/state.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
dimas-b: Iterative approach sounds good to me.

sungwy:
Authorizer refactoring is being tracked through https://github.com/apache/polaris/issues/377

dimas-b:
@sungwy : I guess you meant to reference #3779 :)

sungwy:
Yes @dimas-b - that was a bad copy paste :)

Issue #2896, opened 2025-10-24
Support `overwrite` option in RegisterTable


HonahX:
### Is your feature request related to a problem? Please describe.

From a slack [discussion](https://apache-polaris.slack.com/archives/C084QSKD6S2/p1761327874167969?thread_ts=1761326421.638729&cid=C084QSKD6S2)

IRC spec added an `overwrite` option for [registerTable](https://github.com/apache/iceberg/blob/6b86a75d3a0f0123d1a19f2db0b266bef05c2a5b/open-api/rest-catalog-open-api.yaml#L3543-L3543)

This will enable use case where updating the entire metadata is needed.

It would be good for polaris to support this to remain compliant with IRC spec

CodingBangboo:
I'm interested in adding this new feature. May I work on it? Thanks a lot.

HonahX:
@CodingBangboo Assign to you! Thanks for taking this!

PR #3682, opened 2026-02-06
[DRAFT] feat: add RegisterTable overwrite support; preserve correct auth for overwrites

sririshindra:
Fixes https://github.com/apache/polaris/issues/2896

Add support for the new overwrite boolean on RegisterTableRequest (default: false) so clients can register a metadata location that replaces an existing table pointer. Implement register-table overwrite semantics:
If overwrite=false (default): preserve existing behavior — attempting to register a table that already exists returns a conflict/error. If overwrite=true:
If the table does not exist: create it (normal register). If the table exists: update the table's metadata-location to the provided value (do not throw AlreadyExists). Ensure safety/backward-compatibility:
overwrite defaults to false so existing clients are unaffected. Validate provided metadata location where possible before committing to avoid corrupting catalog state.

Details / rationale:
The change implements the REST Catalog spec extension that adds an overwrite flag to RegisterTableRequest. This enables clients to atomically point an existing table identifier at a new metadata file (useful for moving or restoring table metadata). Because overwrite changes the set of operations allowed on an existing table, we had to be careful with authorization and entity resolution: When performing an overwrite against an existing table, the handler enforces the UPDATE_TABLE privilege (mapped to TABLE_WRITE_PROPERTIES) rather than REGISTER_TABLE. This prevents callers who only have create permissions from silently replacing another principal's table pointer. We fixed a subtle distinction where lack of read privileges could be mistaken for a non-existent table. The handler now distinguishes "truly not found" from "exists but unreadable" and applies the correct required privilege accordingly. Resolution manifest and passthrough-path population were adjusted so downstream overwrite logic can locate the namespace and table entries reliably. Files / components touched (high level):

Tests added/updated:
- DTO unit tests: overwrite deserialization for `RegisterTableRequest` (true/false/missing).
- Integration / behavior tests (integration-tests module):
  - Register new table with `overwrite=false` → success.
  - Register existing table with `overwrite=false` → conflict / exception as before.
  - Register existing table with `overwrite=true` → success and the table's metadata-location is updated atomically.
- Authorization tests: updated/added unit tests to assert `UPDATE_TABLE` (TABLE_WRITE_PROPERTIES) is enforced for overwrite against existing tables.
- Unit tests: added `CatalogHandlerUtilsTest` to cover handler helper logic.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
@@ -64,6 +64,7 @@ import org.apache.iceberg.catalog.SupportsNamespaces;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.catalog.ViewCatalog;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
+import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
@@ -348,7 +349,21 @@ public class CatalogHandlerUtils {
     request.validate();
 
     TableIdentifier identifier = TableIdentifier.of(namespace, request.name());
-    Table table = catalog.registerTable(identifier, request.metadataLocation());
+    // Determine whether the client requested overwrite semantics. For catalogs that
+    // understand overwrite (our {@link IcebergCatalog}) we invoke the overwrite-capable
+    // registration path. For other catalog implementations, explicitly reject overwrite
+    // so callers don't accidentally replace an existing table pointer.
+    boolean shouldOverwrite = RegisterTableRequestContext.getOverwrite();
+    Table table;
+    if (shouldOverwrite && catalog instanceof IcebergCatalog polarisCatalog) {
+      table = polarisCatalog.registerTable(identifier, request.metadataLocation(), true);
+    } else if (shouldOverwrite && catalog.tableExists(identifier)) {
+      throw new BadRequestException(
+          "Register table overwrite is not supported for catalog type: %s",
+          catalog.getClass().getName());
+    } else {
+      table = catalog.registerTable(identifier, request.metadataLocation());
+    }
     if (table instanceof BaseTable baseTable) {
       return LoadTableResponse.builder()
           .withTableMetadata(baseTable.operations().current())
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -287,6 +287,26 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
   @Override
   public Table registerTable(TableIdentifier identifier, String metadataFileLocation) {
+    return registerTable(identifier, metadataFileLocation, false);
+  }
+
+  /**
+   * Register a table with optional overwrite semantics.
+   *
+   * <p>When {@code overwrite} is false (the default) this behaves like a normal register and will
+   * fail if the table already exists. When {@code overwrite} is true and the named table already
+   * exists, this method updates the table's stored metadata-location to point at the provided
+   * metadata file. The overwrite path performs additional validation to ensure the supplied
+   * metadata file and its location are consistent with the table's resolved storage configuration.
+   *
+   * @param identifier the table identifier
+   * @param metadataFileLocation the metadata file location
+   * @param overwrite if true, update existing table metadata; if false, throw exception if table
+   *     exists
+   * @return the registered table
+   */
+  public Table registerTable(
+      TableIdentifier identifier, String metadataFileLocation, boolean overwrite) {
     Preconditions.checkArgument(
         identifier != null && isValidIdentifier(identifier), "Invalid identifier: %s", identifier);
     Preconditions.checkArgument(
@@ -299,19 +319,26 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
         "Invalid metadata file location; metadata file location must be absolute and contain a '/': %s",
         metadataFileLocation);
 
-    // Throw an exception if this table already exists in the catalog.
-    if (tableExists(identifier)) {
+    boolean tableExists = tableExists(identifier);
+    if (!overwrite && tableExists) {
       throw new AlreadyExistsException("Table already exists: %s", identifier);
     }
 
     String locationDir = metadataFileLocation.substring(0, lastSlashIndex);
+    if (!tableExists) {
+      return registerNewTable(identifier, metadataFileLocation, locationDir);
+    }
 
+    return overwriteRegisteredTable(identifier, metadataFileLocation, locationDir);
+  }
+
+  private Table registerNewTable(
+      TableIdentifier identifier, String metadataFileLocation, String locationDir) {
     TableOperations ops = newTableOps(identifier);
 
     PolarisResolvedPathWrapper resolvedParent =
         resolvedEntityView.getResolvedPath(identifier.namespace());
     if (resolvedParent == null) {
-      // Illegal state because the namespace should've already been in the static resolution set.
       throw new IllegalStateException(
           String.format("Failed to fetch resolved parent for TableIdentifier '%s'", identifier));
     }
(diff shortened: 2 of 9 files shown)


adutra:
Hi @sririshindra thanks for opening this PR! 

Unfortunately I don't think this is the right way to bring support for overwrite in register table requests. In fact, support for this parameter seems to be missing from the Iceberg artifact. So, I went ahead and created a PR to bring support for it in Iceberg: https://github.com/apache/iceberg/pull/15248.

If/when that PR is accepted, then we will be able to adapt Polaris properly to the new parameter.

runtime/service/src/main/java/org/apache/polaris/service/config/Serializers.java
```diff
@@ -58,6 +61,7 @@ public static void registerSerializers(ObjectMapper mapper) {
         GrantCatalogRoleRequest.class, new GrantCatalogRoleRequestDeserializer());
     module.addDeserializer(AddGrantRequest.class, new AddGrantRequestDeserializer());
     module.addDeserializer(RevokeGrantRequest.class, new RevokeGrantRequestDeserializer());
+    module.addDeserializer(RegisterTableRequest.class, new RegisterTableRequestDeserializer());
```
adutra, line 64: This is going to conflict with `org.apache.iceberg.rest.RESTSerializers.RegisterTableRequestDeserializer`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
```diff
@@ -348,7 +349,21 @@ public LoadTableResponse registerTable(
     request.validate();
 
     TableIdentifier identifier = TableIdentifier.of(namespace, request.name());
-    Table table = catalog.registerTable(identifier, request.metadataLocation());
+    // Determine whether the client requested overwrite semantics. For catalogs that
```
adutra, line 352: This class was copied from Iceberg. I'm not sure it's a good idea to modify it in a divergent way. I would place the Polaris-specific logic in `IcebergCatalogHandler` or `IcebergCatalog` directly.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +357,80 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+
+    // Build updated entity from parsed metadata and persist the update.
+    Map<String, String> storedProperties = buildTableMetadataPropertiesMap(metadata);
+    IcebergTableLikeEntity updatedEntity =
+        new IcebergTableLikeEntity.Builder(existingEntity)
+            .setInternalProperties(storedProperties)
+            .setMetadataLocation(metadataFileLocation)
+            .build();
+
+    updateTableLike(identifier, updatedEntity);
```
adutra, line 426: Is this switching the table UUID to the new UUID, or keeping the old UUID? From this ML discussions, I think it should be the new UUID:

https://lists.apache.org/thread/b5k7vdng904zr3n3q8wv83y8l30rnd4c
https://lists.apache.org/thread/k3595bttvohb6c3ms36o16gppdfllqmp

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -634,11 +640,110 @@ public LoadTableResponse createTableStaged(
...
+  }
+
+  private void authorizeUpdateTableOverwriteOrThrow(TableIdentifier identifier) {
+    authorizeBasicTableLikeOperationsOrThrow(
+        EnumSet.of(PolarisAuthorizableOperation.UPDATE_TABLE),
+        PolarisEntitySubType.ICEBERG_TABLE,
+        identifier);
 
-    return catalogHandlerUtils.registerTable(baseCatalog, namespace, request);
+    if (!(authorizer instanceof PolarisAuthorizerImpl authorizerImpl)) {
```
adutra, line 698: This is odd, the exact authorizer being used shouldn't matter. The rest of this method feels too low-level for this class.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/RegisterTableRequestContext.java
```diff
@@ -0,0 +1,53 @@
...
+ * Thread-local context that carries additional parameters for a {@code RegisterTableRequest} which
+ * are not present on the generated Iceberg request type.
+ *
+ * <p>The current implementation only records the caller's intent to "overwrite" an existing table.
+ * Using a ThreadLocal keeps the flag local to the request lifecycle and avoids changing the
+ * generated DTOs. Callers must clear the flag (typically in a finally block) to avoid leaking state
+ * across requests.
+ */
+public class RegisterTableRequestContext {
+  private static final ThreadLocal<Boolean> SHOULD_OVERWRITE = ThreadLocal.withInitial(() -> false);
```
adutra, line 31: ThreadLocals are not recommended with Quarkus because there is no guarantee that a request will be handled by a single thread for the whole request scope.

sririshindra:
Thanks @adutra for your review.
I noticed that this this parameter is missing from the Iceberg artifact, so I implemented this PR in such a way that the polaris still supports it even if it was missing from Iceberg. But, I now realize that is the wrong approach to take. Thanks for opening the PR in the Iceberg repository adding support for overwrite. Once your PR get's merged, I will update this PR based on your review comments. 

I have an additional question. My understanding is that once your PR in the iceberg repository gets merged and released as part of the next Iceberg update. We would have to update https://github.com/apache/polaris/blob/main/gradle/libs.versions.toml#L23 to pick up that artifact. Correct?   
So we can only merge this feature in polaris after the next Iceberg release and after polaris picks up the updated artifact.. Is my understanding correct?

adutra:
> I have an additional question. My understanding is that once your PR in the iceberg repository gets merged and released as part of the next Iceberg update. We would have to update https://github.com/apache/polaris/blob/main/gradle/libs.versions.toml#L23 to pick up that artifact. Correct?   

You are correct in saying that this feature will have to wait for Iceberg 1.11.

But development can happen ahead of schedule! I just created the `feature/iceberg-1.11` branch. 

Once the overwrite parameter is merged in Iceberg you would be able to rebase your current work on top of that branch and open a PR against it. 

Would that work for you?

sririshindra:
> > I have an additional question. My understanding is that once your PR in the iceberg repository gets merged and released as part of the next Iceberg update. We would have to update https://github.com/apache/polaris/blob/main/gradle/libs.versions.toml#L23 to pick up that artifact. Correct?
> 
> You are correct in saying that this feature will have to wait for Iceberg 1.11.
> 
> But development can happen ahead of schedule! I just created the `feature/iceberg-1.11` branch.
> 
> Once the overwrite parameter is merged in Iceberg you would be able to rebase your current work on top of that branch and open a PR against it.
> 
> Would that work for you?

Yes, I will update my PR in the meantime. Thank you!

adutra:
The Iceberg PR has been merged FYI: https://github.com/apache/iceberg/pull/15248

We should be able to work on this feature against the `feature/iceberg-1.11` branch as soon as we have the next Iceberg nightly build.

sririshindra:
> The Iceberg PR has been merged FYI: [apache/iceberg#15248](https://github.com/apache/iceberg/pull/15248)
> 
> We should be able to work on this feature against the `feature/iceberg-1.11` branch as soon as we have the next Iceberg nightly build.

Thanks @adutra I will update the Pr shortly.

sririshindra:
@adutra , I have created https://github.com/apache/polaris/pull/3719/changes on the feature branch. I built Iceberg locally to develop it. Once Iceberg nightly builds become available the test in that PR should pass. I think I have addressed all your review comments, but if I miss anything please let me know.

runtime/service/src/main/java/org/apache/polaris/service/config/Serializers.java
sririshindra: Removed in https://github.com/apache/polaris/pull/3719

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
sririshindra: Removed in https://github.com/apache/polaris/pull/3719

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: Yes, a new UUID will be created. Added to the test to verify this in https://github.com/apache/polaris/pull/3719

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sririshindra: Fixed in the latest PR https://github.com/apache/polaris/pull/3719

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/RegisterTableRequestContext.java
sririshindra: Removed this entirely in https://github.com/apache/polaris/pull/3719

PR #3684, opened 2026-02-06
[Iceberg 1.11] Temporarily disable incompatible tests

adutra:
`CatalogTests.testLoadTableWithMissingMetadataFile` is not compatible with REST catalogs, see https://github.com/apache/iceberg/pull/15192. It should never be re-enabled.

The register view tests are failing for now since the register view endpoint is not implemented. These tests should be re-enabled later.

This PR is the minimum we must do to get a successful build in the feature branch `feature/iceberg1.11`.

## Checklist

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisApplicationIntegrationTest.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisApplicationIntegrationTest.java
@@ -281,8 +281,7 @@ public class PolarisApplicationIntegrationTest {
   public void testConfigureCatalogCaseSensitive() {
     assertThatThrownBy(() -> newSessionCatalog("TESTCONFIGURECATALOGCASESENSITIVE"))
         .isInstanceOf(RESTException.class)
-        .hasMessage(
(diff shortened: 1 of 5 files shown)


snazy:
I wonder, shall we have CI on `feature/*` branches and PRs against those?
Not thinking of protecting these branches (think: not require "green" CI), but having CI results feels nice.

adutra:
> I wonder, shall we have CI on `feature/*` branches and PRs against those? Not thinking of protecting these branches (think: not require "green" CI), but having CI results feels nice.

Agreed, we should at least have some CI results.

jbonofre:
I makes sense to have CI on features branches.

By the way, we plan to "extract" atomic PRs based on this branch right ?

jbonofre (approved):

snazy:
> I makes sense to have CI on features branches.
> 
> By the way, we plan to "extract" atomic PRs based on this branch right ?

I think we can whatever is appropriate:
* Squash/rebase the feature branch when needed
* Split into reasonable commits
* Extract PRs to `main`

I don't think we should basically allow "everything that's reasonable" on feature branches. It's rather "experimental work" for now.

PR #3689, opened 2026-02-06
[Iceberg 1.11] View registration

adutra:
This PR implements the new `registerView` endpoint.

## Checklist


adutra:
> Wonder whether we should focus on the necessary changes in the feature branch. Register-view is a new functionality, wonder whether we should do this change directly against main once the feature branch is merged. WDYT?

I think it would be nicer, if possible, to start working ahead of phase. 

Register view is just one of the items we need to work on. There is also "register table with overwrite", cf. #3682. And then, of course, there is remote signing, if the spec changes get merged before 1.11 is cut.

But I get the problem with merging new features into the `feature/iceberg-1.11` branch: the individual commits would end up squashed when merged to `main`, which is not great.

Wdyt about enabling "Rebase & Merge" so that we could keep working on the feature branch, then rebase the final result onto `main`, keeping each new feature in a separate commit?

adutra:
The other option is to keep working against the feature branch, but refraining from merging new features into it, and wait instead until `main` is ready.

snazy:
I think we can keep those things as separate commits (rebase/squash as needed) on the feature branch and then "peal out" individual PRs from the feature branch. That makes reviewing the "final PRs" easier.

snazy:
I think we can keep those things as separate commits (rebase/squash as needed) on the feature branch and then "peal out" individual PRs from the feature branch. That makes reviewing the "final PRs" easier.

snazy:
I think we can keep those things as separate commits (rebase/squash as needed) on the feature branch and then "peal out" individual PRs from the feature branch. That makes reviewing the "final PRs" easier.

adutra:
This is now ready for review.

dimas-b:
I might have missed the discussion about the plan for merging this feature branch to `main`, but it makes sense to me to merge individual PRs into the feature branch as work progresses.

If we want to redo those PRs against `main` later, that would be fine too from my POV.

snazy (commented):
It seems the change is missing a "location overlap" check to prevent that any table/view share the the same "directory tree", which would lead to data security issues (vending credentials granting access to other tables/views).

adutra:
> It seems the change is missing a "location overlap" check to prevent that any table/view share the the same "directory tree", which would lead to data security issues (vending credentials granting access to other tables/views).

This validation is already done in `BasePolarisViewOperations.doCommit()`.

adutra:
Rebased to fix conflicts.

adutra:
I think this PR has waited long enough. If there are no concerns, I am going to merge this on Monday.

PR #3693, opened 2026-02-06
Correctly Propagate Errors from Drop Entities

travis-bowen:
During drop commands, Apache Polaris would make assumptions based on the entity type as to why the entity could not be dropped which could lead to incorrect user error messaging and handling.

Here I tackle updating error handling for DropTable, DropNamespace, and DropView.

The Catalog API javadocs by Iceberg indicate false should indicate [TableNotExists] (and simila for View) (https://github.com/apache/iceberg/blob/a8ece055ba93adc0c046db29b6b7b5edaf35d4da/api/src/main/java/org/apache/iceberg/catalog/Catalog.java#L292).

For Namespaces [it's less clear]((https://github.com/apache/iceberg/blob/a8ece055ba93adc0c046db29b6b7b5edaf35d4da/api/src/main/java/org/apache/iceberg/catalog/SupportsNamespaces.java#L120) ), but I assume the same holds and we treat it to follow the same logic that false = Not Found in CatalogHandlerUtils.

**Before**
Any non success in Polaris's IcebergCatalog impl would return false.
Polaris's [CatalogHandlerUtils](https://github.com/apache/polaris/blob/6a516e0976f549ec4326f1de6b1324520c2a1f38/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java#L364) would interpret this false to mean the entity did not exist and would throw the appropriate does not exist error

However, the reason that it returned false, may not truly be that it didn't exist.

**After**
The IcebergCatalog impl in Polaris uses the return status of the persistence result to determine whether or not the drop failure was due to being not found. If not found, it returns false per the interface definition, which the CatalogHandlerUtil transforms into the appropriate not found exception. Otherwise it throws it's own error corresponding to the cause.

Ran tests locally + S3 Tests with drop
Screenshot 2026-02-06 at 3 57 35 PM
Screenshot 2026-02-06 at 3 57 21 PM

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -434,7 +434,7 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
         default:
           throw new ServiceFailureException(
-              "Unexpected error dropping table %s: %s. ExtraInfo: %s",
+              "Failed to drop table %s, status=%s, extraInfo=%s",
               tableIdentifier,
               dropEntityResult.getReturnStatus(),
               dropEntityResult.getExtraInformation());
@@ -674,8 +674,10 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
         default:
           throw new ServiceFailureException(
-              "Unexpected error dropping namespace %s: %s. ExtraInfo: %s",
-              namespace, dropEntityResult.getReturnStatus(), dropEntityResult.getExtraInformation());
+              "Failed to drop namespace %s, status=%s, extraInfo=%s",
+              namespace,
+              dropEntityResult.getReturnStatus(),
+              dropEntityResult.getExtraInformation());
       }
     }
 
@@ -874,7 +876,7 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
         default:
           throw new ServiceFailureException(
-              "Unexpected error dropping view %s: %s. ExtraInfo: %s",
+              "Failed to drop view %s, status=%s, extraInfo=%s",
               identifier,
               dropEntityResult.getReturnStatus(),
               dropEntityResult.getExtraInformation());
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
@@ -1987,7 +1987,7 @@ public abstract class AbstractIcebergCatalogTest extends CatalogTests<IcebergCat
 
     Assertions.assertThatThrownBy(() -> spiedCatalog.dropNamespace(NS))
         .isInstanceOf(ServiceFailureException.class)
-        .hasMessageContaining("Unexpected error dropping namespace")
+        .hasMessageContaining("Failed to drop namespace")
         .hasMessageContaining("UNEXPECTED_ERROR_SIGNALED");
   }
 
@@ -2012,7 +2012,7 @@ public abstract class AbstractIcebergCatalogTest extends CatalogTests<IcebergCat
 
     Assertions.assertThatThrownBy(() -> spiedCatalog.dropTable(TABLE, false))
         .isInstanceOf(ServiceFailureException.class)
-        .hasMessageContaining("Unexpected error dropping table")
+        .hasMessageContaining("Failed to drop table")
         .hasMessageContaining("UNEXPECTED_ERROR_SIGNALED");
   }
 
@@ -2024,8 +2024,7 @@ public abstract class AbstractIcebergCatalogTest extends CatalogTests<IcebergCat
     // Create a spy of the metaStoreManager to simulate an undroppable entity error
     PolarisMetaStoreManager spiedManager = spy(metaStoreManager);
     doReturn(
-            new DropEntityResult(
-                BaseResult.ReturnStatus.ENTITY_UNDROPPABLE, "Entity is protected"))
+            new DropEntityResult(BaseResult.ReturnStatus.ENTITY_UNDROPPABLE, "Entity is protected"))
         .when(spiedManager)
         .dropEntityIfExists(any(), anyList(), any(), anyMap(), anyBoolean());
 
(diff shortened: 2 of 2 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -423,7 +423,22 @@ public boolean dropTable(TableIdentifier tableIdentifier, boolean purge) {
...
+          return false;
+
+        case BaseResult.ReturnStatus.ENTITY_UNDROPPABLE:
+          throw new ForbiddenException(
+              "Table %s cannot be dropped: %s",
+              tableIdentifier, dropEntityResult.getExtraInformation());
+
+        default:
+          throw new ServiceFailureException(
+              "Unexpected error dropping table %s: %s. ExtraInfo: %s",
```
flyrain: Minor: 
```suggestion
              "Failed to drop table %s, status=%s, extraInfo=%s",
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -648,12 +663,23 @@ public boolean dropNamespace(Namespace namespace) throws NamespaceNotEmptyExcept
...
+        case BaseResult.ReturnStatus.NAMESPACE_NOT_EMPTY:
+        case BaseResult.ReturnStatus.CATALOG_NOT_EMPTY:
+          throw new NamespaceNotEmptyException("Namespace %s is not empty", namespace);
+
+        case BaseResult.ReturnStatus.ENTITY_NOT_FOUND:
+          return false;
+
+        default:
+          throw new ServiceFailureException(
+              "Unexpected error dropping namespace %s: %s. ExtraInfo: %s",
```
flyrain: Minor: "Failed to drop namespace %s, status=%s, extraInfo=%s"?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -835,8 +861,26 @@ public boolean dropView(TableIdentifier identifier) {
...
+        case BaseResult.ReturnStatus.ENTITY_NOT_FOUND:
+          return false;
+
+        case BaseResult.ReturnStatus.ENTITY_UNDROPPABLE:
+          throw new ForbiddenException(
+              "View %s cannot be dropped: %s", identifier, dropEntityResult.getExtraInformation());
+
+        default:
+          throw new ServiceFailureException(
+              "Unexpected error dropping view %s: %s. ExtraInfo: %s",
```
flyrain: Minor: Failed to drop view %s, status=%s, extraInfo=%s?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -648,12 +663,23 @@ public boolean dropNamespace(Namespace namespace) throws NamespaceNotEmptyExcept
                 Map.of(),
                 realmConfig.getConfig(FeatureConfiguration.CLEANUP_ON_NAMESPACE_DROP));
 
-    if (!dropEntityResult.isSuccess() && dropEntityResult.failedBecauseNotEmpty()) {
-      throw new NamespaceNotEmptyException("Namespace %s is not empty", namespace);
+    if (!dropEntityResult.isSuccess()) {
+      switch (dropEntityResult.getReturnStatus()) {
+        case BaseResult.ReturnStatus.NAMESPACE_NOT_EMPTY:
+        case BaseResult.ReturnStatus.CATALOG_NOT_EMPTY:
```
flyrain, line 669: Is `CATALOG_NOT_EMPTY` a valid status here?

HonahX: Same question : )

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -423,7 +423,22 @@ public boolean dropTable(TableIdentifier tableIdentifier, boolean purge) {
         dropTableLike(
             PolarisEntitySubType.ICEBERG_TABLE, tableIdentifier, storageProperties, purge);
     if (!dropEntityResult.isSuccess()) {
-      return false;
+      switch (dropEntityResult.getReturnStatus()) {
+        case BaseResult.ReturnStatus.ENTITY_NOT_FOUND:
+          return false;
+
+        case BaseResult.ReturnStatus.ENTITY_UNDROPPABLE:
+          throw new ForbiddenException(
```
HonahX, line 431: Do we need this for table/view? I remember this is for `service_admin` and `catalog_admin` role which are not droppable.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
travis-bowen: So I also think it feels weird, but it is what the existing method uses, so I kept it to minimize logic changes.
https://github.com/apache/polaris/blob/7de070d59008afa6d369268cec88f27d555ead7b/polaris-core/src/main/java/org/apache/polaris/core/persistence/dao/entity/DropEntityResult.java#L76

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
travis-bowen: You're right - we don't need it. I lean towards keeping it since we can't be certain of any custom implementations but I'm not tied to keeping it if you feel strongly.

flyrain (approved):

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
HonahX: I am ok to keep it. I feel like we probably should not group all the returnStatus in a single place.. but as Yufei mentioned that's a follow-up thing

flyrain:
Thanks @travis-bowen for the fix, and thanks @HonahX for the review.

PR #3694, opened 2026-02-07
Docs: Fix the misleading deprecation labels

flyrain:
We shouldn't make these configs as deprecation, more details are in https://lists.apache.org/thread/5rf8q2k8g54f9x78h1lwyvr3yb3pllnd

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -128,7 +128,8 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(true)
           .buildFeatureConfiguration();
 
-  @SuppressWarnings("deprecation")
+  @SuppressWarnings(
+      "deprecation") // catalogConfigUnsafe() is deprecated, not the feature config itself
```
dimas-b, line 132: nit: maybe put the comment on a new line for nicer formatting?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -164,7 +165,8 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
-  @SuppressWarnings("deprecation")
+  @SuppressWarnings(
+      "deprecation") // catalogConfigUnsafe() is deprecated, not the feature config itself
```
dimas-b, line 169: optional: even better, I'd prefer un-deprecating `catalogConfigUnsafe()` because:

1) it is actually used as a first-class method in `polaris-core` to ensure backward compatibility at the user level

2) downstream projects are not likely to abuse the `catalogConfigUnsafe()` method due to the "unsafe" mark in the method name

3) the method may be useful downstream for the same reason it is used in `polaris-core`

dimas-b: optional: IMHO, it would make sense to deprecate `catalogConfigUnsafe()` only when  `polaris-core` has no use cases for it.

adutra: I've always advocated for un-deprecating this method: what is deprecated is the property name, not the method that we use to handle deprecated names.

I doubt the method will ever disappear: more property names could be deprecated in the future. The current method name `catalogConfigUnsafe` focuses on one specific case of deprecated names that happened a while ago; I would actually rename it to `deprecatedCatalogConfig` and un-deprecate it.

My 2 cents 😄

flyrain: I agreed with both of you. +1 on un-deprecating and rename to deprecatedCatalogConfig. This PR focuses on the doc change. We can un-deprecate the method later, which may need another discussion on dev ML. WDYT?

dimas-b: The "problem" this PR attempts to fix, exists because `catalogConfigUnsafe()` is deprecated in the first place. Since we all agree that it should not be deprecated, lets un-deprecate it and then doc changes are not necessary.

Renaming can happen later, I suppose, but could be done together with un-deprecation.

flyrain: The site doc change has to happen, no matter un-deprecation happens or not, right?

dimas-b: Sorry, my statement was not accurate. Doc changes in this PR LGTM, but I suppose code comment change are not necessary.... but I guess the whole `@SuppressWarnings` will be removed when `catalogConfigUnsafe()` is un-deprecated, so I'm fine merging "as is".

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: Follow-up PR: #3735

PR #1480, opened 2025-04-28
Remove unused PolarisCredentialVendor.validateAccessToLocations()

dimas-b:

PR #3699, opened 2026-02-08
core: Credential Vending Refactor

tokoko:
- Decouples credential vending from `StorageCredentialCache` by moving orchestration logic into `StorageAccessConfigProvider`
- Remove credential vending from MetaStoreManager: PolarisCredentialVendor interface, StorageCredentialsVendor, and getSubscopedCredsForEntity() implementations are removed from MetaStoreManager implementations. Credential vending no longer round-trips through the persistence layer.
- Moves caching into storage integrations: Each PolarisStorageIntegration subclass now owns its StorageCredentialCache interaction and builds cloud-specific cache keys.
- Simplify StorageAccessConfigProvider: Now directly resolves the storage integration and delegates to it, instead of going through StorageCredentialsVendor → PolarisCredentialVendor → MetaStoreManager → persistence layer → integration provider. Entity is no longer re-loaded from persistence during credential vending.


polaris-core/src/main/java/org/apache/polaris/core/persistence/TransactionWorkspaceMetaStoreManager.java
```diff
@@ -317,31 +313,6 @@ public void deletePrincipalSecrets(
     throw illegalMethodError("loadTasks");
   }
 
-  @Override
-  public @Nonnull ScopedCredentialsResult getSubscopedCredsForEntity(
```
dimas-b: While I agree that this change makes sense in general, `polaris-core` is traditionally sensitive to API changes (despite the standing [evolution](https://polaris.apache.org/in-dev/unreleased/evolution/) guidelines).

I'd propose to keep this method for now, even if it will be unused in OSS code.

dimas-b: The NoSQL impl. can be replaced with a `throw new UnsupportedOperationException()`, but let's keep related code in `polaris-core`.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -95,94 +77,27 @@ private long maxCacheDurationMs(RealmConfig realmConfig) {
...
-              scopedCredentialsResult.getExtraInformation());
-        };
-    return cache.get(key, loader).toAccessConfig();
+  public StorageAccessConfig getOrLoad(
+      StorageCredentialCacheKey key,
+      RealmConfig realmConfig,
+      Supplier<StorageAccessConfig> loader) {
+    long maxCacheDurationMs = maxCacheDurationMs(realmConfig);
+    return cache
+        .get(
```
dimas-b: The `Supplier` in this `.get` method in Caffeine is supposed to be fast, but I believe in our case it may hit network (e.g. STS), which is not desirable... It might be best to go back to `LoadingCache`, which allows loaders to be blocking.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
```diff
@@ -95,94 +77,27 @@ private long maxCacheDurationMs(RealmConfig realmConfig) {
...
+            key,
+            k -> {
+              LOGGER.atDebug().log("StorageCredentialCache::load");
+              StorageAccessConfig accessConfig = loader.get();
```
dimas-b: Ideally, I think this should be something like `loader.get(key)`. That is to ensure we produce `StorageAccessConfig` from exactly the same data that is used as the cache key, i.e. any change in key results in regeneration of `StorageAccessConfig` and any two `StorageAccessConfig` objects are logically equivalent if their keys match. WDYT?

dimas-b (commented):
Thanks for keeping this refactoring going, @tokoko !

From my POV this PR is moving in the right direction. I've got just a couple of relatively minor comments.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
tokoko: while I agree that it would be ideal, we probably still wouldn't be able to make a loader dependent 100% on the key, for example the key only contains catalogId, rather than a resolved polaris entity. In the current codebase, while key is available in the loader, it's only partially used. My thinking was that unless we make loader fully dependent on the key it makes little sense to make the code more confusing when some of the params depend on the key and others don't.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
tokoko: I can revert, but I have to say my limited (ai-assisted) research convinced me that there's no point in using a `LoadingCache` with a dummy loader if you're always calling a get with an explicit loader. The behavior should be the same with either. might be wrong, though. happy to trust you on this.

dimas-b: I agree, that the old code was using `LoadingCache` in a strange way... yet, I believe `LoadingCache` in itself is the right tool here :sweat_smile:

dimas-b: This is related to using the "key" as input for Storage Config generation. If we could achieve that, `LoadingCache` would fit naturally.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: Re: confusing code - TBH, IMHO the _current_ (on `main`) code is very confusing because the reader has to think very carefully about which parameters go through the cache and which go outside and whether they are consistent :sweat_smile: 

I think we're pretty close to "ideal" here. The "entity" this code deals with, has already been "found" and loaded, we only need to extract the storage configuration JSON from it in front of the cache.

`PolarisStorageIntegration` is loaded based on the storage config, which can be done past the cache.

In each call, the key will be used in full as required for a particular storage backend + feature flags. Different calls will have different data in their keys, of course, but that is the idea :) It will allow one cache per JVM to handle all use cases. We could even have different expiry strategies per storage type/config, but that's looking to far into the future :sweat_smile: 

Would renaming `StorageCredentialCacheKey` to `StorageAccessConfigParameters` make the code clearer?

dimas-b:
@tokoko : it looks like this PR has a lot of conflicts now... CI will not run because of that :sweat_smile:

tokoko:
@dimas-b that's fine, this one was more for you to take a look :laughing:  I'll take care of the conflicts and bunch of other stuff that's still unfinished

dimas-b:
@tokoko : The general approach in this PR is fine from my POV :+1: If you have time I think we can iterate.

tokoko:
yeah, I'm gonna rebase and make it ready for a review this weekend.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -91,74 +145,53 @@ public PolarisStorageIntegrationProviderImpl(
...
-                  @Nonnull Set<String> allowedReadLocations,
-                  @Nonnull Set<String> allowedWriteLocations,
-                  @Nonnull PolarisPrincipal polarisPrincipal,
-                  Optional<String> refreshCredentialsEndpoint,
-                  @Nonnull CredentialVendingContext credentialVendingContext) {
-                // FILE storage does not support credential vending
-                return StorageAccessConfig.builder().supportsCredentialVending(false).build();
-              }
+    return (PolarisStorageIntegration<T>)
+        switch (polarisStorageConfigurationInfo.getStorageType()) {
```
dimas-b: How about we rename `getStorageIntegrationForConfig()` to `getStorageIntegrationForType()`?

I do not think we can expect the bulk of config data to be used in producing a `PolarisStorageIntegration` instance.

I believe old callers of this method (e.g. `AbstractTransactionalPersistence.createStorageIntegration()`)  are not very meaningful now that storage config is generated by application-scoped beans... so we might as well make it explicit.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -31,56 +31,110 @@
...
       RealmConfig realmConfig,
-      Clock clock) {
-    this.storageConfiguration = storageConfiguration;
-    this.stsClientProvider = stsClientProvider;
-    this.stsCredentials = Optional.empty();
-    this.gcpCredsProvider = storageConfiguration.gcpCredentialsSupplier(clock);
-    this.realmConfig = realmConfig;
+      Clock clock,
+      StorageCredentialCache cache) {
+    Supplier<RealmConfig> realmConfigSupplier = () -> realmConfig;
```
dimas-b: I'm not sure we need a `Supplier` here. `realmConfig` is going to be a CDI proxy bridging from the application scope of this bean to the current request scope of `RealmConfig`. We should be able to pass that proxy to `AwsCredentialsStorageIntegration` (and others) "as is".

The `Supplier` does not change the fact that the config is a CDI proxy.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -131,14 +144,30 @@ public StorageAccessConfig getStorageAccessConfig(
...
+    RealmConfig realmConfig = callContext.getRealmConfig();
+
+    // Get the right integration for this storage type
+    String storageConfigStr =
+        storageInfoEntity
+            .getInternalPropertiesAsMap()
+            .get(PolarisEntityConstants.getStorageConfigInfoPropertyName());
+    PolarisStorageConfigurationInfo storageConfig =
+        PolarisStorageConfigurationInfo.deserialize(storageConfigStr);
+    var integration = storageIntegrationProvider.getStorageIntegrationForConfig(storageConfig);
```
dimas-b: This `integration` feeds directly into the `getOrLoadSubscopedCreds()` call below. I think we can further simplify the flow and obtain `StorageAccessConfig` directly from `storageIntegrationProvider`.

See my other comment about older callers to `getStorageIntegrationForConfig()` becoming useless (as far as I can tell).

dimas-b (commented):
I like the idea of using application-scoped beans for producing `StorageAccessConfig` in runtime.

Posting some preliminary comments below. I did not review every single line, but the general approach LGTM 👍

flyrain:
Thanks @tokoko for working on it. Is there a dev mailing thread for it? If not, I'd recommend to have one for better visibility. It is generally a good practice to engage with the community earlier to avoid last minute surprises.

dimas-b:
@flyrain : This PR was mentioned on the S3 Tables thread: https://lists.apache.org/thread/c5nc7144zlsb5vdjmpsgbc0h0g5no9c0

Do you think we need a separate discussion for this?

tokoko:
@flyrain sure, I'll make a bit more changes, change PR description to be more accurate with latest changes and send out an email on the devlist. For me, the main goal of this PR is to demo the overall direction and build consensus on it. P.S. I'd be okay to split these changes up over a series of smaller PRs if people think that's more appropriate.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
tokoko: yeah, that makes sense.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
tokoko: you're right, that's most likely redundant. I don't have my java brain on today :smile:

polaris-core/src/main/java/org/apache/polaris/core/persistence/TransactionWorkspaceMetaStoreManager.java
flyrain: cc @dennishuo

dennishuo: Thanks for the ping! Yes, I agree with @dimas-b that this is a sensitive API surface. Though we should indeed be able to *decouple* the various interfaces pulled in/aggregated by `PolarisMetaStoreManager`:

        PolarisSecretsManager,
        PolarisGrantManager,
        PolarisCredentialVendor,
        PolarisPolicyMappingManager,
        PolarisEventManager,
        PolarisMetricsManager

those interfaces were specifically written as part of SPIs, and at least the same *functionality* provided by those extensibility points needs to be preserved more strictly than syntax evolution of those interfaces, unless a much deeper attempt is made amongst Polaris users to agree on a migration path off of such functionality.

There's some discussion about the list of core SPIs in here: https://lists.apache.org/thread/0nj24zro7kyctqfnlml08ppo7zs9xcqs

The way that applies to this PR is that the interaction with the persistence layer is intentional, especially for the TransactionalMetaStoreManager path.

The basic use case if if a custom Polaris deployment uses subclasses that leverage those SPIs in order to perform additional book-keeping around storage-integration functionality and needs that to be transactional with entity creation/updates. This is best highlighted by thinking about the interaction between the subset of `PolarisCredentialVendor` method implementations with `IntegrationPersistence` which has these methods:

    createStorageIntegration
    persistStorageIntegrationIfNeeded
    loadPolarisStorageIntegration

These are in the interface specifically to define the SPI by which users customize a "lease, commit, use" model for secondary metadata necessary for StorageIntegrations. In this current PR, it's not clear what happens to those methods now -- will `IntegrationPersistence::loadPolarisStorageIntegration` just be ignored?

Overall, I think the refactoring on the credential caching layer should still be possible, while preserving the `PolarisCredentialVendor` interface so that implementations of `PolarisCredentialVendor` still work correctly.

There are some nuances to this that may be better discussed on the mailing list thread.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegrationProvider.java
```diff
@@ -21,11 +21,11 @@
...
 
 /**
- * Factory interface that knows how to construct a {@link PolarisStorageIntegration} given a {@link
+ * Factory interface that knows how to return a {@link PolarisStorageIntegration} for a given {@link
  * PolarisStorageConfigurationInfo}.
  */
 public interface PolarisStorageIntegrationProvider {
   <T extends PolarisStorageConfigurationInfo>
-      @Nullable PolarisStorageIntegration<T> getStorageIntegrationForConfig(
+      @Nullable PolarisStorageIntegration<T> getStorageIntegration(
```
dennishuo: This rename might actually help highlight the main discrepancy that will cause issues here. The SPI contract via `IntegrationPersistence` defines the ability to support stateful StorageIntegration objects. In the current status quo, that means `PolarisStorageIntegrationProvider.getStorageIntegrationForConfig` is more like `constructProspectiveNewStorageIntegrationBasedOnConfig`.

It's an unfortunate shortcut that the same `PolarisStorageIntegration` returned there is what holds a `getSubscopedCreds` method.

tokoko:
@dennishuo @dimas-b I changed the PR and reinstated PolarisCredentialVendor SPI, but the thing that bothers me now is that PolarisCredentialVendor.getStorageAccessConfig and PolarisStorageIntegration.getSubscopedCreds signatures are converging to essentially the same thing — both take locations, actions, refresh endpoint, context and return StorageAccessConfig. I think the current design is coupling PolarisStorageIntegrationProvider (which effectively includes PolarisStorageIntegration) and PolarisCredentialVendor too much. Having two SPIs that mirror each other adds complexity without buying us anything.

The root cause is that PolarisCredentialVendor is trying to be a unified SPI that wraps both "get the right integration" and "vend credentials" into one call. So what if instead of putting PolarisStorageIntegrationProvider behind PolarisCredentialVendor, we make PolarisStorageIntegrationProvider a first-class citizen?

Planned flow will look like this:
Vending:

- PolarisStorageIntegrationProvider.getStorageIntegration(List<PolarisEntity> resolvedEntityPath)
    - oss: resolves StorageConfig from full entity path and gets cached per-config PolarisStorageIntegration singleton from the cache.
    - custom: gets entity (with findInHierarchy) and calls IntegrationPersistence.loadPolarisStorageIntegration directly
    
- StorageIntegration.getSubscopedCreds(...) - since persistence question is out of the way and PolarisStorageIntegrationProvider SPI gives an implementer full control over the integration objects that are retrieved, I think we no longer need another PolarisCredentialVendor SPI for the next step. it would essentially be identical to PolarisStorageIntegration anyway.
    
CreateCatalog:
- createStorageIntegration / persistStorageIntegrationIfNeeded:
    - oss: we make both a no-op, createStorageIntegration no longer defers to PolarisStorageIntegrationProvider since PolarisStorageIntegrationProvider signature has changed and current impl discards the result anyway.
     - custom: this is unchanged, implements lease-commit steps here.

This change would give us one SPI (PolarisStorageIntegrationProvider) instead of two that mirror each other and also keep IntegrationPersistence storage methods around for custom impls that need them.

Curious what you guys think...

dennishuo:
Sorry for the delay responding to this! I've been digesting this more and I think there's a bigger convergence we can align with.

I agree with what you said that if we only reintroduce PolarisCredentialVendor in the proposed way, it's a bit confusing whether a `StorageIntegration` itself is also fundamentally a CredentialVendor interface (with partially-bound request-scoped state). I'm mulling over whether what you describe is topologically equivalent to the other long-term direction we wanted to go, just with different class names.

**The TL;DR:** We should incrementally try to converge with how the analogous ConnectionConfig flow works for federation, especially since the sigv4-to-Glue federation credentials are pretty much exactly the same as sigv4-to-s3 storage credentials. Your proposal does seem to align somewhat, so I'd say go ahead and try to have StorageIntegration be the first-class entrypoint. We should consider accordingly moving the items related to request scope (`realmConfig`, `storageConfig`) out of the `getSubscopedCreds` arguments and force them to be bound at construction-time, if `StorageIntegrationProvider.getStorageIntegration(resolvedEntityPath)` is intended to be the stateful lookup.

Extended analysis:

Very abstractly, there are these four things:

1. "User-Requested Config" -- Basically the StorageConfingInfo that comes in from a CreateCatalog request initially
2. "Full config with system-assigned internal references" -- This is currently a confusing "hidden" ghost entity but  only hinted at due to `createStorageIntegration` and `persistStorageIntegrationIfNeeded` and the fact that `PolarisCredentialVendor` takes the `PolarisEntity` when getting subscoped credentials. The `StorageIntegration` object itself kind of doubles as an effective "Data Access Object" (DAO) that wraps this "ghost entity"
3. "Handle into the entity-specific credential vending layer" - The `StorageIntegration`, by virtue of being the DAO into the hidden layer, basically fills this role. Importantly, the `IntegrationPersistence.loadPolarisStorageIntegration` method is implied to interact with persistence because it takes `PolarisBaseEntity` as an argument. 
4. "Factory for the credential-vending handle, that requires full config with system-assigned state" - This is the `IntegrationPersistence.loadPolarisStorageIntegration` method basically

The fact that StorageIntegrationProvider at `HEAD` only takes what amounts to (1) - the user-requested config - and serves the purpose of (4) to provide (3) is what causes confusion.

What I remembered is that we actually already did try to redesign the flow of such credential-handling to accommodate this same concept in a more explicit way. See:

1. Secrets management for Catalog Federation: https://docs.google.com/document/d/1JPNx5vL4vM8DqwRwnBIPiQxwN4MXOdGx4Ki0j7vgwSM/edit?tab=t.0
2. The sigv4 support for that secrets management that is really the same flow as storage-credential vending, but for a different set of IAM actions: https://github.com/apache/polaris/pull/2190

Reposting @XJDKC 's diagram here:

polaris_connection_config_sigv4

In this world:
1. User-Requested config is the `ConnectionConfigInfoModel`
2. Config with System-assigned references is the `ConnectionConfigInfoDpo` that contains an "identityReference"
3. Handle into credential-vending - This is actually multi-step - [AwsIamServiceIdentityCredential](https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/identity/credential/AwsIamServiceIdentityCredential.java) (concrete `ServiceIdentityCredential` class) is basically the handle into the *service identity* that can be used to further mint short-lived access, and [SigV4ConnectionCredentialVendor](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/SigV4ConnectionCredentialVendor.java) (concrete `ConnectionCredentialVendor ` class) is the thing that actually gets the short-lived credentials, as evidenced by it holding an `StsClient`. (Not explicitly depicted in the diagram, have to look at the code)
4. Factory that takes internal config/reference to get the handle - [ServiceIdentityProvider](https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/identity/provider/ServiceIdentityProvider.java#L72) (not exactly in the diagram since it's outdated, but roughly, the interaction between "ServiceIdentityRegistry" and "ResolvedServiceItentity" is captured in `ServiceIdentityProvider`)

I guess the rough equivalences are:

- `ServiceIdentityProvider` is kind of like `StorageIntegrationProvider` but it's also kind of like `IntegrationPersistence` since it also defines the `allocate` method
- `ServiceIdentityCredential` is like `StorageIntegration`
- `ConnectionCredentialVendor ` is like a combination of `StorageIntegration`, `PolarisCredentialVendor`, `StorageCredentialsVendor`, and `StorageAccessConfigProvider`

One key difference in behavior is in the ConnectionConfig path, I guess we ended up putting the following all the way in `PolarisAdminService`:

        Optional<ServiceIdentityInfoDpo> serviceIdentityInfoDpoOptional =
            serviceIdentityProvider.allocateServiceIdentity(connectionConfigInfo);

        entity =
            new CatalogEntity.Builder(entity)
                .setConnectionConfigInfoDpoWithSecrets(
                    connectionConfigInfo,
                    processedSecretReferences,
                    serviceIdentityInfoDpoOptional.orElse(null))
                .build();

This is in contrast to putting the "allocate" hook into the inner transaction-aware block in PolarisMetaStoreManager. So it doesn't support as good of transactional semantics as how we handle StorageConfig, but on the plus side, it forces the referential service-identity info to be first-class in the `CatalogEntity` itself, so we don't have fully invisible "ghost entities" representing the underlying credential allocation.

Overall, fully merging the Storage credential flow into the Connection credential flow is probably too big of an undertaking to do here, and maybe we also want to adjust the Connection credential code at the same time, but I think any incremental refactoring to make it structurally compatible is probably the right direction.

All this to say, yes, I think your proposal is somewhat structurally compatible, with a couple key points:

1. StorageIntegrationProvider changes from being semantically `allocateForConfig(userSpecifiedConfig)` to being `fetchForEntity(entityPath)`
2. StorageIntegration *instances* should be thought of as fully representing the resolved entityPath already, rather than being application/type-scoped -- this might mean we want `realmConfig` and `storageConfig` to be constructor-time bindings rather than arguments to `getSubscopedCreds` (where the `storageConfig` will actually be derived from `entityPath` during construction by `StorageIntegrationProvider`)

tokoko:
@dennishuo thanks for the explanation. some sort of convergence between normal credential vending and federation makes sense. I'm not too familiar with federation codebase so I'll leave that out of scope for now, but I'll keep that in mind. that would probably make something like #4052 cleaner.

I went ahead and made quite a few changes. technically StorageIntegration instances returned by `StorageIntegrationProvider` now are storageConfig-scoped rather than entity-scoped which probably doesn't make much of a difference as of now. (other than the fact that storage locations need to be passed as parameters, which they might not have been otherwise, not sure about that though). 

I'm thinking of trimming PolarisStorageIntegration itself next:

- @dimas-b I saw you had previously removed #1480 from Vendor but left it on the integration layer, any reason for that? from what I saw it's completely unused and can be deleted from there as well. with that gone `InMemoryStorageIntegration` class also no longer makes sense to be kept around. The only active code there is a static method that can be refactored somewhere else.
- `getStorageIdentifierOrId` is another piece of unused dead code that I think can be removed.
- On the other hand, I'm thinking of splitting current `PolarisStorageIntegration` into a slimmer interface and renaming current code to `CachingStorageIntegration`. I think that should be better for persistence-based credentials that are unlikely to require caching (?). Also in general if we're making `PolarisStorageIntegration` a primary SPI, it should really be an interface.
- All in all, this is the interface I'm going for (also factoring in @dimas-b's suggestion of passing raw locations and actions rather than separate read/write locations):
```
public interface PolarisStorageIntegration {

    StorageAccessConfig getStorageAccessConfig(
        @Nonnull Set<String> locations,
        @Nonnull Set<PolarisStorageActions> actions,
        @Nonnull Optional<String> refreshEndpoint,
        @Nonnull CredentialVendingContext context);
  }
```

dimas-b:
> [@dimas-b](https://github.com/dimas-b) I saw you had previously removed [#1480](https://github.com/apache/polaris/pull/1480) from Vendor but left it on the integration layer, any reason for that? from what I saw it's completely unused and can be deleted from there as well. [...]

I did not touch `PolarisStorageIntegration.validateAccessToLocations()` because it is considered part of the historical "core" and I am now aware of the full implications of such a change.

From my POV that method can be deleted, but I'd appreciate it if @dennishuo could share his view on that.

polaris-core/src/main/java/org/apache/polaris/core/persistence/IntegrationPersistence.java
```diff
@@ -137,16 +142,4 @@ <T extends PolarisStorageConfigurationInfo> void persistStorageIntegrationIfNeed
...
-  /**
-   * Load the polaris storage integration for a polaris entity (Catalog,Namespace,Table,View)
-   *
-   * @param callContext the polaris call context
-   * @param entity the polaris entity
-   * @return a polaris storage integration
-   */
-  @Nullable
-  <T extends PolarisStorageConfigurationInfo>
-      PolarisStorageIntegration<T> loadPolarisStorageIntegration(
```
dimas-b, line 150: @dennishuo : Are you ok with removing this method?

dimas-b:
The current state of this PR looks reasonable to me. If @dennishuo is ok with the general direction, I'll take a deeper look :wink:

polaris-core/src/main/java/org/apache/polaris/core/persistence/IntegrationPersistence.java
tokoko: just to clarify, the reason I removed them was that they were no longer called from anywhere in oss code anymore. The removal from the api in oss does not mean they need to be removed from concrete implementations though, of course. I figured that would be fine for the use case Dennis outlined, but I can reinstate if that's not the case.

dimas-b: +1 to @tokoko points. I also do not see any use case for this method in OSS, still @dennishuo , please respond if you have any concerns from your side.

dennishuo:
Sorry for the delays, took a some deep diving archaeology to assess the impact of all the suggested changes downstream, though I only just realized earlier this evening that you were just saying you'd remove `getStorageIdentifierOrId` in a future PR, not this one. I've also always disliked that method, and confirmed there will be some downstream refactoring required in custom code if that's removed, but it should be feasible to do (the cleaner way to achieve the more general functionality there is to add a method that abstracts out injecting "book-keeping" info into a PolarisEntity based on the StorageIntegration, like mentioned in the Catalog ConnectionConfig stuff, but that's for a longer-term refactor).

For the current PR:

Removing `loadPolarisStorageIntegration` is feasible, but since it is indeed an SPI which is intended to be called by encouraged customizations (e.g. custom MetaStoreManager classes calling ((IntegrationPersistence) ms).loadPolarisStorageIntegration), it would be good to exercise the intended SPI-evolution process, which is to put it to a vote on the dev mailing list in case others want to assess impact as well.

For the suggested changes:

Removing getStorageIdentiferOrId - workable for the customization cases I know of, but since this is also used by customizations of MetaStoreManager, if we're going to have a dev mailing list vote for `loadPolarisStorageIntegration` anyways, perhaps we can also include this method's removal in that same vote thread (so that we don't have to have multiple threads in case it's most likely anyone impacted would feel equally about both methods)

Removing `validateAccessToLocations` - this one seems fairly clean to remove, since the TODO about plumbing the check through to a persistence impl instead of conjuring an InMemoryStorageIntegration never materialized. Personally I could go either way on whether to include this one in the vote. I suppose one could argue that since the intended caller of the `validateAccessToLocations` wouldn't have been from `polaris-core` but rather `polaris-service` (i.e. `BasePolarisCatalog`) that `validateAccessToLocations` isn't part of the SPIs that customizations against `polaris-core` would rely on. It's a bit of a blurry line here.

The main deciding factor would be starting from the core SPIs called out in https://lists.apache.org/thread/0nj24zro7kyctqfnlml08ppo7zs9xcqs and going one level deeper into their arguments and if the interaction between those core classes and the argument (including a PolarisStorageIntegration) is part of the core SPI then we should have it in the vote. 

**Regarding removal of read vs write and just having one locations list** - I'm not sure I understand this proposal. Subscoped policy strings give different sets of allowed locations for read vs write locations. @dimas-b what was the suggestion about no longer differentiating between read vs write? I must have missed this

tokoko:
@dennishuo thanks for the detailed response. If we're going in the ML vote direction, we might as well bundle all SPI changes in a single vote. we can then either remove everything that's no longer needed or I can simply deprecate them as part of this PR (storage id for example) and push removal for a follow up if you think that'll make adjusting to the changes easier for you. let me know.

to summarize: 
- we are dropping `PolarisCredentialVendor` (and `StorageCredentialsVendor`) SPI entirely.
- `PolarisStorageIntegrationProvider` is made into a top-level SPI with a single method: `PolarisStorageIntegration getStorageIntegration(List<PolarisEntity> resolvedEntityPath);`
- we are dropping `loadPolarisStorageIntegration` from `IntegrationPersistence` because it's not called from oss code.
- we are dropping everything from `PolarisStorageIntegration` except for `StorageAccessConfig getStorageAccessConfig(...)` method.

The only thing left is to agree on `getStorageAccessConfig` signature. let me know if I missed anything.

> I'm not sure I understand this proposal. Subscoped policy strings give different sets of allowed locations for read vs write locations

let me try to make the case for changing current params from:
```
boolean allowListOperation,
@Nonnull Set<String> allowedWriteLocations
```
to
```
@Nonnull Set<String> locations,
@Nonnull Set<PolarisStorageActions> actions,
```

the motivation for the change stems from two observations: 
- we never set `allowedReadLocations` and `allowedWriteLocations` to different sets of strings. the two lists are always identical in practice. the fact that storage policies (like iam policy strings) expect them to be supplied in some particular way is an implementation detail.
- credential vending flow only allows read/write locations and a list boolean to be set (which is extremely ugly as well imho :laughing:) when `PolarisStorageActions` is actually more expressive, it also contains a delete action. One hypothetical use case that would benefit from passing `PolarisStorageActions.DELETE` down to the storage integration layer is more fine-grained control and distinction between write/delete actions which are currently impossible to distinguish. For example a normal iceberg writer principal would get only `PutObject` privilege, while one that also can remove orphans files or expire snapshots would get `DeleteObject` privilege as well. (`PutObject` can also be destructive unless we mandate conditional writes like [this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes-enforce.html) but that's different story and would probably require upstream iceberg change that would allow setting `s3.write.if-none-match=true` in `S3FileIO` to signal clients to use conditional writes) Anyway, The point is that the current design discards important details (like DELETE action) and duplicates storage locations that are (at least currently) always identical.   
- If we want to be even more future-proof we could introduce something like this that would allow setting different storage actions for different sets of locations, but I couldn't really come up with a use case:
```
public record LocationGrant(
        @Nonnull Set<String> locations, 
        @Nonnull Set<PolarisStorageActions> actions
)

public interface PolarisStorageIntegration {
    StorageAccessConfig getStorageAccessConfig(
        @Nonnull List<LocationGrant> grants,
        @Nonnull Optional<String> refreshEndpoint,
        @Nonnull CredentialVendingContext context);
}
```

I get that this will introduce even more changes and headache for downstream implementations. It's just that if we're breaking things anyway and trying to build consensus around new SPI design, it's unlikely we will have better opportunity in the future for these sorts of changes.

dennishuo:
@tokoko +1 to bundling up SPI changes for a vote.

For the locations, I'm okay with refactoring a more advanced way to set different storage actions for different sets of locations. Keeping the muscle under the hood for policy-scoping logic to handle different sets of actions for different locations will be important to avoid that door inadvertently closing with other architectural decisions that start assuming a uniform action-set, even if we failed to properly capitalize on it so far.

The original use case in mind was in the case of cloned tables or other tables that originate from some legacy dataset in an "old" location that is no longer the new write-location, it's important to keep the legacy location "read-only" and only provide write access to the new location. It's a layer of protection against misbehaving client-side engines that may unexpectedly delete things from the legacy dataset location, for example.

In general with the future of better intent-conveyance in the APIs (e.g. loadTable for a subsequent write vs just for a SELECT), we'd also want to better distinguish whether to include write permissions at all. Some finer-grained operations, such as things that only want to compact a snapshot might only need write access to the metadata/ directory but not the data/ directory.

With better specification of intent in the API spec, there's some discussion of providing finer-grained privileges under TABLE_WRITE_PROPERTIES that distinguish between data-writes vs metadata-writes, and even within metadata writes, distinguishing between some sensitive ones like `setLocation` from more innocuous ones like snapshot tags.

In some cases you might even want to allow DELETE but not READ on a subdirectory. For example, you might run a shared snapshot-expiry maintenance engine on a highly sensitive dataset. It would need read + write on `metadata/` files but you might even only want to give it DELETE on `data/`.

dimas-b:
Re: voting for SPI changes - I think it would be fine, however, I'd suggest to have concrete changes in a PR first, then review the PR, then vote. This is because voting is supposed to be on specific changes. If some adjustments are called for during a vote, the whole process has to restart, so we'd better sort out all concerns before the vote even starts :sweat_smile: 

That said, from my personal POV posting a notice about changes on the `dev` ML and approving in GH is just as good.

persistence/nosql/persistence/metastore/src/testFixtures/java/org/apache/polaris/persistence/nosql/metastore/CdiProducers.java
```diff
@@ -35,9 +34,8 @@ public class CdiProducers {
   PolarisStorageIntegrationProvider producePolarisStorageIntegrationProvider() {
     return new PolarisStorageIntegrationProvider() {
       @Override
-      public @Nullable <T extends PolarisStorageConfigurationInfo>
-          PolarisStorageIntegration<T> getStorageIntegrationForConfig(
-              PolarisStorageConfigurationInfo polarisStorageConfigurationInfo) {
+      public @Nullable PolarisStorageIntegration getStorageIntegration(
+          java.util.List<org.apache.polaris.core.entity.PolarisEntity> resolvedEntityPath) {
```
dimas-b: nit: import

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -1251,31 +1250,23 @@ private List<PolarisPolicyMappingRecord> fetchPolicyMappingRecords(
...
+  public PolarisStorageIntegration createStorageIntegration(
+      @Nonnull PolarisCallContext callCtx,
+      long catalogId,
+      long entityId,
+      PolarisStorageConfigurationInfo polarisStorageConfigurationInfo) {
+    // No-op in OSS: the storage integration is resolved at credential-vending time via
+    // PolarisStorageIntegrationProvider.getStorageIntegration(resolvedEntityPath). This hook
+    // remains available for custom deployments that need to allocate/lease external state
+    // atomically with the catalog-creation transaction.
+    return null;
```
dimas-b, line 1273: NoSQL code throws in this case... I think JDBC should throw too for the sake of consistency.

As far as I can tell this method is not called is OSS anyway.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
```diff
@@ -1251,31 +1250,23 @@ private List<PolarisPolicyMappingRecord> fetchPolicyMappingRecords(
...
-  @Nullable
-  @Override
-  public <T extends PolarisStorageConfigurationInfo>
-      PolarisStorageIntegration<T> loadPolarisStorageIntegration(
-          @Nonnull PolarisCallContext callContext, @Nonnull PolarisBaseEntity entity) {
-    PolarisStorageConfigurationInfo storageConfig =
-        BaseMetaStoreManager.extractStorageConfiguration(diagnostics, entity);
-    return storageIntegrationProvider.getStorageIntegrationForConfig(storageConfig);
-  }
+      @Nullable PolarisStorageIntegration storageIntegration) {}
```
dimas-b, line 1280: same here: let's throw an `UnsupportedOperationException` for clarity

persistence/relational-jdbc/src/test/java/org/apache/polaris/persistence/relational/jdbc/MetricsReportPersistenceTest.java
```diff
@@ -63,9 +62,8 @@ void setUp() throws SQLException {
     PolarisStorageIntegrationProvider storageProvider =
         new PolarisStorageIntegrationProvider() {
           @Override
-          public <T extends PolarisStorageConfigurationInfo>
-              PolarisStorageIntegration<T> getStorageIntegrationForConfig(
-                  PolarisStorageConfigurationInfo config) {
+          public PolarisStorageIntegration getStorageIntegration(
+              java.util.List<org.apache.polaris.core.entity.PolarisEntity> resolvedEntityPath) {
```
dimas-b: nit: import

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapTransactionalPersistenceImpl.java
```diff
@@ -532,24 +530,16 @@ public void deletePrincipalSecretsInCurrentTxn(
...
+  public @Nullable PolarisStorageIntegration createStorageIntegrationInCurrentTxn(
+      @Nonnull PolarisCallContext callCtx,
+      long catalogId,
+      long entityId,
+      PolarisStorageConfigurationInfo polarisStorageConfigurationInfo) {
+    // No-op in OSS: the storage integration is resolved at credential-vending time via
+    // PolarisStorageIntegrationProvider.getStorageIntegration(resolvedEntityPath). This hook
+    // remains available for custom deployments that need to allocate/lease external state
+    // atomically with the catalog-creation transaction.
+    return null;
```
dimas-b, line 550: let's throw here too... it will provide a clear indication for needing extra work if new code changes touch this call path.

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: I believe this comment thread  still applies... WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/StorageAccessConfigProvider.java
```diff
@@ -154,68 +103,88 @@ public StorageAccessConfig getStorageAccessConfig(
...
+      @Nonnull Set<String> locations,
+      @Nonnull Set<PolarisStorageActions> storageActions,
+      @Nonnull Optional<String> refreshCredentialsEndpoint) {
+
+    boolean skipCredentialSubscopingIndirection =
+        callContext
+            .getRealmConfig()
+            .getConfig(FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
+    if (skipCredentialSubscopingIndirection) {
+      return StorageAccessConfig.builder().build();
```
dimas-b: Should we use `.supportsCredentialVending(false)` here too?

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -45,120 +49,136 @@
...
+      @Nonnull List<PolarisEntity> resolvedEntityPath) {
+    return PolarisStorageConfigurationInfo.findStorageInfoFromHierarchy(resolvedEntityPath)
+        .map(entity -> BaseMetaStoreManager.extractStorageConfiguration(diagnostics, entity))
+        .map(this::getOrCreateIntegration)
+        .orElse(null);
+  }
+
+  private PolarisStorageIntegration getOrCreateIntegration(
+      PolarisStorageConfigurationInfo storageConfig) {
+    return integrationCache.computeIfAbsent(storageConfig, this::createIntegration);
```
dimas-b: Is the `integrationCache` Map worth having? `CachingStorageIntegration` sub-classes a light-weight objects. Creating them on the spot is probably non-consequential. 

The heavy part is the `StorageCredentialCache`, which is shared already.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -57,86 +60,174 @@
...
+        PolarisStorageActions.WRITE,
+        PolarisStorageActions.DELETE,
+        PolarisStorageActions.ALL);
+  }
+
+  private static Set<String> locationsFor(
+      List<LocationGrant> grants, PolarisStorageActions... wantedActions) {
+    EnumSet<PolarisStorageActions> wanted = EnumSet.noneOf(PolarisStorageActions.class);
+    for (PolarisStorageActions a : wantedActions) {
+      wanted.add(a);
```
dimas-b: `addAll()`?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: ping :)

runtime/admin/src/main/java/org/apache/polaris/admintool/config/AdminToolProducers.java
```diff
@@ -67,9 +66,9 @@ public PolarisStorageIntegrationProvider storageIntegrationProvider() {
     return new PolarisStorageIntegrationProvider() {
       @Override
       @Nullable
-      public <T extends PolarisStorageConfigurationInfo>
-          PolarisStorageIntegration<T> getStorageIntegrationForConfig(
-              PolarisStorageConfigurationInfo polarisStorageConfigurationInfo) {
+      public PolarisStorageIntegration getStorageIntegration(
+          @jakarta.annotation.Nonnull
+              java.util.List<org.apache.polaris.core.entity.PolarisEntity> resolvedEntityPath) {
```
dimas-b: nit: import

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogRustFSSpecialIT.java
```diff
@@ -383,7 +383,6 @@ private void assertLoadTableWithVendedCredentialsFailsWithKmsError(TableIdentifi
                     id,
                     "ALL",
                     Map.of("X-Iceberg-Access-Delegation", VENDED_CREDENTIALS.protocolValue())))
-        .hasMessageContaining("Failed to get subscoped credentials")
```
dimas-b, line 386: Can we assert something else to be sure the error comes from where it is expected to come?

dimas-b (commented):
LGTM overall 👍 Please fix conflicts 🙂

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
tokoko: I think this was deliberate, otherwise I'd push for deletion of these methods as well. JdbcMetaStoreManagerFactory uses AtomicOperationMetaStoreManager which invokes both storage persistence methods.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
tokoko: same here

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
tokoko: I guess we settled on `getStorageIntegration`. I'll resolve this one

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogRustFSSpecialIT.java
tokoko: `software.amazon.awssdk.services.sts.model.StsException: (Service: Sts, Status Code: 400, Request ID: null) (SDK Attempt Count: 1)` this is all that is returned by rustfs. I can add `.hasMessageContaining("Service: Sts")`

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
tokoko: you're probably right. I'm not sure about `ServiceOptions.getFromServiceLoader(HttpTransportFactory.class, NetHttpTransport::new)` in gcp branch though. I can move it out of the constructor though.

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogRustFSSpecialIT.java
dimas-b: WDYT about keeping the old `Failed to get subscoped credentials` messge?

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
tokoko: maybe, I think we have a different problem now that each integration might have a different cache key... LoadingCache is supposed to have a single predefined supplier method, right? now that we have effectively split each integration to act independent from one another, the only realistic way to achieve this would be to have a single LoadingCache per integration (cached but not shared). wdyt?

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogRustFSSpecialIT.java
tokoko: changed

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
tokoko: refactored w/o integrationCache

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogRustFSSpecialIT.java
tokoko: this was kinda unintended, but we no longer "repackage" exceptions from integrations now that we got rid of some the intermediate steps. the exceptions that surface now are actual integration exceptions.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: AtomicOperationMetaStoreManager invokes this method conditionally on a property that is never set in OSS 😉  ... but it's a minor point. I'm ok to keep this method.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcBasePersistenceImpl.java
dimas-b: let's keep

polaris-core/src/main/java/org/apache/polaris/core/persistence/transactional/TreeMapTransactionalPersistenceImpl.java
dimas-b: pls ignore

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
dimas-b: if providers (e.g. `AwsCredentialsStorageIntegration`) are light-weight objects, we can create them inside the shared loader method from "type" information stored in the key... WDYT?

runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogRustFSSpecialIT.java
dimas-b: Yeah... having `Failed to get subscoped credentials` in the message is probably relevant only on the server side, but severs should really log the full stack trace, which will make the call path clear.

I'm fine with removing exception wrapping on the call path.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -45,120 +47,119 @@
...
-    }
-    return storageIntegration;
+  private PolarisStorageIntegration createIntegration(
+      PolarisStorageConfigurationInfo storageConfig) {
+    return switch (storageConfig.getStorageType()) {
+      case S3 -> awsFactory.apply((AwsStorageConfigurationInfo) storageConfig);
+      case GCS -> gcpFactory.apply((GcpStorageConfigurationInfo) storageConfig);
+      case AZURE -> azureFactory.apply((AzureStorageConfigurationInfo) storageConfig);
+      case FILE -> FILE_INTEGRATION;
+      default ->
```
dimas-b: IJ warns that this `default` is unnecessary 😅

polaris-core/src/main/java/org/apache/polaris/core/storage/cache/StorageCredentialCache.java
tokoko: Don't you have a "chicken and egg" problem there? you need `StorageIntegration` for cache key first... Or do you mean we can create them twice effectively?

dimas-b: Sorry, I do not see the "chicken", only the "egg" :wink:  which is the key. It is created first, then we go through the cache, then the loader function takes the key, creates a light-weight `AwsCredentialsStorageIntegration` and feeds the data from the key into it.

tokoko: I took a shot at it. basically added a `load` method to the cache key interface. each key carries it's own loader method and that's what loader method of the LoadingCache is calling. let me know what you think.

dimas-b: Sorry, not quite there yet from my POV :sweat_smile: 

So we derive storage config like this: entity -> PolarisStorageConfigurationInfo -> (1) PolarisStorageIntegration -> include CredentialVendingContext -> (cache) -> StorageAccessConfig

I believe we need to first combine `PolarisStorageConfigurationInfo` and `CredentialVendingContext` into a cache key. This requires (2) a concrete `PolarisStorageIntegration` (same as step 1 above). However, the key should only contain the data that acts as direct input into the config generation. So, it should not hold `PolarisStorageIntegration`. The cache loader function will re-derive (3) `PolarisStorageIntegration` from the cache key at load time (if it happens). This is a fast operation as far as I understand.

This is a fine point, I admit, but I think it is critical for caching correctness and indirectly for future improvement in this code.

tokoko: you're right, that was a poor implementation, I shouldn't have included integration itself as an aux. still I don't like the idea of re-deriving entire PolarisStorageIntegration object. the reason is that the first time it's instantiated, the point it to build a key, the second time around you build a credential. inputs for each are not necessarily identical, I changed session name derivation in AWS to highlight it. I switched to deriving session name during cache key generation and then simply using resolved string in the cache key directly. in other words, cache key generation needs entire CredentialVendingContext (which includes principal name), while credential compute needs only the session name string that's already resolved. Reinstantiating PolarisStorageIntegration with entire CredentialVendingContext is therefore wasteful.

I switched to providing fields like `stsClientProvider` and `credentialResolver` to cache key as aux fields instead of integration itself. the key then calls a **static** `compute` method on the integration which takes in only what it absolutely needs and params are coming from the fields in the cache key. let me know what you think.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
```diff
@@ -0,0 +1,118 @@
...
+  String roleSessionName();
+
+  @Value.Parameter(order = 8)
+  List<Tag> sessionTags();
+
+  // ---- aux: app-scoped invariants, excluded from equals/hashCode ----
+
+  @Value.Parameter(order = 9)
+  @Value.Auxiliary
+  StsClientProvider stsClientProvider();
```
dimas-b, line 70: I'm still a bit hesitant about keeping this state (effectively application-scoped) in the cache key, but current logic looks correct nonetheless, so I'm fine with this... Maybe we can improve more later :)

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Value.Parameter(order = 10)
+  @Value.Auxiliary
+  Function<AwsStorageConfigurationInfo, Optional<AwsCredentialsProvider>> credentialsResolver();
+
+  @Value.Parameter(order = 11)
+  @Value.Auxiliary
+  AwsStorageConfigurationInfo storageConfig();
```
dimas-b: This config is used inside the `.compute()` method... Why is it `Auxiliary`? If `storageConfig` is different, the value should be re-computed, right? 🤔 

Shouldn't we break it down into individual elements and use only the latter inside `compute()`?

polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java
```diff
@@ -68,41 +72,96 @@
...
+        allowList(grants), readLocations(grants), writeLocations(grants), refreshEndpoint, context);
+  }
+
+  private static boolean allowList(List<LocationGrant> grants) {
+    return grants.stream()
+        .flatMap(g -> g.actions().stream())
+        .anyMatch(a -> a == PolarisStorageActions.LIST || a == PolarisStorageActions.ALL);
+  }
+
+  private static Set<String> readLocations(List<LocationGrant> grants) {
```
dimas-b, line 112: Passing through feedback from artificial helpers (I have not reviewed this finding yet myself 😅 )

> High polaris-core/src/main/java/org/apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java:112:
    readLocations() includes every grant location regardless of requested action. Then polaris-core/src/main/java/org/
    apache/polaris/core/storage/azure/AzureCredentialsStorageIntegration.java:178 turns on SAS read permission whenever
    that set is non-empty. A write-only LocationGrant therefore receives read permission too, which contradicts the new
    action model and the existing test intent at polaris-core/src/test/java/org/apache/polaris/core/storage/azure/
    AzureCredentialStorageIntegrationTest.java:300. For Blob SAS this can over-grant container read access. Filter read
    locations by READ/ALL instead of flattening all grants.

tokoko: That's azure limitation and technically present even in current design. Azure can't put locations with different permissions in a single token.

dimas-b: Correct... I should have reviewed :facepalm:

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
tokoko: because there's already storageConfigSerializedStr which doesn't implement `equals/hashCode`. I should just add equals method and merge them into a single value. individual elements sounds a bit excessive.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
tokoko: not a huge fan either, but couldn't really find a cleaner way to inject, especially in core. my understanding is that the boundary between quarkus and non-quarkus world is at IntegrationProvider so that's the point where injections are allowed to happen. from there on, everything needs to be effectively passed through,

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
dimas-b: Noted (missed that before).

However, having a string representation of the config and the config object itself looks a bit obscure.

Why do you think individual elements are excessive? Immutable objects are pretty efficient in memory. If we limit t he key to the data we truly need, the same amount of data will be "live" on the call stack anyway.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
dimas-b: Yeah... it's a bit convoluted, indeed... not a big deal for now.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsStorageCredentialCacheKey.java
tokoko: it would simply be a lot of fields for aws, roughly 15. but that's not a good reason not to do it of course :laughing: I'll see what I can do.

dimas-b: On the other hand, `AwsStorageConfigurationInfo`  is immutable... why not use it as a "normal" field in the cache key?

tokoko: good catch, switched to use concrete storage-specific classes in all keys.

dimas-b (approved):

dennishuo (approved):
Verified all the previous discussion points match what I had in mind, thanks!

dimas-b:
I propose to merge tomorrow (hoping to not run into merge conflicts :sweat_smile: )

@flyrain : FYI, if you're interested

flyrain (approved):
LGTM. Thanks @tokoko for authoring. Thanks everyone for the review.

PR #3713, opened 2026-02-09
[Iceberg 1.11] Replace deprecated encodeNamespace and decodeNamespace methods

adutra:
## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
@@ -33,6 +33,15 @@ public class NamespaceEntity extends PolarisEntity implements LocationBasedEntit
   // RESTUtil-encoded parent namespace.
   public static final String PARENT_NAMESPACE_KEY = "parent-namespace";
 
+  /**
+   * The URL-encoded UTF-8 representation of the namespace separator character (0x1F).
+   *
+   * <p>This constant mirrors the {@code
+   * org.apache.iceberg.rest.RESTUtil#NAMESPACE_SEPARATOR_URLENCODED_UTF_8} constant which is not
+   * public.
+   */
+  public static final String NAMESPACE_SEPARATOR_URLENCODED_UTF_8 = "%1F";
+
   public NamespaceEntity(PolarisBaseEntity sourceEntity) {
     super(sourceEntity);
     Preconditions.checkState(
@@ -55,7 +64,7 @@ public class NamespaceEntity extends PolarisEntity implements LocationBasedEntit
     if (encodedNamespace == null) {
       return Namespace.empty();
     }
-    return RESTUtil.decodeNamespace(encodedNamespace);
+    return RESTUtil.decodeNamespace(encodedNamespace, NAMESPACE_SEPARATOR_URLENCODED_UTF_8);
   }
 
   public Namespace asNamespace() {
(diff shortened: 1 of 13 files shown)


adutra:
> We should also guard against any change to the separator.

The separator is declared as a constant in `NamespaceEntity` and there is currently no way to specify a different separator. Do you think we need more safeguards?

snazy:
> > We should also guard against any change to the separator.
> 
> The separator is declared as a constant in `NamespaceEntity` and there is currently no way to specify a different separator. Do you think we need more safeguards?

I _think_ there's an Iceberg config property for this, which could leak into Polaris. Just wanna be sure that such situations cannot happen.

adutra:
Rebased to fix conflicts.

adutra:
> I _think_ there's an Iceberg config property for this, which could leak into Polaris. Just wanna be sure that such situations cannot happen.

You are right: there is a new field `RESTCatalogProperties.NAMESPACE_SEPARATOR`. I added this property to the /config endpoint as an override.

polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
```diff
@@ -33,6 +33,15 @@ public class NamespaceEntity extends PolarisEntity implements LocationBasedEntit
   // RESTUtil-encoded parent namespace.
   public static final String PARENT_NAMESPACE_KEY = "parent-namespace";
 
+  /**
+   * The URL-encoded UTF-8 representation of the namespace separator character (0x1F).
+   *
+   * <p>This constant mirrors the {@code
+   * org.apache.iceberg.rest.RESTUtil#NAMESPACE_SEPARATOR_URLENCODED_UTF_8} constant which is not
```
flyrain, line 40: Is `org.apache.iceberg.rest.RESTUtil#NAMESPACE_SEPARATOR_URLENCODED_UTF_8` deprecated in the Iceberg repo as well? If not, we could reuse it.

polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
```diff
@@ -33,6 +33,15 @@ public class NamespaceEntity extends PolarisEntity implements LocationBasedEntit
...
+   * public.
+   */
+  public static final String NAMESPACE_SEPARATOR_URLENCODED_UTF_8 = "%1F";
```
flyrain, line 43: Are we going to support configurable separator in the future?

polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
adutra: It's not deprecated, but unfortunately it's not public either.

polaris-core/src/main/java/org/apache/polaris/core/entity/NamespaceEntity.java
adutra: I personally think it's not possible/desirable.

dimas-b: IIRC, the separator may need to be different only in very specific and rare edge cases. I personally never ran into them in practice. 

Unless someone knows a specific use case where a different separator is needed, let's keep the default one.

PR #3719, opened 2026-02-10
[Iceberg 1.11] feat: add RegisterTable overwrite support; preserve correct auth for overwrites

sririshindra:
Add support for the new overwrite boolean on RegisterTableRequest (default: false) so clients can register a metadata location that replaces an existing table pointer. Implement register-table overwrite semantics:
- If overwrite=false (default): preserve existing behavior 
- attempting to register a table that already exists returns a conflict/error. If overwrite=true:
- If the table does not exist: create it (normal register). 
- If the table exists: update the table's metadata-location to the provided value (do not throw AlreadyExists). 

Ensure safety/backward-compatibility:
- overwrite defaults to false so existing clients are unaffected. Validate provided metadata location where possible before committing to avoid corrupting catalog state.

Details / rationale:
The change implements the REST Catalog spec extension that adds an overwrite flag to RegisterTableRequest. This enables clients to atomically point an existing table identifier at a new metadata file (useful for moving or restoring table metadata). 

Tests added/updated:
- Integration / behavior tests (integration-tests module):
  - Register new table with `overwrite=false` → success.
  - Register existing table with `overwrite=false` → conflict / exception as before.
  - Register existing table with `overwrite=true` → success and the table's metadata-location is updated atomically.
- Authorization tests: updated/added unit tests to assert (TABLE_WRITE_PROPERTIES) is enforced for overwrite against existing tables.

## Checklist
- [+] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [+] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [+] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [+] 💡 Added comments for complex logic
- [-] 🧾 Updated `CHANGELOG.md` (if needed)
- [-] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,8 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Added support for S3-compatible storage that does not have KMS (use `kmsUavailable: true` in catalog storage configuration)
 - Added support for storage-scoped AWS credentials, allowing different AWS credentials to be configured per named storage. Enable with the `RESOLVE_CREDENTIALS_BY_STORAGE_NAME` feature flag (default: false). Storage names can be set explicitly via the `storageName` field on storage configuration, or inferred from the first allowed location's host.
 - Added support for persisting Iceberg metrics (ScanReport, CommitReport) to the database. Enable by setting `polaris.iceberg-metrics.reporting.type=persisting` in configuration. Metrics tables are included in the main JDBC schema.
+- Added support for `register table` overwrite semantics in the Iceberg REST catalog flow (`overwrite=true`) for internal Polaris catalogs. With overwrite enabled, existing table pointers can be updated to a new metadata location while preserving default behavior for `overwrite=false`.
+- Added `REGISTER_TABLE_OVERWRITE` authorization operation mapped to `TABLE_FULL_METADATA` for deterministic overwrite authorization.
 
 ### Changes
 


sririshindra:
cc: @adutra @flyrain

snazy:
Honestly, I think this 'overwrite` flag can be dangerous, because it allows users to change an existing table in a breaking way. The ability to replace an existing table should be guarded by a separate privilege. We should also think about what happens to the existing, old (meta)data files and how changes to the base-location and write-(meta)data properties are handled.

I do not mind adding this to the feature branch, but it would be really good to have a separate discussion about the behavior before this change goes into the `main` branch.

The safest approach IMO would be to initially just error out when `overwrite == true`.

flyrain:
It's an operation users need to be careful with. Adding an addition privilege sounds a good solution to me. In which case, the operation is blocked by default. It is only possible when a special privilege is granted.

sririshindra:
Thanks for the feedback @snazy and @flyrain. I understand the concern about safety.

1. On Auth: 
- I agree TABLE_WRITE_PROPERTIES might be too loose. Would you prefer I restrict this to CATALOG_MANAGE_CONTENT (catalog admins), or should we introduce a specific TABLE_REGISTER_OVERWRITE privilege?

2. On Behavior: 
- Currently, this simply performs a pointer swap. Old metadata/data files are left untouched (not deleted) to prevent accidental data loss. Since the old metadata location is not actually deleted, users can always use that to restore back to the previous state in critical cases. We can log the previous metadata location if necessary for data recovery purposes. 
- Regarding base-location: My current implementation leaves the base-location unchanged (pointing to the original table path). However, if the new metadata file lives in a different location (e.g. during a restore from a backup path), should I update the entity's base-location to match the new metadata's parent directory? 

As a next step, would you prefer I update the PR to throw an error when overwrite=true (disabling it for now) so we can merge the API changes first, or should I implement the stricter auth checks in this PR? My preference would be implement the auth checks in this PR itself, but I am open to suggestions.

flyrain:
> I agree TABLE_WRITE_PROPERTIES might be too loose. Would you prefer I restrict this to CATALOG_MANAGE_CONTENT (catalog admins), or should we introduce a specific TABLE_REGISTER_OVERWRITE privilege?

We will need a new operation in enum `PolarisAuthorizableOperation`, otherwise, external PDP won't work.  For built-in RBAC, I think `TABLE_FULL_METADATA` should be good to use, as it covers both `TABLE_CREATE` and `TABLE_DROP`.

sririshindra:
@flyrain , @snazy Could you please take another look at this PR when you get a chance. I have updated the PR.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +614,80 @@ public LoadTableResponse createTableStaged(
...
+        authorizeCreateTableLikeUnderNamespaceOperationOrThrow(op, identifier);
+      }
+    } else {
+      // Creating new table requires REGISTER_TABLE privilege
+      PolarisAuthorizableOperation op = PolarisAuthorizableOperation.REGISTER_TABLE;
+      authorizeCreateTableLikeUnderNamespaceOperationOrThrow(op, identifier);
+    }
+
+    // Handle Polaris-specific overwrite logic
+    if (overwrite) {
```
nandorKollar, line 608: nit: I think this overwrite can be merged with the previous one.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +614,80 @@ public LoadTableResponse createTableStaged(
...
+      resolutionManifest.addPassthroughPath(tablePath, identifier);
+      resolutionManifest.resolveAll();
+
+      boolean tableExists =
+          resolutionManifest.getResolvedPath(
+                  identifier,
+                  PolarisEntityType.TABLE_LIKE,
+                  PolarisEntitySubType.ICEBERG_TABLE,
+                  true)
+              != null;
```
snazy: Why is all this needed for this change but other functions (loadTable or the table-exists one) don't need it?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +614,80 @@ public LoadTableResponse createTableStaged(
...
+    if (overwrite) {
+      if (baseCatalog instanceof IcebergCatalog icebergCatalog) {
+        // Use the overwrite-capable registration path for IcebergCatalog
+        Table table = icebergCatalog.registerTable(identifier, request.metadataLocation(), true);
+        if (table instanceof BaseTable baseTable) {
+          return LoadTableResponse.builder()
+              .withTableMetadata(baseTable.operations().current())
+              .build();
+        }
+        throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
```
snazy: Looks like duplicated code.

snazy:
I do not mind having this in the feature branch, but I do think the overwrite behavior deserves more thoughts wrt authZ and the behavior when table-attributes like the base-location or the write-(meta)data-locations are different.

flyrain:
The new operation(`REGISTER_TABLE_OVERWRITE(TABLE_FULL_METADATA)`) looks good to me.

The location concern is also valid, I think it should be similar to location checking in `updateTable`, more details could be find here https://github.com/apache/polaris/blob/24c9f238d5945d172321e6580730cef1eefb5f21/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1541.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sririshindra: Good catch, thanks! Addressed in the latest commit — the two if (overwrite) blocks have been merged. Authorization and the catalog call are now handled in a single branch: authorizeRegisterTableOverwriteOrCreate followed by registerTableWithOverwrite, with the non-overwrite path falling straight through to the standard catalogHandlerUtils().registerTable(...).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sririshindra: Fair question — loadTable and tableExists can use the standard helpers (authorizeBasicTableLikeOperationsOrThrow / authorizeCreateTableLikeUnderNamespaceOperationOrThrow) because they always operate against either a known-existing table or always against the namespace. Here the required privilege and auth target depend on whether the table already exists at call time: an existing table needs REGISTER_TABLE_OVERWRITE (TABLE_FULL_METADATA) targeted at the table entity, while a non-existent table falls back to REGISTER_TABLE (TABLE_CREATE) targeted at the namespace. Neither standard helper handles both outcomes. Additionally, baseCatalog.tableExists() can't be called at this point because initializeCatalog() hasn't run yet — it's always invoked at the end of an auth helper. So we resolve the namespace and an optional table path in one pass and branch on the outcome. Added a Javadoc on authorizeRegisterTableOverwriteOrCreate calling this out explicitly.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sririshindra: You're right, there were two separate entity lookups — one via getResolvedPath(identifier.namespace()) for storage validation, and a second via getPassthroughResolvedPath(identifier, ...) to fetch the raw entity. Fixed in the latest commit: overwriteRegisteredTable now does a single getPassthroughResolvedPath upfront and uses that resolved path for both location/storage validation (validateLocationForTableLike, loadFileIOForTableLike) and the entity update.

sririshindra:
> I do not mind having this in the feature branch, but I do think the overwrite behavior deserves more thoughts wrt authZ and the behavior when table-attributes like the base-location or the write-(meta)data-locations are different.

Totally agree this deserves careful thought — a few notes on what this PR covers:

- AuthZ: We settled on REGISTER_TABLE_OVERWRITE mapped to TABLE_FULL_METADATA. The rationale is that replacing a table's metadata pointer is semantically equivalent to simultaneously dropping the old pointer and registering a new one, so it warrants the strictest table-level privilege rather than just TABLE_WRITE_PROPERTIES (which covers normal metadata updates). TABLE_FULL_METADATA also encompasses both TABLE_CREATE and TABLE_DROP individually. Happy to revisit if there's a more appropriate granularity in mind.

- Location changes: The implementation mirrors what doCommit does when the table location changes. It calls StorageUtil.getLocationsUsedByTable(metadata.location(), metadata.properties()) to collect all locations referenced by the new metadata — including write.data.path and write.metadata.path if set — and then runs them through CatalogUtils.validateLocationsForTableLike (checks against the resolved storage config) and validateNoLocationOverlap (checks for conflicts with existing tables). So a metadata file pointing to a different base location or write paths will be validated and rejected if those fall outside the allowed storage configuration.


> The new operation(`REGISTER_TABLE_OVERWRITE(TABLE_FULL_METADATA)`) looks good to me.
> 
> The location concern is also valid, I think it should be similar to location checking in `updateTable`, more details could be find here
> 
> https://github.com/apache/polaris/blob/24c9f238d5945d172321e6580730cef1eefb5f21/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1541
> 
> .

agreed it should mirror updateTable. This is addressed in the latest commit: overwriteRegisteredTable now performs the same three checks that doCommit does when the table location changes:

 1. CatalogUtils.validateLocationsForTableLike — validates all data locations against the resolved storage configuration
 2. validateNoLocationOverlap — ensures the new metadata location doesn't overlap with an existing table
 3. validateMetadataFileInTableDir — ensures the metadata file is within the table's directory structure

The resolved entity used for storage context is now obtained via getPassthroughResolvedPath on the table itself (rather than getResolvedPath on the namespace), which also aligns with how doCommit resolves storage credentials for an existing table.

sririshindra:
@adutra There is a gradle check failing in this PR. I am not sure if that failure has anything do with the changes in this PR. I tried debugging it but I didn't quite get it. Could you help lead me in the right direction please. 
Edit: I realized my branch is a bit behind and missing some commits that might be causing the gradle failure. I rebased and pushed the latest branch. Hopefully that will resolve this issue.  

@snazy @flyrain Could you please take another look at this PR when you get a chance.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,155 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+        metadata.location());
+
+    LOGGER.debug(
+        "Validating metadata.location()={} for identifier={}", metadata.location(), identifier);
+    validateLocationForTableLike(identifier, metadata.location(), resolvedPath);
+    validateMetadataFileInTableDir(identifier, metadata);
+    LOGGER.debug(
+        "Metadata file validated to be within table directory for identifier={}", identifier);
+
+    PolarisResolvedPathWrapper resolvedStorageEntity = resolvedPath;
```
nandorKollar: nit: this is not needed, we can use `resolvedPath`

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -287,25 +287,75 @@ protected Map<String, String> properties() {
...
         lastSlashIndex != -1,
         "Invalid metadata file location; metadata file location must be absolute and contain a '/': %s",
         metadataFileLocation);
 
     // Throw an exception if this table already exists in the catalog.
-    if (tableExists(identifier)) {
+    boolean tableExists = tableExists(identifier);
+    LOGGER.debug("tableExists for identifier {} = {}", identifier, tableExists);
+    if (!overwrite && tableExists) {
+      LOGGER.debug("Table already exists and overwrite is false for identifier={}", identifier);
```
flyrain: Nit: we can  remove the debug log as we throw an exception right after it.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
```diff
@@ -127,6 +128,7 @@ public enum PolarisAuthorizableOperation {
   CREATE_TABLE_STAGED(TABLE_CREATE),
   CREATE_TABLE_STAGED_WITH_WRITE_DELEGATION(EnumSet.of(TABLE_CREATE, TABLE_WRITE_DATA)),
   REGISTER_TABLE(TABLE_CREATE),
+  REGISTER_TABLE_OVERWRITE(TABLE_FULL_METADATA),
```
flyrain, line 131: Given that Polaris doesn't persist the value of this enum, it is probably fine to insert a new one in the middle.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +615,133 @@ public LoadTableResponse createTableStaged(
...
+              tableTarget,
+              null /* secondary */);
+      initializeCatalog();
+      LOGGER.debug(
+          "registerTable: overwrite=true, authorized for REGISTER_TABLE_OVERWRITE on existing table {}",
+          identifier);
+      return;
+    }
+
+    // Table doesn't exist, fall back to standard register-table authorization.
```
nandorKollar: I'm wondering how important this fallback is. If the table already exists, then eventually the caller needs `REGISTER_TABLE_OVERWRITE` role, but it turns out only when they attempted the operation. I'd expect anyone, who calls this endpoint to have `REGISTER_TABLE_OVERWRITE`, regardless of the 'state of the world'. What do you think?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +615,133 @@ public LoadTableResponse createTableStaged(
...
+  }
+
+  private LoadTableResponse registerTableWithOverwrite(
+      TableIdentifier identifier, RegisterTableRequest request) {
+    LOGGER.debug(
+        "registerTableWithOverwrite: identifier={}, metadataLocation={}",
+        identifier,
+        request.metadataLocation());
+    // Handle Polaris-specific overwrite logic
+    if (baseCatalog instanceof IcebergCatalog icebergCatalog) {
```
nandorKollar, line 628: I think this won't work with federated catalog, where `baseCatalog` is a `RESTCatalog`. We should probably document this limitation.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -287,25 +287,75 @@ protected Map<String, String> properties() {
...
+   * metadata file. The overwrite path performs additional validation to ensure the supplied
+   * metadata file and its location are consistent with the table's resolved storage configuration.
+   *
+   * @param identifier the table identifier
+   * @param metadataFileLocation the metadata file location
+   * @param overwrite if true, update existing table metadata; if false, throw exception if table
+   *     exists
+   * @return the registered table
+   */
+  public Table registerTable(
```
nandorKollar, line 313: This method is not part of `Catalog` interface in Iceberg, in fact, there's no such method which corresponds to the register table overwrite semantic. Would it make sense to extend Iceberg Catalog interface with a corresponding method (`registerTable(boolean overwrite)`)?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
flyrain: Agreed with @nandorKollar, I think a more consistent behavior here is still to use operation `REGISTER_TABLE_OVERWRITE`.

flyrain: We can also consolidate the code by 

```
var target = getResolvedPath()
if (target == null) {
    target = resolutionManifest.getResolvedPath(identifier.namespace(), true);
}

 authorizer()
          .authorizeOrThrow(
              polarisPrincipal(),
              resolutionManifest.getAllActivatedCatalogRoleAndPrincipalRoles(),
              PolarisAuthorizableOperation.REGISTER_TABLE_OVERWRITE,
              target,
              null);
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +615,133 @@ public LoadTableResponse createTableStaged(
...
+   * Authorize registerTable with overwrite=true.
+   *
+   * <p>If the table exists, require REGISTER_TABLE_OVERWRITE on the table; otherwise require
+   * REGISTER_TABLE on the parent namespace.
+   *
+   * <p>Resolve both the namespace and an optional passthrough table path in one pass because the
+   * standard helpers either assume the table exists or always authorize against the namespace.
+   * Also, baseCatalog.tableExists() cannot be used here since initializeCatalog() has not run.
+   */
+  private void authorizeRegisterTableOverwriteOrCreate(TableIdentifier identifier) {
```
flyrain: I think we may just call it `authorizeRegisterTableOverwrite`, as register-table itself implies creation.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +615,133 @@ public LoadTableResponse createTableStaged(
...
+  private LoadTableResponse registerTableWithOverwrite(
+      TableIdentifier identifier, RegisterTableRequest request) {
+    LOGGER.debug(
+        "registerTableWithOverwrite: identifier={}, metadataLocation={}",
+        identifier,
+        request.metadataLocation());
+    // Handle Polaris-specific overwrite logic
+    if (baseCatalog instanceof IcebergCatalog icebergCatalog) {
+      LOGGER.debug(
+          "registerTableWithOverwrite: using IcebergCatalog.registerTable for {}", identifier);
```
flyrain: this debug message seems not useful, can we remove it?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +615,133 @@ public LoadTableResponse createTableStaged(
...
+      if (table instanceof BaseTable baseTable) {
+        TableMetadata metadata = baseTable.operations().current();
+        LOGGER.debug(
+            "registerTableWithOverwrite: registered and loaded metadata for {} (metadataLocation={})",
+            identifier,
+            metadata.location());
+        return LoadTableResponse.builder().withTableMetadata(metadata).build();
+      }
+      LOGGER.debug(
+          "registerTableWithOverwrite: unexpected table implementation for {}", identifier);
```
flyrain: Can we remove this debug log?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -612,13 +615,133 @@ public LoadTableResponse createTableStaged(
...
+      throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+    }
+
+    // For non-Polaris catalogs, reject overwrite to prevent accidental overwrites
+    LOGGER.debug(
+        "registerTableWithOverwrite: unsupported baseCatalog type {} for overwrite on {}",
+        baseCatalog.getClass().getName(),
+        identifier);
```
flyrain: Can we remove the debug log as the next line throws?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,155 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+
+    // Refresh TableOperations so the in-memory table reflects the new metadata.
+    LOGGER.debug("Refreshing TableOperations for identifier={}", identifier);
+    TableOperations ops = newTableOps(identifier);
+    ops.refresh();
+    LOGGER.debug("TableOperations refresh completed for identifier={}", identifier);
+    Table result = new BaseTable(ops, fullTableName(name(), identifier), metricsReporter());
+    LOGGER.debug(
+        "Returning BaseTable for identifier={} with tableName={}", identifier, result.name());
+    return result;
```
flyrain:
```suggestion
    return new BaseTable(ops, fullTableName(name(), identifier), metricsReporter());
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,155 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+    } else if (rawEntity.getSubType() == PolarisEntitySubType.GENERIC_TABLE) {
+      LOGGER.debug(
+          "Existing entity is a generic table for identifier={} - cannot overwrite as iceberg table",
+          identifier);
+      throw new AlreadyExistsException(
+          "Generic table with same name already exists: %s", identifier);
+    }
+
+    IcebergTableLikeEntity existingEntity = IcebergTableLikeEntity.of(rawEntity);
+    if (existingEntity == null) {
```
flyrain, line 409: is it possible existingEntity is null?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,155 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+      LOGGER.debug(
+          "Existing entity is a view for identifier={} - cannot overwrite as table", identifier);
+      throw new AlreadyExistsException("View with same name already exists: %s", identifier);
+    } else if (rawEntity.getSubType() == PolarisEntitySubType.GENERIC_TABLE) {
+      LOGGER.debug(
+          "Existing entity is a generic table for identifier={} - cannot overwrite as iceberg table",
+          identifier);
+      throw new AlreadyExistsException(
+          "Generic table with same name already exists: %s", identifier);
+    }
```
flyrain: I think we may check if it is an Iceberg table, otherwise, the logic is fragile when there is a new  `PolarisEntitySubType` introduced, like udf. 
```suggestion
    if (rawEntity.getSubType() != PolarisEntitySubType.ICEBERG_TABLE) {
      throw new AlreadyExistsException(...);
    }
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sririshindra: agreed. I documented this in two places: (1) code-level note + explicit BadRequest message, and (2) unreleased federation docs under operational notes. Current behavior is explicit: overwrite register is supported only for internal Polaris catalogs.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: I agree, I will raise a separate PR in the Iceberg repo for this. As the support for overwrite is added in the rest interface https://github.com/apache/iceberg/pull/15248/changes it is probably a good idea to add the same in the core Iceberg as well.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
sririshindra: I agree with this. I simplified the logic so `overwrite=true` always requires `REGISTER_TABLE_OVERWRITE`; the fallback path was removed. That makes auth deterministic and independent of runtime table-existence state.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: Done

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -287,25 +287,75 @@ protected Map<String, String> properties() {
...
+          identifier,
+          metadataFileLocation);
+      return registerNewTable(identifier, metadataFileLocation, locationDir);
+    }
+
+    LOGGER.debug(
+        "Overwriting registered table for identifier={}, metadataFileLocation={}",
+        identifier,
+        metadataFileLocation);
+    return overwriteRegisteredTable(identifier, metadataFileLocation, locationDir);
```
flyingImer: IIUC, this path comes from overwrite=true and tableExists=true, in such case, would additional authz be required? i.e., privileges on the existing table entity, not just the namespace.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,153 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+        identifier,
+        existingEntity.getId(),
+        metadataFileLocation);
+    updateTableLike(identifier, updatedEntity);
+    LOGGER.debug("updateTableLike succeeded for identifier={}", identifier);
+
+    // Refresh TableOperations so the in-memory table reflects the new metadata.
+    LOGGER.debug("Refreshing TableOperations for identifier={}", identifier);
+    TableOperations ops = newTableOps(identifier);
+    ops.refresh();
```
flyrain: I think this call isn't necessary, the ops.current() will automatically check if refresh is needed. 
https://github.com/apache/polaris/blob/daf62741ec4b63443b7a4519ae92f5879b31adcf/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L653

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,153 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+            .setMetadataLocation(metadataFileLocation)
+            .build();
+
+    LOGGER.debug(
+        "Updating stored entity for identifier={} (entityId={}) with new metadataLocation={}",
+        identifier,
+        existingEntity.getId(),
+        metadataFileLocation);
+    updateTableLike(identifier, updatedEntity);
+    LOGGER.debug("updateTableLike succeeded for identifier={}", identifier);
```
flyrain: I think this debug log is confusing, as part of register with overwrite operation. Can we modify it or remove it?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,153 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+    LOGGER.debug(
+        "Validating metadata.location()={} for identifier={}", metadata.location(), identifier);
+    validateLocationForTableLike(identifier, metadata.location(), resolvedPath);
+    validateMetadataFileInTableDir(identifier, metadata);
+    LOGGER.debug(
+        "Metadata file validated to be within table directory for identifier={}", identifier);
+
+    List<PolarisEntity> resolvedNamespace = resolvedPath.getRawParentPath();
+    Set<String> dataLocations =
+        StorageUtil.getLocationsUsedByTable(metadata.location(), metadata.properties());
```
flyrain: `dataLocations` may indicate they are data file locations only, which is not.
```suggestion
    Set<String> tableLocations = StorageUtil.getLocationsUsedByTable(metadata);
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,153 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+          "Failed to convert rawEntity to IcebergTableLikeEntity for identifier={}", identifier);
+      throw new NoSuchTableException("Table does not exist: %s", identifier);
+    }
+
+    // Build updated entity from parsed metadata and persist the update.
+    // NOTE: This updates the Iceberg table UUID to match the new metadata file's UUID
+    // (via buildTableMetadataPropertiesMap which extracts metadata.uuid()).
+    // The Polaris entity ID remains unchanged to preserve grants/permissions.
+    // See https://lists.apache.org/thread/b5k7vdng904zr3n3q8wv83y8l30rnd4c
+    // and https://lists.apache.org/thread/k3595bttvohb6c3ms36o16gppdfllqmp
```
flyrain: Given that `buildTableMetadataPropertiesMap()` includes the UUID, and two Iceberg discussion threads are not quite related here. I'd suggest to remove these comments. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -330,6 +380,153 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+    TableMetadata metadata = TableMetadataParser.read(fileIO, metadataFileLocation);
+    LOGGER.debug(
+        "Parsed TableMetadata for identifier={} -> uuid={}, location={}",
+        identifier,
+        metadata.uuid(),
+        metadata.location());
+
+    LOGGER.debug(
+        "Validating metadata.location()={} for identifier={}", metadata.location(), identifier);
+    validateLocationForTableLike(identifier, metadata.location(), resolvedPath);
```
flyrain: I think this validation duplicates line 458.  `getLocationsUsedByTable()` will return all locations used by the table including the table `location`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: Published https://github.com/apache/iceberg/pull/15525 in Iceberg repo.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -305,13 +325,22 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
       throw alreadyExistsExceptionForTableLikeEntity(
           identifier, PolarisEntitySubType.ICEBERG_TABLE);
     }
 
     String locationDir = metadataFileLocation.substring(0, lastSlashIndex);
+    if (!tableExists) {
+      return registerNewTable(identifier, metadataFileLocation, locationDir);
+    }
+
+    return overwriteRegisteredTable(identifier, metadataFileLocation, locationDir);
```
flyrain: Minor: I'd prefer the following to make it more readable
```suggestion
    if (tableExists) {
         overwriteRegisteredTable(identifier, metadataFileLocation, locationDir);
    } else {
         registerNewTable(identifier, metadataFileLocation, locationDir);
    }    
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
flyrain: The authZ usually happens in the class `IcebergCatalogHandler`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -336,6 +365,65 @@ public Table registerTable(TableIdentifier identifier, String metadataFileLocati
...
+            resolvedPath,
+            new HashMap<>(tableDefaultProperties),
+            Set.of(PolarisStorageActions.READ, PolarisStorageActions.LIST));
+
+    TableMetadata metadata = TableMetadataParser.read(fileIO, metadataFileLocation);
+    validateMetadataFileInTableDir(identifier, metadata);
+
+    List<PolarisEntity> resolvedNamespace = resolvedPath.getRawParentPath();
+    Set<String> tableLocations =
+        StorageUtil.getLocationsUsedByTable(metadata.location(), metadata.properties());
```
flyrain: Nit: fit into one line
```
var tableLocations = StorageUtil.getLocationsUsedByTable(metadata);
```

sririshindra: @flyrain Addressed it in the latest commit.

sririshindra:
> LGTM. Thanks @sririshindra ! Left a few minor comments.

Thanks @flyrain , I addressed your remaining comments as well in the latest commits.

nandorKollar (approved):
LGTM

flyrain:
Thanks @sririshindra for the change. Thanks @snazy @flyingImer @nandorKollar for the review!

PR #3724, opened 2026-02-10
Require table read/write privilege for metrics reporting

adutra:
This PR introduces a new table privilege: `TABLE_REPORT_METRICS`, which is now required for accessing the metrics reporting endpoint.

**UPDATE**: we now require `TABLE_READ_DATA` or `TABLE_WRITE_DATA`, depending on the report type (scan vs commit respectively).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - The (Before/After)CommitTableEvent has been removed.
 - The `PolarisMetricsReporter.reportMetric()` method signature has been extended to include a `receivedTimestamp` parameter of type `java.time.Instant`.
 - The `ExternalCatalogFactory.createCatalog()` and `createGenericCatalog()` method signatures have been extended to include a `catalogProperties` parameter of type `Map<String, String>` for passing through proxy and timeout settings to federated catalog HTTP clients.
-- Metrics reporting now requires the `TABLE_REPORT_METRICS` privilege on the target table.
+- Metrics reporting now requires the `TABLE_READ_DATA` privilege on the target table for read (scan) metrics and `TABLE_WRITE_DATA` for write (commit) metrics.
 
 ### New Features
 
--- a/site/content/in-dev/unreleased/managing-security/access-control.md
+++ b/site/content/in-dev/unreleased/managing-security/access-control.md
@@ -118,7 +118,6 @@ To grant the full set of privileges (drop, list, read, write, etc.) on an object
 | TABLE_FULL_METADATA | Grants all table privileges, except TABLE_READ_DATA and TABLE_WRITE_DATA, which need to be granted individually. |
 | TABLE_ATTACH_POLICY | Enables attaching policy to a table. |
 | TABLE_DETACH_POLICY | Enables detaching policy from a table. |
-| TABLE_REPORT_METRICS | Enables reporting metrics for a table. |
 
 ### View privileges
 


snazy (dismissed):

jbonofre (approved):

singhpk234 (approved):

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
```diff
@@ -144,7 +145,7 @@ public enum PolarisAuthorizableOperation {
   DROP_VIEW(VIEW_DROP),
   VIEW_EXISTS(VIEW_LIST),
   RENAME_VIEW(VIEW_DROP, EnumSet.of(VIEW_LIST, VIEW_CREATE)),
-  REPORT_METRICS(EnumSet.noneOf(PolarisPrivilege.class)),
+  REPORT_METRICS(TABLE_REPORT_METRICS),
```
flyrain: Do we need to introduce a new privilege? My concern is the new one is probably not necessary, and the way we introduce it will also break existing users using the report-metrics endpoint. 

Given the metrics endpoint is per-table, can we use the similar approach as credential refresh endpoint does, which uses the privilge TABLE_READ_PROPERTIES?
https://github.com/apache/polaris/blob/849039371b59c9ffba088f6ba6a0fffb6d9f58fc/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L729


```suggestion
  REPORT_METRICS(TABLE_READ_PROPERTIES),
```

adutra: I'm fine with that; but we must agree that any client that can read the table can also submit metrics. Is everybody on board with that?

dimas-b: Having a separate permission for metrics seems more flexible and more coherent to me... mainly because the behaviour of this endpoint is very different from the endpoints engaged in reading a table.

That said, considering Scan and Commit Reports (the only kinds supported ATM), it probably makes sense to allow anyone who can query a table to be able submit scan reports.

If I understand correctly, `TABLE_REPORT_METRICS` is implied (subsumed) by `CATALOG_MANAGE_CONTENT`, `TABLE_FULL_METADATA`. This set likely ensures that most table readers are able to submit metrics already.

Metrics Reports are supported only since 1.3.0 (which is a very recent release), so given the above, I would not expect major breakage on the client side due to introducing the new permission.

My preference would be to merge the current PR "as is".

snazy: > My preference would be to merge the current PR "as is".

+100

dimas-b: Note: we could also immediately update 1.3.0 release notes with a warning about the metrics reporting permissions.

collado-mike: I could be persuaded either way, but tbh, I think a new privilege here is excessive. Anybody who is adding rows will end up posting metrics and I think we want those. Read metrics are less useful in my mind, but still useful. With this new privilege, a user needs to have full privileges on the table to report anything. For backward compatibility, I'd suggest at least giving anyone with write privilege the implicit new metrics privilege.

dimas-b: Good point about making write privileges imply `TABLE_REPORT_METRICS` :+1:

adutra: If I follow correctly, it seems we should require `TABLE_READ_DATA` for scan reports, and `TABLE_WRITE_DATA` for commit reports. Is that correct?

jbonofre: Let's leverage the existing privileges. I don't have a strong preference, but using existing privileges makes sense to me. I would go with `TABLE_WRITE_DATA` for metrics reporting.
I think it doesn't hurt and we can move forward quickly.

adutra: OK I implemented this idea:

> we should require TABLE_READ_DATA for scan reports, and TABLE_WRITE_DATA for commit reports

flyrain: Write privilege may be too restrictive as that major use cases of a scan report(https://github.com/apache/polaris/blob/849039371b59c9ffba088f6ba6a0fffb6d9f58fc/spec/iceberg-rest-catalog-open-api.yaml#L3788) metrics happen when engines read a table.

But I agreed that the `CommitReport` needs the write privilege. 
https://github.com/apache/polaris/blob/849039371b59c9ffba088f6ba6a0fffb6d9f58fc/spec/iceberg-rest-catalog-open-api.yaml#L3823

We may have to distinguish them.

adutra: I distinguished privileges based on the report type (scan vs commit).

singhpk234: I may be a bit late and it seems like we have a concencus already i don't wanna disturb that !

+1 to not breaking existing flows !

but i am not sure if i understand the rationale completely, for example it seems like we are treating report as same resource as that of table ? does it have to be this way ?
for example when i am dash boarding and i wanna show reports to all the tables do we need to grant it table per table ? or just one grant and we are able to see it, same for collecting metrics for example R/W. 

I understand report creation is table scoped api, but do we wanna make it table scoped for reads the metrics too. for example in events proposal `/v1/{prefix}/events:` as the endpoint proposed imho

flyrain: In terms of metrics consuming, I don't think it is necessary tied to read/write. For metrics consuming, we could still have table level privilege with the combination of namespace-level and catalog-level consuming privileges, most use cases will be covered.  For example, a user granted catalog-level consuming privilege can see all metrics under the catalog.

adutra:
Thanks all for the reviews!

> We may update the doc like https://polaris.apache.org/in-dev/unreleased/managing-security/access-control/, but it could be done as a followup.

We are not introducing a new privilege anymore, so I think the docs are up to date, wdyt?

PR #3725, opened 2026-02-10
fix: production readiness check doesn't take every credential vendor into account

nandorKollar:
This PR fixes the production readiness check when SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES feature is used. The `checkConnectionCredentialVendors` method doesn't inject every credential vendor, but only a single one, which is the `DefaultPolarisCredentialManager`. When the external catalog authentication feature is used `polaris.features."SUPPORTED_EXTERNAL_CATALOG_AUTHENTICATION_TYPES"=["OAUTH"]` then the production readiness check will fail, as it won't find the selected OAUTH provider.

## Checklist


nandorKollar:
cc @jbonofre

dimas-b:
The fix looks correct and I do not see any adverse effects. Merging now. Of course we can reconsider (or fix-forward) on `main` if any adverse effects are found later (which I do not expect to be the case).

dimas-b:
Thanks for the fix, @nandorKollar !

PR #3732, opened 2026-02-10
Update the generated OpenAPI spec bundle

adutra:
It was out of sync.

The changes in this PR are the result of running:

```
redocly bundle spec/polaris-catalog-service.yaml -o spec/generated/bundled-polaris-catalog-service.yaml
```

## Checklist

### DIFF
--- a/spec/generated/bundled-polaris-catalog-service.yaml
+++ b/spec/generated/bundled-polaris-catalog-service.yaml
@@ -168,7 +168,6 @@ paths:
       parameters:
         - $ref: '#/components/parameters/page-token'
         - $ref: '#/components/parameters/page-size'
-        - $ref: '#/components/parameters/prefix'
         - name: parent
           in: query
           description: An optional namespace, underneath which to list namespaces. If not provided or empty, all top-level namespaces should be listed. If parent is a multipart namespace, the parts must be separated by the unit separator (`0x1F`) byte.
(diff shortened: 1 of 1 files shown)


spec/generated/bundled-polaris-catalog-service.yaml
dimas-b, line 1: Is it used in build? in docs?

dimas-b (approved):

spec/generated/bundled-polaris-catalog-service.yaml
adutra: Docs only, from what I could find.

dimas-b: Would it make sense to move the generated YAML out of `spec/` into something under "docs" perhaps?.. in a separate PR, of course.

adutra: I think it could get even more easily out of sync, don't you think?

dimas-b: I was thinking about generating it at docs build time :thinking:

snazy: I am a huge fan of automating "all the things" and let CI "catch all the things".

As mentioned in my other comment, the process to update the OpenAPI is a very manual process in which people can easily miss those steps.

I've been experimenting with a process to generate the OpenAPI spec from Java types, using the microprofile OpenAPI annotations with an additional check that verifies that the generated OpenAPI-yaml has no breaking changes against a reference OpenAPI-yaml.

Starting from Java types has the advantage that there are no more manual steps needed. Currently, one has to update not just the spec but also a bunch of build scripts for most OpenAPI spec changes, leaking spec specific type names into build scripts.

Thinking further, it is pretty much impossible to make REST endpoints work asynchronously/reactive, because the endpoint functions' signatures are dictated by the code generator.

adutra: > the endpoint functions' signatures are dictated by the code generator.

You could change the method signatures in the mustache templates and make everything async or reactive. We are already customizing almost all templates anyways 😅

PR #3735, opened 2026-02-11
Refactor PolarisConfiguration to provide a non-deprecated method for legacy properties.

dimas-b:
Following up on the discussion from #3694:

* Add non-public `legacyConfig()` for building `PolarisConfiguration` objects with support for legacy property names.

Such properties are deprecated for end users, but Polaris code itself still needs to support them, therefore building `PolarisConfiguration` objects with legacy properties is a mainstream use case.

The new method is not public to prevent external code dependencies since its purpose is to deal with Polaris' own properties.

* Keep the old deprecated `catalogConfigUnsafe()` method for backward compatibility.

It is still used by test code from other packages.

* This is a pure refactoring without behaviour changes.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
@@ -173,7 +173,7 @@ public abstract class PolarisConfiguration<T> {
      * @deprecated Use {@link #legacyCatalogConfig(String)} instead.
      */
     @SuppressWarnings("DeprecatedIsStillUsed")
-    @Deprecated
+    @Deprecated(since = "1.0.0-incubating", forRemoval = true)
     public Builder<T> catalogConfigUnsafe(String catalogConfig) {
       return legacyCatalogConfig(catalogConfig);
     }
--- a/runtime/service/src/test/java/org/apache/polaris/service/config/DefaultConfigurationStoreTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/config/DefaultConfigurationStoreTest.java
@@ -194,7 +194,7 @@ public class DefaultConfigurationStoreTest {
             .description(prefix)
             .buildFeatureConfiguration();
 
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings({"deprecation", "removal"})
     FeatureConfiguration<Boolean> unsafeConfig =
         FeatureConfiguration.<Boolean>builder()
             .key(String.format("%s_unsafe", prefix))
@@ -203,7 +203,7 @@ public class DefaultConfigurationStoreTest {
             .description(prefix)
             .buildFeatureConfiguration();
 
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings({"deprecation", "removal"})
     FeatureConfiguration<Boolean> bothConfig =
         FeatureConfiguration.<Boolean>builder()
             .key(String.format("%s_both", prefix))
--- a/polaris-core/src/test/java/org/apache/polaris/service/storage/PolarisConfigurationStoreTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/service/storage/PolarisConfigurationStoreTest.java
@@ -151,7 +151,7 @@ public class PolarisConfigurationStoreTest {
 
   @Test
   public void testEntityOverrides() {
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings({"deprecation", "removal"})
     Function<Integer, FeatureConfiguration<String>> cfg =
         i ->
             PolarisConfiguration.<String>builder()


polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -128,13 +128,11 @@ public static void enforceFeatureEnabledOrThrow(
...
           .buildFeatureConfiguration();
 
-  @SuppressWarnings(
-      "deprecation") // catalogConfigUnsafe() is deprecated, not the feature config itself
   public static final FeatureConfiguration<Boolean> ALLOW_TABLE_LOCATION_OVERLAP =
       PolarisConfiguration.<Boolean>builder()
           .key("ALLOW_TABLE_LOCATION_OVERLAP")
           .catalogConfig("polaris.config.allow.overlapping.table.location")
-          .catalogConfigUnsafe("allow.overlapping.table.location")
+          .legacyConfig("allow.overlapping.table.location")
```
flyrain: -> `legacyCatalogConfig`? It is specified to catalog level config, I assume.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -174,13 +174,16 @@ public Builder<T> catalogConfig(String catalogConfig) {
     }
 
     /**
-     * Used to support backwards compatibility before there were reserved properties. Usage of this
-     * method should be removed over time.
-     *
-     * @deprecated Use {@link #catalogConfig()} instead.
+     * @deprecated Use {@link #catalogConfig()} or {@link #legacyConfig(String)} instead.
```
flyrain:
```suggestion
     * @deprecated Use {@link #legacyConfig(String)} instead.
```
I think the later is the one to replace `catalogConfigUnsafe()`

runtime/service/src/test/java/org/apache/polaris/service/config/DefaultConfigurationStoreTest.java
flyrain, line 211: Can we create a similar test for the new method `legacyConfig`? Once we removed the deprecated methods, there will be no tests for the new method.

flyrain, line 201: Same here

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
flyrain: Can we clarify this method is deprecated at 1.4.0 and will be removed at 1.6 or 1.7?

dimas-b: Updated javadoc. Added deprecation "for removal".

I do not think we specifically need to flag a future version when it will be removed. In the end, who knows what future version numbers will be, eh? ;)

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
dimas-b: good point - renamed

runtime/service/src/test/java/org/apache/polaris/service/config/DefaultConfigurationStoreTest.java
dimas-b: The diff in this PR is pretty clear that `catalogConfigUnsafe` and `legacyCatalogConfig` work the same way, so existing tests apply to `legacyCatalogConfig` as well.

I tend to agree that in general the testing of config lookup can be improved, I'd prefer to make it in a separate PR though and keep this PR as a "pure" refactoring. WDYT?

dimas-b: discussed in https://github.com/apache/polaris/pull/3735#discussion_r2794315172

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
flyrain: yeah, I think we may figure out a deprecation policy that could be used here. Deprecation without removing does seems a half story. I'm fine with current one.

runtime/service/src/test/java/org/apache/polaris/service/config/DefaultConfigurationStoreTest.java
flyrain: I'm fine with a followup. I'm just a bit concerned that we might forget to port the test whenever we removed the deprecated methods. Not a blocker though.

dimas-b: No worries. I will handle tests on `main` right after we merge this... subject to usual distractions, of course :sweat_smile:

dimas-b:
Rebased to resolve conflicts (which happened automatically - no extra change). @flyrain : PTAL.

runtime/service/src/test/java/org/apache/polaris/service/config/DefaultConfigurationStoreTest.java
dimas-b: #3753

PR #3738, opened 2026-02-11
Use simple READ.md for ArtifactHub

MonkeyCanCode:
As we are publishing our [apache polaris](https://artifacthub.io/packages/helm/apache-polaris/polaris) helm chart on artifact hub, the existed `README.md` has a bit too much context. Based on the [ML](https://lists.apache.org/thread/03kw8or2n69olvvn8y51o5cnt08sctqq) discussion, [apache airflow](https://artifacthub.io/packages/helm/apache-airflow/airflow) has a very clean `README.md` as the front page for artifact hub. Also, looking at [apache spark](https://artifacthub.io/packages/helm/spark-kubernetes-operator/spark-kubernetes-operator), it follows the similar layout as well. This PR contains the following changes:
1. Follow the same layout with two documents by keeping the main `README.md` clean as that is the display page on artifact hub and have it reference to our detail pages from public site.
2. Changed existed `README.md.gotmpl` to `helm.md.gotmpl` to continue support the detail helm page from our official site and updated `Makefile` to reflect this change of path
3. Updated `helm/polaris/.helmignore` as existed packaged helm has a lot more things such as ci and tests which are not necessary for a helm chart
4. Updated `.gitignore` to exclude `helm/polaris/helm.md` as that is temporary file right now to power `Makefile` with our detail helm page update.

## Checklist


helm/polaris/README.md
```diff
@@ -1,468 +1,49 @@
...
-ct lint --charts helm/polaris
-```
+## Features
 
-### Running the integration tests
+* Apache Iceberg REST catalog implementing the Iceberg REST specification
+* Multi-engine interoperability for Spark, Trino, Flink, and other Iceberg-compatible engines
+* Centralized security and governance with principals, roles, and fine-grained privileges
+* Centralized catalog, namespace, and table management
+* Kubernetes-native deployment with stateless services and horizontal scalability
```
adutra: nit: the service is not stateless by default.

Makefile
```diff
@@ -219,8 +219,8 @@ helm-doc-generate: DEPENDENCIES := helm-docs
 .PHONY: helm-doc-generate
 helm-doc-generate: check-dependencies ## Generate Helm chart documentation
 	@echo "--- Generating Helm documentation ---"
-	@helm-docs --chart-search-root=helm
-	@$(PYTHON) helm/polaris/tools/prepare_helm_readme.py helm/polaris/README.md site/content/in-dev/unreleased/helm.md
+	@helm-docs --chart-search-root=helm --template-files helm.md.gotmpl --output-file helm.md
```
adutra: If the `helm.md.gotmpl` is now used only for website docs, why not skip the intermediary step that produces `helm.md`?

We could maybe just do:

```
@helm-docs --chart-search-root=helm --template-files helm.md.gotmpl --output-file site/content/in-dev/unreleased/helm.md
```

helm/polaris/README.md
adutra: Also: you could mention support for Ingress and Gateway API.

helm/polaris/README.md
```diff
@@ -1,468 +1,49 @@
...
+* Production-ready observability with health checks and metrics
+* Open source and vendor neutral, governed by the Apache Software Foundation
```
flyrain: I think it is more accurate to say that the project is governed by PPMC, something like this
```
Open source and vendor neutral, governed by the Apache Polaris PPMC under the Apache Software Foundation
```

helm/polaris/README.md
```diff
@@ -1,468 +1,49 @@
...
+- Kubernetes 1.32+ cluster
+- Helm 3.0+
 
-```bash
-ct lint --charts helm/polaris
-```
+## Features
 
-### Running the integration tests
+* Apache Iceberg REST catalog implementing the Iceberg REST specification
```
flyrain, line 35: Nit: we could also mention that Polaris support other table format, like Delta, Huidi, Lance

Makefile
MonkeyCanCode: So couple things here. The `--output-file` is relative path to the same location as template-files...so technically, we can do `helm-docs --chart-search-root=helm --template-files helm.md.gotmpl --output-file ../../site/content/in-dev/unreleased/helm.md` to get it to the right path. It is a bit ugly but I do agreed that is better then have two steps.
The previous reason on why I kept it two steps was due to we are using those blockquote render hooks for hugo which is not actually valid as go template is used by for helm-doc to render (the `<` in the blockquote render hooks). But we can escape those to tell go to treat them as string literal. It will make the doc a bit harder to read but I also think it is better than depends on another python script for the post-processing and extra file copy.

MonkeyCanCode:
@adutra @dimas-b @flyrain thanks for the quick review. All feedbacks had being addressed and the custom python code for post-processing is no longer needed as well. Please take another look when you get a chance.

flyrain (approved):
Thanks @MonkeyCanCode !

PR #3739, opened 2026-02-11
Update .asf.yaml

jbonofre:
This PR updates the `.asf.yaml` (GitHub configuration) with:
- discussions notification to the dev mailing list
- update on the labels

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -29,6 +29,10 @@ github:
     - polaris
     - iceberg
     - apache
+    - lakehouse
+    - opencatalog
+    - metadata
+    - catalog
 
   enabled_merge_buttons:
     merge: false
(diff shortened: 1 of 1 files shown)


.asf.yaml
adutra, line 40:
```suggestion
    rebase: true
```

jbonofre: Not everyone is happy with `rebase: true`. Personally, I have a preference to have `rebase` enabled, but I'm fine either way.

snazy: I think it's better to have a separate discussion on the ML about GH discussions and GH merge strategy.

jbonofre: Yes, agree. The intent of this PR is about the GH discussions forwarded to the dev mailing list.

PR #3726, opened 2026-02-10
Bump HMS version from 3.1.3 to 4.2.0

nandorKollar:
This PR bumps Hive version to 4.2.0.

## Checklist
- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [ ] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3741, opened 2026-02-11
Revert "Bump HMS version from 3.1.3 to 4.2.0"

jbonofre:
Reverts apache/polaris#3726


jbonofre:
@nandorKollar I propose to focus on fixing the build with HMS 3 first and than prepare a separate effort for HMS 4. Thoughts ?

jbonofre:
@nandorKollar FYI, Iceberg is still using Hive 2, I'm not sure there's any plan to upgrade HiveCatalog in Iceberg. Maybe the Hive community can "host" the HiveCatalog and update it.

nandorKollar:
> @nandorKollar FYI, Iceberg is still using Hive 2, I'm not sure there's any plan to upgrade HiveCatalog in Iceberg. Maybe the Hive community can "host" the HiveCatalog and update it.

Actually it seems that Hive already took over the Iceberg related integration parts, including the HiveCatalog: https://github.com/apache/hive/blob/master/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java

We should eventually use this one, instead of one in Iceberg code. I tried to replace it, but hit an exception, which requires further debugging. I agree with you, probably better to find the problematic transitive dependency for Hive 3, and try to exclude it if possible, and handle Hive 4 in a separate change.

PR #3748, opened 2026-02-12
Optimize configuration value casts (#3688)

### LINKED ISSUE #3688: Optimize configuration value casts
See https://github.com/apache/polaris/pull/3573#discussion_r2764052463

> Looks like this function converts every value to a string and then parses that string even if the type matches, which feels overly expensive.

AryanPatel226:
Optimizes `RealmConfigImpl.tryCast()` to avoid unnecessary string conversions when configuration values are already the correct type.

The method was converting every value to string and parsing it back, even when the value already matched the expected type (Boolean, Integer, Long, Double). This adds type checks before string conversion to eliminate unnecessary overhead.

Fixes #3688

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #3688


polaris-core/src/main/java/org/apache/polaris/core/config/RealmConfigImpl.java
```diff
@@ -106,12 +106,24 @@ public <T> T getConfig(PolarisConfiguration<T> config, Map<String, String> catal
     }
 
     if (config.defaultValue() instanceof Boolean) {
+      if (value instanceof Boolean) {
```
dimas-b: Could this be done in a more general way? 🤔

AryanPatel226: Hey @dimas-b I created a [PR](https://github.com/apache/polaris/pull/3805) to solve this in a general way.

PR #3573, opened 2026-01-27
Introduce RealmConfigurationSource

dimas-b:
The existing `PolarisConfigurationStore` interface is a mixture of SPI methods providing config values from the environment and utility methods for various lookup parameter permutations and type casts.

This change adds `RealmConfigurationSource` for the SPI part and moves lookup logic into `RealmConfigImpl`.

Note that existing service code always accesses config via `RealmConfig`.

The old `PolarisConfigurationStore` interface remains for backward compatibility and but redirects actual config lookup to `RealmConfigImpl`.

## Checklist
- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3750, opened 2026-02-12
Improve Access Delegation Mode Selection Algorithm

### LINKED ISSUE #3090: Improve access delegation mode selection algorithm

#2280 is bringing support for S3 request signing.

It introduces an access delegation mode selection algorithm that is roughly like this:

1. If no delegation mode is requested, use `UNKNOWN`
2. If one single delegation mode is requested, use that mode
3. If requested modes include both `VENDED_CREDENTIALS` and `REMOTE_SIGNING`
    a. If credentials subscoping is enabled for the catalog, use `VENDED_CREDENTIALS`
    b. Otherwise, use `REMOTE_SIGNING`
4. Otherwise, throw an error "unsupported mode(s)"

This algorithm aims at being smart but also fast to execute. 

But it may select sub-optimal modes. For example, it doesn't check whether STS is available, because that requires fetching the `AwsStorageConfigurationInfo` for the catalog. So in some cases it may select `VENDED_CREDENTIALS` while `REMOTE_SIGNING` would be a better choice.

_No response_

_No response_

### Additional context

_No response_

iting0321:
## **Summary**
This PR introduces an improved algorithm for selecting the optimal `AccessDelegationMode` when clients request both `VENDED_CREDENTIALS` and `REMOTE_SIGNING`. The new implementation considers actual catalog capabilities (STS availability) rather than using simple heuristics.

## **Problem**
Previously, when a client requested both `VENDED_CREDENTIALS` and `REMOTE_SIGNING` modes via the `X-Iceberg-Access-Delegation` header, Polaris would always prefer `VENDED_CREDENTIALS`. 
However, this could lead to failures when:

- STS is unavailable for the catalog's storage configuration
- Credential subscoping is disabled via SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION configuration

In these cases, `REMOTE_SIGNING` would be the better choice, but the original algorithm had no way to make this determination.

## **Solution**
Introduced `AccessDelegationModeResolver` - a dedicated component that intelligently selects the optimal access delegation mode based on:

- Client request - The modes specified in `X-Iceberg-Access-Delegation` header
- STS availability - Checks `AwsStorageConfigurationInfo.getStsUnavailable()` flag
- Credential subscoping configuration - Respects `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION` setting

## **How to Test**
```
./gradlew :polaris-runtime-service:test --tests "AccessDelegationModeResolverTest"
```
## **Related Issues**
 #3090 - Improve access delegation mode selection algorithm

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #3090 
- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how) - Added `AccessDelegationModeResolverTest` with 15 unit tests

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/AccessDelegationModeResolverTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/AccessDelegationModeResolverTest.java
@@ -31,7 +31,6 @@ import java.util.Optional;
 import org.apache.polaris.core.admin.model.Catalog;
 import org.apache.polaris.core.config.FeatureConfiguration;
 import org.apache.polaris.core.config.RealmConfig;
-import org.apache.polaris.core.context.CallContext;
 import org.apache.polaris.core.entity.CatalogEntity;
 import org.apache.polaris.core.entity.PolarisEntityConstants;
 import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
@@ -53,14 +52,12 @@ import org.mockito.quality.Strictness;
 class AccessDelegationModeResolverTest {
 
   @Mock private RealmConfig realmConfig;
-  @Mock private CallContext callContext;
 
   private AccessDelegationModeResolver resolver;
 
   @BeforeEach
   void setUp() {
-    when(callContext.getRealmConfig()).thenReturn(realmConfig);
-    resolver = new DefaultAccessDelegationModeResolver(callContext);
+    resolver = new DefaultAccessDelegationModeResolver(realmConfig);
   }
 
   /** Helper to set up config mock for tests that need it */


runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,222 @@
...
+ * </ol>
+ *
+ * <p>This resolver improves upon the simple mode selection by considering:
+ *
+ * <ul>
+ *   <li>STS availability from the catalog's {@link AwsStorageConfigurationInfo}
+ *   <li>The {@link FeatureConfiguration#SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION} setting
+ * </ul>
+ */
+public class AccessDelegationModeResolver {
```
adutra: Should this be turned into an interface + CDI bean?

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,222 @@
...
+   * Determines if STS is available for the catalog's storage configuration.
+   *
+   * <p>For AWS storage, this checks the {@link AwsStorageConfigurationInfo#getStsUnavailable()}
+   * flag. For other storage types, STS is considered available by default (since they may have
+   * their own credential vending mechanisms).
+   *
+   * @param catalogEntity The catalog entity to check
+   * @return true if STS is available or the storage type doesn't require STS
+   */
+  private boolean isStsAvailable(@Nonnull CatalogEntity catalogEntity) {
```
adutra: Suggestion: rename this method to something more general: `isCredentialVendingAvailable`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * Convenience method to resolve and return an EnumSet containing just the resolved mode.
+   *
+   * @param requestedModes The set of delegation modes requested by the client
+   * @param catalogEntity The catalog entity
+   * @return An EnumSet containing the single resolved mode
+   */
+  @Nonnull
+  public EnumSet<AccessDelegationMode> resolveToSet(
```
adutra: What's the point of this method? Once the mode is resolved, we should use the resolved mode only.

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,222 @@
...
+      LOGGER.debug(
+          "No catalog entity available for mode resolution, defaulting to VENDED_CREDENTIALS");
+      return VENDED_CREDENTIALS;
+    }
+
+    // Check if credential subscoping is skipped - if so, VENDED_CREDENTIALS won't work properly
+    boolean skipCredentialSubscoping =
+        Boolean.TRUE.equals(
+            realmConfig.getConfig(
+                FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION, catalogEntity));
```
adutra: 1. Contrary to the javadocs, this method cannot return null since there is a default value (false).
2. The `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION` is not overridable at catalog level.
```suggestion
      boolean skipCredentialSubscoping =
          realmConfig.getConfig(FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,222 @@
...
+
+    if (skipCredentialSubscoping) {
+      LOGGER.debug(
+          "Credential subscoping is skipped for catalog {}, selecting REMOTE_SIGNING",
+          catalogEntity.getName());
```
adutra:
```suggestion
          "Credential subscoping is skipped for this realm, selecting REMOTE_SIGNING");
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,222 @@
...
+      return REMOTE_SIGNING;
+    }
+
```
adutra: We are missing one important check here:

```java
      boolean credentialSubscopingAuthorized =
          baseCatalog instanceof IcebergCatalog
              || realmConfig.getConfig(
                  ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING, catalogEntity);

      if (!credentialSubscopingAuthorized) {
        LOGGER.debug(
            "Credential subscoping is not authorized for catalog {}, selecting REMOTE_SIGNING",
            catalogEntity.getName());
        return REMOTE_SIGNING;
      }
```

dimas-b:
Hi @iting0321 : 

> [...] In these cases, REMOTE_SIGNING would be the better choice [...]

I'm not sure Polaris supports remote signing yet :thinking: 

@adutra : WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1307,6 +1317,34 @@ private void checkAllowExternalCatalogCredentialVending(
...
+   * @param requestedModes The set of delegation modes requested by the client
+   * @return The resolved set of delegation modes (containing the single optimal mode)
+   */
+  protected EnumSet<AccessDelegationMode> resolveAccessDelegationModes(
+      EnumSet<AccessDelegationMode> requestedModes) {
+    if (requestedModes.isEmpty()) {
+      return requestedModes;
+    }
+
+    AccessDelegationModeResolver resolver = new AccessDelegationModeResolver(realmConfig());
```
adutra: 1. The resolver itself could be exposed as a method like this:

```java
  @Value.Lazy
  protected AccessDelegationModeResolver accessDelegationModeResolver() {
    return new AccessDelegationModeResolver(realmConfig());
  }
```

2. I would suggest to store the result of the access delegation mode resolution as a field in this class for later retrieval – similar to what we do with the resolution manifest. It's not necessary to pass the access mode around as a parameter in each method.

adutra:
> I'm not sure Polaris supports remote signing yet 🤔
> 
> @adutra : WDYT?

Polaris does not support remote signing yet indeed, but the changes in this PR go in the right direction, and imho, can be adopted _before_ remote signing lands.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: Note: if the resolver becomes a CDI bean, then my suggestion above needs to be changed to:

```java
protected abstract AccessDelegationModeResolver accessDelegationModeResolver();
```

And the `IcebergCatalogHandlerFactory` would have to be modified accordingly to include this new component.

dimas-b:
Improving access delegation mode selection before remote signing is fully supports sounds reasonable to me.

Please consider this check, though: 

https://github.com/apache/polaris/blob/579b271b8c41cd26339bccfd771c1a591ad27039/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java#L258-L261

That is to say, it might be best to do this validation only after the final mode resolution :thinking:

adutra:
> That is to say, it might be best to do this validation only after the final mode resolution 🤔

I think this check is fine; as you noted, we don't support anything else than `VENDED_CREDENTIALS` for the time being.

I see @iting0321's work in this PR as a way to proactively prepare Polaris for remote signing – without actually making remote signing possible.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1332,17 +1341,28 @@ private void checkAllowExternalCatalogCredentialVending(
...
    * @param requestedModes The set of delegation modes requested by the client
-   * @return The resolved set of delegation modes (containing the single optimal mode)
+   * @return The resolved access delegation mode
    */
-  protected EnumSet<AccessDelegationMode> resolveAccessDelegationModes(
+  protected AccessDelegationMode resolveAccessDelegationModes(
       EnumSet<AccessDelegationMode> requestedModes) {
     if (requestedModes.isEmpty()) {
-      return requestedModes;
+      resolvedAccessDelegationMode = AccessDelegationMode.UNKNOWN;
```
adutra: Not for this PR, but I've always thought we should rename this mode to `NONE`, wdyt?

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
adutra: I think the rest of the javadocs starting from this line are rather for the implementation, not for the interface.

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
iting0321: Yes, I agree. I turned `AccessDelegationModeResolver` into interface and added `DefaultAccessDelegationModeResolver.java` for CDI bean implementation.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -253,13 +253,9 @@ public Response updateProperties(
...
-    EnumSet<AccessDelegationMode> delegationModes =
-        AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
-    Preconditions.checkArgument(
-        delegationModes.isEmpty() || delegationModes.contains(VENDED_CREDENTIALS),
-        "Unsupported access delegation mode: %s",
-        accessDelegationMode);
-    return delegationModes;
+    // Parse the modes from the header - validation will happen after mode resolution
+    // in IcebergCatalogHandler.resolveAccessDelegationModes()
+    return AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
```
adutra, line 249: We need to reject `REMOTE_SIGNING` explicitly until it's available, otherwise we could introduce regressions, wdyt?

```suggestion
    var accessDelegationMode = AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
    // TODO remove when remote signing is implemented
    Preconditions.checkArgument(
        accessDelegationMode != REMOTE_SIGNING,
        "Unsupported access delegation mode: %s",
        accessDelegationMode);
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,203 @@
...
+    boolean skipCredentialSubscoping =
+        configurationStore.getConfiguration(
+            realmContext, FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
+
+    if (skipCredentialSubscoping) {
+      LOGGER.debug(
+          "Credential subscoping is skipped for this realm, selecting REMOTE_SIGNING");
+      return REMOTE_SIGNING;
+    }
+
```
adutra, line 170: We are still missing this check:

https://github.com/apache/polaris/pull/3750#discussion_r2799598375

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
iting0321: Thanks for your  feedback ! 
I removed `resolveToSet()`. The resolver now returns a single `AccessDelegationMode`, and the resolved mode is used directly (not wrapped in an EnumSet)

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
iting0321: Thanks for your feedback. I just removed it.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
iting0321: Yes, I think we should reject `REMOTE_SIGNING` explicitly for now.
Since the resolver may accept it but the feature isn’t implemented, clients could see unexpected behavior. Rejecting unsupported modes in the adapter keeps the flow clean and avoids failures later on.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
iting0321: Sorry, I missed that earlier. I just pushed a commit adding the missing check and the corresponding unit test in ```AccessDelegationModeResolverTest.java```
Thanks for pointing it out!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
iting0321: I think NONE is more precise, since it clearly communicates that no delegation mode was requested or selected. I’m happy to add it in a follow-up PR.

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,68 @@
...
+ * <p>The selection algorithm is:
+ *
+ * <ol>
+ *   <li>If no delegation mode is requested, use {@link AccessDelegationMode#UNKNOWN}
+ *   <li>If exactly one delegation mode is requested, use that mode (if supported)
+ *   <li>If the requested modes include both {@link AccessDelegationMode#VENDED_CREDENTIALS} and
+ *       {@link AccessDelegationMode#REMOTE_SIGNING}:
+ *       <ol>
+ *         <li>Check if STS is available for the catalog's storage configuration
+ *         <li>If STS is available and credential subscoping is not skipped, use {@link
```
singhpk234: > credential subscoping is not skipped

is this a feature flag ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -243,13 +244,16 @@ public Response updateProperties(
   }
 
   private EnumSet<AccessDelegationMode> parseAccessDelegationModes(String accessDelegationMode) {
-    EnumSet<AccessDelegationMode> delegationModes =
+    // Parse the modes from the header - validation will happen after mode resolution
```
dimas-b: nit: `header` - nothing in this case binds to any "headers"... The first part of this comment looks a bit confusing to me.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -243,13 +244,16 @@ public Response updateProperties(
...
+    // in IcebergCatalogHandler.resolveAccessDelegationModes()
+    EnumSet<AccessDelegationMode> modes =
         AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
+    // TODO remove when remote signing is implemented
     Preconditions.checkArgument(
-        delegationModes.isEmpty() || delegationModes.contains(VENDED_CREDENTIALS),
+        !modes.contains(REMOTE_SIGNING),
```
dimas-b: This assertion might be too strict. When Polaris supports `VENDED_CREDENTIALS` but does not support `REMOTE_SIGNING`, I do not see any hard in the client requesting both, as long as the resolution algorithm selects the supported more (i.e. `VENDED_CREDENTIALS` in the current state of the code).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1327,6 +1346,39 @@ private void checkAllowExternalCatalogCredentialVending(
...
+    return resolvedAccessDelegationMode;
+  }
+
+  /**
+   * Returns the resolved access delegation mode. Must be called after
+   * {@link #resolveAccessDelegationModes(EnumSet)} has been invoked.
+   *
+   * @return The resolved access delegation mode, or null if not yet resolved
+   */
+  protected AccessDelegationMode getResolvedAccessDelegationMode() {
```
dimas-b: This method is not used. Is it worth adding?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: The original idea behind `AccessDelegationMode.UNKNOWN` was for it to represent cases when the client declared something unparseable. Cf. `AccessDelegationMode.fromProtocolValuesList()`

If we want to definitely resolve to one specific `AccessDelegationMode`  why don't we feed raw protocol strings to `AccessDelegationModeResolver` instead?

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,220 @@
...
+    // Case 3: Both VENDED_CREDENTIALS and REMOTE_SIGNING requested
+    if (effectiveModes.contains(VENDED_CREDENTIALS) && effectiveModes.contains(REMOTE_SIGNING)) {
+      return resolveVendedCredentialsVsRemoteSigning(catalogEntity);
+    }
+
+    // Case 4: Unknown combination - default to VENDED_CREDENTIALS for backward compatibility
+    LOGGER.warn(
+        "Unknown access delegation mode combination: {}, defaulting to VENDED_CREDENTIALS",
+        requestedModes);
+    return VENDED_CREDENTIALS;
```
dimas-b: I'm not sure about defaulting to `VENDED_CREDENTIALS`. The client may have a legitimate reason not to ask for `VENDED_CREDENTIALS` (e.g. due to risk of credential exposure), so forcing sensitive data into the response may not be advisable.

If we want to do this, I'd propose to add a feature flag (with default "off").

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,220 @@
...
+          "No catalog entity available for mode resolution, defaulting to VENDED_CREDENTIALS");
+      return VENDED_CREDENTIALS;
+    }
+
+    // Check if credential subscoping is authorized for this catalog.
+    // For internal catalogs, credential subscoping is always authorized.
+    // For external/federated catalogs, check if ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING is enabled.
+    boolean credentialSubscopingAuthorized =
+        !catalogEntity.isExternal()
+            || configurationStore.getConfiguration(
```
dimas-b: Please use `RealmConfig` (cf. #3573)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1327,6 +1346,39 @@ private void checkAllowExternalCatalogCredentialVending(
...
+   */
+  protected AccessDelegationMode resolveAccessDelegationModes(
+      EnumSet<AccessDelegationMode> requestedModes) {
+    if (requestedModes.isEmpty()) {
+      resolvedAccessDelegationMode = AccessDelegationMode.UNKNOWN;
+      return resolvedAccessDelegationMode;
+    }
+
+    CatalogEntity catalogEntity = getResolvedCatalogEntity();
+    resolvedAccessDelegationMode = accessDelegationModeResolver().resolve(requestedModes, catalogEntity);
```
dimas-b: Is the `resolvedAccessDelegationMode` field really necessary? It does not appear to be read at all (only written). Do we expect more that one "resolve" call per API request?

dimas-b (commented):
Thanks for working on this, @iting0321 !

I agree with @adutra that is makes sense to add a resolver for access delegation modes. This PR seems to be moving in the right direction.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
iting0321: I agree that is misleading. Change it to ```// Parse the access delegation modes - validation will happen after mode resolution```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
iting0321: I moved the validation from the parsing stage to the resolution stage. Now we only reject if `REMOTE_SIGNING` is the resolved/selected mode, not just if it's requested.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
iting0321: Each API endpoint only calls `resolveAccessDelegationModes()` once 
per request, so storing the result in an instance field provides no benefit. 
I've removed the field and the unused getter method.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
iting0321: I just removed it.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
iting0321: I’ve removed the fallback to `VENDED_CREDENTIALS` when catalogEntity is null  and now I throw the error 
```java
    // Case 4: Unknown combination - reject to prevent unintended credential exposure
    throw new IllegalArgumentException(
        "Unsupported access delegation mode combination: " + requestedModes);
  }
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
iting0321: I just replaced all of them to `RealmConfig`.

dimas-b:
@iting0321 : Please resolve conflicts. Thx!

iting0321:
Hi @dimas-b, I just fixed the conflicts.
Thanks !

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -598,15 +605,19 @@ public LoadTableResponse createTableStaged(
       EnumSet<AccessDelegationMode> delegationModes,
       Optional<String> refreshCredentialsEndpoint) {
 
+    // Authorize first with original delegation modes to initialize resolution manifest
     authorizeCreateTableStaged(namespace, request, delegationModes);
 
+    // Resolve the optimal delegation mode based on catalog capabilities (after authorization)
+    AccessDelegationMode resolvedMode = resolveAccessDelegationModes(delegationModes);
```
dimas-b: I believe `resolvedMode` should go into `authorizeCreateTableStaged` (line 609) and trickle down to `checkAllowExternalCatalogCredentialVending(...)`

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,213 @@
...
+      LOGGER.debug(
+          "No catalog entity available for mode resolution, defaulting to VENDED_CREDENTIALS");
+      return VENDED_CREDENTIALS;
+    }
+
+    // Check if credential subscoping is authorized for this catalog.
+    // For internal catalogs, credential subscoping is always authorized.
+    // For external/federated catalogs, check if ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING is
+    // enabled.
+    boolean credentialSubscopingAuthorized =
```
dimas-b: nit: it's not "authorized" but rather "enabled" by this config.

dimas-b: also, the config check is not about "subscoping" but `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING`

dimas-b (commented):
I think this PR is approaching a mergeable state... @adutra : WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
iting0321: 1. Changed the comment to
 ```   
 // Check if credential vending is enabled for this catalog.
 // For internal catalogs, credential vending is always enabled.
```
2. Changed ```credentialSubscopingAuthorized``` to ```credentialVendingEnabled```

File: runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java (hunk 18 above)
Comment (iting0321) [reply runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java]: Thanks for catching this! I've changed `checkAllowExternalCatalogCredentialVending` to use 
`resolvedMode`.

I kept it as a separate call after authorization though the check must happen 
**AFTER** authorization because `resolveAccessDelegationModes()` calls 
`getResolvedCatalogEntity()`, which requires the resolution manifest to be 
initialized by the authorization step first.

Updated all three methods to follow this pattern.

**Comment by iting0321** [comment]
Hi @dimas-b, I have fixed the all the above comments.
Please take a look when you have moment.
Thank you!

File: runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
Comment (dimas-b) [inline runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java]: This error is supposed to happen only for vended credentials, but now it appears to be possible with request signing... WDYT?

Comment (iting0321) [reply runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java]: Oops, I think this function is for both mode.
I just changed the function name from `checkAllowExternalCatalogCredentialVending` to  `checkAllowExternalCatalogAccessDelegation` and made the error message more precise.

File: runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java [hunk 20]
```diff
@@ -246,13 +244,9 @@ public Response updateProperties(
   }
 
   private EnumSet<AccessDelegationMode> parseAccessDelegationModes(String accessDelegationMode) {
-    EnumSet<AccessDelegationMode> delegationModes =
-        AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
-    Preconditions.checkArgument(
-        delegationModes.isEmpty() || delegationModes.contains(VENDED_CREDENTIALS),
```
dimas-b, line 252: It looks like if `accessDelegationMode` contains only invalid values (represented by `UNKNOWN`) we used to fail the request before, but now the `UNKNOWN` delegation mode is carried forward (and will eventually be accepted).

I believe we have to keep old behaviour where requests with invalid access modes failed.

If a request contains some invalid access modes (`UNKNOWN`) and some value, I tend to think we can remove the `UNKNOWN` entry from the `EnumSet` here and carry on.

@adutra WDYT?

adutra: Indeed I think that if the set contains only `UNKNOWN` we should throw some error, probably here.

As for removing the `UNKNOWN` entry from the set, this is already being done in `DefaultAccessDelegationModeResolver` from what I see, so I don't think we need to remove it here.

adutra: I see that `DefaultAccessDelegationModeResolver` also handles the case where the set contains only `UNKNOWN` – maybe we should throw the error there instead.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,213 @@
...
+      return UNKNOWN;
+    }
+
+    // Filter out UNKNOWN mode from consideration for selection logic
+    EnumSet<AccessDelegationMode> effectiveModes = EnumSet.copyOf(requestedModes);
+    effectiveModes.remove(UNKNOWN);
+
+    if (effectiveModes.isEmpty()) {
+      LOGGER.debug("Only UNKNOWN mode requested, returning UNKNOWN");
+      return UNKNOWN;
```
adutra: As said in my previous comments: maybe this should instead throw some error.

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
adutra: The javadocs are still here, FYI.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1293,17 +1320,16 @@ private EnumSet<PolarisAuthorizableOperation> getUpdateTableAuthorizableOperatio
     }
   }
 
-  private void checkAllowExternalCatalogCredentialVending(
-      EnumSet<AccessDelegationMode> delegationModes) {
+  private void checkAllowExternalCatalogAccessDelegation(AccessDelegationMode resolvedMode) {
 
-    if (delegationModes.isEmpty()) {
+    if (resolvedMode == AccessDelegationMode.UNKNOWN) {
```
adutra: This is apparently impossible with the default resolver. But I think it doesn't hurt to keep this `if` block because other implementations may choose to return `UNKNOWN`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
dimas-b: That works too - my main point is to preserve client-visible behaviour with invalid access mode values.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,213 @@
...
+    // If no catalog entity available, default to VENDED_CREDENTIALS for backward compatibility
+    if (catalogEntity == null) {
+      LOGGER.debug(
+          "No catalog entity available for mode resolution, defaulting to VENDED_CREDENTIALS");
+      return VENDED_CREDENTIALS;
+    }
+
+    // Check if credential vending is enabled for this catalog.
+    // For internal catalogs, credential vending is always enabled.
+    // For external/federated catalogs, check if ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING is
```
flyingImer, line 146: I'm a bit confused here, thought `ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING` is meant for external catalogs while `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING` is for federated catalogs

Resolver picks VENDED_CREDENTIALS (because `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING`=true) but the handler's subsequent check rejects it (if `ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING`=false), seems one of them to be redundant

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1313,13 +1339,53 @@ private void checkAllowExternalCatalogCredentialVending(
...
+   * capabilities.
+   *
+   * @param requestedModes The set of delegation modes requested by the client
+   * @return The resolved access delegation mode
+   */
+  protected AccessDelegationMode resolveAccessDelegationModes(
+      EnumSet<AccessDelegationMode> requestedModes) {
+    if (requestedModes.isEmpty()) {
+      return AccessDelegationMode.UNKNOWN;
+    }
```
flyingImer: nit: do we need to duplicate this logic here as well? where IcebergCatalogHandler already covers it

runtime/service/src/test/java/org/apache/polaris/service/catalog/AccessDelegationModeResolverTest.java
```diff
@@ -0,0 +1,353 @@
...
+            org.mockito.ArgumentMatchers.<CatalogEntity>any()))
+        .thenReturn(allowFederatedCredentialVending);
+
+    // Mock SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION (realm-level only)
+    when(realmConfig.getConfig(org.mockito.ArgumentMatchers.<PolarisConfiguration<Boolean>>any()))
+        .thenReturn(skipCredentialSubscoping);
+  }
+
+  @Test
+  void resolveEmptyModes_returnsUnknown() {
```
flyingImer: is it the same as `resolve_emptyModes_returnsUnknown`?

runtime/service/src/test/java/org/apache/polaris/service/catalog/AccessDelegationModeResolverTest.java
```diff
@@ -0,0 +1,353 @@
...
+import org.junit.jupiter.params.provider.EnumSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@SuppressWarnings("unchecked")
+class AccessDelegationModeResolverTest {
```
flyingImer, line 52: I believe we missed a negative test for `IllegalArgumentException`

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
adutra: See related dev ML thread: https://lists.apache.org/thread/c4ywj2jh9lxn11nmv6rr9scrqc77p1rx

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
adutra: I think this should throw.

runtime/service/src/test/java/org/apache/polaris/service/catalog/AccessDelegationModeResolverTest.java
iting0321: Added!

runtime/service/src/test/java/org/apache/polaris/service/catalog/AccessDelegationModeResolverTest.java
iting0321: Deleted!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
iting0321: Now I fixed it to 
```java
Preconditions.checkArgument(
        !requestedModes.isEmpty(),
        "requestedModes must not be empty; callers should skip resolution when no delegation is needed");
```
When the requestedModes is {}, it will throw the error `IllegalArgumentException`

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
iting0321: I just remove the `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING` check from the resolver.
The authorization gate for external catalogs remains in `IcebergCatalogHandler.checkAllowExternalCatalogAccessDelegation()`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
iting0321: Deleted. Sorry for the miss.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -129,22 +125,6 @@ private AccessDelegationMode resolveVendedCredentialsVsRemoteSigning(
       return VENDED_CREDENTIALS;
     }
 
-    // Check if credential vending is enabled for this catalog.
```
adutra: Why was this check removed?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -478,10 +478,12 @@ public LoadTableResponse createTableDirect(
     authorizeCreateTableDirect(namespace, request, delegationModes);
 
     // Resolve the optimal delegation mode based on catalog capabilities (after authorization)
-    AccessDelegationMode resolvedMode = resolveAccessDelegationModes(delegationModes);
-
-    // Check if external catalog access delegation is allowed for the resolved mode
-    checkAllowExternalCatalogAccessDelegation(resolvedMode);
+    AccessDelegationMode resolvedMode = AccessDelegationMode.UNKNOWN;
```
adutra: I suppose this new change is a consequence of my suggestion to throw an exception from the resolver rather than returning `UNKNOWN`.

Unfortunately, I think this is becoming a bit too verbose. I apologize for having made that suggestion. Let me think about this some more.

adutra:
@iting0321 and @dimas-b so I tried a few things and thought a bit more about access delegation mode resolution. Here are my findings:

1) There is a chicken-and-egg problem that needs attention: authorization requires access mode resolution, but access mode resolution requires catalog entity resolution, but catalog entity resolution requires authorization. I have a proposal below to fix this properly.
2) I think the `UNKNOWN` access mode is getting in the way. I propose to drop unparseable modes and log a warning instead. Keeping this constant around is not useful for business logic, but complicates implementation.
3) I think that `AccessDelegationModeResolver.resolve()` should actually return `Optional<AccessDelegationMode>`. Returning empty clearly states the fact that no actionable access delegation mode could be extracted from the request, and for all intents and purposes, Polaris should assume that no access delegation should happen.

With that said, I think the general order of method invocations should be:

1. Call the relevant `authorizeXYZ()` method with all the parsed modes (`EnumSet<AccessDelegationMode>`): without the `UNKNOWN` constant, we can now be sure that if the set is empty, then no access delegation should happen. **This information is enough for `authorize()` methods**: they don't need to know the actual resolved delegation mode, they only need to know if delegation should happen or not. **This solves the chicken and egg problem.**
2. Resolve the access delegation mode by calling `AccessDelegationModeResolver.resolve()`.
3. Use the returned `Optional<AccessDelegationMode>` for access delegation checks and for response builders, since these components must know what the exact resolved mode is.

I implemented the above ideas in this commit: feel free to use or cherry-pick it:

https://github.com/adutra/polaris/commit/915b5d98ad38c0b1b7e1c191f4da762d3fab48df

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
adutra: @iting0321 let's keep both checks for now. There is an ongoing discussion about these checks and we don't know yet if it makes sense to have them or not. So, I'd suggest that we leave this topic out of this PR.

dimas-b:
Re: checked-and-egg problem: I believe only `checkAllowExternalCatalogCredentialVending()` needs to consider delegation mode parameters.

Could we lift that logic into the "mode resolver"?

Basically, if the client requests both REMOTE_SIGNING and VENDED_CREDENTIALS for a Federated catalog, we can simply drop VENDED_CREDENTIALS and continue execution. If only VENDED_CREDENTIALS is requested, it's an error (same as now).

+1 to dropping `UNKNOWN` and using `Optional` where one specific mode is needed.

adutra:
> Re: checked-and-egg problem: I believe only `checkAllowExternalCatalogCredentialVending()` needs to consider delegation mode parameters.

Well, there is also `buildLoadTableResponseWithDelegationCredentials()` that (obviously) needs the resolved mode.

> Basically, if the client requests both REMOTE_SIGNING and VENDED_CREDENTIALS for a Federated catalog, we can simply drop VENDED_CREDENTIALS and continue execution. If only VENDED_CREDENTIALS is requested, it's an error (same as now).

Hmm I think I see what you mean, you would like to move the check for `FeatureConfiguration.ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING` into the resolver, where we already have the check for `FeatureConfiguration.ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING`.

I like the idea 👍  This places these two checks side by side in the same place, which I think will help us eventually decide on merging them together.

adutra:
Hi @iting0321 are you still working on this feature?

iting0321:
> Hi @iting0321 are you still working on this feature?

Sorry that I oversaw the previous reply, I will follow-up on this one recently. Thank you for the kindly reminder.

iting0321:
Hi @adutra, I cherry-picked the commit and moved the `ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING` check from `IcebergCatalogHandler` into `DefaultAccessDelegationModeResolver`, placing it next to the existing `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING` check.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -1293,31 +1298,33 @@ private EnumSet<PolarisAuthorizableOperation> getUpdateTableAuthorizableOperatio
...
+   * Resolves the access delegation mode by delegating to the configured {@link
+   * AccessDelegationModeResolver}.
+   *
+   * <p>Callers must ensure that {@code requestedModes} is not empty before invoking this method. If
+   * no delegation is needed, callers should skip resolution entirely rather than passing an empty
+   * set.
+   *
+   * @param requestedModes The non-empty set of delegation modes requested by the client
+   * @return The resolved access delegation mode
+   * @throws IllegalArgumentException if {@code requestedModes} is empty
```
adutra: These javadocs are now a bit obsolete. Suggestion:

```suggestion
   * @param requestedModes The non-empty set of delegation modes requested by the client
   * @return The resolved access delegation mode, or empty if no delegation mode was resolved
```

iting0321:
> Thank you for the quick turnaround, @iting0321 !
> 
> I think this PR is almost ready, I only had one minor remark.

Hi @adutra, I just fixed these javadocs. 
Thank you for the feedback!

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    // Check if credential vending is enabled for this catalog.
+    // For internal catalogs, credential vending is always enabled.
+    // For external/federated catalogs, check if ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING is
+    // enabled. Note: reaching here means ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING is true.
+    boolean credentialVendingEnabled =
+        !catalogEntity.isExternal()
+            || realmConfig.getConfig(
+                FeatureConfiguration.ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING, catalogEntity);
```
adutra, line 151: I still see one small inconsistency now that we have both feature flags checked here:

When a single `VENDED_CREDENTIALS` is requested, only `ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING` is checked. But when both modes are requested, both `ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING` and `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING` are checked before selecting `VENDED_CREDENTIALS`.

IOW this means: if a client requests only `VENDED_CREDENTIALS` on an external catalog where `ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING=true` but `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING=false`, the resolver will happily return `VENDED_CREDENTIALS`. But if that same client had requested both modes, the resolver would have fallen back to `REMOTE_SIGNING`.

I think that the single-mode path above should also check `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING` before accepting `VENDED_CREDENTIALS` — and if it fails, throw a `ForbiddenException` (since there’s no fallback mode available).

adutra: And one last small request: let's list `ForbiddenException`  as a possible `@throws` in the javadocs of this method.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * <p>The resolution logic:
+ *
+ * <ol>
+ *   <li>If no delegation mode is requested, returns empty.
+ *   <li>If exactly one delegation mode is requested:
+ *       <ul>
+ *         <li>If {@link AccessDelegationMode#VENDED_CREDENTIALS} is requested but the catalog is
+ *             external and {@link FeatureConfiguration#ALLOW_EXTERNAL_CATALOG_CREDENTIAL_VENDING}
+ *             is disabled, throws {@link org.apache.iceberg.exceptions.ForbiddenException}.
```
adutra: Would you mind updating the logic and mention the newly-added check for `ALLOW_FEDERATED_CATALOGS_CREDENTIAL_VENDING`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationModeResolver.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Resolves the optimal access delegation mode based on the requested modes and catalog
+   * capabilities.
+   *
+   * @param requestedModes The set of delegation modes requested by the client
+   * @param catalogEntity The catalog entity, used to determine storage configuration and
+   *     capabilities
+   * @return The resolved access delegation mode, or empty if no suitable mode is available.
+   * @throws org.apache.iceberg.exceptions.ForbiddenException if the client requests a mode that is
```
dimas-b: I'd prefer Polaris SPIs to not rely on Iceberg exceptions (narrower scope, easier maintenance).

Also, Iceberg's `ForbiddenException` has javadoc like `HTTP 403 Forbidden - Failed authorization checks.`, which does not appear to fit this use case. Polaris Authorizers are not involved at all.

I'd think a 400 response (Bad Request) is more appropriate because the caller ought to know the nature of the catalog for which it requests access delegation. That is to say, asking for credential vending from a federated catalog is not an authorization failure but a setup/configuration mistake.

For example, `buildLoadTableResponseWithDelegationCredentials()` throws an `IllegalArgumentException`, when credential vending has been requested, but cannot be performed. I'd think we should throw an `IllegalArgumentException` in this case too.

That said, do we have to declare a specific exception type here at all?

adutra: If we throw `ForbiddenException` imho we should mention it here because it is a quite specific exception.

I agree though that a 400 here would also make sense. I therefore suggest that we change the exception to `IllegalArgumentException`.

dimas-b:
@flyingImer : I do not think the described behaviour with `REMOTE_SIGNING` is problematic. Polaris never supported `REMOTE_SIGNING`.

When `REMOTE_SIGNING` support is added, the default resolver's behaviour will be correct, as far as I can tell.

Indeed, this PR feels a bit odd in what it is trying to improve access mode resolution before we have support for multiple modes. However, I think it is fine do this preparatory work before `REMOTE_SIGNING` is available.

Related message: https://github.com/apache/polaris/pull/3750#issuecomment-3891957835

adutra:
@iting0321 do you mind rebasing so that we can re-approve and merge? Thanks 🙏 

@flyingImer +1 to what Dmitri said: this PR is a bit ahead of phase wrt to the work on remote signing, but it is a good and solid preparatory work for that.

iting0321:
Hi @adutra, I have already fixed the conflict!

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
```diff
@@ -105,6 +109,9 @@ public abstract class AbstractIcebergCatalogHandlerAuthzTest extends PolarisAuth
 
   @Inject LocalCatalogFactory localCatalogFactory;
   @Inject IcebergCatalogHandlerFactory icebergCatalogHandlerFactory;
+  @Inject Instance<FederatedCatalogFactory> externalCatalogFactories;
+  @Inject CatalogPrefixParser prefixParser;
+  @Inject AccessDelegationModeResolver accessDelegationModeResolver;
```
dimas-b: Do we have to inject these variables? They appear to be unused 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -222,6 +225,10 @@ public abstract class IcebergCatalogHandler extends CatalogHandler implements Au
   // e.g., no pagination, location transformation, or multi-table transactions.
   private boolean isFederated = false;
 
+  // Resolved access delegation mode, initialized via resolveAccessDelegationModes()
+  @SuppressWarnings("immutables:incompat")
+  private AccessDelegationMode resolvedAccessDelegationMode = null;
```
adutra: I thought we eventually decided to not store this info in a field. Is this maybe an artifact of your recent rebase?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
adutra: +1, probably something that re-appeared during rebase.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,241 @@
...
+ *             returns {@link AccessDelegationMode#REMOTE_SIGNING}.
+ *         <li>If credential subscoping is skipped, returns {@link
+ *             AccessDelegationMode#REMOTE_SIGNING}.
+ *         <li>If STS is unavailable for the catalog's AWS storage, returns {@link
+ *             AccessDelegationMode#REMOTE_SIGNING}.
+ *         <li>Otherwise returns {@link AccessDelegationMode#VENDED_CREDENTIALS}.
+ *       </ul>
+ * </ol>
+ */
+@RequestScoped
```
adutra: Didn't we agree to make this `@ApplicationScoped`? (I don't recall)

But since this bean holds no mutable state, I don't see why it's request-scoped.

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
```diff
@@ -0,0 +1,241 @@
...
+public class DefaultAccessDelegationModeResolver implements AccessDelegationModeResolver {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(DefaultAccessDelegationModeResolver.class);
+
+  private final RealmConfig realmConfig;
+
+  @Inject
+  public DefaultAccessDelegationModeResolver(CallContext callContext) {
```
adutra: Again I think I mentioned this before... you can inject `RealmConfig` directly.

dimas-b: +1 to direct injection

runtime/service/src/main/java/org/apache/polaris/service/catalog/DefaultAccessDelegationModeResolver.java
dimas-b: SGTM :+1:

adutra:
Thank you @iting0321 !

PR #3755, opened 2026-02-12
Synchronize on PolarisConfiguration.registerConfiguration()

adutra:
The `ALL_CONFIGURATIONS` field is using `ArrayList` which is not thread-safe.

The field is only accessed via `registerConfiguration()`, and all configurations are currently registered during static initialization.

But we have two subclasses, `FeatureConfiguration` and `BehaviorChangeConfiguration`, that both call that method.

Consider this scenario:

1. Thread A triggers loading of `FeatureConfiguration`;
2. Thread B triggers loading of `BehaviorChangeConfiguration`;
3. Both classes' static initializers run in parallel;
4. Both call `registerConfiguration()` on the shared list concurrently.

In order to avoid this, this PR <strike>changes the list to `CopyOnWriteArrayList`</strike> makes the `registerConfiguration` method synchronized.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
@@ -21,10 +21,14 @@ package org.apache.polaris.core.config;
 import jakarta.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * An ABC for Polaris configurations that alter the service's behavior TODO: deprecate unsafe
@@ -34,8 +38,11 @@ import java.util.stream.Stream;
  */
 public abstract class PolarisConfiguration<T> {
 
-  // Only mutated via the `registerConfiguration` method, which is synchronized.
-  private static final List<PolarisConfiguration<?>> ALL_CONFIGURATIONS = new ArrayList<>();
+  private static final Logger LOGGER = LoggerFactory.getLogger(PolarisConfiguration.class);
+
+  private static final Map<String, PolarisConfiguration<?>> ALL_CONFIGURATIONS =
+      new ConcurrentHashMap<>();
+  private static final Set<String> ALL_CATALOG_CONFIGS = new ConcurrentSkipListSet<>();
 
   private final String key;
   private final String description;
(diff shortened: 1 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -50,7 +52,7 @@ public abstract class PolarisConfiguration<T> {
    * configs.
    */
   private static void registerConfiguration(PolarisConfiguration<?> configuration) {
-    for (PolarisConfiguration<?> existingConfiguration : allConfigurations) {
+    for (PolarisConfiguration<?> existingConfiguration : ALL_CONFIGURATIONS) {
       if (existingConfiguration.key.equals(configuration.key)) {
         throw new IllegalArgumentException(
             String.format("Config '%s' is already in use", configuration.key));
```
dimas-b: The change looks good to me... yet, the use case described in the PR makes me wonder if perhaps this check may be circumvented in the concurrent load situation 😅 not a big deal, IMHO.

adutra: Hmm that's right. We might just as well make this method synchronized. I wanted to avoid that because it's less future-proof than using `CopyOnWriteArrayList`, but 🤷‍♂️

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -71,12 +72,12 @@ private static void registerConfiguration(PolarisConfiguration<?> configuration)
...
       }
     }
-    allConfigurations.add(configuration);
+    ALL_CONFIGURATIONS.add(configuration);
   }
 
   /** Returns a list of all PolarisConfigurations that have been registered */
   public static List<PolarisConfiguration<?>> getAllConfigurations() {
-    return List.copyOf(allConfigurations);
+    return List.copyOf(ALL_CONFIGURATIONS);
```
snazy: The writes to `ALL_CONFIGURATIONS` via `registerConfiguration` are fine, but this one can still yield wrong results (no fence/barrier here).

I suspect that the there are maybe just a handful calls to `registerConfiguration`.

WDYT of using a copy-on-write but save the `List.copyOf()` on every read?

```java
private static volatile List<PolarisConfiguration<?>> allConfigurations = List.of();

// The `synchronized` is only needed to to have no concurrent executions of registerConfiguration(), not needed as a memory fence/barrier..
synchronized void registerConfiguration(...) {
  var updatedConfigs = new ArrayList(allConfigurations);
  // register dance
  allConfigurations = List.copyOf(updatedConfigs);
}

static List<PolarisConfiguration<?>> getAllConfigurations() {
    // reads the current immutable list object
    return allConfigurations;
}
```

dimas-b: +1

adutra: Isn't it simpler to just synchronize `getAllConfigurations`? It's only used in `org.apache.polaris.service.config.ReservedProperties.AllowlistHolder`.

snazy: Hm, could the allow-list holder functionality be moved into PolarisConfiguration and make the allow-list part of the builder?

dimas-b: I'm fine with simple `synchronized` methods. The affected code is not on the hot path, AFAIK.

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
adutra: Let's merge this as is and see if we can improve `ReservedProperties.AllowlistHolder` later. I suspect there is a  ton of improvement opportunities.

adutra: Well I had to fix a conflict, so I seized the opportunity to remove `AllowlistHolder`.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -55,33 +59,33 @@ public abstract class PolarisConfiguration<T> {
...
+            String.format(
+                "Catalog config '%s' is already in use", configuration.catalogConfigUnsafe()));
       }
     }
-    allConfigurations.add(configuration);
   }
 
-  /** Returns a list of all PolarisConfigurations that have been registered */
+  /** Returns a list of all PolarisConfigurations that have been registered. */
   public static List<PolarisConfiguration<?>> getAllConfigurations() {
```
adutra, line 82: Now unused.

polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -55,33 +59,33 @@ public abstract class PolarisConfiguration<T> {
...
    */
   private static void registerConfiguration(PolarisConfiguration<?> configuration) {
-    for (PolarisConfiguration<?> existingConfiguration : allConfigurations) {
-      if (existingConfiguration.key.equals(configuration.key)) {
+    if (ALL_CONFIGURATIONS.putIfAbsent(configuration.key(), configuration) != null) {
+      throw new IllegalArgumentException(
+          String.format("Config '%s' is already in use", configuration.key));
+    }
+    if (configuration.hasCatalogConfig()) {
+      if (!ALL_CATALOG_CONFIGS.add(configuration.catalogConfig())) {
```
dimas-b, line 67: nit: I think this is still a bit racy if we want to be strict. A configuration may be added `ALL_CONFIGURATIONS`, but fail the catalog property checks, which will lead to inconsistent entries.

Not a big deal, since failures are likely to break the whole runtime in the current state of the code... but the intent of this PR is to ensure multi-threaded correctness, so I wonder if it is worth the trouble to use async collections compared to simple `synchronized` methods? (not a blocker)

adutra: > since failures are likely to break the whole runtime in the current state of the code

Yes, the assumption is that any error here is fatal and will make the JVM exit. I don't think we could bring the collections to contain inconsistent data.

dimas-b: fair point

dimas-b (approved):

PR #3756, opened 2026-02-12
Helm chart: use the chart version as default image tag

adutra:
Since the chart release cycle is tied to the server's it makes sense to default the image tag to the chart version, which is also the Polaris server version.

## Checklist

### DIFF
--- a/helm/polaris/values.yaml
+++ b/helm/polaris/values.yaml
@@ -26,8 +26,8 @@ image:
   repository: apache/polaris
   # -- The image pull policy.
   pullPolicy: IfNotPresent
-  # -- The image tag.
-  tag: "latest"
+  # -- The image tag. If not set, the chart version is used.
+  tag: ""
   # -- The path to the directory where the application.properties file, and other configuration
   # files, if any, should be mounted.
   configDir: /deployments/config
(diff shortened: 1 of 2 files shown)


helm/polaris/values.yaml
```diff
@@ -26,8 +26,8 @@ image:
   repository: apache/polaris
   # -- The image pull policy.
   pullPolicy: IfNotPresent
-  # -- The image tag.
-  tag: "latest"
+  # -- The image tag. If not set, the chart version is used.
+  tag: ""
```
MonkeyCanCode, line 30: The only concern I have here is users who are on main won't be able to do helm install without doing a local image build as with this, the tag will be resolved to `apache/polaris:1.4.0-incubating-SNAPSHOT` which doesn't existed anywhere. Even if a user did performed a local image build, it will be `apache/polaris:1.4.0-incubating` instead.

snazy: I guess the image build could be adjusted to tag the images as `x.y.z-incubating` and `latest`?

snazy: (If we want to keep `latest`, which would be confusing when s/o builds from an older commit?)

adutra: It seems we've been bumping the chart's `version` and `appVersion` manually. How about we include this bump in the release workflow? In which case we could make the chart's `version` match the Docker tag, and the chart's `appVersion` match the Maven version:

```yaml
version: 1.4.0-incubating
appVersion: 1.4.0-incubating-SNAPSHOT
```

adutra:
I'm closing this in favor of changing the tag during the release process: #3796 .

PR #3757, opened 2026-02-13
Core: Add Java doc for the class PolarisPrivilege

flyrain:
To clarify PolarisPrivilege may only needed by the built-in RBAC

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
@@ -26,12 +26,15 @@ import java.util.List;
 
 /**
  * Enumerates the privileges used by the built-in RBAC authorizer ({@link
- * org.apache.polaris.core.auth.PolarisAuthorizerImpl}). These privileges are granted to roles and
- * checked against securables during authorization.
+ * org.apache.polaris.core.auth.PolarisAuthorizerImpl}). In Polaris' RBAC model, a
+ * <em>securable</em> is a Polaris entity (such as a catalog, namespace, table, or policy) on which
+ * access may be controlled by granting privileges to a grantee (typically a catalog role or
+ * principal role). Each privilege in this enum targets a specific {@link PolarisEntityType} of
+ * securable and is checked against that securable during authorization.
  *
  * <p>Alternative authorizer implementations such as the OPA-based authorizer may not use these
  * privileges. They operate at the {@link org.apache.polaris.core.auth.PolarisAuthorizableOperation}
- * level and delegates all privilege/permission logic to external PDPs.
+ * level and delegate all privilege/permission logic to external PDPs.
  */
 public enum PolarisPrivilege {
   SERVICE_MANAGE_ACCESS(1, PolarisEntityType.ROOT),


polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
```diff
@@ -24,7 +24,15 @@
...
 
-/** List of privileges */
+/**
+ * Enumerates the privileges used by the built-in RBAC authorizer ({@link
+ * org.apache.polaris.core.auth.PolarisAuthorizerImpl}). These privileges are granted to roles and
+ * checked against securables during authorization.
+ *
+ * <p>Alternative authorizer implementations such as the OPA-based authorizer may not use these
+ * privileges. They operate at the {@link org.apache.polaris.core.auth.PolarisAuthorizableOperation}
+ * level and delegates all privilege/permission logic to external PDPs.
```
adutra:
```suggestion
 * level and delegate all privilege/permission logic to external PDPs.
```

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
```diff
@@ -24,7 +24,15 @@
 import jakarta.annotation.Nullable;
 import java.util.List;
 
-/** List of privileges */
+/**
+ * Enumerates the privileges used by the built-in RBAC authorizer ({@link
+ * org.apache.polaris.core.auth.PolarisAuthorizerImpl}). These privileges are granted to roles and
+ * checked against securables during authorization.
```
dimas-b: nit: "securable" is not a common term (not in my experience)... It might be good to define it or link to where it is defined, or rephrase using more common terms.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
```diff
@@ -24,7 +24,15 @@
...
+ *
+ * <p>Alternative authorizer implementations such as the OPA-based authorizer may not use these
```
dimas-b, line 35: nit: While this statement is valid, is it relevant to this class? Why should the reader be concerned with OPA in the context of `PolarisPrivilege`? The paragraph above already scopes it down to the Internal Authorizer 🤔

dimas-b (commented):
Thanks for the javadoc effort, @flyrain ! Please consider my comments optional and only an attempt to reach greater clarity.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
flyrain: it was defined here, https://polaris.apache.org/in-dev/unreleased/managing-security/access-control/#securable-object, should I link it? Or what other terms do you perfer?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
flyrain: OPA is an example here, I understand that it looks like duplication, just to clarify more to avoid confusions. I can remove this paragraph if you feel strongly about it.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
dimas-b: My point is that it is not self-evident to the reader of the java code. I do not see links to Polaris' own site from code as particularly useful as code will evolve independently from the site. I'd prefer a javadoc description or a link to javadoc in another class.

flyrain:
Hi @adutra @dimas-b, updated the PR per your suggestion. Can you take another look?

adutra (approved):

PR #3760, opened 2026-02-14
core: Refactor PolarisAuthorizer SPI (Phase 1): new interfaces and utilities

sungwy:
This is the first PR in a series to refactor the PolarisAuthorizer SPI so authorizer implementations can control whether Polaris-native principal/RBAC resolution is required.

Summary of changes in this PR:
  - Added `authorizeDecision(...)` (returns `AuthorizationDecision`) and a new `authorizeOrThrow(...)` overload that uses `AuthorizationState` + `AuthorizationRequest`.
  - Added request-scoped `RequestAuthorizationState` CDI bean implementing `AuthorizationState`.
  - Implemented resolver selection planning so `resolveSelections(...)` can skip caller principal/role resolution when not requested.
  - Included new selection-based resolution tests in ResolverTest.
  - New classes introduced: `AuthorizationDecision`, `AuthorizationState`, `AuthorizationRequest`

In this PR, handlers still use the legacy call sites, and hence no behavior change is expected. The new SPI methods remain unimplemented in `PolarisAuthorizerImpl` and `OpaPolarisAuthorizer`.

Once this SPI is accepted, follow-up PRs will implement the new SPI methods in `PolarisAuthorizerImpl` and `OpaPolarisAuthorizer`.

RFC Google Doc on Authorization Refactor: [Link](https://docs.google.com/document/d/1OaiQG_C4-yUe0ihaDBxtw_mEcOOzUBnWPazzVbjQi5U/edit?tab=t.0#heading=h.dyow25dt9w1)
Parent tracking Issue: https://github.com/apache/polaris/issues/3779

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -247,7 +264,7 @@ public ResolverStatus resolveAll() {
     int count = 0;
     ResolverStatus status;
     do {
-      status = runResolvePass();
+      status = runResolvePass(plan);
       count++;
```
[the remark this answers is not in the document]
sungwy: This bug seems to predate my change - taking a look to see if it would make sense to fix this as a part of this PR

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -0,0 +1,98 @@
...
+  @Nonnull
+  PolarisAuthorizableOperation getOperation();
+
+  /** Returns the primary target securables, if any. */
+  @Nullable
+  List<PolarisSecurable> getTargets();
+
+  /** Returns the secondary securables, if any. */
+  @Nullable
+  List<PolarisSecurable> getSecondaries();
```
dimas-b: What is a "secondary" securable really? 😅 I looked at them in the core RBAC code many times, but the idea does not stick in my head 😅 Since this is an attempt to refactor the authZ SPI, could you add javadoc to describe targets and secondaries in more details, please?

I know the concept predates this effort, but it's a good opportunity to clarify this... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationState.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.polaris.core.persistence.resolver.PolarisResolutionManifest;
+
+/**
+ * Request-scoped authorization state shared across authorization phases.
+ *
+ * <p>Distinct from {@link org.apache.polaris.core.context.CallContext}. This state is intended to
+ * carry authorization-specific data, such as a {@link PolarisResolutionManifest} created by the
+ * caller and populated by the authorizer.
+ */
+public interface AuthorizationState {
```
dimas-b: Do we expect multiple different implementations of this "state"? Why not make it a class?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -0,0 +1,52 @@
...
+    return ImmutablePolarisSecurable.builder().entityType(entityType).nameParts(nameParts).build();
+  }
+
+  /** Returns the entity type of the securable. */
+  @Nonnull
+  PolarisEntityType getEntityType();
+
+  /** Returns the name parts that identify the securable in hierarchical order. */
+  @Nonnull
+  List<String> getNameParts();
```
dimas-b, line 46: nit: maybe `getNameElements()`?... (really minor)

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /** Returns the dot-joined name derived from {@link #getNameParts()}. */
+  default @Nonnull String getName() {
```
dimas-b: Is it intended for any non-informational use? Joining with dot separators is not a reversible mapping WRT `getNameParts()`, I believe 🤔

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationState.java
sungwy: I did this intentionally because I wanted to keep `AuthorizationState` in `polaris-core` and the request-scoped CDI in `polaris-runtime`. But yes, I agree that we would not intend for there to be multiple implementations

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: Thank you for the scrutiny!

I assumed we'd need this because of the complexity in handling multi-part namespaces in OPA inputs, but I now I remember that we've actually decided to move forward with clearly preserving the parent hierarchy through an ordered list. I think it'll be safer to remove this so that we don't introduce an unnecessary string conversion that could be confusing in the future.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
sungwy: I found this very confusing as well... I agree this is a good place for the Java doc to explain the difference between the two. Let me add it

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationState.java
dimas-b: We can still keep a "plain" impl. class in `polaris-core` and have a producer method in `polaris-runtime` with appropriate CDI annotations. It might cause some booiler-plate code for (injected) passing parameters to the constructor/builder, but hopefully not too much. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
dimas-b: removing `getName()` sounds good (not hard to re-add when / if we need it).

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -0,0 +1,116 @@
...
+   * <ul>
+   *   <li>Table rename: the destination namespace (target is the source table).
+   *   <li>Role grants: the grantee role/principal (target may be the role or the resource being
+   *       granted on).
+   *   <li>Policy attach/detach: the catalog/namespace/table being attached to (target is the
+   *       policy).
+   * </ul>
+   */
+  @Nullable
+  List<PolarisSecurable> getSecondaries();
```
dimas-b: Thanks for the javadoc! It certainly helped me refresh my memory 😉 

Given current use cases, I wonder if a List-to-List relationship between targets and secondary is actually correct in general. I'd imagine in all relevant cases it is a 1:1 association. When there are multiple targets, there are going to be multiple target/secondary pairs, right?

I know it might be a bit of rift in the existing code pattern, but since we're introducing a new SPI, WDYT about modeling this as `List<Pair<PolarisSecurable, PolarisSecurable>> getTargets()`? Of course, we have have a dedicated small immutable object with target/secondary attributes instead of the `Pair`.

dimas-b (dismissed):

dimas-b:
I clicked "approve" too eagerly perhaps given my comment above, but I do not want to retract it as the PR LGTM overall :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
sungwy: > I know it might be a bit of rift in the existing code pattern, but since we're introducing a new SPI, WDYT about modeling this as List<Pair<PolarisSecurable, PolarisSecurable>> getTargets()? Of course, we have have a dedicated small immutable object with target/secondary attributes instead of the Pair.

Yeah sure thing - Now that you mention it, I think it'll be helpful to create a new class to represent this, than using a `Pair`

> Thanks for the javadoc! It certainly helped me refresh my memory 😉

Now that I'm thinking about it, I'm not a fan of the `secondaries` convention at all - this makes the `authorization` checks quite awkward and complicated. Instead what I think we should have done is to just break them into two authorization checks. As an example, for a check like renaming a table that requires table1 and table2, we should have just checked if the principal can rename from table1, and then subsequently to table2. However, I think this will result in an even larger refactoring and I want to keep it out of scope of this effort.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationState.java
sungwy: adopted this feedback - thanks @dimas-b !

runtime/service/src/main/java/org/apache/polaris/service/auth/RequestAuthorizationState.java
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.
+ */
+package org.apache.polaris.service.auth;
+
+import jakarta.enterprise.context.RequestScoped;
+import jakarta.enterprise.inject.Produces;
+import org.apache.polaris.core.auth.AuthorizationState;
+
+/** Request-scoped producer for {@link AuthorizationState} in runtime services. */
+@RequestScoped
```
dimas-b: nit: I do not think this annotation is required... If it is, it should probably be a `@Singleton`, WDYT?

runtime/service/src/main/java/org/apache/polaris/service/auth/RequestAuthorizationState.java
```diff
@@ -0,0 +1,33 @@
...
+public class RequestAuthorizationState {
+  @Produces
+  @RequestScoped
+  public AuthorizationState get() {
+    return new AuthorizationState();
```
dimas-b: nit: traditionally, these methods are collected in `ServiceProducers`, but having a dedicated producer class is ok too.

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
dimas-b: I'm fine keeping secondaries in this PR. `AuthorizationTargetBinding` already makes things a lot clearer at the SPI level.

runtime/service/src/main/java/org/apache/polaris/service/auth/RequestAuthorizationState.java
sungwy: yep - you are right, this was a lousy follow-up your feedback. This should be a Singleton :)

runtime/service/src/main/java/org/apache/polaris/service/auth/RequestAuthorizationState.java
sungwy: I think following convention is important in a project, so will adopt this feedback as well. Although I think the ServiceProducer is becoming very crowded...

I’d like to keep convention for now in this PR to minimize churn, but do you see a world where we separate out the ServiceProducer into domain specific ones? (like: `AuthProducers`/`PersistenceProducers`) to improve discoverability?

Curious to hear your thoughts

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -22,11 +22,47 @@
...
+  AuthorizationDecision authorizeDecision(
+      @Nonnull AuthorizationState authzState, @Nonnull AuthorizationRequest request);
+
+  /**
+   * Convenience method that throws a {@link ForbiddenException} when authorization is denied.
+   *
+   * <p>Implementations should provide allow/deny decisions via {@link #authorizeDecision}.
+   */
+  default void authorizeOrThrow(
+      @Nonnull AuthorizationState authzState, @Nonnull AuthorizationRequest request) {
```
singhpk234: why is `authZState` an arg, can it not be injected ?

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationDecision.java
```diff
@@ -0,0 +1,57 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.auth;
+
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+
+/** Authorization decision returned by authorizer implementations. */
+public final class AuthorizationDecision {
```
adutra: nit: this might benefit from being a `@PolarisImmutable` + usage of `Optional` for the message, in which case I don't think we need the `getMessageOrDefault` method:

```java
@PolarisImmutable
public interface AuthorizationDecision {

  static AuthorizationDecision allow() {
    return ImmutableAuthorizationDecision.of(true, Optional.empty());
  }

  static AuthorizationDecision deny(@Nullable String message) {
    return ImmutableAuthorizationDecision.of(false, Optional.ofNullable(message));
  }

  @Value.Parameter(order = 1)
  boolean isAllowed();

  @Value.Parameter(order = 2)
  Optional<String> getMessage();
}
```

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -0,0 +1,106 @@
...
+  @Nonnull
+  List<AuthorizationTargetBinding> getTargetBindings();
+
+  /**
+   * Returns the primary target securables, if any.
+   *
+   * <p>Compatibility accessor derived from {@link #getTargetBindings()}.
+   */
+  @Nonnull
+  default List<PolarisSecurable> getTargets() {
```
adutra, line 66:
```suggestion
  @Value.Derived
  default List<PolarisSecurable> getTargets() {
```

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -0,0 +1,106 @@
...
+    return getTargetBindings().stream().map(AuthorizationTargetBinding::getTarget).toList();
+  }
+
+  /**
+   * Returns secondary securables, if any.
+   *
+   * <p>Compatibility accessor derived from {@link #getTargetBindings()}.
+   */
+  @Nonnull
+  default List<PolarisSecurable> getSecondaries() {
```
adutra, line 77:
```suggestion
  @Value.Derived
  default List<PolarisSecurable> getSecondaries() {
```

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
```diff
@@ -0,0 +1,56 @@
...
+   * <ul>
+   *   <li>Table rename: the destination namespace (target is the source table).
+   *   <li>Role grants: the grantee role/principal (target may be the role or the resource being
+   *       granted on).
+   *   <li>Policy attach/detach: the catalog/namespace/table being attached to (target is the
+   *       policy).
+   * </ul>
+   */
+  @Nullable
+  PolarisSecurable getSecondary();
```
adutra, line 55: Two questions, but not for this PR:

1. Should we plan for the future and add the possibility of 1-N secondaries for a given target?
2. Wouldn't this be better modeled by inheritance? e.g.

```java
public interface TableRenameAuthorizationTargetBinding extends AuthorizationTargetBinding {
  PolarisSecurable getDestinationNamespace();
}
```

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -22,11 +22,47 @@
...
+      @Nonnull AuthorizationState authzState, @Nonnull AuthorizationRequest request);
+
+  /**
+   * Core authorization entry point for the new SPI.
+   *
+   * <p>Implementations should rely on any required state in {@link AuthorizationState} and the
+   * intent captured by {@link AuthorizationRequest} (principal, operation, and target securables).
+   */
+  @Nonnull
+  AuthorizationDecision authorizeDecision(
```
adutra: nit: the current name (at least for me) implies that the authorization has been done already when this method is called, and that it merely returns the authorization decision. If this method actually performs the authorization, I would suggest renaming it to `authorize`.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
```diff
@@ -22,11 +22,47 @@
...
+  /**
+   * Convenience method that throws a {@link ForbiddenException} when authorization is denied.
+   *
+   * <p>Implementations should provide allow/deny decisions via {@link #authorizeDecision}.
+   */
+  default void authorizeOrThrow(
+      @Nonnull AuthorizationState authzState, @Nonnull AuthorizationRequest request) {
+    AuthorizationDecision decision = authorizeDecision(authzState, request);
+    if (!decision.isAllowed()) {
+      String message = decision.getMessageOrDefault("Authorization denied");
```
adutra:
```suggestion
      String message = decision.getMessage().orElse("Authorization denied");
```

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java
```diff
@@ -856,6 +888,68 @@ private ResolverStatus resolveReferenceCatalog(
     return new ResolverStatus(ResolverStatus.StatusEnum.SUCCESS);
   }
 
+  private static final class ResolvePlan {
```
adutra: nit: very good candidate for a record class.

runtime/service/src/main/java/org/apache/polaris/service/auth/RequestAuthorizationState.java
adutra: I'm still not convinced we need to create a CDI bean for `AuthorizationState`, but why not.

In any case I would personally be in favor of adding it to `ServiceProducers` for now, and then later look into splitting `ServiceProducers` into concern-specific components as you suggested.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationState.java
```diff
@@ -0,0 +1,54 @@
...
+
+import jakarta.annotation.Nonnull;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.polaris.core.persistence.resolver.PolarisResolutionManifest;
+
+/**
+ * Request-scoped authorization state shared across authorization phases.
+ *
+ * <p>Distinct from {@link org.apache.polaris.core.context.CallContext}. This state is intended to
```
adutra: nit: not sure we need to call out  a comparison with `CallContext` here. We have many other request-scoped beans, and that's fine.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationDecision.java
sungwy: Good suggestion @adutra

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
sungwy: yeah I initially had it as `authorize` and then renamed it because I thought the new name implied that we are returning an `AuthorizationDecision`... but I think I agree with you that it's more confusing. I'll take the suggestion and revert it back

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationState.java
sungwy: Very agreeable nit :)

runtime/service/src/main/java/org/apache/polaris/service/auth/RequestAuthorizationState.java
dimas-b: +1 to refactoring `ServiceProducers` later.

In general (since we're on the subject), there is no technical difference where producer methods are located, so it's all about convenience of the code maintainer :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
sungwy: Thanks @adutra - I've adopted this feedback

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
sungwy: I can see the value in doing that - but I agree with you that it would be good to scope it out from this refactoring. I think it would be easier to complete this refactoring by keeping feature parity, and then entertain the idea of changing the `AuthorizationTargetBinding` implementations.

Changing its members will result in a backward incompatible change in the `OpaPolarisAuthorizer`'s OPA request input json (e.g. `destinationNamespace` vs `secondaries` as the field name).

But I'm definitely open to changes in the OPA request input structure to make it more intuitive, as it is currently an experimental feature, and it would be best to make those changes before it becomes a stable contact.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: I thought about this, but I think  `getNameParts` sounds better because we do refer to Namespaces with multiple segments as `multi-part` Namespace:

https://github.com/apache/iceberg/blob/9534c9b3adc29d127ecc541ce131f49fd72f1980/open-api/rest-catalog-open-api.yaml#L5012-L5014

dimas-b: fair enough :+1:

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizer.java
dimas-b: +1

sungwy:
@adutra @dennishuo @dimas-b - how shall we proceed from here - can we move forward with this PR?

sungwy:
@dimas-b - I think we should hold off on merging this. While implementing Phase 3 - implementing the SPI `OpaPolarisAuthorizer`, I realized that the process constructing the parent hierarchy for an AuthorizationTarget is actually persistence driven, instead of being lexicological...

I think that topic of how parent hierarchy should be constructed for a target based on an intent-based authorization argument should be discussed first.

sungwy:
I just looked through the code, and discussed with @dimas-b offline.

But it looks like the `Resolver` just looks through the individual parts of a name to construct its parent hierarchy, and optionally includes [ROOT_CONTAINER](https://github.com/apache/polaris/blob/483e1317d3737e862db1b74171325aaa3bed2355/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java#L35), and/or the [referenceCatalog](https://github.com/apache/polaris/blob/483e1317d3737e862db1b74171325aaa3bed2355/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java#L361). I think it should be possible to introduce a similar logic to iterate through the `AuthorizationTargetBinding`'s name parts to construct the parents needed for the OPA request as well.

https://github.com/apache/polaris/blob/483e1317d3737e862db1b74171325aaa3bed2355/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolver.java#L695-L726

No issues on merging this PR in @dimas-b - thank you for talking me through this

dimas-b:
SGTM :+1:

PR #3773, opened 2026-02-15
Unify keycloak docker tags

snazy:
This aligns with `getting-started/keycloak/docker-compose.yml`, prevent duplicate Renovate PRs against the different tags.

### DIFF
--- a/runtime/test-common/src/main/resources/org/apache/polaris/test/commons/keycloak/Dockerfile-keycloak-version
+++ b/runtime/test-common/src/main/resources/org/apache/polaris/test/commons/keycloak/Dockerfile-keycloak-version
@@ -19,4 +19,4 @@
 
 # Dockerfile to provide the image name and tag to a test.
 # Version is managed by Renovate - do not edit.
-FROM keycloak/keycloak:26.5.3
+FROM quay.io/keycloak/keycloak:26.5.3


runtime/test-common/src/main/resources/org/apache/polaris/test/commons/keycloak/Dockerfile-keycloak-version
```diff
@@ -19,4 +19,4 @@
 
 # Dockerfile to provide the image name and tag to a test.
 # Version is managed by Renovate - do not edit.
-FROM keycloak/keycloak:26.5.3
+FROM quay.io/keycloak/keycloak:26.5.3
```
adutra, line 22: Shouldn't we do the opposite and standardize on `keycloak/keycloak`?

snazy: quay.io is the "primary"  registry for Keycloak.
Docker Hub has rate limiting in place, which _should_ be irrelevant for GH actions, but not for other usage.
It's usually nicer to explicitly specify the registry instead of having the Docker daemon do the "try all the registries" dance.

adutra (approved):

PR #3782, opened 2026-02-16
Remove unused functions from CatalogHandlerUtils

snazy:


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
```diff
@@ -220,12 +214,6 @@ public CreateNamespaceResponse createNamespace(
         .build();
   }
 
-  public void namespaceExists(SupportsNamespaces catalog, Namespace namespace) {
```
adutra, line 223: This file is copied from Iceberg. I fear that, by removing unused methods, we would make future merges from the upstream file more difficult, wdyt?

snazy: Not sure, we've already changed this class for various Polaris specific reasons in the past, so I suspect it already diverged a lot.

adutra: > I suspect it already diverged a lot.

I can confirm 😅 – and it will get worse with Iceberg 1.11.

PR #1149, opened 2025-03-10
Fix annotations in `PolarisEntityManager`

dimas-b:
Following up on #1131:

* Fix annotations in `PolarisEntityManager`

* Add unit tests for running without `EntityCache`

PR #3788, opened 2026-02-16
Use more CDI in AbstractIcebergCatalogTest

snazy:
Remove the unnecessary "no entity cache" test class, which is not relevant for production workloads.

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
@@ -103,7 +103,6 @@ import org.apache.polaris.core.auth.PolarisPrincipal;
 import org.apache.polaris.core.config.FeatureConfiguration;
 import org.apache.polaris.core.config.RealmConfig;
 import org.apache.polaris.core.context.CallContext;
-import org.apache.polaris.core.context.RealmContext;
 import org.apache.polaris.core.entity.CatalogEntity;
 import org.apache.polaris.core.entity.NamespaceEntity;
 import org.apache.polaris.core.entity.PolarisBaseEntity;
@@ -138,6 +137,7 @@ import org.apache.polaris.service.catalog.io.ExceptionMappingFileIO;
 import org.apache.polaris.service.catalog.io.FileIOFactory;
 import org.apache.polaris.service.catalog.io.MeasuredFileIOFactory;
 import org.apache.polaris.service.catalog.io.StorageAccessConfigProvider;
+import org.apache.polaris.service.context.catalog.RealmContextHolder;
 import org.apache.polaris.service.events.EventAttributes;
 import org.apache.polaris.service.events.PolarisEvent;
 import org.apache.polaris.service.events.PolarisEventMetadataFactory;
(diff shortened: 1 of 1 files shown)


runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -293,32 +278,6 @@ public void before(TestInfo testInfo) {
     QuarkusMock.installMockForType(realmContext, RealmContext.class);
```
adutra: You probably could replace this mock with:

```java
  @Inject RealmContextHolder realmContextHolder;

  @BeforeEach
  public void before(TestInfo testInfo) {
    ...
    RealmContext realmContext = () -> realmName;
    realmContextHolder.set(realmContext);
  }
```

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -289,36 +275,9 @@ public void before(TestInfo testInfo) {
                 testInfo.getTestMethod().map(Method::getName).orElse("test"), System.nanoTime());
     bootstrapRealm(realmName);
 
-    RealmContext realmContext = () -> realmName;
-    QuarkusMock.installMockForType(realmContext, RealmContext.class);
+    realmContextHolder.set(() -> realmName);
     polarisContext = callContext.getPolarisCallContext();
 
-    EntityCache entityCache = createEntityCache(diagServices, realmConfig, metaStoreManager);
```
dimas-b, line 296: I believe testing without the `EntityCache` is still relevant since the impl. of `EntityCache` may not always be present (the system can be set up without it). Therefore, I think it makes sense to validate that Polaris still works without it.

NoSQL test probably cover that use case to some extent, but it is still possible with JDBC. I suppose it is not unrealistic to run JDBC Persistence without `EntityCache` with (hypothetically) some other Persistence-level cache.

snazy: Not sure, there's no way to configure Polaris to run without that cache.

dimas-b: It is doable in code, though... NoSQL Persistence runs without an `EntityCache`, so other custom impl. may elect to run without it too. My point is that it's probably worth testing it in the general case, not only via NoSQL Persistence.

dimas-b: The "no entity cache" tests were added in #1149 specifically to ensure core code correctness in that situation.

snazy: The option to run with no entity-cache was introduced for NoSQL.

> other custom impl. may elect

It's rather hypothetical. Any custom implementation would require its own tests anyway.

dimas-b: I still think the "no entity cache" tests make sense even with JDBC Persistence because the cache is exercised in the core code, which is supposed to work with any persistence, but I do not insist.

PR #3793, opened 2026-02-17
Move `decodeNamespace` to a utility class

snazy:
This makes the `decodeNamespace` function reusable without directly accessing Iceberg code.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
@@ -19,8 +19,9 @@
 
 package org.apache.polaris.service.catalog.common;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.net.URLEncoder;
-import java.nio.charset.Charset;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
```diff
@@ -94,4 +98,8 @@ public static void validateLocationsForTableLike(
               }
             });
   }
+
+  public static Namespace decodeNamespace(String namespace) {
+    return RESTUtil.decodeNamespace(URLEncoder.encode(namespace, Charset.defaultCharset()));
```
adutra: `Charset.defaultCharset()` is unreliable, let's seize the opportunity to change this:

```suggestion
    return RESTUtil.decodeNamespace(URLEncoder.encode(namespace, StandardCharsets.UTF_8));
```

snazy: Yea, I just initially didn't want to add that change as well. But since you're concerned about this as well, I'll update it.

adutra (approved):

PR #3798, opened 2026-02-17
Helm chart: add support for NoSQL persistence

adutra:
## Checklist
- [ ] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [ ] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [ ] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3800, opened 2026-02-17
Renovate: manage dependencies in helm charts and values

snazy:


.github/renovate.json5
```diff
@@ -53,6 +53,7 @@
         "gradle", "gradle-wrapper",
         "pip_requirements", "pip_setup", "uv", "setup-cfg",
         "dockerfile", "devcontainer", "docker-compose",
+        "helm-requirements", "helm-values",
```
adutra, line 56: I don't mind adding those but I'm not sure they will be very useful: `helm-values` updates docker images in values.yaml, but we only have one docker image, Polaris itself (and the tag is always `latest`). As for `helm-requirements` I think it's mostly useful when the chart has subcharts, which is not the case for Polaris.

snazy: Was mostly thinking about the mongo image (#3798) and the postgres image.

adutra: I'm not sure these managers will match those files, afaict they only match `values.ya?ml`.

adutra: Also: files like `helm/polaris/ci/fixtures/postgres.yaml` are Kubernetes YAML files, not Helm values files.

PR #3805, opened 2026-02-18
Optimize configuration value casts (#3688)

### LINKED ISSUE #3688: Optimize configuration value casts

AryanPatel226:
Optimizes `tryCast` to avoid unnecessary string conversions when configuration values are already the correct type.

The method was converting every value to string and parsing it back, even when the value already matched the expected type (Boolean, Integer, Long, Double). This adds type checks before string conversion to eliminate unnecessary overhead.

Fixes #3688

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #3688

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
@@ -177,17 +177,24 @@ public abstract class PolarisConfiguration<T> {
       return defaultValue();
     }
 
-    if (defaultValue() instanceof Boolean) {
+    Object defaultValue = defaultValue();
+
+    // If value is already the same type as the default, no conversion needed
+    if (defaultValue.getClass().isInstance(value)) {
+      return cast(value);
+    }
+
+    if (defaultValue instanceof Boolean) {
       return cast(Boolean.valueOf(String.valueOf(value)));
-    } else if (defaultValue() instanceof Integer) {
+    } else if (defaultValue instanceof Integer) {
       return cast(Integer.valueOf(String.valueOf(value)));
-    } else if (defaultValue() instanceof Long) {
+    } else if (defaultValue instanceof Long) {
       return cast(Long.valueOf(String.valueOf(value)));
-    } else if (defaultValue() instanceof Double) {
+    } else if (defaultValue instanceof Double) {
       return cast(Double.valueOf(String.valueOf(value)));
-    } else if (defaultValue() instanceof Float) {
+    } else if (defaultValue instanceof Float) {
       return cast(Float.valueOf(String.valueOf(value)));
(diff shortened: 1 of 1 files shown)


polaris-core/src/main/java/org/apache/polaris/core/config/PolarisConfiguration.java
```diff
@@ -173,17 +173,24 @@ public T resolveValue(
...
 
-    if (defaultValue() instanceof Boolean) {
+    Object defaultValue = defaultValue();
+
+    // If value is already the same type as the default, no conversion needed
+    if (defaultValue.getClass().isInstance(value)) {
+      return cast(value);
+    }
+
+    if (defaultValue instanceof Boolean) {
```
nandorKollar, line 187: nit: Would it make sense to turn this `if-else if` to switch expression, something like this:
```
    return switch (defaultValue()) {
          case Boolean ignored -> cast(Boolean.valueOf(String.valueOf(value)));
          case Integer ignored -> cast(Integer.valueOf(String.valueOf(value)));
          case Long ignored -> cast(Long.valueOf(String.valueOf(value)));
          case Double ignored -> cast(Double.valueOf(String.valueOf(value)));
          case Float ignored -> cast(Float.valueOf(String.valueOf(value)));
          case List<?> ignored -> cast(new ArrayList<>((List<?>) value));
          case null, default -> cast(value);
      };
```

AryanPatel226: @nandorKollar 
Good suggestion, but I think this compiles with --release 17, and switch pattern matching requires Java 21+. So we probably need to keep the if-else. But let me know if you have some suggestion.
Thank you!!

nandorKollar: Ah okay, that makes sense. Do we plan to compile with Java 21 in the near future?

jbonofre: @nandorKollar https://lists.apache.org/thread/v8ghm3k5nh9xp9x0xdo2kknk5ph0rnmb

nandorKollar: Thanks @jbonofre I just noticed that we recently moved to Java 17 minimum version for the clients.

dimas-b (approved):
Thanks for the contribution, @AryanPatel226 ! Changes LGTM 👍

nandorKollar (approved):

PR #3812, opened 2026-02-18
Helm docs overhaul

adutra:
This PR completely redesigns the Helm chart docs, and breaks the existing doc page into several pages: Production, Persistence, Authentication, Networking, Development, and Reference.

The values reference page contains the generated helm-docs output, but it is now split into many sections for readability. Non-leaf nodes in values.yaml file are now excluded from the generated output because they tend to produce unreadable content.

Screenshot of the new chart landing page:

image

## Checklist

### DIFF
--- a/helm/polaris/ci/nosql-persistence-values.yaml
+++ b/helm/polaris/ci/nosql-persistence-values.yaml
@@ -31,6 +31,17 @@ logging:
     json: true
     storage:
       size: 50Mi
+  categories:
+    org.apache.iceberg.rest: DEBUG
+    org.apache.polaris:
+      core: WARN
+      persistence.nosql: DEBUG
+  mdc:
+    env.name: ci
+    env.region: us-west-2
+    application:
+      name: polaris
+      version: 0.1.0
 
 persistence:
   type: nosql
@@ -40,3 +51,10 @@ persistence:
     secret:
       name: polaris-nosql-persistence
       connectionString: connectionString
+
+advancedConfig:
+  quarkus.mongodb.connect-timeout: 10S
+  quarkus:
+    mongodb:
+      write-concern:
+        safe: true


adutra:
Screenshot of the values reference page:

image

site/content/in-dev/unreleased/helm-chart/_index.md
```diff
@@ -0,0 +1,60 @@
...
+
+[![Artifact Hub](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/apache-polaris)](https://artifacthub.io/packages/search?repo=apache-polaris)
+
+The Apache Polaris project provides an official Helm chart for deploying Polaris on Kubernetes. The chart supports a wide range of configuration options, including persistence backends, authentication, scaling, observability, and more.
+
+The chart is available in the [official Apache Polaris Helm repository](https://downloads.apache.org/incubator/polaris/helm-chart).
+
+## Documentation
+
+- **[Production Installation]({{% relref "production" %}})**: Instructions for deploying Polaris in a production environment, including how to add the Helm repository and set up scaling.
```
dimas-b, line 34: does `relref` work inside the multi-version docs build? I remember seeing some odd issues related to using the same link name in multiple doc versions... but I do not recall exactly 😅

adutra: I deployed the Hugo site locally and everything is working. Should I do some more checks?

Makefile
```diff
@@ -287,6 +290,32 @@ helm-lint: check-dependencies ## Run Helm chart lint check
...
+	@kubectl wait --namespace polaris --for=condition=ready pod --selector=app.kubernetes.io/name=postgres --timeout=120s
+	@kubectl wait --namespace polaris --for=condition=ready pod --selector=app.kubernetes.io/name=mongodb --timeout=120s
+	@echo "--- Fixtures deployed and ready ---"
+
+helm-fixtures-cleanup: DEPENDENCIES := kubectl
+.PHONY: helm-fixtures-cleanup
+helm-fixtures-cleanup: check-dependencies ## Remove fixtures and namespace for Helm chart testing
+	@echo "--- Removing fixtures and namespace ---"
+	@kubectl delete --namespace polaris -f helm/polaris/ci/fixtures/ --ignore-not-found
+	@kubectl delete namespace polaris --ignore-not-found
```
MonkeyCanCode: As we are deleting namespace here already, we may as well just do `kubectl delete namespace polaris --wait=true --ignore-not-found` so it will do cascade delete for resources within the namespace?

site/content/in-dev/unreleased/helm-chart/production.md
```diff
@@ -0,0 +1,215 @@
...
+---
+
+This guide provides instructions for configuring the Apache Polaris Helm chart for a production environment. For full list of chart values, see the [Chart Reference]({{% relref "reference" %}}) page.
+
+## Prerequisites
+
+- A Kubernetes cluster (1.33+ recommended)
+- Helm 3.x or 4.x installed
+- `kubectl` configured to access your cluster
+- A PostgreSQL or MongoDB database
```
MonkeyCanCode: Should we promo MongoDB here as it is still under beta thus not necessary a production setup.

site/content/in-dev/unreleased/helm-chart/production.md
```diff
@@ -0,0 +1,215 @@
...
+#### Static Replicas
+
+`replicaCount` must be set to the desired number of pods:
+
+```yaml
+replicaCount: 3
+```
+
+#### Autoscaling
+
```
MonkeyCanCode, line 119: nit: extra newline

site/content/in-dev/unreleased/helm-chart/production.md
```diff
@@ -0,0 +1,215 @@
...
+```bash
+kubectl run polaris-bootstrap \
+  -n polaris \
+  --image=apache/polaris-admin-tool:latest \
+  --restart=Never \
+  --rm -it \
+  --env="polaris.persistence.type=nosql" \
+  --env="polaris.persistence.nosql.backend=MongoDb" \
+  --env="quarkus.mongodb.database=polaris" \
```
MonkeyCanCode, line 198: I could be wrong here, I thought db name is case sensitive in mongodb. In this case, do we need to change  to `quarkus.mongodb.database=polaris` otherwise, we will need to change realm on line 201 to `polaris` from `POLARIS` instead?

site/content/in-dev/unreleased/helm-chart/production.md
```diff
@@ -0,0 +1,215 @@
...
+  --rm -it \
+  --env="polaris.persistence.type=relational-jdbc" \
+  --env="quarkus.datasource.username=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.username}' | base64 --decode)" \
+  --env="quarkus.datasource.password=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.password}' | base64 --decode)" \
+  --env="quarkus.datasource.jdbc.url=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.jdbcUrl}' | base64 --decode)" \
+  -- \
+  bootstrap -r POLARIS -c POLARIS,root,$ROOT_PASSWORD
+```
+
+Example for the NoSQL (MongoDB) backend:
```
MonkeyCanCode: As NoSQL is still in beta (but in the values file we stated `Only MongoDb is supported for production use.`), do we need clarify this is in beta only? From persitence.md, we do clarify this. Thus wondering if that is worth mentioned it here again.

site/content/in-dev/unreleased/helm-chart/production.md
```diff
@@ -0,0 +1,215 @@
...
+  -n polaris \
+  --image=apache/polaris-admin-tool:latest \
+  --restart=Never \
+  --rm -it \
+  --env="polaris.persistence.type=relational-jdbc" \
+  --env="quarkus.datasource.username=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.username}' | base64 --decode)" \
+  --env="quarkus.datasource.password=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.password}' | base64 --decode)" \
+  --env="quarkus.datasource.jdbc.url=$(kubectl get secret polaris-persistence -n polaris -o jsonpath='{.data.jdbcUrl}' | base64 --decode)" \
+  -- \
+  bootstrap -r POLARIS -c POLARIS,root,$ROOT_PASSWORD
```
MonkeyCanCode: Should we add a statement here to clarify the relationship between `db-name` and `realm` (which we already stated in http://localhost:1313/in-dev/unreleased/realm/). For people who may just jump to `production.md` and `persistence.md`, the magic word "POLARIS" may be confusing and run into issues.

site/content/in-dev/unreleased/helm-chart/dev.md
```diff
@@ -0,0 +1,264 @@
...
+---
+
+This guide provides instructions for developers who want to work with the Polaris Helm chart locally, including setting up a local Kubernetes cluster, building from source, and running tests.
+
+## Local Development with Minikube
+
+### Prerequisites
+
+- [Minikube](https://minikube.sigs.k8s.io/docs/start/) installed
+- [Helm](https://helm.sh/docs/intro/install/) 3.x ou 4.x installed
```
MonkeyCanCode: typo on `ou`

site/content/in-dev/unreleased/helm-chart/dev.md
```diff
@@ -0,0 +1,264 @@
...
+
+```bash
+kubectl run polaris-bootstrap \
+  -n polaris \
+  --image=apache/polaris-admin-tool:latest \
+  --restart=Never \
+  --rm -it \
+  --env="polaris.persistence.type=nosql" \
+  --env="polaris.persistence.nosql.backend=MongoDb" \
+  --env="quarkus.mongodb.database=polaris" \
```
MonkeyCanCode, line 172: Same here for the database name. Will lower case `polaris` as database name works for bootstrap on line 170?

site/content/in-dev/unreleased/helm-chart/dev.md
```diff
@@ -0,0 +1,264 @@
...
+
+The above command will build and load the images into Minikube, deploy the fixtures, and run all `ct install` tests.
+
+## Cleanup
+
+Uninstall the chart, remove resources and delete the namespace:
+
+```bash
+helm uninstall --namespace polaris polaris
+kubectl delete --namespace polaris -f helm/polaris/ci/fixtures/
```
MonkeyCanCode: same posted on the makefile, should we just delete namespace directly?

MonkeyCanCode:
Thanks for the PR @adutra . This looks very clean and a lot better than the original length doc. The new added section annotations is great for helping new users to get familiar with the values file. I left couple comments. Overall, looks good to me.

site/content/in-dev/unreleased/helm-chart/reference.md
```diff
@@ -17,476 +17,324 @@
...
-| oidc | object | `{"authServeUrl":"","client":{"id":"polaris","secret":{"key":"clientSecret","name":""}},"principalMapper":{"idClaimPath":"","nameClaimPath":"","type":"default"},"principalRolesMapper":{"filter":"","mappings":[],"rolesClaimPath":"","type":"default"}}` | Polaris OIDC configuration. Only relevant when at least one realm is configured for external (or mixed) authentication. The currently supported configuration is for a single, default OIDC tenant. For more complex scenarios, including OIDC multi-tenancy, you will need to provide the relevant configuration using the `advancedConfig` section. |
+
+### Realm Context
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| realmContext.type | string | `"default"` | The type of realm context resolver to use. Two built-in types are supported: default and test; test is not recommended for production as it does not perform any realm validation. |
+| realmContext.realms | list | `["POLARIS"]` | List of valid realms, for use with the default realm context resolver. The first realm in the list is the default realm. Realms not in this list will be rejected. |
+
+### Features
```
MonkeyCanCode: nit: this section may looks odd to new users i think as the only value we have is `features.realmOverrides` and all other features are from default application config.

Makefile
adutra: I think it's better to delete the resources one by one (at least for those we know that they exist), then deleting the namespace, wdyt?

site/content/in-dev/unreleased/helm-chart/production.md
adutra: In this setup, `polaris` is the MongoDB database name, and `POLARIS` is the realm name – but the former doesn't need to match the latter.

site/content/in-dev/unreleased/helm-chart/production.md
adutra: I think there is some confusion with the word "backend".

We have 3 metastores:

- in-memory
- relational-jdbc
- nosql

And the `nosql` metastore has 2 backends:

- InMemory
- MongoDb

Indeed I used the word "backend" to mean "metastore", I will improve that.

But in any case, the `nosql` metastore is indeed in beta status today.

site/content/in-dev/unreleased/helm-chart/production.md
adutra: OK I will rename the realm `POLARIS` to `polaris-realm1` 👍

site/content/in-dev/unreleased/helm-chart/reference.md
adutra: True, let me try to improve this.

site/content/in-dev/unreleased/helm-chart/_index.md
dimas-b: It's fine... I do not remember the exact old error anyway :sweat_smile: we can fix when/if it breaks again.

Makefile
MonkeyCanCode: But deleting namespaces would delete everything within regardless no unless users deploy somethings with finalizers. Not a must change. I was just wondering if there is a specific reason on why we do so as delete ns will delete everything within.

adutra: OK you convinced me 😄 – I changed all the sites where we were deleting the namespace.

site/content/in-dev/unreleased/helm-chart/production.md
MonkeyCanCode: Yeah, my mistake here. Somehow I keep thinking they need to be the same but going over the code, that is not really the case. Sorry for the confusion.

site/content/in-dev/unreleased/helm-chart/reference.md
MonkeyCanCode: Not sure what should we do here? Put some default values here to make this section more read-able for end-users? But if we do so, we will then need to maintain the full list of default values going forward.

adutra: @MonkeyCanCode I came up with a different approach: I switched to HTML rendering for the docs, which allows me to easily include examples in the documentation.

There is one drawback though: the examples need to be written in HTML and they appear as (garbled) HTML content inside the JSON schema file.

Let me know what you think!

MonkeyCanCode: This is really interesting route and something that I didn't know that is possible to do. IMO, having sample values is more user friendly if the end-users is only looking at the site and not looking at the values file. However, it does appears this add good amount of potential overhead to the values file as you mentioned above (also seeing html tags in values file appear to be a bit overwhelming for the first impression and make the values file a bit more harder to read IMO). Based on my experiences, I spent way more time reading values files as opposed to checking the rendered values files on browser. I am not sure if this is truly worth it for this potential overhead or user experience. As this really only apply to one configuration, perhaps we should just pick a default enabled feature and have it not commented out such as
```
polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES" = [ "S3", "Azure", "GCP", "FILE"]
```
Plus, we already have document in-placed for production configuration to warm user to turn off the FILE one for this config). Thoughts?

adutra: I agree with you, it was a bit too much. I partially reverted the changes and went back to markdown tables.

I also added an example for `polaris.features` but the example I added was: `polaris.features."SUPPORTED_CATALOG_STORAGE_TYPES" = [ "S3", "Azure", "GCP"]` without `FILE` since that is the default value for this flag.

Let me know what you think!

adutra: The helm-docs project has an issue open for adding support for an Example column:

https://github.com/norwoodj/helm-docs/issues/298

I may try to contribute that feature.

MonkeyCanCode: Looks good. thanks @adutra

adutra:
Thanks @MonkeyCanCode for the thorough review and many nice suggestions! 🙏

PR #3723, opened 2026-02-10
Add Apache Paimon support to Polaris Spark Catalog

iting0321:
### Summary

This PR adds first-class support for Apache Paimon tables in the Polaris Spark Catalog plugin, enabling users to manage Paimon tables alongside Iceberg, Delta Lake, and Hudi tables through a unified catalog interface.

### How It Works

Paimon tables are stored using the existing **Generic Table API** with `format: "paimon"`. The implementation follows the same pattern as Delta and Hudi support:

1. Tables created with `USING paimon` are routed through `PaimonHelper`
2. `PaimonHelper` dynamically loads the Paimon Spark Catalog (`org.apache.paimon.spark.SparkCatalog`)
3. Paimon catalog delegates to `PolarisSparkCatalog` for metadata management
4. Paimon handles format-specific operations (snapshots, manifests, etc.)

### Usage Example

```sql
-- Configure Spark with Polaris and Paimon
spark.sql.catalog.polaris = org.apache.polaris.spark.SparkCatalog
spark.sql.catalog.polaris.uri = http://localhost:8181/api/catalog
spark.jars.packages = org.apache.paimon:paimon-spark-3.5:1.0.0

-- Create a Paimon table through Polaris
CREATE TABLE polaris.my_namespace.my_paimon_table (
    id INT,
    name STRING
) USING paimon
LOCATION 's3://bucket/path/to/table';

-- Query works seamlessly
SELECT * FROM polaris.my_namespace.my_paimon_table;
```

---

## Checklist

- [x] 🛡️ **Don't disclose security issues!** - N/A, this is a feature addition
- [x] 🔗 **Clearly explained why the changes are needed** - Enables multi-format catalog support as requested by users who want unified catalog management
- [x] 🧪 **Added/updated tests with good coverage** - Added `paimon` to parameterized test `testCreateAndLoadGenericTable`, created `NoopPaimonCatalog` mock, added Paimon format detection mock in test setup
- [x] 💡 **Added comments for complex logic** - Added Javadoc comments explaining PaimonHelper's purpose and the delegation pattern
- [ ] 🧾 **Updated CHANGELOG.md** - *To be added if required*
- [ ] 📚 **Updated documentation** - *Documentation update for site/content may be needed*

---

### Unit Tests
```bash
./gradlew :polaris-spark-3.5_2.12:test --tests "org.apache.polaris.spark.SparkCatalogTest"
```
---

## Related Issues

#3343 - Support Apache Paimon as a first-class citizen in the catalog

PR #3820, opened 2026-02-18
Integrate Paimon with Polaris without delegation mode

iting0321:
## Summary
The original implementation attempted to configure Polaris as a delegate catalog (following the Delta/Hudi integration pattern). However, this approach is not compatible with Paimon, **as Paimon's SparkCatalog does not implement the delegation model.**

Paimon's architecture is fundamentally different:

- Delta/Hudi: Extend Spark's `DelegatingCatalogExtension`, allowing them to delegate catalog operations (namespace, table listing) to another catalog while handling format-specific operations themselves.
- Paimon: Has its own `SparkCatalog` that directly manages its warehouse. 

## Solution
Instead of delegation, we implement a hybrid approach:

- Paimon operations: A Paimon catalog (NoopPaimonCatalog) handles all Paimon-specific operations directly against its own warehouse.

- Polaris registration: After Paimon creates a table, we register it as an external table in Polaris so it appears in the unified catalog view.

- Warehouse separation: Paimon tables use a separate warehouse directory (paimon-warehouse) configured via `spark.sql.catalog.<catalog>.paimon-warehouse`

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: #3723 


plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkPaimonIT.java
```diff
@@ -0,0 +1,279 @@
...
+    assertThat(results.get(1)).isEqualTo(new Object[] {2, "bobby", 31});
+    assertThat(results.get(2)).isEqualTo(new Object[] {3, "charlie", 25});
+
+    sql("DROP TABLE %s", targetTable);
+    sql("DROP TABLE %s", sourceTable);
+  }
+
+  @Test
+  public void testShowTables() {
+    // Note: Paimon tables managed by Paimon SparkCatalog are not visible via SHOW TABLES
```
flyingImer: This is odd to me, as the solution proposed in the description suggests making it appears in the unified catalog view. Which behavior is desired?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -143,7 +143,29 @@ public Table loadTable(Identifier ident) throws NoSuchTableException {
     try {
       return this.icebergsSparkCatalog.loadTable(ident);
     } catch (NoSuchTableException e) {
-      return this.polarisSparkCatalog.loadTable(ident);
+      // For generic tables, first try to load from Polaris to check the format
```
flyingImer: I found the logic is a bit tangled here - paimon loading strategy impl is a bit intrusive:
- when entering generic tables, should polaris to further delegate to other catalogs? and in which order? IMHO it worth a dev mail list discussion if not already
- lets say polaris wants to try other catalogs before returning `this.polarisSparkCatalog.loadTable(ident)`, can we simplify the impl to:
```java
String provider = table.properties().get(PolarisCatalogUtils.TABLE_PROVIDER_KEY);
if (PolarisCatalogUtils.usePaimon(provider)) {
    try {
        // For Paimon tables, use Paimon's SparkCatalog to load the table
        // This ensures proper handling of Paimon's metadata and schema files
        TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog(this.catalogName);
        return paimonCatalog.loadTable(ident);
    } catch (Exception paimonException) {
        // better paimonException to polarisException mapping
    }
} else if (...other catalog) {...}
return this.polarisSparkCatalog.loadTable(ident); // exception bubbles up as previous
```

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -186,30 +211,56 @@ public Table alterTable(Identifier ident, TableChange... changes) throws NoSuchT
...
-        return deltaCatalog.alterTable(ident, changes);
-      } else if (PolarisCatalogUtils.useHudi(provider)) {
-        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
-        return hudiCatalog.alterTable(ident, changes);
-      } else if (PolarisCatalogUtils.usePaimon(provider)) {
-        TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog(this.polarisSparkCatalog);
-        return paimonCatalog.alterTable(ident, changes);
-      } else {
-        return this.polarisSparkCatalog.alterTable(ident);
+      // Try to load from Polaris first
```
flyingImer: similarly here as well

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -186,30 +211,56 @@ public Table alterTable(Identifier ident, TableChange... changes) throws NoSuchT
...
     }
   }
 
   @Override
   public boolean dropTable(Identifier ident) {
-    return this.icebergsSparkCatalog.dropTable(ident) || this.polarisSparkCatalog.dropTable(ident);
+    boolean dropped =
+        this.icebergsSparkCatalog.dropTable(ident) || this.polarisSparkCatalog.dropTable(ident);
+
+    // Also try to drop from Paimon catalog if it exists
```
flyingImer: curious - why not using the provider to identify the paimon catalog first (just like you did for load) and when try to drop?

also, would it possible that such impl accidentally drops unrelated paimon tables? like happens to share the same identity name? (given I don't have full visibility on the other side)

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -186,30 +211,56 @@ public Table alterTable(Identifier ident, TableChange... changes) throws NoSuchT
...
+    try {
+      TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog(this.catalogName);
+      if (paimonCatalog.dropTable(ident)) {
+        dropped = true;
+      }
+    } catch (Exception e) {
```
flyingImer: nit: at minimal, we should have log some warning

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
```diff
@@ -19,75 +19,125 @@
...
+   * <p>Paimon SparkCatalog requires a warehouse path to manage table locations. This method
+   * initializes Paimon with the warehouse path configured via the paimon-warehouse option.
    *
-   * @param polarisSparkCatalog the Polaris Spark Catalog to set as delegate
+   * @param catalogName the name of the catalog
    * @return the configured Paimon TableCatalog
+   * @throws IllegalArgumentException if paimon-warehouse is not configured
    */
-  public TableCatalog loadPaimonCatalog(PolarisSparkCatalog polarisSparkCatalog) {
+  public TableCatalog loadPaimonCatalog(String catalogName) {
```
flyingImer: hmmm, should the return type be updated as well?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
```diff
@@ -19,75 +19,125 @@
...
     if (this.paimonCatalog != null) {
```
flyingImer, line 69: I'm a bit concerned about this impl in the context of concurrent calls, resulting in duplicate initialization.

Do we have a pattern already? if not, we should have a separate story to follow up

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkPaimonIT.java
```diff
@@ -0,0 +1,279 @@
...
+
+  @Override
+  protected SparkSession buildSparkSession() {
+    // Get Paimon warehouse path - use parent of Spark warehouse to create a sibling directory
+    paimonWarehouseDir = warehouseDir.resolve("../paimon_warehouse").normalize().toString();
+
+    return SparkSessionBuilder.buildWithTestDefaults()
+        .withExtensions(
+            "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,"
+                + "io.delta.sql.DeltaSparkSessionExtension,"
```
flyingImer: info: is this intentional? seems odd to me it loads delta deps for a paimon test

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
iting0321: We didn’t have this pattern. `HudiHelper` and `DeltaHelper` also have this issue. 
I can help with the follow-up.

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkPaimonIT.java
iting0321: I just removed Delta dependency. Thanks for point it out.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
iting0321: updated!

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: Fix!

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: 1. I’ve refactored `dropTable` to resolve the Paimon catalog via the provider first, consistent with how load works.
2. Added a validation to ensure the table belongs to the Paimon catalog before deleting, to avoid accidentally dropping unrelated tables with the same identifier.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: The design resolves the provider via `polarisSparkCatalog.loadGenericTable(ident)` first, then delegates based on the result, no speculative multi-catalog probing. Delegation is deterministic, it only happens once the provider is known, and keeping Polaris as the single entry point.

Happy to discuss the broader routing strategy on the dev mailing list if needed.

plugins/spark/v3.5/integration/src/intTest/java/org/apache/polaris/spark/quarkus/it/SparkPaimonIT.java
iting0321: The original implementation is contradicted the unified catalog goal, now this change introduces explicit dual registration that Paimon remains the physical owner of metadata and data files, while Polaris maintains the logical catalog entry so that listing operations are unified.

flyingImer: Hmm, have such pattern been discussed? I'm concerned about this behavior discrepancy due to per provider implementation nuance. I wonder if possible to "enable" Paimon tables managed by Paimon SparkCatalog via SHOW TABLES? what is the drawback?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
flyingImer: yea, lets start a ML thread on this. 

IMHO l if this is the pattern Polaris is going forward, we should have the spec or a rule deterministically suggests the resolving order. And technically, making them pluggable and consistent ordering across all operations

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -155,6 +185,32 @@ public Table createTable(
...
+        tableLocation = paimonHelper.resolveTableLocation(ident.namespace(), ident.name());
+      }
+      Map<String, String> registrationProps = new java.util.HashMap<>(properties);
+      registrationProps.put(TableCatalog.PROP_LOCATION, tableLocation);
+      try {
+        this.polarisSparkCatalog.createTable(ident, schema, transforms, registrationProps);
+      } catch (Exception e) {
+        LOG.warn(
+            "Failed to register Paimon table {} in Polaris catalog: {}", ident, e.getMessage());
+      }
```
flyingImer: on second thought, I would highly recommend to 1) retry limited times if retryable; 2) roll back paimon table creation and rethrow if not auto recoverable

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -183,33 +234,78 @@ public Table createTable(
...
+      String provider = this.polarisSparkCatalog.getTableFormat(ident);
+
+      if (PolarisCatalogUtils.usePaimon(provider)) {
+        // For Paimon tables, drop from both Polaris and Paimon catalog
+        boolean droppedFromPolaris = this.polarisSparkCatalog.dropTable(ident);
+        try {
+          TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog(this.catalogName);
+          paimonCatalog.dropTable(ident);
+        } catch (Exception e) {
+          LOG.warn("Failed to drop table {} from Paimon catalog: {}", ident, e.getMessage());
```
flyingImer: Current impl would make the table becomes an orphan if Polaris successfully drops but Paimon fails. We should do better on the failure mode here: 1) drop Paimon first then Polaris; 2) or drop Polaris first and roll back Polaris if Paimon drop fails

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -183,33 +234,78 @@ public Table createTable(
...
-        TableCatalog hudiCatalog = hudiHelper.loadHudiCatalog(this.polarisSparkCatalog);
-        return hudiCatalog.alterTable(ident, changes);
-      } else if (PolarisCatalogUtils.usePaimon(provider)) {
-        TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog(this.polarisSparkCatalog);
+      // Not an Iceberg table, fall through to handle as generic table
+    }
+
+    // For generic tables, check the format/provider to decide delegation
+    // Use getTableFormat to avoid triggering Spark DataSource resolution for routing decisions
+    String provider = this.polarisSparkCatalog.getTableFormat(ident);
```
flyingImer: info: it is not quite clear to me on why the pattern change from `String provider = table.properties().get(PolarisCatalogUtils.TABLE_PROVIDER_KEY)` to `String provider = this.polarisSparkCatalog.getTableFormat(ident)`, mind elaborating?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
```diff
@@ -81,13 +112,48 @@ public TableCatalog loadPaimonCatalog(PolarisSparkCatalog polarisSparkCatalog) {
...
+   *
+   * @param namespace the namespace (database) of the table
+   * @param tableName the name of the table
+   * @return the resolved table location path
+   */
+  public String resolveTableLocation(String[] namespace, String tableName) {
+    // Paimon convention: warehouse/database.db/table_name
+    String dbPath = String.join(".", namespace) + ".db";
+    return this.paimonWarehouse + "/" + dbPath + "/" + tableName;
+  }
```
flyingImer: It seems to be us poking around Paimon internals, is there any Paimon offering alternatives? like `paimonTable.properties().get(PROP_LOCATION)`?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: In this scenario, we just need to get the format (e.g. "paimon").
For the previous one will triggers full Spark DataSource resolution which need lots of times. 
But  `getTableFormat` gives the exact same routing information at a fraction of the cost.
We can extract format in
```java
public record GenericTable(
    String name,
    String format,    
    String baseLocation,
    String doc,
    Map<String, String> properties) { ... }
```

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
iting0321: This function was used in `SparkCatalog.java` to reconstruct the path manually(failback). Before this function we will call `String tableLocation = paimonTable.properties().get(PROP_LOCATION);` to take the location.

Now I remove `resolveTableLocation` and throw the error directly.
```
throw new IllegalStateException(
            String.format(
                "Paimon table %s was created but its properties do not contain a location. "
                    + "This is unexpected — Paimon should always populate the location from its CoreOptions.PATH.",
                ident));
      }
      ```

File: plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java (hunk 10 above)
Comment (iting0321) [reply plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java]: Fix for drop Paimon first then Polaris.

File: plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java (hunk 9 above)
Comment (iting0321) [reply plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java]: Fixed!

File: plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java (hunk 2 above)
Comment (iting0321) [reply plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java]: Hi @flyingImer, I just opened a [ML](https://lists.apache.org/thread/8dkv1q15f923nvrbf2fdpjhgqn9c0g4y) to discuss this problem.
Looking forward to your thoughts !

File: plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java (hunk 7 above)
Comment (gh-yzou) [reply plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java:69]: I am little bit confused about the duplicated initialization part, for one client session, the Catalog is suppose to only be initialized once since we cache the paimonCatalog created

Comment (iting0321) [reply plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java:69]: I agree with that.
I just refactored `PaimonHelper` to store the catalogName.

**Comment by dimas-b** [comment]
Thanks for your contribution, @iting0321 ! Do you have time / interest in continuing this PR? I hope we can find consensus in the comment threads :)

**Comment by iting0321** [comment]
> Thanks for your contribution, @iting0321 ! Do you have time / interest in continuing this PR? I hope we can find consensus in the comment threads :)

Yes, I'm definitely interested in continuing this PR!
Actually, I opened an issue in Paimon to ask whether they could support [DelegatingCatalogExtension](https://github.com/apache/paimon/issues/7629), but no one has replied yet. Alternatively, I’m considering implementing DelegatingCatalogExtension for Paimon myself. This would enable end-to-end functionality while avoiding the need for explicit dual registration.

File: plugins/spark/v3.5/integration/build.gradle.kts [hunk 13]
```diff
@@ -93,6 +93,25 @@ dependencies {
     exclude("org.apache.logging.log4j", "log4j-slf4j-impl")
   }
 
+  // Paimon dependency for Paimon table support
+  // Note: Paimon only publishes paimon-spark-3.5 for Scala 2.12, not 2.13
+  if (scalaVersion == "2.12") {
```
dimas-b: Would it be possible to factor Paimon tests out into a separate Gradle module to avoid `if/else` in build scritps?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
dimas-b: If these changes make sense without the new test code, it might be preferable to move them into a separate (smaller) PR to simplify review. WDYT?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -135,16 +145,42 @@ public void initialize(String name, CaseInsensitiveStringMap options) {
...
+
+    // Delegate to the appropriate catalog based on the provider
+    if (PolarisCatalogUtils.usePaimon(provider)) {
+      try {
+        // For Paimon tables, use Paimon's SparkCatalog to load the table
+        // This ensures proper handling of Paimon's metadata and schema files
+        TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog();
+        return paimonCatalog.loadTable(ident);
```
dimas-b: These look like substantial implementation changes, but the PR title mentions only "Add tests" 🤔 Could you update the PR title to be more aligned with actual code changes?

iting0321: I agree this PR has become a mix of Paimon implementation changes and integration test work, so the current title and scope are misleading. I’m planning to remove the integration-test-specific changes from this PR and narrow it to aligning the Paimon implementation with its non-delegation behavior. Once that implementation is settled, I’ll open a follow-up PR for the Paimon integration tests. If you think that sounds like the better approach, I’m happy to proceed that way.

iting0321: Hi @dimas-b, I just deleted the integration test and updated the PR title and description. Looking forward to your feedback!

dimas-b (dismissed):

plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
```diff
@@ -442,21 +443,28 @@ void testCreateAndLoadGenericTable(String format) throws Exception {
     Identifier identifier = Identifier.of(defaultNS, "generic-test-table");
     createAndValidateGenericTableWithLoad(catalog, identifier, defaultSchema, format);
 
-    Identifier[] icebergTables = catalog.listTables(defaultNS);
-    assertThat(icebergTables.length).isEqualTo(1);
-    assertThat(icebergTables[0]).isEqualTo(Identifier.of(defaultNS, "generic-test-table"));
+    // Paimon tables are managed separately by PaimonCatalog and not listed in Polaris
```
flyrain: this skips the unified-view assertion specifically for Paimon — but unified `SHOW TABLES` is the stated goal of the PR. Please add a test that asserts the Paimon table *is* visible via `catalog.listTables(defaultNS)` after create, and missing after drop.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -155,6 +191,68 @@ public Table createTable(
     String provider = properties.get(PolarisCatalogUtils.TABLE_PROVIDER_KEY);
     if (PolarisCatalogUtils.useIceberg(provider)) {
       return this.icebergsSparkCatalog.createTable(ident, schema, transforms, properties);
+    } else if (PolarisCatalogUtils.usePaimon(provider)) {
```
flyingImer: IIUC, this adds the Paimon-specific lifecycle for createTable, but the staged paths still unconditionally delegate to Iceberg. 

Since this catalog implements StagingTableCatalog, Spark routes Create Table As Select(CTAS)/Replace Table As Select(RTAS) through stageCreate/stageReplace rather than this method, so USING paimon can bypass both Paimon creation and Polaris registration.

Is this the desired behavior? Could we either route Paimon staged operations through the corresponding Paimon + Polaris lifecycle, or explicitly reject them before any side effect? It would also be good to cover CTAS/RTAS in regression tests.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -183,33 +276,104 @@ public Table createTable(
...
+    if (this.icebergsSparkCatalog.dropTable(ident)) {
+      return true;
+    }
+
+    // For generic tables, check the provider to delegate to the appropriate catalog
+    // Use getTableFormat to avoid triggering Spark DataSource resolution for routing decisions
+    try {
+      String provider = this.polarisSparkCatalog.getTableFormat(ident);
+
+      if (PolarisCatalogUtils.usePaimon(provider)) {
```
flyingImer: now that Paimon dropTable has a two-catalog lifecycle, purgeTable needs the same provider-aware treatment. Today a Paimon purge falls through to PolarisSparkCatalog.purgeTable(), which only removes the Generic Table record, leaving the physical Paimon table behind.

Should we either define the Paimon → Polaris purge lifecycle here as well, or reject Paimon purge before modifying either side? A regression test that checks both catalog states would help prevent this from drifting again IMO.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -155,6 +191,68 @@ public Table createTable(
...
+      registrationProps.put(TableCatalog.PROP_LOCATION, tableLocation);
+
+      try {
+        Tasks.foreach(registrationProps)
+            .retry(MAX_POLARIS_REGISTRATION_RETRIES)
+            .exponentialBackoff(RETRY_MIN_WAIT_MS, RETRY_MAX_WAIT_MS, RETRY_TOTAL_TIMEOUT_MS, 2.0)
+            .shouldRetryTest(
+                e ->
+                    !(e instanceof TableAlreadyExistsException)
+                        && !(e instanceof NoSuchNamespaceException))
```
flyingImer: I don't think this retry classifier can currently see the exception types it is testing for. TableAlreadyExistsException and NoSuchNamespaceException are wrapped in RuntimeException inside the task, so shouldRetryTest receives the wrapper and treats both as retryable.

Also, Tasks.retry(3) means the initial attempt plus three retries. Could we preserve/unwrap the original cause for classification and add a regression test for the attempt count and single rollback behavior?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
```diff
@@ -155,6 +191,68 @@ public Table createTable(
...
+      // so we don't require a location to be specified.
+      TableCatalog paimonCatalog = paimonHelper.loadPaimonCatalog();
+      // Ensure namespace exists in Paimon before creating table
+      paimonHelper.ensureNamespaceExists(ident.namespace());
+      Table paimonTable = paimonCatalog.createTable(ident, schema, transforms, properties);
+
+      // Register the table in Polaris as a generic table so it's visible via SHOW TABLES.
+      // Paimon's SparkTable.properties() populates "location" from CoreOptions.PATH,
+      String tableLocation = paimonTable.properties().get(TableCatalog.PROP_LOCATION);
+      if (tableLocation == null) {
```
flyingImer: The physical Paimon table has already been created before this validation, but this failure is outside the rollback-protected scope. Even if location is expected to be an invariant for the default Paimon implementation, violating that invariant should not leave a successfully created physical table behind while the overall create reports failure.

I feel we make the rollback guard cover every failure after paimonCatalog.createTable(...) succeeds, including this validation, WDYT?

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
```diff
@@ -81,13 +113,33 @@ public TableCatalog loadPaimonCatalog(PolarisSparkCatalog polarisSparkCatalog) {
...
+   *
+   * @param namespace the namespace to ensure exists
+   */
+  public void ensureNamespaceExists(String[] namespace) {
+    if (this.paimonCatalog instanceof SupportsNamespaces) {
+      SupportsNamespaces nsSupport = (SupportsNamespaces) this.paimonCatalog;
+      if (!nsSupport.namespaceExists(namespace)) {
+        try {
+          nsSupport.createNamespace(namespace, new HashMap<>());
+        } catch (Exception e) {
```
flyingImer: This currently treats every namespace-creation failure as an already-exists race. That can hide authorization, storage/I/O, or unsupported-operation failures and let table creation continue with the original cause lost.

I feel we should only ignore NamespaceAlreadyExistsException (or re-check existence for that race) and propagate other failures?

plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
```diff
@@ -442,21 +443,28 @@ void testCreateAndLoadGenericTable(String format) throws Exception {
...
+    if (!PolarisCatalogUtils.usePaimon(format)) {
```
flyingImer: This skip seems to contradict the contract introduced by this PR. Paimon owns the physical table, but the explicit Polaris registration is added specifically so the table participates in the unified catalog view/SHOW TABLES.

Could we include Paimon in the listing, duplicate-name, and post-drop assertions? As written, a regression that stops registering the Paimon table in Polaris could still pass this test.

flyingImer (commented):
Thanks for carrying the earlier review feedback through. The current version is much closer to the direction discussed on the dev list, and I’m fine treating a generalized provider router and broader idempotency improvements as follow-ups rather than blockers here.

I still don’t think this is ready to merge yet. I left a few inline comments on lifecycle correctness and failure handling that I think need to be addressed first.

One additional merge requirement is that this branch predates the Spark common-module refactor, so the implementation should be rebased and ported to the current shared Spark catalog structure, with coverage for both Spark 3.5 and 4.0.

From my side, the merge bar is:

- address the lifecycle/failure-handling issues called out inline;
- add regression coverage for the Paimon unified-catalog behavior introduced by this PR;
- rebase/port the implementation to the current shared Spark catalog structure and verify both supported Spark versions;
- document the required Paimon configuration and any intentionally unsupported operations.

The more general routing abstraction can remain follow-up work.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: Thanks for your feedbacks!

Now the Paimon stageCreate, stageReplace, and stageCreateOrReplace now fail before Iceberg/Paimon side effects.
And I also add the corresponding regression test for SQL CTAS/RTAS.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: I added provider-aware Paimon purge routing : purge the physical Paimon table first, then remove the Polaris registration. 
Failures preserve or clearly report inconsistent state. Regression tests cover both catalog states and failure paths.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: I preserved the original exception types so the retry classifier now handles them correctly. I also added tests verifying one attempt for non retryable failures, four attempts for retryable failures,and a single rollback.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/SparkCatalog.java
iting0321: I agree with your points. I widened the rollback scope to cover every failure after physical Paimon creation. And added the  regression test for the missing location case.

plugins/spark/v3.5/spark/src/main/java/org/apache/polaris/spark/utils/PaimonHelper.java
iting0321: Agreed. I narrowed the catch to NamespaceAlreadyExistsException and propagate all other failures.

plugins/spark/v3.5/spark/src/test/java/org/apache/polaris/spark/SparkCatalogTest.java
iting0321: I removed the Paimon exclusions so the test now verifies listing, duplicate name rejection, and post drop cleanup for Paimon .

iting0321:
> Thanks for carrying the earlier review feedback through. The current version is much closer to the direction discussed on the dev list, and I’m fine treating a generalized provider router and broader idempotency improvements as follow-ups rather than blockers here.
> 
> I still don’t think this is ready to merge yet. I left a few inline comments on lifecycle correctness and failure handling that I think need to be addressed first.
> 
> One additional merge requirement is that this branch predates the Spark common-module refactor, so the implementation should be rebased and ported to the current shared Spark catalog structure, with coverage for both Spark 3.5 and 4.0.
> 
> From my side, the merge bar is:
> 
> * address the lifecycle/failure-handling issues called out inline;
> * add regression coverage for the Paimon unified-catalog behavior introduced by this PR;
> * rebase/port the implementation to the current shared Spark catalog structure and verify both supported Spark versions;
> * document the required Paimon configuration and any intentionally unsupported operations.
> 
> The more general routing abstraction can remain follow-up work.

Hi @flyingImer, thanks for the review!

I already ported the implementation to the shared Spark catalog module used by both Spark 3.5 and 4.0. 
And also documented the required Paimon warehouse configuration and unsupported staged/rename operations in README.md.
For the regression test, I added unified-catalog, rollback, retry, purge, and CTAS/RTAS regression coverage for both Spark versions.

plugins/spark/common/src/main/java/org/apache/polaris/spark/BasePolarisSparkCatalog.java
```diff
@@ -222,11 +351,47 @@ public void invalidateTable(Identifier ident) {
 
   @Override
   public boolean purgeTable(Identifier ident) {
-    if (this.icebergsSparkCatalog.purgeTable(ident)) {
-      return true;
-    } else {
-      return this.polarisSparkCatalog.purgeTable(ident);
+    String provider;
+    try {
+      provider = this.polarisSparkCatalog.getTableFormat(ident);
```
flyingImer, line 360: IIUC, this makes Iceberg purge depend on Generic Table support. With `ENABLE_GENERIC_TABLES=false`, getTableFormat calls PolarisRESTCatalog.loadGenericTable, whose endpoint check throws UnsupportedOperationException. Since this catch only handles NoSuchTableException, an existing Iceberg table can no longer be purged, although the previous Iceberg-first path supported it. Could we preserve Iceberg purge when Generic Tables are disabled, keep the Paimon cleanup ordering, and add a regression for that configuration?

flyingImer (commented):
Thanks for the update. The earlier six concerns, shared-module port, and documentation requests are addressed. I still see one regression affecting an existing supported deployment configuration, detailed in the new inline, so I'd hold off on merging until that is fixed and covered. The broader routing and idempotency work can remain follow-ups.

plugins/spark/common/src/main/java/org/apache/polaris/spark/BasePolarisSparkCatalog.java
iting0321: Hi @flyingImer, I just let Iceberg purge when Generic Tables are disabled and add corresponding tests. Thanks!!

PR #3826, opened 2026-02-19
Add API Spec for Generic Table credential vending

gh-yzou:
Add API Spec to support credential vending for Generic Table:
1) Add storage-credentials filed in LoadGenericTablesResponse
2) docs all supported storage credentials in description

## Checklist

### DIFF
--- a/spec/polaris-catalog-apis/generic-tables-api.yaml
+++ b/spec/polaris-catalog-apis/generic-tables-api.yaml
@@ -276,7 +276,7 @@ components:
           description: Indicates a storage location prefix where the configuration is relevant. Clients should choose the most 
             specific prefix (by selecting the longest common prefix) if several configurations of the same type are available.
             If multiple configurations share the same longest matching prefix, the client may select any of them. However, 
-            a well-designed server should avoid generating multiple configurations for the same prefix.
+            a server should avoid generating multiple configurations for the same prefix.
         config:
           type: object
           description: |


spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,55 @@ components:
...
+            specific prefix (by selecting the longest prefix) if several credentials of the same type are available.
+        config:
+          type: object
+          description: |
+            Credential configurations for AWS S3, GCP GCS, and Azure ADLS are supported. The following outlines 
+            the currently supported configuration options:
+              
+            ## AWS Configurations
+              
+            The following configurations should be respected when working with tables stored in AWS S3
```
dimas-b: Is "AWS" critical here? There are many S3-compatible implementations.

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,55 @@ components:
...
+              - `s3.access-key-id`: id for credentials that provide access to the data in S3
+              - `s3.secret-access-key`: secret for credentials that provide access to data in S3
+              - `s3.session-token`: if present, this value should be used for as the session token
+              - `s3.session-token-expires-at-ms`: the time the aws session token expires, in milliseconds
+            Extra properties:
+              - `s3.endpoint`: the S3 endpoint to use for requests
```
dimas-b: `endpoint` is technically not a "credential", which is implied by `StorageAccessCredential`. (also applies to other properties below).

Could we structure the API in a way that clearly separates credentials from other configuration?

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,55 @@ components:
...
+            
+            # AZURE Configuration
+            
+            The following configurations should be respected when working with tables stored in AZURE ADLS
+              - `adls.sas-token.<hostname>`: an azure shared access signature token
+              - `adls.sas-token-expires-at-ms.<hostname>`: the expiration time for the access token, in milliseconds
+            Extra properties:
+              - `adls.refresh-credentials-endpoint`: the endpoint to load vended credentials for a table from the catalog
+
+          additionalProperties:
```
dimas-b, line 310: ~~What is the purpose of `additionalProperties`?~~

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,55 @@ components:
...
+          type: object
+          description: |
+            Credential configurations for AWS S3, GCP GCS, and Azure ADLS are supported. The following outlines 
+            the currently supported configuration options:
+              
+            ## AWS Configurations
+              
+            The following configurations should be respected when working with tables stored in AWS S3
+              - `s3.access-key-id`: id for credentials that provide access to the data in S3
+              - `s3.secret-access-key`: secret for credentials that provide access to data in S3
```
dimas-b, line 290: Would it be possible to use declared OpenAPI objec properties for this data, as opposed to relying of a naming convention? I mean something like the `base-location` property on line 230.

spec/polaris-catalog-apis/generic-tables-api.yaml
snazy: I agree, AWS is one vendor, while S3 is used by multiple open source projects and vendors.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: I don’t think the cloud-specific configuration fields are good candidates to be extracted into dedicated fields. Different cloud providers rely on different credential structures, and they regularly introduce new authentication methods or storage access patterns. Keeping these settings in a general configuration map makes it much easier to support new configurations without requiring schema changes.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: that is good point, I updated to S3 configuration to be more precise. Also updated other ones to contain the storage type

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: That is a good point, it seems like those fields are really needed by iceberg FILE IO today. I think for generic table, we don't necessary need this. I can take if off from generic for now, and if we really need those access specific configuration, we can have a separate discussion to see if we need to evolve a separate structure for those information

dimas-b: Fair enough :+1:

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: That's fair.

WDYT about adding a docs page for this aspect of the spec? For example under https://polaris.apache.org/in-dev/unreleased/polaris-api-specs/ ?

Listing specific properties in Open API YAML comments is helpful, but is not normative from the Open API perspective (a change in a comment does not materially affect how the spec file is processes by automated tools). At the same time a proper docs page is probably easier for humans to read and incorporate into their programs. The Open API spec could refer to the docs page for the exact property definitions.

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,53 @@ components:
...
+        - config
+      properties:
+        prefix:
+          type: string
+          description: Indicates a storage location prefix where the credential is relevant. Clients should choose the most 
+            specific prefix (by selecting the longest prefix) if several credentials of the same type are available.
+        config:
+          type: object
+          description: |
+            Credential configurations for AWS S3, GCP GCS, and Azure ADLS are supported. The following outlines
```
dimas-b: I'd like to further generalize "AWS S3" to "AWS S3 and compatible systems", please.

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,53 @@ components:
...
+              
+            ## S3 Configurations
+              
+            The following configurations should be respected when working with tables stored in AWS S3
+              - `s3.access-key-id`: id for credentials that provide access to the data in S3
+              - `s3.secret-access-key`: secret for credentials that provide access to data in S3
+              - `s3.session-token`: if present, this value should be used for as the session token
+              - `s3.session-token-expires-at-ms`: the time the aws session token expires, in milliseconds
+            Extra properties:
+              - `client.region`: region to configure client for making requests to AWS
```
dimas-b: As discussed in a previous comment about `endpoint`, `client.region` and `client.refresh-credentials-endpoint` are not themselves credentials. Reporting them under `StorageAccessCredential` is confusing and may lead to obscure assumptions on the client side.

I think it is fine to expose these properties to clients, but I believe we ought to structure the API in a way to clearly separate sensitive data from general configuration.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: yes, adding release doc is one of the task in supporting of credential vending, I plan to do that once the generic table credential vending work is done.  For this PR, i would like to keep it as spec change, i will make sure we have proper release doc before we announce credential vending is supported for generic table, wdyt?

dimas-b: Actually, I think adding new properties requires a Polaris Server change anyway. If  that is a given, additionally updating object types in Open API is not that hard, IMHO. I wonder what other people think. I'll post this to the `dev` ML for awareness.

spec/polaris-catalog-apis/generic-tables-api.yaml
adutra: I confess I'm torn on this 😅 .

That would require to mimic Iceberg even more and declare a `LoadGenericTableResponse.config` property similar to this:

https://github.com/apache/iceberg/blob/a5ef2392bb8aeed9730ff1937cc515279200a5a7/open-api/rest-catalog-open-api.yaml#L3503-L3506

But tbh, I've always found Iceberg's `LoadTableResult` quite confusing as well, with its two `config` fields: `LoadTableResult.config` (for non-credential stuff) and `LoadTableResult.storage-credentials.config` (for the credential stuff).

adutra: To clarify my point above: I think it's a good thing to distinguish two levels of config (table-wide vs prefix-specific), like Iceberg does, in spite of the fact that the two config levels have boundaries that are not always 100% clear.

dimas-b: It looks like this discussion has a parallel track  on the `dev` ML too :sweat_smile: 

https://lists.apache.org/thread/44xfrnvoq5720cgjfnobnyp3xn1w8tvv

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,53 @@ components:
           items:
             $ref: '../iceberg-rest-catalog-open-api.yaml#/components/schemas/TableIdentifier'
 
+
+    StorageAccessCredential:
```
flyingImer: info: given `StorageCredential` in iceberg-rest-catalog-open-api.yaml:3254-3267, why do we choose `StorageAccessCredential` over `StorageCredential`? Without further context, IMHO keeping consistency may be better

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,53 @@ components:
...
+              
+            ## GCP GCS Configurations
+            
+            The following configurations should be respected when working with tables stored in GCP GCS
+              - `gcs.oauth2.token`: the gcs scoped access token
+              - `gcs.oauth2.token-expires-at`: the time the gcs access token expires, in milliseconds
+            Extra properties:
+              - `gcs.oauth2.refresh-credentials-endpoint`: the endpoint to load vended credentials for a table from the catalog
+            
+            # AZURE ADLS Configuration
```
flyingImer: nit: 
```suggestion
            ## AZURE ADLS Configuration
```

flyingImer (commented):
Wonder `bundled-polaris-catalog-service.yaml` should be re-generated as instructed in spec/README.md:56?

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +262,53 @@ components:
           items:
             $ref: '../iceberg-rest-catalog-open-api.yaml#/components/schemas/TableIdentifier'
 
+
+    StorageAccessCredential:
+      type: object
+      required:
+        - prefix
```
dimas-b, line 271: Does this `prefix` apply to non-credential config properties too?

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: updated

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: Based on the Iceberg thread (https://lists.apache.org/thread/44xfrnvoq5720cgjfnobnyp3xn1w8tvv
), it appears that, for historical reasons, Iceberg currently has overlapping configurations between StorageCredentials and config. Initializing a client requires merging both config and StorageCredentials.

Following recent discussions in the Polaris thread (https://lists.apache.org/thread/9y0xnjfn4n06svkt657gcsplk6sc7dhf
), we have decided to simplify things in Polaris by maintaining a single configuration map that contains all necessary storage access configurations, and renaming StorageCredentials to StorageAccessConfig for clarity.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: Based on the discussion over thread, the prefix also apply to non-credential configs.

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +264,53 @@ components:
...
+    StorageAccessConfig:
+      type: object
+      required:
+        - prefix
+        - config
+      properties:
+        prefix:
+          type: string
+          description: Indicates a storage location prefix where the configuration is relevant. Clients should choose the most 
+            specific prefix (by selecting the longest prefix) if several configurations of the same type are available.
```
dimas-b: for the sake of clarity, please specify the behavior when several prefixes have the same length.

I suggest using the one that goes first in the list of config objects.

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: Hmmm... this line still mentions "AWS" 🤔 I'd propose: `[...] tables stored in S3-compatible systems` (also on line 281).

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +264,53 @@ components:
...
+            the currently supported configuration options:
+              
+            ## S3 Configurations
+              
+            The following configurations should be respected when working with tables stored in AWS S3
+              - `s3.access-key-id`: id for credentials that provide access to the data in S3
+              - `s3.secret-access-key`: secret for credentials that provide access to data in S3
+              - `s3.session-token`: if present, this value should be used for as the session token
+              - `s3.session-token-expires-at-ms`: the time the aws session token expires, in milliseconds
+            Extra properties:
```
dimas-b: What makes there properties "extra"? Do we even have to categorize properties into "main" and "extra"?

I propose describing all properties uniformly because clients will need to process them all.

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b, line 1: Side note: This spec file is not recognized as valid by SwaggerHub 🤷 

`Unable to render editor content` from https://editor.swagger.io/?url=https://raw.githubusercontent.com/apache/polaris/4a2490d7a9b0c491bb722e4e16fb83478e262f38/spec/polaris-catalog-apis/generic-tables-api.yaml

This is not specific to current PR, so we can address this later.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: I updated the sentence to "tables stored on S3, including AWS S3, S3-Compatible systems", and uses S3 for all following description, so that we don't have to repeat AWS S3, S3-Compatible systems everywhere

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: If there are configurations for the same prefix, i think client should be free to choose any of them, given that they are all valid. I felt explicitly stating the first one is over restricted to me. 

Added more clarification for such case.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: make sense. removed the "extra properties"

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: updated to "S3" to be more general, and latter description for actual storage config stats that S3 includes both AWS s3 and S3-compatible system

gh-yzou:
Hi @dimas-b Thanks a lot for your help! I’ve updated the PR to address the comments. Please take another look when you get a chance.

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +264,51 @@ components:
...
+      type: object
+      required:
+        - prefix
+        - config
+      properties:
+        prefix:
+          type: string
+          description: Indicates a storage location prefix where the configuration is relevant. Clients should choose the most 
+            specific prefix (by selecting the longest common prefix) if several configurations of the same type are available.
+            If multiple configurations share the same longest common prefix, the client may select any of them.
```
dimas-b: ` client may select any of them` - I do not think it's a good idea, TBH. The server cannot ensure predicable behaviour in this case.

The idea of a spec is to align all implementation of functional aspects so that behaviours can reasonably be extrapolated across a wide range of systems.

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: thx!

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: :+1:

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: thx!

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: From my perspective, the server’s responsibility is to ensure that it produces valid credentials for accessing the table data. As long as the configurations are valid for accessing files under the prefi, it should not matter which configuration the client chooses to use.
Specifying whether the client uses the first or the last configuration matches does not change the server’s responsibility.

Furthermore, a well-designed server should not produce multiple configurations for the same prefix, as this can be confusing for clients regardless of which one they select.

dimas-b: > a well-designed server should not produce multiple configurations for the same prefix [...]

Then we should state this in the spec. (I agree)

gh-yzou: sg! Added

dimas-b (dismissed):

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: thx - LGTM :+1:

adutra (dismissed):

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -256,6 +264,52 @@ components:
...
+      required:
+        - prefix
+        - config
+      properties:
+        prefix:
+          type: string
+          description: Indicates a storage location prefix where the configuration is relevant. Clients should choose the most 
+            specific prefix (by selecting the longest common prefix) if several configurations of the same type are available.
+            If multiple configurations share the same longest matching prefix, the client may select any of them. However, 
+            a well-designed server should avoid generating multiple configurations for the same prefix.
```
flyrain: nit:
```suggestion
            a server should avoid generating multiple configurations for the same prefix.
```

flyrain (dismissed):
LGTM. Thanks @gh-yzou !

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: updated

flyrain (approved):

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: Note: this update was made after the VOTE on `dev`: https://lists.apache.org/thread/n8nfccb61zb2gy3ktpw8rbrt2yoryo4c

I do not mind this specific change, but what is the point of calling a formal vote if changes are interleaved with the voting process? A vote is supposed to to approve a specific change. How do we know the votes cast before an update to the PR are still relevant and their authors do not want to reconsider?

flyrain: Good catch, @dimas-b. It's a great reminder for the community on how to handle the intersection of PR polish and formal Apache votes. Keeping the 'source of truth' stable during a vote is definitely key to ensuring everyone is actually on the same page.

gh-yzou: I made this change because it's a minor wording update and doesn't conflict with any of the existing comments. 
For larger changes, I agree that we should definitely call them out. However, for small adjustments such as wording improvements, we should remain flexible; otherwise, the process can become unnecessarily tedious.

gh-yzou: I can also have a PR to add that word back if that is preferred.

dimas-b: No worries about this particular change (from my POV). I just wanted to highlight the procedures.

PR #3761, opened 2026-02-14
Allow CatalogAdmin to list Principal Roles

vignesh-manel:
Implements automatic principal role listing for `catalog_admin` users via a new system-managed `catalog_role_manager` role. Fixes #363

## Implementation

   - `catalog_role_manager` created at bootstrap with `PRINCIPAL_ROLE_LIST` privilege (read-only)
   -  Automatically granted to principals when they receive `catalog_admin` on any catalog
   -  Automatically revoked when all `catalog_admin` grants are removed

## Limitations

   - A new system role is introduced just to grant PRINCIPAL_ROLE_LIST for catalog_admin
   - Principal must be assigned to principal role before granting `catalog_admin`. If assigned after, revoke and re-grant `catalog_admin` to trigger auto-grant.
   - No backfill for existing `catalog_admin` grants (requires manual grant or re-grant)
   
   CC: @collado-mike 

## Checklist
- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [x] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #361, opened 2024-10-09
Update privilege model to support granting CatalogRole access to PrincipalRoles

collado-mike:
# Description

This updates the authorization model to allow `catalog_admin` roles to grant other CatalogRoles within their catalog to a PrincipalRole. This addresses https://github.com/apache/polaris/issues/359 , in which it is stated that neither the `catalog_admin` or the `service_admin` roles alone are sufficient to grant catalog access to principal roles.

The fix is implemented by removing the requirement for `PRINCIPAL_ROLE_MANAGE_GRANTS_FOR_GRANTEE` privilege from the `ASSIGN_CATALOG_ROLE_TO_PRINCIPAL_ROLE` operation. Implicitly, this states that a user only needs privileges to manage the grants on the catalog as a securable, but does not need privilege to manage the grants on a PrincipalRole as a grantee.

A caller _does_ require `PRINCIPAL_ROLE_MANAGE_GRANTS_FOR_GRANTEE` to revoke a privilege, which, admittedly, does make this operation a little bit lopsided. However, having the ability to create grants does not implicitly mean the user has the privilege to revoke grants for the same securable. For example, a `service_admin` may delegate `catalog_admin` to another user/role, but may wish to retain `catalog_admin` privileges. Without requiring `PRINCIPAL_ROLE_MANAGE_GRANTS_FOR_GRANTEE`, a `catalog_admin` could revoke grants from any PrincipalRole, even if the grants weren't originally created by the `catalog_admin` itself.

For consistency, I also removed the corresponding `*_MANAGE_GRANTS_FOR_GRANTEE` requirements for other `ASSIGN*` operations. So a user, having privilege to manage grants on a securable, can create a grant for that securable for any grantee without needing special privilege on that grantee. As an example, a user having `CATALOG_MANAGE_ACCESS` at a namespace level can create grants for any entity at that namespace or below and grant it to any CatalogRole in the catalog. The revoke authorization is also consistent with the `ASSIGN_CATALOG_ROLE_TO_PRINCIPAL_ROLE` operation.

I confirmed with tests that the `catalog_admin` and other roles cannot create grant records that cross catalogs (i.e., can't grant `CREATE_TABLE` privilege to a catalog role in another catalog). 

Fixes #359

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] New feature (non-breaking change which adds functionality)
[...]

PR #3852, opened 2026-02-20
Allow CatalogAdmin to list Principal Roles

### LINKED ISSUE #363: [FEATURE REQUEST] CatalogAdmin cannot list principal roles

By default, a principal that has a `catalog_admin` role in a catalog cannot list principal roles using the API. The `catalog_admin`'s responsibility is managing privileges and access to the catalog roles in a catalog. After fixing https://github.com/apache/polaris/issues/359 , the `catalog_admin` has the ability to grant a catalog role to a principal role, but has no ability to see the list of available principal roles. This may be ok for cases where a principal has both the `service_admin` and the `catalog_admin` roles, but if there's an enforced separation, the lack of privilege to list principal roles is a hindrance.

The authorization model typically requires a catalog to be in the scope of a request in order to detect that user has `catalog_admin` on the specified catalog. However, PrincipalRoles are not tied to a catalog, so it's difficult for the current authorization workflow to know if the current user does have admin privilege on any catalog. We can consider a separate API, where the `/principal_roles` endpoint is prefixed by catalog, but that feels cumbersome - especially if the caller is an admin on multiple catalogs.

A more likely solution would be to manage a special PrincipalRole that has limited privileges on PrincipalRoles (and possibly Principals) that a user is automatically added to when granted `catalog_admin`. 

_No response_

### Additional context

_No response_

vignesh-manel:
Duplicate of https://github.com/apache/polaris/pull/3761 which got closed accidentally while rebasing 

Implements automatic principal role listing for `catalog_admin` users via a new system-managed `catalog_role_manager` role. Fixes #363

## Implementation

   - `catalog_role_manager` created at bootstrap with `PRINCIPAL_ROLE_LIST` privilege (read-only)
   -  Automatically granted to principals when they receive `catalog_admin` on any catalog
   -  Automatically revoked when all `catalog_admin` grants are removed

## Limitations

   - A new system role is introduced just to grant PRINCIPAL_ROLE_LIST for catalog_admin
   - Principal must be assigned to principal role before granting `catalog_admin`. If assigned after, revoke and re-grant `catalog_admin` to trigger auto-grant.
   - No backfill for existing `catalog_admin` grants (requires manual grant or re-grant)
   
   CC: @collado-mike 

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
@@ -104,6 +104,23 @@ public class AuthBootstrapUtil {
         rootContainer,
         PolarisPrivilege.SERVICE_MANAGE_ACCESS);
 
+    // create the catalog_role_manager principal role for catalog admins to list principal roles
+    PrincipalRoleEntity catalogRoleManagerPrincipalRole =
+        new PrincipalRoleEntity.Builder()
+            .setId(generateId(metaStoreManager, ctx))
+            .setName(PolarisEntityConstants.getNameOfCatalogRoleManagerPrincipalRole())
+            .setCreateTimestamp(System.currentTimeMillis())
+            .build();
+    metaStoreManager.createEntityIfNotExists(ctx, null, catalogRoleManagerPrincipalRole);
+
+    // grant PRINCIPAL_ROLE_LIST on the rootContainer to the catalogRoleManagerPrincipalRole
+    metaStoreManager.grantPrivilegeOnSecurableToRole(
+        ctx,
+        catalogRoleManagerPrincipalRole,
+        null,
+        rootContainer,
+        PolarisPrivilege.PRINCIPAL_ROLE_LIST);
+
     return metaStoreManager.loadPrincipalSecrets(ctx, rootPrincipal.getClientId());
   }
 
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
@@ -121,6 +121,67 @@ public class AuthBootstrapUtil {
     return metaStoreManager.loadPrincipalSecrets(ctx, rootPrincipal.getClientId());
   }
 
+  /**
+   * Ensures the principal_role_viewer role exists in an already-bootstrapped realm. This is used
+   * for upgrade migrations to add the role to realms that were bootstrapped before this role was
+   * introduced.
+   *
+   * <p>This method is idempotent - it checks if the role already exists and only creates it if
+   * missing.
+   *
+   * @param metaStoreManager the metastore manager for the realm
+   * @param ctx the call context for the realm
+   */
+  public static void ensurePrincipalRoleViewerExists(
+      PolarisMetaStoreManager metaStoreManager, PolarisCallContext ctx) {
+    // Check if the role already exists
+    EntityResult existingRoleResult =
+        metaStoreManager.readEntityByName(
+            ctx,
+            null,
+            PolarisEntityType.PRINCIPAL_ROLE,
+            PolarisEntitySubType.NULL_SUBTYPE,
+            PolarisEntityConstants.getNameOfPrincipalRoleViewerRole());
+
+    if (existingRoleResult.isSuccess() && existingRoleResult.getEntity() != null) {
+      // Role already exists, nothing to do
+      LOGGER.info("principal_role_viewer role already exists, skipping creation");
+      return;
+    }
+
+    LOGGER.info(
+        "principal_role_viewer role not found. Creating it for existing installation upgrade.");
+
+    // Load the root container for granting privileges
+    EntityResult rootContainerResult =
+        metaStoreManager.readEntityByName(
+            ctx,
+            null,
+            PolarisEntityType.ROOT,
+            PolarisEntitySubType.NULL_SUBTYPE,
+            PolarisEntityConstants.getRootContainerName());
+
+    Preconditions.checkState(
+        rootContainerResult.isSuccess() && rootContainerResult.getEntity() != null,
+        "Root container not found - realm may not be bootstrapped");
+    PolarisBaseEntity rootContainer = rootContainerResult.getEntity();
+
+    // Create the principal_role_viewer role
+    PrincipalRoleEntity principalRoleViewer =
+        new PrincipalRoleEntity.Builder()
+            .setId(generateId(metaStoreManager, ctx))
+            .setName(PolarisEntityConstants.getNameOfPrincipalRoleViewerRole())
+            .setCreateTimestamp(System.currentTimeMillis())
+            .build();
+    metaStoreManager.createEntityIfNotExists(ctx, null, principalRoleViewer);
+
+    // Grant PRINCIPAL_ROLE_LIST privilege on the root container
+    metaStoreManager.grantPrivilegeOnSecurableToRole(
+        ctx, principalRoleViewer, null, rootContainer, PolarisPrivilege.PRINCIPAL_ROLE_LIST);
+
+    LOGGER.info("Successfully created principal_role_viewer role for upgrade migration");
+  }
+
   private static long generateId(PolarisMetaStoreManager metaStoreManager, PolarisCallContext ctx) {
     GenerateEntityIdResult res = metaStoreManager.generateNewEntityId(ctx);
     Preconditions.checkState(res.isSuccess(), "Unable to generate id for polaris entity");
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
@@ -33,6 +33,7 @@ import org.apache.polaris.core.entity.PrincipalEntity;
 import org.apache.polaris.core.entity.PrincipalRoleEntity;
 import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
 import org.apache.polaris.core.persistence.dao.entity.CreatePrincipalResult;
+import org.apache.polaris.core.persistence.dao.entity.EntityResult;
 import org.apache.polaris.core.persistence.dao.entity.GenerateEntityIdResult;
 import org.apache.polaris.core.persistence.dao.entity.PrincipalSecretsResult;
 import org.slf4j.Logger;
@@ -104,22 +105,18 @@ public class AuthBootstrapUtil {
         rootContainer,
         PolarisPrivilege.SERVICE_MANAGE_ACCESS);
 
-    // create the catalog_role_manager principal role for catalog admins to list principal roles
-    PrincipalRoleEntity catalogRoleManagerPrincipalRole =
+    // create the principal_role_viewer role for catalog admins to list principal roles
+    PrincipalRoleEntity principalRoleViewer =
         new PrincipalRoleEntity.Builder()
             .setId(generateId(metaStoreManager, ctx))
-            .setName(PolarisEntityConstants.getNameOfCatalogRoleManagerPrincipalRole())
+            .setName(PolarisEntityConstants.getNameOfPrincipalRoleViewerRole())
             .setCreateTimestamp(System.currentTimeMillis())
             .build();
-    metaStoreManager.createEntityIfNotExists(ctx, null, catalogRoleManagerPrincipalRole);
+    metaStoreManager.createEntityIfNotExists(ctx, null, principalRoleViewer);
 
-    // grant PRINCIPAL_ROLE_LIST on the rootContainer to the catalogRoleManagerPrincipalRole
+    // grant PRINCIPAL_ROLE_LIST on the rootContainer to the principalRoleViewer
     metaStoreManager.grantPrivilegeOnSecurableToRole(
-        ctx,
-        catalogRoleManagerPrincipalRole,
-        null,
-        rootContainer,
-        PolarisPrivilege.PRINCIPAL_ROLE_LIST);
+        ctx, principalRoleViewer, null, rootContainer, PolarisPrivilege.PRINCIPAL_ROLE_LIST);
 
     return metaStoreManager.loadPrincipalSecrets(ctx, rootPrincipal.getClientId());
   }
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
@@ -43,6 +43,9 @@ public class PolarisEntityConstants {
   // the name of the principal role we create to manage the entire Polaris service
   private static final String ADMIN_PRINCIPAL_ROLE_NAME = "service_admin";
 
+  // the name of the principal role for catalog admins to list principal roles
+  private static final String CATALOG_ROLE_MANAGER_PRINCIPAL_ROLE_NAME = "catalog_role_manager";
+
   // 24 hours retention before purging. This should be a config
   private static final long RETENTION_TIME_IN_MS = 24 * 3600_000;
 
@@ -89,6 +92,10 @@ public class PolarisEntityConstants {
 
   public static String getNameOfPrincipalServiceAdminRole() {
     return ADMIN_PRINCIPAL_ROLE_NAME;
+  }
+
+  public static String getNameOfCatalogRoleManagerPrincipalRole() {
+    return CATALOG_ROLE_MANAGER_PRINCIPAL_ROLE_NAME;
   }
 
   public static long getRetentionTimeInMs() {
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
@@ -44,7 +44,7 @@ public class PolarisEntityConstants {
   private static final String ADMIN_PRINCIPAL_ROLE_NAME = "service_admin";
 
   // the name of the principal role for catalog admins to list principal roles
-  private static final String CATALOG_ROLE_MANAGER_PRINCIPAL_ROLE_NAME = "catalog_role_manager";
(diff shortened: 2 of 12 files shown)


dimas-b:
@vignesh-manel : please avoid opening multiple PRs for the same feature (cf. https://polaris.apache.org/community/contributing-guidelines/)

You should be able to take the commits from this PR, put them into your local `vignesh-manel:main` branch and force-push into #3761

Once that PR is merged, you'll be able to reuse `main` for a different purpose locally :wink:

vignesh-manel:
> @vignesh-manel : please avoid opening multiple PRs for the same feature (cf. https://polaris.apache.org/community/contributing-guidelines/)
> 
> You should be able to take the commits from this PR, put them into your local `vignesh-manel:main` branch and force-push into #3761
> 
> Once that PR is merged, you'll be able to reuse `main` for a different purpose locally :wink: 

@dimas-b I had tried that, I had forced push the same commits to the main branch of my fork, even though they showed up in the fork,  GitHub wasn't giving me the option to reopen the old PR. Seems to be an issue with GitHub  https://github.com/isaacs/github/issues/361 hence I created this new PR

dimas-b:
@vignesh-manel : no worries, let's continue on this PR :+1:

dimas-b:
`dev` thread for reference: https://lists.apache.org/thread/ws0blghsv8jl9rbwpgfgcbzjs7d38242

dimas-b:
@vignesh-manel : just heads up: I'm willing to facilitate the progress of this PR, but I do not have enough context for the internal Polaris RBAC system and its applications for an approval. 

I hope @dennishuo, @collado-mike, @flyrain could provide feedback on the actual behaviour changes.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
```diff
@@ -43,6 +43,9 @@ public class PolarisEntityConstants {
   // the name of the principal role we create to manage the entire Polaris service
   private static final String ADMIN_PRINCIPAL_ROLE_NAME = "service_admin";
 
+  // the name of the principal role for catalog admins to list principal roles
+  private static final String CATALOG_ROLE_MANAGER_PRINCIPAL_ROLE_NAME = "catalog_role_manager";
```
flyingImer: should it be protected from being droppable?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1534,8 +1544,17 @@ public PrivilegeResult revokeCatalogRoleFromPrincipalRole(
...
-    return metaStoreManager.revokeUsageOnRoleFromGrantee(
-        getCurrentPolarisContext(), catalogEntity, catalogRoleEntity, principalRoleEntity);
+    PrivilegeResult result =
+        metaStoreManager.revokeUsageOnRoleFromGrantee(
+            getCurrentPolarisContext(), catalogEntity, catalogRoleEntity, principalRoleEntity);
+
+    // if revoking catalog_admin, check if principal still has catalog_admin on other catalogs
+    if (result.isSuccess()
+        && PolarisEntityConstants.getNameOfCatalogAdminRole().equals(catalogRoleName)) {
+      revokeCatalogRoleManagerIfNeeded(principalRoleEntity);
```
flyingImer, line 1594: I see the `catalog_role_manager` is revoked only via `revokeCatalogRoleFromPrincipalRole`, but `catalog_role_manager` is granted to principal (not principal roles). Say when a principal is unassigned from a principal role (i.e., revoke principal role from principal API), `catalog_role_manager` granted to the principal would be never revoked.

I feel it should be cleaned up as well to avoid leaking PRINCIPAL_ROLE_LIST privilege

flyingImer: Similarly, when catalog drops, it should be cleaned up

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -2335,4 +2354,195 @@ private static PolarisEntitySubType selectEntitySubType(List<PolarisEntitySubTyp
...
+    EntityResult catalogRoleManagerResult =
+        metaStoreManager.readEntityByName(
+            getCurrentPolarisContext(),
+            null,
+            PolarisEntityType.PRINCIPAL_ROLE,
+            PolarisEntitySubType.NULL_SUBTYPE,
+            PolarisEntityConstants.getNameOfCatalogRoleManagerPrincipalRole());
+
+    if (!catalogRoleManagerResult.isSuccess() || catalogRoleManagerResult.getEntity() == null) {
+      return;
```
flyingImer, line 2422: 2 cents: log something when nothing found. This should be helpful to capture the missing catalog_role_manager role, as currently it is created only during bootstrap IIUC, i.e., we don't have upgrade/backfill path for already bootstrapped realms. I'm fine with temporary gaps for upgrade and backfill, but strongly recommend at least to warn on role not found

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -2335,4 +2354,195 @@ private static PolarisEntitySubType selectEntitySubType(List<PolarisEntitySubTyp
...
```
flyingImer, line 2492: 2 cents: log something when nothing found

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
flyingImer: we should have tests capturing them

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1518,8 +1519,17 @@ public PrivilegeResult assignCatalogRoleToPrincipalRole(
...
-    return metaStoreManager.grantUsageOnRoleToGrantee(
-        getCurrentPolarisContext(), catalogEntity, catalogRoleEntity, principalRoleEntity);
+    PrivilegeResult result =
+        metaStoreManager.grantUsageOnRoleToGrantee(
+            getCurrentPolarisContext(), catalogEntity, catalogRoleEntity, principalRoleEntity);
+
+    // if granting catalog_admin, also grant catalog_role_manager to allow listing principal roles
+    if (result.isSuccess()
+        && PolarisEntityConstants.getNameOfCatalogAdminRole().equals(catalogRoleName)) {
+      grantCatalogRoleManagerIfNeeded(principalRoleEntity);
```
flyingImer, line 1569: catalog role manager should be granted on catalog_admin grant to a role as well, i.e., inside `assignPrincipalRole`

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
vignesh-manel: made it similar to service_admin

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: good catch. made relevant changes to cleanup on catalog drop and also to revoke catalog_role_manager when principal loses a principal role

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: added log

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: added log

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: assignPrincipalRole() is called when a principal is first assigned to a principal role. At that moment, the principal role might not have any catalog_admin grants yet. The auto-grant happens when catalog_admin is granted to the role

dimas-b:
@dennishuo @collado-mike : WDYT?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1518,8 +1531,17 @@ public PrivilegeResult assignCatalogRoleToPrincipalRole(
...
```
flyingImer, line 1569: IIUC, the current fix covers the assign-then-grant order, but I think the reverse order still looks open: if a principal role already holds `catalog_admin` on a catalog and a principal is added later via `assignPrincipalRole`, the new principal does not seem to inherit `catalog_role_manager`. Would it make sense to handle that mirror case as well, or add a test showing why the current lifecycle is still safe?

vignesh-manel: updated to handle this case as well

vignesh-manel:
Gentle Reminder for PR review

dimas-b:
@vignesh-manel : please add a CHANGELOG entry for this since it affects Polaris administration.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
```diff
@@ -104,6 +104,23 @@ public static PrincipalSecretsResult createPolarisPrincipalForRealm(
...
         PolarisPrivilege.SERVICE_MANAGE_ACCESS);
 
+    // create the catalog_role_manager principal role for catalog admins to list principal roles
+    PrincipalRoleEntity catalogRoleManagerPrincipalRole =
+        new PrincipalRoleEntity.Builder()
+            .setId(generateId(metaStoreManager, ctx))
+            .setName(PolarisEntityConstants.getNameOfCatalogRoleManagerPrincipalRole())
+            .setCreateTimestamp(System.currentTimeMillis())
+            .build();
+    metaStoreManager.createEntityIfNotExists(ctx, null, catalogRoleManagerPrincipalRole);
```
dimas-b: As an alternative to adding a new default role in the server code, WDYT about using the new `setup` CLI (#3929) command instead?

This way, the server does not have to manage a lot of default roles, and admin users can still have convenient methods of managing new grants.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
```diff
@@ -104,6 +104,23 @@ public static PrincipalSecretsResult createPolarisPrincipalForRealm(
...
+
+    // grant PRINCIPAL_ROLE_LIST on the rootContainer to the catalogRoleManagerPrincipalRole
+    metaStoreManager.grantPrivilegeOnSecurableToRole(
+        ctx,
+        catalogRoleManagerPrincipalRole,
+        null,
+        rootContainer,
+        PolarisPrivilege.PRINCIPAL_ROLE_LIST);
```
dimas-b: We grant a "principal" privilege to a "catalog admin" role. This looks a bit out of line with the general distinction of principals (which are global to the realm) and catalog-specific roles (scoped under a catalog). Should the new role be named "principal role manager" perhaps?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -1518,8 +1569,17 @@ public PrivilegeResult assignCatalogRoleToPrincipalRole(
...
```
dimas-b, line 1569: Why not grant `PRINCIPAL_ROLE_LIST` to `catalog_admin` directly?

dimas-b (commented):
Posting some review comments from my side. Sorry about the delay.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: As per my understanding this is not possible as `catalog_admin` is a catalog scoped role that cannot hold root-scoped privilege `PRINCIPAL_ROLE_LIST`. That is the reason we need a principal role (catalog_role_manager) which can have `PRINCIPAL_ROLE_LIST` privilege, and this principal role is automatically granted to principals who have catalog_admin.

dimas-b: TIL - thx for the info!

I'm still a bit uncomfortable about the automatic granting part... Is that critical?

dimas-b: The role named `catalog_admin` may exist in every catalog, but conceptually it is scoped down to each catalog individually. However, Principals are global to the realm. So automatically exposing principal role listing to every catalog admin sounds like an overreach to me... WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
vignesh-manel: Yes, this is possible via CLI for a service_admin to create a new principal role with PRINCIPAL_ROLE_LIST privilege and manually assign it to all principals having the catalog_admin catalog role
 (and revoke it when principals no longer have catalog_admin). However, this manual process can be error-prone and operationally cumbersome.
 
 This feature automates the above mentioned grant and revoke process as described in https://github.com/apache/polaris/issues/363. After https://github.com/apache/polaris/pull/361 enabled catalog_admin to grant catalog roles to principal roles, issue #363 was that catalog_admin still cannot list available principal roles, making it difficult to know which roles to grant. This PR solves that problem through automatic role management.

dimas-b: Well, #363 was reported by @collado-mike , but @collado-mike is not active on this PR, unfortunately :shrug:

dimas-b: from my POV, the new role introduced in this PR makes sense. I'm just not sure about automatically granting it to every user with a `catalog_admin` role. My rationale is here: https://github.com/apache/polaris/pull/3852#discussion_r2983661300

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthBootstrapUtil.java
vignesh-manel: We are creating a  principal role which as `PRINCIPAL_ROLE_LIST` and assigning it to all principals which have `catalog_admin` catalog role, in a way we are managing a catalog role(catalog_admin) by adding a new principal role to provide access to principals - based on this I had named it catalog_role_manager. Based on this functionality WDYT would be the right name for this? I can update it accordingly

dimas-b: My rationale: We should name the role according to the "job" its members perform. `PRINCIPAL_ROLE_LIST` allows listing "principal" roles, in a way managing "principal" roles, therefore I think `principal_role_manager` fits this use case. `principal_role_reader` might be more precise.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
flyrain: Do we provide a way to re-bootstrap? AuthBootstrapUtil (line 56-61) basically prevents re-bootstrap. I think we need to figure out a way for upgrade from existing realm. Throwing an error there isn't enough.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityConstants.java
flyrain: The name `catalog_role_manager` feels confusing to me. It is not really about managing catalog roles. In this PR, it is only granted because of `catalog_admin`, but the role itself is about listing principal roles, and the same capability could be reused in other contexts that have nothing to do with catalog roles.

A name like `principal_role_viewer` or `principal_role_reader` seems more accurate and easier to understand. cc @singhpk234 @collado-mike

dimas-b: +1 to `principal_role_viewer`

vignesh-manel: renamed ro `principal_role_viewer`

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: Updated the changes to handle upgrade for existing realm

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -2331,4 +2400,336 @@ private static PolarisEntitySubType selectEntitySubType(List<PolarisEntitySubTyp
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Revokes the principal_role_viewer principal role from all principals assigned to the specified
+   * principal role if they no longer have any catalog_admin grants.
+   */
+  private void revokeCatalogRoleManagerIfNeeded(PrincipalRoleEntity principalRoleEntity) {
```
collado-mike, line 2478: This is an n^2 loop that is very likely to leave the catalog privileges in a weird state. For every principal granted this principal role, you find all principal roles that have been granted that principal and then check the catalog roles of each. If 1000 principals have 1000 roles, you'll make 1,000,000 calls. If the request times out halfway through (likely), you may only remove the list role from half of the principals, leaving the other half with a role they shouldn't have access to. I think something atomic and O(1) would be a better approach to this problem

collado-mike (changes requested):

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
vignesh-manel: @collado-mike can you please suggest the approach which would do this in O(1) complexity?

PR #3860, opened 2026-02-23
Revert "Update dependency com.fasterxml.jackson:jackson-bom to v2.21.1"

jbonofre:
Reverts apache/polaris#3859


snazy:
No need to revert, the jackson-scala artifacts [are there now](https://repo.maven.apache.org/maven2/com/fasterxml/jackson/module/jackson-module-scala_2.12/2.21.1/). The jackson-scala artifacts are usually published a bit later than the "core" Jackson artifacts.

jbonofre:
Artifact is available on Central now. No need to revert.

PR #3864, opened 2026-02-23
Update LICENSE/NOTICE for 1.4.0 release

snazy:
This change effectively prepares the LICENSE files for the 1.4.0 release.

* Requires the changes from #3881 
* Re-adds the license checks removed a while ago. This verifies the dependencies of the _effective_ dependencies in the Polaris Admin tool & Polaris Server Quarkus applications by ensuring that the dependencies are _mentioned_ in the LICENSE file.
* The Polaris distribution assembly's LICENSE & NOTICE files are now generated from the LICENSE & NOTICE file contents of Polaris Admin tool & Polaris Server. In other words: do not need to be maintained separately.

Fixes #2484

PR #3865, opened 2026-02-23
Exclude Jetty from Polaris Server

snazy:
Jetty is a web server and servlet container with its own lifecycle. This can easily confuse or even conflict with consuming application dependencies. If started, it can also cause issues, potentially security issues.

### DIFF
--- a/extensions/federation/hadoop/build.gradle.kts
+++ b/extensions/federation/hadoop/build.gradle.kts
@@ -44,6 +44,7 @@ dependencies {
     exclude("com.sun.jersey", "jersey-server")
     exclude("com.sun.jersey", "jersey-servlet")
     exclude("io.dropwizard.metrics", "metrics-core")
+    exclude(group = "org.eclipse.jetty")
   }
   implementation(libs.hadoop.client.api)
   implementation(libs.hadoop.client.runtime)


snazy:
This is a prerequisite for #3864

runtime/server/build.gradle.kts
```diff
@@ -43,7 +43,9 @@ dependencies {
   runtimeOnly("org.postgresql:postgresql")
   runtimeOnly(project(":polaris-relational-jdbc"))
   runtimeOnly("io.quarkus:quarkus-jdbc-postgresql")
-  runtimeOnly(project(":polaris-extensions-federation-hadoop"))
+  runtimeOnly(project(":polaris-extensions-federation-hadoop")) {
+    exclude(group = "org.eclipse.jetty") // No Jetty server, please
```
adutra: Shouldn't this exclusion go into the federation-hadoop build file? E.g.

```kotlin
  // Hadoop dependencies (for Hadoop catalog support)
  implementation(libs.hadoop.common) {
    exclude("org.slf4j", "slf4j-reload4j")
    ...
    exclude(group = "org.eclipse.jetty")
  }
```

snazy: works as well, yes

dimas-b (approved):

extensions/federation/hadoop/build.gradle.kts
```diff
@@ -44,6 +44,7 @@ dependencies {
     exclude("com.sun.jersey", "jersey-server")
     exclude("com.sun.jersey", "jersey-servlet")
     exclude("io.dropwizard.metrics", "metrics-core")
+    exclude(group = "org.eclipse.jetty") // No Jetty server, please
```
singhpk234: can you please add more descriptive comment (if required), i understand the intent and totally agree.

flyrain: I’m fine either removing the comment entirely or expanding it to clearly explain the rationale, whichever we think adds more clarity for future readers.

dimas-b: +1 to remove comment from code and add rationale to the PR description (commit message)

flyrain: Btw, would it be possible to add the rationale of removal to the PR description, so that reviewers or any other readers can understand the motivation better,  which is also align with the Polaris PR template format.

PR #3866, opened 2026-02-23
Ban `javax.*` dependencies

snazy:
The `javax.*` group ID artifacts are often GPL+CE licensed, which is often concerning, even if it is not a real issue. Technically, we do not need those dependencies in production code. This means that we can savely remove those dependencies from the Quarkus production runtime.
    
Due to the Spark integration tests, as Spark needs the old Servlet API in the `javax.*` namespace, we cannot globally ban the old servlet API, which causes some special handling.

### DIFF
--- a/build-logic/src/main/kotlin/polaris-java.gradle.kts
+++ b/build-logic/src/main/kotlin/polaris-java.gradle.kts
@@ -17,13 +17,13 @@
  * under the License.
  */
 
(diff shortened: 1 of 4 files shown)


snazy:
This is a prerequisite for #3864

adutra (approved):

adutra (dismissed):

PR #3879, opened 2026-02-24
Helm chart: use real URLs in values.yaml instead of Hugo templates

adutra:
Using Hugo templates in values.yaml is not very user-friendly, since users reading that file cannot guess what the URL is.

Instead, this PR proposes to use regular URLs. In the `main` branch, they point to the `https://polaris.apache.org/in-dev/unreleased/` root. When doing a release, the URLs are changed to `https://polaris.apache.org/releases/x.y.z`.

## Checklist

### DIFF
--- a/helm/polaris/values.schema.json
+++ b/helm/polaris/values.schema.json
@@ -323,7 +323,7 @@
             }
         },
         "features": {
-            "description": "Features to enable or disable globally. If a feature is not present in the map, the default for a complete list of available features.",
+            "description": "Features to enable or disable globally. If a feature is not present in the map, the default built-in value is used. Refer to Polaris [Configuration reference](https://polaris.apache.org/in-dev/unreleased/configuration/configuration-reference) for a complete list of available features.",
             "type": "object",
             "properties": {
                 "realmOverrides": {
--- a/helm/polaris/values.yaml
+++ b/helm/polaris/values.yaml
@@ -849,7 +849,7 @@ realmContext:
     - POLARIS
 
 # -- Features to enable or disable globally. If a feature is not present in the map, the default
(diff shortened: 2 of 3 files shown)


dimas-b (dismissed):

helm/polaris/values.schema.json
```diff
@@ -323,7 +323,7 @@
             }
         },
         "features": {
-            "description": "Features to enable or disable globally. If a feature is not present in the map, the default built-in value is used. Refer to Polaris [Configuration reference]({{% ref \"../configuration\" %}}) for a complete list of available features.",
```
flyingImer, line 326: is this intentionally dropping the config reference? if so, mind elaborating?

MonkeyCanCode: Nice catch on this @flyingImer . So this is actually due to parsing with helm-doc via `helm schema -f values.yaml --use-helm-docs --draft 7`.

The issue here is due to we are using `https://polaris.apache.org/in-dev/unreleased/configuration/configuration-reference/#features--behavior` as url where there is two dashes. That somehow break the parsing. The two dashes are due to safe url from hugo when using `&` (in this case `features & behavior` got converted to `#features--behavior`).

Now to fix this problem, here are couple options:
1. Convert this to a single line to avoid rendering issue
2. Update hugo template to handle `&`
3. Avoid use `&` in the headers which will be using as anchor (there is nothing wrong here just issue from the tooling)
4. Refer to other links which doesn't have `--` (e.g. maybe point to `https://polaris.apache.org/in-dev/unreleased/configuration/configuration-reference/` instead?)
5. Check if we can fix bug from the tooling (in this case helm docs and helm schema)

There may other options available. @adutra thoughts?

adutra: Good catch @flyingImer ! I think the easiest fix is to just remove the anchor `#features--behavior`.

PR #3861, opened 2026-02-23
chore: remove versions from LICENSE and NOTICE files to "relax" dependecy updates, improving also the wording

jbonofre:
This closes #2484

PR #3881, opened 2026-02-25
Fix LICENSE and NOTICE for admin tool and server (prerequisite)

snazy:
This change updates the LICENSE and NOTICE files as a prerequisite for #3864.

* Removal of the version mentions in the LICENSE files
* Improved wording from https://github.com/apache/polaris/pull/3861

### DIFF
--- a/runtime/admin/distribution/NOTICE
+++ b/runtime/admin/distribution/NOTICE
@@ -9,7 +9,7 @@ to the ASF by Snowflake Inc. (https://www.snowflake.com/) copyright 2024.
 
 --------------------------------------------------------------------------------
 
-This product bundles Picocli with the following in its NOTICE:
+This product bundles Picocli with the following in its NOTICE file:
 |
 | This project includes one or more documentation files from OpenJDK, licensed under GPL v2 with Classpath Exception.
 |
@@ -18,7 +18,7 @@ This product bundles Picocli with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles gRPC with the following in its NOTICE:
+This product bundles gRPC with the following in its NOTICE file:
 |
 | Copyright 2014 The gRPC Authors
 | 
@@ -85,7 +85,7 @@ This product bundles gRPC with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles Micrometer with the following in its NOTICE:
+This product bundles Micrometer with the following in its NOTICE file:
 |
 | Micrometer
 | 
@@ -135,7 +135,7 @@ This product bundles Micrometer with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles and includes code from Netty with the following in its NOTICE:
+This product bundles and includes code from Netty with the following in its NOTICE file:
 | 
 |                             The Netty Project
 |                             =================
@@ -403,7 +403,7 @@ This product bundles and includes code from Netty with the following in its NOTI
 
 --------------------------------------------------------------------------------
 
-This product bundles Perfmark with the following in its NOTICE:
+This product bundles Perfmark with the following in its NOTICE file:
 | 
 | Copyright 2019 Google LLC
 | 
@@ -448,7 +448,7 @@ This product bundles Perfmark with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles SmallRye with the following in its NOTICE:
+This product bundles SmallRye with the following in its NOTICE file:
 |
 | Copyright 2009-2017 Mark Struberg
 | Copyright 2018 Red Hat, inc.
@@ -458,7 +458,7 @@ This product bundles SmallRye with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles Swagger with the following in its NOTICE:
+This product bundles Swagger with the following in its NOTICE file:
 |
 | Swagger Core - ${pom.name}
 | Copyright (c) 2015. SmartBear Software Inc.
@@ -467,7 +467,7 @@ This product bundles Swagger with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles Jackson JSON Processor with the following in its NOTICE:
+This product bundles Jackson JSON Processor with the following in its NOTICE file:
 |
 | # Jackson JSON processor
 | 
@@ -504,7 +504,7 @@ This product bundles Jackson JSON Processor with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles CDI with the following in its NOTICE:
+This product bundles CDI with the following in its NOTICE file:
 |
 | # Notices for Jakarta Contexts and Dependency Injection
 | 
@@ -559,7 +559,7 @@ This product bundles CDI with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles jakarta.validation-api with the following in its NOTICE:
+This product bundles jakarta.validation-api with the following in its NOTICE file:
 |
 | # Notices for Eclipse Jakarta Validation
 | 
@@ -606,7 +606,7 @@ This product bundles jakarta.validation-api with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles Conscrypt (openjdk uber) with the following in its NOTICE:
+This product bundles Conscrypt (openjdk uber) with the following in its NOTICE file:
 |
 | Copyright 2016 The Android Open Source Project
 | 
@@ -641,7 +641,7 @@ This product bundles Conscrypt (openjdk uber) with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles Microprofile with the following in its NOTICE:
+This product bundles Microprofile with the following in its NOTICE file:
 |
 | =========================================================================
 | ==  NOTICE file corresponding to section 4(d) of the Apache License,   ==
@@ -673,7 +673,7 @@ This product bundles Microprofile with the following in its NOTICE:
 
 --------------------------------------------------------------------------------
 
-This product bundles AWS Java SDK with the following in its NOTICE:
+This product bundles AWS Java SDK with the following in its NOTICE file:
 |
 | AWS SDK for Java
 | Copyright 2010-2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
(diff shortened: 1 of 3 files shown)


snazy:
@jbonofre the first commit is from your PR.

runtime/server/distribution/LICENSE
```diff
@@ -901,404 +1067,463 @@ License: BSD 3-Clause
...
 
 --------------------------------------------------------------------------------
 
-Group: jakarta.resource Name: jakarta.resource-api Version: 2.1.0
+This product bundles Jakarta JSON API.
+
+* Maven group:artifact IDs: jakarta.json:jakarta.json-api
+
+Project URL: https://github.com/eclipse-ee4j/jsonp
+License: Eclipse Public License 2.0 - https://projects.eclipse.org/license/epl-2.0
```
adnanhemani, line 1339: Copying from #3864:

> [Not sure if relevant, but just during diff checking had this question:] Different between EPL and "Eclipse Public License"?

Took a further look, seems like the new website (https://projects.eclipse.org/license/epl-2.0) just redirects to the EPL 2.0 one (https://www.eclipse.org/legal/epl-2.0).

Not sure if we are bound to put whatever is listed in the package, or if we should help standardize.

runtime/server/distribution/LICENSE
```diff
@@ -616,74 +715,105 @@ License: BSD 3-Clause
...
+* Maven group:artifact IDs: com.google.cloud.opentelemetry:exporter-metrics
+* Maven group:artifact IDs: com.google.cloud.opentelemetry:shared-resourcemapping
+
 Project URL: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java
 License: Apache License 2.0 - https://www.apache.org/licenses/LICENSE-2.0.txt
 
 --------------------------------------------------------------------------------
 
-Group: com.google.code.gson Name: gson Version: 2.13.2
+This product bundles GSon.
```
adnanhemani, line 752: Copied from #3864:

> nit: I don't think that's how "Gson" is capitalized. Not sure that this makes a difference though?

Found this in other license files as well. Leaving only one comment for brevity.

runtime/server/distribution/LICENSE
```diff
@@ -819,55 +958,82 @@ License: MIT License
 
 --------------------------------------------------------------------------------
 
-Group: com.nimbusds Name: nimbus-jose-jwt Version: 10.5
+This product bundles Nimbu Jose JWT.
```
adnanhemani: Copied from #3864:

> Nimbus?

runtime/server/distribution/LICENSE
```diff
@@ -819,55 +958,82 @@ License: MIT License
...
+
+* Maven group:artifact IDs: commons-cli:commons-cli
+
 Project URL: https://commons.apache.org/proper/commons-cli/
 License: Apache License 2.0 - https://www.apache.org/licenses/LICENSE-2.0.txt
 
 --------------------------------------------------------------------------------
 
-Group: commons-codec Name: commons-codec Version: 1.19.0
+This product bundles Apache Commons Codec.
```
adnanhemani, line 979: Copied from #3864:

> Not sure on the rules here (I understand if every Maven group must have its own section) - but if not, could we join all the Apache Commons together?

runtime/server/distribution/LICENSE
```diff
@@ -901,404 +1067,463 @@ License: BSD 3-Clause
...
 Project URL: https://github.com/reactor/reactor-core
 License: Apache License 2.0 - https://www.apache.org/licenses/LICENSE-2.0.txt
 
 --------------------------------------------------------------------------------
 
-Group: io.prometheus Name: simpleclient Version: 0.16.0
+This product bundles Prometheus SimpleClient.
+
+* Maven group:artifact IDs: io.prometheus:simpleclient
 Group: io.prometheus Name: simpleclient_common Version: 0.16.0
```
adnanhemani: Copied from #3864:

> Missing these removals/transformations here.

runtime/distribution/LICENSE
```diff
@@ -235,147 +248,43 @@ License: BSD 2-Clause
...
-Group: io.quarkus Name: quarkus-virtual-threads Version: 3.29.4
-Group: io.quarkus Name: quarkus-websockets-next-spi Version: 3.29.4
-Group: io.quarkus.resteasy.reactive Name: resteasy-reactive Version: 3.29.4
-Group: io.quarkus.resteasy.reactive Name: resteasy-reactive-common Version: 3.29.4
-Group: io.quarkus.resteasy.reactive Name: resteasy-reactive-common-types Version: 3.29.4
-Group: io.quarkus.resteasy.reactive Name: resteasy-reactive-jackson Version: 3.29.4
-Group: io.quarkus.resteasy.reactive Name: resteasy-reactive-vertx Version: 3.29.4
-Group: io.quarkus.security Name: quarkus-security Version: 2.2.1
-Group: io.quarkus.vertx.utils Name: quarkus-vertx-utils Version: 3.29.4
+This product bundles Quarkus.
```
adnanhemani, line 251: I'm noticing that we don't have the Maven group packages listing here but we do in the other files. I checked Iceberg's license file and the ASF licensing guide - I don't see any mention of if this is required in a certain way or not. Do we need to do the listing of the Maven group-packages used here?

runtime/admin/distribution/LICENSE
```diff
@@ -202,11 +202,57 @@
 
 --------------------------------------------------------------------------------
 
-This binary artifact bundles the following projects:
+This product bundles and includes code from Apache Iceberg.
```
flyrain, line 205: The term "binary artifact" is more accurate IMHO. Do we want to change it?

jbonofre: This is not what we prefer, it's what we have to use, and consistency across projects.

As we can see here https://infra.apache.org/licensing-howto.html, the wording used is "This product bundles" or "The product includes", as we already use for:

```
This product includes software developed at
The Apache Software Foundation (http://www.apache.org/).
```

So for consistency and align with ASF good practice, I proposed to use `This product bundles` (for dependencies) and `This product includes code` (for code copy).

runtime/admin/distribution/LICENSE
```diff
@@ -202,11 +202,57 @@
...
+
+* Maven group:artifact IDs: org.apache.iceberg:iceberg-api
```
jbonofre, line 207: Personally, I think it's useless to keep the Maven coordinates, but I'm fine to keep it (as soon as there's no version 😄 ).

runtime/admin/distribution/NOTICE
```diff
@@ -9,39 +9,17 @@ to the ASF by Snowflake Inc. (https://www.snowflake.com/) copyright 2024.
...
 --------------------------------------------------------------------------------
 
-This binary artifact bundles the following projects with NOTICE:
-
---------------------------------------------------------------------------------
-
-Group: info.picocli Name: picocli Version: 4.7.7
-
-Relevant information from NOTICE:
+This product bundles Picocli with the following in its NOTICE:
```
jbonofre: I think I did a minor mistake in the wording in my original PR.

We should use `with the following in its NOTICE file:` instead of `with the following in its NOTICE:`.

runtime/distribution/LICENSE
jbonofre: Two comments:
1. Yes, I agree it should be one or another: we have to list or not list the Maven coordinates on the `LICENSE` files.
2. No, there's no requirement to list the Maven coordinates. We have to document in the `LICENSE` the dependencies bundled in our product. It's up to us how we document it. So, just `This product bundles Quarkus` is good enough.

runtime/server/distribution/LICENSE
jbonofre: According to https://github.com/google/gson, `Gson` is the right form.

That said, it's not a big deal 😄

runtime/server/distribution/LICENSE
jbonofre: Good catch, yes, it should be Nimbus.

runtime/server/distribution/LICENSE
jbonofre: As Apache Commons is separate projects (different `LICENSE` files), I prefer to document each Apache Commons atomic.

runtime/server/distribution/LICENSE
jbonofre: Good catch, +1

runtime/server/distribution/LICENSE
jbonofre: If there is a difference between EPL and EDL, EPL IS actually the Eclipse Public License.
What matters is documenting the dependency using the EPL license.
The link (as working) is good enough.

runtime/server/distribution/LICENSE
snazy: ~Not really, those are superfluous and not present in the dependency tree and should be removed.~

runtime/admin/distribution/LICENSE
snazy: +1
We do _both_: bundle _and_ include.

runtime/admin/distribution/LICENSE
snazy: Generally yes, however #3864 checks those coordinates against the dependency tree and "barks" if a dependency is missing to superfluous.

runtime/admin/distribution/NOTICE
snazy: Did a global search + replace.

runtime/server/distribution/LICENSE
snazy: Prometheus dependencies were actually mentioned twice. Removed the duplicate at the end.

runtime/admin/distribution/LICENSE
adnanhemani: That's pretty useful then! So then do we need to add the Maven coordinates to `runtime/distribution/LICENSE` as well?

snazy: No, see description of #3864

jbonofre:
On it.

runtime/distribution/LICENSE
```diff
@@ -202,11 +202,24 @@
 
 --------------------------------------------------------------------------------
 
-This binary artifact bundles the following projects:
+This propduct bundles and includes code from Apache Iceberg.
```
jbonofre, line 205: Typo here on product

snazy: Not an issue - that part's going away with #3864

PR #3874, opened 2026-02-24
Site: Add explicit release relevant ASF docs and policies

snazy:

PR #3883, opened 2026-02-25
Site: release manager role guide

snazy:
This adds a new page about the role of a release manager providing a bunch of detail and background information to explain why we have to follow such a "formal" and "complicated" release process.

This PR depends on #3874

### DIFF
--- a/site/content/community/release-guides/release-manager-guide.md
+++ b/site/content/community/release-guides/release-manager-guide.md
@@ -46,7 +46,7 @@ This page adds more details to the _Release manager_ paragraph on the [_Release 
 
 The RM is a committer who takes **responsibility** for the **mechanics** of a release.
 You aren't "in charge" of what goes into the release (that's a community decision), but you are the **facilitator**
-ensuring that the release process follows the [ASF Release Policy](https://www.apache.org/legal/release-policy.html).
+ensuring that the release process follows the [_ASF Release Policy_](https://www.apache.org/legal/release-policy.html).
 
 As the whole release process is a very formal act, it is strictly required to follow the _ASF Release Policy_.
 Not following the policy can lead to “failed” releases or even more serious consequences.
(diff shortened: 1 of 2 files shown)


site/content/community/release-guides/release-manager-guide.md
```diff
@@ -0,0 +1,319 @@
...
+The ASF provides a lot of additional relevant content on this topic.
+{{< /alert >}}
+
+{{< alert note >}}
+This page adds more details to the _Release manager_ paragraph on the [_Release Guides_ page](../#release-manager).
+{{< /alert >}}
+
+## What is a Release Manager?
+
+The RM is a committer who takes **responsibility** for the **mechanics** of a release.
```
flyrain: AFAIK, the RM doesn't have to be a committer. There is no ASF policy enforcing that as well. In fact, I have seen non-committer as RM often. cc @jbonofre

snazy: While it is true that the RM _role_ does not need to be fulfilled by a _committer_, many of the technical release relevant work require _committer_ privileges (SVN access, workflow triggering, etc).

Strictly speaking, and mentioned in the [Release Creation Process](https://infra.apache.org/release-publishing.html#releasemanager), "_... the PMC in general, and the PMC chair in particular (as an officer of the Foundation), remains responsible for compliance with ASF requirements._" and "_Allowing Committers to publish release artifacts requires project consensus._"

Being a _committer_ is a prerequisite to "push the button."

For clarification I added "typically" to that sentence.

jbonofre: From a "policy standpoint", the RM doesn't have to be a committer. In that case, the RM has to work with a committer/PMC member for the "technical requirements" (dist access, etc).

From a "legal standpoint", the RM doesn't have any responsibility. It's up to us, the PMC members (and especially the chair as VP of the Foundation), to guarantee that all ASF requirements are met: legal etc.

I would rephrase like "The RM is a contributor .... If the RM is not a committer, he has to work with project committer for technical/permission requirements".

snazy: Already rephrased to "is typically a committer"

PR #3908, opened 2026-02-28
Persistence: use correct grant loading method

singhpk234:
### About the change

toResolvedPolarisEntity() was calling loadAllGrantRecordsOnSecurableInCurrentTxn for both the securable and grantee grant records. The grantee line should call loadAllGrantRecordsOnGranteeInCurrentTxn instead. This caused loadResolvedEntities() to return duplicate securable records in place of grantee records for entities that are both grantees and securables (e.g., catalog roles).

## Checklist

### DIFF
--- a/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java
+++ b/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java
@@ -2850,12 +2850,24 @@ public class PolarisTestMetaStoreManager {
     // grantee records: R1 was granted TABLE_READ_DATA on N1, so granteeId must be R1
     Assertions.assertThat(resolved.getGrantRecordsAsGrantee())
         .isNotEmpty()
-        .allSatisfy(g -> Assertions.assertThat(g.getGranteeId()).isEqualTo(R1.getId()));
+        .allSatisfy(
+            g -> {
+              Assertions.assertThat(g.getGranteeId()).isEqualTo(R1.getId());
+              Assertions.assertThat(g.getSecurableId()).isEqualTo(N1.getId());
+              Assertions.assertThat(g.getPrivilegeCode())
+                  .isEqualTo(PolarisPrivilege.TABLE_READ_DATA.getCode());
+            });
 
     // securable records: PR1 was granted CATALOG_ROLE_USAGE on R1, so securableId must be R1
     Assertions.assertThat(resolved.getGrantRecordsAsSecurable())
         .isNotEmpty()
-        .allSatisfy(g -> Assertions.assertThat(g.getSecurableId()).isEqualTo(R1.getId()));
+        .allSatisfy(
+            g -> {
+              Assertions.assertThat(g.getSecurableId()).isEqualTo(R1.getId());
+              Assertions.assertThat(g.getGranteeId()).isEqualTo(PR1.getId());
+              Assertions.assertThat(g.getPrivilegeCode())
+                  .isEqualTo(PolarisPrivilege.CATALOG_ROLE_USAGE.getCode());
+            });
 
     Assertions.assertThat(resolved.getGrantRecordsAsGrantee())
         .doesNotContainAnyElementsOf(resolved.getGrantRecordsAsSecurable());


polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java
```diff
@@ -2801,6 +2801,66 @@ public void testLoadResolvedEntitiesById() {
...
+    PolarisBaseEntity N1 = this.createEntity(List.of(catalog), PolarisEntityType.NAMESPACE, "N1");
+
+    // create a catalog role — both a grantee and a securable
+    PolarisBaseEntity R1 =
+        this.createEntity(List.of(catalog), PolarisEntityType.CATALOG_ROLE, "R1");
+
+    // grant a privilege TO R1 on N1 (R1 is the grantee)
+    grantPrivilege(R1, List.of(catalog, N1), N1, PolarisPrivilege.TABLE_READ_DATA);
+
+    // create a principal role and grant R1 usage to it (R1 is the securable)
```
flyrain, line 2833: I have a hard time to understand R1 is a securable, however, looks like Polaris is using the special privilege(`catalog_role_usage`) to avoid another relation(the relation between catalog role and principal). It should just work, but still I not 100% convinced that R1 is a securable. I'd consider it is a special securable at most.

singhpk234:
> Curious why we didn't hit the correctness issue before.

I am not sure, found it while reading the code :), one guess would be not all metastore imp support this so the in-memory cache plumbing to other things like resolver is still pending ?

persistence/nosql/persistence/metastore/src/test/java/org/apache/polaris/persistence/nosql/metastore/TestNoSqlMetaStoreManager.java
```diff
@@ -116,6 +116,10 @@ protected void testLoadResolvedEntitiesById() {
         .isInstanceOf(UnsupportedOperationException.class);
   }
 
+  @Override
+  @Disabled("loadResolvedEntities is unsupported in the NoSQL metastore implementation")
```
dimas-b, line 120: @snazy : WDYT?

dimas-b: @singhpk234 : Does this mean the new code in `TransactionalMetaStoreManagerImpl` is not usable with NoSQL persistence?

dimas-b: @singhpk234 : Please ignore my previous comment - I was confused.

dimas-b: `loadResolvedEntities` appears to be called only from `InMemoryEntityCache`, which is not used when by NoSQL persistence is active.

Cf. https://github.com/apache/polaris/blob/d95bd6b1e3d77b5db7769e866c0474f08accad2c/persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStoreManagerFactory.java#L128-L131

polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java
```diff
@@ -2801,6 +2801,66 @@ public void testLoadResolvedEntitiesById() {
...
+        .isNotNull()
+        .returns(BaseResult.ReturnStatus.SUCCESS, ResolvedEntitiesResult::getReturnStatus);
+
+    ResolvedPolarisEntity resolved = result.getResolvedEntities().get(0);
+    Assertions.assertThat(resolved).isNotNull();
+
+    // grantee records: R1 was granted TABLE_READ_DATA on N1, so granteeId must be R1
+    Assertions.assertThat(resolved.getGrantRecordsAsGrantee())
+        .isNotEmpty()
+        .allSatisfy(g -> Assertions.assertThat(g.getGranteeId()).isEqualTo(R1.getId()));
```
dimas-b: Could you add asserts for the ID on  "other end" of the grant and privilege code (to be certain)?

Also on line 2858?

persistence/nosql/persistence/metastore/src/test/java/org/apache/polaris/persistence/nosql/metastore/TestNoSqlMetaStoreManager.java
dimas-b: All clear on this comment thread from my side.

singhpk234:
Thanks for the review @flyrain, @dimas-b !

PR #3803, opened 2026-02-18
Add Iceberg REST Idempotency-Key support

huaxingao:
This PR introduces optional Idempotency-Key handling for Iceberg REST mutation endpoints to support safe client retries.

Adds an HTTP request/response filter that:

- validates UUIDv7 keys,
- reserves keys and enforces operation/resource binding (422 on conflict),
- waits for in-progress duplicates and replays finalized responses,
- never finalizes/replays 5xx.

Persists a bounded response body plus an allowlisted set of response headers and replays with correct Content-Type.

Adds optional heartbeats and background purge of expired keys.

Advertises idempotency-key-lifetime via GET /v1/config for Polaris-managed (INTERNAL) catalogs only.

Includes small config knobs for wait/lease TTL and TTL grace.

**Note:** reconciliation/takeover for stale in-progress keys is intentionally deferred to a follow-up PR.

## Checklist
- [ ] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [ ] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [ ] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3915, opened 2026-03-01
Idempotency: JDBC store implementation and SPI

huaxingao:
This PR introduces the persistence SPI for idempotency records and adds a relational-JDBC implementation.
This is split from https://github.com/apache/polaris/pull/3803

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStoreFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Factory for creating an {@link IdempotencyStore} backend.
+ *
+ * <p>Implementations are selected by the runtime wiring layer (e.g. Quarkus CDI) using an
+ * identifier, similar to other Polaris factories.
+ */
+public interface IdempotencyStoreFactory {
+  /** Create an {@link IdempotencyStore} instance. */
+  IdempotencyStore create();
```
dimas-b: This method does not appear to be called 🤔 

What is the intended usage pattern for this factory?

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
```diff
@@ -112,13 +112,12 @@ public String responseSummary() {
...
-   * Serialized representation of a small, whitelisted set of HTTP response headers.
+   * Allowlisted snapshot of HTTP response headers to replay for duplicates.
    *
-   * <p>Stored as a JSON string so that the HTTP layer can replay key headers (such as {@code
-   * Content-Type}) when serving a duplicate idempotent request.
+   * <p>How this map is persisted is store-specific.
    */
   @Override
-  public String responseHeaders() {
+  public Map<String, String> responseHeaders() {
```
dimas-b: HTTP headers can have multiple values. Restricting this map to one value per key may limit the ability of the Polaris Server to represent certain responses.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStoreFactory.java
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.enterprise.inject.literal.NamedLiteral;
+import jakarta.inject.Inject;
+import java.sql.SQLException;
+import javax.sql.DataSource;
+import org.apache.polaris.core.persistence.IdempotencyStore;
+import org.apache.polaris.core.persistence.IdempotencyStoreFactory;
+import org.apache.polaris.persistence.relational.jdbc.RelationalJdbcConfiguration;
+
+@ApplicationScoped
+@Identifier("relational-jdbc")
```
dimas-b: nit: it might be best to defer adding CDI-related classes to the PR that actually uses them.

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStore.java
```diff
@@ -131,29 +132,53 @@ ReserveResult reserve(
...
+   *
+   * <p>This is used for cases where an idempotency key was reserved but the request should not be
+   * tracked for replay (for example, 401/403/408/429 responses per the proposal). Implementations
+   * should only cancel the record if it is still in-progress (i.e., not finalized) and owned by the
+   * given executor.
+   *
+   * @param realmId logical tenant or realm identifier
+   * @param idempotencyKey application-provided idempotency key
+   * @param executorId identifier of the executor that owns the reservation
+   * @return {@code true} if a record was cancelled, {@code false} otherwise
```
dimas-b: It would be nice to specify behaviour in case exceptions are thrown. 

Does the caller care about the success of this operation? In other words, is returning `true` critical?

Note: some exceptions may occur after the JDBC commit completes.

dimas-b (commented):
Thanks for your continuous work on this feature, @huaxingao ! Sorry about the delayed review.

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
```diff
@@ -53,7 +53,7 @@ public record IdempotencyRecord(
     Integer httpStatus,
     String errorSubtype,
     String responseSummary,
-    String responseHeaders,
+    Map<String, String> responseHeaders,
```
dimas-b: Records are supposed to be immutable. Should we ensure this map is immutable?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStoreFactory.java
huaxingao: Sounds good. Deleted this change and defer it to next PR.

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
huaxingao: Good point. Added a compact constructor that wraps it with Map.copyOf() to ensure immutability.

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
huaxingao: For the headers we care about (default: just Content-Type), there's always just one value, so getFirst is fine. If multi-valued header support is needed later, we can change to Map<String, List<String>> in a follow-up.

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStore.java
huaxingao: Cancel is best-effort. The caller doesn't need the return value: the HTTP response is already decided by the time cancel runs. If it fails or throws, the record just expires on its own. Java docs was updated

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStoreFactory.java
huaxingao: The call site is in the next PR, where ServiceProducers calls factory.create() to produce the IdempotencyStore bean. Same pattern as MetaStoreManagerFactory.

huaxingao:
@dimas-b Thanks for your review! I have addressed you comments. Could you please take a look again when you have time?

huaxingao:
also cc @singhpk234 @flyrain @flyingImer

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStoreFactory.java
dimas-b: Does ServiceProducers absolutely have to call `.create()`? This method does not have have any parameters, so the same creation logic can be performed by CDI automatically.

Since "making" objects is essentially a CDI-related activity, I'd prefer deferring the discussion about factory til the PR that make use of them.

In any case, it does not feel like the factory should be in `polaris-core`, because nothing in core needs to call it, as far as I can tell.

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
dimas-b: How do we know which headers are relevant and which are not?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStoreFactory.java
dimas-b: thx!

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStore.java
dimas-b: If the return value does not affect the outcome on the caller's side, why declare a return value for this method at all?

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
huaxingao: When replaying a response, the filter reconstructs it from the stored record: HTTP status, response body (summary), and headers. The client needs Content-Type to know how to parse the body, that's why we have Content-Type as the default. We might also need Location (when a create operation returns the URL of the new resource) and ETag (when the client uses conditional updates like If-Match). Operators can add these via config.

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStoreFactory.java
huaxingao: Makes sense. I deleted `IdempotencyStoreFactory` from this PR and revisit the wiring approach in the next PR.

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStore.java
huaxingao: Changed to `void`

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
dimas-b: Thanks for the context, @huaxingao !

IMHO, this still sounds like we cannot know beforehand what the client or the API layer in the server may need to support for the response headers. Since HTTP allows multi-values headers, I tend to this the Persistence SPI should support that from day 1. Adding multi-valued entries later will be much harder, I think.

polaris-core/src/main/java/org/apache/polaris/core/persistence/IdempotencyStore.java
dimas-b: thx!

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
huaxingao: Makes sense. I updated the PR to support multi-values headers.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
```diff
@@ -147,7 +152,17 @@ static IdempotencyRecord fromRow(String realmId, ResultSet rs) throws SQLExcepti
     Integer httpStatus = (Integer) rs.getObject(HTTP_STATUS);
     String errorSubtype = rs.getString(ERROR_SUBTYPE);
     String responseSummary = rs.getString(RESPONSE_SUMMARY);
-    String responseHeaders = rs.getString(RESPONSE_HEADERS);
+    Map<String, List<String>> responseHeaders = null;
```
dimas-b, line 155: Is Polaris is expected to have any old data for this value?

dimas-b (dismissed):

polaris-core/src/main/java/org/apache/polaris/core/entity/IdempotencyRecord.java
dimas-b: thx!

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
```diff
@@ -190,27 +195,67 @@ public HeartbeatResult updateHeartbeat(
...
+                  idempotencyKey,
+                  ModelIdempotencyRecord.EXECUTOR_ID,
+                  executorId),
+              Map.of(),
+              Map.of(),
+              Set.of(ModelIdempotencyRecord.HTTP_STATUS),
+              Set.of());
+      datasourceOperations.executeUpdate(delete);
+    } catch (SQLException e) {
+      throw new IdempotencyPersistenceException(
```
flyingImer, line 219: I think there are two high-priority issues worth fixing before this moves forward. First, the Postgres IT still uses `Map.of("Content-Type", "application/json")` even though `finalizeRecord()` now expects `Map<String, List<String>>`, which looks like a straight compile/type mismatch. Second, the SPI docs describe `cancelInProgressReservation()` as best-effort and explicitly say failures should not affect the response path, but the JDBC impl now throws on cancel failure. Would it make sense to either make the JDBC path log-and-swallow here, or tighten the SPI contract + callers so the boundary is explicit?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/models/ModelIdempotencyRecord.java
huaxingao: No, I don't think so.

dimas-b: :+1:

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java
dimas-b: Why does the IT test pass with a type mismatch? :thinking:

snazy (commented):
I think this PR should wait until the comments in #3803 are all addressed, because those will affect this spilled-out change.

dimas-b:
So, what's the state of this PR? I assume it is not active anymore (per comments above) :thinking: @huaxingao , if you agree, would you mind marking it as "draft" for the sake of clarity?

PR #3916, opened 2026-03-01
core: rename Resolvable.CATALOG_ROLES to Resolvable.CALLER_CATALOG_ROLES

sungwy:
Resolvable enum was recently introduced to enable the caller to select specific entities for resolution rather than resolving everything. See PR: https://github.com/apache/polaris/pull/3760

This PR renames CATALOG_ROLES to CALLER_CATALOG_ROLES instead to make it obvious whether we are resolving the principal's activated catalog roles vs the catalog roles that are requested as the target of the action (like creating a CATALOG_ROLE).

Also added a private helper to detect if CATALOG_ROLE is requested as a part of the top level entities and would require referenced catalog resolution.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
@@ -36,14 +36,12 @@ public enum Resolvable {
   /** Resolve explicitly registered paths (via addPath/addPassthroughPath). */
   REQUESTED_PATHS,
   /**
-   * Resolve any additional top-level entities explicitly registered via addTopLevelName, such as
-   * catalog/principal/principal-role names used for authorization beyond the caller and reference
-   * catalog.
+   * Resolve any additional top-level entities explicitly registered via {@link
+   * PolarisResolutionManifest#addTopLevelName}, such as catalog/principal/principal-role names used
+   * for authorization beyond the caller and reference catalog.
    *
-   * <p>Note: this currently also covers requested {@code CATALOG_ROLE} names because they share the
-   * same {@link PolarisResolutionManifest} registration surface ({@code addTopLevelName}). Although
-   * grouped here for compatibility, catalog-role name resolution still requires the reference
-   * catalog context.
+   * <p>All entity types resolved here share the same {@link
+   * PolarisResolutionManifest#addTopLevelName} registration surface.
    */
   REQUESTED_TOP_LEVEL_ENTITIES
 }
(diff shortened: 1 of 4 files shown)


polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
```diff
@@ -39,6 +39,11 @@ public enum Resolvable {
    * Resolve any additional top-level entities explicitly registered via addTopLevelName, such as
    * catalog/principal/principal-role names used for authorization beyond the caller and reference
    * catalog.
+   *
+   * <p>Note: this currently also covers requested {@code CATALOG_ROLE} names because they share the
+   * same {@link PolarisResolutionManifest} registration surface ({@code addTopLevelName}). Although
```
sungwy: An alternative is to introduce a larger scope change to add `REQUESTED_CATALOG_ROLES` as a separate enum, and also to add a different registration surface to the `PolarisResolutionManifest` for requested catalog roles

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
```diff
@@ -29,16 +29,19 @@ public enum Resolvable {
   CALLER_PRINCIPAL,
   /** Resolve the caller's activated principal-role entities. */
   CALLER_PRINCIPAL_ROLES,
-  /** Resolve catalog-role entities (e.g., roles attached in the reference catalog). */
-  CATALOG_ROLES,
+  /** Resolve caller-activated catalog-role entities in the reference catalog. */
```
dimas-b, line 32: nit "caller-activated" is understandable, but still a bit cryptic IMHO 😅 How about "Catalog Role entities referenced by the active PolarisPrincipal"?

The "activated" part is generally handled during request authentication, so by the time the request reaches AuthZ, the Principal already has a fixed set of role names in it, IIRC. WDYT?

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
dimas-b: It looks like the latest javadoc does not mention this at all... Has the problem been resolved? :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/Resolvable.java
sungwy: That's exactly what I meant by that - `caller-activated` as in, the catalog-roles that were activated by the caller...

sungwy: since this is a nit, and it would be nice to get this in before the release, I will proceed to merge the change in for now

adutra: This distinction between "activated" vs "granted" comes from the very initial stages of Polaris, afaik.

PR #3824, opened 2026-02-18
Introduce a factory for authz dynamic tests

adutra:
As promised, after #3780, this change further refactors the "authz" tests extending `PolarisAuthzTestBase` to introduce a fluent test builder.

**No function change.**

As an example, these two method calls:

```java
    return doTestSufficientPrivilegeSets(
        "attachPolicyToNamespace",
        List.of(
            Set.of(PolarisPrivilege.POLICY_ATTACH, PolarisPrivilege.NAMESPACE_ATTACH_POLICY),
            Set.of(PolarisPrivilege.CATALOG_MANAGE_METADATA),
            Set.of(PolarisPrivilege.CATALOG_MANAGE_CONTENT)),
        () -> newWrapper(Set.of(PRINCIPAL_ROLE1)).attachPolicy(POLICY_NS1_1, attachPolicyRequest),
        () -> newWrapper(Set.of(PRINCIPAL_ROLE2)).detachPolicy(POLICY_NS1_1, detachPolicyRequest));

    return doTestInsufficientPrivilegeSets(
        "attachPolicyToNamespace",
        List.of(
            Set.of(PolarisPrivilege.POLICY_ATTACH, PolarisPrivilege.CATALOG_ATTACH_POLICY),
            Set.of(PolarisPrivilege.POLICY_ATTACH, PolarisPrivilege.TABLE_ATTACH_POLICY),
            Set.of(PolarisPrivilege.POLICY_ATTACH),
            Set.of(PolarisPrivilege.NAMESPACE_ATTACH_POLICY)),
        () -> newWrapper(Set.of(PRINCIPAL_ROLE1)).attachPolicy(POLICY_NS1_1, attachPolicyRequest));
```

Can now be replaced with:

```java
    return authzTestsBuilder("attachPolicyToNamespace")
        .action(
            () ->
                newHandler(Set.of(PRINCIPAL_ROLE1)).attachPolicy(POLICY_NS1_1, attachPolicyRequest))
        .cleanupAction(
            () ->
                newHandler(Set.of(PRINCIPAL_ROLE2)).detachPolicy(POLICY_NS1_1, detachPolicyRequest))
        .shouldPassWith(PolarisPrivilege.POLICY_ATTACH, PolarisPrivilege.NAMESPACE_ATTACH_POLICY)
        .shouldPassWith(PolarisPrivilege.CATALOG_MANAGE_METADATA)
        .shouldPassWith(PolarisPrivilege.CATALOG_MANAGE_CONTENT)
[...]

PR #3927, opened 2026-03-03
feat: add CATALOG_READ_DATA privilege for data analyst read-only catalog access

Praneeth-Sagar-13:
Introduces a new catalog-level privilege CATALOG_READ_DATA (code 103) that grants read-only access to all data and navigational metadata within a specific catalog. This is intended for data analyst principals who need broad read access across an entire catalog without any write or administrative capabilities.

CATALOG_READ_DATA subsumes the following privileges when granted on a catalog:
- NAMESPACE_LIST and NAMESPACE_READ_PROPERTIES (catalog navigation)
- TABLE_LIST, TABLE_READ_PROPERTIES, and TABLE_READ_DATA (table access)
- VIEW_LIST and VIEW_READ_PROPERTIES (view access)

It is itself subsumed by CATALOG_MANAGE_CONTENT (the existing full-access catalog privilege), preserving the existing privilege hierarchy.

Changes:
- PolarisPrivilege.java: add CATALOG_READ_DATA enum constant (code 103)
- PolarisAuthorizerImpl.java: register subsumption rules in SUPER_PRIVILEGES
- polaris-management-service.yml: add CATALOG_READ_DATA to CatalogPrivilege enum
- PolarisPrivilegeTest.java: update code-to-privilege mapping test

## Checklist

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
@@ -623,6 +623,7 @@ public abstract class AbstractIcebergCatalogHandlerAuthzTest extends PolarisAuth
         .shouldPassWith(PolarisPrivilege.TABLE_READ_DATA)
         .shouldPassWith(PolarisPrivilege.TABLE_WRITE_DATA)
         .shouldPassWith(PolarisPrivilege.CATALOG_MANAGE_CONTENT)
+        .shouldPassWith(PolarisPrivilege.NAMESPACE_READ_DATA)
         .shouldFailWith(PolarisPrivilege.TABLE_READ_PROPERTIES)
         .shouldFailWith(PolarisPrivilege.TABLE_WRITE_PROPERTIES)
         .createTests();
@@ -1803,6 +1804,7 @@ public abstract class AbstractIcebergCatalogHandlerAuthzTest extends PolarisAuth
         .shouldPassWith(PolarisPrivilege.TABLE_FULL_METADATA)
         .shouldPassWith(PolarisPrivilege.CATALOG_MANAGE_CONTENT)
         .shouldPassWith(PolarisPrivilege.CATALOG_MANAGE_METADATA)
+        .shouldPassWith(PolarisPrivilege.NAMESPACE_READ_DATA)
         .shouldFailWith(PolarisPrivilege.TABLE_MANAGE_STRUCTURE)
         .createTests();
   }


collado-mike:
I feel ok with this change. +1 on the ML discussion. Also @dennishuo may have an opinion

Praneeth-Sagar-13:
Started a [DISCUSS] thread :[https://lists.apache.org/thread/p6fjxh7y65s2mdywc330gor6yrj2vngz](url)

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
```diff
@@ -256,6 +256,16 @@ public enum PolarisPrivilege {
...
+  /**
+   * Grants read-only access to all data and navigational metadata within a specific catalog.
+   * Intended for data analyst principals who need to read tables and views across an entire catalog
+   * without any write or administrative capabilities.
+   *
+   * <p>This privilege subsumes: {@link #NAMESPACE_LIST}, {@link #NAMESPACE_READ_PROPERTIES},
+   * {@link #TABLE_LIST}, {@link #TABLE_READ_PROPERTIES}, {@link #TABLE_READ_DATA}, {@link
+   * #VIEW_LIST}, and {@link #VIEW_READ_PROPERTIES}.
+   */
+  CATALOG_READ_DATA(103, PolarisEntityType.CATALOG),
```
dennishuo: I agree with the scoping of this privilege to be those 7 privileges. Given that though, we might want to consider whether to call this privilege `NAMESPACE_READ_DATA` instead, since it doesn't contain any catalog-entity privileges like `CATALOG_READ_PROPERTIES`.

This may also highlight something we can discuss more in the dev list, which is the intentionally separate meanings of the privilege-inheritance from the *securable hierarchy* vs the logical *privilege hierarchy* defined here. As in, despite a privilege being applicable to a leaf entity or middle-layer entity (e.g. TABLE_READ_PROPERTIES, NAMESPACE_READ_PROPERTIES), and child-entity privilege types are also allowed to be granted *on* the securable parent (e.g. `GRANT TABLE_READ_PROPERTIES ON CATALOG c1 TO CATALOG_ROLE cr1`), and whether it's a source of confusion.

dimas-b (dismissed):
LGTM 👍

dimas-b:
@Praneeth-Sagar-13 : Please run `./gradlew spotlessApply` - see CI reports for specific formatting errors.

Praneeth-Sagar-13:
The formatting has been fixed with ./gradlew spotlessApply. Could @dimas-b and @adutra please re-review? The only change since your last approval was the spotless formatting fix.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisPrivilege.java
```diff
@@ -256,6 +256,16 @@ public enum PolarisPrivilege {
...
+  /**
+   * Grants read-only access to all data and navigational metadata within a specific catalog.
+   * Intended for data analyst principals who need to read tables and views across an entire catalog
+   * without any write or administrative capabilities.
+   *
+   * <p>This privilege subsumes: {@link #NAMESPACE_LIST}, {@link #NAMESPACE_READ_PROPERTIES}, {@link
+   * #TABLE_LIST}, {@link #TABLE_READ_PROPERTIES}, {@link #TABLE_READ_DATA}, {@link #VIEW_LIST}, and
+   * {@link #VIEW_READ_PROPERTIES}.
+   */
+  CATALOG_READ_DATA(103, PolarisEntityType.CATALOG),
```
flyrain: The privilege is good to have, otherwise, user have to grant/invoke 7+ privileges for the same purpose.

The name CATALOG_READ_DATA implies data-only access (reading table contents), but it actually subsumes metadata-plane privileges too like TABLE_READ_PROPERTIES, NAMESPACE_READ_PROPERTIES, VIEW_READ_PROPERTIES, and the *_LIST privileges. I am wondering if we can give a more descriptive name, like `CATALOG_READ` or `CATALOG_READ_ALL`. 

A related question: why doesn't it subsume `CATALOG_READ_PROPERTIES`?

spec/polaris-management-service.yml
```diff
@@ -1582,6 +1582,7 @@ components:
         - CATALOG_MANAGE_ACCESS
         - CATALOG_MANAGE_CONTENT
         - CATALOG_MANAGE_METADATA
+        - CATALOG_READ_DATA
```
dennishuo: If we do make this `NAMESPACE_READ_DATA`, we can still allow it to be set at either this `CatalogPrivilege` level *or* at the `NamespacePrivilege` level, which I think could actually be in keeping with the intended use case, where you may want to grant "read all" semantics at some entire namespace subtree even if not granting at the entire catalog level.

dennishuo: And in case it's not clear how, allowing setting at either level should be as simple as adding the same thing on line 1533 under `NamespacePrivilege`

adutra:
> @adutra We used to try (imperfectly) to adhere to having a comprehensive covering of privileges always enumerated in those Authz tests before https://github.com/apache/polaris/pull/3824/changes#top but now I guess in the interest of reducing verbosity of the tests we no longer enumerate all the subsuming privileges so I it's less obvious to developers what should be required to be added to the tests. but this might be something to discuss more deeply on the mailing list - for something like Authz tests we probably _want_ to force ourselves to be explicit about testing which things succeed and which things fail whenever we create a new super-privilege. Auto-generating the "insufficient" sets is probably still a good thing because it could actually help reduce human error, _if_ we don't auto-generate subsuming-privilege "shouldPassWith" tests, since that way the tests can automatically identify all branches that need to be considered when defining the privilege.

Yes, I added two "syntactic sugars":

- All single-privilege negative tests are added automatically (by subtracting all the single-privilege positive tests)
- All subsuming privilege positive tests are added automatically for each single-privilege positive test

So in most cases, it's enough to explicitly declare the "least privilege" positive tests, and the test factory will derive the other tests from that.

I think that generating negative tests automatically is a good thing and should not be removed. It actually helped me identify a few missing positive tests.

Automatically generating the subsuming positive tests is imho also a good thing. Since the subsuming privileges are statically defined in `PolarisAuthorizerImpl`, I don't see much value in listing the subsuming test cases extensively for each test. But I'm open to revisiting this.

adutra:
> The comment there is somewhat scary in its implication, but I'm _fairly_ sure it's just an obsolete comment and that we do indeed scope vended credentials to only be readonly when privileges are only readonly. We may need to do some digging (possibly separately from this PR) to make sure our Authz tests correctly reflect the credential read/write scoping.

I noticed this comment too, and I didn't want to modify it during my recent refactoring. These two tests are actually identical: `testLoadTableWithReadAccessDelegationPrivileges` and `testLoadTableWithWriteAccessDelegationPrivileges`. I think they could be merged together: in this test we are focusing on RBAC, not on the actual credential being vended.

dimas-b:
@dennishuo @flyrain : shall we merge this PR? I see you posted comments, but I'm not sure whether you're ok with merging :thinking:

flyrain:
> @dennishuo @flyrain : shall we merge this PR? I see you posted comments, but I'm not sure whether you're ok with merging 🤔

I think @dennishuo  and I have some comments are not resolved, https://github.com/apache/polaris/pull/3927/changes#r2893040582, https://github.com/apache/polaris/pull/3927/changes#r2897948131

dimas-b:
@flyrain : the links you posted above to not lead to any specific comment (perhaps it's just in my browser :thinking: ). Would you mind posting a fresh review to have new GH comments about the issues that still need addressing?

dimas-b:
@Praneeth-Sagar-13 : I see `Claude` as co-author in some of your commits. Please clarify whether you own the changes and in that case consider removing `Claude` as co-author.

Cf. https://github.com/apache/polaris/pull/4106#issuecomment-4192728124

flyrain:
> the links you posted above to not lead to any specific comment

That's a bit weird.  Let me put them here again. Hopefully it works this time: https://github.com/apache/polaris/pull/3927#discussion_r2899017079, https://github.com/apache/polaris/pull/3927#discussion_r2897948131.  However, I think we easily go through the page to find them.

PR #3931, opened 2026-03-04
fix(helm): add quarkus.http.cors.enabled=true when CORS is configured

### LINKED ISSUE #3930: `quarkus.http.cors.enabled=true` missing from Helm chart CORS configuration
### Describe the bug

The Helm chart exposes a `cors` section in `values.yaml` to configure CORS for Polaris. However, the ConfigMap template generates all `quarkus.http.cors.*` properties but omits `quarkus.http.cors.enabled=true`, which is required by Quarkus to actually activate the CORS filter.

### To Reproduce

1. Deploy Polaris with the Helm chart and configure CORS:
```yaml
cors:
  allowedOrigins: ["http://localhost:5173"]
```

2. Send a CORS preflight request:
```bash
curl -v -X OPTIONS http:///api/catalog/v1/config \
  -H "Origin: http://localhost:5173" \
  -H "Access-Control-Request-Method: GET"
```

### Actual Behavior

The generated ConfigMap contains e.g. `quarkus.http.cors.origins=http://localhost:5173` but not `quarkus.http.cors.enabled=true`. Quarkus requires this property to enable the CORS filter — without it, all other CORS properties are ignored.

### Expected Behavior

When any `cors.*` value is set in `values.yaml`, `quarkus.http.cors.enabled=true` should automatically be added to the generated `application.properties`.

### Additional context

Until this is fixed, use `advancedConfig` to manually set the missing property:
```yaml
advancedConfig:
  quarkus.http.cors.enabled: "true"

cors:
  allowedOrigins: ["http://localhost:5173"]
```

### System information

- Polaris Helm chart version: `1.3.0-incubating`

ddecrulle:
Fixes the Helm chart CORS configuration that was silently ignored by Quarkus because quarkus.http.cors.enabled=true was never generated.                                     

Fixes #3930

## Checklist

### DIFF
--- a/helm/polaris/templates/configmap.yaml
+++ b/helm/polaris/templates/configmap.yaml
@@ -115,25 +115,25 @@ data:
 
     {{- /* CORS */ -}}
     {{- if .Values.cors.enabled -}}
-    {{- $_ = set $map "quarkus.http.cors.enabled" "true" -}}
-    {{- end -}}
-    {{- if .Values.cors.allowedOrigins -}}
-    {{- $_ = set $map "quarkus.http.cors.origins" (join "," .Values.cors.allowedOrigins) -}}
-    {{- end -}}
-    {{- if .Values.cors.allowedMethods -}}
-    {{- $_ = set $map "quarkus.http.cors.methods" (join "," .Values.cors.allowedMethods) -}}
-    {{- end -}}
-    {{- if .Values.cors.allowedHeaders -}}
-    {{- $_ = set $map "quarkus.http.cors.headers" (join "," .Values.cors.allowedHeaders) -}}
-    {{- end -}}
-    {{- if .Values.cors.exposedHeaders -}}
-    {{- $_ = set $map "quarkus.http.cors.exposed-headers" (join "," .Values.cors.exposedHeaders) -}}
-    {{- end -}}
-    {{- if .Values.cors.accessControlMaxAge -}}
-    {{- $_ = set $map "quarkus.http.cors.access-control-max-age" .Values.cors.accessControlMaxAge -}}
-    {{- end -}}
-    {{- if ne .Values.cors.accessControlAllowCredentials nil -}}
-    {{- $_ = set $map "quarkus.http.cors.access-control-allow-credentials" .Values.cors.accessControlAllowCredentials -}}
+      {{- $_ = set $map "quarkus.http.cors.enabled" "true" -}}
+      {{- if .Values.cors.allowedOrigins -}}
+      {{- $_ = set $map "quarkus.http.cors.origins" (join "," .Values.cors.allowedOrigins) -}}
+      {{- end -}}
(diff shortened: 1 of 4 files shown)


site/content/in-dev/unreleased/helm-chart/reference.md
```diff
@@ -317,6 +317,8 @@ weight: 900
 
 ### CORS
 
+When any `cors.*` value is set, `quarkus.http.cors.enabled=true` is automatically added to the generated `application.properties`.
```
adutra: I'm afraid this page is generated, you cannot add this paragraph here. You can generate it with `make helm`.

helm/polaris/templates/configmap.yaml
```diff
@@ -114,6 +114,9 @@ data:
     {{- $_ = set $map "quarkus.management.port" (get (first .Values.managementService.ports) "port") -}}
 
     {{- /* CORS */ -}}
+    {{- if or .Values.cors.allowedOrigins .Values.cors.allowedMethods .Values.cors.allowedHeaders .Values.cors.exposedHeaders .Values.cors.accessControlMaxAge (ne .Values.cors.accessControlAllowCredentials nil) -}}
```
adutra: I wonder if we shouldn't simply introduce a `cors.enabled` property in the Helm chart, wdyt?

ddecrulle: The simplest things are often the best. I do that !

helm/polaris/values.yaml
```diff
@@ -1072,6 +1072,10 @@ oidc:
 
 # Polaris CORS configuration.
 cors:
+  # @schema type:[boolean, null]
```
adutra: No need to add null as a valid type here. You can just remove this line and the schema will be (rightfully) inferred to boolean.

helm/polaris/templates/configmap.yaml
```diff
@@ -114,6 +114,9 @@ data:
     {{- $_ = set $map "quarkus.management.port" (get (first .Values.managementService.ports) "port") -}}
 
     {{- /* CORS */ -}}
+    {{- if .Values.cors.enabled -}}
+    {{- $_ = set $map "quarkus.http.cors.enabled" "true" -}}
+    {{- end -}}
```
adutra: I would actually move the end of this if block to line 138; IOW, if this value is false, no CORS-related properties should be included.

adutra:
@dimas-b do you think this deserves a mention in the changelog?

dimas-b:
It's a bugfix, right? CORS support was added before, just did not work, right? I do not think we have to mention all bugfixes in CHANGELOG, but it won't hurt either :)

dimas-b:
On the other hand, we're adding a new helm flag, so it might be worth a brief line in CHANGELOG :)

adutra:
Thanks @ddecrulle for your contribution!

@dimas-b I will update the changelog.

PR #3939, opened 2026-03-05
fix(schema): Add critical performance indexes to resolve create_namespace latency >30s

### LINKED ISSUE #3685: Performance Critical: `create_namespace` Latency (>30s) - Schema & Query Optimization Required
### **1. Issue Summary**

We are experiencing a critical performance degradation where the `create_namespace` API operation consistently times out (>30s), causing cascading failures in our downstream workload infrastructure (Traefik 504 Gateway Timeouts and pyiceberg have an unexpected behaviour).

### **2. Remediation Status**

To restore production stability, we are manually applying the following schema changes:

* **Applied:** `idx_entities_catalog_id_id` on `entities`. (Result: Latency dropped from ~36s to ~17s).
* **Applying Now:** `idx_grants_realm_grantee` and `idx_grants_realm_securable` on `grant_records` to resolve 100% CPU usage during permission checks and GC.

### **3. Root Cause Analysis (Database Layer)**

Our investigation confirmed that the default schema lacks secondary indexes required for the application's actual query patterns.

**A. The "Grant Records" Bottleneck (High Frequency)**
The application performs thousands of permission checks per operation.

* **Observation:** 100% Sequential Scans.
* **Metrics:** 65,000+ executions of `SELECT ... WHERE grantee_id = ...`.
* **Impact:** Without an index starting with `grantee_id` or `realm_id`, the database reads **13.7 Billion rows** to serve these requests.
* **Fix:** We are adding indexes to support lookups by `grantee_id` and `securable_id`.

**B. The "Entities" Bottleneck (Bulk Fetch)**

* **Observation:** Namespace create operation triggers a query with a massive `IN` clause containing ~7,800 tuples: `WHERE (catalog_id, id) IN ((?,?), (?,?)...)`.
* **Impact:** The query planner could not use the PK (starting with `realm_id`) efficiently, resulting in 17s+ execution time.
* **Fix:** We added `idx_entities_catalog_id_id` to optimize this lookup.

### **4. Request for Engineering: Query Optimization**

While the indexes mitigate the immediate outage, the underlying query patterns appear inefficient and likely need refactoring in the Polaris codebase.

**1. Review "N+1" Permission Checks**
We observed **65,415 executions** of the permission check query during a short window.

* *Question:* Is the application iterating through a list and checking permissions one-by-one instead of performing a bulk permission check? This volume suggests an "N+1" query issue that indexes can only partially band-aid.

**2. Review Bulk Entity Fetch Strategy**
The query `SELECT ... FROM entities WHERE (catalog_id, id) IN (...)` is passing **~15,000 bind parameters** (7,800 pairs).

* *Recommendation:* This pattern scales poorly. Please consider paginating this operation or refactoring the logic to avoid passing thousands of IDs in a single SQL statement.

### **5. SQL for Upstream Merge**

We strongly recommend including these indexes in the next Polaris release to prevent similar degradation for other users:

```sql
-- 1. Optimizes High-Volume Permission Checks (Fixes API CPU load)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_grants_realm_grantee 
ON polaris_schema.grant_records (realm_id, grantee_id);

-- 2. Optimizes Garbage Collection (Fixes background job hangs)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_grants_realm_securable 
ON polaris_schema.grant_records (realm_id, securable_id);

-- 3. Optimizes Bulk Entity Lookups (Fixes Namespace listing latency)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_entities_catalog_id_id 
ON polaris_schema.entities (catalog_id, id);

```

### To Reproduce

1. **Environment:** Deploy Apache Polaris backed by PostgreSQL with a moderate dataset (~170k grant records, ~160k entities).
2. **Action:** Execute the `create_namespace` API operation on a catalog with ~22k namespaces.
3. **Conditions:** Ensure standard concurrency; the issue manifests even without much load due to table scans.
4. **Observation:** Monitor API response time and database activity via `pg_stat_statements`.

### Actual Behavior

* **Performance:** The operation consistently takes **33s - 36s** to complete.
* **Database Execution:** Permission checks trigger a **Sequential Scan** of the entire `grant_records` table for every single lookup. We observed queries reading **~174,000 rows per execution** (100% of the table) thousands of times per minute.
* **Client Experience:** The API request times out. Proxies (e.g., Traefik) return `504 Gateway Timeout` at the 30s mark, and client applications (crawlers) hang or fail silently.

### Expected Behavior

* **Performance:** The `create_namespace` operation should complete in **sub-second time to ~2 seconds**.
* **Database Execution:** Permission checks filtering by `grantee_id` should utilize an **Index Scan**, reading only the specific rows relevant to the user (typically <100 rows).
* **Client Experience:** The request should return `200 OK` well within standard proxy/client timeout windows (30s).

### Additional context

All this information was very consistent along the research we did, after applying the first index `idx_entities_catalog_id_id` we seen a consistent drop on the response time from ~31s into ~16s.
After the index was applied the stats were reset and and pg statements cleaned, we kept the system running for +-24h and the data we see in the ticket is from this period after the resets.

### System information

OS: Linux
Polaris Catalog Version: Docker image 1.3.0 (initial symptoms were present in 1.0.1 and we upgraded to validate if this was caused by the version, the problem persisted)
Object storage & setup: AWS S3

machov:
## Problem

Issue #3685 reports critical performance degradation where `create_namespace` API operations consistently timeout (>30s), causing cascading failures with 504 Gateway Timeouts.

## Root Cause Analysis

Database analysis revealed:
- **65,000+ permission check queries** performing full table scans on `grant_records` 
- **Bulk entity lookups** with ~7,800 tuples causing inefficient query plans
- **13.7 billion rows read** during a single create_namespace operation

## Solution

Added index to existing schema v4 with three performance-critical indexes:

1. **`idx_grants_realm_grantee`** on `grant_records(realm_id, grantee_id)`
   - Eliminates sequential scans for permission checks
   
2. **`idx_grants_realm_securable`** on `grant_records(realm_id, securable_id)`  
   - Optimizes grant record cleanup and GC operations
   
3. **`idx_entities_catalog_id_id`** on `entities(catalog_id, id)`
   - Optimizes bulk entity lookups with large IN clauses

## Performance Impact

Based on issue reporter's testing:
- **Before**: create_namespace consistently 33-36 seconds  
- **After**: Latency dropped to **under 2 seconds** 

## Database Compatibility

- ✅ PostgreSQL schema v4 edited
- ✅ H2 schema v4 edited  
- ✅ Uses `IF NOT EXISTS` for safe deployment

Fixes #3685


dimas-b:
@jbonofre : FYI ... I believe we spoke about this in the Community Sync call today and @machov kindly opened a PR with a fix already :tada:

dimas-b (dismissed):
Thanks for your contribution, @machov ! The change LGTM 👍 

schema diff against v4 for reference:

```
19,21c19,21
< -- Changes from v2:
< --  * Added `events` table
< --  * Added `idempotency_records` table for REST idempotency
---
> -- Changes from v4:
> --  * Added performance-critical indexes for grant_records table to fix create_namespace latency (Issue #3685)
> --  * Added optimized index for entities bulk lookups
31c31
< VALUES ('version', 4)
---
> VALUES ('version', 5)
59a60,61
> -- Additional index for bulk entity lookups (Issue #3685)
> CREATE INDEX IF NOT EXISTS idx_entities_catalog_id_id ON entities (catalog_id, id);
98a101,107
> 
> -- Performance-critical indexes for grant_records (Issue #3685)
> -- These indexes resolve create_namespace latency from 30+ seconds to under 2 seconds
> CREATE INDEX IF NOT EXISTS idx_grants_realm_grantee 
>     ON grant_records (realm_id, grantee_id);
> CREATE INDEX IF NOT EXISTS idx_grants_realm_securable 
>     ON grant_records (realm_id, securable_id);
```

Given that quite a few people are involved in JDBC persistence, let's give this PR a few extra days in review.

dimas-b:
@machov : Do you rely on having this fix an a released version soon? Note: 1.4.0 is in the works ATM.

machov:
Can you elaborate what else needs to be done? I see all tests passed

dimas-b:
@machov : The PR is good to merge from my POV, I merely wanted it to have some more time in review in case other interested people have opinions on the new indexes.

The question about 1.4.0 was basically to check whether you need this fix in the 1.4.0 release or you're ok with merging it after 1.4.0.

singhpk234:
couple of feedbacks : 
- do we need a v5, this is technically an additional command ? 
- do we need to ship these indexes or can we add them as recommendation in our docs or code ? 
- obviously do we need this in in 1.4 ?

dimas-b:
> do we need a v5, this is technically an additional command ?

I'm personally fine with adding the new indexes to v4 DDL files.

However, from a more rigorous perspective, it makes sense to version the schema every time there is a material change. This way, it is easier to track how the Polaris database is expected to behave... For example, we could (hypothetically) deny the expensive operations with a v4 schema.

I do not mean to do that in current PR, just exposing options to consider :slightly_smiling_face:

flyrain (dismissed):
Thanks for the change. Echo @singhpk234, we probably reuse the v4 as 1.4.0 isn't release yet.

dimas-b:
Good point - I missed that v4 schema was added after 1.3.0. Let's update v4 in this PR and merge it before 1.4.0 then. Adding to milestone.

machov:
Do we need a v5 (since this is “just” additional DDL)?
Good questions.
I think we do need v5. v5 schema bump is the safest way to ensure the indexes get applied for *existing* installs. Updating v4 only helps fresh installs; it won’t automatically add indexes to databases already initialized at v4.

Do we need to ship these indexes vs recommending in docs/code?
We should ship them. This change takes create_namespace from 30s+ (timeouts) down to ~<2s, so it’s not an optional tuning tip at scale. Shipping via schema/migrations avoids putting manual DDL burden on operators. (Happy to also add a short doc note explaining the rationale.)

Do we need this in 1.4?
Yes. I can backport it to the 1.4.x line as a post-release schema bump (v5) with migration notes (what indexes are created + expected impact/lock behavior). This way 1.4 users get the fix through the normal upgrade flow without changing already-shipped v4 semantics.

flyrain:
Hi @machov, I think we all agreed on introducing a new schema version for this change. Since 1.4.x has not been released yet, we could simply include this PR as part of the 1.4.0 release. That way there is no need for a backport, and all future releases will automatically include the indexes as well. WDYT?

dimas-b:
@machov : just for my own understanding: Do you use releases of do you deploy Polaris from `main`?

machov:
@flyrain  should i change the target branch then?

@dimas-b  I’m contributing as a community member and don’t operate a Polaris deployment myself.

flyrain:
No need to change the branch ATM, as the 1.4.0 release branch hasn't cut yet.

Hi @adnanhemani, can we include this PR in the 1.4.0?

dimas-b:
@machov : I generally support your point about introducing schema v5 in this case.

However, as @flyrain mentioned, the v4 schema has never been released. It was added after 1.3.0 and 1.4.0 is still in early stages (there's no branch for it ATM).

All in all, I think in this particular case it is preferable to add the new indexes to the v4 schema and release them in 1.4.0 (soon).

Downstream users who pull directly from `main` might have to re-apply the new revision of the v4 schema, but AFAIK, the DDL is idempotent, but it should work.

adnanhemani:
Yes, I think this is reasonable to put in 1.4.0, given we are still waiting on some last license and notice stuff.

And agreed with @dimas-b, @flyrain, and @singhpk234 - given that v4 was never in a public release, we should just merge the index changes onto that. I will hold the release until we are able to get this PR sorted (if possible within this week)!

Thank you for your great work, @machov!

pjanuario:
Once we push this into the 1.4 I will update our instalation and re-enable one of the problematic catalogs that trigger the issue, but i suspect this will not be enough as we did applied those indexes and still the response time decreased dramatically but still was over 30s in some cases. Let's wait and test again after the release.

@machov thanks for the fix.

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
```diff
@@ -97,6 +98,11 @@ COMMENT ON COLUMN grant_records.grantee_catalog_id IS 'catalog id of the grantee
 COMMENT ON COLUMN grant_records.grantee_id IS 'id of the grantee';
 COMMENT ON COLUMN grant_records.privilege_code IS 'privilege code';
 
+CREATE INDEX IF NOT EXISTS idx_grants_realm_grantee 
+    ON grant_records (realm_id, grantee_id);
```
jbonofre, line 102: The index should include `grantee_catalog_id` no ?

persistence/relational-jdbc/src/main/resources/h2/schema-v4.sql
```diff
@@ -97,6 +98,11 @@ COMMENT ON COLUMN grant_records.grantee_catalog_id IS 'catalog id of the grantee
...
+CREATE INDEX IF NOT EXISTS idx_grants_realm_securable 
+    ON grant_records (realm_id, securable_id);
```
jbonofre, line 104: This one seems already covered by the primary key index. 
If we want this index, I think we should include `securable_catalog_id`, no ?

persistence/relational-jdbc/src/main/resources/postgres/schema-v4.sql
```diff
@@ -57,6 +57,7 @@ CREATE TABLE IF NOT EXISTS entities (
 
 -- TODO: create indexes based on all query pattern.
 CREATE INDEX IF NOT EXISTS idx_entities ON entities (realm_id, catalog_id, id);
+CREATE INDEX IF NOT EXISTS idx_entities_catalog_id_id ON entities (catalog_id, id);
```
jbonofre, line 60: Is it not already covered by `idx_entities` index ?

PR #3940, opened 2026-03-05
core: Standardize Path Lookup Key for `PolarisResolutionManifest.getResolvedPath`

sungwy:
Today, plain Java `Object` is used as the Path Lookup key in `PolarisResolutionManifest.getResolvedPath`. This poses a problem, because a  Polaris Entity like `Namespace` or `PolicyIdentifier` could be used as the object key, or other aliases referring to the same object like `ResolverPath` or `PolarisSecurable` could also mistakenly be used as the object key, resulting in an unexpected cache miss in the `Resolver`.

This PR standardizes the path look up key as `ResolverPathKey(List<String> entityNames, PolarisEntityType entityType)`, so that there is a consistent way to look up the same logical Polaris Entity, wherever in the code base we may be fetching the resolved Path.

## Checklist

### DIFF
--- a/runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
+++ b/runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
@@ -76,6 +76,11 @@ public class PolarisPassthroughResolutionView implements PolarisResolutionManife
   @Override
   public PolarisResolvedPathWrapper getResolvedPath(
       ResolvedPathKey key, PolarisEntitySubType subType) {
+    Preconditions.checkState(
+        key.entityType() == PolarisEntityType.TABLE_LIKE
+            || key.entityType() == PolarisEntityType.POLICY,
+        "Trying to getResolvedPath(key, subType) for unsupported key %s",
+        key);
     PolarisResolutionManifest manifest = newResolutionManifest();
     manifest.addPath(new ResolverPath(key.entityNames(), key.entityType()));
     manifest.resolveAll();
@@ -96,6 +101,11 @@ public class PolarisPassthroughResolutionView implements PolarisResolutionManife
   @Override
   public PolarisResolvedPathWrapper getPassthroughResolvedPath(
       ResolvedPathKey key, PolarisEntitySubType subType) {
+    Preconditions.checkState(
+        key.entityType() == PolarisEntityType.TABLE_LIKE
+            || key.entityType() == PolarisEntityType.POLICY,
+        "Trying to getPassthroughResolvedPath(key, subType) for unsupported key %s",
+        key);
     PolarisResolutionManifest manifest = newResolutionManifest();
     manifest.addPassthroughPath(new ResolverPath(key.entityNames(), key.entityType()));
     return manifest.getPassthroughResolvedPath(key, subType);


runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -384,8 +387,7 @@ private PolarisResolutionManifest authorizeGrantOnCatalogRoleToPrincipalRoleOper
       String principalRoleName) {
     PolarisResolutionManifest resolutionManifest = newResolutionManifest(catalogName);
     resolutionManifest.addPath(
-        new ResolverPath(List.of(catalogRoleName), PolarisEntityType.CATALOG_ROLE),
-        catalogRoleName);
+        new ResolverPath(List.of(catalogRoleName), PolarisEntityType.CATALOG_ROLE));
```
adutra: I am observing the following similarity:

```java
private record ResolverPathKey(List<String> entityNames, PolarisEntityType entityType) {}
public  record ResolverPath   (List<String> entityNames, PolarisEntityType lastEntityType, boolean optional) {}
```

Wouldn't it be better to define these records by composition? E.g.:

```java
public record ResolverPathKey(List<String> entityNames, PolarisEntityType entityType) {}
public record ResolverPath(ResolverPathKey key, boolean optional) {}

File: polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifestCatalogView.java [hunk 2]
```diff
@@ -39,13 +40,15 @@ public interface PolarisResolutionManifestCatalogView {
         .orElse(null);
   }
 
-  PolarisResolvedPathWrapper getResolvedPath(Object key);
+  PolarisResolvedPathWrapper getResolvedPath(
+      List<String> entityNames, PolarisEntityType entityType);
```
adutra: Would it be simpler to expose ResolverPathKey publicly and transform these methods as follows:

```java
  PolarisResolvedPathWrapper getResolvedPath(ResolvedPathKey key);

  PolarisResolvedPathWrapper getResolvedPath(
      ResolvedPathKey key, PolarisEntitySubType subType);

  PolarisResolvedPathWrapper getPassthroughResolvedPath(
      ResolvedPathKey key);

  PolarisResolvedPathWrapper getPassthroughResolvedPath(
      ResolvedPathKey key, PolarisEntitySubType subType);
```

Then `ResolvedPathKey` could expose some factory methods to quickly create keys from different types of entities. E.g.

```java
resolvedEntityView.getResolvedPath(
    PolarisCatalogHelpers.tableIdentifierToList(tableIdentifier),
    PolarisEntityType.TABLE_LIKE,
    PolarisEntitySubType.ICEBERG_TABLE);
```

Becomes:

```java
resolvedEntityView.getResolvedPath(
    ResolvedPathKey.ofTable(tableIdentifier),
    PolarisEntitySubType.ICEBERG_TABLE);
```

Just a suggestion :-)

sungwy: yeah I really like this suggestion. let me adopt this

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -57,11 +57,11 @@ public class PolarisResolutionManifest implements PolarisResolutionManifestCatal
   private final Resolver primaryResolver;
   private final PolarisDiagnostics diagnostics;
 
-  private final Map<Object, Integer> pathLookup = new HashMap<>();
+  private final Map<ResolvedPathKey, Integer> pathLookup = new HashMap<>();
```
dimas-b, line 60: I believe this map is generally used only to allow caller to retrieved resolved entities for some "seed" parameters.

Some time ago I tried refactoring this code in #499 with the idea that callers would get back "handle" objects for "seed" parameter they put into the resolver flow.

That PR did not merge, but I still wanted to mention it again. I wonder what you think about it. This is just an idea for consideration. No pressure 😉 

I believe callers never need to use arbitrary lookup keys (at least they did not do that at the time of that old PR), they always obtain the resolved object for some kind of input parameter they already have, so the "handle" objects can usually be carried around as local variables.

dimas-b: Ah, about why "handle" objects are beneficial: the callers can directly obtain the resolved entity from the handle object, no need to create `ResolvedPathKey` objects just for the sake of map lookup.

Also, the handle is always directly related to the result on the caller side (local var.), which reduces risk of confusion when an unrelated key is used for lookup.

dimas-b: Actually, what I wrote above is not exactly what the old PR was doing... I guess I was dreaming a bit 😅 

In any case, what I mean is this:

```
Handle h = resolutionManifest.addPath(...);
// resolve
PolarisResolvedPathWrapper r = h.get();
```
WDYT?

sungwy: Hi @dimas-b - funny you mention this, because I was thinking just the same thing. But I fell short of proposing it because:
1. It wasn't absolutely necessary for the `PolarisAuthorizer` SPI refactoring needs
2. I wasn't really sure if there was some hidden sauce reason why we were doing the `Integer` based lookup...

Since that's an underlying implementation detail, how about we leave that out of scope of this PR and revisit it after a careful, dedicated review? cc @dennishuo

dimas-b: Sure. Totally fine to leave this out.

dimas-b:
I second comments from @adutra above, but in general this PR LGTM :+1: 

Since it touches `polaris-core`, I hope @dennishuo could review too.

sungwy:
Hi @dennishuo - just a bump to get your attention on this PR

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolverPath.java
```diff
@@ -24,14 +24,14 @@
...
- *     table_like entity versus a namespace
+ * @param key canonical path key (entity names + terminal entity type)
  * @param optional true if this path is optional, i.e. failing to fully resolve it is not an error
  */
-public record ResolverPath(
-    List<String> entityNames, PolarisEntityType lastEntityType, boolean optional) {
+public record ResolverPath(ResolvedPathKey key, boolean optional) {
+
+  public ResolverPath {
+    key = new ResolvedPathKey(key.entityNames(), key.entityType());
```
adutra: nit: do we need this defensive copy?

dimas-b: `ResolvedPathKey` is already immutable, so why bother, indeed 🤔

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolvedPathKey.java
```diff
@@ -0,0 +1,60 @@
...
+  public ResolvedPathKey {
+    entityNames = List.copyOf(entityNames);
+  }
+
+  public static ResolvedPathKey of(List<String> entityNames, PolarisEntityType entityType) {
+    return new ResolvedPathKey(entityNames, entityType);
+  }
+
+  public static ResolvedPathKey of(ResolverPath path) {
+    return new ResolvedPathKey(path.entityNames(), path.lastEntityType());
```
dimas-b: nit: `return path.key()`?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -112,29 +112,24 @@ public void addTopLevelName(String entityName, PolarisEntityType entityType, boo
...
-  /**
-   * Adds a path that is allowed to be dynamically resolved with a new Resolver when
-   * getPassthroughResolvedPath is called. These paths are also included in the primary static
-   * resolution set resolved during resolveAll().
-   */
-  public void addPassthroughPath(ResolverPath path, Object key) {
-    addPath(path, key);
-    passthroughPaths.put(key, path);
+  private static ResolvedPathKey resolvedPathKey(ResolverPath path) {
+    return ResolvedPathKey.of(path);
```
dimas-b: nit: why not inline this trivial method redirect?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -342,25 +316,38 @@ public PolarisEntitySubType getLeafSubType(Object key) {
...
+    if (!getIsPassthroughFacade()
+        && resolvedPath.getRawLeafEntity() != null
+        && subType != PolarisEntitySubType.ANY_SUBTYPE
+        && resolvedPath.getRawLeafEntity().getSubType() != subType) {
+      return null;
+    }
+    return resolvedPath;
+  }
 
+  private PolarisResolvedPathWrapper getResolvedPathByIndex(
```
dimas-b: nit: there is only one trivial caller... do we need a separate method for this? 🤔

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/PolarisResolutionManifest.java
```diff
@@ -342,25 +316,38 @@ public PolarisEntitySubType getLeafSubType(Object key) {
...
         pathLookup.containsKey(key),
         "never_registered_key_for_resolved_path",
         "key={} pathLookup={}",
         key,
         pathLookup);
+    return getResolvedPathByIndex(pathLookup.get(key), prependRootContainer);
+  }
+
+  public PolarisResolvedPathWrapper getResolvedPath(
+      ResolvedPathKey key, PolarisEntitySubType subType, boolean prependRootContainer) {
```
dimas-b: nit: moving this to approx. old line 392 will probably simplify the diff.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -303,10 +305,10 @@ private PolarisResolutionManifest authorizeBasicCatalogRoleOperationOrThrow(
       PolarisAuthorizableOperation op, String catalogName, String catalogRoleName) {
     PolarisResolutionManifest resolutionManifest = newResolutionManifest(catalogName);
     resolutionManifest.addPath(
-        new ResolverPath(List.of(catalogRoleName), PolarisEntityType.CATALOG_ROLE),
-        catalogRoleName);
+        new ResolverPath(List.of(catalogRoleName), PolarisEntityType.CATALOG_ROLE));
```
dimas-b: nit: `ResolverPathKey.ofCatalogRole(catalogRoleName).toPath()`?

runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
```diff
@@ -65,89 +60,34 @@ public PolarisResolvedPathWrapper getResolvedReferenceCatalogEntity() {
...
-          new ResolverPath(Arrays.asList(namespace.levels()), PolarisEntityType.NAMESPACE),
-          namespace);
-      manifest.resolveAll();
-      return manifest.getResolvedPath(namespace);
-    } else {
-      throw new IllegalStateException(
-          String.format(
-              "Trying to getResolvedPath(key) for %s with class %s", key, key.getClass()));
-    }
+    manifest.addPath(new ResolverPath(key.entityNames(), key.entityType()));
```
dimas-b, line 71: Should we assert that the type inside `key` is still `NAMESPACE` for backward compatibility?... although I personally think the new code is more correct.

dimas-b: @dennishuo : is this critical?

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolverPath.java
sungwy: Good catch, I'll remove it since it's already immutable

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolvedPathKey.java
sungwy: thank you for the nit. I didn't update this after I took the suggestion to adopt composition

runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
sungwy: I wasn’t fully sure why this test fixture previously restricted it to Namespace only, since passthrough resolution can conceptually work for other path types too. With `ResolvedPathKey`, we now have a uniform key, so I generalized it. 

That said, I agree we should preserve backward compatibility in this PR and keep focus on the core key-standardization changes.

We can revisit test fixture related refactors separately.

runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
```diff
@@ -65,89 +62,42 @@ public PolarisResolvedPathWrapper getResolvedReferenceCatalogEntity() {
...
-      manifest.addPath(
-          new ResolverPath(
-              PolarisCatalogHelpers.identifierToList(
-                  policyIdentifier.namespace(), policyIdentifier.name()),
-              entityType),
-          policyIdentifier);
-      manifest.resolveAll();
-      return manifest.getResolvedPath(policyIdentifier, entityType, subType);
-    } else {
-      throw new IllegalStateException(
```
dimas-b, line 105: now that the Namespace type check got restored, should we restore type checks here as well?.

runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
```diff
@@ -65,89 +62,42 @@ public PolarisResolvedPathWrapper getResolvedReferenceCatalogEntity() {
...
-    } else if (key instanceof PolicyIdentifier policyIdentifier) {
-      manifest.addPassthroughPath(
-          new ResolverPath(
-              PolarisCatalogHelpers.identifierToList(
-                  policyIdentifier.namespace(), policyIdentifier.name()),
-              entityType),
-          policyIdentifier);
-      return manifest.getPassthroughResolvedPath(policyIdentifier, entityType, subType);
-    } else {
-      throw new IllegalStateException(
```
dimas-b, line 147: restore this check too?

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolvedPathKey.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.polaris.core.entity.PolarisEntityType;
+
+/** Canonical lookup key for resolved paths in a {@link PolarisResolutionManifest}. */
+public record ResolvedPathKey(List<String> entityNames, PolarisEntityType entityType) {
+
+  public ResolvedPathKey {
+    entityNames = List.copyOf(entityNames);
+  }
+
+  public static ResolvedPathKey of(List<String> entityNames, PolarisEntityType entityType) {
```
sungwy, line 35: note to reviewer: this constructor isn't used right now outside of tests, but I anticipate this will be used when we construct the key from within the `PolarisAuthorizerImpl`

runtime/service/src/testFixtures/java/org/apache/polaris/service/catalog/PolarisPassthroughResolutionView.java
sungwy: sure thing. Thank you!

polaris-core/src/main/java/org/apache/polaris/core/persistence/resolver/ResolvedPathKey.java
dimas-b: WFM :+1:

dimas-b (approved):

PR #3941, opened 2026-03-05
Add OPTION and HEAD to quarkus.http.cors.methods

### LINKED ISSUE #3938: Helm chart: CORS `allowedMethods` default does not include HEAD, breaking DuckDB connectivity

The Helm chart comment for `allowedMethods` states:
```yaml
# If this is not set or empty, all requested methods are considered allowed.
allowedMethods: []
```

In practice, leaving this empty breaks clients like **DuckDB** that issue 
`HEAD` requests to the Iceberg REST Catalog — those requests are rejected 
with a CORS error.

## Root cause

Quarkus does not include `HEAD` in its CORS allowed methods when the property 
is left empty, despite its own documentation suggesting otherwise. This may be 
an intentional choice in Quarkus, but it means Polaris's default Helm 
configuration silently breaks clients that rely on `HEAD`.

## Impact

Clients like **DuckDB** (including the WASM build) issue `HEAD` requests when 
probing Iceberg REST Catalog endpoints. Without `HEAD` explicitly listed, 
these requests fail with a CORS error, preventing DuckDB-based UIs from 
connecting to Polaris.

The Helm chart should ship with `HEAD` explicitly included in the defaults:
```yaml
allowedMethods: [GET, HEAD, PUT, POST, DELETE, OPTIONS]
```

_No response_

### Additional context

_No response_

nandorKollar:
Fixes #3938

Add HEAD and OPTIONS to default allowed method list for CORS.

## Checklist

### DIFF
--- a/runtime/defaults/src/main/resources/application.properties
+++ b/runtime/defaults/src/main/resources/application.properties
@@ -57,12 +57,13 @@ quarkus.http.auth.proactive=false
 quarkus.http.body.handle-file-uploads=false
 quarkus.http.limits.max-body-size=10240K
 
-quarkus.http.cors.origins=http://localhost:8080
-quarkus.http.cors.methods=PATCH, POST, DELETE, GET, PUT, HEAD, OPTIONS
-quarkus.http.cors.headers=*
-quarkus.http.cors.exposed-headers=*
-quarkus.http.cors.access-control-max-age=PT10M
-quarkus.http.cors.access-control-allow-credentials=true
+quarkus.http.cors.enabled=false
+#quarkus.http.cors.origins=http://localhost:8080
+#quarkus.http.cors.methods=PATCH, POST, DELETE, GET, PUT, HEAD, OPTIONS
+#quarkus.http.cors.headers=*
+#quarkus.http.cors.exposed-headers=*
+#quarkus.http.cors.access-control-max-age=PT10M
+#quarkus.http.cors.access-control-allow-credentials=true
 
 quarkus.http.port=8181
 quarkus.http.test-port=0
--- a/site/content/in-dev/unreleased/configuration/configuring-polaris.md
+++ b/site/content/in-dev/unreleased/configuration/configuring-polaris.md
@@ -87,25 +87,19 @@ For a comprehensive reference of all Polaris configuration options, see the [Con
 Some often useful Quarkus configuration properties are listed below.
 Please refer to the [Quarkus documentation](https://quarkus.io/guides/) for more details.
 
-| Configuration Property                               | Default Value                                  | Description                                                                                    |
-|------------------------------------------------------|------------------------------------------------|------------------------------------------------------------------------------------------------|
-| `quarkus.log.level`                                  | `INFO`                                         | Define the root log level.                                                                     |
-| `quarkus.log.category."org.apache.polaris".level`    |                                                | Define the log level for a specific category.                                                  |
-| `quarkus.default-locale`                             | System locale                                  | Force the use of a specific locale, for instance `en_US`.                                      |
-| `quarkus.http.port`                                  | `8181`                                         | Define the HTTP port number.                                                                   |
-| `quarkus.http.auth.basic`                            | `false`                                        | Enable the HTTP basic authentication.                                                          |
-| `quarkus.http.limits.max-body-size`                  | `10240K`                                       | Define the HTTP max body size limit.                                                           |
-| `quarkus.http.cors.enabled`                          | `false`                                        | Enable the HTTP CORS filter. Must be set to `true` for any other CORS property to take effect. |
-| `quarkus.http.cors.origins`                          |                                                | Define the HTTP CORS origins.                                                                  |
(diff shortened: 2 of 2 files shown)


site/content/in-dev/unreleased/configuration/configuring-polaris.md
```diff
@@ -87,25 +87,25 @@ For a comprehensive reference of all Polaris configuration options, see the [Con
...
+| `quarkus.http.auth.basic`                            | `false`                                        | Enable the HTTP basic authentication.                                                          |
+| `quarkus.http.limits.max-body-size`                  | `10240K`                                       | Define the HTTP max body size limit.                                                           |
+| `quarkus.http.cors.enabled`                          | `false`                                        | Enable the HTTP CORS filter. Must be set to `true` for any other CORS property to take effect. |
+| `quarkus.http.cors.origins`                          |                                                | Define the HTTP CORS origins.                                                                  |
+| `quarkus.http.cors.methods`                          | `PATCH, POST, DELETE, GET, PUT, HEAD, OPTIONS` | Define the HTTP CORS covered methods.                                                          |
```
adutra: What "Default Value" means here is ambiguous. I would rename the column to "Default Value in Polaris" to clarify things.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -58,7 +58,7 @@ quarkus.http.body.handle-file-uploads=false
 quarkus.http.limits.max-body-size=10240K
 
 quarkus.http.cors.origins=http://localhost:8080
-quarkus.http.cors.methods=PATCH, POST, DELETE, GET, PUT
+quarkus.http.cors.methods=PATCH, POST, DELETE, GET, PUT, HEAD, OPTIONS
```
adutra: That is the default value in Quarkus, so I wonder if it's simpler to just comment this out?

Or better yet: I think we should include `quarkus.http.cors.enabled` (it's probably an oversight), and comment out all the rest:

```properties
quarkus.http.cors.enabled=false
#quarkus.http.cors.origins=http://example.com:8080
#quarkus.http.cors.methods=PATCH, POST, DELETE, GET, PUT, OPTIONS, HEAD
#quarkus.http.cors.headers=*
#quarkus.http.cors.exposed-headers=*
#quarkus.http.cors.access-control-max-age=PT10M
#quarkus.http.cors.access-control-allow-credentials=true
```

WDYT?

nandorKollar: Hmm, should we disable CORS filters? The rational would be that CORS on REST endpoints it isn't less critical, than let's say on a website?

adutra: They are disabled by default. The default value of `quarkus.http.cors.enabled` is `false`.

nandorKollar: Oh okay true, the doc also says that. Yes, in this case better to comment out comment out those configs. I think we should also remove those from `configuring-polaris.md` too, as Quarkus defaults are used then.

site/content/in-dev/unreleased/configuration/configuring-polaris.md
dimas-b: +1

runtime/defaults/src/main/resources/application.properties
adutra: > I think we should also remove those from `configuring-polaris.md` too, as Quarkus defaults are used then.

Well this section attempts to present common configs that most users would want to configure. I think CORS deserves to be there, wdyt?

nandorKollar: It does, but only those, which have some non-default values no? If we don't set any default for `quarkus.http.cors.*`, just add `quarkus.http.cors.enabled=false`, then I think it is sufficient do document only `quarkus.http.cors.enabled` no?

I'm wondering why `quarkus.http.cors.*` has any non-default value in our application properties, when cors is turned off? Is it because if someone turns it on, then we think that these values are the safest option?

dimas-b: `quarkus.http.cors.*` values were added for [Polaris UI](https://github.com/apache/polaris-tools/tree/main/console), IIRC.

Yet, if we switch to Quarkus default (which have wider scope), it should not hurt UI, I think.

I think it is sufficient to document only what we explicitly change (or instruct users to change) in Polaris compared to Quarkus defaults. Interested users can always go to Quarkus config docs (which are pretty easy to navigate in my experience).

adutra: > I'm wondering why `quarkus.http.cors.*` has any non-default value in our application properties, when cors is turned off? Is it because if someone turns it on, then we think that these values are the safest option?

No, imho these values were meant as (dummy) examples.

dimas-b:
This change looks backward-compatible to me and worth having in 1.4.0, so I'm going to merge now. Happy to reconsider if concerns are raised after merging.

dimas-b:
Thanks for the fix, @nandorKollar !

Issue #3855, opened 2026-02-20
Refactor Admin-CLI such that the shell script can farm out commands to each module


keithgchapman:
### Is your feature request related to a problem? Please describe.

Currently, the Admin CLI is built and packaged as a fully self-contained unit. It would be beneficial to restructure it so that dependencies can be added as needed, rather than bundling everything by default. For example, the NoSQL commands could be moved into the NoSQL module, with the Admin CLI invoking them from there.

This will become increasingly important as we continue adding more commands to the Admin CLI. Now would be a good time to revisit and improve this aspect of the design.

This discussion came up in this [dev ML thread](https://lists.apache.org/thread/9djtyorskydb6yrzjby14mkr8b7czo8r)

dimas-b:
@flyrain : I believe this relates to your proposal for Admin / Server refactor in #3340. Let's reset and resume this work from scratch. I'm very interested in sorting dependency issues out.

jbonofre:
What's the status on this one? I bumped to 1.6.0 for now, waiting for an update.

PR #3947, opened 2026-03-06
Split Admin CLI into modules per technology

### LINKED ISSUE #3867: Add build-time flags to produce Admin CLI without NoSQL dependencies

This is a short-term follow-up to #3395.

In some environments, it may not be desirable to pull in NoSQL dependencies into the Admin CLI command.

This can be done by modularizing individual CLI commands and having downstream project rebuild the Admin CLI with only required dependencies.

However, that creates some burden on the downstream project in terms of developing the build system.

This issue proposes for Polaris to offer built-time flags so that Polaris build scripts could be used to compile the Admin CLI with (OSS default) or without NoSQL commands as user's discretion. 

This would be similar to the existing `NonRESTCatalogs` build flag.

_No response_

_No response_

### Additional context

_No response_

dimas-b:
* Introduce `polairs-admin-base` for shared commands

* Move NoSQL commands to `polaris-persistence-nosql-admin`

* Make NoSQL commands optional with build property: `ExcludeNoSqlPersistence` as proposed in #3867

* Move JDBC-specific dependencies for admin commands to `polaris-relational-jdbc-admin`

* CDI-based Sub-command discovery is handled by `AdminCliConfiguration`

* Add dedicated admin test modules per persistence technology (JDBC / NoSQL)

* Move default quarkus properties to `runtime-admin-defaults` to allow downstream projects to _not_ include them.

* Downstream builds are advised to depend on individual command modules (noted above) and have local `application.properties`, but not depend on `runtime-admin`, which serves as the Admin CLI distribution assembly module for Apache Polaris releases.

Closes #3867

This change also addresses most of concerns from #3855.

## Checklist


persistence/nosql/admin/cmd/src/main/java/org/apache/polaris/admintool/nosql/Placeholder.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.polaris.admintool.nosql;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+/**
+ * Something with a "bean-producing" annotation is required by Quarkus to "notice" this module, yet
+ * Quarkus will find CLI commands only if they are `@Default`, so `@ApplicationScoped` cannot be
+ * applied to them... hence the workaround with this placeholder just to hold the
+ * `@ApplicationScoped` annotation.
+ */
```
jbonofre, line 29: I understand the workaround. I just wonder if it's actually required ?

persistence/nosql/admin/cmd/src/main/java/org/apache/polaris/admintool/nosql/BaseNoSqlCommand.java
```diff
@@ -25,6 +25,10 @@
 public abstract class BaseNoSqlCommand extends BaseCommand {
   @Inject protected Backend backend;
 
+  public BaseNoSqlCommand() {
```
jbonofre: nit: Do we really need this constructor ?

persistence/nosql/admin/cmd/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
```diff
@@ -34,6 +36,8 @@
     },
     mixinStandardHelpOptions = true,
     description = "Sub-commands specific to NoSQL persistence.")
+@AdminSubCommand
+@Default
 public class NoSqlCommand extends BaseNoSqlCommand {
   @Inject protected Backend backend;
```
jbonofre: `backend` is already defined in `BaseNoSqlCommand`. Do we really need to override it there ?

persistence/nosql/admin/cmd/src/main/java/org/apache/polaris/admintool/nosql/Placeholder.java
dimas-b: Yes. Quarkus does not "see" custom qualifiers (like `@AdminSubCommand`). Since `persistence/nosql/admin/cmd` does not have any standard "bean-defining" annotations, Quarkus will not consider it when looking for `@AdminSubCommand` implementations.

I guess this is an interplay of some not-so-standard Quarkus approach to implementing CLI (note that Quarkus considered only `@Default` beans for CLI commands) and me wanting to leverage `@AdminSubCommand` to allow pluggable CLI modules.

persistence/nosql/admin/cmd/src/main/java/org/apache/polaris/admintool/nosql/BaseNoSqlCommand.java
dimas-b: thx - leftover... removing

persistence/nosql/admin/cmd/src/main/java/org/apache/polaris/admintool/nosql/NoSqlCommand.java
dimas-b: Probably not... It was there before, I just did not notice 😅 Thx for the pointer - removing.

dimas-b:
Note: I rebased to include new tests from #3352

flyrain (commented):
Thanks a lot for working on it, @dimas-b! 

Some environments should be able to build the Admin CLI without pulling in NoSQL dependencies. But I think this PR addresses that with too much structural churn. Splitting the admin CLI into this many new modules increases build, test, and dependency complexity quite a bit.

A simpler approach would be to keep the main Admin CLI structure intact and support this through Polaris-owned build-time selection. That could mean a build flag and minimal isolation for the NoSQL-specific maintenance commands sources, for example via a small optional NoSQL admin command module or a separate source set, rather than introducing so many new Gradle projects.

Another option would be using the shell script to redirect to the persistence-specific command. For example, when `polaris-admin-tool.jar nosql  maintenance-run` was invoked, the admin script[1] can direct to a NoSQL maintenance related jar file. In that case, we didn't have to introduce a build flag while keeping persistence-specific commands in their own modules. 

WDTY?

1. https://github.com/apache/polaris/blob/83200d53202b7b66071b873fc34e7e74d9124c26/runtime/distribution/bin/admin#L20

dimas-b:
@flyrain : I'm not sure I understand what you mean exactly, but I'm open to alternatives. If you feel like opening an alternative PR, I think that can allow for different approaches to be compared more easily.

PR #3948, opened 2026-03-06
doc: introduce guidelines for AI-generated contributions

flyingImer:
This PR proposes adding a short guideline around AI-assisted contributions. The goal is not to restrict how contributors use development tools, but to clarify contributor accountability.

In particular, the guideline emphasizes that the human submitting a pull request:
- remains the author of the change
- must understand the implementation end-to-end
- must be able to justify the design and code during review

This is similar in spirit to recent discussions and updates in other ASF projects (for example Apache Iceberg). This PR is intended as a starting point for discussion and wording refinement.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,7 +127,7 @@ Contributors may use a variety of tools when preparing changes to Polaris, inclu
 * The pull request author **must understand the implementation end-to-end** and be able to **explain and justify the design and code** during review.
 * Tools, including AI systems, **are not** considered contributors. **Responsibility and authorship remain with the human** submitting the change.
 * Contributors are encouraged to **disclose** significant AI assistance in the pull request description for transparency.
-* **Respect ASF policy**. Ensure generated content does not introduce incompatible licenses or undisclosed third-party code; review the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) and licensing rules when in doubt.
+* **Respect ASF policy**. Contributors are responsible for ensuring generated code can be contributed to Polaris under ASF license and [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).
 
 ## Java version requirements
 


CONTRIBUTING.md
```diff
@@ -119,6 +119,16 @@ Tips:
...
 
+#### Guidelines for AI-assisted Contributions
+
+Contributors may use a variety of tools when preparing changes to Polaris, including AI systems (e.g., large language models or code assistants). Contributors using such systems are expected to adhere to the following principles:
+
+* Regardless of how a change is produced, the individual submitting the pull request is considered the author of the contribution and is fully responsible for it.
+* The pull request author **must understand the implementation end-to-end** and be able to **explain and justify the design and code** during review.
+* Tools, including AI systems, **are not** considered contributors. **Responsibility and authorship remain with the human** submitting the change.
+* Contributors are encouraged to **disclose** significant AI assistance in the pull request description for transparency.
+* **Respect ASF policy**. Ensure generated content does not introduce incompatible licenses or undisclosed third-party code; review the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) and licensing rules when in doubt.
```
dimas-b: `does not introduce incompatible licenses` seems a bit vague IMHO. How about `the author is responsible for ensuring that the generated code can be contributed to Polaris under the ASF license and the Generative Tooling policy ...`?

flyingImer: love it!

dimas-b (approved):

dimas-b:
@jbonofre : wdyt?

CONTRIBUTING.md
```diff
@@ -119,6 +119,16 @@ Tips:
...
 
+#### Guidelines for AI-assisted Contributions
+
+Contributors may use a variety of tools when preparing changes to Polaris, including AI systems (e.g., large language models or code assistants). Contributors using such systems are expected to adhere to the following principles:
+
+* Regardless of how a change is produced, the individual submitting the pull request is considered the author of the contribution and is fully responsible for it.
+* The pull request author **must understand the implementation end-to-end** and be able to **explain and justify the design and code** during review.
+* Tools, including AI systems, **are not** considered contributors. **Responsibility and authorship remain with the human** submitting the change.
+* Contributors are encouraged to **disclose** significant AI assistance in the pull request description for transparency.
```
flyrain, line 129: Given that AI generated code and PRs put a lot of burden on the reviewers. Should we add something like

Authors must ensure that AI assisted pull requests remain focused, well explained, and of a size that can be reasonably reviewed by humans. Submissions that significantly increase reviewer burden may be asked to be reduced in scope or revised before review.

dimas-b: That applies to all PRs, IMHO :wink:

flyrain: it is, however, generated code tends to cover more things than manually written code, which make the situation worse. Also an explicit rule make it clear that keeping PRs smaller are prefereed

jbonofre: Something important also to remember is that we can't have copyright on generated source (AI or other tool).
So ASF header should not be added on a class where the majority of the code is generated.

jbonofre: Maybe we should ask to put a header like:

```
/**
 * This code has been generated by AI Foo Model
 */
```

dimas-b: Re: ASF header: It says `Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements`. 

If we can assume that the contributor did due diligence per https://www.apache.org/legal/generative-tooling.html and ensured that the generated code is not copyrightable by another entity, I suppose using the ASF header should be fine :thinking: 

Of course, if there is other copyright on the generated code, then the appropriate header should be used.

jbonofre: Let's remove AI, let's talk about code generating code (like OpenAPI): we don't include header here because the code is generated. 
So, if we consider the code as generated by AI, ASF header should not be added.
If the contributor, review and/or modify the code, then yes ASF header is possible.
If the tool (AI or other) set copyright (usually not), then it has to be kept.

flyingImer: Love your discussions ❤️! I think there are two different topics here.

For this PR, I’d prefer to keep the scope narrow and human-facing: the submitting human remains accountable, must understand the change, and AI tools are not contributors.

Questions like reviewer burden, generated-code header treatment, or contributions from AI-controlled GitHub accounts feel real, but broader. My bias is that those belong in a separate discussion rather than in this `CONTRIBUTING.md` update.

So I’d prefer to keep this PR focused on human contributor accountability and spin the broader agent/harness questions out separately.

WDYT?

dimas-b: The header problem exists, I think. So I think we need to resolve it and provide clear guidelines for contributors who want to use AI-generated code. Deferring it will make reviewing PRs with known AI-generated code pretty much impossible.

Re: comparison with OpenAPI code generation, from my POV the key distinction is that build-time tools (like OpenAPI generator) produce code that is not meant to be edited after generation. Any updates to sources require re-generating that code. On the other hand, AI-generated code is meant to be committed to the source repo like any human-written code, so it will be subject to editing and copying by (other) people after the initial contribution. Not having a copyright header will make the status of subsequent manual edits obscure. I believe manual edits should be covered by the normal ASF license, so a header would be valuable in these files... WDYT?

I agree that this discussion is growing in scope, but I think it is beneficial to the project even if it takes time to figure this out.

flyingImer: @dimas-b @jbonofre I agree the header/copyright question is important! My concern is that this feels bigger than a Polaris-local CONTRIBUTING.md rule.

My bias is that Polaris should not try to define project-specific copyright/header treatment for AI-generated code ahead of clearer ASF-level guidance or discussion there.

For this PR, I’d prefer to keep the scope narrow: the submitting human remains accountable, contributors must understand and stand behind the change, and submissions should follow the ASF generative tooling guidance.

If we still feel project-specific header/provenance rules are needed after that, I think that should be handled as a separate follow-up discussion.

I'm inclined to merge as is and bring the discussion to ASF level. WDYT?

dimas-b: From my POV if we encourage contributors to disclose AI contributions and at the same time we're not sure about how those contributions are supposed to be incorporated into the codebase, we're putting reviewers into a logical deadlock.

flyrain: The encouraging disclosure part isn't practical AFAIK. I'd remove them if it causes issues.

jbonofre: The discussion already happens at ASF level.
I'm fine to merge as is, but the question I have is: what are the "core" differences with https://www.apache.org/legal/generative-tooling.html ? More practical ?

dimas-b: I guess individual contributors are more likely to notice the local Polaris guidelines than the ASF-wide document.

From my POV, it might be preferable to emphasize  the ASF doc more (via a link at the top of this section) and then offer Polaris-specific interpretation and points about the human being responsible to understanding the change , etc...

jbonofre: Don't get me wrong: I'm very in favor of this PR. I just want that we bring value for our contributors. So, explaining and illustrating the ASF guidelines for Polaris "local".

dimas-b:
@flyrain : let's merge as discussed (just one last check after the community sync meeting :) )

CONTRIBUTING.md
dimas-b: recap from the Community Sync call (IIRC): re: headers, I believe we're fine with using the usual ASF header on files in PRs that are submitted by a human.

If the whole PR is submitted by an AI agent, let's hold it and have a broader discussion (on `dev`) about how to handle it.

flyrain:
Thanks @flyingImer for driving the discussion. Thanks all for the review.

PR #3956, opened 2026-03-09
"Force" ktfmt version to 0.61 to avoid kotlin "version" issue depending of the kotlin version in gradle

jbonofre:


snazy:
Honestly, this is a noop-change.
`com.diffplug.gradle.spotless.BaseKotlinExtension#ktfmt()` calls `com.diffplug.spotless.kotlin.KtfmtStep#defaultVersion` that returns `com.diffplug.spotless.kotlin.KtfmtStep#DEFAULT_VERSION`, which is a constant for `"0.61"`.

jbonofre:
Good catch. I had a old version zombie in previous tasks. My bad.

PR #3969, opened 2026-03-10
[helm] added httproute config for management endpoints

cccs-cat001:
Adding configs to the Helm charts HTTPRoute to allow for the management endpoints for liveness, readiness and metrics to also be exposed. We've got a use-case where we want to see the readiness/liveness/health/metrics endpoints from outside the cluster, so this is how we can enable that feature. The default is false, so we don't expose that service for a normal deployment. 

## Checklist

### DIFF
--- a/helm/polaris/values.schema.json
+++ b/helm/polaris/values.schema.json
@@ -430,7 +430,7 @@
                     "type": "boolean"
                 },
                 "exposeManagementEndpoints": {
-                    "description": "Enable to expose the health endpoints on your httproute. This is required if you want to use the management service for health checks and metrics scraping external to the cluster.",
+                    "description": "Enable to expose the management endpoints on your httproute. This is required if you want to use the management service for health checks and metrics scraping external to the cluster.",
                     "type": "boolean"
                 },
                 "gatewayName": {
--- a/site/content/in-dev/unreleased/helm-chart/reference.md
+++ b/site/content/in-dev/unreleased/helm-chart/reference.md
@@ -152,7 +152,7 @@ weight: 900
 | httproute.gatewayNamespace | string | `"default"` | Namespace where the Gateway is deployed. Required. |
 | httproute.sectionName | string | `""` | Section name within the gateway to use (optional). |
 | httproute.hosts | list | `["chart-example.local"]` | A list of hostnames that the HTTPRoute should match. |
-| httproute.exposeManagementEndpoints | bool | `false` | Enable to expose the health endpoints on your httproute. This is required if you want to use the management service for health checks and metrics scraping external to the cluster. |
+| httproute.exposeManagementEndpoints | bool | `false` | Enable to expose the management endpoints on your httproute. This is required if you want to use the management service for health checks and metrics scraping external to the cluster. |
 
 ### Resources and Autoscaling
 


cccs-cat001:
Yeah... I'm not seeing why it keeps wanting to put this new config under `Other Values` when I run helm-doc-generate... any ideas?

helm/polaris/templates/httproute.yaml
```diff
@@ -41,8 +41,21 @@ spec:
       sectionName: {{ .Values.httproute.sectionName }}
       {{- end }}
   rules:
-    # We don't specify a matches block here, so the default is a prefix path match on "/" (match every HTTP request)
-    # The backend (Service) to send matching requests to
+    {{- if .Values.httproute.exposeManagementEndpoints }}
+    - matches:
+      - path:
+          type: PathPrefix
+          value: "/mgmt"
```
adutra: Should we make this prefix configurable?

helm/polaris/values.yaml
```diff
@@ -472,6 +472,8 @@ httproute:
   # @section -- HTTPRoute
   hosts:
     - chart-example.local
+  
+  exposeManagementEndpoints: false
```
adutra, line 477: You are missing a description and a section declaration here:

```yaml
  # -- description here
  # @section -- HTTPRoute
  exposeManagementEndpoints: false
```

Then please regenerate the docs and the schema by running `make helm`.

helm/polaris/templates/httproute.yaml
```diff
@@ -41,8 +41,21 @@ spec:
...
+      backendRefs:
+        - name: {{ include "polaris.fullnameWithSuffix" (list . "mgmt") }}
+          port: {{ get (first .Values.managementService.ports) "port" }}
+      filters:
+        - type: URLRewrite
+          urlRewrite:
+            path:
+              type: ReplacePrefixMatch
+              replacePrefixMatch: "/"
```
adutra: Note: this is also exposing metrics under `/q/metrics`, but the PR description mentions only "liveness/readiness/health".

helm/polaris/templates/httproute.yaml
```diff
@@ -41,8 +41,21 @@ spec:
       sectionName: {{ .Values.httproute.sectionName }}
       {{- end }}
   rules:
-    # We don't specify a matches block here, so the default is a prefix path match on "/" (match every HTTP request)
-    # The backend (Service) to send matching requests to
+    {{- if .Values.httproute.exposeManagementEndpoints }}
```
adutra, line 44: Can you please add unit tests for this?

helm/polaris/templates/httproute.yaml
adutra: If we want just healthchecks, maybe something like this is better:

```yaml
    - matches:
      - path:
          type: PathPrefix
          value: "/q/health"
      backendRefs:
        - name: {{ include "polaris.fullnameWithSuffix" (list . "mgmt") }}
          port: {{ get (first .Values.managementService.ports) "port" }}
      filters:
        - type: URLRewrite
          urlRewrite:
            path:
              type: ReplacePrefixMatch
              replacePrefixMatch: "/q/health"
```

helm/polaris/templates/httproute.yaml
adutra: And actually, do we need a new prefix at all?

helm/polaris/values.yaml
cccs-cat001: Ah! That's why the docs are being uncooperative!

helm/polaris/templates/httproute.yaml
cccs-cat001: updated the description to include metrics

adutra: But do we want to expose metrics to the outside world? Do you think there could be a use case for that?

cccs-cat001: from the sounds of our use case they'll want to scrape the metrics too eventually

helm/polaris/templates/httproute.yaml
cccs-cat001: right... I could just direct all `/q` endpoints to that service...

adutra: That's what I was thinking too 😄

adutra: Maybe a safer approach is to just expose a boolean to enable the feature. If it's true then `/q` -> management port, otherwise `/q` requests are dropped:

```yaml
  rules:
    # Always intercept /q — either route to mgmt service or drop
    - matches:
        - path:
            type: PathPrefix
            value: "/q"
      {{- if .Values.httproute.exposeManagementEndpoints }}
      backendRefs:
        - name: {{ include "polaris.fullnameWithSuffix" (list . "mgmt") }}
          port: {{ get (first .Values.managementService.ports) "port" }}
      {{- end }}
    # Catch-all: everything else goes to the main service
    - backendRefs:
        - name: {{ $fullName }}
          port: {{ get (first .Values.service.ports) "port" }}
```

cccs-cat001: I don't think that's really necessary. a /q request going to the regular service results in a 404 anyways

adutra: OK, I thought it was more expressive if we declared the `/q` prefix always, but I won't die on this hill 😅

helm/polaris/values.yaml
```diff
@@ -472,6 +472,9 @@ httproute:
   # @section -- HTTPRoute
   hosts:
     - chart-example.local
+  # -- Enable to expose the health endpoints on your httproute. This is required if you want to use the management service for health checks and metrics scraping external to the cluster.
```
```suggestion
  # -- Enable to expose the management endpoints on your httproute. This is required if you want to use the management service for health checks and metrics scraping external to the cluster.
```

adutra (approved):

Issue #2573, opened 2025-09-15
Provide a mechanism to purge the events table


adutra:
### Is your feature request related to a problem? Please describe.

The `events` table is currently configured for indefinite growth, lacking a purge mechanism. Neither the Polaris server, nor any tool provided by the Apache Polaris project, offers a solution for purging this table.

To improve the experience for open-source users, a more robust strategy for managing the `events` table is needed.

PR #3973, opened 2026-03-11
Enable multiple event listeners to consume for Polaris events

nandorKollar:
This PR enables multiple event listeners watching for Polaris events.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/events/EventBusConfigurer.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/events/EventBusConfigurer.java
@@ -27,7 +27,7 @@ import jakarta.enterprise.event.Observes;
 
 @ApplicationScoped
 public class EventBusConfigurer {
-  private static final String LOCAL_CODEC_NAME = "local";
+  private static final String LOCAL_CODEC_NAME = "polaris-events-local";
 
   void configureEventBus(@Observes StartupEvent ev, EventBus eventBus) {
     eventBus.registerDefaultCodec(PolarisEvent.class, new LocalEventBusCodec<>(LOCAL_CODEC_NAME));


nandorKollar:
@jbonofre @adutra @adnanhemani would you mind share your thoughts about this approach?

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -73,35 +75,46 @@ public AwsCloudWatchEventListener(
     this.clock = clock;
   }
 
-  @PostConstruct
+  @VisibleForTesting
```
adutra: Isn't this change going to break the listener?

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -73,35 +75,46 @@ public AwsCloudWatchEventListener(
...
   void start() {
     this.client = createCloudWatchAsyncClient();
     ensureLogGroupAndStream();
   }
 
+  private CloudWatchLogsAsyncClient client() {
```
adutra: Oh I see – you want lazy init out of fear that this client could be initialized, even if the listener is inactive. Is that right? I don't think you need to worry about that, application-scoped beans are lazily loaded by default.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
```diff
@@ -30,4 +31,15 @@ public interface PolarisEventListenerConfiguration {
    * identifier.
    */
   String type();
+
+  /**
+   * Comma separated list of event listers, each item must be a registered {@link
+   * PolarisEventListener} identifier.
+   */
+  @WithDefault("no-op")
```
adutra: I think the default should be no listener at all. Not every Polaris deployment wants / needs events.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
```diff
@@ -30,4 +31,15 @@ public interface PolarisEventListenerConfiguration {
...
+  String types();
+
+  default String[] enabledPolarisEventListener() {
```
adutra: Nit: Maybe return `Set<String>`.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.polaris.service.events.listeners.PolarisEventListener;
+
+@ApplicationScoped
+public class PolarisEventListeners {
+  @Inject EventBus eventBus;
+  @Inject @Any Instance<PolarisEventListener> eventListeners;
+  @Inject PolarisEventListenerConfiguration configuration;
+
+  public void onStartup(@Observes StartupEvent event) {
+    for (PolarisEventListener listener : eventListeners) {
```
adutra: Ah – this might trigger the initialization of all listener beans. I'm not sure. Maybe we need to change this code and invert the logic: iterate over the active listener names, and select/materialize only the ones we want to activate.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
nandorKollar: Well, the problem is something like that yes. The problem is this: we can't use `@PostConstruct` here. As you spotted here: https://github.com/apache/polaris/pull/3973/changes#r2932831063 it will trigger bean initialisation is triggered in `PolarisEventListeners`, I guess because we use methods of it, so Quarkus lazy bean init kicks in. Therefore, this post construct is triggered even when doesn't use CloudWatch, and a nice fat exception is raised then. I'm thinking about how to avoid touching and triggering every Listener initialisation, but I don't have a breakthrough ide yet. One ide what's in my mind is that we can maybe introduce a custom annotation, which can be used to filter beans, instead of calling `identifier` method. What do you think?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
nandorKollar: How about this: reuse the Identifier annotation (or use our own, Inherited one):
```
            Class<?> listenerClass = listener.getClass();
            if (listener instanceof ClientProxy) {
                listenerClass = listenerClass.getSuperclass();
            }

            if (Arrays.stream(configuration.enabledPolarisEventListener())
                    .toList()
                    .contains(listenerClass.getAnnotation(Identifier.class).value())) {
                Handler<Message<PolarisEvent>> handler = e -> listener.onEvent(e.body());
                eventBus.localConsumer(POLARIS_EVENT_CHANNEL, handler);
            }
```

What do you think?

nandorKollar: Or let's just make it simple by following your approach :) :
```
        for (String enabledEventListener : configuration.enabledPolarisEventListener()) {
            PolarisEventListener listener = eventListeners.select(Identifier.Literal.of(enabledEventListener)).get();
            Handler<Message<PolarisEvent>> handler = e -> listener.onEvent(e.body());
            eventBus.localConsumer(POLARIS_EVENT_CHANNEL, handler);
        }
```

Thanks, I think this will be the right solution.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
nandorKollar: Reverted, with the new approach it isn't needed.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
nandorKollar: Reverted, with the new approach it isn't needed.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
nandorKollar: Latest commit should make both new listeners and old listener config optional, and allow using no listener at all.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
nandorKollar: This was refactored a bit, the logic is now moved to PolarisEventListeners, and is rewritten to use Set there.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -0,0 +1,62 @@
...
+@ApplicationScoped
+public class PolarisEventListeners {
+  @Inject EventBus eventBus;
+  @Inject @Any Instance<PolarisEventListener> eventListeners;
+  @Inject PolarisEventListenerConfiguration configuration;
+
+  public void onStartup(@Observes StartupEvent event) {
+    String listerTypes = configuration.types().orElse("");
+    if (listerTypes.isBlank()) {
+      return;
```
jbonofre: If we don't have types here, it means we never reach type() merging bellow (line 51).

If a user only set `polaris.event-listener.type` (the old config) without setting `types`, the `onStartup()` method just returns directly, and no listener is registered.

It means it's a breaking change.

Imho, the `type()` merge logic should be before the early return, or this check should check both `types` and `type` (to be backward compatible).

nandorKollar: Good point, thanks! This is indeed broken, fixed in 1a0aaa510d94a44f23202ef9b9b7c54c7f0af351

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
```diff
@@ -29,5 +30,11 @@ public interface PolarisEventListenerConfiguration {
    * The type of the event listener to use. Must be a registered {@link PolarisEventListener}
    * identifier.
    */
-  String type();
+  Optional<String> type();
```
adutra, line 37: We don't need this one anymore, right?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
```diff
@@ -29,5 +30,11 @@ public interface PolarisEventListenerConfiguration {
...
+
+  /**
+   * Comma separated list of event listers, each item must be a registered {@link
+   * PolarisEventListener} identifier.
+   */
+  Optional<String> types();
```
adutra:
```suggestion
  Optional<List<String>> types();
```

It's a bit odd to see an `Optional<List<...>` – but that's how SmallRye Config wants it 😅

adutra: Update: `Optional<Set<String>> types();` is probably better.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
nandorKollar: I think we need it to read the old config: `polaris.event-listener.type`. It is optional to allow one to omit. The value of this config (if present), and the new one `polaris.event-listener.types` should be merged. We should also indicate that `polaris.event-listener.type` is deprecated, should not be used.

runtime/defaults/src/main/resources/application.properties
```diff
@@ -157,8 +157,6 @@ quarkus.rds.sync-client.type=apache
 
 polaris.file-io.type=default
 
-polaris.event-listener.type=no-op
```
nandorKollar, line 160: As there's no need to set this property to no-op, should we also delete `NoOpPolarisEventListener`? It will be a breaking change, because anyone who still refers to no-op with the old `polaris.event-listener.type` (or even with the new `polaris.event-listener.types`) will hit the following error: `No bean found for required type [interface org.apache.polaris.service.events.listeners.PolarisEventListener] and qualifiers [[@jakarta.enterprise.inject.Any(), @io.smallrye.common.annotation.Identifier(value="no-op")]]`

We can mark it deprecated though.

nandorKollar:
> @nandorKollar this is look promising thanks!
> 
> I think we should avoid calling `dispatch()` altogether (and avoid materializing the `PolarisEvent` instance on the heap) if we know the event type is not needed. I would like us to consider that, if a user deactivates all listeners, they would get a performance comparable to what they would get if there were no events in Polaris at all.

Thanks, this makes sense, however I've further questions related to this. Do you happen to know why does [AwsCloudWatchEventListener](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java#L149) filter only for a single event type? I think event listeners should accept any event type, and allow users to configure a particular event listener to listen for only a certain events, but the listener code itself should not assume that it is working only with event type xyz.

adutra:
> Do you happen to know why does [AwsCloudWatchEventListener](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java#L149) filter only for a single event type? I think event listeners should accept any event type, and allow users to configure a particular event listener to listen for only a certain events, but the listener code itself should not assume that it is working only with event type xyz.

Completely agree here. Listeners should NOT filter events, and especially not in a hard-coded fashion – that's a user decision. @adnanhemani may know why the CoudWatch listener was designed that way. Imho it's not very useful – it can only be useful if you are interested in after-refresh-table events, and hence, in certain table optimization use cases.

runtime/defaults/src/main/resources/application.properties
adutra: Yes, I think we could delete the no-op listener. Regarding breaking changes, our evolution guidelines accept that: https://polaris.apache.org/releases/1.3.0/evolution/

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
adutra: Hmm indeed our evolution guidelines considers changes to the configuration as breaking changes: https://polaris.apache.org/releases/1.3.0/evolution/

So, I think we indeed need to deprecate this config but keep it around for some time.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.polaris.service.events.listeners.PolarisEventListener;
+
+@ApplicationScoped
+public class PolarisEventListeners {
+  @Inject EventBus eventBus;
+  @Inject @Any Instance<PolarisEventListener> eventListeners;
+  @Inject PolarisEventListenerConfiguration configuration;
+
+  public void onStartup(@Observes StartupEvent event) {
+    Set<String> listenerTypeSet = configuration.types().orElseGet(HashSet::new);
```
adutra, line 45: I'd be careful here, `configuration.types()` will likely return an _immutable_ set, if present. This is probably safer:

```java
    Set<String> listenerTypeSet =
        Stream.concat(configuration.types().stream(), configuration.type().stream())
            .collect(Collectors.toSet());
```

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
nandorKollar: How can we mark it deprecated?

nandorKollar:
> > Do you happen to know why does [AwsCloudWatchEventListener](https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java#L149) filter only for a single event type? I think event listeners should accept any event type, and allow users to configure a particular event listener to listen for only a certain events, but the listener code itself should not assume that it is working only with event type xyz.
> 
> Completely agree here. Listeners should NOT filter events, and especially not in a hard-coded fashion – that's a user decision. @adnanhemani may know why the CoudWatch listener was designed that way. Imho it's not very useful – it can only be useful if you are interested in after-refresh-table events, and hence, in certain table optimization use cases.

@adutra How about introducing the event listener configuration in a follow-up PR, together with the optimization you mentioned earlier (i.e., avoiding materializing the PolarisEvent)?

In that PR, we could address the following points:
- Make each event listener configurable. For example, CloudWatch could be configured to receive only a subset of events, while at the same time every event is also persisted in the database.
- Ensure the event dispatcher only emits events when there is a registered listener for the given event type. This was partially implemented in #3442 — we could revive that work.
- Avoid assumptions in the event listener implementation about handling only a specific event type.

What do you think?

nandorKollar:
Oh and we shouldn't forget #2573 either. We need to find a way to trigger event cleanup, or even set retention period for events persisted in the database.

adutra:
> @adutra How about introducing the event listener configuration in a follow-up PR, together with the optimization you mentioned earlier (i.e., avoiding materializing the PolarisEvent)?
> 
> In that PR, we could address the following points:
> 
> * Make each event listener configurable. For example, CloudWatch could be configured to receive only a subset of events, while at the same time every event is also persisted in the database.
> * Ensure the event dispatcher only emits events when there is a registered listener for the given event type. This was partially implemented in [Create PolarisEvent only when event type is actually supported by listener #3442](https://github.com/apache/polaris/pull/3442) — we could revive that work.
> * Avoid assumptions in the event listener implementation about handling only a specific event type.
> 
> What do you think?

I'm all in for breaking the work in many smaller PRs, but I'm a little confused: you said you would like to introduce the event listener configuration in a follow-up PR – but you also wrote you would like to make each event listener configurable in this PR. Could you clarify this small item?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
adutra: You should:

1. Annotate the method with `@Deprecated`
2. Add a note to the changelog.

nandorKollar:
> > @adutra How about introducing the event listener configuration in a follow-up PR, together with the optimization you mentioned earlier (i.e., avoiding materializing the PolarisEvent)?
> > In that PR, we could address the following points:
> > 
> > * Make each event listener configurable. For example, CloudWatch could be configured to receive only a subset of events, while at the same time every event is also persisted in the database.
> > * Ensure the event dispatcher only emits events when there is a registered listener for the given event type. This was partially implemented in [Create PolarisEvent only when event type is actually supported by listener #3442](https://github.com/apache/polaris/pull/3442) — we could revive that work.
> > * Avoid assumptions in the event listener implementation about handling only a specific event type.
> > 
> > What do you think?
> 
> I'm all in for breaking the work in many smaller PRs, but I'm a little confused: you said you would like to introduce the event listener configuration in a follow-up PR – but you also wrote you would like to make each event listener configurable in this PR. Could you clarify this small item?

I recommend to target only focus on making multiple event listeners work in this PR, and rather implement the rest in a second PR. The only configuration parameter for now is deprecating `polaris.event-listener.type` and adding `polaris.event-listener.types`. Of course if you think we should avoid event materialization in this PR, I can continue working on that in this change too.

adutra:
> I recommend to target only focus on making multiple event listeners work in this PR, and rather implement the rest in a second PR. The only configuration parameter for now is deprecating `polaris.event-listener.type` and adding `polaris.event-listener.types`. Of course if you think we should avoid event materialization in this PR, I can continue working on that in this change too.

That works for me. I think this PR is good to go then, but it would still be nice to add a note to the changelog about the deprecation of `polaris.event-listener.type`.

nandorKollar:
> > I recommend to target only focus on making multiple event listeners work in this PR, and rather implement the rest in a second PR. The only configuration parameter for now is deprecating `polaris.event-listener.type` and adding `polaris.event-listener.types`. Of course if you think we should avoid event materialization in this PR, I can continue working on that in this change too.
> 
> That works for me. I think this PR is good to go then, but it would still be nice to add a note to the changelog about the deprecation of `polaris.event-listener.type`.

Should we add it only to the changelog?

_EDIT_: never mind, found the way to deprecate it in the doc too. :)

runtime/service/src/main/java/org/apache/polaris/service/events/EventBusConfigurer.java
```diff
@@ -16,13 +16,20 @@
...
-@QuarkusIntegrationTest
-public class CockroachRestCatalogIT extends PolarisRestCatalogFileIntegrationTest {}
+import io.quarkus.runtime.StartupEvent;
+import io.quarkus.vertx.LocalEventBusCodec;
+import io.vertx.core.eventbus.EventBus;
+import jakarta.enterprise.event.Observes;
+
+public class EventBusConfigurer {
+  private static final String LOCAL_CODEC_NAME = "local";
+  private static final LocalEventBusCodec<PolarisEvent> LOCAL_CODEC =
```
dimas-b: nit: is this constant really required?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
dimas-b: We could probably add a config redirect too.

Cf. https://github.com/apache/polaris/blob/087209314acb4abb10f3ca3fe178abac38eebb3f/persistence/nosql/persistence/cdi/quarkus/src/main/java/org/apache/polaris/persistence/nosql/quarkus/backend/MongoDBConfigSourceFactory.java#L45

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisServiceBusEventDispatcher.java
```diff
@@ -16,13 +16,25 @@
...
+import jakarta.inject.Inject;
+
+@ApplicationScoped
+@DefaultBean
+public class PolarisServiceBusEventDispatcher implements PolarisEventDispatcher {
+  public static final String POLARIS_EVENT_CHANNEL = "polaris-events";
+
+  @Inject EventBus eventBus;
+
+  //  @Inject PolarisEventListeners polarisEventListeners;
```
dimas-b: nit: leftover?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -326,7 +332,8 @@ public void before(TestInfo testInfo) {
         .thenReturn((PolarisStorageIntegration) storageIntegration);
 
     this.catalog = initCatalog("my-catalog", ImmutableMap.of());
-    testPolarisEventListener = (TestPolarisEventListener) polarisEventListener;
+    testPolarisEventListener =
+        (TestPolarisEventListener) polarisEventListener.select(Identifier.Literal.of("test")).get();
```
dimas-b: Would it work if we had `@Identifier` on the `polarisEventListener` field without `Intance<>`?

runtime/service/src/testFixtures/java/org/apache/polaris/service/events/listeners/TestPolarisEventDispatcher.java
```diff
@@ -16,26 +16,32 @@
...
 
-public class CockroachRelationalJdbcProfile implements QuarkusTestProfile {
-  @Override
-  public Map<String, String> getConfigOverrides() {
-    return Map.of(
-        "polaris.persistence.auto-bootstrap-types",
-        "relational-jdbc",
-        "polaris.persistence.relational.jdbc.database-type",
-        "cockroachdb");
+public class TestPolarisEventDispatcher implements PolarisEventDispatcher {
```
dimas-b: nit: the `Test` prefix makes this class look like a JUnit test class, which it is not. How about `InMemoryEventCollector`?

nandorKollar: `InMemoryEventCollector` suggests that it can be used outside of tests too no? This class is indented to be used only in tests though.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
nandorKollar: Yes, looks like it works! Thanks, great recommendation!

runtime/service/src/testFixtures/java/org/apache/polaris/service/events/listeners/TestPolarisEventDispatcher.java
dimas-b: The name describes what the class does :shrug: Where it can be used is controlled by the location of the class in the sources tree (ATM under `/test`).

If we (hypothetically) wanted to expose this functionality to mainstream code, we'd only have to move the class without renaming it.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
dimas-b: WDYT about config redirects? My code pointer above may not be an exact match for this use case, but I hope it can be a starting point 🙂

nandorKollar: Well, I’m trying to understand how we can use that. What would be the benefit of a config redirect? The old property is already mapped to the new one, I think. Would ConfigSourceFactory make this mapping cleaner, or is there a specific scenario that wouldn’t work with the current approach?

dimas-b: Where is the old config mapped to the new one (I might have missed it)?

What I mean is that ideally we should treat the old `polaris.event-listener.type` values as a single list entry for the new `polaris.event-listener.types` config (if the new one is not set explicitly). The `no-op` old type value can be ignored (mapped to empty list).

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisServiceBusEventDispatcher.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.polaris.service.events;
+
+import io.quarkus.arc.DefaultBean;
+import io.vertx.core.eventbus.EventBus;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+@ApplicationScoped
+@DefaultBean
+public class PolarisServiceBusEventDispatcher implements PolarisEventDispatcher {
```
flyrain, line 30: Can we add a Java doc for this class?

CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 - The configuration option `polaris.rate-limiter.token-bucket.window` is no longer supported and should be removed.
 - `PolarisConfigurationStore` has been deprecated for removal.
+- The configuration option `polaris.event-listener.type` is no longer supported and should be removed. Please use `polaris.event-listener.types` instead.
```
flyrain: ~~It mighty be OK to make this breaking change, given event listener are not widely used ATM. Could we share it in the mailing list to gather more feedback to double check? cc @adnanhemani .~~

~~If it turns out to be fine, we should call it out here that it's a breaking change.~~

runtime/service/src/main/java/org/apache/polaris/service/events/EventBusConfigurer.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.polaris.service.events;
+
+import io.quarkus.runtime.StartupEvent;
+import io.quarkus.vertx.LocalEventBusCodec;
+import io.vertx.core.eventbus.EventBus;
+import jakarta.enterprise.event.Observes;
+
+public class EventBusConfigurer {
```
flyrain, line 29: Do we need `@ApplicationScoped` for this class?

CHANGELOG.md
flyrain: NVM. It seems to be deprecated. I think the word here isn't accurate. 
```suggestion
- The configuration option `polaris.event-listener.type` is deprecated and will be removed later. Please use `polaris.event-listener.types` instead.
```

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
adutra: We can also leverage a ConfigSourceInterceptor for that. Example from another project:

https://github.com/dremio/iceberg-auth-manager/blob/main/oauth2/core/src/main/java/com/dremio/iceberg/authmgr/oauth2/config/ConfigRelocationInterceptor.java

nandorKollar: > Where is the old config mapped to the new one (I might have missed it)?
> 
> What I mean is that ideally we should treat the old `polaris.event-listener.type` values as a single list entry for the new `polaris.event-listener.types` config (if the new one is not set explicitly). The `no-op` old type value can be ignored (mapped to empty list).

No, you didn't miss it, indeed the old config value isn't explicitly mapped to the new one. The old property is transformed (merged) with the new one at the place where it is used.

nandorKollar: > Where is the old config mapped to the new one (I might have missed it)?
> 
> What I mean is that ideally we should treat the old `polaris.event-listener.type` values as a single list entry for the new `polaris.event-listener.types` config (if the new one is not set explicitly). The `no-op` old type value can be ignored (mapped to empty list).

Exactly, this is what I'm considering to use too.

nandorKollar: Although `RelocateConfigSourceInterceptor` sounds really promising, I'm afraid it won't work for us. It relocates the property, but we need to merge the values: if both old and new property is set, the values should be merged. We can alternatively say, that the new property overrides the old one, if both are set:
_Option 1:_
```
polaris.event-listener.type=aws-cloudwatch
polaris.event-listener.types=persistence-in-memory-buffer

```
result is: `persistence-in-memory-buffer, aws-cloudwatch`

_Option 2:_
```
polaris.event-listener.type=aws-cloudwatch
polaris.event-listener.types=persistence-in-memory-buffer

```
result is `persistence-in-memory-buffer`, because the new property overrides the old one

The second approach is super simple with `FallbackConfigSourceInterceptor`, the first one needs a bit more work. Due to simplicity I'd vote for option 2, if both are set (why would someone set both btw? maybe we can emit a warning if both are set btw, or when the old property is still used) What do you think?

jbonofre:
nit: I updated the PR title (typo on lister instead of listener 😄 ).

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
dimas-b: Option 2 LGTM. My rationale: if the admin user sets the new property, we can assume the user is aware of the old config deprecation and took care to set the new value correctly.

dimas-b: Re: warning - IMHO it's not worth the effort (same rationale applies), but if it's simple to implement, why not?

runtime/service/src/main/java/org/apache/polaris/service/events/EventBusConfigurer.java
dimas-b: IDK off the top of my head 😅 ... but it does not hurt, I'm pretty sure.

dimas-b: `ServiceProducers` have the `maybeBootstrap()` observer without `@ApplicationScoped`, so I'd guess it is NOT required here.

dimas-b: If we drop `@ApplicationScoped` it might be worth trying to make the observer method `static`.

dimas-b: This is not a blocker from my POV

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
dimas-b: It looks like Option 2 got implemented already :+1: 

The warning is optional from my POV.

nandorKollar:
Great, thanks @dimas-b , @flyrain would you mind also have a look at the latest version? @jbonofre @adutra is there any unresolved comment on your end that I may have forgotten to address?

adnanhemani:
I will look into this later today, please hold merging until then.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
adnanhemani: I'm also ok with Option 2. Given that `type` will be deprecated with this PR, it doesn't make much sense to  keep respecting it if `types` is already defined.

runtime/service/src/main/java/org/apache/polaris/service/events/EventBusConfigurer.java
```diff
@@ -0,0 +1,35 @@
...
+
+import io.quarkus.runtime.StartupEvent;
+import io.quarkus.vertx.LocalEventBusCodec;
+import io.vertx.core.eventbus.EventBus;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
+
+@ApplicationScoped
+public class EventBusConfigurer {
+  private static final String LOCAL_CODEC_NAME = "local";
```
adnanhemani: I'm not an expert but after searching around, it seems like this value is supposed to be unique per codec. So shouldn't we name this something like "polaris-events-local" or something to ensure there is no potential future clash naming-wise?

polaris-core/src/main/java/org/apache/polaris/core/connection/OAuthClientCredentialsParametersDpo.java
```diff
@@ -43,9 +42,6 @@
  * API model.
  */
 public class OAuthClientCredentialsParametersDpo extends AuthenticationParametersDpo {
-
-  private static final Joiner COLON_JOINER = Joiner.on(":");
```
adnanhemani, line 47: spurrious?

extensions/auth/opa/tests/src/intTest/java/org/apache/polaris/extension/auth/opa/test/OpaTestProfiles.java
```diff
@@ -19,7 +19,6 @@
 package org.apache.polaris.extension.auth.opa.test;
 
 import io.quarkus.test.junit.QuarkusTestProfile;
-import io.quarkus.test.junit.QuarkusTestProfile.TestResourceEntry;
```
adnanhemani, line 22: spurrious?

polaris-core/src/main/java/org/apache/polaris/core/connection/OAuthClientCredentialsParametersDpo.java
nandorKollar: This was an unused constant. Not related to the event listeners, but could be deleted in this PR I think.

extensions/auth/opa/tests/src/intTest/java/org/apache/polaris/extension/auth/opa/test/OpaTestProfiles.java
nandorKollar: Unused import, I think it is safe to do this cleanup here.

nandorKollar:
> Looks almost there, just a few questions on my end - and then one comment from @flyrain about adding a Javadoc. Will approve it after those are done!

@adnanhemani which Javadoc do you refer to? I already added Javadoc to `PolarisServiceBusEventDispatcher`.

polaris-core/src/main/java/org/apache/polaris/core/connection/OAuthClientCredentialsParametersDpo.java
adnanhemani: Personally would prefer not to do spurrious cleanups in files that would be otherwise untouched.

nandorKollar:
Thank you all for the help and reviews!

PR #3991, opened 2026-03-12
Remove unused PolarisAuthorizableOperation values

dimas-b:
* `REVOKE_PRINCIPAL_GRANT_FROM_PRINCIPAL_ROLE`
* `REVOKE_PRINCIPAL_ROLE_GRANT_FROM_PRINCIPAL_ROLE`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Changed from Poetry to UV for Python package management.
 - Exclude KMS policies when KMS is not being used for S3.
 - Improved default KMS permission handling to better distinguish read-only and read-write access.
+- Removed unused `PolarisAuthorizableOperation` values: `REVOKE_PRINCIPAL_GRANT_FROM_PRINCIPAL_ROLE`, `REVOKE_PRINCIPAL_ROLE_GRANT_FROM_PRINCIPAL_ROLE`.
 
 ### Deprecations
 


flyrain:
cc @collado-mike @dennishuo

nandorKollar:
Just a quick question: why do we remove only these two operations? Among others, `ADD_PRINCIPAL_ROLE_GRANT_TO_PRINCIPAL_ROLE` and `ADD_PRINCIPAL_GRANT_TO_PRINCIPAL_ROLE` seems also unused.

flyingImer (commented):
I think there is one compatibility concern worth fixing before this moves forward.

IIUC, the key question here is not whether Polaris core still emits these operations, but whether `PolarisAuthorizableOperation` is part of the current public auth vocabulary. From `AuthorizationRequest`, `PolarisAuthorizer`, and the OPA adapter, it seems the answer is yes.

If that is the boundary, then removing enum members feels more like narrowing a public SPI than cleaning up unused internal code. External authorizer implementations may still reference these values even if core no longer produces them itself.

My bias would be to preserve that vocabulary unless we want to make an explicit breaking-change decision. Would it make sense to keep these as deprecated compatibility values for now, and handle actual removal through a clearer deprecation path later?

2 cents: that seems like the safer current-scope move, while still leaving room to evolve the auth SPI more intentionally over time.

dimas-b:
> Just a quick question: why do we remove only these two operations [...]

Just what I personally noticed :sweat_smile: Please feel free to open similar cleanup PR for removing other unused code.

dimas-b:
> the key question here is not whether Polaris core still emits these operations, but whether PolarisAuthorizableOperation is part of the current public auth vocabulary [...]

Current [Polaris Evolution](https://polaris.apache.org/in-dev/unreleased/evolution/) guidelines do not consider java code (whether public or not) to be part of the API surface in the SemVer sense.

Additionally, since these values were not used in Polaris code, any softly-linked user-driven configuration that may reference them can probably be expected to remain operational (albeit pointless).

dimas-b:
Related `dev` discussion: https://lists.apache.org/thread/gocs3lbl4w6x3tpso0wjos6sr8ctffl8

flyrain:
Can we update it in the change log?

dimas-b:
Rebased. Added an entry to CHANGELOG.

CHANGELOG.md
```diff
@@ -72,6 +72,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Changed from Poetry to UV for Python package management.
 - Exclude KMS policies when KMS is not being used for S3.
 - Improved default KMS permission handling to better distinguish read-only and read-write access.
+- Removed unused `PolarisAuthorizableOperation` values: `REVOKE_PRINCIPAL_GRANT_FROM_PRINCIPAL_ROLE`, `REVOKE_PRINCIPAL_ROLE_GRANT_FROM_PRINCIPAL_ROLE`.
```
nandorKollar, line 75: Should we maybe add these to the list of breaking changes?

jbonofre: I think it's fine here. As it was not used anymore, not sure we can consider as a strict breaking change.

flyrain (approved):

CHANGELOG.md
dimas-b: I do not see any potential for breakage - commented about that in the main comment thread above and on `dev`: https://lists.apache.org/thread/gocs3lbl4w6x3tpso0wjos6sr8ctffl8

dimas-b:
Consensus appears to exist based on GH comments and `dev` discussions:
* https://lists.apache.org/thread/gocs3lbl4w6x3tpso0wjos6sr8ctffl8
* https://lists.apache.org/thread/1tp3zsotv07058gh4n63v8dwr701y0fo

PR #3996, opened 2026-03-13
feat(gcp): Add Hierarchical Namespace (HNS) support for GCS bucket 

varunarya002:
## Problem

GCS buckets with Hierarchical Namespace (HNS) enabled treat folders as first-class resources, not name prefixes. Iceberg table creation and Spark ingestion fail on HNS buckets because:

1. Folder objects must exist before files can be written into them.
2. Folder creation requires `storage.folders.create` — not granted by the standard write roles in our vended credentials.

## Approach

Two complementary changes, with a clear split between server-side and client-side responsibility:

### 1. Server-side: pre-createTable hook to materialize the folder hierarchy

A new SPI, `prepareLocations(List<String>)`, is exposed on `PolarisStorageIntegration` with a no-op default. GCS overrides it.

`IcebergCatalogHandler.createTableDirect()` and `createTableStaged()` now invoke `integration.prepareLocations([tableLocation, metadataLocation, dataLocation])` **before** delegating to Iceberg's `Catalog.createTable()`.

The GCS implementation (`GcsStorageLocationPreparer`):
- Calls `Storage.get(bucket, HIERARCHICAL_NAMESPACE)` to detect HNS per bucket (auto-detection — HNS is bucket-immutable, so it's safe to call once per request; caching is a follow-up).
- For HNS buckets only, walks the path hierarchy (`warehouse` → `warehouse/ns1` → `warehouse/ns1/table1`) and calls `StorageControlClient.createFolder()` for each segment, idempotent against `AlreadyExistsException`.
- For non-HNS buckets, no-op.

Generic URI parsing and hierarchy-building lives in `HierarchicalFolderLocationPreparer` (base class) so Azure ADLS Gen2 can reuse it later.

### 2. Client-side: vended-credential scope adds folders.create + folders.get for HNS buckets

`GcpCredentialsStorageIntegration.generateAccessBoundaryRules()` now takes a `Predicate<String> isHnsBucket`. For each write bucket where the predicate is true, it adds a third access-boundary rule containing exactly two permissions: `storage.folders.create` and `storage.folders.get`, scoped via `resource.name.startsWith('projects/_/buckets/<bucket>/folders/<path>')`.

This allows Spark to create deeper partition folders (e.g. for `year=2024/month=01/`) at write time without needing admin credentials, while non-HNS catalogs are unaffected (no additional permissions vended).

The narrow permission set (vs. the predefined `roles/storage.folderAdmin` role, which includes `setIamPolicy/getIamPolicy/delete/rename/list`) preserves least-privilege.

## Tests
- HNS folder rule emission gated by predicate
- HNS with multiple buckets and partial writes
- HNS without writes (no folder rules)
- HNS with separate metadata and data buckets
- HNS predicate-gating (mixed HNS/non-HNS in one catalog)
- Folder rule uses narrow `folders.create` + `folders.get` permissions (verified explicitly)
- Non-HNS bucket emits no folder rule
- `GcsStorageLocationPreparer` HNS detection and folder creation (full suite preserved from previous round)

## Checklist
- [x] 🧪 Added/updated tests with good coverage
- [x] 🧾 Updated \`CHANGELOG.md\`
- [x] 📚 Updated documentation in \`site/content/in-dev/unreleased\`

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
@@ -55,4 +55,14 @@ public interface PolarisStorageIntegration {
       @NonNull List<LocationGrant> grants,
       @NonNull Optional<String> refreshEndpoint,
       @NonNull CredentialVendingContext context);
+
+  /**
+   * Hook invoked before a table is created to pre-materialize storage locations that the underlying
+   * storage system requires to exist before files can be written into them.
+   *
+   * <p>The default is a no-op. Storage types that need explicit folder creation (e.g. GCS with
+   * Hierarchical Namespace enabled) override this. Passed locations include the table location and
+   * its metadata/data subpaths.
+   */
+  default void prepareLocations(@NonNull List<String> locations) {}
 }
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
@@ -28,6 +28,10 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.iam.credentials.v1.GenerateAccessTokenRequest;
 import com.google.cloud.iam.credentials.v1.GenerateAccessTokenResponse;
 import com.google.cloud.iam.credentials.v1.IamCredentialsClient;
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
@@ -43,6 +47,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.polaris.core.StructuredLogKeys;
@@ -55,6 +60,7 @@ import org.apache.polaris.core.storage.PolarisStorageActions;
 import org.apache.polaris.core.storage.PolarisStorageIntegration;
 import org.apache.polaris.core.storage.StorageAccessConfig;
 import org.apache.polaris.core.storage.StorageAccessProperty;
+import org.apache.polaris.core.storage.StorageLocationPreparer;
 import org.apache.polaris.core.storage.StorageUri;
 import org.apache.polaris.core.storage.cache.StorageCredentialCache;
 import org.apache.polaris.core.storage.cache.StorageCredentialCacheKey;
@@ -83,6 +89,7 @@ public class GcpCredentialsStorageIntegration
   private final HttpTransportFactory transportFactory;
   private final GcpCredentialOps credentialOps;
   private final Optional<GcpAttributionParams> attributionParams;
+  private final StorageLocationPreparer folderPreparer;
 
   public GcpCredentialsStorageIntegration(
       GoogleCredentials sourceCredentials,
@@ -96,7 +103,8 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         GcpCredentialOps.DEFAULT,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
   }
 
   public GcpCredentialsStorageIntegration(
@@ -112,7 +120,8 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         GcpCredentialOps.DEFAULT,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
   }
 
   public GcpCredentialsStorageIntegration(
@@ -128,7 +137,8 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         credentialOps,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
   }
 
   public GcpCredentialsStorageIntegration(
@@ -145,7 +155,45 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         credentialOps,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
+  }
+
+  public GcpCredentialsStorageIntegration(
+      GoogleCredentials sourceCredentials,
+      HttpTransportFactory transportFactory,
+      StorageCredentialCache cache,
+      GcpStorageConfigurationInfo storageConfig,
+      RealmConfig realmConfig,
+      @NonNull StorageLocationPreparer folderPreparer) {
+    this(
+        sourceCredentials,
+        transportFactory,
+        cache,
+        storageConfig,
+        realmConfig,
+        GcpCredentialOps.DEFAULT,
+        resolveAttributionParams(realmConfig),
+        folderPreparer);
+  }
+
+  public GcpCredentialsStorageIntegration(
+      GoogleCredentials sourceCredentials,
+      HttpTransportFactory transportFactory,
+      StorageCredentialCache cache,
+      GcpStorageConfigurationInfo storageConfig,
+      RealmConfig realmConfig,
+      GcpCredentialOps credentialOps,
+      @NonNull StorageLocationPreparer folderPreparer) {
+    this(
+        sourceCredentials,
+        transportFactory,
+        cache,
+        storageConfig,
+        realmConfig,
+        credentialOps,
+        resolveAttributionParams(realmConfig),
+        folderPreparer);
   }
 
   /**
@@ -161,6 +209,26 @@ public class GcpCredentialsStorageIntegration
       RealmConfig realmConfig,
       GcpCredentialOps credentialOps,
       Optional<GcpAttributionParams> attributionParams) {
+    this(
+        sourceCredentials,
+        transportFactory,
+        cache,
+        storageConfig,
+        realmConfig,
+        credentialOps,
+        attributionParams,
+        locations -> {});
+  }
+
+  public GcpCredentialsStorageIntegration(
+      GoogleCredentials sourceCredentials,
+      HttpTransportFactory transportFactory,
+      StorageCredentialCache cache,
+      GcpStorageConfigurationInfo storageConfig,
+      RealmConfig realmConfig,
+      GcpCredentialOps credentialOps,
+      Optional<GcpAttributionParams> attributionParams,
+      @NonNull StorageLocationPreparer folderPreparer) {
     super(cache, realmConfig, storageConfig);
     // Needed for when environment variable GOOGLE_APPLICATION_CREDENTIALS points to google service
     // account key json
@@ -169,6 +237,7 @@ public class GcpCredentialsStorageIntegration
     this.transportFactory = transportFactory;
     this.credentialOps = credentialOps;
     this.attributionParams = attributionParams;
+    this.folderPreparer = Objects.requireNonNull(folderPreparer, "folderPreparer");
   }
 
   /**
@@ -209,6 +278,11 @@ public class GcpCredentialsStorageIntegration
             realmConfig.getConfig(FeatureConfiguration.GCS_PRINCIPAL_ATTRIBUTION_WIF_AUDIENCE),
             realmConfig.getConfig(FeatureConfiguration.GCS_PRINCIPAL_ATTRIBUTION_SIGNING_KEY_FILE),
             realmConfig.getConfig(FeatureConfiguration.GCS_PRINCIPAL_ATTRIBUTION_SIGNING_KEY_ID)));
+  }
+
+  @Override
+  public void prepareLocations(@NonNull List<String> locations) {
+    folderPreparer.prepareLocations(locations);
   }
 
   @Override
@@ -312,7 +386,11 @@ public class GcpCredentialsStorageIntegration
         resolveSourceCredentials(key, gcpStorageConfig, sourceCredentials, credentialOps);
 
     CredentialAccessBoundary accessBoundary =
-        generateAccessBoundaryRules(readLocations, listLocations, writeLocations);
+        generateAccessBoundaryRules(
+            readLocations,
+            listLocations,
+            writeLocations,
+            bucket -> isHnsBucket(bucket, credentialsToDownscope));
     DownscopedCredentials credentials =
         DownscopedCredentials.newBuilder()
             .setHttpTransportFactory(transportFactory)
@@ -387,7 +465,7 @@ public class GcpCredentialsStorageIntegration
    * Returns the credential to be used as the source for downscoping. If a specific service account
    * is configured, it impersonates that account first.
    */
-  private static GoogleCredentials getBaseCredentials(
+  public static GoogleCredentials getBaseCredentials(
       GcpStorageConfigurationInfo storageConfig,
       GoogleCredentials sourceCredentials,
       GcpCredentialOps credentialOps) {
@@ -451,8 +529,32 @@ public class GcpCredentialsStorageIntegration
       @NonNull Set<String> allowedReadLocations,
       @NonNull Set<String> allowedListLocations,
       @NonNull Set<String> allowedWriteLocations) {
+    return generateAccessBoundaryRules(
+        allowedReadLocations, allowedListLocations, allowedWriteLocations, bucket -> false);
+  }
+
+  @VisibleForTesting
+  public static CredentialAccessBoundary generateAccessBoundaryRules(
+      boolean allowListOperation,
+      @NonNull Set<String> allowedReadLocations,
+      @NonNull Set<String> allowedWriteLocations,
+      @NonNull Predicate<String> isHnsBucket) {
+    return generateAccessBoundaryRules(
+        allowedReadLocations,
+        allowListOperation ? allowedReadLocations : Set.of(),
+        allowedWriteLocations,
+        isHnsBucket);
+  }
+
+  @VisibleForTesting
+  public static CredentialAccessBoundary generateAccessBoundaryRules(
+      @NonNull Set<String> allowedReadLocations,
+      @NonNull Set<String> allowedListLocations,
+      @NonNull Set<String> allowedWriteLocations,
+      @NonNull Predicate<String> isHnsBucket) {
     Map<String, LinkedHashSet<String>> readConditionsByBucket = new LinkedHashMap<>();
     Map<String, LinkedHashSet<String>> writeConditionsByBucket = new LinkedHashMap<>();
+    Map<String, LinkedHashSet<String>> folderConditionsByBucket = new LinkedHashMap<>();
     HashSet<String> bucketsWithList = new HashSet<>();
 
     Stream.concat(allowedReadLocations.stream(), allowedListLocations.stream())
@@ -490,6 +592,9 @@ public class GcpCredentialsStorageIntegration
           writeConditionsByBucket
               .computeIfAbsent(bucket, key -> new LinkedHashSet<>())
               .add(resourceNameStartsWithExpression(bucket, path));
+          folderConditionsByBucket
+              .computeIfAbsent(bucket, key -> new LinkedHashSet<>())
+              .add(folderNameStartsWithExpression(bucket, path));
         });
 
     CredentialAccessBoundary.Builder accessBoundaryBuilder = CredentialAccessBoundary.newBuilder();
@@ -526,13 +631,61 @@ public class GcpCredentialsStorageIntegration
           builder.setAvailablePermissions(List.of("inRole:roles/storage.legacyBucketWriter"));
           accessBoundaryBuilder.addRule(builder.build());
         });
+    folderConditionsByBucket.forEach(
+        (bucket, conditions) -> {
+          if (conditions.isEmpty() || !isHnsBucket.test(bucket)) {
+            return;
+          }
+          CredentialAccessBoundary.AccessBoundaryRule.Builder builder =
+              CredentialAccessBoundary.AccessBoundaryRule.newBuilder();
+          builder.setAvailableResource(bucketResource(bucket));
+          builder.setAvailabilityCondition(
+              CredentialAccessBoundary.AccessBoundaryRule.AvailabilityCondition.newBuilder()
+                  .setExpression(String.join(" || ", conditions))
+                  .build());
+          builder.setAvailablePermissions(List.of("inRole:roles/storage.folderAdmin"));
+          accessBoundaryBuilder.addRule(builder.build());
+        });
     return accessBoundaryBuilder.build();
+  }
+
+  @VisibleForTesting
+  static boolean isHnsBucket(String bucket, GoogleCredentials sourceCredentials) {
+    try {
+      Storage storage = newStorageClient(sourceCredentials);
+      Bucket bucketMetadata =
+          storage.get(
+              bucket, Storage.BucketGetOption.fields(Storage.BucketField.HIERARCHICAL_NAMESPACE));
+      if (bucketMetadata == null) {
+        return false;
+      }
+      BucketInfo.HierarchicalNamespace hns = bucketMetadata.getHierarchicalNamespace();
+      return hns != null && Boolean.TRUE.equals(hns.getEnabled());
+    } catch (RuntimeException e) {
+      LOGGER
+          .atWarn()
+          .addKeyValue("bucket", bucket)
+          .log("Failed to determine HNS status; defaulting to non-HNS", e);
+      return false;
+    }
+  }
+
+  @VisibleForTesting
+  static Storage newStorageClient(GoogleCredentials sourceCredentials) {
+    return StorageOptions.newBuilder().setCredentials(sourceCredentials).build().getService();
   }
 
   @VisibleForTesting
   static String resourceNameStartsWithExpression(String bucket, String path) {
     return String.format(
         "resource.name.startsWith('projects/_/buckets/%s/objects/%s')",
+        escapeCelLiteral(bucket), escapeCelLiteral(path));
+  }
+
+  @VisibleForTesting
+  static String folderNameStartsWithExpression(String bucket, String path) {
+    return String.format(
+        "resource.name.startsWith('projects/_/buckets/%s/folders/%s')",
         escapeCelLiteral(bucket), escapeCelLiteral(path));
   }
 
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
@@ -28,6 +28,10 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.iam.credentials.v1.GenerateAccessTokenRequest;
 import com.google.cloud.iam.credentials.v1.GenerateAccessTokenResponse;
 import com.google.cloud.iam.credentials.v1.IamCredentialsClient;
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
@@ -43,6 +47,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.polaris.core.config.FeatureConfiguration;
@@ -54,6 +59,7 @@ import org.apache.polaris.core.storage.PolarisStorageActions;
 import org.apache.polaris.core.storage.PolarisStorageIntegration;
 import org.apache.polaris.core.storage.StorageAccessConfig;
 import org.apache.polaris.core.storage.StorageAccessProperty;
+import org.apache.polaris.core.storage.StorageLocationPreparer;
 import org.apache.polaris.core.storage.StorageUri;
 import org.apache.polaris.core.storage.cache.StorageCredentialCache;
 import org.apache.polaris.core.storage.cache.StorageCredentialCacheKey;
@@ -82,6 +88,7 @@ public class GcpCredentialsStorageIntegration
   private final HttpTransportFactory transportFactory;
   private final GcpCredentialOps credentialOps;
   private final Optional<GcpAttributionParams> attributionParams;
+  private final StorageLocationPreparer folderPreparer;
 
   public GcpCredentialsStorageIntegration(
       GoogleCredentials sourceCredentials,
@@ -95,7 +102,8 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         GcpCredentialOps.DEFAULT,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
   }
 
   public GcpCredentialsStorageIntegration(
@@ -111,7 +119,8 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         GcpCredentialOps.DEFAULT,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
   }
 
   public GcpCredentialsStorageIntegration(
@@ -127,7 +136,8 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         credentialOps,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
   }
 
   public GcpCredentialsStorageIntegration(
@@ -144,7 +154,45 @@ public class GcpCredentialsStorageIntegration
         storageConfig,
         realmConfig,
         credentialOps,
-        resolveAttributionParams(realmConfig));
+        resolveAttributionParams(realmConfig),
+        locations -> {});
+  }
+
+  public GcpCredentialsStorageIntegration(
+      GoogleCredentials sourceCredentials,
+      HttpTransportFactory transportFactory,
+      StorageCredentialCache cache,
+      GcpStorageConfigurationInfo storageConfig,
+      RealmConfig realmConfig,
+      @NonNull StorageLocationPreparer folderPreparer) {
+    this(
+        sourceCredentials,
+        transportFactory,
+        cache,
+        storageConfig,
+        realmConfig,
+        GcpCredentialOps.DEFAULT,
+        resolveAttributionParams(realmConfig),
+        folderPreparer);
+  }
+
+  public GcpCredentialsStorageIntegration(
+      GoogleCredentials sourceCredentials,
+      HttpTransportFactory transportFactory,
+      StorageCredentialCache cache,
+      GcpStorageConfigurationInfo storageConfig,
+      RealmConfig realmConfig,
+      GcpCredentialOps credentialOps,
+      @NonNull StorageLocationPreparer folderPreparer) {
+    this(
+        sourceCredentials,
+        transportFactory,
+        cache,
+        storageConfig,
+        realmConfig,
+        credentialOps,
+        resolveAttributionParams(realmConfig),
+        folderPreparer);
   }
 
   /**
@@ -160,6 +208,26 @@ public class GcpCredentialsStorageIntegration
       RealmConfig realmConfig,
       GcpCredentialOps credentialOps,
       Optional<GcpAttributionParams> attributionParams) {
+    this(
+        sourceCredentials,
+        transportFactory,
+        cache,
+        storageConfig,
+        realmConfig,
+        credentialOps,
+        attributionParams,
+        locations -> {});
+  }
+
+  public GcpCredentialsStorageIntegration(
+      GoogleCredentials sourceCredentials,
+      HttpTransportFactory transportFactory,
+      StorageCredentialCache cache,
+      GcpStorageConfigurationInfo storageConfig,
+      RealmConfig realmConfig,
+      GcpCredentialOps credentialOps,
+      Optional<GcpAttributionParams> attributionParams,
+      @NonNull StorageLocationPreparer folderPreparer) {
     super(cache, realmConfig, storageConfig);
     // Needed for when environment variable GOOGLE_APPLICATION_CREDENTIALS points to google service
     // account key json
@@ -168,6 +236,7 @@ public class GcpCredentialsStorageIntegration
     this.transportFactory = transportFactory;
     this.credentialOps = credentialOps;
     this.attributionParams = attributionParams;
+    this.folderPreparer = Objects.requireNonNull(folderPreparer, "folderPreparer");
   }
 
   /**
@@ -208,6 +277,11 @@ public class GcpCredentialsStorageIntegration
             realmConfig.getConfig(FeatureConfiguration.GCS_PRINCIPAL_ATTRIBUTION_WIF_AUDIENCE),
             realmConfig.getConfig(FeatureConfiguration.GCS_PRINCIPAL_ATTRIBUTION_SIGNING_KEY_FILE),
             realmConfig.getConfig(FeatureConfiguration.GCS_PRINCIPAL_ATTRIBUTION_SIGNING_KEY_ID)));
+  }
+
+  @Override
+  public void prepareLocations(@NonNull List<String> locations) {
+    folderPreparer.prepareLocations(locations);
   }
 
   @Override
@@ -311,7 +385,11 @@ public class GcpCredentialsStorageIntegration
         resolveSourceCredentials(key, gcpStorageConfig, sourceCredentials, credentialOps);
 
     CredentialAccessBoundary accessBoundary =
-        generateAccessBoundaryRules(readLocations, listLocations, writeLocations);
+        generateAccessBoundaryRules(
+            readLocations,
+            listLocations,
+            writeLocations,
+            bucket -> isHnsBucket(bucket, credentialsToDownscope));
     DownscopedCredentials credentials =
         DownscopedCredentials.newBuilder()
             .setHttpTransportFactory(transportFactory)
@@ -386,7 +464,7 @@ public class GcpCredentialsStorageIntegration
    * Returns the credential to be used as the source for downscoping. If a specific service account
    * is configured, it impersonates that account first.
    */
-  private static GoogleCredentials getBaseCredentials(
+  public static GoogleCredentials getBaseCredentials(
       GcpStorageConfigurationInfo storageConfig,
       GoogleCredentials sourceCredentials,
       GcpCredentialOps credentialOps) {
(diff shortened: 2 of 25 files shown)


spec/polaris-management-service.yml
```diff
@@ -1193,6 +1193,13 @@ components:
             gcsServiceAccount:
               type: string
               description: a Google cloud storage service account
+            hierarchicalNamespace:
```
dimas-b: nit: a similar field in Azure is called simply `hierarchical`

dimas-b (commented):
Thanks for your contribution, @varunarya002 !

Since this PR affects REST API parameters, please open a corresponding discussion on the `dev` [ML](https://polaris.apache.org/community/), which is a usual practice in Polaris.

(this is not a complete review from my side 😅 ... just noting a couple of points for a start)

dimas-b:
@varunarya002 : Do you have capacity to push this PR forward? If you do not have time, it's fine. We can build something based of your work and still give you attribution.

varunarya002:
Hi @dimas-b. I will be starting discussion about this change on dev ML.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
dimas-b, line 1: Hi @varunarya002 !

I did a quick test in my env. with an HNS bucket in GCS, but unfortunately, I hit a 403 when Polaris tries to write metadata:

Catalog:
```
curl -H "Authorization: Bearer $POLARIS_TOKEN" -H 'Content-Type: application/json' \
  -X POST http://localhost:8181/api/management/v1/catalogs \
  -d '{"type":"INTERNAL","name":"polaris","properties": \
     {"default-base-location":"gs://***"},"storageConfigInfo": \
     {"storageType":"GCS", "hierarchicalNamespace":true}}'
```

In Spark:
```
create table ns.t1 (n string);
```

Polaris log:
```
2026-04-03 09:15:35,895 INFO  [org.apa.pol.ser.exc.IcebergExceptionMapper] [06b3d4a3-10a0-49be-a783-61f624b0603d_0000000000000000011,POLARIS] [,,,] (executor-thread-1) Handling runtimeException 403 Forbidden
	|> PUT https://storage.googleapis.com/upload/storage/v1/b/***/o?name=ns/t1/metadata/00000-5cf5ef8c-cf72-4a3e-95d1-964d8d2e36e3.metadata.json&uploadType=resumable&upload_id=***
	|> content-range: bytes 0-722/723
	|> x-goog-hash: ****
	|> x-goog-gcs-idempotency-token: ****
	|  
	|< HTTP/1.1 403 Forbidden
	|< content-length: 552
	|< content-type: application/json; charset=UTF-8
	|< x-guploader-uploadid: ****
	|  
```

I did not debug the code, though 😅 

Did I miss anything in the config? Do you have any suggestions for how to use the new feature?

varunarya002: Let me check it

varunarya002: Hi @dimas-b, updated the pr with pre-constructed directories needed for metadata and data locations.
Please try again.

PS: Refactor HNS detection logic to auto rather being explicit as a table can have metadata and data on multiple buckets, hence auto detection will create necessary directories on HNS buckets.

dimas-b: Thanks for the update, @varunarya002 ! The PR works in my local env now :tada: ... will review the code presently.

polaris-core/src/test/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegrationTest.java
```diff
@@ -384,6 +384,118 @@ protected AccessToken refreshAccessToken(DownscopedCredentials credentials) {
...
+        GcpCredentialsStorageIntegration.generateAccessBoundaryRules(
+            true, Set.of("gs://bucket1/path/to/data"), Set.of("gs://bucket1/path/to/data"));
+    assertThat(credentialAccessBoundary).isNotNull();
+    // Should have 3 rules: read + write + folder admin
+    assertThat(credentialAccessBoundary.getAccessBoundaryRules()).hasSize(3);
+    ObjectMapper mapper = JsonMapper.builder().build();
+    JsonNode parsedRules = mapper.convertValue(credentialAccessBoundary, JsonNode.class);
+    JsonNode refRules = readResource(mapper, "gcp-testGenerateAccessBoundaryHnsEnabled.json");
+    assertThat(parsedRules)
+        .usingRecursiveComparison(
```
dimas-b: nit: does simple `equals` not work in this case?

runtime/service/src/main/java/org/apache/polaris/service/storage/gcs/GcsStorageLocationPreparer.java
```diff
@@ -0,0 +1,367 @@
...
+
+    List<FolderToCreate> allFolders = resolveAllFoldersToCreate(tableLocation, tableProperties);
+    if (allFolders.isEmpty()) {
+      return;
+    }
+
+    // Group folders by bucket and create only for HNS-enabled buckets
+    allFolders.stream()
+        .collect(Collectors.groupingBy(FolderToCreate::bucketName))
+        .forEach(this::createFoldersForBucketIfHnsEnabled);
```
dimas-b: It looks like (at least in my env) that with pre-created folders, changes in `GcpCredentialsStorageIntegration` are not actually required... WDYT?

varunarya002: For unpartitioned table it works fine, but for partitioned table it does not create deeper structures like partition subdirectories (e.g., data/year=2024/month=01/). On HNS enabled buckets, if Iceberg clients need to create partition folders at write time, they'd need folder creation permissions in the downscoped token.

dimas-b: Good point :+1:

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  private StorageLocationPreparerFactory() {
+    this.storageConfiguration = null;
+    this.clock = Clock.systemUTC();
+  }
+
+  public StorageLocationPreparer create(@Nonnull PolarisStorageConfigurationInfo storageConfig) {
+    if (storageConfig instanceof GcpStorageConfigurationInfo && storageConfiguration != null) {
+      return new GcsStorageLocationPreparer(storageConfiguration.gcpCredentialsSupplier(clock));
```
dimas-b: Could we use `StorageAccessConfigProvider` for storage credentials?

runtime/service/src/main/java/org/apache/polaris/service/storage/gcs/GcsStorageLocationPreparer.java
```diff
@@ -0,0 +1,367 @@
...
+   * table properties take precedence over defaults.
+   */
+  private List<String> resolveAllPaths(String tableLocation, Map<String, String> tableProperties) {
+    String metadataPath =
+        Optional.ofNullable(tableProperties.get(TableProperties.WRITE_METADATA_LOCATION))
+            .or(
+                () ->
+                    Optional.ofNullable(
+                        tableProperties.get(
+                            IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY)))
```
dimas-b: This does not look specific to GCS. Could we handle this part in a more general way?

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparer.java
```diff
@@ -0,0 +1,25 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.storage;
+
+import java.util.Map;
+
+public interface StorageLocationPreparer {
+  void prepareTableLocation(String tableLocation, Map<String, String> tableProperties);
```
dimas-b: I wonder if we could simplify this SPI 🤔 I suppose we could simply request a certain set of folders to be created. Do you envision more sophisticated work to be done (in some future cases) for preparing table locations?

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    this.clock = clock;
+  }
+
+  private StorageLocationPreparerFactory() {
+    this.storageConfiguration = null;
+    this.clock = Clock.systemUTC();
+  }
+
+  public StorageLocationPreparer create(@Nonnull PolarisStorageConfigurationInfo storageConfig) {
+    if (storageConfig instanceof GcpStorageConfigurationInfo && storageConfiguration != null) {
```
dimas-b: Let's leverage CDI via storage type-based `@Identifier` annotations... similar to `ServiceProducers.polarisAuthorizerFactory()`

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
varunarya002: It's not feasible because of different trust levels (admin creds for HNS folder creation vs subscoped client tokens).

runtime/service/src/main/java/org/apache/polaris/service/storage/gcs/GcsStorageLocationPreparer.java
varunarya002: Extracted AbstractStorageLocationPreparer base class with generic URI parsing, hierarchy building, bucket grouping.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparer.java
varunarya002: Changed to void prepareLocations(List<String>) — no Iceberg concepts leak into the SPI.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
varunarya002: Factory uses Instance<StorageLocationPreparer> lookup by StorageType.name(). GCS preparer annotated with @Identifier("GCS").

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
dimas-b: Hmmm... sorry, but I'm a bit confused now... Clients (e.g. Spark) should be able to create folders (for new data files with Iceberg partitioning) using the vended credentials, right?

dimas-b:
Related `dev` thread: https://lists.apache.org/thread/jyx21cmbprttb5p6lr2q0q1ytn0jgv0w

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -271,6 +273,28 @@ public static CredentialAccessBoundary generateAccessBoundaryRules(
...
+            return;
+          }
+          CredentialAccessBoundary.AccessBoundaryRule.Builder builder =
+              CredentialAccessBoundary.AccessBoundaryRule.newBuilder();
+          builder.setAvailableResource(bucketResource(bucket));
+          builder.setAvailabilityCondition(
+              CredentialAccessBoundary.AccessBoundaryRule.AvailabilityCondition.newBuilder()
+                  .setExpression(String.join(" || ", folderConditions))
+                  .build());
+          builder.setAvailablePermissions(List.of("inRole:roles/storage.folderAdmin"));
```
dimas-b, line 295: We add this unconditionally in this code, but it is required only for HNS, right?

I suppose it would be more robust to check `resolveHnsStatus()` here or have an HNS flag in the Storage Config.... I'm kind of leaning toward the flag, even though we can auto-detect this... WDYT?

My thinking is that users are generally aware of the the HNS flag in their GCS storage (or al least they should be) and it is not likely to change in runtime.

runtime/service/src/main/java/org/apache/polaris/service/storage/AbstractStorageLocationPreparer.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Base class for storage location preparers that handles URI parsing, folder hierarchy resolution,
+ * and bucket-level grouping. Subclasses implement the storage-specific folder creation logic.
+ *
+ * <p>The template method {@link #prepareLocations(List)} drives the pipeline: filter by scheme,
+ * resolve folder hierarchies, group by bucket, and delegate to {@link
+ * #createFoldersForBucket(String, List)} for storage-specific operations.
+ */
+public abstract class AbstractStorageLocationPreparer implements StorageLocationPreparer {
```
sungwy, line 40: I feel this name is a bit too abstract (pun intended 🙂). Would it make sense to give it a more descriptive name like `HierarchicalFolderLocationPreparer` or `ObjectStoreFolderPreparer`?

I also wonder whether introducing this abstraction is a bit premature. Do we expect these methods to be reusable across other cloud providers, or is this really tailored to the GCS HNS case for now?

If it is mainly the latter, would it make sense to fold this into `GcsStorageLocationPreparer` for now and only extract a shared abstraction once we have a second provider implementation and can validate the actual common ground?

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
```diff
@@ -0,0 +1,59 @@
...
+  private StorageLocationPreparerFactory() {
+    this.preparers = null;
+  }
+
+  public StorageLocationPreparer create(@Nonnull PolarisStorageConfigurationInfo storageConfig) {
+    if (preparers == null) {
+      return NO_OP;
+    }
+    String key = storageConfig.getStorageType().name();
+    Instance<StorageLocationPreparer> selected = preparers.select(Identifier.Literal.of(key));
```
sungwy, line 52: I like the `StorageLocationPreparer` as an abstraction, but instead of introducing it as a separate standalone selection path, would it make more sense for it to be exposed by `PolarisStorageIntegration` instead?

[PolarisStorageIntegration is already selected](https://github.com/apache/polaris/blob/23b45f9a0197d87afe3cc0898a37e8834fabf07f/runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java#L116) from the storage configuration type, so grouping storage-specific behavior there seems cleaner and makes the config-driven selection more obvious and consistent.

I do think it still makes sense to model `StorageLocationPreparer` as a separate capability. I’m just wondering whether `PolarisStorageIntegration` should be single the interface that exposes that capability, rather than introducing a parallel factory and dispatch path for storage-specific behavior.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
sungwy: I agree with @dimas-b 's comment - is it a concern that we are introducing this permission for any GCS backed tables? That IMO goes against the principle of granting least-privileged access, so I think it would be helpful to review if that exposes any unnecessary actions for non HNS enabled tables.

sungwy (commented):
Hi @varunarya002 thank you so much for putting together this PR. The new `StorageLocationPreparer` abstraction is a highly reusable idea, and I think it is directionally correct. I took a first pass at reviewing this PR and left some comments.

Also, could we add a bit more context in the PR description about this model? We are effectively introducing a pre-createTable hook into the table creation flow, so it might be worth calling that out explicitly.

dimas-b:
Hi @varunarya002 : Thanks again for this PR. Do you have capacity to resolve conflicts and comments?

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
varunarya002: Addressed in the latest force-push. Two changes here:

1. The folder rule is now **gated on per-bucket HNS auto-detection** — `generateAccessBoundaryRules()` takes a `Predicate<String> isHnsBucket` and only emits the folder rule for buckets where it returns `true`. `GcpCredentialsStorageIntegration.isHnsBucket()` queries `Storage.get(bucket, HIERARCHICAL_NAMESPACE)` per bucket. We went with auto-detection rather than a config flag so a single catalog can serve a mix of HNS and non-HNS buckets correctly (a catalog's `default-base-location` is often one bucket, but `write.data.path`/`write.metadata.path` can point elsewhere — the auto-detect handles that per-bucket without admins having to enumerate which buckets are HNS).

2. **Permissions narrowed** — replaced `inRole:roles/storage.folderAdmin` with the two individual permissions we actually need: `storage.folders.create` and `storage.folders.get`. The role would have additionally granted `setIamPolicy`/`getIamPolicy`/`delete`/`rename`/`list` on both `folders/` and `managedFolders/` resources, none of which clients need for write paths.

Also dropped the `managedFolders/` resource condition — we use the Storage Control API's `folders/` resource path, not managed folders.

varunarya002: Addressed. The folder rule now uses individual permissions (`storage.folders.create` + `storage.folders.get`) instead of `inRole:roles/storage.folderAdmin`, and is only emitted for buckets where HNS is detected via an injected `Predicate<String> isHnsBucket`. Non-HNS GCS-backed tables get no extra permissions, restoring least-privilege.

The `managedFolders/` resource condition was also dropped since we use HNS folders (`folders/`), not managed folders.

runtime/service/src/main/java/org/apache/polaris/service/storage/AbstractStorageLocationPreparer.java
varunarya002: Renamed to `HierarchicalFolderLocationPreparer` — using your suggestion verbatim. We kept the abstract base (rather than folding it into `GcsStorageLocationPreparer`) because Azure ADLS Gen2 is the obvious next reuse case (also has hierarchical-namespace folder semantics), and the generic URI parsing + path-hierarchy logic is genuinely cloud-agnostic.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
varunarya002: Done. `prepareLocations(List<String>)` is now a method on `PolarisStorageIntegration` with a no-op default; non-GCS storage types pay nothing.

`GcpCredentialsStorageIntegration` takes a `Consumer<List<String>> folderPreparer` constructor parameter and delegates `prepareLocations` to it. `PolarisStorageIntegrationProviderImpl` (which already switches on storage type) injects `GcsStorageLocationPreparer` when constructing the GCP integration.

This deletes `StorageLocationPreparerFactory` and the `@Identifier("GCS")` CDI lookup entirely — single selection mechanism (the provider's type switch), and the module boundary is respected: `polaris-core` only declares the interface signature, the runtime-side preparer with `StorageControlClient` stays in `runtime/service`. `IcebergCatalogHandler` calls `integration.prepareLocations(locations)` directly.

runtime/service/src/main/java/org/apache/polaris/service/storage/StorageLocationPreparerFactory.java
varunarya002: Yes, exactly — Spark *does* create folders with the vended credentials, and the new design preserves that. Two distinct folder-creation responsibilities, with different credential trust levels:

**1. Server-side, at table creation time** — using the catalog's service-account credentials (admin-level), `GcsStorageLocationPreparer` pre-creates the table's known top-level folders (`table/`, `table/metadata/`, `table/data/`). This happens in `IcebergCatalogHandler.createTableDirect/createTableStaged` *before* Iceberg writes the initial table metadata.

**2. Client-side, at write time** — Spark needs to create *deeper* folders that the server can't enumerate ahead of time (partition subdirectories like `data/year=2024/month=01/`). For this, the vended downscoped credentials now include `storage.folders.create` + `storage.folders.get` (narrowly scoped to the catalog's write locations) when the target bucket has HNS enabled. That's the change in `GcpCredentialsStorageIntegration.generateAccessBoundaryRules()`.

Earlier I conflated these two paths in my `StorageAccessConfigProvider` reply — apologies for the confusion. The vended credentials *do* let Spark create folders; they just don't include the unrelated admin operations that `roles/storage.folderAdmin` would have granted (e.g. `setIamPolicy`).

polaris-core/src/test/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegrationTest.java
varunarya002: The recursive-comparison pattern matches the existing convention in this file — see `testGenerateAccessBoundary` (line 202) and its peers, which all use `usingRecursiveComparison` + the `recursiveEquals` helper. I kept the new HNS tests on the same pattern for consistency.

That said, the new HNS tests also have stricter assertions where it matters — e.g. `testGenerateAccessBoundaryFolderRuleUsesNarrowPermissions` explicitly asserts `containsExactlyInAnyOrder("storage.folders.create", "storage.folders.get")` and that the expression doesn't reference `managedFolders/`. The JSON fixtures cover the structural rule emission shape.

spec/polaris-management-service.yml
varunarya002: The `hierarchicalNamespace` field has been removed from the OpenAPI spec entirely. HNS is now auto-detected per bucket at credential-vending time, so no user-facing flag is needed (which also means a single catalog can serve a mix of HNS and non-HNS buckets correctly). If we add the flag back in a follow-up for any reason, we'll use `hierarchical` to match the Azure naming.

varunarya002:
Thanks @dimas-b and @sungwy for the thorough review. Just pushed a v2 that addresses every open thread. Summary of changes since the last round:

**Design changes**
- HNS is now **auto-detected per bucket at credential-vending time** via `Storage.get(bucket, HIERARCHICAL_NAMESPACE)`. `generateAccessBoundaryRules()` accepts a `Predicate<String> isHnsBucket` and the folder rule is emitted *only* for HNS buckets. Non-HNS buckets get no additional permissions — least-privilege preserved.
- The folder rule uses **narrow individual permissions** `storage.folders.create` + `storage.folders.get`, replacing the previous overly-broad `roles/storage.folderAdmin`. Dropped the `managedFolders/` resource condition entirely (we use the Storage Control API's `folders/` resource, not managed folders).
- `StorageLocationPreparer` is now **exposed via `PolarisStorageIntegration.prepareLocations(List<String>)`** with a no-op default. The standalone `StorageLocationPreparerFactory` and `@Identifier("GCS")` CDI lookup are deleted — single selection path through `PolarisStorageIntegrationProviderImpl`'s existing storage-type switch. Module boundary respected: only the interface signature lives in `polaris-core`; the runtime-side preparer with `StorageControlClient` stays in `runtime/service`.
- `AbstractStorageLocationPreparer` renamed to **`HierarchicalFolderLocationPreparer`** (using @sungwy's suggested name). Kept as a base class because Azure ADLS Gen2 is the obvious next reuse case — the URI parsing + path-hierarchy logic is genuinely cloud-agnostic.

**PR description rewrite**: now explicitly explains the pre-createTable hook model and the server-side / client-side responsibility split, per @sungwy's ask.

**Rebase**: branch is rebased onto current `apache/main` (was 200+ commits behind). Force-pushed as one squashed commit. Per-thread replies posted on every open comment.

Tests added for predicate gating, narrow-permission shape, multi-bucket scenarios, and the mixed HNS/non-HNS-in-one-catalog case. All build green locally on Java 21 (`:polaris-core:test` and `:polaris-runtime-service:test --tests "*GcsStorageLocationPreparerTest*"`).

Ready for another look when you have time.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -261,9 +296,67 @@ public static CredentialAccessBoundary generateAccessBoundaryRules(
...
+   * Detects whether a bucket has Hierarchical Namespace (HNS) enabled by fetching its metadata.
+   * Called per-bucket at credential-vending time. HNS status is bucket-immutable, so caching is a
+   * safe follow-up optimization; not implemented here to keep this PR focused.
+   *
+   * <p>On any failure (network, missing bucket, permission), returns {@code false} — non-HNS rules
+   * still apply, so the worst case is that an HNS bucket misses the folder-permission grant and
+   * subsequent writes fail with a clear 403, rather than silently over-permissioning.
+   */
+  @VisibleForTesting
+  boolean isHnsBucket(String bucket) {
```
dimas-b: I guess it might be worth covering this with a `FeatureConfiguration` flag (enabled by default) in case some prior deployments do not wish to execute the extra call for some internal reason... WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -236,6 +241,67 @@ private CatalogEntity getResolvedCatalogEntity() {
...
+      CatalogEntity catalog,
+      TableIdentifier tableIdentifier,
+      @Nullable String requestLocation,
+      @Nullable Map<String, String> tableProperties) {
+    PolarisStorageConfigurationInfo storageConfig = catalog.getStorageConfigurationInfo();
+    if (storageConfig == null) {
+      return;
+    }
+    PolarisStorageIntegration<?> integration =
+        storageIntegrationProvider().getStorageIntegrationForConfig(storageConfig);
```
dimas-b: This looks like a conflict WRT #3699 , which removes this method 🤔  

I tend to think that #3699 is beneficial and would like that refactoring to land.

@varunarya002 : Will @tokoko 's approach work for you?

runtime/service/src/main/java/org/apache/polaris/service/storage/HierarchicalFolderLocationPreparer.java
```diff
@@ -0,0 +1,139 @@
...
+ * the storage-specific folder-creation call.
+ *
+ * <p>Today there is one concrete implementation (GCS HNS). Azure ADLS Gen2 is the next likely reuse
+ * case, hence the abstraction lives in this package rather than inside the GCS class.
+ *
+ * <p>The template method {@link #prepareLocations(List)} drives the pipeline: filter by scheme,
+ * resolve folder hierarchies, group by bucket, and delegate to {@link
+ * #createFoldersForBucket(String, List)}.
+ */
+public abstract class HierarchicalFolderLocationPreparer implements Consumer<List<String>> {
```
dimas-b: Let's define a dedicated interface for `Consumer<List<String>>` for ease of navigation in IDEs.

runtime/service/src/main/java/org/apache/polaris/service/storage/gcs/GcsStorageLocationPreparer.java
```diff
@@ -0,0 +1,179 @@
...
+  private void createFolderIfNotExists(
+      StorageControlClient controlClient, String parent, String folderPath) {
+    try {
+      CreateFolderRequest request =
+          CreateFolderRequest.newBuilder().setParent(parent).setFolderId(folderPath).build();
+
+      controlClient.createFolder(request);
+      LOGGER.atDebug().addKeyValue("folder", folderPath).log("Created HNS folder");
+    } catch (AlreadyExistsException e) {
+      LOGGER.atDebug().addKeyValue("folder", folderPath).log("HNS folder already exists, skipping");
```
dimas-b: Is `folder` actually logged even though the message does not reference it? (just to double check).

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
```diff
@@ -114,6 +115,16 @@ public abstract StorageAccessConfig getSubscopedCreds(
...
 
+  /**
+   * Hook invoked before a table is created to pre-materialize storage locations that the underlying
+   * storage system requires to exist before files can be written into them.
+   *
+   * <p>The default is a no-op. Storage types that need explicit folder creation (e.g. GCS with
+   * Hierarchical Namespace enabled) override this. Passed locations include the table location and
+   * its metadata/data subpaths.
+   */
+  public void prepareLocations(@Nonnull List<String> locations) {}
```
dimas-b: Passing a note from artificial helpers:
> High: the new pre-create hook runs before Polaris has validated the requested table locations, so a rejected create request can still mutate GCS. In both
     direct and staged flows, prepareStorageForTable(...) is invoked before any of the location checks that happen later in IcebergCatalog
     (validateLocationsForTableLike, overlap checks, metadata-in-table-dir checks). That means a caller can point location, write.metadata.location, or
     write.data.location at a path Polaris should reject, and Polaris will still try to create HNS folders there with the service account first. See runtime/
     service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java:252, runtime/service/src/main/java/org/apache/polaris/service/
     catalog/iceberg/IcebergCatalogHandler.java:535, runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java:658,
     and the later validation in runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java:1541.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -236,6 +241,67 @@ private CatalogEntity getResolvedCatalogEntity() {
...
+    }
+    String effectiveLocation = resolveTableLocation(catalog, tableIdentifier, requestLocation);
+    if (effectiveLocation == null) {
+      return;
+    }
+    Map<String, String> props = tableProperties != null ? tableProperties : Map.of();
+    integration.prepareLocations(resolveStorageLocations(effectiveLocation, props));
+  }
+
+  private @Nullable String resolveTableLocation(
```
dimas-b: Passing feedback from artificial helpers:
> High: resolveTableLocation() guesses the default table path from catalog.getBaseLocation(), but actual table creation does not use that rule. For implicit
     locations, Iceberg resolves through namespace-aware logic in defaultWarehouseLocation(...), including namespace-specific locations; staged create also
     derives the final location from the catalog builder. As written, HNS folder creation will target the wrong path whenever the namespace has its own
     location or other catalog-side location logic applies, so the new feature breaks exactly for non-default namespace layouts. See runtime/service/src/main/
     java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java:274, runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/
     IcebergCatalogHandler.java:576, and the real location resolution in runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/
     IcebergCatalog.java:366.

dimas-b (commented):
Thanks for pushing this PR forward, @varunarya002 ! Some more comments below.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
dimas-b: Alternatively we could add a GCS storage config property `hierarchical` with values `yes`, `no`, `auto` (default)... WDYT? This could be done in a separate PR, of course.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: this seems to have been addresses, resolving

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
dimas-b: this seems to have been addresses, resolving

runtime/service/src/main/java/org/apache/polaris/service/storage/gcs/GcsStorageLocationPreparer.java
```diff
@@ -0,0 +1,182 @@
...
+                    String.format(
+                        "GCS bucket '%s' not found. Cannot proceed with table creation.",
+                        bucketName)));
+  }
+
+  private void createFolderIfNotExists(
+      StorageControlClient controlClient, String parent, String folderPath) {
+    try {
+      CreateFolderRequest request =
+          CreateFolderRequest.newBuilder().setParent(parent).setFolderId(folderPath).build();
```
dimas-b: > P1 runtime/service/src/main/java/org/apache/polaris/service/storage/gcs/GcsStorageLocationPreparer.java:160: CreateFolderRequest sends
    folderId(folderPath) values like warehouse/ns/table, but the Storage Control API requires folder_id to end with /, and documents the folder field
    as required. The values are produced without trailing slashes in runtime/service/src/main/java/org/apache/polaris/service/storage/
    HierarchicalFolderLocationPreparer.java:131, so real HNS folder creation will fail for the main feature path.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -385,16 +473,64 @@ public static CredentialAccessBoundary generateAccessBoundaryRules(
...
+            return;
+          }
+          CredentialAccessBoundary.AccessBoundaryRule.Builder builder =
+              CredentialAccessBoundary.AccessBoundaryRule.newBuilder();
+          builder.setAvailableResource(bucketResource(bucket));
+          builder.setAvailabilityCondition(
+              CredentialAccessBoundary.AccessBoundaryRule.AvailabilityCondition.newBuilder()
+                  .setExpression(String.join(" || ", conditions))
+                  .build());
+          builder.setAvailablePermissions(List.of("storage.folders.create", "storage.folders.get"));
```
dimas-b: > P1 polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java:488: the new CAB rule uses raw
    permissions storage.folders.create / storage.folders.get. Google’s Credential Access Boundary docs say availablePermissions must be role
    identifiers prefixed with inRole: and that permission names cannot be specified directly. HNS credential vending will fail or produce unusable
    downscoped credentials.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -209,7 +266,11 @@ static StorageAccessConfig compute(GcpStorageCredentialCacheKey key) {
         getBaseCredentials(gcpStorageConfig, sourceCredentials, credentialOps);
```
dimas-b: This might have been discussed before... sorry for repeating in that case, but I wonder whether the `--service-account` point is relevant 🤔 

> P2 polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java:265 and runtime/service/src/main/java/
    org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java:100: HNS detection and server-side folder creation use the original
    GCP provider credentials, not the configured catalog gcpServiceAccount credentials used for impersonation. Catalogs that rely on --service-
    account can have valid vended/storage credentials but still fail HNS checks or omit folder rules.

runtime/service/src/main/java/org/apache/polaris/service/storage/HierarchicalFolderLocationPreparer.java
```diff
@@ -0,0 +1,135 @@
...
+      return List.of();
+    }
+
+    return buildPathHierarchy(objectPath).stream()
+        .map(folder -> new FolderToCreate(bucketName, folder))
+        .toList();
+  }
+
+  /** Strips the scheme, bucket, and slashes from a storage URI, returning the object path. */
+  protected static String extractObjectPath(String storageUri) {
```
dimas-b, line 105: > P2 runtime/service/src/main/java/org/apache/polaris/service/storage/HierarchicalFolderLocationPreparer.java:105: URI.create(...).getPath() drops
    valid object-path content after ? or #. Polaris already has StorageUri specifically to preserve those characters, so HNS preparation can create
    folders for the wrong truncated path.

dimas-b (commented):
Thanks for the update, @varunarya002 !

I'm posting some more comments from artificial helpers. I did not have time to dig deeper into them, but they seem valid at first glace. I'd appreciate your assessment of them. Thx!

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
varunarya002: I left this PR on auto-detection rather than adding a `hierarchical: yes/no/auto` storage-config knob. Auto-detection is per bucket and now uses the same resolved/impersonated credentials as credential vending / folder preparation, so a single catalog can still serve mixed HNS and non-HNS buckets without extra user-facing config. An explicit override could make sense as a follow-up if we find auto-detection latency or required `storage.buckets.get` access is a practical problem, but I’d prefer not to expand the config surface in this PR unless you feel strongly.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageIntegration.java
```diff
@@ -55,4 +55,14 @@ StorageAccessConfig getStorageAccessConfig(
...
+
+  /**
+   * Hook invoked before a table is created to pre-materialize storage locations that the underlying
+   * storage system requires to exist before files can be written into them.
+   *
+   * <p>The default is a no-op. Storage types that need explicit folder creation (e.g. GCS with
+   * Hierarchical Namespace enabled) override this. Passed locations include the table location and
+   * its metadata/data subpaths.
+   */
+  default void prepareLocations(@NonNull List<String> locations) {}
```
flyingImer, line 67: Could we first re-validate whether this hook is required for correctness? Google’s current HNS behavior automatically creates missing parent folders during object creation; explicit pre-creation is documented as a performance optimization rather than a prerequisite.

Since Polaris loads the WRITE-capable FileIO immediately after this hook, I think the default should be to solve the folder-resource CAB and remove this second storage-mutation path unless a real HNS test shows otherwise. If we retain it as an optimization, we also need explicit failure and cleanup semantics because a later create failure can leave empty folders behind.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -526,16 +631,64 @@ public static CredentialAccessBoundary generateAccessBoundaryRules(
...
+            return;
+          }
+          CredentialAccessBoundary.AccessBoundaryRule.Builder builder =
+              CredentialAccessBoundary.AccessBoundaryRule.newBuilder();
+          builder.setAvailableResource(bucketResource(bucket));
+          builder.setAvailabilityCondition(
+              CredentialAccessBoundary.AccessBoundaryRule.AvailabilityCondition.newBuilder()
+                  .setExpression(String.join(" || ", conditions))
+                  .build());
+          builder.setAvailablePermissions(List.of("inRole:roles/storage.folderAdmin"));
```
flyingImer, line 646: Can we avoid `roles/storage.folderAdmin` here? The PR description says the token receives only `storage.folders.create` and `storage.folders.get`, but this role’s permission upper bound also includes folder delete, list, and rename.

`roles/storage.objectCreator` already contains `storage.folders.create`; could we use the narrowest predefined or custom role set that a real HNS client trace proves is required, and update the PR description to match the implementation?

runtime/service/src/main/java/org/apache/polaris/service/storage/HierarchicalFolderLocationPreparer.java
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Builds the parent hierarchy for an object path: hierarchical storage systems require every
+   * parent folder to exist before a child can be created.
+   *
+   * <p>For example, {@code "warehouse/ns1/table1"} produces {@code ["warehouse", "warehouse/ns1",
+   * "warehouse/ns1/table1"]}.
+   */
+  public static List<String> buildPathHierarchy(String objectPath) {
+    String[] segments =
+        Arrays.stream(objectPath.split(PATH_SEPARATOR)).filter(NOT_BLANK).toArray(String[]::new);
```
flyingImer, line 124: Could this preserve consecutive slashes rather than filtering empty path segments? GCS HNS treats every `/` as a folder boundary, so `a//b` is a different hierarchy from `a/b`, but this implementation maps both to the same folder tree.

If explicit creation remains, passing the exact raw folder ID and using `recursive=true` would also avoid reimplementing the provider’s hierarchy semantics here.

polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java
```diff
@@ -526,16 +631,64 @@ public static CredentialAccessBoundary generateAccessBoundaryRules(
...
+        return false;
+      }
+      BucketInfo.HierarchicalNamespace hns = bucketMetadata.getHierarchicalNamespace();
+      return hns != null && Boolean.TRUE.equals(hns.getEnabled());
+    } catch (RuntimeException e) {
+      LOGGER
+          .atWarn()
+          .addKeyValue("bucket", bucket)
+          .log("Failed to determine HNS status; defaulting to non-HNS", e);
+      return false;
```
flyingImer, line 669: Should an HNS metadata lookup failure be explicit rather than silently classified as non-HNS? A transient failure or missing `storage.buckets.get` permission currently produces credentials without the folder rule, so vending succeeds and the user only sees a later write 403.

The server-side preparation path treats the same uncertainty differently. If auto-detection remains, I think we need one documented and tested failure contract across both paths.

flyingImer (commented):
Thanks for continuing to iterate on this PR. I’m requesting changes because I think the current design needs one more premise check against the current GCS HNS contract, rather than another round of implementation polish.

The key merge question is: what exact operation still fails when the supported client uses a correctly scoped folder-resource CAB on an HNS bucket? GCS now documents that object creation automatically creates missing parent folders, while explicit pre-creation is a performance optimization. The existing integration test does not exercise a real HNS bucket, so I do not think we yet have enough evidence to introduce a new core SPI and a second server-side storage mutation path.

For this to be mergeable, I think we need:

1. Real HNS integration evidence covering initial metadata creation, partitioned writes, and custom metadata/data locations.
2. The exact minimum folder permissions and CAB resource conditions derived from that test.
3. Removal of server-side pre-creation unless the test proves it is required for correctness; if retained as an optimization, explicit failure, configuration, and orphan-folder semantics.
4. Exact preservation of GCS path semantics, including consecutive slashes.
5. A consistent HNS-detection failure contract and documentation of any new service-account permissions.
6. PR description, docs, and the related dev-list discussion updated to match the final design, followed by a normal CI run on the resulting head.

Once the underlying shape is settled, the next review should be much more focused.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -677,6 +706,7 @@ public LoadTableResponse createTableStaged(
 
     if (baseCatalog instanceof LocalIcebergCatalog polarisCatalog) {
       polarisCatalog.validateStagedTableCreate(ident, metadata);
+      prepareStorageForTable(ident, metadata);
```
flyrain, line 709: Is it possible to let engines handle this, given that the normal table creation was done by engines I assume?

PR #3997, opened 2026-03-14
IRC: Optimize Credential Vending 

singhpk234:
### About the change 

  The /credentials endpoint currently performs a full table metadata read from object storage to vend credentials.
  This is expensive and unnecessary — the table locations needed for credential vending are already available in the
  entity's internal properties stored in the metastore.

  This PR introduces an optimized credential vending path via loadCredentials that reads locations directly from the
  metastore, avoiding the object storage round-trip. When the entity lacks the required internal properties (e.g. tables
  created before the property was stored), it gracefully falls back to the existing loadTableWithAccessDelegation path.

  The optimization is restricted to native Polaris catalogs only — federated/external catalogs fall back to the standard
  path since their entity internal properties may not be in sync with the remote catalog's actual table metadata.

  Key changes:
  - New loadCredentials method in IcebergCatalogHandler that reads locations from entity internal properties
  - Falls back to full loadTableWithAccessDelegation when: entity lacks location properties, or catalog is external/federated
  - Fails fast with NoSuchTableException if the entity is not found in the metastore
  - Authorization tests include both positive and negative privilege cases

  co-author : @adutra

## Checklist
- [ ] 🧾 Updated `CHANGELOG.md` (if needed) --later

### DIFF
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
@@ -624,8 +624,8 @@ public abstract class AbstractIcebergCatalogHandlerAuthzTest extends PolarisAuth
             CATALOG_NAME, CATALOG_ROLE1, PolarisPrivilege.TABLE_WRITE_DATA));
 
     // Wrap the real factory to spy on the catalog it creates
-    CallContextCatalogFactory spyFactory =
-        manifest -> Mockito.spy(callContextCatalogFactory.createCallContextCatalog(manifest));
+    LocalCatalogFactory spyFactory =
+        manifest -> Mockito.spy(localCatalogFactory.createCatalog(manifest));
     IcebergCatalogHandler handler = newHandler(Set.of(), CATALOG_NAME, spyFactory);
 
     // Call the optimized credential vending path — authorizeLoadTable inside
@@ -647,7 +647,7 @@ public abstract class AbstractIcebergCatalogHandlerAuthzTest extends PolarisAuth
 
     // Provide a factory that returns a non-IcebergCatalog to simulate an external catalog
     Catalog mockExternalCatalog = Mockito.mock(Catalog.class);
-    CallContextCatalogFactory externalFactory = manifest -> mockExternalCatalog;
+    LocalCatalogFactory externalFactory = manifest -> mockExternalCatalog;
     IcebergCatalogHandler handler = newHandler(Set.of(), CATALOG_NAME, externalFactory);
 
     // The optimized path should detect the non-IcebergCatalog and fall back,


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -512,15 +513,20 @@ public Response loadCredentials(
...
     Namespace ns = decodeNamespace(namespace);
     TableIdentifier tableIdentifier = TableIdentifier.of(ns, RESTUtil.decodeString(table));
+    Optional<String> refreshEndpoint =
+        Optional.of(new PolarisResourcePaths(prefix).credentialsPath(tableIdentifier));
     return withCatalog(
         securityContext,
         prefix,
         catalog -> {
+          // Vend credentials directly from entity properties, avoiding a full table metadata read
+          if (realmConfig.getConfig(FeatureConfiguration.OPTIMIZED_CREDENTIAL_VENDING)) {
```
flyingImer: IIUC, `realmConfig.getConfig(OPTIMIZED_CREDENTIAL_VENDING)` here resolves against an empty catalog-properties map (`RealmConfigImpl.getConfig(config)` -> `getConfig(config, emptyMap)`), so the `catalogConfig` key `polaris.config.optimized-credential-vending` never participates in this check.

That means the two new `@CatalogConfig` tests below are effectively exercising the same branch: both still see the default `false`, so `testLoadCredentialsEndpointWithOptimizedCredentialVending()` never proves that the optimized path is actually entered.

Would it make sense to resolve this flag with the catalog entity, consistent with the other catalog-scoped feature checks in `IcebergCatalogHandler`? Otherwise this lands the code path and docs, but the flag read and its coverage are both silently disconnected from the catalog-level config surface.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -740,6 +738,71 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
         refreshCredentialsEndpoint);
   }
 
+  /**
+   * Vend credentials for a table using location data from entity internal properties, avoiding a
+   * full table metadata read from object storage. Falls back to the standard
+   * loadTableWithAccessDelegation path if the entity lacks the required properties.
+   */
+  public ImmutableLoadCredentialsResponse loadCredentialsFromEntityProperties(
```
travis-bowen: Do we need to also check the validateRemoteTableLocations when reading from props?

This seems important since the allowed locations can occur outside of a table write where the location is currently checked.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
travis-bowen: can we pass the catalog so it matches the definition of this being defined per catalog?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -64,6 +64,18 @@ public static void enforceFeatureEnabledOrThrow(
...
+  public static final FeatureConfiguration<Boolean> OPTIMIZED_CREDENTIAL_VENDING =
+      PolarisConfiguration.<Boolean>builder()
+          .key("OPTIMIZED_CREDENTIAL_VENDING")
+          .catalogConfig("polaris.config.optimized-credential-vending")
+          .description(
+              "When enabled, the loadCredentials endpoint vends storage credentials using "
+                  + "location data from entity internal properties, avoiding a full table metadata "
+                  + "read from object storage. Falls back to the standard path if the entity lacks "
+                  + "the required properties.")
+          .defaultValue(false)
```
dimas-b: nit: I'd default to `true` ... Are there any adverse effects with that? In case of (unexpected) bugs, people can revert manually.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
```diff
@@ -597,6 +555,16 @@ Stream<DynamicNode> testLoadTableWithReadAccessDelegationPrivileges() {
...
+  }
+
+  @TestFactory
+  Stream<DynamicNode> testLoadCredentialsFromEntityPropertiesPrivileges() {
+    return authzTestsBuilder("loadCredentialsFromEntityProperties")
+        .action(
+            () -> newHandler().loadCredentialsFromEntityProperties(TABLE_NS1A_2, Optional.empty()))
+        .shouldPassWith(PolarisPrivilege.TABLE_READ_DATA)
+        .shouldPassWith(PolarisPrivilege.TABLE_WRITE_DATA)
         .shouldPassWith(PolarisPrivilege.CATALOG_MANAGE_CONTENT)
```
dimas-b, line 614: Should we add some negative cases too?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
flyrain: I’d prefer not to introduce a flag. The logic should be robust enough to handle different scenarios, regardless of whether locations are cached in the catalog entity. It should fall back to metadata.json if not in the catalog entity.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
```diff
@@ -806,6 +807,37 @@ public void testLoadTableWithAccessDelegationForExternalCatalogWithConfigEnabled
...
   }
 
+  @CatalogConfig(properties = {"polaris.config.optimized-credential-vending", "false"})
+  @Test
+  public void testLoadCredentialsEndpoint() {
+    doTestLoadCredentialsEndpoint();
+  }
+
+  @CatalogConfig(properties = {"polaris.config.optimized-credential-vending", "true"})
+  @Test
```
flyrain: We don't need these annotation once we remove the flag.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
flyrain: Since the method `loadCredentialsFromEntityProperties()` handles the fallback already, the condition on a flag seems redundant. 

Instead, I'd keep this file untouched and put the new logic to the original method `loadTableWithAccessDelegation()`.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -740,6 +738,71 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
...
+          .addKeyValue("tableIdentifier", tableIdentifier)
+          .log(
+              "Entity missing location in internal properties, requires backfill "
+                  + "as it was likely not updated with stored property changes. "
+                  + "Falling back to full loadTable path");
+      return fallbackToFullLoadTable(tableIdentifier, refreshCredentialsEndpoint);
+    }
+
+    Set<String> tableLocations =
+        StorageUtil.getLocationsUsedByTable(baseLocation, internalProperties);
```
flyrain, line 793: We need a strong guarantee at table property write time that all relevant locations are persisted, including base location, `write.metadata.path` and `write.data.path`. I assume this is already handled, but any future changes could silently break this guarantee. It would be helpful to add a test case to validate this behavior at write time.

flyrain: It would be nice to cache the full table metadata. That would significantly reduce the chance of inconsistency.

flyrain:
Thanks a lot for adding this, @singhpk234! I think it will skip the IO for the credential endpoint, which saves a lot, in terms of cost and time. Great idea!

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
singhpk234: Totally agree as part of adding these props to entity, i already a test case which asserts post write the props are there https://github.com/apache/polaris/pull/3226/changes#diff-cdc22279a7436a32d96de60cd410f08d953be2a9e287d81a4fcd5db524f5b864R2309

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
singhpk234: I wonder if we should even allow RemoteCatalog aka External Catalogs ? external writes can happen independently of the external catalog without going through Polaris and our entities might not be consistent ?

I added a check for now to only do this for catlaog owned by Polaris

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
singhpk234: remove this config entirely :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapter.java
```diff
@@ -512,21 +511,15 @@ public Response loadCredentials(
...
-                  Optional.of(new PolarisResourcePaths(prefix).credentialsPath(tableIdentifier)));
-          return Response.ok(
-                  ImmutableLoadCredentialsResponse.builder()
-                      .credentials(loadTableResponse.credentials())
-                      .build())
-              .build();
-        });
+        catalog ->
+            Response.ok(
+                    catalog.loadCredentialsFromEntityProperties(tableIdentifier, refreshEndpoint))
```
flyrain: I like the new changes here. I think the method name isn't accurate as it could fall back to table metadata.json. Rename it to something like `loadCredentials()`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -740,6 +738,82 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
...
+
+    IcebergTableLikeEntity entity = getTableEntity(tableIdentifier);
+    if (entity == null) {
+      LOGGER
+          .atDebug()
+          .addKeyValue("tableIdentifier", tableIdentifier)
+          .log(
+              "Entity not found in metastore (external catalog?), "
+                  + "falling back to full loadTable path");
+      return fallbackToFullLoadTable(tableIdentifier, refreshCredentialsEndpoint);
```
flyrain: Should we fail fast here? If entity could NOT be found in the metadata store, it doesn't seem necessary to fall back to full load. WDYT?

singhpk234: Agree better to fail fast !

dimas-b:
@singhpk234 : please update PR description to match the latest state of the code :)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -740,6 +739,76 @@ public Optional<LoadTableResponse> loadTableWithAccessDelegationIfStale(
...
+        ImmutableLoadCredentialsResponse.builder();
+
+    if (!credentialConfig.isEmpty()) {
+      responseBuilder.addCredentials(
+          ImmutableCredential.builder().prefix(baseLocation).config(credentialConfig).build());
+    } else {
+      Boolean skipCredIndirection =
+          realmConfig().getConfig(FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
+      Preconditions.checkArgument(
+          !storageAccessConfig.supportsCredentialVending() || skipCredIndirection,
```
dimas-b, line 813: We should probably make this check when we exit via line 790 too (needs a refactoring)

dimas-b (commented):
Sorry, one more last minute comment 😅

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: I propose to just use an empty `storageAccessConfig` on line 790 rather than `return`.

singhpk234: That makes sense :) done !

dimas-b (approved):

singhpk234:
Thanks for review @dimas-b @flyrain !

Issue #3779, opened 2026-02-15
Tracking Issue: Refactor `PolarisAuthorizer` SPI to skip native Principal and RBAC resolution requirement in Polaris


sungwy:
### Overview
  This issue tracks the multi‑phase refactor of the PolarisAuthorizer SPI described in [RFC](https://docs.google.com/document/d/1OaiQG_C4-yUe0ihaDBxtw_mEcOOzUBnWPazzVbjQi5U/edit?tab=t.0#heading=h.dyow25dt9w1). The goal of this refactoring is three folds:
- move resolution from Handler (like `CatalogHandler` or `PolarisAdminService`) call sites to `PolarisAuthorizer` to enable `PolarisAuthorizer` implementations to decide what to resolve (Principal, RBAC, entities, paths, etc)
- update Authorization SPI to conform to broadly accepted standard authorization intent as inputs to enhance its flexibility in introducing integrations to more authorization standards as an interface
- Introduce a breaking change in `OpaPolarisAuthorizer`'s behavior to remove dependency on Polaris native Principal and RBAC.

### Phases

- Phase 1 - New `PolarisAuthorizer` SPI + selective resolution API (not called by any call sites): https://github.com/apache/polaris/pull/3760
- Phase 1.5:
  - standardize `getResolvedPath` lookup keys: https://github.com/apache/polaris/pull/3940
  - decouple rbac privileges and semantics from `PolarisAuthorizableOperations`: https://github.com/apache/polaris/pull/4006
  - refactor `PolarisSecurable` with `PathSegment`: https://github.com/apache/polaris/pull/4016
- Phase 2 - Implement new SPI in `PolarisAuthorizerImpl` (not called by any call sites)
- Phase 3 - Implement new SPI in`OpaPolarisAuthorizer` (not called by any call sites) - Complete
- Phase 4 - Handler migration (resolution before authorization) - Complete
- Phase 5 - Handler migration (`authorizeOrThrow`): https://github.com/apache/polaris/pull/5194/

SBylemans:
Hey, what's the status of the RFC and implementation wise?

sungwy:
Hi @SBylemans - sorry for the delayed response. We've completed implementing the changes in `polaris-core`, and in `runtime`.

That means that if you bring your own `PolarisAuthorizer` implementation that skips resolving RBAC in `resolveAuthorizationInputs`, then you will be able to avoid having to sync principals into `Polaris`'s internal RBAC system.

[...]

PR #3928, opened 2026-03-04
Implement Apache Ranger Authorization Plugin for Apache Polaris (#274)

sneethiraj:
RFC :  https://docs.google.com/document/d/10UIpPMeWVU3VA0goGz_y8OAbXhIDigah/edit?usp=sharing&ouid=103452742845206345322&rtpof=true&sd=true

This RFC proposes integrating Apache Ranger as an external authorization plugin for Apache Polaris by implementing the PolarisAuthorizer SPI. By leveraging Ranger’s mature ecosystem, Polaris can provide enterprises with centralized policy management, fine-grained resource-based access control (RBAC/ABAC), and comprehensive audit logging without replacing its internal authorization model. This integration addresses key enterprise challenges by providing a uniform, centralized security and governance platform for policy management and data access auditing.

## Checklist
- [X] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [X] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [X] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [X] 💡 Added comments for complex logic
- [X] 🧾 Updated `CHANGELOG.md` (if needed)
- [X] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #3999, opened 2026-03-15
core: implement new authorize SPI in `PolarisAuthorizerImpl` and `OpaPolarisAuthorizer`

sungwy:
This PR implements phases 2 and 3 of the authorization refactor tracking issue. (https://github.com/apache/polaris/issues/3779)

It introduces request-based authorization with `AuthorizationRequest`, in both `PolarisAuthorizerImpl` and `OpaPolarisAuthorizer`. The new SPI method `authorize` is still not called by Polaris endpoint handlers. That will be done in phase 4 and 5 of the authorization refactoring effort. Hence, no behavior changes are expected in existing endpoint handling, except for `ROOT` handling in `OpaPolarisAuthorizer` noted below.

In order to decouple RBAC semantics from Polaris's core, we move RBAC operation semantics like _Privileges_ and newly introduced _Scope_ semantic into `RbacOperationSemantics`. The PR also adds `FullyQualifiedPath` as a lexical path with parent representation that `OpaPolarisAuthorizer` can use to construct the OPA request input payload.

### Planned Breaking change
OPA input payloads created via `authorize` no longer include `ROOT` in resource parent chains. This is intentional, since `ROOT` is an internal Polaris RBAC semantic and should not be part of the external OPA request contract. This PR does not yet enable this breaking change, but a future PR to make use of `authorize` instead of `authorizeOrThrow` at Handler callsites will surface this change.

## Checklist

### DIFF
--- a/extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
+++ b/extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
@@ -126,7 +126,8 @@ class OpaPolarisAuthorizer implements PolarisAuthorizer {
                 toResourceEntitiesFromSecurables(request.getSecondaries())));
     return allowed
         ? AuthorizationDecision.allow()
-        : AuthorizationDecision.deny("OPA denied authorization");
+        : AuthorizationDecision.deny(
+            "OPA denied authorization for " + request.formatForAuthorizationMessage());
   }
 
   /**
@@ -183,7 +184,9 @@ class OpaPolarisAuthorizer implements PolarisAuthorizer {
                 toResourceEntitiesFromResolvedPaths(targets),
                 toResourceEntitiesFromResolvedPaths(secondaries)));
     if (!allowed) {
-      throw new ForbiddenException("OPA denied authorization");
+      throw new ForbiddenException(
+          "OPA denied authorization for operation=%s principal=%s targets=%s secondaries=%s",
+          authzOp, polarisPrincipal.getName(), targets, secondaries);
     }
   }
 
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
@@ -21,6 +21,7 @@ package org.apache.polaris.core.auth;
 import jakarta.annotation.Nonnull;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.immutables.PolarisImmutable;
 import org.immutables.value.Value;
@@ -82,6 +83,27 @@ public interface AuthorizationRequest {
         .map(AuthorizationTargetBinding::getSecondary)
         .filter(Objects::nonNull)
         .toList();
+  }
+
+  /**
+   * Returns a stable debug string for authorization messages.
+   *
+   * <p>Includes the operation, principal name, formatted targets, and formatted secondaries.
+   */
+  @Nonnull
+  default String formatForAuthorizationMessage() {
+    return String.format(
+        "operation=%s principal=%s targets=%s secondaries=%s",
+        getOperation(),
+        getPrincipal().getName(),
+        formatSecurables(getTargets()),
+        formatSecurables(getSecondaries()));
+  }
+
+  private static String formatSecurables(List<PolarisSecurable> securables) {
+    return securables.stream()
+        .map(PolarisSecurable::formatForAuthorizationMessage)
+        .collect(Collectors.joining(", ", "[", "]"));
   }
 
   default boolean hasSecurableType(PolarisEntityType... types) {
--- a/polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisAuthorizerImplTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisAuthorizerImplTest.java
@@ -22,6 +22,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
@@ -31,6 +32,7 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.iceberg.exceptions.ForbiddenException;
 import org.apache.polaris.core.config.RealmConfig;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.core.entity.PolarisPrivilege;
@@ -54,6 +56,27 @@ public class PolarisAuthorizerImplTest {
             ? PolarisAuthorizerImpl.SUPER_PRIVILEGES.get(privilege)
             : EnumSet.of(privilege);
     assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test
+  void resolveAuthorizationInputsResolvesAll() {
+    PolarisAuthorizerImpl authorizer = new PolarisAuthorizerImpl(mock(RealmConfig.class));
+    AuthorizationState authzState = new AuthorizationState();
+    PolarisResolutionManifest manifest = mock(PolarisResolutionManifest.class);
+    AuthorizationRequest request =
+        AuthorizationRequest.of(
+            PolarisPrincipal.of("alice", Map.of(), Set.of("role")),
+            PolarisAuthorizableOperation.GET_CATALOG,
+            List.of(
+                AuthorizationTargetBinding.of(
+                    PolarisSecurable.of(new PathSegment(PolarisEntityType.CATALOG, "catalog")),
+                    null)));
+
+    authzState.setResolutionManifest(manifest);
+
+    authorizer.resolveAuthorizationInputs(authzState, request);
+
+    verify(manifest).resolveAll();
   }
 
   @Test
@@ -187,5 +210,40 @@ public class PolarisAuthorizerImplTest {
             eq(PolarisAuthorizableOperation.LIST_NAMESPACES),
             eq(List.of(namespaceWrapper)),
             eq(null));
+  }
+
+  @Test
+  void authorizeReturnsDenyDecision() {
+    PolarisAuthorizerImpl authorizer = spy(new PolarisAuthorizerImpl(mock(RealmConfig.class)));
+    AuthorizationState authzState = new AuthorizationState();
+    PolarisResolutionManifest manifest = mock(PolarisResolutionManifest.class);
+    PolarisResolvedPathWrapper catalogWrapper = mock(PolarisResolvedPathWrapper.class);
+
+    authzState.setResolutionManifest(manifest);
+    when(manifest.getResolvedTopLevelEntity("catalog", PolarisEntityType.CATALOG))
+        .thenReturn(catalogWrapper);
+    when(manifest.getAllActivatedCatalogRoleAndPrincipalRoles()).thenReturn(Set.of());
+    doThrow(new ForbiddenException("missing privilege"))
+        .when(authorizer)
+        .authorizeOrThrow(
+            any(PolarisPrincipal.class),
+            ArgumentMatchers.any(),
+            eq(PolarisAuthorizableOperation.GET_CATALOG),
+            ArgumentMatchers.any(),
+            ArgumentMatchers.<List<PolarisResolvedPathWrapper>>any());
+
+    AuthorizationRequest request =
+        AuthorizationRequest.of(
+            PolarisPrincipal.of("alice", Map.of(), Set.of("role")),
+            PolarisAuthorizableOperation.GET_CATALOG,
+            List.of(
+                AuthorizationTargetBinding.of(
+                    PolarisSecurable.of(new PathSegment(PolarisEntityType.CATALOG, "catalog")),
+                    null)));
+
+    AuthorizationDecision decision = authorizer.authorize(authzState, request);
+
+    assertThat(decision.isAllowed()).isFalse();
+    assertThat(decision.getMessage()).hasValue("missing privilege");
   }
 }
--- a/extensions/auth/opa/impl/src/test/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizerTest.java
+++ b/extensions/auth/opa/impl/src/test/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizerTest.java
@@ -662,7 +662,15 @@ public class OpaPolarisAuthorizerTest {
     AuthorizationDecision decision = authorizer.authorize(authzState, request);
 
     assertThat(decision.isAllowed()).isFalse();
-    assertThat(decision.getMessage()).hasValue("OPA denied authorization");
+    assertThat(decision.getMessage())
+        .hasValueSatisfying(
+            message ->
+                assertThat(message)
+                    .contains("OPA denied authorization")
+                    .contains("operation=GET_CATALOG")
+                    .contains("principal=alice")
+                    .contains("targets=[CATALOG:catalog-1]")
+                    .contains("secondaries=[]"));
   }
 
   @Test
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
@@ -21,6 +21,7 @@ package org.apache.polaris.core.auth;
 import com.google.common.base.Preconditions;
 import jakarta.annotation.Nonnull;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.polaris.core.entity.PolarisEntityType;
 import org.apache.polaris.immutables.PolarisImmutable;
 import org.immutables.value.Value;
@@ -68,6 +69,19 @@ public interface PolarisSecurable {
     Preconditions.checkState(
         !pathSegments.isEmpty(), "PathSegments must contain at least one segment");
     return pathSegments.subList(0, pathSegments.size() - 1);
+  }
+
+  /**
+   * Returns a stable debug string for authorization messages.
+   *
+   * <p>For example, a table securable may render as {@code
+   * CATALOG:catalog1.NAMESPACE:ns1.TABLE_LIKE:table1}.
+   */
+  @Nonnull
+  default String formatForAuthorizationMessage() {
+    return getPathSegments().stream()
+        .map(segment -> segment.entityType() + ":" + segment.name())
+        .collect(Collectors.joining("."));
   }
 
   @Value.Check


polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.common.base.Preconditions;
+import jakarta.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.polaris.core.entity.PolarisEntityType;
+import org.apache.polaris.core.persistence.PolarisResolvedPathWrapper;
+import org.apache.polaris.core.persistence.ResolvedPolarisEntity;
+
+/** Lexical fully qualified resource path split into leaf and ordered parent segments. */
+public record FullyQualifiedPath(PathSegment leaf, List<PathSegment> parents) {
```
sungwy: Note to reviewer: this class was needed to absorb the logic required in converting a `PolarisSecurable` to a fully qualified path representation that included the correct parents and their respective `PolarisEntityType` that could be constructed and inserted into the OPA request payload.

I'm on the fence as to whether we'd want this to be the representation in `AuthorizationRequest` instead of `PolarisSecurable`. Leaning towards keeping them separate for now.

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
```diff
@@ -0,0 +1,111 @@
...
+    PathSegment leaf =
+        new PathSegment(resolvedLeaf.getEntity().getType(), resolvedLeaf.getEntity().getName());
+    List<PathSegment> parents = new ArrayList<>();
+    List<ResolvedPolarisEntity> resolvedParents = wrapper.getResolvedParentPath();
+    if (resolvedParents != null) {
+      for (ResolvedPolarisEntity resolvedParent : resolvedParents) {
+        // Omit ROOT from fully qualified paths. It is an internal RBAC-only ancestor
+        if (resolvedParent.getEntity().getType() == PolarisEntityType.ROOT) {
+          continue;
+        }
```
sungwy: Note to reviewer: important behavior change to remove `ROOT` semantics from OPA request payload, since this is an RBAC-only semantic

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
sungwy, line 1: Note to reviewer: large refactoring here to move `PolarisPrivilege` semantics out of `PolarisAuthorizableOperation`, and into a completely separate module specific to RBAC instead. See `RbacOperationSemantics.java`

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +755,75 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
+  }
+
+  private List<PolarisResolvedPathWrapper> getResolvedSecurables(
+      PolarisResolutionManifest resolutionManifest,
+      RbacOperationSemantics semantics,
+      List<PolarisSecurable> securables,
+      TargetType targetType) {
+    boolean prependRootContainer =
+        switch (targetType) {
+          case TARGET -> semantics.targetScope() == PathEvaluationScope.ROOT;
```
jbonofre: I'm a bit confused here. Every operation uses `PathEvaluationScope.ROOT` except the policy attach/detach operations which use `CATALOG`. 

Is this correct for catalog-scoped operations like `LIST_NAMESPACES`, `LIST_TABLES`, ...

These operations semantically happen within a catalog context. If the existing `authorizeOrThrow` callers already prepend the root container themselves, then `ROOT` make sense for backward compatibility. Maybe worth to add a comment here to clarify.

sungwy: Hi @jbonofre - thanks for the early feedback on the PR :)

I was confused initially as well, so your confusion isn't unfounded. the `ROOT` scope here refers to whether `ROOT` container should be prepended to the resolved path, and hence whether the `ROOT` container resource should be included for privilege evaluation. 

For example, `PolarisAuthorizableOperation.LIST_NAMESPACES` prepends root container:
https://github.com/apache/polaris/blob/00061ff0dc44570eba8d3ce72835ab66758baa7f/runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java#L135

`PolarisAuthorizableOperation.GET_APPLICABLE_POLICIES_ON_CATALOG` sets prependRootContainer to `false`.

https://github.com/apache/polaris/blob/00061ff0dc44570eba8d3ce72835ab66758baa7f/runtime/service/src/main/java/org/apache/polaris/service/catalog/policy/PolicyCatalogHandler.java#L195

I think the name `PathEvaluationScope` may be misleading. I'll think of a better name for this enum

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
adutra: But this class is in polaris-core, which tends to indicate that it's not only for the OPA authorizer. Having both models in polaris-core seems a bit overkill.

This model is imho richer than `PolarisSecurable`. I wonder if you couldn't merge both models together, e.g.:

```java
@PolarisImmutable
public interface PolarisSecurable {

  /** Returns the path segments that identify the securable, in hierarchical order. */
  @Nonnull
  List<PathSegment> getPathSegments();

  default PathSegment getLeafPathSegment() {
    return getPathSegments().getLast();
  }

  @Value.Check
  default void validate() {
    Preconditions.checkState(!getPathSegments().isEmpty(), "path segments must not be empty");
  }
}
```

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
adutra: Ack, but we are introducing a specific behavior in polaris-core. Shouldn't we instead keep the `ROOT` entity here, and let specific authorizers filter it out? I feel like the `ROOT` case deserves some more thinking. It's also troubling that the current code sometimes prepends it, and sometimes not.

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -56,6 +67,13 @@ static AuthorizationRequest of(
   @Nonnull
   List<AuthorizationTargetBinding> getTargetBindings();
 
+  /** Returns the request reference catalog name, if provided by the caller. */
+  @Nullable
+  @Value.Default
+  default String getReferenceCatalogName() {
+    return null;
+  }
```
adutra: This should be enough:

```suggestion
  @Nullable
  String getReferenceCatalogName();
```

But how about using an `Optional` here?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
adutra: Nice!

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
```diff
@@ -0,0 +1,713 @@
...
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+/** Polaris RBAC-specific interpretation of an authorizable operation. */
+record RbacOperationSemantics(
+    EnumSet<PolarisPrivilege> targetPrivileges,
+    EnumSet<PolarisPrivilege> secondaryPrivileges,
+    ResolvedPathRooting targetScope,
+    ResolvedPathRooting secondaryScope) {
+
+  private RbacOperationSemantics(
```
adutra: Not sure we need this constructor, why don't you update `RBAC_SEMANTICS_BY_OPERATION` using `EnumSet` as below:

```java
 RBAC_SEMANTICS_BY_OPERATION.put(
        PolarisAuthorizableOperation.LIST_NAMESPACES,
        new RbacOperationSemantics(EnumSet.of(NAMESPACE_LIST), null, ResolvedPathRooting.ROOT, null));
```

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
```diff
@@ -0,0 +1,713 @@
...
+import java.util.EnumSet;
+import java.util.List;
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+/** Polaris RBAC-specific interpretation of an authorizable operation. */
+record RbacOperationSemantics(
+    EnumSet<PolarisPrivilege> targetPrivileges,
+    EnumSet<PolarisPrivilege> secondaryPrivileges,
+    ResolvedPathRooting targetScope,
+    ResolvedPathRooting secondaryScope) {
```
adutra: `secondaryScope` is either null (90% of cases), or identical to `targetScope`. Can we use one single `ResolvedPathRooting` maybe?

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
sungwy: > Having both models in polaris-core seems a bit overkill.

I thought about this as well, I'm currently leaning towards keeping it in polaris-core because new PolarisAuthorizer implementations like the Ranger Authorizer may also need this.

> This model is imho richer than PolarisSecurable. I wonder if you couldn't merge both models together, e.g.:

I'm open to this idea. In order to correctly construct the parent though, we'd need to pass the `referenceCatalogName` to an arg referenceCatalogName arg here like

```
default PathSegment getParents(@Nullable String referenceCatalogName) {
    ...
}
```
Would that make sense?

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
sungwy: I fully agree. 

```
public static FullyQualifiedPath of(PolarisResolvedPathWrapper wrapper)
```

This is only a temporary constructor introduced for compatibility, and will be removed from `FullyQualifiedPath` in the future. Which means `ROOT` semantics will no longer be in this module.

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
sungwy: It was more for convenience / beautification. Do you think using EnumSet directly is better? I'm okay either way

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
dimas-b: +1 to this sub-section of the current PR. Does it depend on other changes a lot? I think it might be preferable to do this part in a dedicated PR. This way, OPA changes will hopefully be isolated from the internal RBAC changes in that commit (or at least less intertwined) WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
dimas-b: nit: swap arguments maybe? In JRE's `Path` parents go first 🤔

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
sungwy: Honestly, I had that thought as well. I'm happy to do it in a dedicated PR...

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
sungwy: Yeah that's a great insight, but I also didn't want to make assumptions so I went with a safer approach of keeping them separate (feels more logical to do so)

On second thought, maybe we keep them as the same for now, since this is an internally used SPI, maybe it's okay to change it to decouple them in the future if we need to

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
dimas-b: I tend to think `PolarisSecurable` must know its catalog name anyway. Permission checks in Polaris are always performed within the context of a catalog 🤔 WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizableOperation.java
dimas-b: sweet, let's merge that first :)

polaris-core/src/main/java/org/apache/polaris/core/auth/FullyQualifiedPath.java
sungwy: Yeah that's a good point.

I'm starting to think that the shape of `FullyQualifiedPath` is more in line with what we'd want to have in place of `PolarisSecurable` in the `AuthorizationRequest`, instead of having it represented as a `PolarisEntityType` and a List of String as its name.

Having clear PathSegment with ordered parents and reference catalog is a better representation of the resource... What do you folks think about just replacing `PolarisSecurable` with `FullyQualifiedPath`'s shape instead?

dimas-b: > What do you folks think about just replacing PolarisSecurable with FullyQualifiedPath's shape instead?

Works for me :+1:

jbonofre: It sounds good yeah.

sungwy: Good point, but `parents` could be an empty list though. Feels counterintuitive to have an empty list as the first argument as well

dimas-b: How about we only have one `List<PathSegment>` in this record and add convenience methods for creating child paths based on an existing parent path?

sungwy: That sounds great actually

adutra: Would a linked-list-like approach be more appropriate? E.g.

```java
public record PathSegment(PolarisEntityType entityType, String name, @Nullable PathSegment parent) {

  public PathSegment(PolarisEntityType entityType, String name) {
    this(entityType, name, null);
  }

  public String fullyQualifiedName() {
    return parent == null ? name : parent.fullyQualifiedName() + "/" + name;
  }
}
```

Then we might not need `FullyQualifiedPath` at all (but path traversals may be less efficient).

dimas-b: We can probably change recursion to iteration if it becomes critical.

dimas-b: I'm fine with either approach. 

Minor comment: if we go with linked lists, I think it might be preferable to name the object something like `EntityPath` rather than `PathSegment` because it contains full path data down to the root.

flyrain:
Hi @sungwy , is this PR ready for review?

sungwy:
Hi @flyrain - not yet, I will need to update the branch with some recent changes.

I've been at KubeCon this week, I'll get back to it soon :)

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -75,18 +78,15 @@ default List<PolarisSecurable> getTargets() {
...
-    for (AuthorizationTargetBinding targetBinding : getTargetBindings()) {
-      if (targetBinding.getSecondary() != null) {
-        secondaries.add(targetBinding.getSecondary());
-      }
-    }
-    return secondaries;
+    return getTargetBindings().stream()
+        .map(AuthorizationTargetBinding::getSecondary)
+        .filter(Objects::nonNull)
+        .toList();
```
sungwy, line 85: beautification / consistency with above

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
```diff
@@ -18,23 +18,22 @@
...
-import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import org.apache.polaris.immutables.PolarisImmutable;
 
 /** A resource binding containing a primary target and optional secondary. */
 @PolarisImmutable
 public interface AuthorizationTargetBinding {
   static AuthorizationTargetBinding of(
-      @Nonnull PolarisSecurable target, @Nullable PolarisSecurable secondary) {
+      @Nullable PolarisSecurable target, @Nullable PolarisSecurable secondary) {
```
sungwy, line 30: updated to support operations like `ADD_ROOT_GRANT_TO_PRINCIPAL_ROLE` that expect null target (ROOT) and a valid target (principal role) in its TargetBinding.

I wonder if we could have chosen a single target binding (_PRINCIPAL_ROLE_A_, _null_) instead of (_ROOT_, _PRINCIPAL_ROLE_A_) for this type of operation instead.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,84 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
+    RbacOperationSemantics semantics = RbacOperationSemantics.forOperation(request.getOperation());
+    boolean prependRootContainer = semantics.rooting() == ResolvedPathRooting.ROOT;
+    try {
+      List<PolarisSecurable> targets = request.getTargets();
+      List<PolarisResolvedPathWrapper> resolvedTargets =
+          !targets.isEmpty()
+              ? getResolvedSecurables(resolutionManifest, targets, prependRootContainer)
+              : prependRootContainer
+                  ? List.of(resolutionManifest.getResolvedRootContainerEntityAsPath())
+                  : null;
```
sungwy: I am not a fan of this fallback logic to replace empty target with ROOT. I think we should really review what `prependingRootContainer` boolean means, and review whether we can safely combine that semantic with whether ROOT container should be the top level container for all resource targets (target & secondary) when evaluating a specific operation.

The semantic as it is asks two separate questions:
1. should we prepend Root container when resolving path from the resolution manifest?
2. should we have ROOT container in place of an empty target for this operation?

Maintaining these two is complicated, and we should know if the overhead of managing them as separate semantics is absolutely necessary.

sungwy: cc @dennishuo @dimas-b

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
```diff
@@ -245,108 +259,118 @@ private boolean queryOpaCheckResponse(ClassicHttpResponse response) throws IOExc
...
+    List<ResourceEntity> parents = new ArrayList<>();
+    List<ResolvedPolarisEntity> resolvedParents = path.getResolvedParentPath();
+    if (resolvedParents != null) {
+      for (ResolvedPolarisEntity resolvedParent : resolvedParents) {
+        parents.add(
+            ImmutableResourceEntity.builder()
+                .type(resolvedParent.getEntity().getType().name())
+                .name(resolvedParent.getEntity().getName())
+                .build());
       }
```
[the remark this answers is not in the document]
sungwy: this is only the case in the legacy API and is the current behavior.

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
```diff
@@ -158,9 +175,15 @@ public void authorizeOrThrow(
...
+    boolean allowed =
+        queryOpa(
+            buildOpaAuthorizationInput(
+                polarisPrincipal,
+                authzOp,
+                toResourceEntitiesFromResolvedPaths(targets),
+                toResourceEntitiesFromResolvedPaths(secondaries)));
     if (!allowed) {
-      throw new ForbiddenException("OPA denied authorization");
+      throw new ForbiddenException("%s", "OPA denied authorization");
```
dimas-b: nit: why such an odd message construct? 🤔

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
dimas-b: Reversing those arguments sounds reasonable to me, but it might be too late 🤔 

I wonder about @dennishuo 's view on this.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,84 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
+              : getResolvedSecurables(resolutionManifest, secondaries, prependRootContainer);
+      authorizeOrThrow(
+          request.getPrincipal(),
+          resolutionManifest.getAllActivatedCatalogRoleAndPrincipalRoles(),
+          request.getOperation(),
+          resolvedTargets,
+          resolvedSecondaries);
+      return AuthorizationDecision.allow();
+    } catch (ForbiddenException e) {
+      return AuthorizationDecision.deny(e.getMessage());
```
dimas-b, line 795: nit: Ideally we should have the exception-throwing method delegate to the non-exception method (less runtime overhead), but I guess we can do that optimization later.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
dimas-b: I'll defer to @dennishuo 's opinion on this.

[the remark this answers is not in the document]
dimas-b: Is it a valid concern? @sungwy , WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
sungwy: If it's too late, totally fine to move forward with this shape, and make target in AuthorizationTargetBinding nullable as well

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
sungwy: great eye - let me fix this :)

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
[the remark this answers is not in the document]
sungwy: It is indeed a valid concern, and I think a cleaner semantic would have been for us to have a metadata on the operation that specifies if target resource binding is expected for a given PolarisAuthorizableOperation.

That's harder to achieve in the current state where root-grant add and revoke operations have a primary target that is essentially empty (root container primary target). I've made note of this on another comment regarding the ADD_ROOT_GRANT_TO_PRINCIPAL_ROLE.

Given that it's hard to introduce a clear semantic at PolarisAuthorizableOperation to validate the AuthorizationTargetBinding values based on the operation, I think IllegalStateException is an acceptable fallback exception type (we shouldn't expect it to happen in any case)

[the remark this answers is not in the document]
dimas-b: Sounds good to me :+1:

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
dimas-b, line 795: nit: maybe add a DEBUG log for the exception (just in case someone needs it for trouble-shooting)?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,84 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
       @Nonnull AuthorizationState authzState, @Nonnull AuthorizationRequest request) {
-    throw new UnsupportedOperationException(
-        "authorize is not implemented yet for PolarisAuthorizerImpl");
+    PolarisResolutionManifest resolutionManifest = authzState.getResolutionManifest();
+    RbacOperationSemantics semantics = RbacOperationSemantics.forOperation(request.getOperation());
+    boolean prependRootContainer = semantics.rooting() == ResolvedPathRooting.ROOT;
+    try {
+      List<PolarisSecurable> targets = request.getTargets();
+      List<PolarisResolvedPathWrapper> resolvedTargets =
+          !targets.isEmpty()
```
dimas-b: nit: negation of the ternary operator's condition makes the larger expression a bit hard to follow

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
```diff
@@ -18,23 +18,22 @@
...
     return ImmutableAuthorizationTargetBinding.builder()
         .target(target)
         .secondary(secondary)
         .build();
   }
 
-  /** Returns the primary target securable for the binding. */
-  @Nonnull
+  /** Returns the primary target securable for the binding, if any. */
+  @Nullable
```
adutra, line 38: Is `AuthorizationTargetBinding.of(null, null)` a valid call? If not, I'd suggest adding a check method.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,84 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
+            })
+        .toList();
+  }
+
+  private PolarisResolvedPathWrapper getResolvedSecurable(
+      PolarisResolutionManifest resolutionManifest,
+      PolarisSecurable securable,
+      boolean prependRootContainer) {
+    if (securable.getLeaf().entityType().isTopLevel()) {
+      return resolutionManifest.getResolvedTopLevelEntity(
```
adutra, line 825:
```suggestion
      // Ignore prependRootContainer for top-level entities.
      return resolutionManifest.getResolvedTopLevelEntity(
```

(just for clarity)

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,84 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
+      List<PolarisResolvedPathWrapper> resolvedTargets =
+          !targets.isEmpty()
+              ? getResolvedSecurables(resolutionManifest, targets, prependRootContainer)
+              : prependRootContainer
+                  ? List.of(resolutionManifest.getResolvedRootContainerEntityAsPath())
+                  : null;
+      List<PolarisSecurable> secondaries = request.getSecondaries();
+      List<PolarisResolvedPathWrapper> resolvedSecondaries =
+          semantics.secondaryPrivileges().isEmpty() || secondaries.isEmpty()
+              ? null
```
adutra, line 778: So, there seems to be a disymmetry between targets and secondaries: we never prepend the ROOT container for secondaries. Is this on purpose or just mirroring the previous status quo?

polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisAuthorizerImplTest.java
```diff
@@ -39,4 +55,140 @@ void subsumingPrivilegesOf(PolarisPrivilege privilege) {
...
+    PolarisResolutionManifest manifest = mock(PolarisResolutionManifest.class);
+    PolarisResolvedPathWrapper rootWrapper = mock(PolarisResolvedPathWrapper.class);
+    PolarisResolvedPathWrapper principalRoleWrapper = mock(PolarisResolvedPathWrapper.class);
+
+    authzState.setResolutionManifest(manifest);
+    when(manifest.getResolvedRootContainerEntityAsPath()).thenReturn(rootWrapper);
+    when(manifest.getResolvedTopLevelEntity("analytics-admin", PolarisEntityType.PRINCIPAL_ROLE))
+        .thenReturn(principalRoleWrapper);
+    when(manifest.getAllActivatedCatalogRoleAndPrincipalRoles())
+        .thenReturn(Set.<PolarisBaseEntity>of());
```
adutra: Nit:
```suggestion
        .thenReturn(Set.of());
```
Same for other occurrences below.

polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisAuthorizerImplTest.java
```diff
@@ -39,4 +55,140 @@ void subsumingPrivilegesOf(PolarisPrivilege privilege) {
...
+    doNothing()
+        .when(authorizer)
+        .authorizeOrThrow(
+            any(PolarisPrincipal.class),
+            org.mockito.ArgumentMatchers.<Set<PolarisBaseEntity>>any(),
+            eq(PolarisAuthorizableOperation.ADD_ROOT_GRANT_TO_PRINCIPAL_ROLE),
+            org.mockito.ArgumentMatchers.<List<PolarisResolvedPathWrapper>>any(),
+            org.mockito.ArgumentMatchers.<List<PolarisResolvedPathWrapper>>any());
```
adutra: Nit:
```suggestion
    doNothing()
        .when(authorizer)
        .authorizeOrThrow(
            any(PolarisPrincipal.class),
            ArgumentMatchers.any(),
            eq(PolarisAuthorizableOperation.ADD_ROOT_GRANT_TO_PRINCIPAL_ROLE),
            ArgumentMatchers.any(),
            ArgumentMatchers.<List<PolarisResolvedPathWrapper>>any());
```
Same for other occurrences below.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
sungwy: This behavior mirrors the previous status quo. ROOT container is only added to replace an empty target in primary target, but not in secondary (otherwise, we'd always have ROOT container in secondaries wherever `semantics.rooting() == ResolvedPathRooting.ROOT`

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationTargetBinding.java
sungwy: That's a good idea. Thanks @adutra

polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisAuthorizerImplTest.java
sungwy: sure - that's a lot cleaner

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
dimas-b: I think it is reasonable to change `PolarisAuthorizerImpl` internals as little as possible as we go through the SPI refactoring. So current `ROOT` processing looks good from my POV.

dimas-b (approved):

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
```diff
@@ -98,18 +101,32 @@ public OpaPolarisAuthorizer(
...
+    boolean allowed =
+        queryOpa(
+            buildOpaAuthorizationInput(
+                request.getPrincipal(),
+                request.getOperation(),
+                toResourceEntitiesFromSecurables(request.getTargets()),
+                toResourceEntitiesFromSecurables(request.getSecondaries())));
+    return allowed
+        ? AuthorizationDecision.allow()
+        : AuthorizationDecision.deny("OPA denied authorization");
```
flyrain: Question(not a blocker): are we going to provide more detailed message for the deny case? My understanding of having a dedicated class `AuthorizationDecision` is to allow more information passed through instead of less.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,95 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
   @Override
   public void resolveAuthorizationInputs(
       @Nonnull AuthorizationState authzState, @Nonnull AuthorizationRequest request) {
-    throw new UnsupportedOperationException(
-        "resolveAuthorizationInputs is not implemented yet for PolarisAuthorizerImpl");
+    PolarisResolutionManifest resolutionManifest = authzState.getResolutionManifest();
```
flyrain, line 753: There seems be no test for this new logic. Could you add a few?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisAuthorizerImpl.java
```diff
@@ -747,15 +750,95 @@ public PolarisAuthorizerImpl(RealmConfig realmConfig) {
...
+    resolutionManifest.resolveAll();
   }
 
   @Override
+  @Nonnull
   public AuthorizationDecision authorize(
```
flyrain, line 759: Could you add tests for the deny case?
```
  AuthorizationDecision decision = authorizer.authorize(authzState, request);

  assertThat(decision.isAllowed()).isFalse();
```

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaPolarisAuthorizer.java
sungwy: yes, I was just mimicking the status quo, but I think it's a good idea to at least string format the `AuthorizationRequest` input. In the future, I think we could also enhance this to include the status response of OPA request as well (e.g. `200` & `false` vs `403` vs `401`)

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -75,18 +79,36 @@ default List<PolarisSecurable> getTargets() {
...
+
+  /**
+   * Returns a stable debug string for authorization messages.
+   *
+   * <p>Includes the operation, principal name, formatted targets, and formatted secondaries.
+   */
+  @Nonnull
+  default String formatForAuthorizationMessage() {
+    return String.format(
+        "operation=%s principal=%s targets=%s secondaries=%s",
```
dimas-b, line 96: This is used in _denied_ responses 🤔 I'm inclined to be cautious about including a lot of information in them to avoid accidental exposure to untrustworthy clients. I believe it would be preferable to log details for the admin user to read later and send a UUID for the logged details to the client.

That said, I'll leave the decision to you @sungwy . In any case this kind of usability adjustments might be best done in a separate PR to avoid mixing pure refactoring related to SPI changes and functional changes.

dimas-b: Moreover, I'd like to standardize AuthZ error reporting across all authorizers to avoid exposing the configured authZ to clients too (future work).

dimas-b: Cf. https://github.com/apache/polaris/pull/3928#discussion_r2983548645

sungwy: That's a great point. I'm just thinking loudly, but wouldn't a denied user already know what targets, and secondaries they are trying to access, and what operation, and principal they are using? So I feel that this is rather safe information to expose to the user - even if we decide to log the `AuthorizationDecision` message in the handlers in the future.

sungwy: I also think it's likely that this will be logged in the Polaris logs, but excluded in the Polaris API response sent back to the user. In that case, that feels even more safe to keep this in the `AuthorizationDecision` message, as it'll be up to the system integrator to decide whether those logs should be exposed to the end user.

I appreciate your suggestion to merge in this PR and make adjustments in the followup PRs. I'll add an item in the Polaris Authorization Sync to discuss this as well.

flyrain: > this will be logged in the Polaris logs, but excluded in the Polaris API response sent back to the user. 

Agreed. We should be careful about the exception message. For example, https://github.com/polaris-catalog/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/exception/IcebergExceptionMapper.java#L109-L109, it sends back as part of response. Otherwise, we should be good.

dimas-b: > [...] wouldn't a denied user already know what targets, and secondaries they are trying to access, and what operation, and principal they are using?

Partially. My concern is not about about legitimate clients, but about malicious clients. I suppose the message could be used to deduce the existence of certain entities.

Re: principal, if the auth token is stolen, the malicious client will not know the principal name, but may discover it via this message.

Re: operation, I'm thinking that exposing it to clients is more like exposing internal impl. details. They make sense to the admin user / developer, but if clients start relying on observing that data in responses (and they will) it will complicate future internal refactorings.

It's not a critical issue, though, IMHO, just trying to be proactive in tightening security :)

dimas-b: All in all, I do not wish to block this PR, but I'd be more comfortable if we did not have this level of message details in this PR (essentially technical SPI change), but improved this area in a separate PR dedicated to AuthZ usability.

sungwy: I forgot about the fact that the token is a part of the `PolarisPrincipal` - I hope we can revise that at some point.

I'll move forward with this PR, but maybe we could discuss this idea more thoroughly on the PR where we switch the handler call sites to use `authorize` instead of `authorizeOrThrow` - thanks for the thoughtful discussion @dimas-b

dimas-b: +1 to @flyrain message (the link did not work for me, BTW), `IcebergExceptionMapper` does appear to pass the exception message to clients verbatim.

PR #4003, opened 2026-03-16
CLI: Add summarize subcommand

MonkeyCanCode:
Per [ML](https://lists.apache.org/thread/35zzzh2jgorhx7q2xksp7rwxnt6gl2zx), this PR adds some basic summary/statistics for Polaris entities which can make the CLI more useful for day to day operations. As agreed on the ML, we will skip anything that involves FileIO to keep this lightweight. We can come back later to add more interesting stats if desired.

For this PR, I added `summarize` for `catalogs`, `catalog-roles`, `principals`, `principal-roles`, and `namespaces` (`policies` and `privileges` classes don't have `summarize` subcommand as they doesn't make much sense to add it there as it is covered by other entities already from this PR or lack off a quick way to avoid traversal all entities such as policies on affected tables etc.).

Setup:
```
./polaris --profile dev setup apply site/content/guides/assets/polaris/reference-setup-config.yaml
```

Sample output - catalogs:
```
➜  polaris git:(cli_summary_subcommand) ./polaris --profile dev catalogs summarize quickstart_catalog
Catalog: quickstart_catalog
Metadata
  Type:                          INTERNAL
  Storage Type:                  FILE
  Base Location:                 file:///var/tmp/quickstart_catalog/
  Allowed Locations:             file:///var/tmp/quickstart_catalog/
  Created:                       2026-03-16 02:08:03 UTC
  Modified:                      2026-03-16 02:08:03 UTC
  Version:                       1
Inventory
  Namespaces:                    2
  Tables:                        0
  Views:                         0
Access Control
  Catalog Roles:                 4
  Assigned Principal Roles:      4
Applicable Policies
  - orphan-file-policy
```

Sample output - catalog-roles:
```
➜  polaris git:(cli_summary_subcommand) ./polaris --profile dev catalog-roles summarize --catalog quickstart_catalog dev_catalog_role
Catalog Role: dev_catalog_role [Catalog: quickstart_catalog]
Metadata
  Created:                       2026-03-16 02:08:03 UTC
  Modified:                      2026-03-16 02:08:03 UTC
  Version:                       1
Assignments
  Assigned Principal Roles:      1
  Total Grants:                  3
Privileges
  - TABLE_FULL_METADATA on dev_namespace
  - TABLE_READ_DATA on dev_namespace
  - TABLE_WRITE_DATA on dev_namespace
```

Sample output - principals:
```
➜  polaris git:(cli_summary_subcommand) ./polaris --profile dev principals summarize dev_user
Principal: dev_user
Metadata
  Client ID:                     25a74cefb3da5fdb
  Created:                       2026-03-16 02:08:03 UTC
  Modified:                      2026-03-16 02:08:03 UTC
  Version:                       1
Assigned Roles
  - dev_user_role
Accessible Catalogs
  - quickstart_catalog
```

Sample output - principal-roles:
```
➜  polaris git:(cli_summary_subcommand) ./polaris --profile dev principal-roles summarize dev_user_role
Principal Role: dev_user_role
Metadata
  Created:                       2026-03-16 02:08:03 UTC
  Modified:                      2026-03-16 02:08:03 UTC
  Version:                       1
Assigned Principals
  - dev_user
Assigned Catalog Roles
  - dev_catalog_role (Catalog: quickstart_catalog)
```

Sample output - namespaces:
```
➜  polaris git:(cli_summary_subcommand) ./polaris --profile dev namespaces summarize --catalog quickstart_catalog dev_namespace.sub_namespace
Namespace: dev_namespace.sub_namespace
Metadata
  Level:                         2
  Parent:                        dev_namespace
  Sub-namespaces:                0
  Tables:                        0
  Views:                         0
Effective Policies
  - orphan-file-policy (Inherited from dev_namespace)
  - snapshot-expiry-policy (Inherited from dev_namespace)
```

## Checklist

### DIFF
--- a/site/content/in-dev/unreleased/command-line-interface.md
+++ b/site/content/in-dev/unreleased/command-line-interface.md
@@ -688,7 +688,7 @@ options:
 ##### Examples
 
 ```
-polaris principal-roles summarize some_catalog
+polaris principal-roles summarize data_engineer
 ```
 
 ### Catalog Roles
@@ -1007,7 +1007,7 @@ options:
 ##### Examples
 
 ```
-polaris namespaces stats --catalog my_catalog a.b
+polaris namespaces summarize --catalog my_catalog a.b
 ```
 
 ### Privileges


dimas-b (dismissed):
LGTM 👍 Thanks, @MonkeyCanCode !

jbonofre (dismissed):
LGTM, thanks @MonkeyCanCode !

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -268,6 +269,24 @@ polaris catalogs list
 polaris catalogs list --principal-role some_user
 ```
 
+#### summary
+
+The `summary` subcommand is used to display summary for a catalog.
```
flyrain: Should we use a verb here? like `summarize` or `sum` for short.

flyrain: Minor:  What's the different between `get` a Polaris entity and `summarize` one? Does it mean that `summarize` provides more informations? For example, when we get a catalog, we never get its inventory and applied policies, etc. Also maybe we should move this close to the `get`, so that people can understand both better.

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -268,6 +269,24 @@ polaris catalogs list
...
+options:
+  summary
+    Positional arguments:
+      catalog
+```
+
+##### Examples
+
+```
+polaris catalogs stats some_catalog
```
flyrain: stats -> summary?

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -268,6 +269,24 @@ polaris catalogs list
...
+#### summary
+
+The `summary` subcommand is used to display summary for a catalog.
+
+```
+input: polaris catalogs summary --help
+options:
+  summary
+    Positional arguments:
+      catalog
```
flyrain, line 281: I'd suggest to provide all fields it supports ATM, including inventory etc.

site/content/in-dev/unreleased/command-line-interface.md
```diff
@@ -469,6 +488,24 @@ polaris principals reset --new-client-secret ${NEW_CLIENT_SECRET} some_user
...
+#### summary
+
+The `summary` subcommand is used to display summary for a principal.
+
+```
+input: polaris principals summary --help
+options:
+  summary
+    Positional arguments:
+      principal
```
flyrain, line 500: same here for adding supported fields

flyrain:
Some idea to throw: do we think about summarize a Polaris entity by just providing its identifier. It could be useful when we got an identifier but not knowing its type(table/view/policy). Not a blocker.

site/content/in-dev/unreleased/command-line-interface.md
MonkeyCanCode: +1 for using `summarize` over current `summary`

IMO, the current `get` only display the fact/metadata about a given resource (taking `catalogs get xxxx` as an example, it will get us the catalog properties for the requested catalog but not provide us high level metrics/stats such as num of entities within this catalog). So the idea here is to give a high level view of a given resources and its reference via `summarize`. All of those are achieve-able via REST or using SQL via compute engines, but the idea here is why do i need to get to a diff tool or write another one-off script if the CLI can provide us those important info already.

The ones I have current is still  basic as phrase one, but I think we can extend this and provide more quick insights when people are doing admin work on check the inventory and relation info among entities within Polaris.

Among those summarize sub-commands, I think the only one which doesn't do those enhancement is `policies` as those are pretty much the same as the one we have in `get`. This is because there is no api for us to collect those stats without traversal through all ns, table, and views (which can be expensive but something that we can add if desired). Maybe we should drop the `summarize` for `policies`? (already removed for now, maybe we can add more interesting stuff on phase two instead to avoid the confusion with standard `get`).

I am not sure if we should merge this with `get` as get should be getting me the info of a given entity instead of showing me all of the relations for a given entity.  What do you think?

site/content/in-dev/unreleased/command-line-interface.md
MonkeyCanCode: Good catch. Yes, I started with `stats` then changed to `summary`. Updated to `summarize` now.

site/content/in-dev/unreleased/command-line-interface.md
MonkeyCanCode: Do you mind be more specific? Currently this is only for summarize a target catalog, which fields are you referring to here?

site/content/in-dev/unreleased/command-line-interface.md
MonkeyCanCode: Same as above. I would need a bit more clarification.

MonkeyCanCode:
> Some idea to throw: do we think about summarize a Polaris entity by just providing its identifier. It could be useful when we got an identifier but not knowing its type(table/view/policy). Not a blocker.

That is a great idea. Finding out what an entity is can be challenge especially when there are multiple of them using same name but different entity type. It would be nice to have an option to do "polaris summary IDENTIFIER" then we perform some sort of fuzzy search on this then display all related info for that given IDENTIFIER (e.g. an end-user may have principal name be the same as principal role name when doing one to one mapping and got a request to find out all related entities to a IDENTIFIER such as `sales`). However, the current `summarize` in this PR is a sub-command so user will have to know what kind the resource is as that will be from the main command.

To support this, we can introduce `summarize` as a main command then have option of `get|find|lookup` for a given entity based on fuzzy IDENTIFIER, once a fuzzy match is found, it should then display the result of the `summarize` sub-command for all matched. The alternative route which is also support is perform a setup export then grep for the IDENTIFIER (one reason why I was pushing on the setup load/export functionality earlier)

I have couple more things I would like to add into this in phase two. So I can for sure pick this up along with that.

site/content/in-dev/unreleased/command-line-interface.md
flyrain: One way to merge with `get` is to add another parameter to extend its default capability, like `get --extended`. I don't see a big different between the extra parameter and a new command(`summarize`). I'm fine with either way. I'm slightly leaning to a separate command, as it tends to do much more than just get the entity itself.

site/content/in-dev/unreleased/command-line-interface.md
flyrain: I thought it's an output which carries more fields. This is just input, which is fine. Sorry for the confusion.
```
➜  polaris git:(cli_summary_subcommand) ./polaris --profile dev catalogs summarize quickstart_catalog
Catalog: quickstart_catalog
Metadata
  Type:                          INTERNAL
  Storage Type:                  FILE
  Base Location:                 file:///var/tmp/quickstart_catalog/
  Allowed Locations:             file:///var/tmp/quickstart_catalog/
  Created:                       2026-03-16 02:08:03 UTC
  Modified:                      2026-03-16 02:08:03 UTC
  Version:                       1
Inventory
  Namespaces:                    2
  Tables:                        0
  Views:                         0
Access Control
  Catalog Roles:                 4
  Assigned Principal Roles:      4
Applicable Policies
  - orphan-file-policy
```

flyrain:
> That is a great idea. Finding out what an entity is can be challenge especially when there are multiple of them using same name but different entity type. It would be nice to have an option to do "polaris summary IDENTIFIER" then we perform some sort of fuzzy search on this then display all related info for that given IDENTIFIER (e.g. an end-user may have principal name be the same as principal role name when doing one to one mapping and got a request to find out all related entities to a IDENTIFIER such as `sales`). However, the current `summarize` in this PR is a sub-command so user will have to know what kind the resource is as that will be from the main command.
> 
> To support this, we can introduce `summarize` as a main command then have option of `get|find|lookup` for a given entity based on fuzzy IDENTIFIER, once a fuzzy match is found, it should then display the result of the `summarize` sub-command for all matched. The alternative route which is also support is perform a setup export then grep for the IDENTIFIER (one reason why I was pushing on the setup load/export functionality earlier)
> 
> I have couple more things I would like to add into this in phase two. So I can for sure pick this up along with that.

Phase two SGTM. We may need to define a "Polaris identifier", which somehow express all nodes within one layer, so that we could narrow down search space. For example, `c1.ns1.table1`, `c1.ns1.view1`, `c1.ns1.policy1` should all belongs to `c1.ns1`. Nested namespaces is another complexity we have to deal with, but I didn't see a big issue with it.

site/content/in-dev/unreleased/command-line-interface.md
MonkeyCanCode: Resolved this as the other one is resolved as well.

MonkeyCanCode:
> "## TL;DR\n- ✅ Added `summarize` subcommand coverage for `catalogs`, `catalog-roles`, `principals`, `principal-roles`, and `namespaces` is a useful CLI usability improvement.\n- ✅ Implementation style is consistent with existing command patterns and includes meaningful formatting helper (`format_timestamp`) and policy/count summaries.\n\n## Findings\n\n1) (docs) `principal-roles` summarize example uses the wrong positional argument name\n- Location: `site/content/in-dev/unreleased/command-line-interface.md` (Principal Roles -> summarize examples)\n- Current text: `polaris principal-roles summarize some_catalog`\n- Impact: Misleading documentation; readers may pass a catalog name where principal role name is required.\n- Suggested fix: use a principal role name, e.g. `polaris principal-roles summarize data_analyst_role`.\n\n2) (docs) `namespaces` summarize example uses a non-existent subcommand `stats`\n- Location: `site/content/in-dev/un-release/.../command-line-interface.md` (Namespaces -> summarize examples)\n- Current text: `polaris namespaces stats --catalog my_catalog a.b`\n- Impact: Confusing/incorrect command in docs; this subcommand is newly added by the PR as `summarize`.\n- Suggested fix: `polaris namespaces summarize --catalog my_catalog a.b`\n\n\nOverall this looks good and can be approved once these doc examples are corrected.\n"

Seems to be some review generated by AI tool. But it appears I may had used wrong values in the sample doc. Let me review those later tonight then update the PR.

site/content/in-dev/unreleased/command-line-interface.md
MonkeyCanCode: 100%. Lets proceed with sub-command if no other concern.

MonkeyCanCode:
> > That is a great idea. Finding out what an entity is can be challenge especially when there are multiple of them using same name but different entity type. It would be nice to have an option to do "polaris summary IDENTIFIER" then we perform some sort of fuzzy search on this then display all related info for that given IDENTIFIER (e.g. an end-user may have principal name be the same as principal role name when doing one to one mapping and got a request to find out all related entities to a IDENTIFIER such as `sales`). However, the current `summarize` in this PR is a sub-command so user will have to know what kind the resource is as that will be from the main command.
> > To support this, we can introduce `summarize` as a main command then have option of `get|find|lookup` for a given entity based on fuzzy IDENTIFIER, once a fuzzy match is found, it should then display the result of the `summarize` sub-command for all matched. The alternative route which is also support is perform a setup export then grep for the IDENTIFIER (one reason why I was pushing on the setup load/export functionality earlier)
> > I have couple more things I would like to add into this in phase two. So I can for sure pick this up along with that.
> 
> Phase two SGTM. We may need to definite a "Polaris identifier", which somehow express all nodes within one layer, so that we could narrow down search space. For example, `c1.ns1.table1`, `c1.ns1.view1`, `c1.ns1.policy1` should all belongs to `c1.ns1`. Nested namespaces is another complexity we have to deal with, but I didn't see a big issue with it.

Another thing I am thinking is to add some support around table summarized (e.g. view table location and schema). I am bit hesitated here as we don't hav `table` as a main-command and we will likely need to define a scope of this main-command (e.g. not to support create table etc.? as those seems to be barely useful). What I was thinking was following:
1. show me the table stats with read-only (e.g. table location, schema, iceberg version, etc. but avoid FileIO as discussed on the ML)
2. show me the effective policy on this table (similar to the one show above for namespace)
3. delete table (de-register): so this may seems dangerous to add but I think this is very handy. For example, when an end-user accidentally messed up references files for a given table on the backend storage, all compute engines (e.g. spark and trino) will prevent the delete/de-register from happening. Now assuming an end-user need to bypass this safety check, there is no way it can be done without calling the api directly. With de-register, we will only remove reference on the catalog and not touch the files on backend. Thoughts?

flyrain:
> show me the table stats with read-only (e.g. table location, schema, iceberg version, etc. but avoid FileIO as discussed on the ML)

this could be done by engines, or lib like PyIceberg. I guess it doesn't hurt to allow CLI to do so. 

> show me the effective policy on this table (similar to the one show above for namespace)

Good to have. 

> delete table (de-register): 

Not sure I understand correctly. Are you saying that engines have to load a table before deleting it? In case of table corruption, there is no way to delete a table from engines. That seems a valid use case to me.

MonkeyCanCode:
> > show me the table stats with read-only (e.g. table location, schema, iceberg version, etc. but avoid FileIO as discussed on the ML)
> 
> this could be done by engines, or lib like PyIceberg. I guess it doesn't hurt to allow CLI to do so.

that is correct, but my point is, if we can do all of these without expensive FileIO, why not implement those in Polaris CLI. With compute engines, those all work but just a lot tools switching to get some basic info.

> 
> > show me the effective policy on this table (similar to the one show above for namespace)
> 
> Good to have.
> 
> > delete table (de-register):
> 
> Not sure I understand correctly. Are you saying that engines have to load a table before deleting it? In case of table corruption, there is no way to delete a table from engines. That seems a valid use case to me.

Yes. I had seeing couple cases where someone accidentally deleted files on s3 (or bug on older version of iceberg runtime) which then caused the table to be in a corrupted state. For production, we for sure want to recover those (there are a couple of way for recovery a corrupted tables...but very manual and a lot hacky things to do). However, for non-prod, it makes sense to just de-register the table and recreate so people can be unlocked. While had to deal with this issue couple of times, I noticed both spark/trino will all try to load metadata first before allow a drop table to happen. In this case, as the table is corrupted, the metadata lookup will break thus not allow an end-user to drop. As far as I know, the only way to get this over with is calling the API to delete the data which bypass all of the checks performed by compute engines.

In case if u want to try out, create a table with our docker-compose then delete the table path, then try to drop table, it will throw error regarding manifest not found etc.

MonkeyCanCode:
> > "## TL;DR\n- ✅ Added `summarize` subcommand coverage for `catalogs`, `catalog-roles`, `principals`, `principal-roles`, and `namespaces` is a useful CLI usability improvement.\n- ✅ Implementation style is consistent with existing command patterns and includes meaningful formatting helper (`format_timestamp`) and policy/count summaries.\n\n## Findings\n\n1) (docs) `principal-roles` summarize example uses the wrong positional argument name\n- Location: `site/content/in-dev/unreleased/command-line-interface.md` (Principal Roles -> summarize examples)\n- Current text: `polaris principal-roles summarize some_catalog`\n- Impact: Misleading documentation; readers may pass a catalog name where principal role name is required.\n- Suggested fix: use a principal role name, e.g. `polaris principal-roles summarize data_analyst_role`.\n\n2) (docs) `namespaces` summarize example uses a non-existent subcommand `stats`\n- Location: `site/content/in-dev/un-release/.../command-line-interface.md` (Namespaces -> summarize examples)\n- Current text: `polaris namespaces stats --catalog my_catalog a.b`\n- Impact: Confusing/incorrect command in docs; this subcommand is newly added by the PR as `summarize`.\n- Suggested fix: `polaris namespaces summarize --catalog my_catalog a.b`\n\n\nOverall this looks good and can be approved once these doc examples are corrected.\n"
> 
> Seems to be some review generated by AI tool. But it appears I may had used wrong values in the sample doc. Let me review those later tonight then update the PR.

This had been fixed. Thanks for catching those.

MonkeyCanCode:
@jbonofre @flyrain @dimas-b do you mind take another look to see if anything else should be address by this PR? I will do a ML for the phase two with feedbacks/ideas provided by @flyrain and hopefully get more ideals from community as well on what to include.

MonkeyCanCode:
> Thanks a lot for working on it, @MonkeyCanCode . Should we add some tests to file
> 
> https://github.com/apache/polaris/blob/4543b821860b6991b547331bf0b4d0aa4a1cdf82/client/python/integration_tests/test_management_apis.py#L37
> 
> , and
> https://github.com/apache/polaris/blob/4543b821860b6991b547331bf0b4d0aa4a1cdf82/client/python/integration_tests/test_catalog_apis.py#L46

I think both tests are related to testing the APIs only. If you would like, I can for sure add the workflows used by `summarize` over to ensure we get the same result as test cases. We do have a test for `test_cli_parsing.py` if we want to add the test for cli parsing. What do you think?

Also, speaking of `test_cli_parsing.py`, it appears it is a bit outdated as it doesn't include those newly added options lately.

flyrain:
SGTM to add this to the parsing tests. We can add integration tests as a follow up. It would also be nice to file an issue, this could attract new contributors.

MonkeyCanCode:
> SGTM to add this to the parsing tests. We can add integration tests as a follow up. It would also be nice to file an issue, this could attract new contributors.

Sounds good. Let me file an issue for this for new contributors.

MonkeyCanCode:
> SGTM to add this to the parsing tests. We can add integration tests as a follow up. It would also be nice to file an issue, this could attract new contributors.

Here is the tracker: https://github.com/apache/polaris/issues/4017

dimas-b (approved):

PR #4006, opened 2026-03-17
core: decouple rbac privileges and semantics from `PolarisAuthorizableOperations`

sungwy:
Refactoring to decouple RBAC specific privilege based authorization semantics from `PolarisAuthorizableOperations`. The PR includes a unit test suite that verifies that all  `PolarisAuthorizableOperations` have their corresponding `RbacOperationSemantics` defined with a `targetPrivilege`.

No behavior change is introduced in this PR.

Note: This is an off-shoot from https://github.com/apache/polaris/pull/3999.

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
```diff
@@ -0,0 +1,644 @@
...
+    Preconditions.checkArgument(!targetPrivileges.isEmpty(), "targetPrivileges must be non-empty");
+  }
+
+  /**
+   * Determines whether RBAC should prepend the root container to the resolved path before
+   * evaluating privileges.
+   *
+   * <p>{@code ROOT} means include the root container in the resolved path. {@code CATALOG} means do
+   * not prepend the root container.
+   */
```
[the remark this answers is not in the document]
sungwy: Thanks co-pilot, but I think the risk of a system integrator writing custom code to do `semantics.targetPrivileges().clear()` is quite low.

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
```diff
@@ -0,0 +1,644 @@
...
+import java.util.EnumMap;
+import java.util.EnumSet;
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+/** Polaris RBAC-specific interpretation of an authorizable operation. */
+record RbacOperationSemantics(
+    EnumSet<PolarisPrivilege> targetPrivileges,
+    EnumSet<PolarisPrivilege> secondaryPrivileges,
+    ResolvedPathRooting rooting) {
+
```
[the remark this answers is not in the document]
sungwy: Thanks co-pilot, but I think the risk of a system integrator writing custom code to do `semantics.targetPrivileges().clear()` is quite low. Also a duplicate of below.

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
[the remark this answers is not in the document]
adutra: FYI, Guava has a `com.google.common.collect.Sets.immutableEnumSet()` method.

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
```diff
@@ -0,0 +1,644 @@
...
+   *
+   * <p>{@code ROOT} means include the root container in the resolved path. {@code CATALOG} means do
+   * not prepend the root container.
+   */
+  enum ResolvedPathRooting {
+    ROOT,
+    CATALOG
+  }
+
+  private static final EnumMap<PolarisAuthorizableOperation, RbacOperationSemantics>
```
adutra, line 137: It could be worth adding a comment here clarifying that there is no thread-safety issue in using a non-concurrent map because of safe publication via class initialization.

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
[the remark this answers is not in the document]
dimas-b: @sungwy : I agree the risk of `EnumSet` contamination at immediate getter call sites is minimal, but if the set is passed to other methods, it will not be apparent that it is not supposed to be modified.

Given current RBAC code, I suppose the level of optimization `EnumSet` offers over general (immutable) sets is rather minimal, so I'd personally go with keeping immutable views over `EnumSet` in this record. WDYT?

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
dimas-b: to be absolutely sure, I'd propose doing `RBAC_SEMANTICS_BY_OPERATION = initOpMap()`, where `initOpMap()` is equivalent to the current `static` init block, but returns the new `Map`.

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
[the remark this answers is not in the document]
sungwy: Sounds good, point taken

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
sungwy: I see the concern, but I think the current static initialization is already sufficient here. The static final EnumMap is safely published through class initialization, and after the Set.copyOf(...) change the semantics values are also immutable, so there is no concurrent mutation concern after initialization that would make this unsafe.

If we really feel a comment explaining that would help, we could add one, but I don’t think it’s necessary.

dimas-b: I'm fine with current code, without any comment about initialization.

dimas-b:
@sungwy : it looks like we got conflicts from the PRs removing unused operation values :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/auth/RbacOperationSemantics.java
```diff
@@ -0,0 +1,633 @@
...
+import java.util.EnumMap;
+import java.util.EnumSet;
+import java.util.Set;
+import org.apache.polaris.core.entity.PolarisPrivilege;
+
+/** Polaris RBAC-specific interpretation of an authorizable operation. */
+record RbacOperationSemantics(
+    Set<PolarisPrivilege> targetPrivileges,
+    Set<PolarisPrivilege> secondaryPrivileges,
+    ResolvedPathRooting rooting) {
```
dimas-b, line 115: How is `rooting` intended to be used? I do not see any readers of this value 🤔

dimas-b: Is it used in a follow-up PR?

sungwy: Yes it is intended for use in #3999

Maybe it would make sense for me to remove this in the scope of this PR. And add it in #3999 instead

dimas-b: up to you... I was just wondering about the usage pattern... I'll check the other PR.

dimas-b: I'm still a bit fuzzy about the "prepend root container" logic, but it looks like it existed since day 1, so we're not changing it in these PRs, just making inputs nicer :+1:

sungwy: yes, exactly - currently the logic is embedded into the Handler call sites and I think it should be defined here as an RBAC semantic instead.

flyrain:
Thanks a lot for the change, @sungwy ! The PR is in the right direction. Checked with @dennishuo a bit yesterday.  He is interested in taking a look. Given the impact of this refactor, can we give a bit more time for reviewers?

dimas-b:
@flyrain : this PR has been merged, but follow-up comments and changes are welcome, of course... or are you proposing to revert it?

sungwy:
Hi @flyrain, thank you for the note.

As @dimas-b mentioned, I went ahead and merged this PR after the series of reviews here and on https://github.com/apache/polaris/pull/3999. Since no use facing changes are introduced here, I felt it was reasonable to move things forward once it was approved.

That said, I’m very open to feedback. If there are concerns about the direction of the refactor, I’m happy to pause any follow-up changes while we take a closer look.

@dennishuo I would really appreciate your thoughts when you have a chance. Happy to adjust based on your feedback.

PR #4007, opened 2026-03-17
Add optional traversal limit for commit offset lookup

### LINKED ISSUE #4000: Add safeguard to limit commit history traversal in CommitsImpl when using offset
Describe the bug

CommitsImpl in the NoSQL persistence implementation traverses commit history to honor an offset without any safeguard on how much history it will walk. The code has TODOs about adding such a safeguard, but none is implemented, which can lead to unbounded work and performance / resource issues for large histories or pathological offset values.

To Reproduce

Use the NoSQL persistence implementation with a large commit log.
Call an API that ultimately uses CommitsImpl with an offset far back in history.
CommitsImpl walks the commit chain from head toward the offset with no traversal limit, making the operation potentially very expensive.

Expected Behavior

Commit history traversal for an offset should be bounded (e.g., via a configurable maximum traversal limit). If the limit is exceeded before finding the requested offset, the operation should fail fast with a clear error/status instead of continuing unbounded work.

Additional context

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
There are TODOs: // TODO add safeguard to limit the work done when finding the commit with ID 'offset'.
I’d like to work on this by adding a traversal limit for offset lookup and tests for both normal and “limit exceeded” behavior.

System information

OS: macOS
Polaris Catalog Version: current main
Object storage & setup: Not specific; this concerns the NoSQL commit traversal logic.

NssGourav:
## Summary
- Add a new `polaris.persistence.commit-offset-lookup-max-traversal` parameter (default `0` = unlimited) to optionally bound commit history traversal while resolving offsets.
- Enforce the limit in `CommitsImpl.commitLog()` and `CommitsImpl.commitLogReversed()` during offset lookup, failing fast with a clear exception when exceeded.
- Add tests covering the fail-fast behavior for both variants.

## Rationale
`CommitsImpl` currently walks commit history unbounded when resolving offsets. In pathological cases (very large histories or missing/far-back offsets) this can lead to unbounded work.

The safeguard is **opt-in** (default unlimited) to avoid breaking background/maintenance flows that may legitimately traverse the full log.

## Test plan
- `./gradlew :persistence:nosql:persistence-impl:test`

Fixes #4000

### DIFF
--- a/persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/PersistenceParams.java
+++ b/persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/PersistenceParams.java
@@ -83,6 +83,14 @@ public interface PersistenceParams {
   @WithDefault(DEFAULT_MAX_SERIALIZED_VALUE_SIZE_STRING)
   MemorySize maxSerializedValueSize();
 
+  /**
+   * Maximum number of commit-IDs to traverse while resolving an offset for commit log iteration.
+   *
+   * <p>A value of {@code 0} disables the safeguard (unlimited traversal).
+   */
+  @WithDefault("0")
+  long commitOffsetLookupMaxTraversal();
+
   @PolarisImmutable
   interface BuildablePersistenceParams extends PersistenceParams {
     static ImmutableBuildablePersistenceParams.Builder builder() {
@@ -129,6 +137,12 @@ public interface PersistenceParams {
     @Value.Default
     default MemorySize maxSerializedValueSize() {
       return DEFAULT_MAX_SERIALIZED_VALUE_SIZE;
+    }
+
+    @Override
+    @Value.Default
+    default long commitOffsetLookupMaxTraversal() {
+      return 0L;
     }
   }
 }
--- a/persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
+++ b/persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
@@ -32,6 +32,7 @@ import java.util.OptionalLong;
 import org.agrona.collections.LongArrayList;
 import org.apache.polaris.persistence.nosql.api.Persistence;
 import org.apache.polaris.persistence.nosql.api.commit.Commits;
+import org.apache.polaris.persistence.nosql.api.exceptions.CommitOffsetTraversalLimitExceededException;
 import org.apache.polaris.persistence.nosql.api.obj.BaseCommitObj;
 import org.apache.polaris.persistence.nosql.api.obj.ObjRef;
 
@@ -69,7 +70,7 @@ final class CommitsImpl implements Commits {
     }
 
     visited.add(head.id());
-    var traversed = 1L;
+    var traversed = 0L;
     var tail = head.tail();
     outer:
     while (tail.length != 0) {
@@ -77,18 +78,11 @@ final class CommitsImpl implements Commits {
         if (tailId == offset) {
           break outer;
         }
-        visited.add(tailId);
         if (maxTraversal != UNLIMITED_TRAVERSAL && ++traversed > maxTraversal) {
-          throw new IllegalStateException(
-              "Exceeded commit offset traversal limit while resolving offset "
-                  + offset
-                  + " for ref '"
-                  + refName
-                  + "'. Traversed="
-                  + traversed
-                  + ", max="
-                  + maxTraversal);
+          throw new CommitOffsetTraversalLimitExceededException(
+              refName, offset, traversed, maxTraversal);
         }
+        visited.add(tailId);
       }
 
       while (!visited.isEmpty()) {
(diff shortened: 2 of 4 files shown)


Subham-KRLX:
A few things worth addressing like the limit is still inside CommitsImpl, but @snazy mentioned the TODO was meant for the call sites (REST/service layer), not the implementation itself worth clarifying with him before going further. There's also a counter bug where commitLog and commitLogReversed count traversals differently since traversed starts at 0 in one and 1 in the other, so the same maxTraversal value behaves inconsistently. On the exception side, IllegalStateException feels off for an expected operational limit a dedicated exception would be easier to catch specifically. For the tests, the .hasNext() in the commitLog test is misleading since the exception fires during the commitLog() call itself, not lazily, and there's also no test for the success case where the offset is found within the limit. Overall the approach is reasonable, just needs a quick sync with the maintainer on the architectural question first.

NssGourav:
Thanks agreed on the feedback.

I understand the TODOs may be intended for call-site guardrails; I implemented an opt-in safeguard in `CommitsImpl` (commit-offset-lookup-max-traversal, default 0 = unlimited) to avoid impacting maintenance flows unless configured.

Fixed traversal counting so commitLog and commitLogReversed behave consistently.
Replaced IllegalStateException with CommitOffsetTraversalLimitExceededException for easier catching.

Updated tests to assert at the right point and added success path coverage.

Subham-KRLX:
@dimas-b can u please review this pr.

dimas-b:
The idea looks reasonable to me at first glance... Should the PR be non-draft for a wide review, though :wink: 

Please also rebase (just to have a recent base commit) and fix CI :)

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
```diff
@@ -55,27 +57,31 @@ public <C extends BaseCommitObj> Iterator<C> commitLogReversed(
...
     visited.add(head.id());
+    var traversed = 0L;
     var tail = head.tail();
     outer:
     while (tail.length != 0) {
       for (var tailId : tail) {
         if (tailId == offset) {
           break outer;
         }
+        if (maxTraversal != UNLIMITED_TRAVERSAL && ++traversed > maxTraversal) {
```
dimas-b: nit: please make the `++` operation more prominent (for ease of following the logic), e.g. do it on a separate line.

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
```diff
@@ -55,27 +57,31 @@ public <C extends BaseCommitObj> Iterator<C> commitLogReversed(
...
+          throw new CommitOffsetTraversalLimitExceededException(
+              refName, offset, traversed, maxTraversal);
+        }
         visited.add(tailId);
```
dimas-b, line 85: isn't `traversed` equivalent to `visited.size()`?

persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/PersistenceParams.java
```diff
@@ -83,6 +83,14 @@ default RetryConfig retryConfig() {
   @WithDefault(DEFAULT_MAX_SERIALIZED_VALUE_SIZE_STRING)
   MemorySize maxSerializedValueSize();
 
+  /**
+   * Maximum number of commit-IDs to traverse while resolving an offset for commit log iteration.
+   *
+   * <p>A value of {@code 0} disables the safeguard (unlimited traversal).
+   */
+  @WithDefault("0")
+  long commitOffsetLookupMaxTraversal();
```
dimas-b: maybe `OptionalLong`?

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
```diff
@@ -146,8 +152,7 @@ public <C extends BaseCommitObj> Iterator<C> commitLog(
 
     var head = headOpt.get();
     var type = head.type().id();
-
-    // TODO add safeguard to limit the work done when finding the commit with ID 'offset'
+    var maxTraversal = persistence.params().commitOffsetLookupMaxTraversal();
```
dimas-b: The guardrail in this case is different in nature from `commitLogReversed`. Here we do not accumulate memory, AFAIK. It might be preferable to use different config settings for these cases.

Also, `commitLogReversed` is not called in "prod" code ATM.

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/commits/CommitsImpl.java
```diff
@@ -157,6 +162,7 @@ public <C extends BaseCommitObj> Iterator<C> commitLog(
         return singletonList(head).iterator();
```
dimas-b, line 162: @snazy : is this correct? 🤔

Subham-KRLX:
@NssGourav Could you please take a look at the recent review comments from @dimas-b and apply them to the current code? Also it looks like there are a couple of CI checks failing right now that will need to be fixed. Let us know if you need any help!

dimas-b:
@NssGourav : Please rebase this PR to get recent CI job changes.

snazy:
I'm still a bit concerned about adding a global setting that trickles down to all call sites (currently only maintenance).
Maintenance has its own limit and if configured with a higher value than a global one would be confusing, technically even wrong.

NssGourav:
Thanks, that makes sense, a global traversal limit in persistence can definitely become confusing if maintenance already has a separate limit, and I agree that would be the wrong layering.

 I’ll revise the PR to avoid introducing a global setting here and keep the safeguard scoped to the specific caller/use case instead, then I’ll rerun CI and update the PR.

PR #4010, opened 2026-03-17
(Proposal) REST API for Table Metrics

obelix74:
This proposal defines a REST API for querying table scan and commit metrics in Apache Polaris.

**Split from PR #3924** as suggested by @dimas-b to allow independent progress on events and metrics proposals.

## Overview

The Table Metrics REST API provides endpoints for retrieving metrics reports that have been collected from query engines. This enables:
- Performance analysis and optimization
- Query pattern analysis
- Capacity planning
- SLA monitoring

## Key Design Decisions

- **Single unified `/metrics` endpoint** with type filter parameter (scan/commit)
- **Pagination support** with cursor-based tokens
- **Authorization** aligned with `TABLE_READ_DATA` privilege

## Related

- Events REST API proposal: PR #3924 (now Events-only)
- Discussion thread: https://lists.apache.org/thread/yh73rq4fqj1c3t40j7ltysh5f1gvg2g4

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues

### DIFF
--- a/site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
+++ b/site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
@@ -1,0 +1,720 @@
+---
+title: Table Metrics REST API
+linkTitle: Table Metrics REST API
+weight: 110
+---
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+-->
+
+# Proposal: REST API for Querying Table Metrics
+
+**Author:** Anand Sankaran
+**Date:** 2026-03-02
+**Status:** Draft Proposal
+**Target:** Apache Polaris
+
+---
+
+## Abstract
+
+This proposal defines REST API endpoints for querying table metrics (scan reports, commit reports) from Apache Polaris. The endpoints expose data already being persisted via the existing JDBC persistence model (`scan_metrics_report`, `commit_metrics_report` tables) and follow established Polaris API patterns.
+
+---
+
+## Table of Contents
+
+1. [Motivation](#1-motivation)
+2. [Use Cases](#2-use-cases)
+3. [Design Principles](#3-design-principles)
+4. [API Specification](#4-api-specification)
+5. [Authorization](#5-authorization)
+6. [OpenAPI Schema](#6-openapi-schema)
+7. [Implementation Notes](#7-implementation-notes)
+
+---
+
+## 1. Motivation
+
+Apache Polaris currently persists table metrics (scan reports, commit reports) to the database, but provides no REST API to query this data. Users must access the database directly to retrieve metrics information.
+
+Adding read-only REST endpoints enables:
+- Programmatic access to metrics without database credentials
+- Integration with monitoring dashboards and alerting systems
+- Consistent authorization via Polaris RBAC
+- Pagination and filtering without writing SQL
+
+---
+
+## 2. Use Cases
+
+### 2.1 Table Health Monitoring
+- Track write patterns: files added/removed per commit, record counts, duration trends
+- Identify tables with high commit frequency or unusually large commits
+- Detect issues indicating need for compaction (many small files) or optimization
+
+### 2.2 Query Performance Analysis
+- Understand read patterns: files scanned vs skipped, planning duration
+- Identify inefficient queries with low manifest/file pruning ratios
+- Correlate performance with filter expressions and projected columns
+
+### 2.3 Capacity Planning & Chargeback
+- Aggregate metrics by table, namespace, or principal over time
+- Track storage growth trends (`total_file_size_bytes`)
+- Attribute usage to teams/users via `principal_name`
+
+### 2.4 Debugging & Troubleshooting
+- Correlate metrics with distributed traces (`otel_trace_id`, `otel_span_id`)
+- Investigate specific commits by `snapshot_id`
+- Trace operations via `request_id`
+
+---
+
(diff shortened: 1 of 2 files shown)


site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
```diff
@@ -0,0 +1,720 @@
...
+- Attribute usage to teams/users via `principal_name`
+
+### 2.4 Debugging & Troubleshooting
+- Correlate metrics with distributed traces (`otel_trace_id`, `otel_span_id`)
+- Investigate specific commits by `snapshot_id`
+- Trace operations via `request_id`
+
+---
+
+## 3. Design Principles
```
dimas-b, line 88: I believe it would be valuable to make this feature optional at the Polaris Admin user's discretion.

Persisting scan/commits metrics in the Polaris database is ultimately and admin decision (IIRC it's already optional). If metrics are not persistent there is not reason even to make the new API available.

I believe it can be addressed via CDI / Quarkus. Two main points:
1. If the jar containing the impl. of the API is not on the Polaris server's class path at build time (meaning downstream builds), all other functionality should remain operational. The new API endpoints will naturally receive 404 responses.
2. In the default Apache Polaris distribution, add a config flag to short-circuit the new endpoints into a "do nothing" impl. (producing 404 or 501) without the extra burden of RBAC checks. This is optional.

WDYT?

site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
```diff
@@ -0,0 +1,720 @@
...
+
+---
+
+## 4. API Specification
+
+### 4.1 Endpoint Summary
+
+| Method | Path | Description |
+|--------|------|-------------|
+| GET | `/api/metrics-reports/v1/catalogs/{catalogName}/namespaces/{namespace}/tables/{table}` | List metrics for a table (type specified via query parameter) |
```
dimas-b, line 141: This path is modeled after similar IRC endpoints, which is probably going to be familiar to existing IRC users.

However, from the perspective of building a new API from scratch, is it critical to have `namespace` and `table` segments?

How about using a simple full name path to the entity? e.g. `/api/metrics-reports/v1/catalogs/{catalogName}/{path_to_table}`?

`path_to_table` would be encoded the same way as `namespace`.

Other entity types will not require separate endpoints (if they ever receive metrics).

Actual entity type can be reflected in the response payload.

WDYT?

site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
```diff
@@ -0,0 +1,720 @@
...
+
+> **Note:** The metrics API uses a dedicated `/api/metrics-reports/v1/` namespace since it exposes pre-populated records rather than managing catalog state.
+
+### 4.2 Path Parameters
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `catalogName` | string | Name of the catalog |
+| `namespace` | string | Namespace (URL-encoded, multi-level separated by `%1F`) |
```
dimas-b, line 150: The separator (encoding rules) apparently changed in Iceberg 1.11 🤔 I did not delve into details, though 😅

site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
obelix74: Updated the proposal to make this explicit. Section 3.2 now defines three levels of optionality: 

(1) classpath-based opt-out — if the metrics API module jar is  absent, endpoints return 404 naturally with no core changes needed; 
(2) a polaris.metrics-api.enabled config flag that short-circuits to 501 without RBAC overhead; 
(3) runtime  detection of a non-persisting metrics reporter, which also short-circuits to 501 with a diagnostic message rather than returning silently empty results. Deployers who don't  persist metrics incur zero cost.

site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
obelix74: Updated to  /api/metrics-reports/v1/catalogs/{catalogName}/{path_to_table} where path_to_table encodes the full entity path (namespace levels + table name) using the same unit-separator  encoding as namespace in the Iceberg REST spec. The entity type is reflected in the response payload. Updated the endpoint summary, path parameters, example requests, and  OpenAPI paths block accordingly.

site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
obelix74: Thanks for flagging. Added a note in the path_to_table parameter description: implementations should be aware of the encoding rule change in Iceberg 1.11 and handle both  conventions.

obelix74:
@flyingImer     

1. Addressed in section 3.2. The proposal now defines explicit behavior for all three deployment modes: module absent (404), config flag disabled (501), and non-persisting  reporter active (501 with diagnostic hint). The OpenAPI spec also now includes a 501 response code with a description. Silently empty responses are explicitly ruled out for the   non-persisting case.
2. Reconcile envelope design vs. concrete schemas:
Resolved by rewriting section 6.2 to use  the envelope as the canonical schema: each report item is a MetricsReportEnvelope with stable top-level fields (id, timestampMs, catalog, realm, actor, request, object) and type-specific data nested under payload.data, discriminated by payload.type + payload.version. Also collapsed the two separate response types (ListScanMetricsResponse /  ListCommitMetricsResponse) into a single ListMetricsResponse using MetricsReportEnvelope[] items — the metricType field echoes the request. Example JSON responses in section  4.4 are updated to match.

site/content/in-dev/unreleased/proposals/table-metrics-rest-api.md
dimas-b: Correction: the separator is configurable in IRC. The default is the same, though, AFAIK.

We do not have to support configurable separators in this API because Polaris does not support that in its IRC impl., though.

However, when we make actual OpenAPI spec, we probably need to review namespace encoding carefully.

dimas-b:
I put this on tomorrow's Community Sync cal agenda too.

obelix74:
> I put this on tomorrow's Community Sync cal agenda too.

Thanks Dmitry, it is doubtful I will attend tomorrow's sync, I will try my best. Too many conflicts.

dimas-b:
Unfortunately, we did not get to this item during the Community Sync :man_shrugging: Let's continue in GH and `dev` ML for now.

obelix74:
@dimas-b Implementation of this proposal in https://github.com/apache/polaris/pull/4115

obelix74:
I am closing this proposal. Implementation of this proposal is in https://github.com/apache/polaris/pull/4115. Let us continue review over that PR.

PR #4013, opened 2026-03-17
IcebergCatalog- updateNamespaceProperties on single commit

rambleraptor:
First time Polaris contributor here! I've been working mostly on the Iceberg side.

We ran the PyIceberg Catalog Tests against Polaris (https://github.com/apache/iceberg-python/pull/3106) and discovered that Polaris can't handle adding / removing namespace properties at the same time.

This is because Polaris attempts to handle this as two commits (which run into concurrency issues), when the Iceberg spec supports a single commit.

This PR adds support for creating this single commit for Iceberg Catalogs.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
@@ -239,13 +239,17 @@ public class CatalogHandlerUtils {
     Map<String, String> startProperties = catalog.loadNamespaceMetadata(namespace);
     Set<String> missing = Sets.difference(removals, startProperties.keySet());
 
-    if (!updates.isEmpty()) {
-      catalog.setProperties(namespace, updates);
-    }
+    if (catalog instanceof IcebergCatalog polarisCatalog) {
+      polarisCatalog.updateProperties(namespace, removals, updates);
+    } else {
+      if (!updates.isEmpty()) {
+        catalog.setProperties(namespace, updates);
+      }
 
-    if (!removals.isEmpty()) {
-      // remove the original set just in case there was an update just after loading properties
-      catalog.removeProperties(namespace, removals);
+      if (!removals.isEmpty()) {
+        // remove the original set just in case there was an update just after loading properties
+        catalog.removeProperties(namespace, removals);
+      }
     }
 
     return UpdateNamespacePropertiesResponse.builder()
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -735,6 +735,63 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
     return true;
   }
 
+  /**
+   * Update properties for a namespace.
+   *
+   * @param namespace the namespace to update properties for
+   * @param removals the set of properties to remove
+   * @param updates the map of properties to update
+   * @return true if the properties were updated, false otherwise
+   */
+  public boolean updateProperties(
+      Namespace namespace, Set<String> removals, Map<String, String> updates)
+      throws NoSuchNamespaceException {
+    PolarisResolvedPathWrapper resolvedEntities = resolvedEntityView.getResolvedPath(namespace);
+    if (resolvedEntities == null) {
+      throw noSuchNamespaceException(namespace);
+    }
+    PolarisEntity entity = resolvedEntities.getRawLeafEntity();
+    Map<String, String> newProperties = new HashMap<>(entity.getPropertiesAsMap());
+
+    // Merge new properties into existing map.
+    newProperties.putAll(updates);
+    // Remove properties.
+    removals.forEach(newProperties::remove);
+
+    PolarisEntity updatedEntity =
+        new PolarisEntity.Builder(entity).setProperties(newProperties).build();
+
+    if (!realmConfig.getConfig(FeatureConfiguration.ALLOW_NAMESPACE_LOCATION_OVERLAP)) {
+      LOGGER.debug("Validating no overlap with sibling tables or namespaces");
+      validateNoLocationOverlap(
+          NamespaceEntity.of(updatedEntity), resolvedEntities.getRawParentPath());
+    } else {
+      LOGGER.debug("Skipping location overlap validation for namespace '{}'", namespace);
+    }
+    if (!realmConfig.getConfig(
+        BehaviorChangeConfiguration.ALLOW_NAMESPACE_CUSTOM_LOCATION, catalogEntity)) {
+      if (updates.containsKey(PolarisEntityConstants.ENTITY_BASE_LOCATION)) {
+        validateNamespaceLocation(NamespaceEntity.of(entity), resolvedEntities);
+      }
+    }
+
+    List<PolarisEntity> parentPath = resolvedEntities.getRawFullPath();
+    PolarisEntity returnedEntity =
+        Optional.ofNullable(
+                getMetaStoreManager()
+                    .updateEntityPropertiesIfNotChanged(
+                        getCurrentPolarisContext(),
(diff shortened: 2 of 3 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -742,11 +735,67 @@ public boolean setProperties(Namespace namespace, Map<String, String> properties
...
+   *
+   * @param namespace the namespace to update properties for
+   * @param removals the set of properties to remove
+   * @param updates the map of properties to update
+   * @return true if the properties were updated, false otherwise
+   */
+  public boolean updateProperties(
+      Namespace namespace, Set<String> removals, Map<String, String> updates)
+      throws NoSuchNamespaceException {
+    PolarisResolvedPathWrapper resolvedEntities = resolvedEntityView.getResolvedPath(namespace);
```
jbonofre, line 749: Why not using `resolvedEntityView.getResolvedPath(ResolvedPathKey.ofNamespace(namespace));` like in other methods here ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
```diff
@@ -239,13 +239,17 @@ public UpdateNamespacePropertiesResponse updateNamespaceProperties(
     Map<String, String> startProperties = catalog.loadNamespaceMetadata(namespace);
     Set<String> missing = Sets.difference(removals, startProperties.keySet());
 
-    if (!updates.isEmpty()) {
-      catalog.setProperties(namespace, updates);
-    }
+    if (catalog instanceof IcebergCatalog polarisCatalog) {
```
jbonofre, line 242: Due to that we have a direct dependency with `IcebergCatalog`.
As `CatalogHandlerUtils` implements the `SupportNamespace`, it should not be "linked" to any catalog specific implementation.

I think it should be cleaner to isolate this in a specific method (for clarity).

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
```diff
@@ -239,13 +239,17 @@ public UpdateNamespacePropertiesResponse updateNamespaceProperties(
...
+      polarisCatalog.updateProperties(namespace, removals, updates);
```
nandorKollar, line 243: nit: I think if both `updates` and `removals` are empty, there's no need to call `updateProperties` no?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -742,11 +735,67 @@ public boolean setProperties(Namespace namespace, Map<String, String> properties
...
 
+  /**
+   * Update properties for a namespace.
+   *
+   * @param namespace the namespace to update properties for
+   * @param removals the set of properties to remove
+   * @param updates the map of properties to update
+   * @return true if the properties were updated, false otherwise
+   */
+  public boolean updateProperties(
```
adutra, line 746: The returned boolean is never used / checked.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -742,11 +735,67 @@ public boolean setProperties(Namespace namespace, Map<String, String> properties
...
+          NamespaceEntity.of(updatedEntity), resolvedEntities.getRawParentPath());
+    } else {
+      LOGGER.debug("Skipping location overlap validation for namespace '{}'", namespace);
+    }
+    if (!realmConfig.getConfig(
+        BehaviorChangeConfiguration.ALLOW_NAMESPACE_CUSTOM_LOCATION, catalogEntity)) {
+      if (updates.containsKey(PolarisEntityConstants.ENTITY_BASE_LOCATION)) {
+        validateNamespaceLocation(NamespaceEntity.of(entity), resolvedEntities);
+      }
+    }
```
adutra, line 776: These checks appear twice, can we extract a common utility method?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -2614,4 +2615,30 @@ public void testPaginatedListNamespaces() {
...
+    catalog.createNamespace(
+        NS,
+        ImmutableMap.of(
+            "prop1", "val1",
+            "prop2", "val2",
+            "prop3", "val3"));
+
+    catalog.updateProperties(
+        NS,
+        ImmutableSet.of("prop1", "prop2", "missing_prop"),
```
adutra, line 2630: Nit: let's keep one old prop to assert that it is preserved after the update:

```suggestion
        ImmutableSet.of("prop1", "missing_prop"),
```

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogTest.java
```diff
@@ -2614,4 +2615,30 @@ public void testPaginatedListNamespaces() {
       }
     }
   }
+
+  @Test
+  public void testUpdateNamespacePropertiesAtomic() {
```
dimas-b, line 2620: It might be preferable to promote this test to `PolarisRestCatalogIntegrationBase`, which acts almost like a TCK for Polaris. WDYT?

dimas-b, line 2620: What in this test ensures that the operation is "atomic"? 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
```diff
@@ -239,13 +239,17 @@ public UpdateNamespacePropertiesResponse updateNamespaceProperties(
     Map<String, String> startProperties = catalog.loadNamespaceMetadata(namespace);
     Set<String> missing = Sets.difference(removals, startProperties.keySet());
 
-    if (!updates.isEmpty()) {
-      catalog.setProperties(namespace, updates);
```
dimas-b, line 248: Does this mean Polaris propagates namespace property updates to Federated catalogs even if `CatalogEntity.isStaticFacade()` is `true`? @dennishuo WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
dimas-b: Yes, having `instanceof` checks in this case does not look nice from the design POV.

If we have to have custom handling for the internal catalog (which is ok), perhaps it is time to refactor this code to use proper polymorphic `catalog` objects and allow implementations to decide how to best handle updates.

Federated catalogs will have to delegate to the appropriate API.

dimas-b:
Thanks for you contribution, @rambleraptor :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CatalogHandlerUtils.java
flyrain: Agreed with @jbonofre  and @dimas-b, please consider replacing `catalog instanceof IcebergCatalog` with a dedicated capability interface.

nandorKollar:
Not stale, I think this PR is a valuable contribution.

dimas-b:
@rambleraptor : The PR does look valid. Do you have capacity to address review comments posted so far?

flyrain:
It's a valid fix. Thanks for picking this up, @rambleraptor ! Let us know if you are still working on it.

PR #4016, opened 2026-03-18
core: refactor `PolarisSecurable` with `PathSegment`

sungwy:
While working on https://github.com/apache/polaris/pull/3999 we realized that `PolarisSecurable` requires an Entity representation with its parent relationships and their corresponding `EntityType`.

This PR introduces a new record  type `PathSegment` that has `name` and `EntityType` that represents a segment of the resource path, that can be ordered to construct a `PolarisSecurable`.

This is not yet a breaking change, because a release hasn't been made since `PolarisSecurable` was introduced.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
@@ -77,11 +77,9 @@ public interface PolarisSecurable {
     Preconditions.checkState(
         getPathSegments().get(0).entityType().isTopLevel(),
         "PathSegments must start with a top-level entity");
-    for (PathSegment parent : getParents()) {
+    for (PathSegment segment : getPathSegments()) {
       Preconditions.checkState(
-          parent.entityType() != PolarisEntityType.ROOT,
-          "PathSegments must not include ROOT for securable leaf=%s",
-          getLeaf());
+          segment.entityType() != PolarisEntityType.ROOT, "PathSegments must not include ROOT");
     }
     if (getLeaf().entityType().isTopLevel()) {
       Preconditions.checkState(
--- a/polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisSecurableTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisSecurableTest.java
@@ -87,6 +87,13 @@ public class PolarisSecurableTest {
   }
 
   @Test
+  void securableRejectsRootLeaf() {
+    assertThatThrownBy(() -> PolarisSecurable.of(new PathSegment(PolarisEntityType.ROOT, "root")))
+        .isInstanceOf(IllegalStateException.class)
+        .hasMessageContaining("must not include ROOT");
+  }
+
+  @Test
   void securableRejectsInvalidInChainParentHierarchy() {
     assertThatThrownBy(
             () ->


polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,82 @@
...
+  }
+
+  @Value.Check
+  default void validate() {
+    Preconditions.checkState(getPathSegments() != null, "PathSegments must be non-null");
+    Preconditions.checkState(
+        !getPathSegments().isEmpty(), "PathSegments must contain at least one segment");
+    Preconditions.checkState(
+        getPathSegments().get(0).entityType().isTopLevel(),
+        "PathSegments must start with a top-level entity");
```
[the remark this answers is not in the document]
sungwy: thanks Copilot. I'll add null validation in `PathSegment`

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,82 @@
...
+      Preconditions.checkState(
+          getParents().isEmpty(),
+          "top-level securable leaf=%s must not declare parents",
+          getLeaf());
+    } else {
+      for (int i = 1; i < getPathSegments().size(); i++) {
+        PathSegment parent = getPathSegments().get(i - 1);
+        PathSegment child = getPathSegments().get(i);
+        Preconditions.checkState(
+            child.entityType().getParentType() == parent.entityType(),
```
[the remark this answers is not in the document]
sungwy: yep - that's a good edge case. Thanks Copilot

polaris-core/src/test/java/org/apache/polaris/core/auth/PolarisSecurableTest.java
```diff
@@ -0,0 +1,86 @@
...
+            new PathSegment(PolarisEntityType.CATALOG, "catalogA"),
+            new PathSegment(PolarisEntityType.NAMESPACE, "ns1"));
+    assertThat(table.getPathSegments())
+        .containsExactly(
+            new PathSegment(PolarisEntityType.CATALOG, "catalogA"),
+            new PathSegment(PolarisEntityType.NAMESPACE, "ns1"),
+            new PathSegment(PolarisEntityType.TABLE_LIKE, "table1"));
+  }
+
+  @Test
```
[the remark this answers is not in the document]
sungwy: adding more test coverage

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,84 @@
...
+  @Nonnull
+  List<PathSegment> getPathSegments();
+
+  /** Returns the leaf segment of the path. */
   @Nonnull
-  PolarisEntityType getEntityType();
+  @Value.Derived
+  default PathSegment getLeaf() {
+    List<PathSegment> pathSegments = getPathSegments();
+    return pathSegments.get(pathSegments.size() - 1);
```
adutra, line 60:
```suggestion
    return getPathSegments().getLast();
```

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,84 @@
...
-  List<String> getNameParts();
+  @Value.Derived
+  default List<PathSegment> getParents() {
+    List<PathSegment> pathSegments = getPathSegments();
+    return pathSegments.subList(0, pathSegments.size() - 1);
+  }
+
+  @Value.Check
+  default void validate() {
+    Preconditions.checkState(getPathSegments() != null, "PathSegments must be non-null");
```
adutra: Already validated by the immutables library:

```suggestion
```

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,84 @@
...
+    Preconditions.checkState(
+        !getPathSegments().isEmpty(), "PathSegments must contain at least one segment");
+    Preconditions.checkState(
+        getPathSegments().get(0).entityType().isTopLevel(),
+        "PathSegments must start with a top-level entity");
+    for (PathSegment parent : getParents()) {
+      Preconditions.checkState(
+          parent.entityType() != PolarisEntityType.ROOT,
+          "PathSegments must not include ROOT for securable leaf=%s",
```
adutra: I'm not following, shouldn't the highest ancestor path segments be allowed to be `ROOT` for some securables?

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,84 @@
...
+          getLeaf());
+    }
+    if (getLeaf().entityType().isTopLevel()) {
```
adutra, line 84: I think we are missing unit tests for some of the validation failures.

polaris-core/src/test/java/org/apache/polaris/core/auth/AuthorizationRequestTest.java
```diff
@@ -36,7 +37,9 @@ void hasSecurableTypeReturnsTrueForPrincipalTarget() {
             PolarisAuthorizableOperation.LOAD_TABLE,
             List.of(
                 AuthorizationTargetBinding.of(
-                    PolarisSecurable.of(PolarisEntityType.PRINCIPAL, List.of("alice")), null)));
+                    PolarisSecurable.of(
+                        List.of(new PathSegment(PolarisEntityType.PRINCIPAL, "alice"))),
```
adutra: nit:

```suggestion
                    PolarisSecurable.of(new PathSegment(PolarisEntityType.PRINCIPAL, "alice")),
```

(same for other occurrences in this class)

polaris-core/src/test/java/org/apache/polaris/core/auth/AuthorizationRequestTest.java
```diff
@@ -64,11 +70,18 @@ void hasSecurableTypeReturnsTrueForCatalogRoleAcrossMultipleBindings() {
...
+                        List.of(
+                            new PathSegment(PolarisEntityType.CATALOG, "catalog"),
+                            new PathSegment(PolarisEntityType.NAMESPACE, "ns"))),
                     null),
                 AuthorizationTargetBinding.of(
-                    PolarisSecurable.of(PolarisEntityType.CATALOG, List.of("catalog")),
-                    PolarisSecurable.of(PolarisEntityType.CATALOG_ROLE, List.of("catalog-role")))));
+                    PolarisSecurable.of(
+                        List.of(new PathSegment(PolarisEntityType.CATALOG, "catalog"))),
+                    PolarisSecurable.of(
```
adutra: Given the usage patterns, I think we could have a new method with varargs:


```java
  static PolarisSecurable of(PathSegment first, PathSegment... rest) {
    return ImmutablePolarisSecurable.builder().addPathSegment(first).addPathSegments(rest).build();
  }
```

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: I'm rooting for removing `ROOT` in the parent chain in `PolarisSecurable`, because including `ROOT` as a parent is a strange RBAC specific semantic. I don't think external PolarisAuthorizer implementations like OPA or Ranger would care about a `ROOT` parent.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: Happy to add more coverage

polaris-core/src/test/java/org/apache/polaris/core/auth/AuthorizationRequestTest.java
sungwy: I love the suggestion. Thank you @adutra

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: noted

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
dimas-b: I agree that `ROOT` should not be a factor in authZ decisions in general. Existing native RBAC uses it, which is fine... historical reasons.

It's like appending `.` to DNS names - it makes them formally absolute, but it makes no practical difference in most cases :sweat_smile: 

Let's standardize on starting paths at the catalog in AuthZ SPI.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: Looks like we are still compiling at 17 with `polaris-client`: https://github.com/apache/polaris/blob/f39a9a38c6b6d08b26f2f90713e54808f4062218/build-logic/src/main/kotlin/polaris-client.gradle.kts#L24

Maybe we can keep this PR with the current syntax and raise a DISCUSS thread to assess moving `polaris-client` compilation to 21?

dimas-b: Java 21 discussion is on-going (although slowly 😅 ): https://lists.apache.org/thread/v8ghm3k5nh9xp9x0xdo2kknk5ph0rnmb

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,87 @@
...
+  List<PathSegment> getPathSegments();
+
+  /** Returns the leaf segment of the path. */
   @Nonnull
-  PolarisEntityType getEntityType();
+  @Value.Derived
+  default PathSegment getLeaf() {
+    List<PathSegment> pathSegments = getPathSegments();
+    Preconditions.checkState(
+        !pathSegments.isEmpty(), "PathSegments must contain at least one segment");
```
dimas-b, line 59: This check is a bit pointless, given the `validate()` method 🤔

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
```diff
@@ -18,30 +18,87 @@
...
+  }
+
+  @Value.Check
+  default void validate() {
+    Preconditions.checkState(
+        !getPathSegments().isEmpty(), "PathSegments must contain at least one segment");
+    Preconditions.checkState(
+        getPathSegments().get(0).entityType().isTopLevel(),
+        "PathSegments must start with a top-level entity");
+    for (PathSegment parent : getParents()) {
```
dimas-b: Why check only parents? I guess `ROOT` is not meaningful as a leaf element either 🤔

polaris-core/src/test/java/org/apache/polaris/core/auth/AuthorizationRequestTest.java
```diff
@@ -36,7 +37,8 @@ void hasSecurableTypeReturnsTrueForPrincipalTarget() {
             PolarisAuthorizableOperation.LOAD_TABLE,
             List.of(
                 AuthorizationTargetBinding.of(
-                    PolarisSecurable.of(PolarisEntityType.PRINCIPAL, List.of("alice")), null)));
+                    PolarisSecurable.of(new PathSegment(PolarisEntityType.PRINCIPAL, "alice")),
```
dimas-b, line 40: nit: maybe add a convenience `.of()` method for this? It looks like a common pattern.

polaris-core/src/main/java/org/apache/polaris/core/auth/PathSegment.java
```diff
@@ -0,0 +1,30 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.auth;
+
+import com.google.common.base.Preconditions;
+import org.apache.polaris.core.entity.PolarisEntityType;
+
+/** One segment in a fully qualified resource path. */
+public record PathSegment(PolarisEntityType entityType, String name) {
```
dimas-b, line 25: nit: maybe add a convenience `.pathSegment()` builder method? Callers would be able to statically import it and have less verbose code.

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: yeah I guess we could expand it to the leaf as well

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
sungwy: I agree, but I thought of just keeping this to make the error more descriptive in the future, if our API evolves... it shouldn't ever happen and `PathSegments must contain at least one segment` feels more descriptive than `ArrayIndexOutOfBoundsException`.

polaris-core/src/test/java/org/apache/polaris/core/auth/AuthorizationRequestTest.java
sungwy: that's a nice neat trick :)

sungwy: hmm I thought about this a bit more, and I'm thinking of just keeping the plain constructors for now. I think it's explicit and consistent for securables with single PathSegment and multiple segments

dimas-b: ok, no worries

polaris-core/src/main/java/org/apache/polaris/core/auth/PolarisSecurable.java
dimas-b: well, I do not see how this check can ever fail 🤔 I believe the failure will happen in the `validate()` method and the object will never get constructed for this method to be called on 🤔

sungwy:
Requesting for re-review @dimas-b - I think this would be safe to merge since this is a new, unused class

dimas-b (approved):
I'm fine with merging now. If @adutra has any follow-up feedback we can always adjust this code later.

PR #4018, opened 2026-03-18
[NoSQL] Ensure index elements with non-`null` values

snazy:
Within the current implementation, it is possible that the iterators of the `Index` API return elements with `null` values. This may cause NPEs.

This change enforces that index elements returned by the `Index` API iterators can never yield `null` values. The actual fix is in `IndexSpi` using the new iterator base class. The only place where `null` values could "escape" from the index API was in the `.(reverse)iterator()` functions in `IndexSpi`. The iterator implementation has been updated to skip elements with `null` values.

While the mentione change alone would have been sufficient, it seemed better to make the null-safety guarantees visible at the API level.

Chosing `Map.Entry` as the type for index elements seemed to be logical at the time of writing. As both `Map.Entry.getKey()` and `Map.Entry.getValue()` are nullable, it makes it difficult to "see" potential non-nullable violations in the code base. To address this issue, the `Map.Entry` interface has been replaced in the API with a new `Index.Element` interface with `@Nonnull` annotated functions. The implementation internal and package-private interface `IndexElement` was renamed to `InternalIndexElement`, now implementing `Index.Element`.

Background information: When an "embedded" serialized index becomes too big, it is spilled out to separately stored index stripes. The embedded index then tracks the changes on top of the spilled out index stripes. Such a change can be an update (a different non-null value) or a remove. In case of a removal, the value becomes `null`. Such removed elements must be visible within the index implementation, but never be exposed via the API.

### DIFF
--- a/persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
+++ b/persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
@@ -34,6 +34,7 @@ abstract class AbstractIndexElement<V> implements InternalIndexElement<V> {
       return key().equals(other.key()) && Objects.equals(valueNullable(), other.valueNullable());
     }
     if (o instanceof Index.Element<?> other) {
+      // This case is just for tests, not for actual production usage.
       return key().equals(other.key()) && Objects.equals(valueNullable(), other.value());
     }
     return false;
--- a/persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/InternalIndexElement.java
+++ b/persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/InternalIndexElement.java
@@ -39,9 +39,9 @@ import org.apache.polaris.persistence.nosql.api.index.IndexValueSerializer;
  * index segments.
  *
  * <p>The only place where instances of this type can "escape" is via the {@link Iterator} returning
- * functions of the {@link Index} interface. The implementations of the {@code iterator()} and {@code
- * reverseIterator()} functions in {@link IndexSpi} uses {@link IndexElementIter} to ensure that
- * only elements with non-{@code null} values "escape".
+ * functions of the {@link Index} interface. The implementations of the {@code iterator()} and
+ * {@code reverseIterator()} functions in {@link IndexSpi} uses {@link IndexElementIter} to ensure
+ * that only elements with non-{@code null} values "escape".
  *
  * @param <V> element value type
  */


persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
```diff
@@ -0,0 +1,67 @@
...
+
+package org.apache.polaris.persistence.nosql.api.index;
+
+import jakarta.annotation.Nonnull;
+
+/**
+ * Package-private implementation of {@link Index.Element} for {@link Index.Element#of(IndexKey,
+ * Object)}.
+ */
+final class IndexElem<V> implements Index.Element<V> {
```
adutra: Any reason not to use a record?

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
```diff
@@ -20,27 +20,30 @@
...
+    }
+    if (o instanceof Index.Element<?> other) {
+      return key().equals(other.key()) && Objects.equals(valueNullable(), other.value());
+    }
+    return false;
   }
 
   @Override
   public int hashCode() {
     @Var var h = 5381;
```
adutra, line 45: nit: `IndexElem` and `AbstractIndexElement` have different hashcode algorithms. This means that two instances of these classes could be considered equal (per the method above), but have different hashcodes.

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/InternalIndexElement.java
```diff
@@ -0,0 +1,61 @@
...
+ * <p>This interface intentionally <em>breaks</em> the contract of {@link Index.Element} by allowing
+ * {@code null} values, within the indexes implementation code. This is for space and performance,
+ * aka to use the same object instance in the indexes API and the implementation.
+ *
+ * <p>Within a {@linkplain AbstractLayeredIndexImpl} index, the embedded reference index contains
+ * {@code null} values for keys that have been removed but are still present in the spilled out
+ * index segments.
+ *
+ * <p>The only place where instances of this type can "escape" is via the {@link Iterator} returning
+ * functions of the {@link Index} interface. The implemntations of the {@code iterator()} and {@code
```
adutra:
```suggestion
 * functions of the {@link Index} interface. The implementations of the {@code iterator()} and {@code
```

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/IndexElementIter.java
```diff
@@ -16,15 +16,34 @@
...
+  }
 
-  int contentSerializedSize(IndexValueSerializer<V> ser);
+  @Override
+  protected Index.Element<V> computeNext() {
+    while (delegate.hasNext()) {
+      var next = delegate.next();
+      var value = next.valueNullable();
+      if (value != null) {
+        return Index.Element.of(next.key(), value);
```
adutra: `InternalIndexElement` also implements `Index.Element`. Isn't it enough to just return `next`?

```suggestion
        return next;
```

persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
dimas-b: +1

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
dimas-b: actually that would not be a "nit". Hash codes may collide for unequal objects, but equal objects _must_ have the same hash code.

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
```diff
@@ -20,27 +20,30 @@
...
-    }
     if (o == this) {
       return true;
     }
-    return getKey().equals(other.getKey()) && Objects.equals(getValue(), other.getValue());
+    if (o instanceof InternalIndexElement<?> other) {
+      return key().equals(other.key()) && Objects.equals(valueNullable(), other.valueNullable());
+    }
+    if (o instanceof Index.Element<?> other) {
+      return key().equals(other.key()) && Objects.equals(valueNullable(), other.value());
```
dimas-b, line 38: If the comparison is reversed and `other` happens to be an `IndexElem`, then it will use `value()` from this `AbstractIndexElement`... so it looks like comparison is not commutative 🤔

dimas-b: Do we actually need polymorphic comparison for these classes?

Why not consider different sub-types as "not equal" all the time?

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractLayeredIndexImpl.java
```diff
@@ -81,10 +81,10 @@ public int estimatedSerializedSize() {
   public boolean contains(IndexKey key) {
     var u = embedded.getElement(key);
     if (u != null) {
-      return u.getValue() != null;
+      return u.valueNullable() != null;
```
dimas-b, line 84: nit: maybe add `u.hasValue()` and avoid nullable getters?

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
snazy: The comparison is commutative. Yes, the "intentionally broken contract" by `InternalIndexElement` makes it a bit difficult to reason about.

Internally, the indexes implementation works only with instances of `InternalIndexElement`, aka `AbstractIndexElement`. You cannot `put()` an `*Element` instance to the index.

With [this](https://github.com/apache/polaris/pull/4018#discussion_r2953703120), call sites can get instances of the internal representation, but as `Index.Element` and never with a `null` value.

So from a call-site the comparison is commutative.

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractLayeredIndexImpl.java
snazy: Hm, I'd like to avoid any unnecessary calls with in this performance sensitive part.

persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/Index.java
```diff
@@ -37,7 +36,7 @@
  * @see UpdatableIndex
  * @see IndexContainer
  */
-public interface Index<V> extends Iterable<Map.Entry<IndexKey, V>> {
```
flyingImer: IIUC the actual bug here is internal tombstone elements leaking through the iterator path. Why do we need to change the public iteration contract from `Map.Entry<IndexKey, V>` to `Index.Element<V>` for that? This feels like a much broader API break than the runtime fix, so my bias would be to keep the external contract stable unless we intentionally want an API migration here.

persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.errorprone.annotations.Var;
+import jakarta.annotation.Nonnull;
+
+/**
+ * Package-private implementation of {@link Index.Element} for {@link Index.Element#of(IndexKey,
+ * Object)}.
+ */
+record IndexElem<V>(IndexKey key, V value) implements Index.Element<V> {
+
+  IndexElem(@Nonnull IndexKey key, @Nonnull V value) {
```
flyingImer: If the goal is to make the non-null guarantee visible at the API level, should `Element.of()`/`IndexElem` enforce it as well? Right now the `@Nonnull` annotations are advisory only, so `new IndexElem<>(key, null)` still succeeds and `value()` can hand back `null` until we trip over it in `hashCode()`/`equals()`. Would it make sense to `requireNonNull()` both fields here and add a tiny test for that contract?

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractLayeredIndexImpl.java
dimas-b: Is `u.hasValue()` less performant than `u.valueNullable() != null`?

dimas-b: My point is that if we have `.hasValue()` perhaps `valueNullable()` will not be necessary at all? WDYT?

snazy: It's an implementation internal type - so :shrug:

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
dimas-b: Considering this test code:
```
  @Test
  void compareWithIndexElem() {
    IndexKey key = IndexKey.key("test");
    DirectIndexElement<Object> d = new DirectIndexElement<>(key, null);
    Index.Element<String> e = Index.Element.of(key, "test");
    assertThat(d.equals(e)).isFalse();
    assertThat(e.equals(d)).isFalse();
  }
```
The first assertion passes, but the second assertion throws a `NullPointerException`. This is what I mean by non-commutative comparison.

It may not be happening in runtime now, but in general, I tend to think, it would be preferable to avoid this kind of code pattern :thinking:

snazy: The test is crafted to that cannot succeed. And it's using an *internal only* type that is not exposed to any `Index` call site.

dimas-b: I played with this PR a bit (full Polaris Sever + Spark client) and I did not find any calls paths leading to comparison of `IndexElem`. Do we need these `.equals()` methods at all?

persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
dimas-b, line 31: I did not see any uses of this record in runtime (create/insert/select in Spark). Do we need this as a  top-level class?

dimas-b: How about we keep the value nullable in the concrete class and non-null in the `Index.Element.value()` interface method? 

Internal use cases can deal with the impl. class., external - with the interface. 

I hope the code can be simplified this way. WDYT?

persistence/nosql/persistence/impl/src/main/java/org/apache/polaris/persistence/nosql/impl/indexes/AbstractIndexElement.java
dimas-b: I'm fine merging "as is". Opened #4049 for follow-up.

PR #4019, opened 2026-03-18
Allow setting ARN role and external ID on catalogs that had none

adutra:
Fixes: https://github.com/apache/polaris/discussions/3885

Catalogs created without a role ARN (e.g. not using STS) could not be updated to add one, because the validation treated going from null to a non-null value as a forbidden AWS account ID change. The same applied to external IDs.

This change relaxes the validation in `validateUpdateCatalogDiffOrThrow()` so that setting a previously-null role ARN or external ID is allowed, while still preventing changes to the AWS account ID or external ID once set.

This change also changes the error message from the misleading "Cannot modify Role ARN" (the role name *can* be changed) to the more accurate "Cannot modify AWS account ID".

## Checklist

### DIFF
--- a/site/content/in-dev/unreleased/configuration/config-sections/flags-polaris_features.md
+++ b/site/content/in-dev/unreleased/configuration/config-sections/flags-polaris_features.md
@@ -157,6 +157,15 @@ If set to true, allow one table's location to reside within another table's loca
 
 ---
 
+##### `polaris.features."ALLOW_UNRESTRICTED_STORAGE_CONFIG_ROLE_CHANGES"`
+
+If set to true, allows unrestricted changes to storage configuration role properties when updating a catalog, including changing the AWS account ID, role ARN, external ID, or Azure tenant ID. When false (default), these changes are restricted: new values can be set when none were previously configured, but existing values cannot be changed to different ones.
+
+- **Type:** `Boolean`
+- **Default:** `false`
+
+---
+
 ##### `polaris.features."ALLOW_UNSTRUCTURED_TABLE_LOCATION"`
 
 If set to true, allows unstructured table locations.


runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -852,16 +852,19 @@ private void validateUpdateCatalogDiffOrThrow(
     if (currentStorageConfig instanceof AwsStorageConfigurationInfo currentAwsConfig
         && newStorageConfig instanceof AwsStorageConfigurationInfo newAwsConfig) {
 
-      if (!Objects.equals(currentAwsConfig.getAwsAccountId(), newAwsConfig.getAwsAccountId())) {
+      // Allow setting a role ARN when none was previously configured, but prevent changing the
+      // AWS account ID once it has been set.
+      if (currentAwsConfig.getAwsAccountId() != null
+          && !currentAwsConfig.getAwsAccountId().equals(newAwsConfig.getAwsAccountId())) {
```
singhpk234: This means we allow updating the role A to role B if they are same account ? any reason why

adutra: Yes. This has always been like this, presumably for security reasons (i.e. prevent a malicious client from changing to a role in another AWS account). Maybe @dennishuo or @collado-mike know more.

singhpk234: I would love to know about the history, because i think malicious client can also point it to a role within the same account but different privledge if they have the access to catalog ? want to understand the attack vector more (not a blocker for this pr)

adutra: Me too 😄 The original discussion https://github.com/apache/polaris/discussions/3885 was precisely about that.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisAdminService.java
```diff
@@ -852,16 +852,19 @@ private void validateUpdateCatalogDiffOrThrow(
     if (currentStorageConfig instanceof AwsStorageConfigurationInfo currentAwsConfig
         && newStorageConfig instanceof AwsStorageConfigurationInfo newAwsConfig) {
 
-      if (!Objects.equals(currentAwsConfig.getAwsAccountId(), newAwsConfig.getAwsAccountId())) {
+      // Allow setting a role ARN when none was previously configured, but prevent changing the
+      // AWS account ID once it has been set.
```
dimas-b: TBH, I do not see much value in allowing changes from unset to a new roles in some account, but preventing changing to a role in another account.

The security implications are the same for changing from unset to some role as for changing from one account to another, IMHO.

If the new role has more access to the data in storage than the Polaris Administrator considers valid (for whatever reason), and Polaris is able to assume that role with current credentials, it is a mistake in the authorization setup in the new role's account. 

This can hardly be a concern when the Polaris owner (who controls Polaris storage access credentials) and the catalog owner is the same person or entity.

I can see this being a concern when the Polaris owner is different from the catalog admin.

Therefore, I'd prefer adding a feature flag to enable any role changes (at the discretion of the Polaris owner). If the flag is not set, let's keep current behaviour. The default should probably be `off` (current, more restricted behaviour).

WDYT?

adutra: I don't disagree 😄 

Since we don't have feedback from the original Polaris authors, let's go with the feature flag approach.

dimas-b: Thx! I believe the feature flag will allow maintaining  the old (current) behaviour when the Polaris owner so wishes, and allow more flexibility in fresh deployments.

adutra:
This PR has been in review for 2 weeks. Let's merge it by today EOD if there are no objections.

PR #4023, opened 2026-03-19
feat(storage): inline polaris.storage.name overrides for namespace/table

sririshindra:
### Summary
This PR adds inline storage-name overrides for namespaces and tables using the polaris.storage.name property, so credential vending can select named credentials per entity without introducing new management APIs.

It also refactors hierarchy lookup logic to remove duplication and adds focused test coverage for override behavior and credential vending.

### Why this change is needed
Today, catalog-level storage credentials are often too coarse for multi-tenant or team-isolated workloads. This change enables namespace/table-scoped credential selection by storing an inline override and resolving it through the existing hierarchy at runtime.

### What changed
1. Added storage-name validation and normalization utilities.
2. Added a helper to clone storage configuration while replacing only storageName.
3. Allowed polaris.storage.name through reserved-property filtering.
4. Implemented override handling for:
   1. namespace create
   2. namespace setProperties
   3. namespace removeProperties
   4. table metadata/property flows
5. Refactored hierarchy traversal into shared utility logic to avoid duplicate implementations.
6. Updated and expanded tests, including an end-to-end credential-vending assertion where leaf override credentials are vended.

### Behavioral notes
1. polaris.storage.name accepts alphanumeric, hyphen, underscore, up to 128 chars.
2. Blank values normalize to null.
3. Override resolution uses nearest ancestor storage config and replaces only storageName.
4. Credential vending uses the resolved inline storage configuration for the target entity.

### Testing
1. Added/updated unit and integration tests for validation, inheritance, persistence, and vending behavior.
2. Verified targeted runtime-service tests for override + vending paths passed.

### Reviewer notes
1. This PR intentionally keeps storage type inheritance unchanged: override changes storageName, not storage type.
2. A follow-up PR is planned for cross-type override scenarios (for example AWS catalog overridden by Azure/GCS at leaf scope).

### Checklist
- [ ] 🧾 Updated CHANGELOG.md (if needed)
- [ ] 📚 Updated documentation in site/content/in-dev/unreleased (if needed)

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
@@ -657,6 +657,18 @@ public class FeatureConfiguration<T> extends PolarisConfiguration<T> {
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> ALLOW_STORAGE_NAME_OVERRIDE =
+      PolarisConfiguration.<Boolean>builder()
+          .key("ALLOW_STORAGE_NAME_OVERRIDE")
+          .description(
+              "If set to true, allows namespaces and tables to set the polaris.storage.name "
+                  + "property to override which server-configured named storage profile is used "
+                  + "for credential vending. The override is resolved leaf-to-root at access time. "
+                  + "Whether the resolved name actually changes vended credentials additionally "
+                  + "depends on RESOLVE_CREDENTIALS_BY_STORAGE_NAME being enabled. Default: false.")
+          .defaultValue(false)
+          .buildFeatureConfiguration();
+
   public static final FeatureConfiguration<Boolean> ALLOW_UNRESTRICTED_STORAGE_CONFIG_ROLE_CHANGES =
       PolarisConfiguration.<Boolean>builder()
           .key("ALLOW_UNRESTRICTED_STORAGE_CONFIG_ROLE_CHANGES")
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
@@ -166,13 +166,7 @@ public abstract class PolarisStorageConfigurationInfo {
 
   public static Optional<LocationRestrictions> forEntityPath(
       RealmConfig realmConfig, List<PolarisEntity> entityPath) {
-    return findStorageInfoFromHierarchy(entityPath)
-        .map(
-            storageInfo ->
-                deserialize(
-                    storageInfo
-                        .getInternalPropertiesAsMap()
-                        .get(PolarisEntityConstants.getStorageConfigInfoPropertyName())))
+    return StorageConfigOverrideResolver.resolveEffectiveConfig(entityPath)
         .map(
             configInfo -> {
               List<PolarisEntity> entityPathReversed = new ArrayList<>(entityPath);
@@ -207,18 +201,6 @@ public abstract class PolarisStorageConfigurationInfo {
                 return new LocationRestrictions(configInfo);
               }
             });
-  }
-
-  public static @NonNull Optional<PolarisEntity> findStorageInfoFromHierarchy(
-      List<PolarisEntity> entityPath) {
-    for (int i = entityPath.size() - 1; i >= 0; i--) {
-      PolarisEntity e = entityPath.get(i);
-      if (e.getInternalPropertiesAsMap()
-          .containsKey(PolarisEntityConstants.getStorageConfigInfoPropertyName())) {
-        return Optional.of(e);
-      }
-    }
-    return Optional.empty();
   }
 
   /** Subclasses must provide the Iceberg FileIO impl associated with their type in this method. */
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
@@ -39,8 +39,11 @@ import org.apache.polaris.core.entity.CatalogEntity;
 import org.apache.polaris.core.entity.PolarisEntity;
 import org.apache.polaris.core.entity.PolarisEntityConstants;
 import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
+import org.apache.polaris.core.storage.aws.ImmutableAwsStorageConfigurationInfo;
 import org.apache.polaris.core.storage.azure.AzureStorageConfigurationInfo;
+import org.apache.polaris.core.storage.azure.ImmutableAzureStorageConfigurationInfo;
 import org.apache.polaris.core.storage.gcp.GcpStorageConfigurationInfo;
+import org.apache.polaris.core.storage.gcp.ImmutableGcpStorageConfigurationInfo;
 import org.immutables.value.Value;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
@@ -88,6 +91,38 @@ public abstract class PolarisStorageConfigurationInfo {
   public abstract String getStorageName();
 
   public abstract StorageType getStorageType();
+
+  /**
+   * Return a copy of {@code base} with its {@code storageName} replaced by {@code storageName}. All
+   * other fields are preserved. Used by the storage-name override path so that an entity-level
+   * override flows through credential vending and into the per-backend cache key without mutating
+   * the original config.
+   *
+   * @throws IllegalArgumentException if {@code base} is not one of the four supported subtypes
+   */
+  public static PolarisStorageConfigurationInfo withStorageName(
+      PolarisStorageConfigurationInfo base, @Nullable String storageName) {
+    if (base instanceof AwsStorageConfigurationInfo aws) {
+      return ImmutableAwsStorageConfigurationInfo.builder().from(aws).storageName(storageName).build();
+    }
+    if (base instanceof AzureStorageConfigurationInfo azure) {
+      return ImmutableAzureStorageConfigurationInfo.builder()
+          .from(azure)
+          .storageName(storageName)
+          .build();
+    }
+    if (base instanceof GcpStorageConfigurationInfo gcp) {
+      return ImmutableGcpStorageConfigurationInfo.builder().from(gcp).storageName(storageName).build();
+    }
+    if (base instanceof FileStorageConfigurationInfo file) {
+      return ImmutableFileStorageConfigurationInfo.builder()
+          .from(file)
+          .storageName(storageName)
+          .build();
+    }
+    throw new IllegalArgumentException(
+        "Unsupported PolarisStorageConfigurationInfo subtype: " + base.getClass().getName());
+  }
 
   private static final ObjectMapper DEFAULT_MAPPER;
 
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
@@ -103,7 +103,10 @@ public abstract class PolarisStorageConfigurationInfo {
   public static PolarisStorageConfigurationInfo withStorageName(
       PolarisStorageConfigurationInfo base, @Nullable String storageName) {
     if (base instanceof AwsStorageConfigurationInfo aws) {
-      return ImmutableAwsStorageConfigurationInfo.builder().from(aws).storageName(storageName).build();
+      return ImmutableAwsStorageConfigurationInfo.builder()
+          .from(aws)
+          .storageName(storageName)
+          .build();
     }
     if (base instanceof AzureStorageConfigurationInfo azure) {
       return ImmutableAzureStorageConfigurationInfo.builder()
@@ -112,7 +115,10 @@ public abstract class PolarisStorageConfigurationInfo {
           .build();
     }
     if (base instanceof GcpStorageConfigurationInfo gcp) {
-      return ImmutableGcpStorageConfigurationInfo.builder().from(gcp).storageName(storageName).build();
+      return ImmutableGcpStorageConfigurationInfo.builder()
+          .from(gcp)
+          .storageName(storageName)
+          .build();
     }
     if (base instanceof FileStorageConfigurationInfo file) {
       return ImmutableFileStorageConfigurationInfo.builder()
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.storage;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.polaris.core.entity.PolarisEntity;
+import org.apache.polaris.core.entity.PolarisEntityConstants;
+import org.jspecify.annotations.NonNull;
+
+/**
+ * Resolves the effective {@link PolarisStorageConfigurationInfo} for an entity hierarchy by
+ * applying any nearest-ancestor {@code storage_name_override} on top of the catalog's base storage
+ * configuration.
+ *
+ * <p>The input list is in root-to-leaf order (matching {@link
+ * PolarisStorageConfigurationInfo#findStorageInfoFromHierarchy(List)}). The walk runs leaf-to-root
+ * and:
+ *
+ * <ol>
+ *   <li>Captures the first {@code storage_name_override} encountered (nearest wins).
+ *   <li>Returns the first {@code storage_configuration_info} encountered — the base config.
+ * </ol>
+ *
+ * <p>If a base config exists and an override was captured, the result is {@link
+ * PolarisStorageConfigurationInfo#withStorageName} applied to the base. If no base config exists,
+ * the result is empty regardless of any override.
+ */
+public final class StorageConfigOverrideResolver {
+
+  private StorageConfigOverrideResolver() {}
+
+  /**
+   * Walk {@code entityPath} leaf-to-root and return the effective storage configuration. The base
+   * config is sourced from the nearest entity that has {@code storage_configuration_info} in its
+   * internal properties; if any closer ancestor (including the leaf itself) carries {@code
+   * storage_name_override}, that name replaces the base config's {@code storageName}.
+   *
+   * @param entityPath entities in root-to-leaf order; the catalog is at index 0
+   * @return effective config, or empty if no entity in the chain has a base storage config
+   */
+  public static Optional<PolarisStorageConfigurationInfo> resolveEffectiveConfig(
+      @NonNull List<? extends PolarisEntity> entityPath) {
+    String override = null;
+    for (int i = entityPath.size() - 1; i >= 0; i--) {
+      Map<String, String> internalProps = entityPath.get(i).getInternalPropertiesAsMap();
+      if (override == null) {
+        String candidate =
+            internalProps.get(PolarisEntityConstants.getStorageNameOverridePropertyName());
+        if (candidate != null && !candidate.isEmpty()) {
+          override = candidate;
+        }
+      }
+      String serializedBase =
+          internalProps.get(PolarisEntityConstants.getStorageConfigInfoPropertyName());
+      if (serializedBase != null) {
+        PolarisStorageConfigurationInfo base =
+            PolarisStorageConfigurationInfo.deserialize(serializedBase);
+        if (override != null) {
+          return Optional.of(PolarisStorageConfigurationInfo.withStorageName(base, override));
+        }
+        return Optional.of(base);
+      }
+    }
+    return Optional.empty();
+  }
+}
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
@@ -30,9 +30,8 @@ import org.jspecify.annotations.NonNull;
  * applying any nearest-ancestor {@code storage_name_override} on top of the catalog's base storage
  * configuration.
  *
- * <p>The input list is in root-to-leaf order (matching {@link
- * PolarisStorageConfigurationInfo#findStorageInfoFromHierarchy(List)}). The walk runs leaf-to-root
- * and:
+ * <p>The input list is in root-to-leaf order (the catalog is at index 0). The walk runs
+ * leaf-to-root and:
  *
  * <ol>
  *   <li>Captures the first {@code storage_name_override} encountered (nearest wins).
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
@@ -57,8 +57,7 @@ public final class StorageNameValidator {
           "polaris.storage.name length must be at most " + MAX_LENGTH + " characters");
     }
     if (!VALID.matcher(trimmed).matches()) {
-      throw new IllegalArgumentException(
-          "polaris.storage.name must match pattern [a-zA-Z0-9_-]+");
+      throw new IllegalArgumentException("polaris.storage.name must match pattern [a-zA-Z0-9_-]+");
     }
     return trimmed;
   }
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.storage;
+
+import java.util.regex.Pattern;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Validates and normalizes user-supplied {@code polaris.storage.name} property values used to
+ * select a server-configured named storage profile for credential vending.
+ *
+ * <p>Blank input (null, empty, or whitespace-only) normalizes to {@code null}, which callers
+ * interpret as "clear the override".
+ */
+public final class StorageNameValidator {
+
+  public static final int MAX_LENGTH = 128;
+
+  private static final Pattern VALID = Pattern.compile("^[a-zA-Z0-9_-]+$");
+
+  private StorageNameValidator() {}
+
+  /**
+   * Trim, validate, and normalize a candidate storage name.
+   *
+   * @param value raw user input
+   * @return the trimmed name if valid, or {@code null} if the input is blank
+   * @throws IllegalArgumentException if the trimmed value exceeds {@link #MAX_LENGTH} or contains
+   *     characters outside {@code [a-zA-Z0-9_-]}
+   */
+  public static @Nullable String normalizeAndValidate(@Nullable String value) {
+    if (value == null) {
+      return null;
+    }
+    String trimmed = value.trim();
+    if (trimmed.isEmpty()) {
+      return null;
+    }
+    if (trimmed.length() > MAX_LENGTH) {
+      throw new IllegalArgumentException(
+          "polaris.storage.name length must be at most " + MAX_LENGTH + " characters");
+    }
+    if (!VALID.matcher(trimmed).matches()) {
+      throw new IllegalArgumentException(
+          "polaris.storage.name must match pattern [a-zA-Z0-9_-]+");
+    }
+    return trimmed;
+  }
+}
--- a/polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
+++ b/polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
@@ -1,0 +1,154 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.core.storage;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
+import org.apache.polaris.core.storage.azure.AzureStorageConfigurationInfo;
+import org.apache.polaris.core.storage.gcp.GcpStorageConfigurationInfo;
+import org.junit.jupiter.api.Test;
+
+class WithStorageNameTest {
+
+  @Test
+  void awsConfigPreservesAllFieldsAndSwapsStorageName() {
+    AwsStorageConfigurationInfo base =
+        AwsStorageConfigurationInfo.builder()
+            .addAllowedLocations("s3://foo/bar", "s3://no/where")
+            .roleARN("arn:aws:iam::123456789012:role/polaris-test")
+            .region("no-where-1")
+            .externalId("external-id")
+            .storageName("default")
+            .build();
+
+    PolarisStorageConfigurationInfo result =
+        PolarisStorageConfigurationInfo.withStorageName(base, "team-a");
+
+    assertThat(result).isInstanceOf(AwsStorageConfigurationInfo.class);
+    AwsStorageConfigurationInfo aws = (AwsStorageConfigurationInfo) result;
+    assertThat(aws.getStorageName()).isEqualTo("team-a");
+    assertThat(aws.getRoleARN()).isEqualTo(base.getRoleARN());
+    assertThat(aws.getRegion()).isEqualTo(base.getRegion());
+    assertThat(aws.getExternalId()).isEqualTo(base.getExternalId());
+    assertThat(aws.getAllowedLocations()).containsExactlyElementsOf(base.getAllowedLocations());
+    assertThat(aws).isNotEqualTo(base);
+    assertThat(base.getStorageName()).isEqualTo("default"); // base unchanged
+  }
+
+  @Test
+  void awsConfigAcceptsNullClearingStorageName() {
+    AwsStorageConfigurationInfo base =
+        AwsStorageConfigurationInfo.builder()
+            .addAllowedLocations("s3://foo/bar")
+            .roleARN("arn:aws:iam::123456789012:role/polaris-test")
+            .region("us-east-1")
+            .storageName("default")
+            .build();
+
+    AwsStorageConfigurationInfo result =
+        (AwsStorageConfigurationInfo) PolarisStorageConfigurationInfo.withStorageName(base, null);
+
+    assertThat(result.getStorageName()).isNull();
+    assertThat(result.getRoleARN()).isEqualTo(base.getRoleARN());
+  }
+
+  @Test
+  void azureConfigPreservesFieldsAndSwapsStorageName() {
+    AzureStorageConfigurationInfo base =
+        AzureStorageConfigurationInfo.builder()
+            .addAllowedLocations("abfs://foo@bar.baz/", "abfss://boo@meep.buzz/")
+            .tenantId("tenant-id")
+            .storageName("default")
+            .build();
+
+    AzureStorageConfigurationInfo result =
+        (AzureStorageConfigurationInfo)
+            PolarisStorageConfigurationInfo.withStorageName(base, "team-azure");
+
+    assertThat(result.getStorageName()).isEqualTo("team-azure");
+    assertThat(result.getTenantId()).isEqualTo("tenant-id");
+    assertThat(result.getAllowedLocations()).containsExactlyElementsOf(base.getAllowedLocations());
+  }
+
+  @Test
+  void gcpConfigPreservesFieldsAndSwapsStorageName() {
+    GcpStorageConfigurationInfo base =
+        GcpStorageConfigurationInfo.builder()
+            .addAllowedLocations("gs://foo/bar")
+            .gcpServiceAccount("svc@project.iam.gserviceaccount.com")
+            .storageName("default")
+            .build();
+
+    GcpStorageConfigurationInfo result =
+        (GcpStorageConfigurationInfo)
+            PolarisStorageConfigurationInfo.withStorageName(base, "team-gcp");
+
+    assertThat(result.getStorageName()).isEqualTo("team-gcp");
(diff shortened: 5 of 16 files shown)


tokoko:
This should definitely go behind a feature flag since there's currently no way to limit what a user can put in `polaris.storage.name` property. 

I hate to be adding to the feature flag proliferation in polaris, but maybe there should be a way to configure at what level override is allowed to happen. I can see an admin wanting to allow override on namespace level, but not on table level since `create table` permissions are generally more liberally granted than `create namespace` ones. A more proper fix is to integrate `polaris.storage.name` property in the authorization step, but I'm not sure that's practical for internal rbac.

sririshindra:
> This should definitely go behind a feature flag since there's currently no way to limit what a user can put in `polaris.storage.name` property.
> 
> I hate to be adding to the feature flag proliferation in polaris, but maybe there should be a way to configure at what level override is allowed to happen. I can see an admin wanting to allow override on namespace level, but not on table level since `create table` permissions are generally more liberally granted than `create namespace` ones. A more proper fix is to integrate `polaris.storage.name` property in the authorization step, but I'm not sure that's practical for internal rbac.

Thanks for the review, @tokoko! I'll go ahead and put this behind a feature flag. I intentionally kept authorization out of scope for this phase, but we can definitely address integrating `polaris.storage.name` into the authz step (and adding level-based configuration) in a follow-up PR.

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.polaris.core.storage.gcp.GcpStorageConfigurationInfo;
+import org.junit.jupiter.api.Test;
+
+class WithStorageNameTest {
+
+  @Test
+  void awsFieldsPreserved() {
+    AwsStorageConfigurationInfo base =
+        AwsStorageConfigurationInfo.builder()
+            .allowedLocations(List.of("s3://bucket/path"))
```
adnanhemani: nit: would prefer if you pulled all the variables out and referred to them twice. It would make the test more readable and maintainable :)

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
```diff
@@ -0,0 +1,171 @@
...
+    PolarisStorageConfigurationInfo result =
+        PolarisStorageConfigurationInfo.withStorageName(base, "file-storage");
+
+    assertThat(result).isInstanceOf(FileStorageConfigurationInfo.class);
+    assertThat(result.getStorageName()).isEqualTo("file-storage");
+    assertThat(result.getAllowedLocations()).containsExactly("file:///tmp/warehouse");
+  }
+
+  @Test
+  void serializationRoundTrip() {
```
adnanhemani: what is this test trying to test? That an overriden `PolarisStorageConfigurationInfo` can be serialized/deserialized? Is there any custom logic that puts this into doubt?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -755,8 +806,19 @@ public boolean removeProperties(Namespace namespace, Set<String> properties)
...
-        new PolarisEntity.Builder(entity).setProperties(updatedProperties).build();
+    PolarisEntity.Builder updatedEntityBuilder =
+        new PolarisEntity.Builder(entity).setProperties(updatedProperties);
+
+    // If polaris.storage.name is being removed, clear the storage config override
+    if (properties.contains(POLARIS_STORAGE_NAME_PROPERTY)) {
+      LOGGER.debug("Clearing storage name override on namespace {}", namespace);
+      Map<String, String> updatedInternalProperties =
+          new HashMap<>(entity.getInternalPropertiesAsMap());
+      updatedInternalProperties.remove(PolarisEntityConstants.getStorageConfigInfoPropertyName());
```
adnanhemani: could it be potentially easier to just transform `POLARIS_STORAGE_NAME_PROPERTY` into `PolarisEntityConstants.getStorageConfigInfoPropertyName()` prior to removing all the properties the first time?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  /**
+   * Normalizes a blank or empty string to {@code null}. Non-blank values are trimmed.
+   *
+   * @param value the input value
+   * @return {@code null} if the input is null, empty, or blank; otherwise the trimmed value
+   */
+  public static @Nullable String normalizeBlankToNull(@Nullable String value) {
```
adnanhemani: Is there a reason this shouldn't be called from within `validate`?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/StorageNameOverrideTest.java
```diff
@@ -0,0 +1,309 @@
...
+ */
+class StorageNameOverrideTest {
+
+  private static final String STORAGE_CONFIG_KEY =
+      PolarisEntityConstants.getStorageConfigInfoPropertyName();
+
+  @Test
+  void storageConfigFromPropertyOverride_noProperty_returnsNull() throws Exception {
+    // Use reflection since the method is package-private/static
+    var method =
```
adnanhemani: really not a fan of reflection...

would much rather prefer doing `@VisibleForTest` on the method.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/StorageNameOverrideTest.java
```diff
@@ -0,0 +1,309 @@
...
+        IcebergCatalog.class.getDeclaredMethod(
+            "storageConfigFromPropertyOverride", Map.class, List.class);
+    method.setAccessible(true);
+
+    Map<String, String> properties = new HashMap<>();
+    properties.put(IcebergCatalog.POLARIS_STORAGE_NAME_PROPERTY, "ns-storage");
+
+    assertThatThrownBy(() -> method.invoke(null, properties, List.of()))
+        .hasCauseInstanceOf(org.apache.iceberg.exceptions.BadRequestException.class)
```
adnanhemani: no need for the full class path, import instead.

adnanhemani (commented):
Would recommend @dimas-b and @flyrain also take a look at this :)

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -180,6 +180,63 @@ public static Optional<LocationRestrictions> forEntityPath(
...
+   *
+   * @param baseConfig the storage configuration to copy
+   * @param storageName the new storage name (may be null to clear an override)
+   * @return a new configuration instance with the updated storage name
+   * @throws IllegalArgumentException if the base config type is not supported
+   */
+  public static PolarisStorageConfigurationInfo withStorageName(
+      PolarisStorageConfigurationInfo baseConfig, @Nullable String storageName) {
+    if (baseConfig instanceof AwsStorageConfigurationInfo awsConfig) {
+      return AwsStorageConfigurationInfo.builder()
```
dimas-b: Maybe `.builder().from(awsConfig)`, etc... ?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -169,6 +169,22 @@ public static void enforceFeatureEnabledOrThrow(
           .defaultValue(false)
           .buildFeatureConfiguration();
 
+  public static final FeatureConfiguration<Boolean> ALLOW_STORAGE_NAME_OVERRIDE =
+      PolarisConfiguration.<Boolean>builder()
+          .key("ALLOW_STORAGE_NAME_OVERRIDE")
```
dimas-b: Please add a CHANGELOG entry for this new flag.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -520,15 +534,35 @@ private void createNamespaceInternal(
...
+      LOGGER.debug(
+          "Applying storage name override '{}' on namespace {}",
+          namespaceStorageConfig.getStorageName(),
+          namespace);
+      entity =
+          new NamespaceEntity(
+              new PolarisEntity.Builder(nsBuilder.build())
+                  .addInternalProperty(
+                      PolarisEntityConstants.getStorageConfigInfoPropertyName(),
+                      namespaceStorageConfig.serialize())
```
dimas-b: Why embed a copy of the config into the entity here? 

Based on previous `dev` discussion (link below) my impression is that we'd want to keep (named) storage configs at the catalog level for ease of management.

https://lists.apache.org/thread/nwcc8krhvm35olgc22676mm12cjkpgn5

I'd expect entities to only reference storage config by name, and the config resolution to happen when requested 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1745,7 +1811,78 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
...
+      return null;
+    }
+
+    String rawValue = properties.get(POLARIS_STORAGE_NAME_PROPERTY);
+    String storageName = StorageNameValidator.normalizeBlankToNull(rawValue);
+    if (storageName != null) {
+      StorageNameValidator.validate(storageName);
+    }
+
+    PolarisStorageConfigurationInfo baseConfig = deserializeStorageConfigFromEntityPath(parentPath);
```
dimas-b: This config is loaded regardless on the "name", so I'm not sure what the purpose of adding `POLARIS_STORAGE_NAME_PROPERTY` to the entity properties might be 🤔

dimas-b (commented):
Hi @sririshindra , thanks for the PR, but I'm a bit confused by the new code... specific comments below. Sorry, if I missed something obvious 😅

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
sririshindra: Done — extracted the expected values into local variables and referenced them in both the setup and assertion.

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
sririshindra: You're right. The intent was to verify that a config produced by withStorageName() survives serialize() → deserialize() since these objects use Jackson's @JsonTypeInfo/@JsonSubTypes for polymorphic deserialization. But in practice, builder().from() produces the same concrete type, so Jackson will always serialize/deserialize it correctly. Looks like awsFieldsPreserved test already covers field preservation. Happy to remove the round-trip test if you think it's redundant.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
sririshindra: refactored withStorageName() to use builder().from() for each storage type.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
sririshindra: added a CHANGELOG entry for ALLOW_STORAGE_NAME_OVERRIDE

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
sririshindra: They serve different purposes: validate() is a pure check, while normalizeBlankToNull() is a transformation. Combining them would make validate() impure — callers wouldn't expect a validation method to also mutate/transform the input. That said, I could rename normalizeBlankToNull to something clearer like toNullIfBlank or inline it at the call sites if you'd prefer. Open to suggestions.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: Agreed — this is fixed in the latest revision. Entities now store only a lightweight storage_name_override string reference, not a full config copy. The full config resolution happens on-demand at access time.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: I am not sure I understood your point correctly. But from what I understood, the challenge is that polaris.storage.name is a user-facing property (in getPropertiesAsMap()) while storage_name_override is an internal property (in getInternalPropertiesAsMap()). These are separate maps, so a key transformation before the first removal pass wouldn't help.
We'd still need to reach into both maps. The special if-block is the way we bridge the user-facing property removal to its internal property counterpart.
But maybe I am getting confused. Could you elaborate your point a bit more.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
sririshindra: In the updated implementation, polaris.storage.name is the user-facing API property that gets translated into an internal storage_name_override reference. At resolution time, StorageConfigResolver finds the nearest override in the hierarchy and applies it to the catalog's base config via withStorageName(). The purpose is to allow different namespaces/tables to use different named storage configurations while the actual configs remain managed centrally at the catalog level. The property is no longer redundant — it's the mechanism by which users select which storage configuration to use.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/StorageNameOverrideTest.java
sririshindra: Agreed — replaced the reflection with @VisibleForTesting on the method.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/StorageNameOverrideTest.java
sririshindra: Fixed — removed the fully qualified name

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -538,4 +542,76 @@ default Optional<PrincipalRoleEntity> findPrincipalRoleByName(
...
+
+    // Build the entity chain (leaf-to-root) by walking parentId links.
+    long catalogId = entity.getCatalogId();
+    List<PolarisBaseEntity> chain = new ArrayList<>();
+    chain.add(entity);
+
+    long currentParentId = entity.getParentId();
+    while (currentParentId != PolarisEntityConstants.getNullId() && currentParentId != catalogId) {
+      EntityResult parentResult =
+          loadEntity(callCtx, catalogId, currentParentId, PolarisEntityType.NAMESPACE);
```
dimas-b: I'm not sure about this loop, TBH... So far the usual workflow for requests was to use the `Resolver` to find all relevant entities based on API parameters upfront, then proceed with request execution.

So, normally, all the parents up to the catalog should already be loaded by the time `StorageAccessConfig` is required. Could we reuse existing data from the resolution manifest.

Note: this may overlap with #3699 CC: @tokoko

tokoko: I think this PR would no longer have to solve this dual resolution issue after #3699 since everything happens in `StorageAccessConfigProvider` there. honestly this feels like we're putting bandaids on bandaids :smile: ideally we should have a single place (StorageAccessConfigProvider most likely) that would resolve storageInfo, resolve storage name, get additional properties from config per storage name and return effective storage configuration. I don't think that's possible to do now w/o first getting rid of all these unnecessary layers.

dimas-b: From my POV, I'd prefer to merge #3699 first, then rebase this PR.

@sririshindra : WDYT?

flyingImer (commented):
Thanks, the newer override-resolution tests are useful.

I still think there is one high-priority correctness issue worth fixing before this moves forward.

IIUC doCommit() still resolves commit-time storage access from the pre-override hierarchy, and only later materializes polaris.storage.name into the table state. If so, the post-commit state may look correct while the first metadata write already used the previous credentials.

That feels like the key invariant here: the effective storage for the write path should be resolved before the commit-time metadata write happens, not only after the override has been persisted.

I’d be much more comfortable if we either move that resolution earlier in the flow, or add explicit coverage proving the first create/update write already uses the overridden credentials.

@flyrain WDYT?

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
sririshindra: Sounds good @dimas-b . I will wait for https://github.com/apache/polaris/pull/3699 to be merged and rebase this PR. I will review it as well.

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
adnanhemani: Yes, I think we should remove the test IMO. I'm all for adding tests but this one looks like we're basically just testing the Jackson library.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
adnanhemani: I understand your point but given that this function is always only called right before validate, maybe there's a way to combine both the functions to reduce the total amount of functions and reduce caller code as well.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
adnanhemani: I see what you are saying now. LGTM.

polaris-core/src/test/java/org/apache/polaris/core/storage/StorageConfigResolverTest.java
```diff
@@ -0,0 +1,181 @@
...
+    Optional<PolarisStorageConfigurationInfo> result =
+        StorageConfigResolver.resolve(List.of(table, namespace));
+
+    assertThat(result).isEmpty();
+  }
+
+  @Test
+  void overridePreservesBaseConfigFields() {
+    PolarisBaseEntity table = entityWithOverride("overridden-storage", 3L, 2L);
+    String roleArn = "arn:aws:iam::123456789012:role/test-role";
```
adnanhemani: Move magic variable to the top. I'd also suggest moving all the `ns-storage`, `catalog-storage`, etc. to the top as well.

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
```diff
@@ -538,4 +542,76 @@ default Optional<PrincipalRoleEntity> findPrincipalRoleByName(
...
+    // Only create a synthetic entity if the resolved config differs from the catalog's base
+    // (i.e., an override was actually applied). If no override exists, the resolved config is
+    // the catalog's own config and should not be stamped onto a non-catalog entity.
+    String catalogConfigJson =
+        catalogResult
+            .getEntity()
+            .getInternalPropertiesAsMap()
+            .get(PolarisEntityConstants.getStorageConfigInfoPropertyName());
+    String resolvedJson = resolved.get().serialize();
+    if (resolvedJson.equals(catalogConfigJson)) {
```
adnanhemani: Could this be brittle in terms of Jackson not serializing/de-serializing the JSON with the same ordering of keys, given that JSON objects don't have any guarantees on ordering of keys?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java
```diff
@@ -18,35 +18,98 @@
...
+  }
+
+  /**
+   * Deserializes and returns the effective storage configuration from the hierarchy, applying any
+   * {@code storageNameOverride} found on descendant entities.
+   *
+   * @param entityPath a list of entities ordered root-to-leaf
+   * @return the effective {@link PolarisStorageConfigurationInfo} or null if not found
+   */
+  public static PolarisStorageConfigurationInfo deserializeStorageConfigFromEntityPath(
```
adnanhemani: Can we potentially rename this since it does apply the overrides? The main concern is the same function name is present in `IcebergCatalog.java` but doesn't apply the overrides there.

IMO this function is doing more than just "deserializing" the storage config here, so it should be named appropriately.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/StorageNameOverrideTest.java
```diff
@@ -0,0 +1,350 @@
...
+import org.apache.polaris.service.events.PolarisEventMetadataFactory;
+import org.apache.polaris.service.task.TaskExecutor;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for the static helper methods in {@link IcebergCatalog} that handle {@code
+ * polaris.storage.name} property overrides. These test the pure logic without requiring a full
+ * Quarkus test context.
+ */
+class StorageNameOverrideTest {
```
adnanhemani, line 50: `enforceStorageNameOverrideEnabledForTableCommit` has a special case: if `priorValue == requestedValue`, it should not throw even when the flag is disabled (so existing tables aren't broken if the flag gets turned off).

Could we add a test in StorageNameOverrideTest covering this path — flag disabled, same value in both base and metadata, assert no exception?

adnanhemani (commented):
@flyingImer's pointing out of the "cold start" issue is a known issue with approaching this feature unfortunately (we discussed it at the last Polaris Community Sprint).

I think we are ok with saying that the credential can only be used once the table is pre-created. But @sririshindra - can you explicitly list (within this PR) the set of operations in which users can do this to avail this feature?

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
sririshindra: Thanks, agreed. I removed the serializationRoundTrip test.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
sririshindra: I combined normalization and validation into a single helper, normalizeAndValidate, and updated all call sites to use it.

polaris-core/src/test/java/org/apache/polaris/core/storage/StorageConfigResolverTest.java
sririshindra: Done. I extracted repeated storage-name and ARN literals into top-level constants.

polaris-core/src/main/java/org/apache/polaris/core/persistence/PolarisMetaStoreManager.java
sririshindra: Great catch. I replaced string equality on serialized JSON with structural comparison using ObjectMapper and JsonNode equality. This makes the comparison robust to field ordering differences.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java
sririshindra: Done. I renamed the FileIO helper to make behavior explicit: resolveEffectiveStorageConfigFromEntityPath. The old name implied pure deserialization, while this path actually resolves and applies overrides.

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/StorageNameOverrideTest.java
sririshindra: Added a test for flag disabled plus priorValue equals requestedValue, and it asserts no exception is thrown. This protects existing tables when the feature flag is later turned off.

runtime/service/src/main/java/org/apache/polaris/service/storage/PolarisStorageIntegrationProviderImpl.java
```diff
@@ -129,8 +129,7 @@ public PolarisStorageIntegrationProviderImpl(
   @Override
   public @Nullable PolarisStorageIntegration getStorageIntegration(
       @NonNull List<PolarisEntity> resolvedEntityPath) {
-    return PolarisStorageConfigurationInfo.findStorageInfoFromHierarchy(resolvedEntityPath)
```
tokoko, line 132: can we get rid of `PolarisStorageConfigurationInfo.findStorageInfoFromHierarchy`? `StorageConfigOverrideResolver.resolveEffectiveConfig` seems like a more correct replacement for it now.

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -657,6 +657,18 @@ public static void enforceFeatureEnabledOrThrow(
...
+          .description(
+              "If set to true, allows namespaces and tables to set the polaris.storage.name "
```
dimas-b, line 664: What is the value of this flag? If a user points to a non-existent storage name, the adverse effect is localized to that user's change.

Adding storage names is a Catalog Management action and should probably be covered by AuthZ checks.

I'd think setting/changing `polaris.storage.name` probably needs an new AuthZ operation rather than a global feature flag.

WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1937,6 +2002,88 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
...
+   */
+  @VisibleForTesting
+  void processStorageNameOverrideOnWrite(
+      Map<String, String> userProperties,
+      Map<String, String> internalProperties,
+      @Nullable String priorOverride) {
+    if (!userProperties.containsKey(POLARIS_STORAGE_NAME_PROPERTY)) {
+      return;
+    }
+    String rawValue = userProperties.remove(POLARIS_STORAGE_NAME_PROPERTY);
```
dimas-b, line 2038: Modifying parameters is not intuitive... Can we avoid this?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1937,6 +2002,88 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
...
+    String requested;
+    try {
+      requested = StorageNameValidator.normalizeAndValidate(rawValue);
+    } catch (IllegalArgumentException e) {
+      throw new BadRequestException(
```
dimas-b, line 2043: `IllegalArgumentException` should automatically result in a 400 response at the REST API layer... Why do we need `BadRequestException` here?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1937,6 +2002,88 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
...
+   * @param internalProperties internal-property map that will be persisted onto the entity; the
+   *     {@code storage_name_override} key is set or removed in place based on the request.
+   * @param priorOverride the override value already persisted on the entity (or {@code null} for a
+   *     fresh create). Used to decide whether the request is an idempotent re-set.
+   */
+  @VisibleForTesting
+  void processStorageNameOverrideOnWrite(
+      Map<String, String> userProperties,
+      Map<String, String> internalProperties,
+      @Nullable String priorOverride) {
```
dimas-b, line 2034: Is `priorOverride` actually an "override"? What does it override? 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1937,6 +2002,88 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
...
+    if (changing && !realmConfig.getConfig(FeatureConfiguration.ALLOW_STORAGE_NAME_OVERRIDE)) {
+      throw new BadRequestException(
+          "Setting %s requires feature %s to be enabled",
+          POLARIS_STORAGE_NAME_PROPERTY, FeatureConfiguration.ALLOW_STORAGE_NAME_OVERRIDE.key());
+    }
+    String storageOverrideKey = PolarisEntityConstants.getStorageNameOverridePropertyName();
+    if (requested == null) {
+      internalProperties.remove(storageOverrideKey);
+    } else {
+      internalProperties.put(storageOverrideKey, requested);
```
dimas-b, line 2056: This logic effectively make the table metadata control the storage name value on the server side. I'm not sure it is ideal.

Ultimately, the catalog owns storage configuration. Why should table metadata have precedence over catalog-side settings?

I'd imagine Polaris CLI being the tool for setting storage name 🤔 

Perhaps we need to dive deeper into this on the `dev` ML 🤔

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -89,6 +92,44 @@ public boolean validatePrefix() {
...
+   * other fields are preserved. Used by the storage-name override path so that an entity-level
+   * override flows through credential vending and into the per-backend cache key without mutating
+   * the original config.
+   *
+   * @throws IllegalArgumentException if {@code base} is not one of the four supported subtypes
+   */
+  public static PolarisStorageConfigurationInfo withStorageName(
+      PolarisStorageConfigurationInfo base, @Nullable String storageName) {
+    if (base instanceof AwsStorageConfigurationInfo aws) {
+      return ImmutableAwsStorageConfigurationInfo.builder()
```
dimas-b, line 106: I'd propose to make an polymorphic `.toBuilder()` instance method and then do `base.toBuilder().storageName(storageName).build()` without `instanceof` checks.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
```diff
@@ -0,0 +1,83 @@
...
+   * storage_name_override}, that name replaces the base config's {@code storageName}.
+   *
+   * @param entityPath entities in root-to-leaf order; the catalog is at index 0
+   * @return effective config, or empty if no entity in the chain has a base storage config
+   */
+  public static Optional<PolarisStorageConfigurationInfo> resolveEffectiveConfig(
+      @NonNull List<? extends PolarisEntity> entityPath) {
+    String override = null;
+    for (int i = entityPath.size() - 1; i >= 0; i--) {
+      Map<String, String> internalProps = entityPath.get(i).getInternalPropertiesAsMap();
```
dimas-b, line 62: nit: we assume indexed access is efficient here... WDYT about `List.reversed()`?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
```diff
@@ -0,0 +1,83 @@
...
+          override = candidate;
+        }
+      }
+      String serializedBase =
+          internalProps.get(PolarisEntityConstants.getStorageConfigInfoPropertyName());
+      if (serializedBase != null) {
+        PolarisStorageConfigurationInfo base =
+            PolarisStorageConfigurationInfo.deserialize(serializedBase);
+        if (override != null) {
+          return Optional.of(PolarisStorageConfigurationInfo.withStorageName(base, override));
```
dimas-b, line 76: I'm not sure I understand the purpose of reusing the same config with a different name in different sub-trees 🤔 How do you envision this to be used?

My impression from the ML [discussion](https://lists.apache.org/thread/nwcc8krhvm35olgc22676mm12cjkpgn5) is that the idea was to define multiple distinct Storage Configurations at the Catalog level and let entities choose one of them by name 🤔

polaris-core/src/test/java/org/apache/polaris/core/storage/WithStorageNameTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
+import org.apache.polaris.core.storage.azure.AzureStorageConfigurationInfo;
+import org.apache.polaris.core.storage.gcp.GcpStorageConfigurationInfo;
+import org.junit.jupiter.api.Test;
+
+class WithStorageNameTest {
```
dimas-b, line 29: This test class name is a bit confusing IMHO 🤔 Why not add the new test cases to `PolarisStorageConfigurationInfoTest`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java
```diff
@@ -18,35 +18,57 @@
...
+    }
+    Optional<PolarisEntity> baseEntity =
+        path.reversed().stream()
             .filter(
                 e ->
                     e.getInternalPropertiesAsMap()
                         .containsKey(PolarisEntityConstants.getStorageConfigInfoPropertyName()))
             .findFirst();
-    return storageInfoEntity;
+    if (baseEntity.isEmpty()) {
```
dimas-b, line 64: Why repeat the search already performed by `StorageConfigOverrideResolver.resolveEffectiveConfig()`? Can we be sure the outcome is / will be the same?

Why not have `StorageConfigOverrideResolver` return a tuple of all the data we need here?

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java
```diff
@@ -18,35 +18,57 @@
...
+   * <p>This method walks the path leaf-to-root, locates the first entity carrying {@code
+   * storage_configuration_info} (the catalog), and — if any closer ancestor carries a {@code
+   * storage_name_override} — returns a synthetic copy of the base entity whose serialized
+   * configuration has its {@code storageName} replaced. Callers that read the resulting entity's
+   * internal properties (e.g., to stash on a purge task) therefore see the effective config.
    *
    * @param resolvedStorageEntity the resolved entity wrapper containing the hierarchical path
-   * @return an {@link Optional} containing the entity with storage config, or empty if not found
+   * @return an {@link Optional} containing the (possibly synthetic) entity with the effective
+   *     storage config, or empty if no entity in the chain has a base storage config
```
dimas-b, line 47: Callers of this method only need to know the presence of this entity and if present, its properties... Why bother with synthetic entities? I'd propose to refactor this code to return an `Optional<some properties class>`... WDYT?

dimas-b (commented):
Thanks for resuming this PR, @sririshindra !

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -89,6 +92,44 @@ public boolean validatePrefix() {
...
 
+  /**
+   * Return a copy of {@code base} with its {@code storageName} replaced by {@code storageName}. All
+   * other fields are preserved. Used by the storage-name override path so that an entity-level
+   * override flows through credential vending and into the per-backend cache key without mutating
+   * the original config.
+   *
+   * @throws IllegalArgumentException if {@code base} is not one of the four supported subtypes
+   */
+  public static PolarisStorageConfigurationInfo withStorageName(
```
adutra, line 103: Conceptually, this should be an abstract instance method:

```java
public abstract PolarisStorageConfigurationInfo withStorageName(@Nullable String storageName);
```

Each subclass would implement accordingly.

Then callers would do:

```java
PolarisStorageConfigurationInfo base = ...
return base.withStorageName("some-storage");
```

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageNameValidator.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final Pattern VALID = Pattern.compile("^[a-zA-Z0-9_-]+$");
+
+  private StorageNameValidator() {}
+
+  /**
+   * Trim, validate, and normalize a candidate storage name.
+   *
+   * @param value raw user input
+   * @return the trimmed name if valid, or {@code null} if the input is blank
```
adutra, line 43: What do we mean by "normalize"? I think it could be counter-intuitive if Polaris changed a user-supplied identifier without notifying the user about the change, even if it's just trimming spaces.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -157,6 +158,13 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
   private static final Joiner SLASH = Joiner.on("/");
 
+  /**
+   * User-facing property key on namespaces and tables that selects a server-configured named
+   * storage profile to use for credential vending. Validated and translated into the internal
+   * property {@link PolarisEntityConstants#getStorageNameOverridePropertyName()}.
+   */
+  static final String POLARIS_STORAGE_NAME_PROPERTY = "polaris.storage.name";
```
adutra, line 166: Just thinking out loud: if I get this correctly, we would be tying a namespace or table property to a storage profile that must be declared in the server configuration. IOW, the namespace or table in question becomes unusable if the catalog configuration changes and does not contain the profile anymore. Probably OK, but this requires some coordination between two personas: the user that created the table, and the admin that manages the server configuration.

runtime/service/src/main/java/org/apache/polaris/service/catalog/io/FileIOUtil.java
adutra: +1, there is some overlap between this class and `StorageConfigOverrideResolver`.

adutra (commented):
Thanks @sririshindra for this PR.

I must say, I have some reservations regarding the mechanism this PR introduces for passing properties back and forth between user and internal spaces. I am concerned that this approach might set a problematic precedent for future implementations.

Was this topic already discussed? My apologies if it was; I am currently catching up on a significant backlog of PR reviews. 😓

While the PR description suggests this method is better than creating new management API endpoints, I am questioning if dedicated endpoints wouldn't actually be a superior solution for configuring items like storage profile names.

As I mentioned in a separate comment, allowing users to specify which storage profile to use leads to a "split-brain" scenario where Polaris admins define the profiles, but the profile selection responsibility falls to the users.

Furthermore, although this PR addresses storage profile selection, I am unclear on how we intend to handle the application of a full `PolarisStorageConfigurationInfo` at the table or namespace level. Relying on additional user properties for such use cases does not seem like a scalable strategy.

Apologies if I'm off track here.

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageConfigOverrideResolver.java
tokoko: I think you might be misreading this (or maybe I am). storage names are set effectively from server configuration similar to how they are handled after #3409. changing storage name of a config from table/namespace metadata simply changes the pointer to the predefined storages that are part of server config. 

P.S. I also agree the entire approach isn't exactly ideal, but my understanding of the devlist consensus was also how it's implemented in the PR.

dimas-b: I'd expect different storage names to be associated with different `PolarisStorageConfigurationInfo` objects (semantically). The storage name would act as a user-level identifier for a specific `PolarisStorageConfigurationInfo` piece of data.

Simply updating the name in a `PolarisStorageConfigurationInfo` in runtime looks odd to me :thinking:

tokoko: I remember that introducing a separate crud resource for storages was rejected, but iiuc what you're talking about is allowing multiple `PolarisStorageConfigurationInfo` to be defined on catalog level and then tables/namespaces would be able to choose between them, is that right? I suppose that's a reasonably clean approach.

One thing to note... `storagaName` field from #3409 is more of a pointer atm rather than a unique name designation for a specific `PolarisStorageConfigurationInfo`. I suppose it can be both, but it might be a little awkward: for example what if two catalogs define a storage config with same names, but want to resolve different credentials from the server config. I guess it shouldn't be a problem as long as it's well-documented.

dimas-b: I guess the whole idea for multiple storage configs/names is applicable only to self-managed deployments. So the admin should be able to sort out name overlaps there.

I'm not sure how this idea can fit in SaaS deployments, TBH :thinking:

adutra: Exactly, I don't think it's a good idea to let users define the storage profile name on a namespace or table in a managed environment.

> I remember that introducing a separate crud resource for storages was rejected

You mean for storage profile names or for `PolarisStorageConfigurationInfo`? I think we need to settle on this topic; I am leaning more towards the idea of leveraging the management API. That's the direction suggested in [this design doc](https://docs.google.com/document/d/1hbDkE-w84Pn_112iW2vCnlDKPDtyg8flaYcFGjvD120/edit?tab=t.0) by @sririshindra.

dimas-b: Heads up: this PR is on the agenda for the Community Sync call today :wink:

flyingImer (commented):
Hi @sririshindra, catching up on this one: IIUC the direction after the dev-list discussion was to keep #4023 on hold until the catalog-level multiple-storage-config work lands, then rebase this on top of that.

Is that still the plan for this PR, or are you planning to continue moving the current approach forward independently?

If the former, I'm happy to hold off on further review here and take another pass once it's rebased.

sririshindra:
> Hi @sririshindra, catching up on this one: IIUC the direction after the dev-list discussion was to keep #4023 on hold until the catalog-level multiple-storage-config work lands, then rebase this on top of that.
> 
> Is that still the plan for this PR, or are you planning to continue moving the current approach forward independently?
> 
> If the former, I'm happy to hold off on further review here and take another pass once it's rebased.

HI @flyingImer , Sorry I didn't see your message earlier. 
I wrote https://docs.google.com/document/d/1TSaZ0G2oEoWq6lDMSPIIfHLwXxW9VZmhUNE3s_jfxgI/edit?usp=sharing . Could you please review that instead

PR #4031, opened 2026-03-19
Use appropriate scope in producer methods

adutra:
Some producer methods are not annotated, which means that the produced beans get the `@Dependent` scope by default. This is incorrect for some of the beans. This PR fixes it.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
@@ -202,6 +202,7 @@ public class ServiceProducers {
   // Polaris service beans - selected from @Identifier-annotated beans
 
   @Produces
+  @Singleton // used in instanceof checks
   public RealmContextResolver realmContextResolver(
       RealmContextConfiguration config, @Any Instance<RealmContextResolver> realmContextResolvers) {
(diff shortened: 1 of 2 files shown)


runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -202,6 +202,7 @@ public ResolutionManifestFactory resolutionManifestFactory(
   // Polaris service beans - selected from @Identifier-annotated beans
 
   @Produces
+  @Singleton // used in instanceof checks
```
dimas-b, line 205: `instanceof` on a CDI bean looks like an anti-pattern to me. This PR is still a good fix for the scope problem, but we should probably revisit those `instanceof` checks... WDYT?

dimas-b (approved):

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
adutra: I agree. We have instanceof checks in production AND test code... but most test code. It's going to take some time to fix them all.

dimas-b: Sure. Let's do it gradually after this PR :)

PR #4034, opened 2026-03-20
Add MinIO coverage for AWS-shaped KMS vended credentials

castanhas98:
### Summary
This PR adds focused MinIO integration coverage for the AWS-shaped vended-credentials/KMS edge case.

It adds a matrix that varies `region` and `roleArn` independently with `kmsUnavailable=false` and verifies that only the `region + roleArn` case fails during `loadTable`, with the expected MinIO/KMS-scoping error.

It also adds a companion test showing that the same AWS-shaped configuration succeeds when `kmsUnavailable=true`, and refactors the test helper so the failing path still cleans up table state correctly.

### Why this change is needed
The existing MinIO coverage exercised standard vended-credentials behavior, but it did not cover the AWS-shaped read-only KMS path that is triggered only when both `region` and `roleArn` are set while KMS remains enabled.

That path is important because Polaris will add the wildcard KMS ARN for read-only AWS-style access, and MinIO STS rejects that resource as invalid. This PR makes that edge case explicit and locks in the expected `kmsUnavailable` behavior.

## Checklist

### DIFF
--- a/runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
+++ b/runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
@@ -104,8 +104,7 @@ public class RestCatalogMinIOSpecialIT {
   private static final String MINIO_ACCESS_KEY = "test-ak-123";
   private static final String MINIO_SECRET_KEY = "test-sk-123";
   private static final String TEST_REGION = "us-west-2";
-  private static final String TEST_ROLE_ARN =
-      "arn:aws:iam::000000000000:role/polaris-access-role";
+  private static final String TEST_ROLE_ARN = "arn:aws:iam::000000000000:role/polaris-access-role";
   private static String adminToken;
 
   public static class Profile implements QuarkusTestProfile {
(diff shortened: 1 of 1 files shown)


runtime/service/src/intTest/java/org/apache/polaris/service/it/RestCatalogMinIOSpecialIT.java
```diff
@@ -265,8 +301,89 @@ public void testCreateTableVendedCredentials(boolean pathStyle) throws IOExcepti
...
+            Optional.of(endpoint),
+            true,
+            Optional.empty(),
+            Optional.of(VENDED_CREDENTIALS),
+            true,
+            includeRegion ? Optional.of(TEST_REGION) : Optional.empty(),
+            includeRoleArn ? Optional.of(TEST_ROLE_ARN) : Optional.empty(),
+            Optional.of(false))) {
+      TableIdentifier id = createTableAndVerifyMetadata(restCatalog);
+      if (expectFailure) {
```
dimas-b: Would you mind refactoring expected failures and successes into two separate test methods? I believe it would be easier for the reader to follow the test logic and analyse failures, should they occur.

castanhas98: Refactored this into separate success and failure tests so the expected behavior is easier to scan and diagnose. I also reran the targeted MinIO/KMS integration tests locally after the split. Thanks for the suggestion.

dimas-b:
@castanhas98 : please run spotless checks locally - I believe the first CI run failed because of that.

castanhas98:
@dimas-b Thanks for the review! I ran spotless on the previous commit, so we should be good on that front.

PR #4039, opened 2026-03-23
Set Quarkus config profile `it` explicitly

snazy:
Integration tests using Quarkus versions before 3.32.4 somehow implicitly got the `quarkus.profile=it` setting. This seemed to be changed in 3.32.4, so that integration tests start with the `prod` (= default) profile, cauting integration tests to fail, because the necessary configurations are just not there.

Looking at the [Quarkus config reference](https://quarkus.io/guides/config-reference#default-profiles), it seems that there is no `it` profile.

Explicitly setting the config profile to `it`, although not sure why that changed in 3.32.4, seems safer anyways.


adutra (approved):

snazy:
#4040 is passing, so this should unblock #4020

PR #4043, opened 2026-03-23
Add Polaris-Generic-Table-Access-Delegation header in the Generic Table API spec for credential vending

gh-yzou:
Add Polaris-Generic-Table-Access-Delegation header in the Generic Table API spec for credential vending

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/CatalogGenericTableEventServiceDelegator.java
@@ -67,7 +67,12 @@ public class CatalogGenericTableEventServiceDelegator
                 .put(EventAttributes.CREATE_GENERIC_TABLE_REQUEST, createGenericTableRequest)));
     Response resp =
         delegate.createGenericTable(
-            prefix, namespace, createGenericTableRequest, polarisGenericTableAccessDelegation, realmContext, securityContext);
+            prefix,
+            namespace,
+            createGenericTableRequest,
+            polarisGenericTableAccessDelegation,
+            realmContext,
+            securityContext);
     polarisEventDispatcher.dispatch(
         new PolarisEvent(
             PolarisEventType.AFTER_CREATE_GENERIC_TABLE,
@@ -157,7 +162,13 @@ public class CatalogGenericTableEventServiceDelegator
                 .put(EventAttributes.NAMESPACE_NAME, namespace)
                 .put(EventAttributes.GENERIC_TABLE_NAME, genericTable)));
     Response resp =
-        delegate.loadGenericTable(prefix, namespace, genericTable, polarisGenericTableAccessDelegation, realmContext, securityContext);
+        delegate.loadGenericTable(
+            prefix,
+            namespace,
+            genericTable,
+            polarisGenericTableAccessDelegation,
+            realmContext,
+            securityContext);
     polarisEventDispatcher.dispatch(
         new PolarisEvent(
             PolarisEventType.AFTER_LOAD_GENERIC_TABLE,


runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogAdapter.java
```diff
@@ -72,6 +72,7 @@ public Response createGenericTable(
       String prefix,
       String namespace,
       CreateGenericTableRequest createGenericTableRequest,
+      String xGenericTableAccessDelegation,
```
jbonofre: I don't see this argument used anywhere. I guess it's a "preparation" PR, just for the header, not yet actually used.

gh-yzou: Yes, this isn’t implemented yet. The initial goal of this PR was just to update the API spec, but the build started failing since adding new parameters affects the generated interface.

I’ll follow up with additional PRs to implement the credential vending support.

jbonofre: Totally fine. Thanks !

adutra:
Shouldn't we use the `X-Polaris` prefix?

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -182,6 +186,26 @@ components:
         type: string
       example: "sales"
 
+    generic-table-data-access:
+      name: X-Generic-Table-Access-Delegation
```
dimas-b: Even HTTP no longer encourages using the `X-` prefix in headers, AFAIK.

I'd propose `Polaris-Access-Delegation`. It could apply to views (hypothetically) too.

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -182,6 +186,26 @@ components:
...
+      in: header
+      description: >
+        Optional signal to the server that the client supports delegated access
+        via a comma-separated list of access mechanisms. The server may choose
+        to supply access via any or none of the requested mechanisms.
+
+
+        When `vended-credentials` is included, the server may return scoped
```
dimas-b: Re: `may return` - do you mean that the server is not allowed to return credentials in `storage-access-configs` when this header value is not set?

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -182,6 +186,26 @@ components:
         type: string
       example: "sales"
 
+    generic-table-data-access:
+      name: X-Generic-Table-Access-Delegation
+      in: header
+      description: >
+        Optional signal to the server that the client supports delegated access
+        via a comma-separated list of access mechanisms. The server may choose
+        to supply access via any or none of the requested mechanisms.
```
dimas-b: Re: `none` - in the Polaris IRC impl., if the client requests credential vending, but the server cannot provide them, the server will return an error response. 

Do we want the server to silently ignore credential vending requests in the Generic Tables API?

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -182,6 +186,26 @@ components:
...
+
+
+        When `vended-credentials` is included, the server may return scoped
+        storage credentials in the `storage-access-configs` field of the response.
+      required: false
+      schema:
+        type: string
+        enum:
```
dimas-b, line 201: `enum` means exactly one of many values can be specified, right? How can it work for a comma-separated list (line 194)?

spec/polaris-catalog-apis/generic-tables-api.yaml
flyrain: +1 on removing the "X-" prefix. Since it is a `generic-table-data-access`, maybe `Generic-Table-Access-Delegation` makes more sense. Besides, if any other catalogs implements the generic table REST spec, `Polaris-Access-Delegation` would not be an appropriate name. WDYT?

flyrain: > It could apply to views (hypothetically) too.

What do you mean by view? Iceberg view should be covered by the Iceberg parameters. And if there is generic table view, it would be covered by `Generic-Table-Access-Delegation` as well.

spec/polaris-catalog-apis/generic-tables-api.yaml
flyrain: +1 on it. I think it shouldn't support comma-separated list

spec/polaris-catalog-apis/generic-tables-api.yaml
dimas-b: I meant "view" as something that is not a table. If the Generic Tables API is mean only for tables, please ignore my comment :)

dimas-b: Re: prefix - "Polaris" is the project that owns the API spec, so using the project name as the namespace for headers seems logical. Using "Generic" as the prefix is too generic :wink: 

`Polaris-Generic-Table-Access-Delegation` would work, from my POV,  though.

My personal preference (non-binding) is with `Polaris-Access-Delegation` because it is shorter and could be extended to other APIs owned by Polaris.

flyrain: It is shorter. I think it's a good name. My only concern is that whether Polaris should  honor it for Iceberg tables, when clients send `Polaris-Access-Delegation`. If we are OK with it, I'm onboard.

gh-yzou: I think the direction of the “generic table” concept is still evolving, and it has the potential to expand beyond structured data to also cover unstructured data.

From my perspective, the naming carries important implications. If we call it generic-table-access-delegation, it suggests that the header is scoped specifically to generic table APIs and applies only to entities managed (if generic table api covers views in the future, it also applies) within that context. On the other hand, polaris-access-delegation implies a broader, cross-cutting concept within Polaris that could be reused by future APIs. As Yufei mentioned, this could also extend to use cases like Iceberg tables.

Since the overall direction is still not fully defined, I’m fine with either name for now, might more prefer to use generic-table prefix to restrict the scope, and let things evolve. However, I want to make sure we’re aligned on the underlying concept and intended scope before finalizing it.

@dimas-b @flyrain @adutra please let me know what is your preference.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: i think we should be consistent on this across polaris, i updated the description to be more clear

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: Yes, enum can only be one of them, the iceberg spec description is not accurate, i updated the description to be more accurate.

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: sorry, i was just trying to use a wording that is more neural. i updated to use *must* to be more clear.

spec/polaris-catalog-apis/generic-tables-api.yaml
flyrain: The `Polaris-Generic-Table-Access-Delegation` is more accurate, but either `Polaris-Generic-Table-Access-Delegation` or `Polaris-Access-Delegation` works for me.

jbonofre: I also have a preference for `Polaris-Generic-Table-Access-Delegation` for the completness.

dimas-b: Let's go with `Polaris-Generic-Table-Access-Delegation`

gh-yzou: Thanks all! i will update the title and code, and send out a vote

spec/polaris-catalog-apis/generic-tables-api.yaml
```diff
@@ -182,6 +186,26 @@ components:
...
 
+    polaris-generic-table-data-access:
+      name: Polaris-Generic-Table-Access-Delegation
+      in: header
+      description: >
+        Optional header indicating that the client supports delegated access
+        using a specific mechanism. If the header is present, the server must
+        either provide access using the requested mechanism or fail the request.
+
+
```
dimas-b: nit: two empty lines?

flyrain:
looks like this failed all PRs' CI
```
#4 [internal] load metadata for docker.io/apache/spark:3.5.6-java17-python3
#4 ERROR: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 500 Internal Server Error: an unknown error occurred during the request
------
 > [internal] load metadata for docker.io/apache/spark:3.5.6-java17-python3:
------
Dockerfile:20

--------------------

  18 |     #

  19 |     

  20 | >>> FROM docker.io/apache/spark:3.5.6-java17-python3

  21 |     

  22 |     ARG POLARIS_HOST=polaris

--------------------

failed to solve: docker.io/apache/spark:3.5.6-java17-python3: failed to resolve source metadata for docker.io/apache/spark:3.5.6-java17-python3: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 500 Internal Server Error: an unknown error occurred during the request
```

spec/polaris-catalog-apis/generic-tables-api.yaml
gh-yzou: removed

gh-yzou:
Nothing changed, run ./gradlew spotlessApply to fix a format issue

Issue #3222, opened 2025-12-05
[feature request] storing additional business-oriented metadata


kevinjqliu:
### Is your feature request related to a problem? Please describe.

Hey folks, 

I want to see if there's interest from the community to use Polaris also as a metadata catalog for storing additional business-oriented metadata. 

Examples of such metadata include:
- Table description
- Column desciption
- Owner information
- Creation date
- Tags and classifications
- Lineage details
- Data quality scores

This type of metadata differs from traditional Iceberg metadata, as it is primarily intended for business users, rather than query engines.

In some ways, [Polaris Policies](https://polaris.apache.org/in-dev/unreleased/policy/) already act as additional metadata since they can reference tables. Extending this concept to include richer metadata could make Polaris more valuable as a unified catalog for both technical and business contexts.

I’d love to start a conversation around this idea and see if there’s interest from the community to pursue it. Thoughts?

dimas-b:
Hi @kevinjqliu : as for me, this idea sounds useful and interesting in terms of opening new use cases. My only concern is ensuring that the new feature does not interfere with existing code paths. For example, storing new information in current Polaris entity properties will affect all table operations (more data to process) and may not be desirable. 

Would you mind opening a corresponding discussion thread on `dev`? I believe email is easier for discussing new feature design than GH comments 🙂

rishii-19-works:
Hi @kevinjqliu  I would like to contribute to this issue as part of my open-source learning journey

Before I begin working I want to confirm the expected scope of the change:-

1. Add a new enum value "ORDINAL" to "PolarisEventType"
2. Update all related usages and serialization/deserialization logic
3. Add or update unit tests to cover the new event type
4. Update any related documentation if needed

Please let me know if this understanding is correct or if there are additional requirements
If approved I'll start working on the implementation and submit a PR

Thank you!!

PR #4048, opened 2026-03-24
POC: Add labels to LoadTableResponse from entity internal properties

laskoviymishka:
Surfaces catalog-scoped metadata as labels in the IRC LoadTableResponse. Labels are derived from Polaris entity internalProperties with 'label.' prefix — these are catalog-managed, not part of Iceberg table state.

Addresses apache/polaris#3222 (business-oriented metadata in Polaris). Labels provide a standard IRC protocol mechanism for surfacing ownership, classification, domain, and other catalog context to any consuming engine.

New files: LoadTableResponseWithLabels.java (Jackson wrapper)
Modified: IcebergCatalogHandler (getLabelsForTable), IcebergCatalogAdapter (enrichment)

Proof-of-concept for the IRC Labels proposal: https://github.com/apache/iceberg/issues/15521.

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
@@ -684,6 +684,25 @@ public abstract class IcebergCatalogHandler extends CatalogHandler implements Au
     return null; // could be an external catalog
   }
 
+  /**
+   * Extract labels from a table entity's internal properties. Properties with the "label." prefix
+   * are catalog-scoped metadata (not part of Iceberg table state) and are surfaced as labels in the
+   * LoadTableResponse.
+   */
+  public Map<String, String> getLabelsForTable(TableIdentifier tableIdentifier) {
+    IcebergTableLikeEntity entity = getTableEntity(tableIdentifier);
+    if (entity == null) {
+      return Map.of();
+    }
+    Map<String, String> labels = new java.util.HashMap<>();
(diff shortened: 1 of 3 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -684,6 +684,25 @@ public void reportMetrics(TableIdentifier identifier, ReportMetricsRequest reque
...
+   * LoadTableResponse.
+   */
+  public Map<String, String> getLabelsForTable(TableIdentifier tableIdentifier) {
+    IcebergTableLikeEntity entity = getTableEntity(tableIdentifier);
+    if (entity == null) {
+      return Map.of();
+    }
+    Map<String, String> labels = new java.util.HashMap<>();
+    entity.getInternalPropertiesAsMap().forEach((key, value) -> {
+      if (key.startsWith("label.")) {
```
dimas-b, line 699: I do not think internal properties were meant to be exposed to clients at all, but using a prefix to filter them might be ok. This probably deserves a discussion on the Polaris `dev` ML.

dimas-b (commented):
Thanks for your contribution, @laskoviymishka ! Interesting idea 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
laskoviymishka: that's a POC, i think if this would be accepted at IRC Spec we may find a better way, but using a special prefix may be an acceptable solution too.

PR #3729, opened 2026-02-10
feat: Allows usage of Iceberg's GoogleAuthManager rather than just OAuth

PhillHenry:
feat: Be Able to Configure an External Catalog to Use org.apache.iceberg.gcp.auth.GoogleAuthManager as it's AuthManager (#3451). It allows an "End-User Credentials (Pass-Through) Approach" in that Polaris provides credentials for accessing storage rather than taking them from the BigLake metastore.

As part of the work, stopped a transitive dependency pulling in incompatible versions of proto-google-cloud-iamcredentials-v1 (protobuf 4.33.2) vs. protobuf-java (4.32.1). ProtobufSmokeTest should stop future regressions.

New DTO and DPO created but no changes were made to others.

PR #4050, opened 2026-03-24
Add BigQuery Metastore federation support

joyhaldar:
Adds federation support for BigQuery Metastore catalogs, enabling Polaris to serve as a unified REST catalog interface for Iceberg tables managed in BigQuery Metastore.

### Changes
- `BigQueryMetastoreFederatedCatalogFactory` - factory for creating BigQuery Metastore catalog handles
- `BigQueryMetastoreConnectionConfigInfoDpo` - connection configuration with all BigQueryProperties
- `BIGQUERY` connection type in `ConnectionType`
- OpenAPI schema for `BigQueryMetastoreConnectionConfigInfo`
- Support for service account impersonation via Iceberg's BigQueryProperties

## Checklist


runtime/server/build.gradle.kts
```diff
@@ -43,6 +43,10 @@ dependencies {
     runtimeOnly(project(":polaris-extensions-federation-hive"))
   }
 
+  if ((project.findProperty("NonRESTCatalogs") as String?)?.contains("BIGQUERY") == true) {
+    runtimeOnly(project(":polaris-extensions-federation-bigquery"))
```
dimas-b, line 47: This means `BIGQUERY` will NOT be included into the standard Apache Polaris distributions... Is that intended?

spec/polaris-management-service.yml
```diff
@@ -933,6 +934,7 @@ components:
           ICEBERG_REST: "#/components/schemas/IcebergRestConnectionConfigInfo"
           HADOOP: "#/components/schemas/HadoopConnectionConfigInfo"
           HIVE: "#/components/schemas/HiveConnectionConfigInfo"
+          BIGQUERY: "#/components/schemas/BigQueryMetastoreConnectionConfigInfo"
```
dimas-b, line 942: Please open a `dev` ML discussion for this (it's customary for all REST API changes).

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
```diff
@@ -0,0 +1,79 @@
...
+
+    String warehouse =
+        ((BigQueryMetastoreConnectionConfigInfoDpo) connectionConfigInfoDpo).getWarehouse();
+    Map<String, String> mergedProperties =
+        RESTUtil.merge(
+            catalogProperties != null ? catalogProperties : Map.of(),
+            connectionConfigInfoDpo.asIcebergCatalogProperties(polarisCredentialManager));
+
+    BigQueryMetastoreCatalog bigQueryMetastoreCatalog = new BigQueryMetastoreCatalog();
+    bigQueryMetastoreCatalog.initialize(warehouse, mergedProperties);
```
dimas-b, line 77: How / where does it get BigQuery connection credentials?

dimas-b (commented):
Thanks for your contribution, @joyhaldar ! The PR LGTM 👍 Still, I believe we need to open a `dev` discussion for it before merging as it includes REST API changes.

runtime/server/build.gradle.kts
nandorKollar: I think - unlike with Hadoop and Hive catalog - this doesn't bring in any new direct dependency, which we have concern with. We should be good to include this extension in the standard distribution no?

dimas-b: From my POV it seems reasonable to include it by default in Apache Polaris.

Downstream builds can make their own decisions since the impl. is well-isolated in terms of build modules.

joyhaldar:
> Thanks for your contribution, @joyhaldar ! The PR LGTM 👍 Still, I believe we need to open a `dev` discussion for it before merging as it includes REST API changes.

Thank you for your review @dimas-b. I have an existing dev ML thread for this. Replied with the PR link: https://lists.apache.org/thread/n3hh5s1zxn6yn7cbowfgf8p029z6m7g1

Would you like me to start a new discussion?

spec/polaris-management-service.yml
joyhaldar: Thank you for your review Dmitri.

I have an existing dev ML thread for this. Replied with the PR link: https://lists.apache.org/thread/n3hh5s1zxn6yn7cbowfgf8p029z6m7g1

Would you like me to start a new discussion?

runtime/server/build.gradle.kts
joyhaldar: Should I remove the conditional statement and include it by default in the standard distribution?

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: Thank you for your review Dmitri.

BigQuery Metastore uses Application Default Credentials via Iceberg's [BigQueryProperties](https://github.com/apache/iceberg/blob/main/bigquery/src/main/java/org/apache/iceberg/gcp/bigquery/BigQueryProperties.java#L146-L175).

Credentials are resolved automatically from the environment, `GOOGLE_APPLICATION_CREDENTIALS`, `gcloud auth`, or attached Service Account.

joyhaldar:
CI failure is unrelated to this PR, MongoDB testcontainer timeout if I understand correctly.

runtime/server/build.gradle.kts
dimas-b: Yes, please go with an unconditional dependency.

joyhaldar: Done.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,230 @@
...
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("connectionTypeCode", getConnectionTypeCode())
+        .add("uri", getUri())
+        .add("warehouse", warehouse)
+        .add("gcpProjectId", gcpProjectId)
+        .add("gcpLocation", gcpLocation)
+        .add("listAllTables", listAllTables)
+        .add("impersonateServiceAccount", impersonateServiceAccount)
```
jbonofre: Is it ok to include `impersonateServiceAccount` for `toString()` ?
It could include sensitive data I guess, so it could potentially land in log messages.
Maybe we should remove the account details or offuscate ?

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
```diff
@@ -0,0 +1,81 @@
...
+      ConnectionConfigInfoDpo connectionConfigInfoDpo,
+      PolarisCredentialManager polarisCredentialManager,
+      Map<String, String> catalogProperties) {
+
+    // Currently, Polaris supports BigQuery Metastore federation only via IMPLICIT authentication.
+    // Hence, prior to initializing the configuration, ensure that the catalog uses
+    // IMPLICIT authentication.
+    AuthenticationParametersDpo authenticationParametersDpo =
+        connectionConfigInfoDpo.getAuthenticationParameters();
+    if (authenticationParametersDpo.getAuthenticationTypeCode()
```
jbonofre, line 53: Should we include a `null` check here ?

If `connectionConfigInfoDpo.getAuthenticationParameters()` returns `null`, then `authenticationParametersDpo.getAuthenticationTypeCode()` will throw NPE.

Since the constructor marks `authenticationParameters` as `required = false` I guess it can happen.

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionConfigInfoDpo.java
```diff
@@ -210,6 +213,30 @@ public static ConnectionConfigInfoDpo fromConnectionConfigInfoModelWithSecrets(
...
                 null /*Service Identity Info*/);
         break;
+      case BIGQUERY:
+        BigQueryMetastoreConnectionConfigInfo bigqueryConfigModel =
+            (BigQueryMetastoreConnectionConfigInfo) connectionConfigurationModel;
+        authenticationParameters =
+            AuthenticationParametersDpo.fromAuthenticationParametersModelWithSecrets(
+                bigqueryConfigModel.getAuthenticationParameters(), secretReferences);
+        String bigqueryUri =
+            bigqueryConfigModel.getUri() != null
```
jbonofre, line 243: nit: If URI is `null`, we default to `https://bigquery.googleapis.com`. Maybe this default should live in DPO for visibility (or maybe OpenAPI as you started a discussion to update it 😄 ).

jbonofre (commented):
If this PR is solid, I have a concern about the potential NPE. Can you please clarify ?

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: Thank you for your review JB.

I copied this pattern from [HiveFederatedCatalogFactory](https://github.com/apache/polaris/blob/49766cca72e6a21ce21cc002a571d485a4586f44/extensions/federation/hive/src/main/java/org/apache/polaris/extensions/federation/hive/HiveFederatedCatalogFactory.java#L49) and [HadoopFederatedCatalogFactory](https://github.com/apache/polaris/blob/49766cca72e6a21ce21cc002a571d485a4586f44/extensions/federation/hadoop/src/main/java/org/apache/polaris/extensions/federation/hadoop/HadoopFederatedCatalogFactory.java#L50) which also don't have null checks. Should I add the null check here and open a separate PR to fix Hive and Hadoop as well? Or keep it consistent with existing code for now?

polaris-core/src/main/java/org/apache/polaris/core/connection/ConnectionConfigInfoDpo.java
joyhaldar: Done. I have moved the default URI to a public constant in `BigQueryMetastoreConnectionConfigInfoDpo`.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: Thank you for your review JB.

Done. I have removed **all impersonation fields** from `toString()`.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
dimas-b: So, if storage is also in GCP, the BigQuery connection and storage connections will have to use the same credentials, effectively (in current Polaris code).

It's not a big deal ATM, I think... just trying to understand the full picture :)

spec/polaris-management-service.yml
dimas-b: I opened https://lists.apache.org/thread/m05xm7szd7znrm9yos1rnld5ljx04y0h specifically for community awareness of REST API changes.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
dimas-b: I'd say let's make new code robust immediately and fix old code later.

dimas-b: Note: IIRC, Hive federation is not even bundled by default: https://github.com/apache/polaris/blob/b67e1a594f740c580fc95b9d7a6cb78810bf6aeb/runtime/server/build.gradle.kts#L42-L43

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,228 @@
...
+        listAllTables,
+        impersonateServiceAccount,
+        impersonateLifetimeSeconds,
+        impersonateScopes,
+        impersonateDelegates,
+        serviceIdentityInfo);
+  }
+
+  @Override
+  public ConnectionConfigInfo asConnectionConfigInfoModel(
```
dimas-b, line 105: Do we have a test to validate no data is lost on round-trips through these different config objects?

spec/polaris-management-service.yml
```diff
@@ -965,6 +967,44 @@ components:
...
+          type: array
+          items:
+            type: string
+          description: OAuth scopes for impersonation (default = cloud-platform)
+        impersonateDelegates:
+          type: array
+          items:
+            type: string
+          description: Delegation chain for impersonation
+      required:
```
flyingImer: IIUC, `BIGQUERY` can still omit `authenticationParameters` at the API boundary, but the current implementation only supports `IMPLICIT` auth and later code paths assume the field is present. Would it make sense to tighten the BIGQUERY contract here instead, so the boundary reflects what the runtime actually supports today?

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,228 @@
...
+        .setUri(getUri())
+        .setWarehouse(warehouse)
+        .setGcpProjectId(gcpProjectId)
+        .setGcpLocation(gcpLocation)
+        .setListAllTables(listAllTables)
+        .setImpersonateServiceAccount(impersonateServiceAccount)
+        .setImpersonateLifetimeSeconds(impersonateLifetimeSeconds)
+        .setImpersonateScopes(impersonateScopes)
+        .setImpersonateDelegates(impersonateDelegates)
+        .setAuthenticationParameters(
```
flyingImer, line 111: Related contract gap: the readback/model-conversion path also seems to assume `authenticationParameters` is non-null here. So even if the create/update boundary accepts a BIGQUERY payload without an auth block, we can still fail later when converting the stored config back to the API model.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
flyingImer: IIUC, This path seems to rely on `authenticationParameters` being present, while the public BIGQUERY contract still appears to allow it to be omitted. I wonder if that mismatch should be resolved at validation / schema time rather than here in the runtime path.

flyingImer (commented):
I’m aligned with the direction here overall. I think there are two high-priority issues worth tightening before this moves forward:

1. The current BIGQUERY auth contract does not look aligned yet. IIUC, the API/model path still allows `authenticationParameters` to be omitted, while the runtime only supports `IMPLICIT` auth and later dereferences that field as if it were always present. That feels like more than a local null-safety issue, the boundary contract still looks looser than the implementation.

2. Since this adds a new federation connector across OpenAPI, model conversion, and runtime wiring, I think we should have at least one round-trip/ wiring test for `BIGQUERY` before the contract settles.

2 cents: this PR also seems like a useful point to clarify whether federation here is aiming for a common Polaris contract over native Iceberg catalogs, or for backend-specific adapters behind one REST surface. I think that answer affects how strict/explicit the connector contract should be.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: Sorry about that. I have added `testBigQueryMetastoreConnectionConfig` in `ConnectionConfigInfoDpoTest`.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: I have added null check in `asConnectionConfigInfoModel`.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: Yes, that's my understanding also. Both BigQuery Metastore and GCS storage will use the same ADC credentials, or impersonated SA if configured. Thank you for clarifying the full picture. Sorry for the late reply.

spec/polaris-management-service.yml
joyhaldar: Thank you for the feedback. I have added null checks in both `BigQueryMetastoreFederatedCatalogFactory` and `asConnectionConfigInfoModel`. If `authenticationParameters` is null, BigQuery uses ADC by default. Please let me know if you'd like me to change the approach. Sorry for the late reply.

joyhaldar:
> I’m aligned with the direction here overall. I think there are two high-priority issues worth tightening before this moves forward:
> 
> 1. The current BIGQUERY auth contract does not look aligned yet. IIUC, the API/model path still allows `authenticationParameters` to be omitted, while the runtime only supports `IMPLICIT` auth and later dereferences that field as if it were always present. That feels like more than a local null-safety issue, the boundary contract still looks looser than the implementation.
> 2. Since this adds a new federation connector across OpenAPI, model conversion, and runtime wiring, I think we should have at least one round-trip/ wiring test for `BIGQUERY` before the contract settles.
> 
> 2 cents: this PR also seems like a useful point to clarify whether federation here is aiming for a common Polaris contract over native Iceberg catalogs, or for backend-specific adapters behind one REST surface. I think that answer affects how strict/explicit the connector contract should be.

Thank you for the detailed review. I have tried to address both points:
1. Added null checks in the factory and `asConnectionConfigInfoModel` to handle omitted `authenticationParameters`, which defaults to ADC.
2. Added `testBigQueryMetastoreConnectionConfig` in `ConnectionConfigInfoDpoTest` for round-trip validation.

Please let me know if you'd like any changes.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: Done. Added null check here. If `authenticationParameters` is null, BigQuery uses ADC by default.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,230 @@
...
+import org.apache.polaris.core.credentials.connection.ConnectionCredentials;
+import org.apache.polaris.core.identity.dpo.ServiceIdentityInfoDpo;
+import org.apache.polaris.core.identity.provider.ServiceIdentityProvider;
+
+/**
+ * The internal persistence-object counterpart to {@link
+ * org.apache.polaris.core.admin.model.BigQueryMetastoreConnectionConfigInfo} defined in the API
+ * model.
+ */
+public class BigQueryMetastoreConnectionConfigInfoDpo extends ConnectionConfigInfoDpo {
```
dimas-b, line 44: Just wondering: Could this class be `@PolarisImmutable`? I guess it might save a good chunk of boiler-plate code 🤔

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+      ConnectionConfigInfoDpo connectionConfigInfoDpo,
+      PolarisCredentialManager polarisCredentialManager,
+      Map<String, String> catalogProperties) {
+
+    // Currently, Polaris supports BigQuery Metastore federation only via IMPLICIT authentication.
+    // Hence, prior to initializing the configuration, ensure that the catalog uses
+    // IMPLICIT authentication.
+    AuthenticationParametersDpo authenticationParametersDpo =
+        connectionConfigInfoDpo.getAuthenticationParameters();
+    if (authenticationParametersDpo != null
```
dimas-b: What's the use case for permitting federated catalogs without an `AuthenticationParametersDpo` object?

spec/polaris-management-service.yml
```diff
@@ -965,6 +967,44 @@ components:
...
+        impersonateScopes:
+          type: array
+          items:
+            type: string
+          description: OAuth scopes for impersonation (default = cloud-platform)
+        impersonateDelegates:
+          type: array
+          items:
+            type: string
+          description: Delegation chain for impersonation
```
flyrain: How often does the BQMS config surface change? For example, when new fields are introduced or existing fields are updated. Locking every field into the spec creates a maintenance burden, since each upstream change requires a Polaris spec PR, client regeneration, and a new release.

Instead, we could consider using a bag of open properties to keep the model more flexible.
```
        properties:
          type: object
          additionalProperties:
            type: string
```

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
flyrain: Same concern here. Should we fail fast when `authenticationParametersDpo` is null?

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
flyrain: Thanks for the explanation, @joyhaldar. Can we comment out this "hidden" behavior somewhere?

flyrain (commented):
Thanks @joyhaldar for adding this. I think it would be really useful. The PR is in the right direction. Left some comments.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,230 @@
...
+  // BigQuery Metastore catalog property keys
+  private static final String GCP_BIGQUERY_PROJECT_ID = "gcp.bigquery.project-id";
+  private static final String GCP_BIGQUERY_LOCATION = "gcp.bigquery.location";
+  private static final String GCP_BIGQUERY_LIST_ALL_TABLES = "gcp.bigquery.list-all-tables";
+  private static final String IMPERSONATE_SERVICE_ACCOUNT =
+      "gcp.bigquery.impersonate.service-account";
+  private static final String IMPERSONATE_LIFETIME_SECONDS =
+      "gcp.bigquery.impersonate.lifetime-seconds";
+  private static final String IMPERSONATE_SCOPES = "gcp.bigquery.impersonate.scopes";
+  private static final String IMPERSONATE_DELEGATES = "gcp.bigquery.impersonate.delegates";
```
XJDKC: For these property keys, should we just use the iceberg sdk's properties?

https://github.com/apache/iceberg/blob/main/bigquery/src/main/java/org/apache/iceberg/gcp/bigquery/BigQueryProperties.java

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,230 @@
...
+
+  private final String warehouse;
+  private final String gcpProjectId;
+  private final String gcpLocation;
+  private final Boolean listAllTables;
+  private final String impersonateServiceAccount;
+  private final Integer impersonateLifetimeSeconds;
+  private final List<String> impersonateScopes;
+  private final List<String> impersonateDelegates;
```
XJDKC: These are authentication related properties, wandering if we need to create a new authentication for GCP?
```
  private final String impersonateServiceAccount;
  private final Integer impersonateLifetimeSeconds;
  private final List<String> impersonateScopes;
  private final List<String> impersonateDelegates;
```

XJDKC (commented):
Side Question: if we can federate to BigLake, we should be able to federate to BigQuery through BigLake right?

jbonofre (commented):
I have two major questions/"concerns":
1. As we introduce new dependencies (for BigQuery and transitive), the LICENSE/NOTICE should be updated according to the second point.
2. The question is: do we ship bigquery metastore federation by default in the server ? If yes, the LICENSE/NOTICE of the server should be updated. If no, no need to update it.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,230 @@
...
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("connectionTypeCode", getConnectionTypeCode())
+        .add("uri", getUri())
+        .add("warehouse", warehouse)
+        .add("gcpProjectId", gcpProjectId)
+        .add("gcpLocation", gcpLocation)
+        .add("listAllTables", listAllTables)
+        .add("authenticationParameters", getAuthenticationParameters())
```
jbonofre: I know I had a comment in this `toString()` method, and you addressed it. Thanks a lot.

Sorry to be a pain, but I have a follow up question 😄 

Here we directly call `getAuthenticationParameters()`. `authenticationParameters` is `@Nullable`.
It's similar to `HiveConnectionConfigInfoDpo` which has the same nullable pattern.
If other DPOs call `toString()` on the auth params, this could throw NPE.

The existing Hadoop DPO has `authenticationParameters` as `@Nonnull` and calls `toString()` explicitly. 

Maybe we should do something similar here to prevent NPE ?

We can do something simple:

```suggestion
  .add("authenticationParameters",
      getAuthenticationParameters() != null ? getAuthenticationParameters().toString() : "null")  
```

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: BigQuery always uses [Application Default Credentials](https://docs.cloud.google.com/docs/authentication/application-default-credentials) or the [GOOGLE_APPLICATION_CREDENTIALS](https://docs.cloud.google.com/docs/authentication/application-default-credentials#GAC) environment variable to automatically pick up the user's credentials or a JSON file pointing to their Service Account. Making `authenticationParameters` required would just force users to type IMPLICIT for no reason IMO, but please let me know what you think, or please let me know if you think I am incorrect.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: The existing `ConnectionConfigInfoDpo` classes (Hadoop, Hive) don't use `@PolarisImmutable`. I can try to refactor and use it if you think it's worthwhile. Would appreciate your inputs before I proceed.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: The property keys in Iceberg's [BigQueryProperties](https://github.com/apache/iceberg/blob/main/bigquery/src/main/java/org/apache/iceberg/gcp/bigquery/BigQueryProperties.java) are package-private, so I had to duplicate them here. I can add a comment referencing the Iceberg source. Let me know what you think.

spec/polaris-management-service.yml
joyhaldar: You are right, I will work on addressing this.

joyhaldar: Done. Refactored to use a properties bag. Only `warehouse` and `gcpProjectId` remain as required fields, everything else goes into the map. Thank you for your help.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: Thank you for the review. These fields are now part of the [properties bag](https://github.com/apache/polaris/pull/4050#discussion_r3029516972).

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: Done. Added the null check as suggested. Thank you again JB!

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
flyrain: I think we will need an `AuthenticationType` regardless of how BigQuery picks up credentials. In this case, it should be `IMPLICIT`(I agreed with line 58 here). However, from a syntax perspective, we may not want users to specify it explicitly. We could default it when the `AuthenticationType` is missing. That said, I would not recommend handling this defaulting logic here. The defaulting to `IMPLICIT` should happen elsewhere. WDYT?

flyrain (commented):
Thanks for continuous working on it, @joyhaldar! I think we are getting closer. Left some comments. Could you fix the CI failures as well?

A few followup items(not a blocker):
1. Document this new feature
2. CLI changes to support BQMS.

runtime/server/build.gradle.kts
joyhaldar: Reverted to optional for now. Making it unconditional was breaking CI due to LICENSE/NOTICE requirements for the new BigQuery dependencies. I can make it unconditional once the license entries are sorted out, wanted to unblock the PR first.

joyhaldar:
> I have two major questions/"concerns":
> 
> 1. As we introduce new dependencies (for BigQuery and transitive), the LICENSE/NOTICE should be updated according to the second point.
> 2. The question is: do we ship bigquery metastore federation by default in the server ? If yes, the LICENSE/NOTICE of the server should be updated. If no, no need to update it.

Made it optional for now to unblock CI. Will follow up on LICENSE/NOTICE updates separately if we decide to ship it by default.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: Done. Added a comment explaining the ADC credential behavior in [this commit](https://github.com/apache/polaris/pull/4050/changes/cef1b0fd0976fc313118df5d96e1df87c37e89b3).

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
```diff
@@ -65,6 +65,8 @@ public Catalog createCatalog(
             catalogProperties != null ? catalogProperties : Map.of(),
             connectionConfigInfoDpo.asIcebergCatalogProperties(polarisCredentialManager));
 
+    // Credentials are resolved via Google Application Default Credentials (ADC).
+    // GCS storage operations use the same ADC credentials.
```
dimas-b: The second statement is not 100% accurate. In principle a user can override GCS storage credentials:

https://github.com/apache/polaris/blob/a6ca7cb2a27eb685522c8dc09d15217683714fde/runtime/service/src/main/java/org/apache/polaris/service/storage/StorageConfiguration.java#L151

I'd propose adding `... by default`.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
dimas-b: No worries. Let's keep existing pattern.

runtime/server/build.gradle.kts
dimas-b: Sounds good. Resolving this comment thread.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
dimas-b: I believe declaring the `IMPLICIT` authentication explicitly 😉 is actually good. It increases clarity about the intended catalog behaviour. It is also more robust in case of future API and backend code changes.

flyrain: Agreed, the point is even we want a default value for better UX. This is not the place to handle the defaulting logic.

flyingImer (commented):
This is directionally better now, but I still think there are two blocker-level contract issues here.

1. BIGQUERY can still accept auth modes that the connector itself does not actually support. The connector is still IMPLICIT-only, but the create/update path is validating auth too generically, so an invalid BIGQUERY catalog can still be accepted and persisted, then fail later at catalog instantiation time.

2. The `authenticationParameters` contract is still inconsistent. The schema/DPO/factory path still treats it as effectively optional, but the create/admin validation path dereferences it as if it were required. So right now a missing auth block can still turn into a null-deref path instead of a deterministic validation/defaulting decision.

My preference would be to make this narrower and explicit before merge:
- validate BIGQUERY auth per connection type at create/update time, not only later in the connector, and
- pick one contract for missing `authenticationParameters` (`required + IMPLICIT` or `missing means IMPLICIT`) and make the spec/validation/DPO path all match.

spec/polaris-management-service.yml
```diff
@@ -965,6 +967,27 @@ components:
...
+          type: string
+          description: The warehouse location for the BigQuery Metastore Catalog
+        gcpProjectId:
+          type: string
+          description: The GCP project ID
+        properties:
+          type: object
+          additionalProperties:
+            type: string
+            description: Additional catalog properties
```
flyrain: Thanks for adding this. Thinking more broadly, this `properties` would be useful for other federated catalog type like HMS. I'd proposed to move it up to its parent `ConnectionConfigInfo`, so that all subtypes can benefit from it. WDYT? cc @XJDKC @HonahX @singhpk234 @dimas-b

dimas-b: Yes, the more I think about it the more I like the idea of supporting general `properties` inside `ConnectionConfigInfo`. That REST API change should be able to support both  BigQuery Catalog federation and #3729 (CC: @PhillHenry).

flyrain: Here is another use case that justifies it. A client for the remote catalog may support more properties than those defined in the spec. Without a freeform `properties` field, we would need to keep updating the spec to accommodate properties already supported by existing clients (for example `HMS` or `BQMS`), which is both unnecessary and heavy. With properties, users can simply update their federated catalog in the metastore instead.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: Updated the comment to say `by default`.

spec/polaris-management-service.yml
joyhaldar: I have moved `properties` to parent `ConnectionConfigInfo` so all federation types can use it.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Override
+  public @Nonnull Map<String, String> asIcebergCatalogProperties(
+      PolarisCredentialManager polarisCredentialManager) {
+    HashMap<String, String> properties = new HashMap<>();
+
+    // Required properties for BigQueryMetastoreCatalog
+    properties.put(GCP_BIGQUERY_PROJECT_ID, gcpProjectId);
+    properties.put(CatalogProperties.WAREHOUSE_LOCATION, warehouse);
```
dimas-b: Why not use the generic `getProperties()` map (line 108) for these parameter too?

flyrain (dismissed):
LGTM. Thanks @joyhaldar !

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: These are required fields in the spec. `projectId` is validated as required in [BigQueryProperties.java](https://github.com/apache/iceberg/blob/9a939d68358de9dac2c6ba9b236b675ebe477490/bigquery/src/main/java/org/apache/iceberg/gcp/bigquery/BigQueryProperties.java#L104-L106), and `warehouse` is validated in [BigQueryMetastoreCatalog.java](https://github.com/apache/iceberg/blob/9a939d68358de9dac2c6ba9b236b675ebe477490/bigquery/src/main/java/org/apache/iceberg/gcp/bigquery/BigQueryMetastoreCatalog.java#L310-L314). The properties bag is for optional settings like `gcp.bigquery.location`. But let me know if this is not appropriate, would love your opinion.

runtime/server/build.gradle.kts
jbonofre: I can help on the LICENSE/NOTICE. I'm fine to do it in a follow up PR.

jbonofre (dismissed):

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
dimas-b: From my POV this creates a skew in the Polaris API, which becomes dependent on particular Iceberg java classes.

Conceptually, this class merely passes some user-defined properties to the Iceberg catalog impl. This is fine. However, do we have to elevate a sub-set of those properties to explicit Polaris OpenAPI properties?.. I'm not sure.

I believe keeping the Polaris API generic and uniform for all federated catalog properties might be preferable.

Implementations (both Iceberg java classes and Polaris classes) can and should perform config validation as appropriate, of course.

dimas-b (commented):
The PR LGTM overall 👍 

Feel free to consider my point about properties as non-blocking.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: Can I do this in a follow-up PR?

I'm trying to figure out the right pattern here, [Hadoop](https://github.com/apache/polaris/blob/main/extensions/federation/hadoop/src/main/java/org/apache/polaris/extensions/federation/hadoop/HadoopFederatedCatalogFactory.java) and [Hive](https://github.com/apache/polaris/blob/main/extensions/federation/hive/src/main/java/org/apache/polaris/extensions/federation/hive/HiveFederatedCatalogFactory.java) have `warehouse` as an explicit field, but `BigQueryMetastoreCatalog` needs both `warehouse` and `projectId`. Should I:

1. Keep `warehouse` explicit (consistent with `Hadoop`/`Hive`), move `projectId` to properties.
2. Move both to properties (fully generic, but diverges from `Hadoop`/`Hive` pattern)
3. Something else?

Would appreciate your guidance.

dimas-b: Retracting API changes is not ideal (in a follow-up). 

I'd propose to remove `gcpProjectId` and `warehouse` from OpenAPI _explicit_ properties and process corresponding Iceberg properties from the generic `properties` bag. Also, add description to OpenAPI YAML to list required properties.

My rationale is that it is fine to accept pass-through properties destined to become Iceberg (federated) catalog properties. Yet, specific processing / validation is a feature of the implementation and can change in new Polaris versions without requiring API changes.

I wonder what other reviewers think about this.

jbonofre: @dimas-b proposal works for me, as soon as clear documented. That said, I'm also fine to keep them as it's "bigquery" specific.

dimas-b: Re: Hadoop and Hive: from my POV the same argument applies to them, but I missed out on the review of that code when it was introduced :sweat_smile: 

This is why my request in this thread is non-blocking.

However, I do not really see a reason to promote an old pattern if we're in agreement about the rationale for generic properties in this case (hence my request for more reviewer opinions).

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
flyrain: @joyhaldar, sorry to miss this one, I think we will actually need this to be resolved. I'd suggest to add it to the method `fromConnectionConfigInfoModelWithSecrets()`, something like this
```
  case BIGQUERY:
    BigQueryMetastoreConnectionConfigInfo bigqueryConfigModel =
        (BigQueryMetastoreConnectionConfigInfo) connectionConfigurationModel;

    // Default to IMPLICIT authentication if not provided
    authenticationParameters =
        bigqueryConfigModel.getAuthenticationParameters() != null
            ? AuthenticationParametersDpo.fromAuthenticationParametersModelWithSecrets(
                  bigqueryConfigModel.getAuthenticationParameters(), secretReferences)
            : new ImplicitAuthenticationParametersDpo();
```

flyrain (commented):
The defaulting logic need to be moved

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: I have moved `warehouse` and `gcpProjectId` to the properties map with validation in the factory and documented required properties in the spec in [this commit](https://github.com/apache/polaris/pull/4050/changes/33ae93a630e15736b29c989cc3a2b98a418ff712).

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: No, I am sorry, I missed it.

I tried adding the defaulting in `fromConnectionConfigInfoModelWithSecrets()` but it doesn't work, `PolarisServiceImpl.validateAuthenticationParameters()` runs before conversion and throws NPE when `authenticationParameters` is null IIUC.

Fixing that would mean changing `PolarisServiceImpl too`, which seems out of scope for this PR.
For now I'm going with @dimas-b's suggestion, users must pass `authenticationParameters: {authenticationType: IMPLICIT}` explicitly ([commit](https://github.com/apache/polaris/pull/4050/changes/3856a7d5a0d5f3ccd8406037614013615c5a67ce)).

Let me know if my understanding is correct. Would appreciate your guidance.

polaris-core/src/main/java/org/apache/polaris/core/connection/hadoop/HadoopConnectionConfigInfoDpo.java
```diff
@@ -51,7 +51,8 @@ public HadoopConnectionConfigInfoDpo(
       @JsonProperty(value = "serviceIdentity", required = false) @Nullable
           ServiceIdentityInfoDpo serviceIdentityInfo,
       @JsonProperty(value = "warehouse", required = false) @Nullable String remoteCatalogName) {
-    super(ConnectionType.HADOOP.getCode(), uri, authenticationParameters, serviceIdentityInfo);
+    super(
+        ConnectionType.HADOOP.getCode(), uri, authenticationParameters, serviceIdentityInfo, null);
```
nandorKollar: nit: I think we can spare adding these ugly 'null' parameters in the super calls. We can overload the public constructor of `ConnectionConfigInfoDpo` like this:

```
  public ConnectionConfigInfoDpo(
          @JsonProperty(value = "connectionTypeCode", required = true) int connectionTypeCode,
          @JsonProperty(value = "uri", required = true) @Nonnull String uri,
          @JsonProperty(value = "authenticationParameters", required = true) @Nullable
          AuthenticationParametersDpo authenticationParameters,
          @JsonProperty(value = "serviceIdentity", required = false) @Nullable
          ServiceIdentityInfoDpo serviceIdentity,
          @JsonProperty(value = "properties", required = false) @Nullable
          Map<String, String> properties) {
    this(connectionTypeCode, uri, authenticationParameters, serviceIdentity, true, properties);
  }

  public ConnectionConfigInfoDpo(
      @JsonProperty(value = "connectionTypeCode", required = true) int connectionTypeCode,
      @JsonProperty(value = "uri", required = true) @Nonnull String uri,
      @JsonProperty(value = "authenticationParameters", required = true) @Nullable
          AuthenticationParametersDpo authenticationParameters,
      @JsonProperty(value = "serviceIdentity", required = false) @Nullable
          ServiceIdentityInfoDpo serviceIdentity) {
    this(connectionTypeCode, uri, authenticationParameters, serviceIdentity, true, null);
  }
```

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
```diff
@@ -0,0 +1,89 @@
...
+      throw new IllegalStateException(
+          "BigQuery Metastore federation only supports IMPLICIT authentication.");
+    }
+
+    Map<String, String> properties = connectionConfigInfoDpo.getProperties();
+    String warehouse = properties.get("warehouse");
+    if (warehouse == null || warehouse.isEmpty()) {
+      throw new IllegalArgumentException("warehouse is required for BigQuery Metastore federation");
+    }
+    if (properties.get("gcp.bigquery.project-id") == null) {
```
nandorKollar: Is empty string accepted for `gcp.bigquery.project-id`?

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.polaris.core.identity.provider.ServiceIdentityProvider;
+
+/**
+ * The internal persistence-object counterpart to {@link
+ * org.apache.polaris.core.admin.model.BigQueryMetastoreConnectionConfigInfo} defined in the API
+ * model.
+ */
+public class BigQueryMetastoreConnectionConfigInfoDpo extends ConnectionConfigInfoDpo {
+
+  public static final String DEFAULT_URI = "https://bigquery.googleapis.com";
```
nandorKollar, line 46: nit: I think we can keep this private.

polaris-core/src/main/java/org/apache/polaris/core/connection/hadoop/HadoopConnectionConfigInfoDpo.java
joyhaldar: I have added the overloaded constructor and updated the subclasses.

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
joyhaldar: I have added empty string validation for `gcp.bigquery.project-id`.

polaris-core/src/main/java/org/apache/polaris/core/connection/bigquery/BigQueryMetastoreConnectionConfigInfoDpo.java
joyhaldar: Actually it's public because it's used in `ConnectionConfigInfoDpo.fromConnectionConfigInfoModelWithSecrets()` for URI defaulting. I kept it in the BQMS DPO since it's specific to BQMS. But let me know if you disagree, would appreciate your inputs.

nandorKollar: Oh okay, I missed that it is used there. Makes sense, thanks!

nandorKollar (approved):
I'm not an expert of BigQuery, but overall LGTM.

flyrain (approved):

spec/polaris-management-service.yml
```diff
@@ -965,6 +972,16 @@ components:
         warehouse:
           type: string
           description: The warehouse location for the hive catalog.
+    
+    BigQueryMetastoreConnectionConfigInfo:
+      type: object
+      description: |
+        Configuration necessary for connecting to a BigQuery Metastore Catalog.
```
dimas-b, line 979: nit: maybe make a note that `properties` are effectively Iceberg's BigQuery catalog properties (for clarity)?

extensions/federation/bigquery/src/main/java/org/apache/polaris/extensions/federation/bigquery/BigQueryMetastoreFederatedCatalogFactory.java
```diff
@@ -0,0 +1,90 @@
...
+      throw new IllegalStateException(
+          "BigQuery Metastore federation only supports IMPLICIT authentication.");
+    }
+
+    Map<String, String> properties = connectionConfigInfoDpo.getProperties();
+    String warehouse = properties.get("warehouse");
+    if (warehouse == null || warehouse.isEmpty()) {
+      throw new IllegalArgumentException("warehouse is required for BigQuery Metastore federation");
+    }
+    String projectId = properties.get("gcp.bigquery.project-id");
```
dimas-b, line 64: nit: make a constant of refer to a constant from Iceberg jars?

dimas-b (approved):
LGTM 👍 Thanks for bearing with me 🙂

joyhaldar:
> LGTM 👍 Thanks for bearing with me 🙂

No, thank you for taking the time to thoroughly review my PR and giving me so much constructive feedback!

flyrain:
Merge is blocked by Github CI. Re-triggered. I will merge it today if there is no more feedback.

dimas-b:
I believe there is an option to re-run failed CI jobs without closing/reopening the PR :thinking: 

+1 to merging today.

flyrain:
Thanks  @joyhaldar for the change! It's very useful feature. Thanks everyone for the review. 

> I believe there is an option to re-run failed CI jobs without closing/reopening the PR 🤔

@dimas-b, got it. Mind sharing the option?

joyhaldar:
> Thanks for continuous working on it, @joyhaldar! I think we are getting closer. Left some comments. Could you fix the CI failures as well?
> 
> A few followup items(not a blocker):
> 
> 1. Document this new feature
> 2. CLI changes to support BQMS.

Sorry for the delay, I will follow-up on the remaining work very soon.

Issue #577, opened 2024-12-17
Amazon S3 Tables Integration


flyrain:
### Describe the solution you'd like

Polaris is designed to act as a **REST facade for S3 tables**, enabling both **read** and **write** operations by interacting with the S3 table API. Polaris registers an S3 table using its **metadata location**. A flag will be needed to label the new Iceberg table is a s3 table. Below is a summary of the proposed approach:
### **Read Path**  
- The **LoadTable** endpoint in Polaris will call the S3 table API `get_table_metadata_location` to fetch the `metadata.json` file.  
- Polaris will then **serialize** the content of the `metadata.json` into a **LoadTableResponse** to return to the client.

### **Write Path**  
When the table is updated, Polaris will:  
1. Gather the changes and generate a new `metadata.json`.  
2. Use the S3 table API [`update-table-metadata-location`](https://docs.aws.amazon.com/cli/latest/reference/s3tables/update-table-metadata-location.html) to commit the new metadata.  

### **AuthZ/AuthN**
We need to ensure that the AWS role used for creating the Polaris catalog has the read and write privileges of the s3 table.

puchengy:
@flyrain Thank you for starting the issue, I have two questions:
1. is it right that with this integration, users will continue to have the option to own the metastore service (that holds source of the truth data)?
2. should it be a concern if "update-table-metadata-location" fail and cause table metadata out of the sync, which could mislead S3 Tables to clean up data as orphan files.

flyrain:
> is it right that with this integration, users will continue to have the option to own the metastore service (that holds source of the truth data)?

That's right. This integration won't change the source of truth(s3 table in this case), and other tools or pipelines against the source catalog should still work as is.

> should it be a concern if "update-table-metadata-location" fail and cause table metadata out of the sync, which could mislead S3 Tables to clean up missing metadata data (and its data files) as orphan files?

Polaris client will get the failure message in that case, then it can retry or just fail itself. The table is still consistent. But the failure may leave orphan files, which is fine as the other clients also leave orphan files in case of failure. 

Thanks for pointing out the wrong link. Updated.

puchengy:
@flyrain 

> Polaris client will get the failure message in that case, then it can retry or just fail itself. The table is still consistent.

Do you mean that "update-table-metadata-location" will be part of the commit to the source catalog (e.g. HMS)?

flyrain:
[...]

PR #4052, opened 2026-03-24
feat: Add S3 Tables credential vending support for federated catalogs

aritragster:
## Summary

When Polaris federates to an S3 Tables Iceberg REST endpoint, the credential vending flow must generate S3 Tables IAM policies (using `s3tables:` actions and table-level ARNs) instead of standard S3 policies.

This is a draft PR to socialize the approach with the community. Feedback on the design is welcome.

Related to #577

## Problem

S3 Tables uses a different IAM action namespace (`s3tables:`) and ARN format (`arn:aws:s3tables:REGION:ACCOUNT:bucket/BUCKET/table/TABLE_ID`) compared to standard S3. When Polaris federates to an S3 Tables Iceberg REST endpoint, the existing credential vending flow generates S3 policies that don't work for S3 Tables data access.

## Approach

- Detect S3 Tables catalogs via the `signingName` property (`"s3tables"`) from the remote endpoint's connection configuration, persisted on the catalog entity
- Capture the `tableId` returned in `loadTable` responses from the remote endpoint using a `ConfigCapturingHTTPClient` wrapper and request-scoped `CapturedConfigHolder`
- Construct proper S3 Tables ARNs from the catalog's base location and captured `tableId`
- Generate per-table scoped session policies with granular read/write actions:
  - Read: `s3tables:GetTableData`, `s3tables:GetTableMetadataLocation`
  - Write (additional): `s3tables:UpdateTableMetadataLocation`, `s3tables:PutTableData`
- Validate at catalog creation time that `signingName: "s3tables"` catalogs use ARN-based `default-base-location` (not `s3://` paths)

## Files Changed (15 files, ~570 lines)

- `AwsCredentialsStorageIntegration` — S3 Tables policy generation, branching on `signingName`
- `CredentialVendingContext` — added `resourceArns` and `signingName` fields
- `AwsStorageConfigurationInfo` — added `signingName` getter
- `CatalogEntity` — `setSigningNameFromConnectionConfig()` + validation
- `CapturedConfigHolder` / `ConfigCapturingHTTPClient` — new classes to capture `tableId` from remote responses
- `IcebergCatalogHandler` — ARN construction from captured config
- `StorageAccessConfigProvider` — pass `resourceArns` through to credential vending
- `IcebergRESTExternalCatalogFactory` — wire in config capture
- `PolarisAdminService` — persist `signingName` at catalog creation
- Test: 3 new unit tests for S3 Tables policy generation (read-only, read-write, fallback)

## AI Disclosure

This implementation was developed with AI assistance (Kiro).

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Related to #577

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
@@ -161,6 +161,18 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
             .setKmsUnavailable(awsConfig.getKmsUnavailable())
             .build();
       }
+      if (configInfo instanceof AwsS3TablesStorageConfigurationInfo s3TablesConfig) {
+        return AwsS3TablesStorageConfigInfo.builder()
+            .setRoleArn(s3TablesConfig.getRoleARN())
+            .setExternalId(s3TablesConfig.getExternalId())
+            .setRegion(s3TablesConfig.getRegion())
+            .setCurrentKmsKey(s3TablesConfig.getCurrentKmsKey())
+            .setAllowedKmsKeys(s3TablesConfig.getAllowedKmsKeys())
+            .setStorageType(StorageConfigInfo.StorageTypeEnum.S3_TABLES)
+            .setAllowedLocations(s3TablesConfig.getAllowedLocations())
+            .setStorageName(s3TablesConfig.getStorageName())
+            .build();
+      }
       if (configInfo instanceof AzureStorageConfigurationInfo azureConfig) {
         return AzureStorageConfigInfo.builder()
             .setTenantId(azureConfig.getTenantId())
@@ -185,18 +197,6 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
             .setStorageType(StorageConfigInfo.StorageTypeEnum.FILE)
             .setAllowedLocations(fileConfigModel.getAllowedLocations())
             .setStorageName(fileConfigModel.getStorageName())
-            .build();
-      }
-      if (configInfo instanceof AwsS3TablesStorageConfigurationInfo s3TablesConfig) {
-        return AwsS3TablesStorageConfigInfo.builder()
-            .setRoleArn(s3TablesConfig.getRoleARN())
-            .setExternalId(s3TablesConfig.getExternalId())
-            .setRegion(s3TablesConfig.getRegion())
-            .setCurrentKmsKey(s3TablesConfig.getCurrentKmsKey())
-            .setAllowedKmsKeys(s3TablesConfig.getAllowedKmsKeys())
-            .setStorageType(StorageConfigInfo.StorageTypeEnum.S3_TABLES)
-            .setAllowedLocations(s3TablesConfig.getAllowedLocations())
-            .setStorageName(s3TablesConfig.getStorageName())
             .build();
       }
       return null;
@@ -346,7 +346,8 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
                     .build();
             break;
           case S3_TABLES:
-            if (defaultBaseLocation == null || !defaultBaseLocation.startsWith("arn:")) {
+            if (defaultBaseLocation == null
+                || !defaultBaseLocation.startsWith("arn:aws:s3tables")) {
               throw new BadRequestException(
                   "S3 Tables catalogs require default-base-location to be an S3 Tables "
                       + "bucket ARN (e.g. arn:aws:s3tables:us-east-1:123456789012:bucket/my-bucket)"
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
@@ -317,6 +317,29 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
                     .build();
             config = awsConfig;
             break;
+          case S3_TABLES:
+            if (defaultBaseLocation == null
+                || !defaultBaseLocation.startsWith(
+                    PolarisStorageConfigurationInfo.S3_TABLES_ARN_PREFIX)) {
+              throw new BadRequestException(
+                  "S3 Tables catalogs require default-base-location to be an S3 Tables "
+                      + "bucket ARN (e.g. arn:aws:s3tables:us-east-1:123456789012:bucket/my-bucket)"
+                      + ", but got: %s",
+                  defaultBaseLocation);
+            }
+            AwsS3TablesStorageConfigInfo s3TablesConfigModel =
+                (AwsS3TablesStorageConfigInfo) storageConfigModel;
+            config =
+                AwsS3TablesStorageConfigurationInfo.builder()
+                    .allowedLocations(allowedLocations)
+                    .storageName(storageConfigModel.getStorageName())
+                    .roleARN(s3TablesConfigModel.getRoleArn())
+                    .externalId(s3TablesConfigModel.getExternalId())
+                    .region(s3TablesConfigModel.getRegion())
+                    .currentKmsKey(s3TablesConfigModel.getCurrentKmsKey())
+                    .allowedKmsKeys(s3TablesConfigModel.getAllowedKmsKeys())
+                    .build();
+            break;
           case AZURE:
             AzureStorageConfigInfo azureConfigModel = (AzureStorageConfigInfo) storageConfigModel;
             config =
@@ -343,28 +366,6 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
                 FileStorageConfigurationInfo.builder()
                     .allowedLocations(allowedLocations)
                     .storageName(storageConfigModel.getStorageName())
-                    .build();
-            break;
-          case S3_TABLES:
-            if (defaultBaseLocation == null
-                || !defaultBaseLocation.startsWith("arn:aws:s3tables")) {
-              throw new BadRequestException(
-                  "S3 Tables catalogs require default-base-location to be an S3 Tables "
-                      + "bucket ARN (e.g. arn:aws:s3tables:us-east-1:123456789012:bucket/my-bucket)"
-                      + ", but got: %s",
-                  defaultBaseLocation);
-            }
-            AwsS3TablesStorageConfigInfo s3TablesConfigModel =
-                (AwsS3TablesStorageConfigInfo) storageConfigModel;
-            config =
-                AwsS3TablesStorageConfigurationInfo.builder()
-                    .allowedLocations(allowedLocations)
-                    .storageName(storageConfigModel.getStorageName())
-                    .roleARN(s3TablesConfigModel.getRoleArn())
-                    .externalId(s3TablesConfigModel.getExternalId())
-                    .region(s3TablesConfigModel.getRegion())
-                    .currentKmsKey(s3TablesConfigModel.getCurrentKmsKey())
-                    .allowedKmsKeys(s3TablesConfigModel.getAllowedKmsKeys())
                     .build();
             break;
           default:
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import org.apache.iceberg.exceptions.BadRequestException;
+import org.apache.polaris.core.admin.model.AwsS3TablesStorageConfigInfo;
 import org.apache.polaris.core.admin.model.AwsStorageConfigInfo;
 import org.apache.polaris.core.admin.model.AzureStorageConfigInfo;
 import org.apache.polaris.core.admin.model.Catalog;
@@ -48,6 +49,7 @@ import org.apache.polaris.core.identity.provider.ServiceIdentityProvider;
 import org.apache.polaris.core.secrets.SecretReference;
 import org.apache.polaris.core.storage.FileStorageConfigurationInfo;
 import org.apache.polaris.core.storage.PolarisStorageConfigurationInfo;
+import org.apache.polaris.core.storage.aws.AwsS3TablesStorageConfigurationInfo;
 import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
 import org.apache.polaris.core.storage.azure.AzureStorageConfigurationInfo;
 import org.apache.polaris.core.storage.gcp.GcpStorageConfigurationInfo;
@@ -183,6 +185,18 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
             .setStorageType(StorageConfigInfo.StorageTypeEnum.FILE)
             .setAllowedLocations(fileConfigModel.getAllowedLocations())
             .setStorageName(fileConfigModel.getStorageName())
+            .build();
+      }
+      if (configInfo instanceof AwsS3TablesStorageConfigurationInfo s3TablesConfig) {
+        return AwsS3TablesStorageConfigInfo.builder()
+            .setRoleArn(s3TablesConfig.getRoleARN())
+            .setExternalId(s3TablesConfig.getExternalId())
+            .setRegion(s3TablesConfig.getRegion())
+            .setCurrentKmsKey(s3TablesConfig.getCurrentKmsKey())
+            .setAllowedKmsKeys(s3TablesConfig.getAllowedKmsKeys())
+            .setStorageType(StorageConfigInfo.StorageTypeEnum.S3_TABLES)
+            .setAllowedLocations(s3TablesConfig.getAllowedLocations())
+            .setStorageName(s3TablesConfig.getStorageName())
             .build();
       }
       return null;
@@ -329,6 +343,27 @@ public class CatalogEntity extends PolarisEntity implements LocationBasedEntity 
                 FileStorageConfigurationInfo.builder()
                     .allowedLocations(allowedLocations)
                     .storageName(storageConfigModel.getStorageName())
+                    .build();
+            break;
+          case S3_TABLES:
+            if (defaultBaseLocation == null || !defaultBaseLocation.startsWith("arn:")) {
+              throw new BadRequestException(
+                  "S3 Tables catalogs require default-base-location to be an S3 Tables "
+                      + "bucket ARN (e.g. arn:aws:s3tables:us-east-1:123456789012:bucket/my-bucket)"
+                      + ", but got: %s",
+                  defaultBaseLocation);
+            }
+            AwsS3TablesStorageConfigInfo s3TablesConfigModel =
+                (AwsS3TablesStorageConfigInfo) storageConfigModel;
+            config =
+                AwsS3TablesStorageConfigurationInfo.builder()
+                    .allowedLocations(allowedLocations)
+                    .storageName(storageConfigModel.getStorageName())
+                    .roleARN(s3TablesConfigModel.getRoleArn())
+                    .externalId(s3TablesConfigModel.getExternalId())
+                    .region(s3TablesConfigModel.getRegion())
+                    .currentKmsKey(s3TablesConfigModel.getCurrentKmsKey())
+                    .allowedKmsKeys(s3TablesConfigModel.getAllowedKmsKeys())
                     .build();
             break;
           default:
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
@@ -202,7 +202,7 @@ public abstract class PolarisStorageConfigurationInfo {
     AZURE(List.of("abfs://", "wasb://", "abfss://", "wasbs://")),
     GCS("gs://"),
     FILE("file://"),
-    S3_TABLES("arn:"),
+    S3_TABLES("arn:aws:s3tables"),
     ;
 
     private final List<String> prefixes;
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
@@ -40,6 +40,7 @@ import org.apache.polaris.core.config.RealmConfig;
 import org.apache.polaris.core.entity.CatalogEntity;
 import org.apache.polaris.core.entity.PolarisEntity;
 import org.apache.polaris.core.entity.PolarisEntityConstants;
+import org.apache.polaris.core.storage.aws.AwsS3TablesStorageConfigurationInfo;
 import org.apache.polaris.core.storage.aws.AwsStorageConfigurationInfo;
 import org.apache.polaris.core.storage.azure.AzureStorageConfigurationInfo;
 import org.apache.polaris.core.storage.gcp.GcpStorageConfigurationInfo;
@@ -59,6 +60,7 @@ import org.slf4j.LoggerFactory;
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
 @JsonSubTypes({
   @JsonSubTypes.Type(value = AwsStorageConfigurationInfo.class),
+  @JsonSubTypes.Type(value = AwsS3TablesStorageConfigurationInfo.class),
   @JsonSubTypes.Type(value = AzureStorageConfigurationInfo.class),
   @JsonSubTypes.Type(value = GcpStorageConfigurationInfo.class),
   @JsonSubTypes.Type(value = FileStorageConfigurationInfo.class),
@@ -200,6 +202,7 @@ public abstract class PolarisStorageConfigurationInfo {
     AZURE(List.of("abfs://", "wasb://", "abfss://", "wasbs://")),
     GCS("gs://"),
     FILE("file://"),
+    S3_TABLES("arn:"),
     ;
 
     private final List<String> prefixes;
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
@@ -196,13 +196,16 @@ public abstract class PolarisStorageConfigurationInfo {
     }
   }
 
+  /** ARN prefix for S3 Tables resources. */
+  public static final String S3_TABLES_ARN_PREFIX = "arn:aws:s3tables";
+
   /** Polaris' storage type, each has a fixed prefix for its location */
   public enum StorageType {
     S3(List.of("s3://", "s3a://")),
+    S3_TABLES(S3_TABLES_ARN_PREFIX),
     AZURE(List.of("abfs://", "wasb://", "abfss://", "wasbs://")),
     GCS("gs://"),
     FILE("file://"),
-    S3_TABLES("arn:aws:s3tables"),
     ;
 
     private final List<String> prefixes;
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
@@ -18,22 +18,16 @@
  */
 package org.apache.polaris.core.storage.aws;
 
-import static org.apache.polaris.core.config.FeatureConfiguration.STORAGE_CREDENTIAL_DURATION_SECONDS;
-import static org.apache.polaris.core.storage.aws.AwsSessionTagsBuilder.buildSessionTags;
-
 import com.google.common.annotations.VisibleForTesting;
 import jakarta.annotation.Nonnull;
 import java.net.URI;
-import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.polaris.core.auth.PolarisPrincipal;
-import org.apache.polaris.core.config.FeatureConfiguration;
 import org.apache.polaris.core.config.RealmConfig;
 import org.apache.polaris.core.storage.CredentialVendingContext;
 import org.apache.polaris.core.storage.InMemoryStorageIntegration;
@@ -41,7 +35,6 @@ import org.apache.polaris.core.storage.StorageAccessConfig;
 import org.apache.polaris.core.storage.StorageAccessProperty;
 import org.apache.polaris.core.storage.StorageUri;
 import org.apache.polaris.core.storage.StorageUtil;
-import org.apache.polaris.core.storage.aws.StsClientProvider.StsDestination;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
@@ -51,9 +44,6 @@ import software.amazon.awssdk.policybuilder.iam.IamPolicy;
 import software.amazon.awssdk.policybuilder.iam.IamResource;
 import software.amazon.awssdk.policybuilder.iam.IamStatement;
 import software.amazon.awssdk.services.sts.StsClient;
-import software.amazon.awssdk.services.sts.model.AssumeRoleRequest;
-import software.amazon.awssdk.services.sts.model.AssumeRoleResponse;
-import software.amazon.awssdk.services.sts.model.Tag;
 
 /** Credential vendor that supports generating */
 public class AwsCredentialsStorageIntegration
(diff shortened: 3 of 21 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -812,6 +816,13 @@ public Optional<LoadTableResponse> loadTable(
     // when data-access is specified but access delegation grants are not found.
     Table table = baseCatalog.loadTable(tableIdentifier);
 
+    // Capture tableId from remote catalog response for S3 Tables ARN construction
+    Optional<String> capturedTableId = capturedConfigHolder().getTableId();
+    List<String> resourceArns = List.of();
+    if (capturedTableId.isPresent()) {
+      resourceArns = List.of(constructS3TablesArn(capturedTableId.get()));
```
dimas-b: This is a smart idea 😉 However, I'm not sure it is viable from the internal APIs POV. This approach appears to circumvent the usual way of passing locations to the storage integration code and instead introduces a new optional way of passing similar information via a different channel (`CredentialVendingContext`).

I believe it would be beneficial to holistically refactor all storage integrations call paths so that location data is handled uniformly.

Note: such a refactoring will probably overlap with #3699.

dimas-b (commented):
Thanks for your contribution, @aritragster !

The idea in this PR is very interesting, but I guess it might be best to have a broach discussion about it on the `dev` [ML](https://polaris.apache.org/community/).

Also, I'm not sure vending storage credentials for federated catalogs is a well-understood use case, which is another point for having a `dev` discussion.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -393,6 +399,36 @@ private static String arnKeyAll(String region, String accountId) {
...
+          .addAction("s3tables:UpdateTableMetadataLocation")
+          .addAction("s3tables:PutTableData");
+    }
+
+    List<String> arns = context.resourceArns().orElse(List.of());
+    for (String arn : arns) {
+      statement.addResource(IamResource.create(arn));
+    }
+
+    return IamPolicy.builder().addStatement(statement.build()).build();
```
singhpk234: I think having a dedidcated storage integration would be really helpful than writing if else for s3 tables

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
aritragster: hey @dimas-b thanks for the initial review on this.

The idea was to get some early feedback if this makes sense, or we should think of a different approach. I also started the discussion on the ML: https://lists.apache.org/thread/sk84l9wdgpkyvk50v0d3wclh7f3575xn

aritragster: I took a look at #3699 and the direction there looks like a natural fit for what S3 Tables needs. The key difference with S3 Tables is that the resource identifier is an ARN (not an s3:// path) and the IAM actions use the s3tables: namespace, but the overall credential vending flow is the same. If #3699's refactored abstractions can accommodate ARN-based resources alongside path-based ones, S3 Tables support should fit in cleanly.

Happy to rebase this PR on top of #3699 once it lands, or collaborate with @tokoko to make sure the refactoring accounts for the S3 Tables case. In the meantime I'll keep this draft up as a reference for the S3 Tables-specific logic (policy generation, signingName detection, tableId capture).

Let me know what you'd prefer — happy to help either way.

flyrain:
Hi @aritragster , thanks for working on it. The PR is in draft now. Is it ready for review?

flyingImer:
Thanks for the draft here. I’m aligned with the direction, and I think this is already useful in showing that S3 Tables can fit the federated catalog path at the protocol level.

My bias would be to treat this as a good proof point, and then use the next iteration to tighten both the shape and the vending semantics a bit more.

In particular, I wonder if the cleaner direction here would be:
- keep the S3 Tables-specific policy generation in a dedicated storage-integration path, rather than growing more branching in the existing S3 logic
- align that with the broader push toward a more unified call path for storage locations, rather than introducing another parallel path here

On the credential vending side, there is also a concrete next-step item that feel worth making explicit:
- IIUC, only `loadTable` currently derives S3 Tables `resourceArns`, while the create paths still flow through with empty resource scope. I think the next iteration should make those create-time semantics explicit: either derive the usable scope there as well, or fail/defer vending rather than returning partially-scoped credentials. This seems especially relevant given the S3 Tables REST docs note that stage-create is not supported.

Also worth aligning with Tornike's PR #3699 on the unified call path for storage locations, seems like natural synergy.

Looking forward to seeing this evolve.

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -200,6 +202,18 @@ private StorageConfigInfo getStorageInfo(Map<String, String> internalProperties)
...
+            .setRoleArn(s3TablesConfig.getRoleARN())
+            .setExternalId(s3TablesConfig.getExternalId())
+            .setRegion(s3TablesConfig.getRegion())
+            .setCurrentKmsKey(s3TablesConfig.getCurrentKmsKey())
+            .setAllowedKmsKeys(s3TablesConfig.getAllowedKmsKeys())
+            .setStorageType(StorageConfigInfo.StorageTypeEnum.S3_TABLES)
+            .setAllowedLocations(s3TablesConfig.getAllowedLocations())
+            .setStorageName(s3TablesConfig.getStorageName())
+            .build();
+      }
```
singhpk234: lets keep this just next to S3 ?

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -200,6 +202,7 @@ public enum StorageType {
     AZURE(List.of("abfs://", "wasb://", "abfss://", "wasbs://")),
     GCS("gs://"),
     FILE("file://"),
+    S3_TABLES("arn:"),
```
singhpk234: how about specifying this till : `arn:aws:s3tables`

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsS3TablesStorageConfigurationInfo.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.polaris.core.storage.PolarisStorageConfigurationInfo;
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
+
+/** AWS S3 Tables storage configuration information. */
+@PolarisImmutable
+@JsonSerialize(as = ImmutableAwsS3TablesStorageConfigurationInfo.class)
+@JsonDeserialize(as = ImmutableAwsS3TablesStorageConfigurationInfo.class)
+@JsonTypeName("AwsS3TablesStorageConfigurationInfo")
+public abstract class AwsS3TablesStorageConfigurationInfo extends PolarisStorageConfigurationInfo {
```
singhpk234, line 37: why not extend AWSS3StorageIntegrations ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/ConfigCapturingHTTPClient.java
```diff
@@ -0,0 +1,120 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A delegating wrapper around the Iceberg {@link RESTClient} that intercepts responses to extract
+ * the {@code config} section from loadTable responses. When a {@link LoadTableResponse} is
+ * received, the config map (containing {@code tableId} for S3 Tables) is captured and stored in the
+ * request-scoped {@link CapturedConfigHolder}.
+ */
+public class ConfigCapturingHTTPClient implements RESTClient {
```
singhpk234, line 41: rather than overriding the rest client i wonder if we can inject the `@RequestScopedBean` we introduced i.e `CapturedConfigHolder` where we are doing actual loadTable ? i am assuming as part of federaton we would be calling irc's loadTable anyways ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -855,8 +865,35 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
...
 
-      // For non polaris' catalog, validate that table locations are within allowed locations
-      if (!(baseCatalog instanceof IcebergCatalog)) {
+      // For S3 Tables catalogs, replace s3:// table locations with the constructed table ARN.
+      // s3tables:* IAM actions require ARN resources, not s3:// paths.
+      CatalogEntity catalogEntity = CatalogEntity.of(getResolvedCatalogEntity());
+      boolean isS3Tables =
+          catalogEntity.getStorageConfigurationInfo() != null
+              && catalogEntity.getStorageConfigurationInfo().getStorageType()
+                  == PolarisStorageConfigurationInfo.StorageType.S3_TABLES;
```
singhpk234: lets move to a private function below, also lets writes an assumption, i would say that call to federated catalog has succeeded and now we want to ..... 
I don't fully understand the contextConfigholder requirement though let me think more about that as well meanwhile.

singhpk234 (commented):
Thank for the update @aritragster !
i left first round of comments

runtime/service/src/main/java/org/apache/polaris/service/catalog/validation/StorageTypeFileIO.java
```diff
@@ -32,6 +32,8 @@ enum StorageTypeFileIO {
 
   FILE("org.apache.iceberg.hadoop.HadoopFileIO", false),
 
+  S3_TABLES("org.apache.iceberg.aws.s3.S3FileIO", true),
```
flyingImer: I do think the PR is moving in the right structural direction by giving S3 Tables a more dedicated path. But this specific change also means `StorageTypeFileIO` is no longer a stable discriminator: both `S3` and `S3_TABLES` now map to the same `S3FileIO`, while the reverse lookup still assumes FileIO class uniquely determines storage type. That makes existing S3 validation behavior depend on enum overwrite/iteration order, which feels too brittle to merge as-is.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -855,8 +865,35 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
...
+
+      if (isS3Tables && capturedTableId.isPresent()) {
+        String tableArn = constructS3TablesArn(catalogEntity, capturedTableId.get());
+        validateS3TablesArn(tableIdentifier, tableArn, catalogEntity);
+        tableLocations = Set.of(tableArn);
+        LOGGER
+            .atDebug()
+            .addKeyValue("tableIdentifier", tableIdentifier)
```
flyingImer: This is the main remaining contract gap for me. If we already know the catalog is `S3_TABLES`, then failing to derive `tableId` / table ARN should fail closed, not warn-and-continue. Otherwise we have improved the dedicated path, but we still have not really resolved the earlier concern around create-time / unresolved-scope semantics: the handler proceeds even though the downstream S3 Tables policy path expects validated table-ARN-shaped input.

runtime/service/src/main/java/org/apache/polaris/service/catalog/validation/StorageTypeFileIO.java
flyingImer: I think the issue here is not the dedicated S3 Tables path itself. That part feels directionally right. The issue is that `StorageTypeFileIO` no longer gives you a stable discriminator once both `S3` and `S3_TABLES` map to the same `S3FileIO`, but the reverse lookup still assumes FileIO impl uniquely determines storage type. Would it make sense to switch this validation path to a discriminator that remains explicit, instead of depending on shared FileIO impl here?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
flyingImer: This is the main remaining gap for me. If we already know the catalog is `S3_TABLES`, I wonder if this should just fail closed once `tableId`/table ARN cannot be derived. Otherwise the dedicated path is better, but the unresolved-scope/create-time story still feels incomplete to me, because the downstream S3 Tables policy path is still being reached without validated table-ARN-shaped input.

flyingImer:
I do think this is directionally better now.

Pulling S3 Tables into a more dedicated path makes sense to me, and I think this is in a better place than the earlier version where the logic was drifting further into the generic S3 vending path.

That said, I still think there are two issues that block this for me:

1. `StorageTypeFileIO` is no longer a stable discriminator once both `S3` and `S3_TABLES` point at the same `S3FileIO`, but the validation path still reverse-maps from FileIO impl back to storage type.
2. Once we know the catalog is `S3_TABLES`, the no-`tableId` path still warns and continues instead of failing closed. So IIUC, the create-time/unresolved-scope gap is still not really closed yet.

So overall, I'm aligned with the structural direction here, but I still don't think the contract is explicit enough yet for merge.

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
aritragster: Done — moved the S3_TABLES block right after S3 to keep AWS storage types grouped.

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
aritragster: Good catch — changed to arn:aws:s3tables to avoid matching non-S3-Tables ARNs. Updated the CatalogEntity base location validation to match

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsS3TablesStorageConfigurationInfo.java
aritragster: This was Intentional choice. S3 Tables doesn't use most of the S3-specific fields (endpoint, pathStyleAccess, stsUnavailable, userARN, endpointInternal). Extending AwsStorageConfigurationInfo would inherit all of those as nullable fields that are never set, which made me think about this. The only shared fields are roleARN, region, externalId, and KMS — and those are explicitly declared on the S3 Tables config class. Happy to reconsider if you feel strongly about it.

runtime/service/src/main/java/org/apache/polaris/service/catalog/validation/StorageTypeFileIO.java
aritragster: Fixed. Set validateAllowedStorageType=false on S3_TABLES so it's excluded from the FileIO reverse lookup map. S3FileIO still resolves to S3. S3_TABLES validation happens via storage type, not FileIO impl.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
aritragster: Agreed. I changed it from warn-and-continue to BadRequestException. If the catalog is S3_TABLES and no tableId was captured, we now refuse to vend credentials rather than proceeding with unscoped permissions.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
aritragster: Done on the private function - extracted to isS3TablesCatalog() with a comment documenting the assumption that the federated catalog loadTable call has already succeeded.

On the CapturedConfigHolder: the problem it solves is that the S3 Tables tableId only exists in the config section of the remote loadTable REST response. When Polaris calls baseCatalog.loadTable(), the Iceberg RESTCatalog internally deserializes the HTTP response, consumes the config map (for token refresh, endpoint config, etc.), and returns a BaseTable object. The BaseTable/TableMetadata doesn't expose the config section - by the time our handler sees the table, the tableId is gone.

The CapturedConfigHolder is a request-scoped bean that bridges this gap. The ConfigCapturingHTTPClient wraps the RESTClient at the HTTP layer (the only point where the config is visible), extracts the tableId, and stashes it in the holder. After baseCatalog.loadTable() returns, the handler reads the tableId from the holder to construct the table ARN.

Let me know what you think of this.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
dimas-b: @aritragster : just to sync up: I tend to prefer completing #3699 first... I hope it does not cause too much delay on this PR. Please let us know if this is timeline is problematic.

#3699 may not provide all the features needed by this PR, but at least I hope we can perform the main refactoring in #3699 and later add remaining features for S3 tables in this PR.

polaris-core/src/main/java/org/apache/polaris/core/entity/CatalogEntity.java
```diff
@@ -356,6 +370,28 @@ public Builder setStorageConfigurationInfo(
                     .storageName(storageConfigModel.getStorageName())
                     .build();
             break;
+          case S3_TABLES:
+            if (defaultBaseLocation == null
+                || !defaultBaseLocation.startsWith("arn:aws:s3tables")) {
```
singhpk234: `arn:aws:s3tables` may be move this to the constants ?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsS3TablesCredentialsStorageIntegration.java
```diff
@@ -0,0 +1,220 @@
...
+
+  @Override
+  public @Nonnull Map<String, Map<PolarisStorageActions, ValidationResult>>
+      validateAccessToLocations(
+          @Nonnull RealmConfig realmConfig,
+          @Nonnull AwsS3TablesStorageConfigurationInfo storageConfig,
+          @Nonnull Set<PolarisStorageActions> actions,
+          @Nonnull Set<String> locations) {
+    // No-op stub for S3 Tables — real validation deferred to a follow-up PR.
+    return Map.of();
```
singhpk234: can you please add an issue and reference the issue link here ? 
follow-up PR is not a meaningful name 
happy to assign it to you since you have already something working

polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java
```diff
@@ -200,6 +202,7 @@ public enum StorageType {
     AZURE(List.of("abfs://", "wasb://", "abfss://", "wasbs://")),
     GCS("gs://"),
     FILE("file://"),
+    S3_TABLES("arn:aws:s3tables"),
```
singhpk234: you can move it before ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -837,7 +846,8 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
       TableMetadata tableMetadata,
       EnumSet<AccessDelegationMode> delegationModes,
       Set<PolarisStorageActions> actions,
-      Optional<String> refreshCredentialsEndpoint) {
+      Optional<String> refreshCredentialsEndpoint,
+      Optional<String> capturedTableId) {
```
singhpk234: do we need to pass this as Args ? since this a request scoped bean its applicable to this Request only right ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRESTFederatedCatalogFactory.java
```diff
@@ -54,9 +62,11 @@ public Catalog createCatalog(
...
             context,
             (config) ->
-                HTTPClient.builder(config)
-                    .uri(config.get(org.apache.iceberg.CatalogProperties.URI))
-                    .build());
+                new ConfigCapturingHTTPClient(
+                    HTTPClient.builder(config)
+                        .uri(config.get(org.apache.iceberg.CatalogProperties.URI))
+                        .build(),
+                    capturedConfigHolder));
```
singhpk234, line 69: Might have to think more about this one : if there is a better way to capture the config part of loadTable response

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -892,6 +930,45 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
...
+    List<String> allowedLocations = storageConfig.getAllowedLocations();
+    boolean isAllowed =
+        allowedLocations.stream().anyMatch(allowed -> tableArn.startsWith(allowed + "/"));
+    if (!isAllowed) {
+      throw new ForbiddenException(
+          "Table '%s' has ARN '%s' which is outside the catalog's allowed locations: %s",
+          tableIdentifier, tableArn, allowedLocations);
+    }
+  }
+
```
singhpk234: can we capture all of these in a dedicate util class

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -855,8 +865,36 @@ private LoadTableResponse.Builder buildLoadTableResponseWithDelegationCredential
...
+      } else if (isS3Tables) {
+        // Fail closed: S3 Tables catalogs require a tableId to construct the table ARN
+        // for scoped credential vending. Without it, we cannot generate a properly scoped
+        // IAM session policy.
+        throw new BadRequestException(
+            "Cannot vend credentials for S3 Tables table '%s': "
+                + "no tableId was captured from the remote catalog response. "
+                + "Ensure the remote S3 Tables endpoint returns tableId in the loadTable config.",
+            tableIdentifier);
+      }
```
singhpk234: I am a bit confused, wouldn't this be completely inside, since this is what marks this catalog as a federated catalog ?
> !(baseCatalog instanceof IcebergCatalog)

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergRESTFederatedCatalogFactory.java
aritragster: I thought about this one a bit. The HTTP interception is really the only way to get the tableId without either changing the Iceberg library or making an extra API call.
The problem is that the S3 Tables Iceberg REST endpoint returns tableId in the config section of the loadTable response, but Iceberg's RESTCatalog consumes that config internally (for token refresh, endpoint routing, etc.) and never exposes it on the BaseTable or TableMetadata. So by the time baseCatalog.loadTable() returns, the tableId is just gone.

The alternatives I considered:
A supplementary s3tables:GetTable API call, but that adds latency and requires extra IAM permissions on the Polaris server.
An upstream Iceberg change to expose response config, but that's a long lead time and not guaranteed to be accepted.
The wrapper itself is pretty thin (~40 lines of delegation) and only captures data, never modifies the response. Happy to revisit if you see a cleaner path though.

singhpk234:
can you resolve the conflicts @aritragster

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsS3TablesCredentialsStorageIntegration.java
```diff
@@ -0,0 +1,232 @@
...
+
+    if (region != null) {
+      accessConfig.put(StorageAccessProperty.CLIENT_REGION, region);
+    }
+
+    refreshCredentialsEndpoint.ifPresent(
+        endpoint ->
+            accessConfig.put(StorageAccessProperty.AWS_REFRESH_CREDENTIALS_ENDPOINT, endpoint));
+
+    return accessConfig.build();
```
singhpk234, line 101: This is mostly same as S3, i would say can we move this to AWSUtil or something ? we can pass the policy string & prinicpal name as args for this since these are the one which differ from s3

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CapturedConfigHolder.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * <p>Thread safety: CDI {@code @RequestScoped} beans are inherently isolated per request. The
+ * {@code volatile} keyword ensures visibility across threads within the same request if async
+ * processing is used.
+ */
+@RequestScoped
+public class CapturedConfigHolder {
+
+  /** Config key for the S3 Tables table identifier in the loadTable response config section. */
+  public static final String TABLE_ID_CONFIG_KEY = "tableId";
```
singhpk234: This is s3Table key, can you prefix it with that or you can have 

Map<String, Map<String, String>> as well

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/ConfigCapturingHTTPClient.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.iceberg.rest.RESTResponse;
+import org.apache.iceberg.rest.responses.ErrorResponse;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A delegating wrapper around the Iceberg {@link RESTClient} that intercepts responses to extract
+ * the {@code config} section from loadTable responses. When a {@link LoadTableResponse} is
+ * received, the config map (containing {@code tableId} for S3 Tables) is captured and stored in the
```
singhpk234: Lets make this generic ! just say intercepting the raw rest responses !

singhpk234 (commented):
This is looking close @aritragster ! 
can you please rebase this pr !

flyingImer:
Coming back after the latest round. You addressed my 4-01 blockers and singhpk234 is on the finish line for local issues. Before this lands, three structural things.

**1. `StorageTypeFileIO`: the current resolution opts out of the contract, it does not fix it.**

My 4-01 point was that `S3` and `S3_TABLES` both mapping to `S3FileIO` breaks the reverse-lookup assumption (`FileIO impl -> storage type`). The resolution here sets `validateAllowedStorageType=false` on `S3_TABLES` so it is excluded from the reverse map. The contract is still "FileIO impl uniquely determines storage type", with `S3_TABLES` as an exception.

Before this PR, the `validate=false` escape was only used by `IN_MEMORY` (test FileIO). Extending it to a production storage type is a different character of decision.

The next S3-family storage type hits the same shape. The structural fix is a dedicated discriminator (the `StorageType` enum is already there) and dropping `FileIO class -> storage type` reverse lookup in the validation path.

Two ways to land this, my preference is A:
- **A (preferred)**: do the discriminator refactor in this PR. `StorageType` enum already exists, the scope is the validation path's reverse lookup.
- **B (acceptable)**: file a follow-up issue and add a TODO next to `S3_TABLES(..., false)` citing the issue, with a commit to do the refactor before another S3-family type lands. This matches singhpk234's 4-25 pattern for `AwsS3TablesCredentialsStorageIntegration:218`.

The one outcome I would push back on is skipping A without filing the follow-up.

**2. `ConfigCapturingHTTPClient` patches an Iceberg-library API gap from inside Polaris.**

You laid this out on 4-27: `tableId` sits in the remote `loadTable` response's `config`, but `RESTCatalog` consumes `config` internally and never exposes it on `BaseTable` or `TableMetadata`. So we wrap the HTTP client to intercept the raw response.

Three issues compound here:
- brittle against Iceberg library version bumps, since internal config consumption is not a stable boundary
- narrow: it hooks on `LoadTableResponse` specifically. Any other response type from another endpoint needs its own wrapper.
- the problem is upstream in Iceberg, the maintenance cost lands in Polaris

You already considered the upstream Iceberg fix on 4-27 and ruled it out as long-lead and not guaranteed to land. I would flip that call. Concrete ask:
- file the Iceberg issue now (happy to help draft if useful)
- add a TODO or `@see` on `ConfigCapturingHTTPClient` class Javadoc linking the upstream issue, so future readers see this is a temporary shim
- keep the interception in this PR on that footing, not as a permanent mechanism

Shipping it without the upstream link signs Polaris up for fork-shaped debt inside OSS.

singhpk234's 4-25 comment on `IcebergRESTExternalCatalogFactory:69` is the same instinct at the local level. Worth lifting it to the boundary.

**3. Sequencing vs #3699.**

You offered on 3-25 to rebase on #3699 once it lands, or to collaborate with tokoko so the refactoring accounts for S3 Tables. +1 to that and to dimas-b 4-13. Let's hold to that sequencing rather than flip to merge-first.

If this merges first, #3699 either carries S3 Tables semantics through the refactor or leaves the S3 Tables path as a legacy branch. Either way the "one coherent vending abstraction" outcome gets worse.

If #3699's abstraction can express ARN-based resources alongside path-based ones, which you flagged as a fit on 3-25, S3 Tables drops in cleanly. Better landing than branching-in-S3 now and retrofitting later.

---

Not trying to block. Iteration since the 3-24 draft has moved from `if/else` branching inside S3 logic to a dedicated S3 Tables integration path. But the discriminator contract and the interception layer are shapes we live with long after this merges, and I would rather get the shape right once.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/CapturedConfigHolder.java
```diff
@@ -0,0 +1,57 @@
...
+ * handler for S3 Tables ARN construction.
+ *
+ * <p>Thread safety: CDI {@code @RequestScoped} beans are inherently isolated per request. The
+ * {@code volatile} keyword ensures visibility across threads within the same request if async
+ * processing is used.
+ */
+@RequestScoped
+public class CapturedConfigHolder {
+
+  /** Config key for the S3 Tables table identifier in the loadTable response config section. */
```
flyrain: "last loadTable wins" is scope-leaky. If a single request ever loads two tables, the second overwrites the first and the first's creds get scoped to the wrong ARN. Either key by `TableIdentifier` or thread `tableId` explicitly through the call. Also a nice security property: a malicious remote can't swap a different table's `tableId` into an unrelated call.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsS3TablesCredentialsStorageIntegration.java
```diff
@@ -0,0 +1,232 @@
...
+      policyBuilder.addStatement(writeStatement.build());
+    }
+
+    // KMS policy if configured
+    AwsCredentialsStorageIntegration.addKmsKeyPolicy(
+        storageConfig.getCurrentKmsKey(),
+        storageConfig.getAllowedKmsKeys(),
+        policyBuilder,
+        canWrite,
+        storageConfig.getRegion(),
```
flyrain: Passing `null` as `accountId` disables the `isAwsS3` branch in `addKmsKeyPolicy` — KMS scoping won't apply. Derive accountId from the role ARN (or the base ARN) and pass it here.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/S3TablesUtil.java
```diff
@@ -0,0 +1,126 @@
...
+          .addKeyValue("tableArn", tableArn)
+          .log("Replaced table locations with S3 Tables ARN for credential vending");
+      return Set.of(tableArn);
+    }
+
+    // Fail closed: S3 Tables catalogs require a tableId to construct the table ARN
+    // for scoped credential vending. Without it, we cannot generate a properly scoped
+    // IAM session policy.
+    throw new BadRequestException(
+        "Cannot vend credentials for S3 Tables table '%s': "
```
flyrain: Fail-closed `BadRequestException` breaks any credential-vending path that doesn't first go through a remote `loadTable`. Confirm this is truly unreachable for S3_TABLES, otherwise relax to skip vending with a clear error.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/S3TablesUtil.java
```diff
@@ -0,0 +1,126 @@
...
+   */
+  public static void validateTableArn(
+      TableIdentifier tableIdentifier, String tableArn, CatalogEntity catalogEntity) {
+    PolarisStorageConfigurationInfo storageConfig = catalogEntity.getStorageConfigurationInfo();
+    if (storageConfig == null) {
+      return;
+    }
+    List<String> allowedLocations = storageConfig.getAllowedLocations();
+    boolean isAllowed =
+        allowedLocations.stream().anyMatch(allowed -> tableArn.startsWith(allowed + "/"));
```
flyrain: `startsWith(allowed + "/")` double-slashes if `allowed` already ends in `/`. Tighten to an exact bucket match: `startsWith(allowedBucketArn + TABLE_ARN_SEGMENT)` — prevents edge cases where allowed is a broader ARN prefix.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/S3TablesUtil.java
```diff
@@ -0,0 +1,126 @@
...
+      validateTableArn(tableIdentifier, tableArn, catalogEntity);
+      LOGGER
+          .atDebug()
+          .addKeyValue("tableIdentifier", tableIdentifier)
+          .addKeyValue("tableArn", tableArn)
+          .log("Replaced table locations with S3 Tables ARN for credential vending");
+      return Set.of(tableArn);
+    }
+
+    // Fail closed: S3 Tables catalogs require a tableId to construct the table ARN
```
flyrain: Log (info) the final ARN used for credential vending so audits can reconstruct what was granted to whom.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsS3TablesCredentialsStorageIntegration.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /**
+   * Builds an IAM policy with s3tables:* actions scoped to the provided ARN resources.
+   *
+   * <p>Read actions ({@code s3tables:GetTableData}, {@code s3tables:GetTableMetadataLocation}) are
+   * scoped to all read and write ARNs. Write actions ({@code s3tables:UpdateTableMetadataLocation},
+   * {@code s3tables:PutTableData}) are scoped to write ARNs only.
+   */
+  private IamPolicy buildS3TablesPolicy(
```
flyrain: Verify in an IT that a read-only client with only `s3tables:*` can actually fetch underlying S3 objects — per the dev-list thread on external REST catalogs + vended credentials, `s3tables:*` actions may be permission-only and `s3:GetObject` on the underlying bucket may still be required. If so, add `s3:GetObject` to the read policy here.

aritragster:
@flyingImer @singhpk234 I've addressed all outstanding feedback, rebased on latest main, and filed the upstream issues. Here's the summary:

## Issues Filed

- **Iceberg upstream:** https://github.com/apache/iceberg/issues/16399 — Expose `tableId` natively in `LoadTableResponse` (removes need for `ConfigCapturingHTTPClient` long-term)
- **Polaris follow-up:** https://github.com/apache/polaris/issues/4486 — Refactor `StorageTypeFileIO` discriminator to support S3 Tables without opt-out

## Review Feedback Addressed

**@singhpk234's feedback:**
- `S3_TABLES` case moved before `GCS` in `PolarisStorageIntegrationProviderImpl` switch
- `TABLE_ID_CONFIG_KEY` renamed to `S3TABLES_TABLE_ID_CONFIG_KEY`
- Shared STS assume-role logic extracted to `AwsStsUtil` — both `AwsCredentialsStorageIntegration` (S3) and `AwsS3TablesCredentialsStorageIntegration` now use it, eliminating duplication
- `ConfigCapturingHTTPClient` Javadoc made generic (removed S3 Tables-specific reference)
- Rebased on latest main, all conflicts resolved

**@flyingImer's feedback:**
- TODO added on `ConfigCapturingHTTPClient` linking https://github.com/apache/iceberg/issues/16399
- TODO added on `StorageTypeFileIO` S3_TABLES entry linking https://github.com/apache/polaris/issues/4486

**Additional cleanup:**
- Fixed `IcebergRESTFederatedCatalogFactory` constructor name mismatch (pre-existing bug exposed by rebase)
- Removed dead inline ARN logic from `IcebergCatalogHandler` (already encapsulated in `S3TablesUtil.resolveTableLocations`)
- Added `@Nullable` annotations on `AwsStsUtil` parameters
- Fixed duplicate `CLIENT_REGION`/`refreshCredentialsEndpoint` setting in S3 integration (util handles STS path, else-block handles non-STS path)

## New Tests

- `AwsStsUtilTest` — 12 tests covering assume-role behavior (parameters, credentials, expiration, region, refresh endpoint, session names, session tags)
- `S3TablesUtilTest` — 9 tests covering ARN construction, validation, fail-closed behavior, and type detection

## On Sequencing with #3699

I've been tracking that PR. It currently has no formal approvals, an unresolved cache design discussion (`LoadingCache` vs `Supplier`-based Caffeine), and a pending dev ML vote on SPI removals. There's no clear timeline for it to land.

This PR is self-contained and additive (~1,200 lines including tests, primarily new files). When #3699 lands with its `LocationGrant` model and `StorageIntegration` SPI, my S3 Tables integration becomes a clean new implementation of that interface — the `AwsS3TablesCredentialsStorageIntegration` maps directly to the new abstraction, and `AwsStsUtil` remains shared. I commit to doing that adaptation when #3699 merges.

Merging this first doesn't add debt to the refactor path. It just means S3 Tables credential vending ships sooner.

dimas-b:
I'm fine merging this before #3699 ... please resolve conflicts for CI to run and enable the next review round :+1:

flyingImer (commented):
Hey @aritragster, circling back now that #3699 has landed. Are you still planning to carry this PR forward?

I think the good news is that the S3 Tables-specific part still looks fairly contained: we need the remote `tableId`, an exact table-scoped resource, and a dedicated integration that translates the normal Polaris storage actions into `s3tables:` IAM actions.

For the next iteration, I would not try to preserve the pre-#3699 vending shape. My preference would be to port this onto the current `LocationGrant`/`PolarisStorageIntegration` flow and keep a single grant channel rather than introducing a parallel resource-ARN path through `CredentialVendingContext`.

One contract question worth exercising during that port is whether `LocationGrant.locations` can cleanly represent a backend-native resource such as an S3 Tables table ARN. If not, I think that is the smallest abstraction we should generalize here rather than adding another side channel.

I don't think this PR needs to take on a broader storage-SPI/provider-dispatch refactor. Keeping `S3_TABLES` as a first-class storage type, a dedicated S3 Tables integration, and the tracked Iceberg response-capture shim all seem reasonable to me for this scope.

If you're still working on this, a rebase/port along those lines would be the most useful next step; I'm happy to do the next detailed review on that shape.

PR #4054, opened 2026-03-25
JDBC: Replace coarse-grained synchronized methods with per-realm locking

singhpk234:
###   Design

  bootstrapRealms  →  synchronized(this)        ← global, serializes DDL
  purgeRealms      →  writeLock(realm)           ← per-realm, blocks requests during purge
  getOrCreateSession      →  readLock(realm)     ← shared, parallel
  getOrCreateEntityCache  →  readLock(realm)     ← shared, parallel
  getOrCreateMetaStoreManager  →  no lock        ← stateless
  getDatasourceOperations →  volatile read       ← cached singleton, no lock on hot path

###   What was removed

  - metaStoreManagerMap (HashMap) — was caching a stateless object
  - sessionSupplierMap (HashMap) — supplier already created new instance per call
  - initializeForRealm() — split into smaller methods
  - synchronized on getOrCreateSession, getOrCreateMetaStoreManager, getOrCreateEntityCache
  - Per-request findRootPrincipal DB query (now memoized in verifiedRealms)

###   What was added

  - realmLocks — per-realm ReentrantReadWriteLock(true) (fair mode)
  - cachedDatasourceOperations — volatile + double-checked locking
  - verifiedRealms — ConcurrentHashMap.newKeySet(), bootstrap check once per realm
  - schemaVersionCache — ConcurrentHashMap, schema version once per realm
  - Complete cache eviction in purgeRealms (old code never evicted entityCacheMap)

###   Known limitations (documented, acceptable)

  - Bootstrap + purge same realm not mutually exclusive (user error, matches old behavior)
  - Session outlives read lock (same as old behavior, needs CDI-scoped fix)
  - Cold-start thundering herd on first request per realm (benign, idempotent DB query)

## Checklist

### DIFF
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -32,6 +32,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import javax.sql.DataSource;
 import org.apache.polaris.core.PolarisCallContext;
 import org.apache.polaris.core.PolarisDiagnostics;
@@ -78,6 +80,14 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   // (checkPolarisServiceBootstrappedForRealm), avoiding redundant DB hits on subsequent calls.
   private final Set<String> verifiedRealms = ConcurrentHashMap.newKeySet();
 
+  // Cached DatasourceOperations — created once, reused across all requests.
+  private volatile DatasourceOperations cachedDatasourceOperations;
+
+  // Per-realm ReadWriteLock: read lock for request-path methods, write lock for purge.
+  // Fair mode ensures purge is not starved under heavy read traffic.
+  // Requests for different realms never contend; purge of one realm only blocks that realm.
+  private final ConcurrentHashMap<String, ReadWriteLock> realmLocks = new ConcurrentHashMap<>();
+
   @Inject Clock clock;
   @Inject PolarisDiagnostics diagnostics;
   @Inject PolarisStorageIntegrationProvider storageIntegrationProvider;
@@ -86,6 +96,10 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   @Inject RealmConfig realmConfig;
 
   protected JdbcMetaStoreManagerFactory() {}
+
+  private ReadWriteLock realmLock(String realmId) {
+    return realmLocks.computeIfAbsent(realmId, k -> new ReentrantReadWriteLock(true));
+  }
 
   protected PrincipalSecretsGenerator secretsGenerator(
       String realmId, @Nullable RootCredentialsSet rootCredentialsSet) {
@@ -127,13 +141,21 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   }
 
   public DatasourceOperations getDatasourceOperations() {
-    DatasourceOperations databaseOperations;
-    try {
-      databaseOperations = new DatasourceOperations(dataSource.get(), relationalJdbcConfiguration);
-    } catch (SQLException sqlException) {
-      throw new RuntimeException(sqlException);
+    DatasourceOperations ops = cachedDatasourceOperations;
+    if (ops == null) {
+      synchronized (this) {
+        ops = cachedDatasourceOperations;
+        if (ops == null) {
+          try {
+            ops = new DatasourceOperations(dataSource.get(), relationalJdbcConfiguration);
+          } catch (SQLException sqlException) {
+            throw new RuntimeException(sqlException);
+          }
+          cachedDatasourceOperations = ops;
+        }
+      }
     }
-    return databaseOperations;
+    return ops;
   }
 
   @Override
@@ -162,7 +184,8 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
         DatasourceOperations datasourceOperations = getDatasourceOperations();
         int currentSchemaVersion =
             JdbcBasePersistenceImpl.loadSchemaVersion(datasourceOperations, true);
-        int requestedSchemaVersion = JdbcBootstrapUtils.getRequestedSchemaVersion(bootstrapOptions);
+        int requestedSchemaVersion =
+            JdbcBootstrapUtils.getRequestedSchemaVersion(bootstrapOptions);
         int effectiveSchemaVersion =
             JdbcBootstrapUtils.getRealmBootstrapSchemaVersion(
                 datasourceOperations.getDatabaseType(),
@@ -206,18 +229,24 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
     Map<String, BaseResult> results = new HashMap<>();
 
     for (String realm : realms) {
-      RealmContext realmContext = () -> realm;
-      PolarisMetaStoreManager metaStoreManager = getOrCreateMetaStoreManager(realmContext);
-      BasePersistence session = getOrCreateSession(realmContext);
+      realmLock(realm).writeLock().lock();
+      try {
+        RealmContext realmContext = () -> realm;
+        PolarisMetaStoreManager metaStoreManager = createNewMetaStoreManager();
+        DatasourceOperations datasourceOperations = getDatasourceOperations();
+        BasePersistence session = createSession(datasourceOperations, realm, null);
 
-      PolarisCallContext callContext = new PolarisCallContext(realmContext, session);
-      BaseResult result = metaStoreManager.purge(callContext);
-      results.put(realm, result);
+        PolarisCallContext callContext = new PolarisCallContext(realmContext, session);
+        BaseResult result = metaStoreManager.purge(callContext);
+        results.put(realm, result);
 
-      // Evict all cached state for this realm so it can be fully re-initialized if needed
-      entityCacheMap.remove(realm);
-      schemaVersionCache.remove(realm);
-      verifiedRealms.remove(realm);
+        // Evict all cached state for this realm
+        entityCacheMap.remove(realm);
+        schemaVersionCache.remove(realm);
+        verifiedRealms.remove(realm);
+      } finally {
+        realmLock(realm).writeLock().unlock();
+      }
     }
 
     return Map.copyOf(results);
@@ -232,34 +261,46 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   @Override
   public BasePersistence getOrCreateSession(RealmContext realmContext) {
     String realmId = realmContext.getRealmIdentifier();
-    DatasourceOperations datasourceOperations = getDatasourceOperations();
+    realmLock(realmId).readLock().lock();
+    try {
+      DatasourceOperations datasourceOperations = getDatasourceOperations();
 
-    // Verify bootstrap once per realm lifetime; skip on subsequent calls
-    if (!verifiedRealms.contains(realmId)) {
-      checkPolarisServiceBootstrappedForRealm(realmContext, datasourceOperations);
+      // Verify bootstrap once per realm lifetime; skip on subsequent calls.
+      // On cold start, multiple threads may verify concurrently — this is benign
+      // (idempotent DB query), trading a few redundant queries for simpler code.
+      if (!verifiedRealms.contains(realmId)) {
+        checkPolarisServiceBootstrappedForRealm(realmContext, datasourceOperations);
+      }
+
+      // Stateless — create a fresh instance on every call; schemaVersion is cached per realm
+      return createSession(datasourceOperations, realmId, null);
+    } finally {
+      realmLock(realmId).readLock().unlock();
     }
-
-    // Stateless — create a fresh instance on every call; schemaVersion is cached per realm
-    return createSession(datasourceOperations, realmId, null);
   }
 
   @Override
   public EntityCache getOrCreateEntityCache(RealmContext realmContext, RealmConfig realmConfig) {
-    // EntityCache is stateful (Caffeine + ConcurrentHashMap) — must be shared across requests
-    return entityCacheMap.computeIfAbsent(
-        realmContext.getRealmIdentifier(),
-        realmId -> {
-          PolarisMetaStoreManager metaStoreManager = createNewMetaStoreManager();
-          return new InMemoryEntityCache(diagnostics, realmConfig, metaStoreManager);
-        });
+    String realmId = realmContext.getRealmIdentifier();
+    realmLock(realmId).readLock().lock();
+    try {
+      // EntityCache is stateful (Caffeine + ConcurrentHashMap) — must be shared across requests
+      return entityCacheMap.computeIfAbsent(
+          realmId,
+          k -> {
+            PolarisMetaStoreManager metaStoreManager = createNewMetaStoreManager();
+            return new InMemoryEntityCache(diagnostics, realmConfig, metaStoreManager);
+          });
+    } finally {
+      realmLock(realmId).readLock().unlock();
+    }
   }
 
   /**
    * In this method we check if Service was bootstrapped for a given realm, i.e. that all the
-   * entities were created (root principal, root principal role, etc) If service was not
-   * bootstrapped we are throwing IllegalStateException exception That will cause service to crash
-   * and force user to run Bootstrap command and initialize MetaStore and create all the required
-   * entities
+   * entities were created (root principal, root principal role, etc) If service was not bootstrapped
+   * we are throwing IllegalStateException exception That will cause service to crash and force user
+   * to run Bootstrap command and initialize MetaStore and create all the required entities
    */
   private void checkPolarisServiceBootstrappedForRealm(
       RealmContext realmContext, DatasourceOperations datasourceOperations) {
--- a/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
+++ b/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
@@ -24,6 +24,7 @@ import io.smallrye.common.annotation.Identifier;
 import jakarta.annotation.Nullable;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Inject;
 import java.sql.SQLException;
 import java.time.Clock;
@@ -78,17 +79,20 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   // (checkPolarisServiceBootstrappedForRealm), avoiding redundant DB hits on subsequent calls.
   private final Set<String> verifiedRealms = ConcurrentHashMap.newKeySet();
 
-  // Cached DatasourceOperations — created once, reused across all requests.
-  private volatile DatasourceOperations cachedDatasourceOperations;
-
   @Inject Clock clock;
   @Inject PolarisDiagnostics diagnostics;
   @Inject PolarisStorageIntegrationProvider storageIntegrationProvider;
-  @Inject Instance<DataSource> dataSource;
-  @Inject RelationalJdbcConfiguration relationalJdbcConfiguration;
+  @Inject DatasourceOperations datasourceOperations;
   @Inject RealmConfig realmConfig;
 
   protected JdbcMetaStoreManagerFactory() {}
+
+  @Produces
+  @ApplicationScoped
+  DatasourceOperations produceDatasourceOperations(
+      Instance<DataSource> dataSource, RelationalJdbcConfiguration relationalJdbcConfiguration) {
+    return new DatasourceOperations(dataSource.get(), relationalJdbcConfiguration);
+  }
 
   protected PrincipalSecretsGenerator secretsGenerator(
       String realmId, @Nullable RootCredentialsSet rootCredentialsSet) {
@@ -104,7 +108,7 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
   }
 
   /** Loads and caches the schema version for the given realm (DB hit only on first call). */
-  private int getOrLoadSchemaVersion(DatasourceOperations datasourceOperations, String realmId) {
+  private int getOrLoadSchemaVersion(String realmId) {
     return schemaVersionCache.computeIfAbsent(
         realmId,
         k ->
@@ -116,10 +120,8 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
 
   /** Creates a new stateless {@link JdbcBasePersistenceImpl} for the given realm. */
   private BasePersistence createSession(
-      DatasourceOperations datasourceOperations,
-      String realmId,
-      @Nullable RootCredentialsSet rootCredentialsSet) {
-    int schemaVersion = getOrLoadSchemaVersion(datasourceOperations, realmId);
+      String realmId, @Nullable RootCredentialsSet rootCredentialsSet) {
+    int schemaVersion = getOrLoadSchemaVersion(realmId);
     return new JdbcBasePersistenceImpl(
         diagnostics,
         datasourceOperations,
@@ -127,24 +129,6 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
         storageIntegrationProvider,
         realmId,
         schemaVersion);
-  }
-
-  public DatasourceOperations getDatasourceOperations() {
-    DatasourceOperations ops = cachedDatasourceOperations;
-    if (ops == null) {
-      synchronized (this) {
-        ops = cachedDatasourceOperations;
-        if (ops == null) {
-          try {
-            ops = new DatasourceOperations(dataSource.get(), relationalJdbcConfiguration);
-          } catch (SQLException sqlException) {
-            throw new RuntimeException(sqlException);
-          }
-          cachedDatasourceOperations = ops;
-        }
-      }
-    }
-    return ops;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -208,45 +219,90 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
...
-    if (!metaStoreManagerMap.containsKey(realmContext.getRealmIdentifier())) {
+  public PolarisMetaStoreManager getOrCreateMetaStoreManager(RealmContext realmContext) {
+    String realmId = realmContext.getRealmIdentifier();
+    // Fast path: lock-free CHM read (common case — realm already initialized)
+    PolarisMetaStoreManager result = metaStoreManagerMap.get(realmId);
+    if (result != null) {
+      return result;
+    }
+    // Slow path: per-realm lock for first-time initialization
+    Object lock = realmLocks.computeIfAbsent(realmId, k -> new Object());
```
dimas-b: Hi @singhpk234 , I have a radical proposal since we're touching this code 😅 

The JDBC persistence objects do not have any state. So, we should be able to create them on demand (as request-scoped objects) without any locking at all.

WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -67,9 +69,18 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
...
-  final Map<String, EntityCache> entityCacheMap = new HashMap<>();
-  final Map<String, Supplier<BasePersistence>> sessionSupplierMap = new HashMap<>();
+  // Per-realm caches — use ConcurrentHashMap for lock-free reads on the fast path
+  final Map<String, PolarisMetaStoreManager> metaStoreManagerMap = new ConcurrentHashMap<>();
+  final Map<String, EntityCache> entityCacheMap = new ConcurrentHashMap<>();
+  final Map<String, Supplier<BasePersistence>> sessionSupplierMap = new ConcurrentHashMap<>();
+
+  // Per-realm lock objects so that initialization of distinct realms can proceed in parallel,
+  // rather than serializing all realms behind a single monitor.
+  private final ConcurrentHashMap<String, Object> realmLocks = new ConcurrentHashMap<>();
```
jbonofre: nit: for the record, we could use a `ReentrantReadWriteLock` per realm (instead of `synchronized` and lock-object removal). We would get the read lock in `getOrCreate*` methods, and write lock for purge.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
jbonofre: I guess `purgeRealms` should be `synchronized` too: it would match the new design and prevent races here.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: That a really nice idea, just pushed a commit for it !

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: Is the case here, a realm being purged concurrently when the request for that realm being processed ? added a writeLock for this case to prevent this being read until the purge completes.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -93,42 +114,48 @@ protected PolarisMetaStoreManager createNewMetaStoreManager() {
...
   public DatasourceOperations getDatasourceOperations() {
-    DatasourceOperations databaseOperations;
-    try {
-      databaseOperations = new DatasourceOperations(dataSource.get(), relationalJdbcConfiguration);
-    } catch (SQLException sqlException) {
-      throw new RuntimeException(sqlException);
+    DatasourceOperations ops = cachedDatasourceOperations;
+    if (ops == null) {
+      synchronized (this) {
+        ops = cachedDatasourceOperations;
```
dimas-b: `DatasourceOperations` is a fairly "thin" object. Is it worth caching? The only potentially "heavy" part in it is the random number generator, but it can be moved to a static field, I guess... WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -178,17 +205,18 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
           throw new RuntimeException(
               String.format("Error executing sql script: %s", e.getMessage()), e);
         }
-        initializeForRealm(
-            datasourceOperations, realmContext, bootstrapOptions.rootCredentialsSet());
+        // Cache the effective schema version for this realm
+        schemaVersionCache.put(realm, effectiveSchemaVersion);
```
dimas-b, line 179: This is fine to keep, but I guess it's not effective in most cases. I believe in real life deployments `bootstrap` is performed via Admin CLI, so the Server will still have to load schema version from the database.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -200,53 +228,81 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
...
-      metaStoreManagerMap.remove(realm);
+        BaseResult result = metaStoreManager.purge(callContext);
+        results.put(realm, result);
+
+        // Evict all cached state for this realm
+        entityCacheMap.remove(realm);
+        schemaVersionCache.remove(realm);
+        verifiedRealms.remove(realm);
+      } finally {
+        realmLock(realm).writeLock().unlock();
```
dimas-b: This pattern is not ideal because `realmLock()` can return different objects on each call (in principle).

I believe a more robust patter would be:
```
Lock lock = realmLock(realm).writeLock();
lock.lock();
try {
  ...
} finally {
  lock.unlock();
}
```

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -200,53 +228,81 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
     Map<String, BaseResult> results = new HashMap<>();
 
     for (String realm : realms) {
-      RealmContext realmContext = () -> realm;
-      PolarisMetaStoreManager metaStoreManager = getOrCreateMetaStoreManager(realmContext);
-      BasePersistence session = getOrCreateSession(realmContext);
+      realmLock(realm).writeLock().lock();
```
dimas-b: Do we gain much by having per-realm locks here now?

1) I believe purges can be expected to run in the Admin CLI (separate JVM) in real life deployments.

2) Concurrent purges (even for different realms) are not likely in the same JVM.

Would it be sufficient to use simple `synchronized` here?

The real concurrency concern this PR addresses is for creating new `EntityCache` objects, but those do not need a lock - separate comment thread.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -200,53 +228,81 @@ public Map<String, BaseResult> purgeRealms(Iterable<String> realms) {
...
-      PolarisMetaStoreManager metaStoreManager = getOrCreateMetaStoreManager(realmContext);
-      entityCacheMap.put(
-          realmContext.getRealmIdentifier(),
-          new InMemoryEntityCache(diagnostics, realmConfig, metaStoreManager));
+  public EntityCache getOrCreateEntityCache(RealmContext realmContext, RealmConfig realmConfig) {
+    String realmId = realmContext.getRealmIdentifier();
+    realmLock(realmId).readLock().lock();
+    try {
+      // EntityCache is stateful (Caffeine + ConcurrentHashMap) — must be shared across requests
+      return entityCacheMap.computeIfAbsent(
```
dimas-b: With `computeIfAbsent()` we do not need a lock, I guess.

dimas-b (commented):
Thanks for the update @singhpk234 , previous changes LGTM... A few more suggestions... I hope you do not mind 🙂

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: I agree this is the only possible codepath in OSS polaris to do this so i am gonna remove it :)

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: I think the connection pool, inference of databaseType etc is what my motivation to cache it, please let me know if you feel strongly about it 

agree on making the random number generator static if not already !

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
singhpk234: > Concurrent purges (even for different realms) are not likely in the same JVM.

Agree for open source polaris this is true ! removed it

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
dimas-b: ... but all the inputs are inferred before this method is called right? We're just creating `DatasourceOperations` as  thin wrapper around some pre-computed data, I think :thinking:

dimas-b: oh, you mean the `dataSource.get()` call below?... TBH, I'd prefer direct injection of the `DataSource` (not using `Instance<DataSource>`, that should fix the resolution cost concern... WDYT?

singhpk234: I am specifically referring to this check 
https://github.com/apache/polaris/blob/1023d76cb8ba24c189707c91bfd40da5ba5abd7d/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/DatasourceOperations.java#L87


this is network call

dimas-b: Good point (I missed it before :facepalm: )

dimas-b: However, I believe we can mitigate that via CDI. For example:
* Have an `@ApplicationScoped` producer for `DatasourceOperations`
* The producer will check both the database type and whether the schema exists (bootstrap check)
* The producer need not have any locks - Quarkus will take care of that.
WDYT?

If you prefer not to add a CDI producer for `DatasourceOperations` we can still do the schema change and database type check in the same place (e.g. in `checkPolarisServiceBootstrappedForRealm()`), which will then free `DatasourceOperations` from heavy operations.

dimas-b: My rationale for this thread is that the information held in `DatasourceOperations` does not look worth the double checked locking pattern. I think we can deal with that in a simpler way :sweat_smile:

singhpk234: That makes sense ! I just pushed a commit addressing this. TBH i don't remember why didn't i create this as an application scoped bean in the first place ? this definitely makes simpler

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -67,19 +69,31 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
...
-  @Inject Instance<DataSource> dataSource;
-  @Inject RelationalJdbcConfiguration relationalJdbcConfiguration;
+  @Inject DatasourceOperations datasourceOperations;
   @Inject RealmConfig realmConfig;
 
   protected JdbcMetaStoreManagerFactory() {}
 
+  @Produces
+  @ApplicationScoped
+  DatasourceOperations produceDatasourceOperations(
```
flyrain: Do we plan to let other classes use this application scoped bean? Somewhere like https://github.com/apache/polaris/blob/8cbfb268c780c976acd59536c60d4ae06089c46f/persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/idempotency/RelationalJdbcIdempotencyStore.java#L47. Not a blocker though, we could do that as a followup.

Also in that case, should we move this bean producer out of this class, as it could be injected by multiple classes. Not a blocker as well.

dimas-b: Should it be `static`? `JdbcMetaStoreManagerFactory` itself requires `DatasourceOperations` to be injected... it looks like a dependency loop 🤔

dimas-b: Reuse by other classes may need to be addressed in #3960 🤔  For this PR it looks ok to me.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -67,19 +69,31 @@ public class JdbcMetaStoreManagerFactory implements MetaStoreManagerFactory {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(JdbcMetaStoreManagerFactory.class);
 
-  final Map<String, PolarisMetaStoreManager> metaStoreManagerMap = new HashMap<>();
-  final Map<String, EntityCache> entityCacheMap = new HashMap<>();
-  final Map<String, Supplier<BasePersistence>> sessionSupplierMap = new HashMap<>();
+  // Stateful per-realm cache — InMemoryEntityCache accumulates entries across requests
+  final Map<String, EntityCache> entityCacheMap = new ConcurrentHashMap<>();
```
dimas-b: `private`?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -93,42 +107,28 @@ protected PolarisMetaStoreManager createNewMetaStoreManager() {
...
+            JdbcBasePersistenceImpl.loadSchemaVersion(
                 datasourceOperations,
-                secretsGenerator(realmId, rootCredentialsSet),
-                storageIntegrationProvider,
-                realmId,
-                schemaVersion));
-
-    PolarisMetaStoreManager metaStoreManager = createNewMetaStoreManager();
-    metaStoreManagerMap.put(realmId, metaStoreManager);
+                realmConfig.getConfig(
```
dimas-b: `realmConfig` is request-scoped. It must be aligned with the `realmId` parameter (represent the same realm). It believe this assumption holds in current code, but it is hard to follow and make sure.

Would you mind obtaining the `SCHEMA_VERSION_FALL_BACK_ON_DNE` value at the same place where `realmId` is obtained and passing both as parameters?

I believe this way the code will be more robust WRT request-scoped data handling.

Note: in `purgeRealms` we'll have to make a hard-coded decision about `SCHEMA_VERSION_FALL_BACK_ON_DNE` because there is no true injected realm context there (cf. #3411). I believe it should be ok for now. It's not a behaviour change, but will make actual behaviour explicit.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -178,75 +177,87 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
-      RealmContext realmContext) {
-    if (!metaStoreManagerMap.containsKey(realmContext.getRealmIdentifier())) {
-      DatasourceOperations datasourceOperations = getDatasourceOperations();
-      initializeForRealm(datasourceOperations, realmContext, null);
-      checkPolarisServiceBootstrappedForRealm(realmContext);
-    }
-    return metaStoreManagerMap.get(realmContext.getRealmIdentifier());
+  public PolarisMetaStoreManager getOrCreateMetaStoreManager(RealmContext realmContext) {
+    // Stateless — create a fresh instance on every call, no caching needed
+    return createNewMetaStoreManager();
```
flyrain, line 231: Question: why not use the one in the entity cache? The caller is request scoped, https://github.com/apache/polaris/blob/8cbfb268c780c976acd59536c60d4ae06089c46f/runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java#L236, I guess the point of caching is to reuse it here. Am I missing something?

dimas-b (commented):
Thanks for bearing with me @singhpk234 🙂 I like the direction this PR is going... just a few smaller comments that I believe are worth addressing before we merge.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
dimas-b: `AtomicOperationMetaStoreManager` is a light-weight java object. It's probably not worth reusing.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -153,8 +153,7 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
 
     for (String realm : bootstrapOptions.realms()) {
       RealmContext realmContext = () -> realm;
-      if (!metaStoreManagerMap.containsKey(realm)) {
-        DatasourceOperations datasourceOperations = getDatasourceOperations();
+      if (!verifiedRealms.contains(realm)) {
```
flyrain, line 154: This check doesn’t make much sense here. The bootstrap is invoked by a CLI tool, so relying on in-memory state isn’t meaningful. Instead, we should make the process idempotent and avoid re-bootstrapping an already initialized realm. That logic should be based on the actual DB state, for example whether the root principal already exists, rather than an in-memory cache that only lives for the duration of the process.

Since the CLI tool is short-lived anyway, I’d suggest reusing `checkPolarisServiceBootstrappedForRealm()` here. We could decide whether split it to two methods to strip out `verifiedRealms.add(realmId);` but that's a minor point.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -178,75 +177,87 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
+
       BaseResult result = metaStoreManager.purge(callContext);
       results.put(realm, result);
 
-      sessionSupplierMap.remove(realm);
-      metaStoreManagerMap.remove(realm);
+      // Evict all cached state for this realm
+      entityCacheMap.remove(realm);
+      schemaVersionCache.remove(realm);
+      verifiedRealms.remove(realm);
```
flyrain, line 222: Same here.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -178,75 +177,87 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
-      PolarisMetaStoreManager metaStoreManager = getOrCreateMetaStoreManager(realmContext);
-      BasePersistence session = getOrCreateSession(realmContext);
+      PolarisMetaStoreManager metaStoreManager = createNewMetaStoreManager();
+      BasePersistence session = createSession(realm, null);
 
       PolarisCallContext callContext = new PolarisCallContext(realmContext, session);
+
+      // Verify the realm is bootstrapped before purging — a non-bootstrapped realm
+      // has no root principal, so purging it is a no-op that should be reported as failure.
+      Optional<PrincipalEntity> rootPrincipal = metaStoreManager.findRootPrincipal(callContext);
```
flyrain, line 209: +1, I think this is the right thing to do.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
flyrain: Why do we put it in the entity cache then?

flyrain (commented):
Thanks a lot for working on this, @singhpk234! It's definitely the right direction to remove the bottleneck. Left some comments.

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
dimas-b: Good question :wink: ... but I think it'll go way outside the scope of this PR.

flyrain (dismissed):

dimas-b (dismissed):
Thanks, @singhpk234 ! Changes LGTM, but there are some CI failures.

singhpk234:
Interesting find by CI :

InMemoryBufferEventListener.flush() runs on a background Mutiny worker thread (no HTTP request scope). It calls
  metaStoreManagerFactory.getOrCreateSession(realmContext), which in your refactored code reads from realmConfig - a @RequestScoped CDI bean. That throws
  ContextNotActiveException, silently dropping all events (the @Fallback just logs the error).

  The old code avoided this because it cached session suppliers during bootstrap (within request scope) and never re-read realmConfig on subsequent calls. The
  refactor removed that caching, exposing the request-scope dependency on every call.

I, Replaced @Inject RealmConfig realmConfig (request-scoped) with @Inject RealmConfigurationSource realmConfigurationSource (application-scoped).
  In getOrCreateSession, we construct a RealmConfigImpl inline using the realmConfigurationSource + the realmContext parameter that's already passed in. This
  matches the pattern used by the other MetaStoreManagerFactory implementations (Local, NoSQL), which don't inject RealmConfig as a field at all.

dimas-b:
> InMemoryBufferEventListener.flush() runs on a background Mutiny worker thread (no HTTP request scope).

Very interesting and, TBH, concerning... `flush()` involves `metaStoreManager.writeEvents()`, which could be a blocking call and I do not think we should call blocking I/O on non-executor threads :thinking: 

@singhpk234 : could you share a bit more details about this call? How can it be reproduced? What's the actual name of the thread?

@adutra : WDYT?

persistence/relational-jdbc/src/main/java/org/apache/polaris/persistence/relational/jdbc/JdbcMetaStoreManagerFactory.java
```diff
@@ -178,75 +175,91 @@ public synchronized Map<String, PrincipalSecretsResult> bootstrapRealms(
...
   }
 
   @Override
-  public synchronized BasePersistence getOrCreateSession(RealmContext realmContext) {
-    if (!sessionSupplierMap.containsKey(realmContext.getRealmIdentifier())) {
-      DatasourceOperations datasourceOperations = getDatasourceOperations();
-      initializeForRealm(datasourceOperations, realmContext, null);
+  public BasePersistence getOrCreateSession(RealmContext realmContext) {
+    String realmId = realmContext.getRealmIdentifier();
+    RealmConfig realmConfig = new RealmConfigImpl(realmConfigurationSource, realmContext);
```
dimas-b, line 237: I suppose this is related to the following comment from the main thread:

> [...] It calls metaStoreManagerFactory.getOrCreateSession(realmContext), which in your refactored code reads from realmConfig - a @RequestScoped CDI bean. That throws ContextNotActiveException

This means the `realmConfig` was from a different request scope. In other words, something was reused across different requests without proper context propagation.

IMHO, this is not a problem in this factory, but in the caller.

We need to review the caller and handle context propagation / activation properly. Cf. #4061

@adutra : WDYT?

dimas-b (approved):
All in all, this PR seems reasonable to me. I think we should still address the context propagation issues, but I believe they are not specific to the JDBC Persistence code. This PR just exposed pre-existing issues. 

I think it's ok to merge this PR and follow up on CDI context problem separately.

@singhpk234 : Please give @adutra some time to review before merging.

singhpk234:
@adutra any further comments ? i propose to merge this by tomorrow EOD if not :)

dimas-b:
From my POV context handling / propagation can still be improved, but current code in this PR seems operational, so I'm be fine with merging "as is" and improving context propagation later.

singhpk234:
Thanks for review @flyrain @dimas-b !

PR #4056, opened 2026-03-25
Unify equals/hashCode for Index.Element

### LINKED ISSUE #4049: Asymmetric equals methods in `Index.Element` implementations
Following up from #4018 

Original comment: https://github.com/apache/polaris/pull/4018/changes#r2954587783

Test code:
```
  @Test
  void compareWithIndexElem() {
    IndexKey key = IndexKey.key("test");
    DirectIndexElement<Object> d = new DirectIndexElement<>(key, null);
    Index.Element<String> e = Index.Element.of(key, "test");
    assertThat(d.equals(e)).isFalse();
    assertThat(e.equals(d)).isFalse();
  }
```

The first assertion passes, but the second assertion throws a NullPointerException. This indicates that this `.equals()` impl. in non-commutative in general.

dimas-b:
* Convert the old `IndexElem` class to the root of the `Index.Element` class hierarchy

* Consolidate equals/hashCode logic for all sub-classes in `IndexElem`

* Support `null` values via a `protected` method - not exposed to public API callers

* Note: Internal index elements having `null` values are never actually compared in runtime ATM

Fixes #4049

## Checklist

### DIFF
--- a/persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
+++ b/persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
@@ -27,8 +27,8 @@ import jakarta.annotation.Nullable;
 import java.util.Objects;
 
 /**
- * Package-private implementation of {@link Index.Element} for {@link Index.Element#of(IndexKey,
- * Object)}.
+ * Base class for {@link Index.Element} implementations that provides coherent {@link
+ * #equals(Object)} and {@link #hashCode()} semantics.
  */
 public abstract class IndexElem<V> implements Index.Element<V> {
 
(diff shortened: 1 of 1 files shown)


persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
```diff
@@ -58,8 +73,9 @@ public boolean equals(Object o) {
     if (o == this) {
       return true;
     }
-    if (o instanceof Index.Element<?> other) {
-      return key.equals(other.key()) && value.equals(other.value());
+    if (o instanceof IndexElem<?> other) {
```
jbonofre, line 81: For context (if other reviewers are asking), this is a change in the semantic here.
The "old" `IndexElem` (record) compared against any `Index.Element<?>`, but the new implementation only matches `instanceof IndexElem<?>`. This means a third-party `Index.Element` implementation that isn't an `IndexElem` subclass will no longer be considered equal, even with the same key/value.

The "old" `AbstractIndexElement` also had a fallback to `Index.Element<?>`.

This is an intentional "breaking change". Maybe worth to add a comment here to avoid "surprising" the reviewers 😄

jbonofre (approved):

persistence/nosql/persistence/api/src/main/java/org/apache/polaris/persistence/nosql/api/index/IndexElem.java
dimas-b: Good point about the comment - will add.

Re: "breaking changes", the code affected by this PR has not been released yet. Do we want to be so strict about changes between individual commits?

dimas-b: comments added

jbonofre: It's all good, thanks !

PR #4061, opened 2026-03-25
feat(service): async context propagation for task executor

### LINKED ISSUE #3444: Provide a robust CDI way to inject request_ids

flyingImer:
This PR adds RequestIdHolder and a concrete context propagation helper for TaskExecutorImpl. Fixes #3444

### Problem

When TaskExecutorImpl schedules async work, the task runs on a different thread with a fresh CDI request scope. Request-scoped context (realm, principal, request ID) was previously propagated via ad-hoc hardcoded logic, and request IDs were not propagated at all, since the only way to read them was through RESTEasy's internal CurrentRequestManager API, which is unavailable on task threads.

### Solution

**RequestIdHolder** is a new @RequestScoped CDI bean replacing the removed ServiceProducers.requestIdSupplier() that depended on RESTEasy internals. It produces RequestIdSupplier via CDI so any component can inject it without depending on JAX-RS types. RequestIdFilter now writes to this holder on each request.

**TaskContextPropagator** is a package-private helper that captures realm, principal, and request ID on the request thread and restores them into the task thread's fresh CDI request scope. It directly injects RealmContextHolder, PolarisPrincipalHolder, and RequestIdHolder. No new SPI or extension point is introduced. The implementation follows the same pattern as Bootstrapper.

CurrentRequestManager is no longer referenced anywhere in the codebase.

### Out of scope (follow-up candidates)

- **MDC propagation**: request ID is not currently written to SLF4J MDC on task threads. Can be added in a follow-up.
- **X-Request-ID header validation**: client-supplied header is used verbatim. Pre-existing behavior in RequestIdFilter, not introduced by this PR.

## Checklist
- [x] Don't disclose security issues! (contact [security@apache.org](mailto:security@apache.org))
- [x] Clearly explained why the changes are needed, or linked related issues: Fixes #3444
- [x] Added/updated tests with good coverage, or manually tested (and explained how)
  - Unit tests for TaskContextPropagator (capture, restore, round-trip)
  - TaskExecutorImplTest updated for new constructor signature
- [x] Added comments for complex logic
- [x] Updated CHANGELOG.md (if needed)
- [ ] Updated documentation in site/content/in-dev/unreleased (if needed)

## Disclaimer
Javadoc is mainly assisted by coding agent.


runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
```diff
@@ -23,24 +23,23 @@
...
-import org.jboss.resteasy.reactive.server.jaxrs.ContainerRequestContextImpl;
+import org.apache.polaris.core.context.RequestIdSupplier;
 
 @ApplicationScoped
 public class PolarisEventMetadataFactory {
 
   @Inject Clock clock;
   @Inject CurrentIdentityAssociation currentIdentityAssociation;
   @Inject Instance<RealmContext> realmContext;
+  @Inject Instance<RequestIdSupplier> requestIdSupplier;
```
dimas-b: Why does it have to be an `Instance`?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
```diff
@@ -100,23 +99,22 @@ private Optional<PolarisPrincipal> getUser() {
...
    */
   private Optional<String> getRequestId() {
-    // See org.jboss.resteasy.reactive.server.injection.ContextProducers
-    ResteasyReactiveRequestContext context = CurrentRequestManager.get();
-    if (context != null) {
-      ContainerRequestContextImpl request = context.getContainerRequestContext();
-      String requestId = (String) request.getProperty(RequestIdFilter.REQUEST_ID_KEY);
-      return Optional.ofNullable(requestId);
+    if (!requestIdSupplier.isResolvable()) {
+      return Optional.empty();
```
dimas-b: I believe `RequestIdSupplier` should always be resolvable, but actual request ID may not be available in all contexts.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -70,22 +67,23 @@ public class TaskExecutorImpl implements TaskExecutor {
...
   private final TaskFileIOSupplier fileIOSupplier;
-  private final RealmContextHolder realmContextHolder;
-  private final PolarisPrincipalHolder polarisPrincipalHolder;
-  private final PolarisPrincipal polarisPrincipal;
   private final List<TaskHandler> taskHandlers = new CopyOnWriteArrayList<>();
   private final Optional<TriConsumer<Long, Boolean, Throwable>> errorHandler;
   private final PolarisEventDispatcher polarisEventDispatcher;
   private final PolarisEventMetadataFactory eventMetadataFactory;
   @Nullable private final Tracer tracer;
+  private final Instance<AsyncContextPropagator> contextPropagators;
```
dimas-b: Why not inject a `List<AsyncContextPropagator>`?

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
```diff
@@ -0,0 +1,83 @@
...
+      PolarisPrincipalHolder polarisPrincipalHolder, Instance<PolarisPrincipal> polarisPrincipal) {
+    this.polarisPrincipalHolder = polarisPrincipalHolder;
+    this.polarisPrincipal = polarisPrincipal;
+  }
+
+  @Nullable
+  @Override
+  public Object capture() {
+    PolarisPrincipal clone = null;
+    if (polarisPrincipal.isResolvable()) {
```
dimas-b: `PolarisPrincipal` should always be resolvable in contexts that initiate principal propagation. If it is not, it would be a bug in Polaris AuthN.

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+    LOGGER.trace("capture principal={}", clone != null ? clone.getName() : null);
+    return clone;
+  }
+
+  @Override
+  public AutoCloseable restore(@Nullable Object capturedState) {
+    LOGGER.trace("restore principal={}", capturedState != null ? ((PolarisPrincipal) capturedState).getName() : null);
+    if (capturedState != null) {
+      polarisPrincipalHolder.set((PolarisPrincipal) capturedState);
```
dimas-b: Would it be possible to use generics and avoid arbitrary type casts here?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -159,27 +153,32 @@ public void addTaskHandlerContext(long taskEntityId, CallContext callContext) {
 
     // Capture the metadata now in order to capture the principal and request ID, if any.
     PolarisEventMetadata eventMetadata = eventMetadataFactory.create();
-    tryHandleTask(taskEntityId, clone, eventMetadata, null, 1);
+
+    // Capture request-scoped context for propagation into the task thread.
+    // Each propagator independently snapshots its own piece of context.
+    List<CapturedPropagator> captured = new ArrayList<>();
+    for (AsyncContextPropagator propagator : contextPropagators) {
+      captured.add(new CapturedPropagator(propagator, propagator.capture()));
```
dimas-b: It looks like handing multiple `AsyncContextPropagator` deserves a helper class, so that callers like this `TaskExecutorImpl` only need to deal with one object (for simplicity). WDYT?

runtime/service/src/main/java/org/apache/polaris/service/task/RequestIdPropagator.java
```diff
@@ -0,0 +1,96 @@
...
+  public AutoCloseable restore(@Nullable Object capturedState) {
+    String requestId = (String) capturedState;
+    LOGGER.trace("restore requestId={}", requestId);
+    requestIdHolder.set(requestId);
+
+    if (requestId == null) {
+      return () -> {};
+    }
+
+    String previous = MDC.get(RequestIdFilter.REQUEST_ID_KEY);
```
dimas-b: Why rely on MDC (thread locals essentially), when we can leverage CDI-based injection that is aware of async processing?

runtime/service/src/main/java/org/apache/polaris/service/task/RealmContextPropagator.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Inject
+  public RealmContextPropagator(RealmContextHolder realmContextHolder) {
+    this.realmContextHolder = realmContextHolder;
+  }
+
+  @Nullable
+  @Override
+  public Object capture() {
```
dimas-b: I believe we do not really need to expose the "value" object to callers.

The propagator can handle state internally and expose an action object to caller. For example:

```
State s = p.capture();
// switch to a different CDI context
s.restore(); // calls realmContextHolder.set(...)
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 
 ### New Features
 
+- Request IDs are now propagated to async task threads via CDI-injectable `RequestIdSupplier`, so task log messages carry the originating request's ID. Context propagation across task boundaries is now handled by the pluggable `AsyncContextPropagator` SPI.
```
dimas-b: nit: I tend to think "New Features" is for end users. This message is targeted to downstream developers, so it might belong in "Changes" instead.

dimas-b (commented):
Hi @flyingImer ! Good idea to normalize the code that deals with async context values propagation! Some comments and suggestions below.

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
flyingImer: If we adopt the State pattern from your earlier comment, the raw Object disappears from the public API entirely i.e., each state object handles its own type internally. I believe that addresses this concern without adding generics to the interface.

On the other hand, adding AsyncContextPropagator<T> would also work, but CDI Instance iteration loses generic type parameters, so the caller would still work with AsyncContextPropagator<?>, the type safety gain would only be within each implementation's own capture/restore pair.

Happy to revisit if you see a case where explicit generics would still add value after the State refactoring.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
flyingImer: IIUC, Instance<T> is the standard CDI programmatic lookup while List<T> injection is a Quarkus extension not part of the CDI spec. Went with Instance for portability, but happy to switch if the project prefers List. Any preference?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
flyingImer: Agreed, this deserves a cleaner abstraction. I think this pairs naturally with the State pattern discussion above. The helper shape depends on whether we go with State.restore() or the current capture()/restore(Object) contract.

I'd prefer to bundle this with that follow-up so we refactor once rather than twice. Sound reasonable?

runtime/service/src/main/java/org/apache/polaris/service/task/RequestIdPropagator.java
flyingImer: IIUC, they serve different purposes:

- RequestIdHolder (CDI) -> programmatic access via @Inject RequestIdSupplier
- MDC -> automatic log correlation — %X{requestId} in the log pattern means every log line carries the request ID without callers needing to inject anything

Without MDC on the task thread, task log lines would silently lose the originating request ID, which I believe this is important for observability. The AutoCloseable cleanup ensures pooled threads don't leak stale IDs to the next task.

runtime/service/src/main/java/org/apache/polaris/service/task/RealmContextPropagator.java
flyingImer: Thanks for the suggestion! I like the direction of making the captured state self-contained.

Those raw objects are internal concerns in current implementation. Callers in TaskExecutorImpl already don't see them.

Your State pattern goes further by eliminating the cast entirely — the state object knows how to restore itself. I'm open to refactoring the SPI to this shape. One consideration: each propagator would need to define an inner class or return a lambda wrapping the state + restore logic, which adds some boilerplate for simple propagators (Realm and Principal are essentially holder.get() / holder.set()).

I think either way works, but want to manage the scope of this pr and stay focused. I'm inclined to have it as a follow-up work. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/task/RequestIdPropagator.java
dimas-b: I do not mind populating MDC with Request IDs. Yet, I think it is preferable to avoid MDC as a source of request IDs for context propagation purposes.

runtime/service/src/main/java/org/apache/polaris/service/task/RealmContextPropagator.java
dimas-b: This code is central to the purpose of this PR. Why would we merge some code now if we know we're going to rework it in a follow-up PR?

jbonofre (commented):
Overall good to me.

Can you please fix the spotless issues ? I will do a new pass after. 

Thanks !

runtime/service/src/main/java/org/apache/polaris/service/task/RealmContextPropagator.java
flyingImer: @dimas-b Fair point. I plan to do the State pattern refactor as an immediate follow-up PR.

My rationale is that the interface lives in `runtime/service` with no downstream consumers, so evolving it across two PRs has zero migration cost. This one stays focused on the propagation mechanism + CDI discovery; the follow-up is a pure shape refactor (`capture() → Object` becomes `capture() → RestoreAction`) with no behavioral change.

I'll open it right after this lands. But if you'd rather bundle it here, I can do that too. Just want to keep the review surface manageable. WDYT?

flyingImer:
Thanks @dimas-b and @jbonofre. I’ve addressed the earlier feedback and fixed the CI issues. The latest run is green now.

When you have a chance, could you please take another look at the current diff? Thanks!

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * Produces a {@link RequestIdSupplier} for the current request scope. The returned supplier reads
+   * the request ID from this holder at the time it is called, so it reflects any value set after
+   * this method is first invoked.
+   */
+  @Produces
+  @RequestScoped
+  @Unremovable
```
dimas-b: Why `@Unremovable`?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
```diff
@@ -23,24 +23,23 @@
...
-import org.jboss.resteasy.reactive.server.jaxrs.ContainerRequestContextImpl;
+import org.apache.polaris.core.context.RequestIdSupplier;
 
 @ApplicationScoped
 public class PolarisEventMetadataFactory {
 
   @Inject Clock clock;
   @Inject CurrentIdentityAssociation currentIdentityAssociation;
   @Inject Instance<RealmContext> realmContext;
+  @Inject RequestIdSupplier requestIdSupplier;
```
dimas-b: Why not `RequestIdHolder`?.. It's simpler, no?

runtime/service/src/main/java/org/apache/polaris/service/task/AsyncContextPropagator.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Restores the captured context into the task thread's active request scope.
+   *
+   * @param capturedState the snapshot returned by {@link #capture()}, may be {@code null}.
+   * @return an {@link AutoCloseable} that is closed after the task finishes. Implementations that
+   *     need no cleanup must return a no-op ({@code () -> {}}). Must not return {@code null}.
+   */
+  @Nonnull
+  AutoCloseable restore(@Nullable Object capturedState);
```
dimas-b: I'm not sure why context restoration should require a closable follow-up action. It seems to complicate the design and intended use case. 

The only currently relevant use case deals with MDC, but I believe it would be preferable to keep MDC out of the (CDI) context propagation code.

It's fine to populate MDC where necessary by specific code (e.g. in `TaskExecutorImpl`). WDYT?

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
dimas-b: So it's another reason to use the state pattern in this PR (as opposed to follow-up), I think 🤔

runtime/service/src/main/java/org/apache/polaris/service/task/RealmContextPropagator.java
dimas-b: This has implications to the `AutoClosable` returned from `restore()` (separate comment thread).

If you agree to the general principles, I believe it would be preferable to do the "state" refactoring in this PR. I hope it can resolve my other concerns "automatically" too.

flyingImer:
@dimas-b I think your latest comments convinced me the shape issue is central enough to address in this PR.

My current cut is to keep the scope narrow, but switch this SPI to a state/action model so callers no longer deal with raw `Object` state, and `TaskExecutorImpl` only deals with one captured object per propagator.

I may still keep a generic cleanup hook with a default no-op, but the intent there would be generic lifecycle cleanup, not letting MDC shape the main propagation contract.

I also plan to clean up the smaller `@Unremovable` and `RequestIdHolder` points while touching this area.

Does that sound like the right cut?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
flyingImer: RequestIdSupplier is the read-only contract and this class only reads. It's also what PersistingMetricsReporter uses, so keeping them consistent felt right. Happy to revisit if you see a reason to diverge.

runtime/service/src/main/java/org/apache/polaris/service/task/AsyncContextPropagator.java
flyingImer: Agreed that MDC should not shape the core SPI. In the updated shape, capture() returns a RestoreAction that encapsulates both state and restore logic. RestoreAction extends AutoCloseable with a default no-op close(), so the base contract has no cleanup requirement.

MDC save/restore is now internal to RequestIdPropagator's action. The SPI itself does not mention MDC, and RealmContextPropagator/PrincipalContextPropagator do not override close().

I considered having TaskExecutorImpl manage MDC directly, but that means the executor needs to know which propagators need cleanup and what that cleanup looks like. That goes against the "caller should only deal with one object" direction. WDYT?

dimas-b:
@flyingImer : I'd prefer to keep MDC out of this PR (feel free to improve MDC data in a follow-up PR). I believe it is a totally different concern, not related to Request Context. Let's keep this PR focused on CDI concerns.

Looking forward to a new diff.

runtime/service/src/main/java/org/apache/polaris/service/task/AsyncContextPropagator.java
```diff
@@ -0,0 +1,78 @@
...
+   */
+  interface RestoreAction extends AutoCloseable {
+
+    /** Shared no-op instance for propagators that have nothing to capture. */
+    RestoreAction NOOP = () -> {};
+
+    /** Restores the captured context into the task thread's active request scope. */
+    void restore();
+
+    /** Optional cleanup after the task finishes. Default is a no-op. */
```
dimas-b: Optional cleanup sounds like a misnomer, TBH... What is optional? Calling it or implementing?

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @Override
+  public RestoreAction capture() {
+    PolarisPrincipal clone = null;
+    try {
+      // Clone to allow task thread get a stable snapshot regardless of the request scope
+      // lifecycle.
+      clone = ImmutablePolarisPrincipal.builder().from(polarisPrincipal.get()).build();
+    } catch (ContextNotActiveException e) {
+      // scope not active
```
dimas-b: Capturing out of scope is a mistake. We should not ignore this exception.

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
```diff
@@ -0,0 +1,78 @@
...
+ * <p>A clone of the principal is captured at submission time so the task thread uses a stable
+ * snapshot that is independent of the originating request scope's lifecycle.
+ */
+@ApplicationScoped
+public class PrincipalContextPropagator implements AsyncContextPropagator {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(PrincipalContextPropagator.class);
+
+  private final PolarisPrincipalHolder polarisPrincipalHolder;
+  private final Instance<PolarisPrincipal> polarisPrincipal;
```
dimas-b: Why `Instance`? Most code is able to deal with simple `PolarisPrincipal` typed fields 🤔

runtime/service/src/main/java/org/apache/polaris/service/task/PrincipalContextPropagator.java
```diff
@@ -0,0 +1,78 @@
...
+    this.polarisPrincipal = polarisPrincipal;
+  }
+
+  @Override
+  public RestoreAction capture() {
+    PolarisPrincipal clone = null;
+    try {
+      // Clone to allow task thread get a stable snapshot regardless of the request scope
+      // lifecycle.
+      clone = ImmutablePolarisPrincipal.builder().from(polarisPrincipal.get()).build();
```
dimas-b: Why not `polarisPrincipalHolder.get()`?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -96,22 +91,18 @@ public TaskExecutorImpl(
       Clock clock,
       MetaStoreManagerFactory metaStoreManagerFactory,
       TaskFileIOSupplier fileIOSupplier,
-      RealmContextHolder realmContextHolder,
       PolarisEventDispatcher polarisEventDispatcher,
       PolarisEventMetadataFactory eventMetadataFactory,
       @Nullable Tracer tracer,
-      PolarisPrincipalHolder polarisPrincipalHolder,
-      PolarisPrincipal polarisPrincipal) {
+      Instance<AsyncContextPropagator> contextPropagators) {
```
dimas-b: I believe a `List<AsyncContextPropagator>` will work, won't it?

Using `Instance` should be reserved to infrastructure-level code. Normal beans can get instances of direct java types injected in most cases.

runtime/service/src/main/java/org/apache/polaris/service/task/AsyncContextPropagator.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    default void close() throws Exception {}
```
dimas-b: Having a `close()` method is logically inconsistent with the idea of restoring state multiple times (javadoc on line 49).

To be clear: I support the idea of repeated calls to `restore()`. I believe this interface should not have a `close()` method.

runtime/service/src/main/java/org/apache/polaris/service/task/RequestIdPropagator.java
```diff
@@ -0,0 +1,95 @@
...
+        LOGGER.trace("restore requestId={}", captured);
+        requestIdHolder.set(captured);
+        previous = MDC.get(RequestIdFilter.REQUEST_ID_KEY);
+        MDC.put(RequestIdFilter.REQUEST_ID_KEY, captured);
+      }
+
+      @Override
+      public void close() {
+        if (previous != null) {
+          MDC.put(RequestIdFilter.REQUEST_ID_KEY, previous);
```
dimas-b: I do not think MDC should be mixed with CDI context propagation. 

Moreover, why is request ID in MDC but Principal is not? This is not to say that Principal should be in MDC, but to illustrate that putting request IDs into MDC goes beyond the scope of CDI context propagation.

If you're striving to have a unified way of setting request IDs in MDC, I propose to do that inside the `requestIdHolder.set()` method (and never remove it from MDC).

If the idea is to ensure try/finally style cleanup for request ID in MDC, I do not think context propagators are the right place for that conceptually. Please consider a different approach. That said, I do not see a reason to ever remove request IDs from MDC. Do you have a use case where that is required?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -260,37 +256,52 @@ protected void handleTask(
...
+        } finally {
+          span.end();
+        }
+      }
+    } finally {
+      // Close in reverse order (LIFO) so that later actions clean up before
+      // earlier ones, matching standard stacked-context conventions.
+      for (int i = restored - 1; i >= 0; i--) {
+        try {
+          actions.get(i).close();
```
dimas-b: I think I understand what you'd like to build, but I believe the term "context propagation SPI" (from PR title) would be really confusing in that case.

If the idea is to establish a convenience framework for running code inside some context and make sure the current thread removes traces of that context on completion, I believe a better approach would to follow the pattern of `Bootstrapper` (suggested new name: `ContrextualTaskRunner`).

https://github.com/apache/polaris/blob/6c1a12bfd25a77b34e9e803295cc92492dda47b6/runtime/service/src/main/java/org/apache/polaris/service/config/Bootstrapper.java#L39

Inject `RequestIdHolder`, `PolarisPrincipalHolder` and `RealmContextHolder` into it.

Use them to propagate context internally. No need for a new SPI.

`TaskExecutorImpl` would call:
1. (on task creation) `ContextualAction action = ContrextualTaskRunner.capture()`
2. (on execution) `action.runWithNewContext( { ... } );`

MDC can be handled internally and cleared upon exit from `runWithContext`

flyingImer: hmmm, on second thought, makes sense. I now agree this PR doesn't have a strong enough case for a new SPI. The main value is the propagation itself, not the extension point.

I'll rework this to a concrete helper along the lines of your ContextualTaskRunner suggestion, wiring in the three holders directly. MDC will be a separate follow-up.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Captures the current request-scoped context for later propagation to a task thread.
+   *
+   * <p>Must be called on a thread with an active CDI request scope (e.g. an HTTP request thread or
+   * a task thread that already has context restored). The principal is cloned into an immutable
+   * snapshot so the captured context is independent of the originating scope's lifecycle.
+   */
+  CapturedTaskContext capture() {
+    return new CapturedTaskContext(
+        realmContextHolder.get(),
```
dimas-b, line 75: nit: for good measure, it might be best to create a copy for the realm context too. It's similar to the principal in many aspects.

Request ID is just a `String`, so it's fine to reuse it.

dimas-b (dismissed):
Thanks for bearing with me, @flyingImer 🙂

flyingImer:
Hi @jbonofre, I addressed the earlier feedback and the latest CI is green now.
Dimas already approved the current diff.
When you have a chance, could you take a quick final look? Thanks!

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyingImer: Thanks for calling this out! Let me tackle it in a follow up pr later

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+  void restore(CapturedTaskContext context) {
+    realmContextHolder.set(context.realmContext());
+    polarisPrincipalHolder.set(context.principal());
+    requestIdHolder.set(context.requestId());
+  }
+
+  /**
+   * Immutable snapshot of request-scoped context captured for propagation across async boundaries.
+   */
+  record CapturedTaskContext(
```
adutra: This imo should be public as it is being exposed outside the package scope: in the protected `handleTaskWithTracing` method.

It should probably be made top-level as well.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -159,27 +149,27 @@ public void addTaskHandlerContext(long taskEntityId, CallContext callContext) {
 
     // Capture the metadata now in order to capture the principal and request ID, if any.
     PolarisEventMetadata eventMetadata = eventMetadataFactory.create();
-    tryHandleTask(taskEntityId, clone, eventMetadata, null, 1);
+
+    // Capture request-scoped context for propagation into the task thread.
+    TaskContextPropagator.CapturedTaskContext captured = taskContextPropagator.capture();
```
adutra: With this change, do we still need to clone `CallContext` on line 148? Do we need `CallContext` in this class _at all_?

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ request adding CHANGELOG notes for breaking (!) changes and possibly other secti
 - Changed from Poetry to UV for Python package management.
 - Exclude KMS policies when KMS is not being used for S3.
 - Improved default KMS permission handling to better distinguish read-only and read-write access.
+- Request IDs are now propagated to async task threads via CDI-injectable `RequestIdSupplier`, so task log messages carry the originating request's ID. Context propagation across task boundaries is now handled by the pluggable `AsyncContextPropagator` SPI.
```
adutra:
```suggestion
- Request IDs are now propagated to async task threads via CDI-injectable `RequestIdSupplier`, so task log messages carry the originating request's ID. Context propagation across task boundaries is now handled by the pluggable `TaskContextPropagator` SPI.
```

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Produces a {@link RequestIdSupplier} for the current request scope. The returned supplier reads
+   * the request ID from this holder at the time it is called, so it reflects any value set after
+   * this method is first invoked.
+   */
+  @Produces
+  @RequestScoped
+  public RequestIdSupplier getRequestIdSupplier() {
+    return () -> requestId.get();
```
adutra: nit:

```suggestion
    return requestId::get
```

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /**
+   * Produces a {@link RequestIdSupplier} for the current request scope. The returned supplier reads
+   * the request ID from this holder at the time it is called, so it reflects any value set after
+   * this method is first invoked.
+   */
+  @Produces
+  @RequestScoped
+  public RequestIdSupplier getRequestIdSupplier() {
```
adutra, line 51: Do we need both a `RequestIdHolder` and a `RequestIdSupplier`? I think the holder could now just expose the raw request ID. The supplier creates an unnecessary indirection, it seems.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
adutra: I think `RequestIdHolder` is enough (see my other comment).

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
dimas-b: `RequestIdSupplier` is for injection into other beans, as far as I can tell. It's more specific than a `Supplier<Integer>`.

IMHO, `RequestIdHolder` is meant to be used only by infrastructure code.

dimas-b: Alternatively, we could produce a `@Named("RequestId") Integer` here and get it injected into code that calls `RequestIdSupplier.get()`.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
dimas-b: Good point. I missed it 🤦

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
adutra: If "holder" interfaces are meant for use in infrastructure code only, in this case I would just rename `RequestIdSupplier` to `RequestId` – it would then look quite similar, in shape and purpose, to the `RealmContext` interface.

IMHO it's easier to have an injectable interface encapsulating the string, rather than injecting the raw string itself annotated with `@Named`.

adutra:
Worth noting: this PR introduces a new `TaskContextPropagator` bean. The bean is functionally equivalent to a custom `org.eclipse.microprofile.context.spi.ThreadContextProvider`. 

I wonder if it wouldn't be cleaner to just wire up a custom `ThreadContextProvider` instead of providing an equivalent class that must be invoked explicitly at the beginning of each task.

The custom `ThreadContextProvider` would still require to clear `ThreadContext.CDI` + `@ActivateRequestContext` + manual population of beans. So it wouldn't look radically different for sure, but we would at least benefit from implicit execution.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
dimas-b: +1 to `RequestId`

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
dimas-b: Yes, in this case the "factory" can be thought of as infrastructure code, so using `RequestIdHolder` is perfectly fine.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
dimas-b: However, `RequestIdSupplier` is an old class (from #3385). It might be best to do that renaming in a separate PR :thinking:

dimas-b: To be more specific about "infrastructure" code, I think `RequestIdSupplier` is for use in `polaris-core` and `runtime/service` classes that "consume" requests IDs without participating in the construction of request-scoped objects.

Factories and CDI producers should be able to use `RequestIdHolder`. Its role is pretty clear in that context, I guess.

dimas-b:
> I wonder if it wouldn't be cleaner to just wire up a custom ThreadContextProvider instead of providing an equivalent class that must be invoked explicitly at the beginning of each task.

+1

dimas-b:
A note on the "Holder" class pattern (`RealmContextHolder`, `PolarisPrincipalHolder`, etc.):

These classes are meant to allow Polaris code to manage corresponding request-scoped data without [relying](https://github.com/apache/polaris/blob/26561f6798b37b168dcb9581a86f32d2c2f802fb/runtime/service/src/main/java/org/apache/polaris/service/context/RealmContextFilter.java#L55) on the REST framework (`ContainerRequestContext`) during async task execution.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
flyingImer: Agreed, separate PR. RequestIdSupplier is a public polaris-core interface, so the rename needs to be done thoughtfully for backwards compatibility.

flyingImer:
> Worth noting: this PR introduces a new `TaskContextPropagator` bean. The bean is functionally equivalent to a custom `org.eclipse.microprofile.context.spi.ThreadContextProvider`.
> 
> I wonder if it wouldn't be cleaner to just wire up a custom `ThreadContextProvider` instead of providing an equivalent class that must be invoked explicitly at the beginning of each task.
> 
> The custom `ThreadContextProvider` would still require to clear `ThreadContext.CDI` + `@ActivateRequestContext` + manual population of beans. So it wouldn't look radically different for sure, but we would at least benefit from implicit execution.

@adutra @dimas-b 
I spiked this. It can work, but the cost outweighs the benefit. You're right that .cleared(CDI) + @ActivateRequestContext + manual holder population stays the same either way. The difference is in what the SPI route adds on top IIUC.

1. The provider is ServiceLoader-instantiated, so no @Inject. Holder lookups go through CDI.current(). That's livable on its own, but resolving the principal during capture triggers Mutiny context propagation, which re-enters the provider. You need a recursion guard to break the cycle.

2. The provider also fires globally, not just for task submissions. Bootstrapper has no request scope when it submits, so you need a scope-active check. Same for any future executor user.

3. There's also the question of who owns the worker's request scope. .cleared(CDI) and the custom provider both want to manage it. Either you rely on SmallRye's ordering between built-in and ServiceLoader providers, or the provider takes over entirely. Neither is clean.

Bottomline: ~50 lines of defensive code to remove ~3 explicit lines from TaskExecutorImpl. The explicit helper avoids all of that because it's a regular CDI bean.

The implicit execution benefit is real but thin here. We're propagating three app-specific values at one call site, not a cross-cutting concern. I'd keep the explicit helper. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
flyingImer: With the holder propagation in place, the worker's fresh request scope can produce a correct CallContext through CDI, so the clone is redundant for context. Removing it means dropping CallContext as a parameter from four methods and switching to CDI-injected access. I feel it deserves a separate pr. WDYT? @dimas-b @adutra

dimas-b:
@flyingImer : so, you prefer to go with the current approach in this PR?

flyingImer:
> @flyingImer : so, you prefer to go with the current approach in this PR?

@dimas-b yes, I believe this is a balanced approach at the moment

adutra:
> That's livable on its own, but resolving the principal during capture triggers Mutiny context propagation, which re-enters the provider. You need a recursion guard to break the cycle.

That's true, I had to implement a re-entrance guard in my own attempt. Not the end of the world, but I agree that it adds some complexity.

OK, I think I'm fine with the current approach.

flyingImer:
@dimas-b @adutra @jbonofre I think the group is aligned on the PR impl directions and they are now reflected in the PR. Could you please take another look once get a chance? Would like to get it merged, so that I can follow up with other corresponding PRs

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
```diff
@@ -96,22 +90,18 @@ public TaskExecutorImpl(
       Clock clock,
       MetaStoreManagerFactory metaStoreManagerFactory,
       TaskFileIOSupplier fileIOSupplier,
-      RealmContextHolder realmContextHolder,
       PolarisEventDispatcher polarisEventDispatcher,
       PolarisEventMetadataFactory eventMetadataFactory,
       @Nullable Tracer tracer,
-      PolarisPrincipalHolder polarisPrincipalHolder,
-      PolarisPrincipal polarisPrincipal) {
+      TaskContextPropagator taskContextPropagator) {
```
adutra, line 96: nit: the injected bean is package-private, but the constructor is public. I'd suggest making `TaskContextPropagator` public, or conversely, making `TaskExecutorImpl` package-privage (class and constructor).

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
adutra, line 148: @flyingImer why do we still need this code? Can we move it to `TaskContextPropagator`?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
flyingImer: Agreed, prefer shrinking `TaskExecutorImpl`. Only the same-package test instantiates it directly; external callers go through `TaskExecutor`. Will make the class and constructors package-private.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
flyingImer: Per my understanding, the clone exists because the worker used to have no request scope: the original `CallContext` would be closed by the time the task ran. With `@ActivateRequestContext` and holder propagation now in place, a worker-produced `CallContext` carries the right values, so the clone is redundant.

It's also carrying some lifecycle coupling today: `CallContext.copy()` reuses the source request's `BasePersistence` session and `RealmConfigurationSource` references across threads. Going through CDI on the worker re-runs the producer chain against the worker's scope.

Rather not fold it into `TaskContextPropagator`. The propagator captures individual request values; `CallContext.copy()` is a bean-lifecycle workaround, and bundling them would still carry the source session into the task. Right fix is dropping the clone and the `CallContext` parameter from `tryHandleTask` / `handleTaskWithTracing` / `handleTask`, then injecting `CallContext` via CDI on the worker. Follow-up so this PR stays on propagation. WDYT?

flyingImer:
@dimas-b @adutra @jbonofre Addressed the latest inline comments: `TaskExecutorImpl` is now package-private (a4750bac5), and the `CallContext.copy()` removal is covered in the thread reply as a follow-up PR. CI is green, open threads resolved. Could you take a final pass?

runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java
adutra: Agreed that the proper fix is remove the parameter and switch to CDI injection. OK to do that in a follow-up task.

adutra (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
```diff
@@ -0,0 +1,64 @@
...
+import jakarta.enterprise.inject.Produces;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.polaris.core.context.RequestIdSupplier;
+
+/**
+ * Request-scoped holder for the request ID.
+ *
+ * <p>On HTTP request threads the request ID is set by {@code RequestIdFilter}. On async task
+ * threads a new CDI request scope is activated by {@code TaskExecutorImpl}, and the request ID from
+ * the originating request is propagated into this holder by {@code RequestIdPropagator}.
```
flyrain: RequestIdPropagator -> TaskContextPropagator?

btw, do we need to mention how this bean be used? I think it isn't necessary.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * <p>Must be called on a thread with an active CDI request scope (e.g. an HTTP request thread or
+   * a task thread that already has context restored). The principal is cloned into an immutable
+   * snapshot so the captured context is independent of the originating scope's lifecycle.
+   */
+  CapturedTaskContext capture() {
+    return new CapturedTaskContext(
+        realmContextHolder.get(),
+        ImmutablePolarisPrincipal.builder().from(polarisPrincipal).build(),
+        requestIdHolder.get());
```
flyrain, line 77: I think we should also include the `callContext` bean here. I'm fine with a followup though.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyrain: `callContext` is cloned at submission and already carries the realm context we care about.  So CapturedTaskContext doesn't need to store realm. In fact, downstream code use the one from the `callContext` in multiple places:
1. https://github.com/flyrain/polaris/blob/cfb175441de39bae1bdc5b362958b2029e56b614/runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java#L215.
2. https://github.com/flyrain/polaris/blob/cfb175441de39bae1bdc5b362958b2029e56b614/runtime/service/src/main/java/org/apache/polaris/service/task/TaskExecutorImpl.java#L290

Looks like we didn't change these places. I'd suggest to avoid having the realmContext here to avoid any future fragmentation, as we have two ways to use realmContext now.

flyrain (commented):
Thanks @flyingImer for the change. The PR looks great overall. The only major concern is the extra realmContext we passed into the task execution.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyingImer: You're right that realm currently flows through two paths on the worker (`callContext.getRealmContext()` at L206/L276 and the holder via captured state). The fragmentation is real but between the clone path (pre-existing) and the holder path (this PR), not within `CapturedTaskContext` itself.

Holder propagation is structurally required regardless: `RangerPolarisAuthorizerFactory` and `PolarisEventMetadataFactory` `@Inject RealmContext` directly, so the holder must be populated even if `callContext` is also passed.

Current `CapturedTaskContext` shape (`{realm, principal, requestId}`) matches the follow-up's target state. The `CallContext.copy()` removal follow-up (thread on `TaskExecutorImpl.java:148`) drops the `callContext` parameter and has the worker produce `CallContext` via CDI, collapsing the two paths into one without changing `CapturedTaskContext`.

Moving realm out now would require adding it back in the follow-up. Principal precedent: principal is captured as a value, not bundled in a bean, so keeping realm symmetric aligns with existing convention.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyingImer: The direction agreed on the `TaskExecutorImpl.java:148` thread is removing `callContext` from the worker path (follow-up PR), not bundling it into `CapturedTaskContext`. See the thread on `TaskContextPropagator.java:75` for the fragmentation discussion.

runtime/service/src/main/java/org/apache/polaris/service/context/catalog/RequestIdHolder.java
flyingImer: Trimmed the paragraph to keep only the contract (f55d384f0). The "populated by" description was implementation detail that already rotted once, and any future rename hits the same problem. Grep finds the setters.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyrain: My comment is NOT related to holder and principal. It is about the duplicated realm context pass-by. Why do we need to add the real context back? We can always visit it via call context.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyrain: Can you post the link? I don't think we can easily remove the call context. There are downstream reference to its field `PolarisCallContext`

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyingImer: The two realm paths today are a transitional overlap: (a) `callContext` clone parameter, (b) `CapturedTaskContext` → `RealmContextHolder`. Path (b) is the mechanism taking over, path (a) is the legacy being retired.

After the follow-up ([thread](https://github.com/apache/polaris/pull/4061#discussion_r3161773807)), `callContext` stops being a cross-thread parcel on the worker side. It becomes an `@Inject`-produced bean in the worker's request scope, the same role it has on HTTP threads. CDI produces it using `RealmContext` from the holder that `TaskContextPropagator` populated. At that point, the only mechanism for realm to reach the worker is captured → holder → CDI. Realm must be in `CapturedTaskContext` for that to work.

So the current shape (`{realm, principal, requestId}` captured) is the end-state shape, not a duplication choice. Dropping realm from `CapturedTaskContext` in this PR would need to re-add it in the follow-up.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyingImer: adutra's ack on the follow-up direction: https://github.com/apache/polaris/pull/4061#discussion_r3161773807

To be precise: `CallContext` the class stays, and downstream code keeps calling `ctx.getPolarisCallContext()` / `ctx.getRealmContext()` unchanged. What changes is how `callContext` appears on the worker.

Today, the submitter clones `callContext` and passes the clone through `tryHandleTask` / `handleTaskWithTracing` / `handleTask` as a parameter. It's a plain object ferried across threads.

After the follow-up, the worker gets `callContext` the same way HTTP threads already get it: via `@Inject CallContext`, with CDI producing it in the worker's request scope from the populated holders. The role shifts from "parcel passed across the boundary" to "bean produced in the local scope", which is its normal role elsewhere in the codebase. Consumer code at the call site is unchanged.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
flyingImer: @flyrain Stepping back to make sure we're aligned on both the near-term shape and the longer-term direction.

**Why `CapturedTaskContext.realmContext` stays**

Reading your concern as "don't let worker threads grow a second realm access path on top of what `CallContext` already provides" — that's a legitimate design principle, and I think the PR's shape actually serves it rather than fighting it.

`RealmContextHolder` isn't introduced here. It's been in the repo since the initial commit, and existing HTTP-path code already injects realm through it (e.g. `RangerPolarisAuthorizerFactory`, `PolarisEventMetadataFactory`). What PR #4061 adds is a capture/restore pair that populates the holder on the worker's fresh request scope, so the existing CDI convention keeps working across the async boundary instead of breaking at it.

`CapturedTaskContext.realmContext` is specifically the value that feeds that holder on the worker side. Dropping it would leave the worker's holder empty, which breaks not just `@Inject RealmContext` but also the `CallContext` producer itself (the producer reads `RealmContext` through the holder). So it's load-bearing, not additive.

**What the follow-up collapses**

The current two-path situation on the worker (clone `.getRealmContext()` vs holder-backed access) is transitional. The follow-up @adutra acked does:

1. Drop the `CallContext` parameter from the three worker methods (`tryHandleTask`, `handleTaskWithTracing`, `handleTask`)
2. Replace it with `@Inject CallContext` as a field
3. Drop the `callContext.copy()` call in `addTaskHandlerContext`

After that, the worker's `CallContext` is produced by CDI from the populated holder, the same way HTTP threads produce it today. One data source (`RealmContextHolder`), two access writings (`@Inject RealmContext` and `ctx.getRealmContext()`) that converge on the same instance. Same composition pattern the HTTP path uses already.

**If "one access pattern" is the goal**

Worth naming that there's a direction here. The two writings aren't symmetric (below sheet created with help of Claude code):

| | `@Inject RealmContext` | `ctx.getRealmContext()` |
|---|---|---|
| Source | `RealmContextHolder` | same, via `CallContext` producer |
| Instance identity | original | original (post-follow-up); lambda repackaging (today's clone path) |
| Dependency declared | narrow | full bag |
| CDI idiomatic | yes | wrapper |
| Covers every realm-access scenario | yes | yes |

`@Inject RealmContext` covers every scenario `ctx.getRealmContext()` does, and it's the narrower, more idiomatic form. The reverse direction (collapsing toward `ctx.getRealmContext()`) would require retrofitting every existing `@Inject RealmContext` site to go through `CallContext`, which widens dependencies and is the path the codebase has historically not taken.

So if we want to reduce to a single access API in the longer run, I'd suggest the right move is to deprecate `CallContext.getRealmContext()` in favor of `@Inject RealmContext`. Separate RFC, not in this PR's scope, but I'm happy to open it after this lands.

**Want to confirm we're aligned on**:

1. `CapturedTaskContext.realmContext` stays in this PR (it's the infrastructure, not a duplicate)
2. The follow-up collapses the transitional two-path state to one data source
3. Any further collapse to a single access API is a separate RFC, and the direction is deprecating `ctx.getRealmContext()` rather than the other way around

If any of those three don't match your read, happy to dig in further.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventMetadataFactory.java
```diff
@@ -100,23 +99,18 @@ private Optional<PolarisPrincipal> getUser() {
...
-    // See org.jboss.resteasy.reactive.server.injection.ContextProducers
-    ResteasyReactiveRequestContext context = CurrentRequestManager.get();
-    if (context != null) {
-      ContainerRequestContextImpl request = context.getContainerRequestContext();
-      String requestId = (String) request.getProperty(RequestIdFilter.REQUEST_ID_KEY);
-      return Optional.ofNullable(requestId);
+    try {
+      return Optional.ofNullable(requestIdHolder.get());
+    } catch (ContextNotActiveException e) {
+      // No active request scope (e.g. background thread without @ActivateRequestContext).
```
dimas-b, line 111: I'm still hesitant about this... This method is called on creating a `PolarisEventMetadata`. If request context is not active at that time, it would be a logical (coding) mistake.

I tend to think we should not try to catch this exception. WDYT?

dimas-b (commented):
LGTM overall, just a couple of minor concerns remaining from my side. One is below. I'll comment separately on the other one.

runtime/service/src/main/java/org/apache/polaris/service/task/TaskContextPropagator.java
dimas-b: I think `CapturedTaskContext.realmContext` should still be a fresh object reusing only the realm ID (`String`) from the parent request context.

`RealmContext` can be a CDI beam, and in that case it will not be reusable after its owner context is terminated.

dimas-b: I'm fine with dealing with `CallContext` in a follow-up PR.

dimas-b:
@flyingImer : I think this PR had valuable changes in it (despite some unresolved comments). If you feel like re-opening it, I'll support working through the remaining issues and merging.

PR #4064, opened 2026-03-26
Create PolarisEvent only when event type is enabled by at least one event listener

nandorKollar:
This change introduces new configuration option to allow each listener to listen only a subset of each Polaris event. The change also includes an optimization: the PolarisEvent doesn't get created and it isn't dispatched if there's no even listener listening on the given event type.

The PR is based on #3442 and incorporates recommendations from @adutra .

Co-authored-by: Oleg Soloviov <40199597+olsoloviov@users.noreply.github.com>

## Checklist


runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -41,13 +41,26 @@ public class PolarisEventListeners {
...
+      if (listenerConfiguration != null && !listenerConfiguration.enabledEventTypes().isEmpty()) {
+        supportedTypes = listenerConfiguration.enabledEventTypes().toArray(PolarisEventType[]::new);
+      }
+      for (var polarisEventType : supportedTypes) {
+        eventsByType.set(polarisEventType.ordinal());
+      }
+      var listener = eventListeners.select(Identifier.Literal.of(enabledEventListener)).get();
       Handler<Message<PolarisEvent>> handler = e -> listener.onEvent(e.body());
-      eventBus.localConsumer(POLARIS_EVENT_CHANNEL, handler);
+      eventBus.localConsumer(POLARIS_EVENT_CHANNEL + "." + event, handler);
```
adutra: This doesn't look correct to me. Appending the startup event to the channel name does not make sense imho.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisServiceBusEventDispatcher.java
```diff
@@ -40,6 +41,11 @@ public class PolarisServiceBusEventDispatcher implements PolarisEventDispatcher
    */
   @Override
   public void dispatch(PolarisEvent event) {
-    eventBus.publish(POLARIS_EVENT_CHANNEL, event);
+    eventBus.publish(POLARIS_EVENT_CHANNEL + "." + event, event);
```
adutra: Same here. There should be one channel per *event type*.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
nandorKollar: Right, this is not correct, will fix it soon.

flyingImer (commented):
Thanks — I walked through the latest revision and I don’t see an obvious correctness blocker after the event-bus address fix.

The main thing I’d still like to see is focused regression coverage for the new behavior introduced here:

- a listener configured with enabledEventTypes only receives that subset;
- when no listener is registered for a type, we really skip the PolarisEvent creation / dispatch path;
- the deprecated polaris.event-listener.type fallback still behaves as expected alongside the new multi-listener config.

This PR is changing both config semantics and the dispatch fast path, so having those cases pinned down in tests would make future refactors a lot safer.

site/content/in-dev/unreleased/configuration/config-sections/smallrye-polaris_event_listener.md
```diff
@@ -25,5 +25,6 @@ build:
...
..._VIEWS, BEFORE_LOAD_VIEW, AFTER_LOAD_VIEW, BEFORE_CHECK_EXISTS_VIEW, AFTER_CHECK_EXISTS_VIEW, BEFORE_DROP_VIEW, AFTER_DROP_VIEW, BEFORE_RENAME_VIEW, AFTER_RENAME_VIEW, BEFORE_REPLACE_VIEW, AFTER_REPLACE_VIEW, BEFORE_REFRESH_VIEW, AFTER_REFRESH_VIEW, BEFORE_LOAD_CREDENTIALS, AFTER_LOAD_CREDENTIALS, BEFORE_COMMIT_TRANSACTION, AFTER_COMMIT_TRANSACTION, BEFORE_SEND_NOTIFICATION, AFTER_SEND_NOTIFICATION, BEFORE_GET_CONFIG, AFTER_GET_CONFIG, BEFORE_CREATE_POLICY, AFTER_CREATE_POLICY, BEFORE_LIST_POLICIES, AFTER_LIST_POLICIES, BEFORE_LOAD_POLICY, AFTER_LOAD_POLICY, BEFORE_UPDATE_POLICY, AFTER_UPDATE_POLICY, BEFORE_DROP_POLICY, AFTER_DROP_POLICY, BEFORE_ATTACH_POLICY, AFTER_ATTACH_POLICY, BEFORE_DETACH_POLICY, AFTER_DETACH_POLICY, BEFORE_GET_APPLICABLE_POLICIES, AFTER_GET_APPLICABLE_POLICIES, BEFORE_CREATE_GENERIC_TABLE, AFTER_CREATE_GENERIC_TABLE, BEFORE_DROP_GENERIC_TABLE, AFTER_DROP_GENERIC_TABLE, BEFORE_LIST_GENERIC_TABLES, AFTER_LIST_GENERIC_TABLES, BEFORE_LOAD_GENERIC_TABLE, AFTER_LOAD_GENERIC_TABLE, BEFORE_ATTEMPT_TASK, AFTER_ATTEMPT_TASK, BEFORE_LIMIT_REQUEST_RATE` | Comma separated list of enabled event types. This event listener will only receive events of  the selected types.  |
```
nandorKollar: To avoid 150 different event types, we could consider each event to have a 'category' or 'entityType'. One can then configure an event listener to consume only a set of event categories, instead of listing all of the event which are related to for example Catalog.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
```diff
@@ -146,9 +145,6 @@ void shutdown() {
 
   @Override
   public void onEvent(PolarisEvent event) {
-    if (event.type() != PolarisEventType.AFTER_REFRESH_TABLE) {
```
nandorKollar, line 149: I removed this, because it should be part of the configuration for cloud watch listener. To remain compatible, we could consider defaulting `AFTER_REFRESH_TABLE` as the only event type for AWS cloudwatch in the application config.

adutra: The feature is flagged as beta/experimental, so I wouldn't really care about that – also, it's a weird default value imho.

site/content/in-dev/unreleased/configuration/config-sections/smallrye-polaris_event_listener.md
adutra: Good idea 👍

dimas-b: +1

dimas-b:
@nandorKollar : since this PR builds on #3442 , would you mind tagging @olsoloviov as a [co-author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) (in the PR description to be later propagated to the commit message)?

nandorKollar:
> co-author

Sure, that's a great idea!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
```diff
@@ -60,52 +60,64 @@ public class PolarisCatalogsEventServiceDelegator implements PolarisCatalogsApiS
   @Override
   public Response createCatalog(
       CreateCatalogRequest request, RealmContext realmContext, SecurityContext securityContext) {
-    polarisEventDispatcher.dispatch(
-        new PolarisEvent(
-            PolarisEventType.BEFORE_CREATE_CATALOG,
-            eventMetadataFactory.create(),
-            new EventAttributeMap()
-                .put(EventAttributes.CATALOG_NAME, request.getCatalog().getName())));
+    if (polarisEventDispatcher.hasListeners(PolarisEventType.BEFORE_CREATE_CATALOG)) {
```
jbonofre, line 63: nit: We have a bunch of boilerplate here around create/delete/etc.

Maybe it would be cleaner to a helper like `dispatchIfListening(Supplier<PolarisEvent>)` or similar to simplify the maintenance.

adutra: > Maybe it would be cleaner to a helper like `dispatchIfListening(Supplier<PolarisEvent>)` or similar to simplify the maintenance.

You would still be creating a lambda + closure if you do this. Only the if block allows to completely avoid creating objects that would be discarded immediately if there are no listeners for them.

nandorKollar: Ideally, we should use AOP for events, with Quarkus `@AroundInvoke` and introducing an `@PolarisEvent` annotation we could make this simple and clean. The problem is that I couldn't find a way to pass the event attribute map.

I'm also not too happy with the bolierplate if-s, it is very easy to forget that later when a new event is added, however I don't know what could be a better alternative. Like @adutra mentioned, using lambdas is also suboptimal approach.

adutra: I thought about AOP too but as you pointed out, the way we accumulate event attributes gets in the way.

I think the if blocks are the least worst solution. If we are concerned about missing new event types, we probably could introduce a test that verifies that every event type is correctly received by a listener accepting all event types.

jbonofre: That's good points, indeed. I agree with @adutra , the "if" chain is probably the least worst.

Thanks !

nandorKollar:
> Thanks — I walked through the latest revision and I don’t see an obvious correctness blocker after the event-bus address fix.
> 
> The main thing I’d still like to see is focused regression coverage for the new behavior introduced here:
> 
> * a listener configured with enabledEventTypes only receives that subset;
> * when no listener is registered for a type, we really skip the PolarisEvent creation / dispatch path;
> * the deprecated polaris.event-listener.type fallback still behaves as expected alongside the new multi-listener config.
> 
> This PR is changing both config semantics and the dispatch fast path, so having those cases pinned down in tests would make future refactors a lot safer.

Thanks @flyingImer for the review. I added some more test coverage, would you mind have a look at the latest version, and share your thoughts?

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
adutra: Some more thoughts on this:

The if-block approach was designed initially with serialization in mind. IOW, the event was created with lots of serialized attributes (either JSON or `toString()` calls - I don't recall exactly). So, creating an event could be expensive.

But now that we have the events attribute map, creating a `PolarisEvent` has become much cheaper — it's just a wrapper around a thin map of references. The if block is now avoiding roughly:

  - One `PolarisEvent` object (~32 bytes)
  - One `EventAttributeMap` object + its internal map (~200 bytes)
  - A handful of `Map.Entry` objects for the `.put()` calls

All of these are short-lived. It's not nothing, but it's not the end of the world either.

One edge case where the guard could matter is if an attribute value is computed lazily solely for the event, e.g., a `.toString()` that creates a new object. Looking at the call sites, I don't think it's happening today (**but it could, in the future**).

The `Supplier<PolarisEvent>` approach imho adds more overhead in the happy path (when there are listeners) to save a small memory overhead in the no-listeners path. All in all, it's not a perfect solution either.

So just to summarize, I'd still stick with the if block, mostly because of the risk of having some expensive-to-compute attribute added to the event.

nandorKollar: Thanks Alex for the helping out with a detailed explanation. I think that the extra if statements don't add much complexity. My only concern is what I mentioned before: one can easily forget wrapping the dispatch calls around. Also, I don't think we can easily verify, that anytime one calls dispatch, it is wrapped around with that if statement. Nonetheless, I also vote keeping this approach.

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventListenersTest.java
```diff
@@ -0,0 +1,237 @@
...
+
+@QuarkusTest
+@TestProfile(PolarisEventListenersTest.PolarisEventListenersTestProfile.class)
+public class PolarisEventListenersTest {
+  static final Set<PolarisEventType> CATALOG_EVENTS =
+      Arrays.stream(PolarisEventType.values())
+          .filter(e -> e.category() == PolarisEventType.Category.CATALOG)
+          .collect(Collectors.toSet());
+
+  static final CountDownLatch latch = new CountDownLatch(5 + 3 * CATALOG_EVENTS.size());
```
adutra: This latch looks a bit fragile here imo (the latch size in particular looks a bit "magic"). Would it be possible to use Awaitility instead?

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventListenersTest.java
```diff
@@ -0,0 +1,237 @@
...
+
+  private abstract static class FilteringEventListener implements PolarisEventListener {
+    private final Predicate<PolarisEvent> predicate;
+
+    List<PolarisEvent> expectedEvents = new ArrayList<>();
```
adutra: I would go with `CopyOnWriteArrayList` here – events can be delivered concurrently.

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventListenersTest.java
```diff
@@ -0,0 +1,237 @@
...
+  @Inject
+  @Identifier("consume-only-catalog-events-listener")
+  PolarisEventListener consumeOnlyCatalogEventListener;
+
+  @Inject
+  @Identifier("consume-catalog-and-after-notification-events-listener")
+  PolarisEventListener consumeCatalogAndNotificationEventsListener;
+
+  @Test
+  public void testEventListenersGetNotified() throws InterruptedException {
```
adutra: One of the tests we had in `PolarisEventListenersMultipleListenersTest` was testing that *two* listeners without any type filtering both receive all events. We don't have this specific test here. Not sure we absolutely must test this scenario, but I wanted to note the difference.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
```diff
@@ -41,4 +44,25 @@ public interface PolarisEventListenerConfiguration {
...
   Optional<Set<String>> types();
+
+  /** Configuration of each event listener type. */
+  @WithParentName
+  Map<String, ListenerConfiguration> listenerConfig();
+
+  interface ListenerConfiguration {
+    /**
+     * Comma separated list of enabled event types. This event listener will only receive events of
+     * the selected types. If both the event types and event category configs are set, the listener
```
adutra, line 55: I think we should also mention what the defaults are:

- If no listener configuration exists, all events are enabled.
- If a listener configuration exists, events are disabled by default (unless explicitlty listed in either `enabled-event-types` or `enabled-event-categories`).

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -41,13 +42,35 @@ public class PolarisEventListeners {
...
+        if (listenerConfiguration.enabledEventCategories().isPresent()) {
+          for (var enabledEventCategory : listenerConfiguration.enabledEventCategories().get()) {
+            enabledEventTypes.addAll(PolarisEventType.typesOfCategory(enabledEventCategory));
+          }
+        }
+        if (listenerConfiguration.enabledEventTypes().isPresent()) {
+          enabledEventTypes.addAll(listenerConfiguration.enabledEventTypes().get());
+        }
+        supportedTypes = enabledEventTypes.toArray(PolarisEventType[]::new);
+      }
```
adutra: Minor suggestion:

```suggestion
      Set<PolarisEventType> supportedTypes;
      if (listenerConfiguration != null) {
        supportedTypes = EnumSet.noneOf(PolarisEventType.class);
        if (listenerConfiguration.enabledEventCategories().isPresent()) {
          for (var enabledEventCategory : listenerConfiguration.enabledEventCategories().get()) {
            supportedTypes.addAll(PolarisEventType.typesOfCategory(enabledEventCategory));
          }
        }
        if (listenerConfiguration.enabledEventTypes().isPresent()) {
          supportedTypes.addAll(listenerConfiguration.enabledEventTypes().get());
        }
      } else {
        // No configuration: enable all event types
        supportedTypes = EnumSet.allOf(PolarisEventType.class);
      }
```

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -41,13 +42,35 @@ public class PolarisEventListeners {
...
+      var listener = eventListeners.select(Identifier.Literal.of(enabledEventListener)).get();
       Handler<Message<PolarisEvent>> handler = e -> listener.onEvent(e.body());
```
adutra: I suggest extracting the delivery logic to a method and surround with exception handling + debug logging, because handlers are not supposed to throw:

```java
  Handler<Message<PolarisEvent>> handler =
          message -> deliverEvent(message.body(), enabledEventListener, listener);
      for (var polarisEventType : supportedTypes) {
        eventsByType.set(polarisEventType.ordinal());
        eventBus.localConsumer(POLARIS_EVENT_CHANNEL + "." + polarisEventType, handler);
      }
    }
  }

  private void deliverEvent(
      PolarisEvent event, String listenerName, PolarisEventListener listener) {
    LOGGER.debug("Delivering {} event to listener '{}' ({})", event.type(), listenerName, listener);
    try {
      listener.onEvent(event);
    } catch (Exception e) {
      LOGGER.error(
          "Error while delivering {} event to listener '{}' ({})",
          event.type(),
          listenerName,
          listener,
          e);
    } finally {
      LOGGER.debug(
          "Delivered {} event to listener '{}' ({})", event.type(), listenerName, listener);
    }
  }
```

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventListenersTest.java
nandorKollar: I haven't heard about Awaitility, is it this one: https://github.com/awaitility/awaitility? It would require an extra dependency, do we need that? I agree the latch size looks indeed magic, we can refactor that to a constant, and write a proper comment for it.

adutra: We already have Awaitility! See here for an example:

https://github.com/apache/polaris/blob/eb28eda6cdde1d7c24280112eef108f3a367bc6c/integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisApplicationIntegrationTest.java#L606-L626

nandorKollar: Awesome, I can try it out in this test too!

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventListenersTest.java
```diff
@@ -0,0 +1,259 @@
...
+      assertEquals(2 + CATALOG_EVENTS.size(), listener.consumedEvents.size());
+      expectEvents(
+          listener.consumedEvents,
+          Stream.concat(
+                  Stream.of(
+                      PolarisEventType.AFTER_SEND_NOTIFICATION,
+                      PolarisEventType.BEFORE_SEND_NOTIFICATION),
+                  CATALOG_EVENTS.stream())
+              .collect(Collectors.toSet()));
+    }
```
adutra, line 221: Minor suggestion to improve readability:


```suggestion
    List<PolarisEvent> allEvents = new ArrayList<>();
    allEvents.add(new PolarisEvent(PolarisEventType.AFTER_SEND_NOTIFICATION, null));
    allEvents.add(new PolarisEvent(PolarisEventType.BEFORE_SEND_NOTIFICATION, null));
    CATALOG_EVENTS.forEach(eventType -> allEvents.add(new PolarisEvent(eventType, null)));

    allEvents.forEach(event -> eventDispatcher.dispatch(event));

    // Only after send notification event
    await().until(() -> afterSendEventListener.receivedEventCount.get() == 1);
    assertEquals(1, afterSendEventListener.expectedEvents.size());
    assertEquals(
        PolarisEventType.AFTER_SEND_NOTIFICATION,
        afterSendEventListener.expectedEvents.getFirst().type());

    // Only before send notification event
    await().until(() -> beforeSendEventListener.receivedEventCount.get() == 1);
    assertEquals(1, beforeSendEventListener.expectedEvents.size());
    assertEquals(
        PolarisEventType.BEFORE_SEND_NOTIFICATION,
        beforeSendEventListener.expectedEvents.getFirst().type());

    // All events received by both consumeAllEventListener and consumeAllEventListener2
    for (var listener :
        new BaseConsumeAllEventListener[] {consumeAllEventListener, consumeAllEventListener2}) {
      await().until(() -> listener.receivedEventCount.get() == allEvents.size());
      assertEquals(allEvents.size(), listener.consumedEvents.size());
      expectEvents(
          listener.consumedEvents,
          allEvents.stream().map(PolarisEvent::type).collect(Collectors.toSet()));
    }
```

runtime/service/src/test/java/org/apache/polaris/service/events/PolarisEventListenersTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+    // There are no unexpected events
+    assertTrue(afterSendEventListener.unexpectedEvents.isEmpty());
+    assertTrue(beforeSendEventListener.unexpectedEvents.isEmpty());
+    assertTrue(consumeOnlyCatalogEventListener.unexpectedEvents.isEmpty());
+    assertTrue(consumeCatalogAndNotificationEventListener.unexpectedEvents.isEmpty());
+  }
+
+  private void expectEvents(List<PolarisEvent> events, Set<PolarisEventType> expectedTypes) {
+    for (var event : expectedTypes) {
```
adutra: nit:
```suggestion
    for (var eventType : expectedTypes) {
```

adutra (dismissed):
Thank you @nandorKollar !

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
dimas-b: Sorry for coming later to this party 😅 

Using AOP is very interesting in this case. However, rather that using it on the leaf call to dispatch an event, we could generate the whole interceptor / delegator class (probably by making a Quarkus plugin) and only call the event-producing code when a listener is present.

This would way outside the scope of current PR, though :)

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisServiceBusEventDispatcher.java
```diff
@@ -40,6 +41,11 @@ public class PolarisServiceBusEventDispatcher implements PolarisEventDispatcher
    */
   @Override
   public void dispatch(PolarisEvent event) {
-    eventBus.publish(POLARIS_EVENT_CHANNEL, event);
+    eventBus.publish(POLARIS_EVENT_CHANNEL + "." + event.type(), event);
```
dimas-b: nit: would it be reasonable to double check that the event type is enabled? I suppose the overhead is not huge, but the benefit is safeguarding against mistakes in enabled checks at call sites 🤔

dimas-b (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListenerConfiguration.java
adnanhemani: nit: "listeners"

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -31,23 +31,68 @@
...
 
 @ApplicationScoped
 public class PolarisEventListeners {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PolarisEventListeners.class);
+
   @Inject EventBus eventBus;
   @Inject @Any Instance<PolarisEventListener> eventListeners;
   @Inject PolarisEventListenerConfiguration configuration;
 
+  private final BitSet eventsByType = new BitSet(PolarisEventType.values().length);
```
adnanhemani: IMO an EnumSet would be much more clear here.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -31,23 +31,68 @@
...
-    Set<String> listenerTypeSet = configuration.types().orElseGet(HashSet::new);
+    var listenerTypeSet = configuration.types().orElseGet(HashSet::new);
     for (String enabledEventListener : listenerTypeSet) {
-      PolarisEventListener listener =
-          eventListeners.select(Identifier.Literal.of(enabledEventListener)).get();
-      Handler<Message<PolarisEvent>> handler = e -> listener.onEvent(e.body());
-      eventBus.localConsumer(POLARIS_EVENT_CHANNEL, handler);
+      var listenerConfiguration = configuration.listenerConfig().get(enabledEventListener);
+      var supportedTypes =
+          listenerConfiguration == null
```
adnanhemani, line 55: This is a bit brittle logic. Say an Event Listener requires some specific configuration (unrelated to event types accepted) but the user does not want to explicitly define all of the PolarisEventTypes. What happens then?

IMO, we should ensure that we check for "enabledEventTypes" not being present at all.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
```diff
@@ -19,189 +19,223 @@
...
 
   public int code() {
     return code;
   }
+
+  public Category category() {
+    return category;
+  }
+
+  public static Set<PolarisEventType> typesOfCategory(Category category) {
```
adnanhemani, line 218: nit: we could probably pre-compute this on startup to reduce this filter+collect logic during runtime since this will always be static.

adnanhemani (commented):
Mostly looking good - but one last major concern in the comments. Thanks for this work @nandorKollar!

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisCatalogsEventServiceDelegator.java
nandorKollar: Thanks Dmitri, it is never too late to get to the party, as long as it still going on! :)

Do you happen to have any example for such a Quarkus plugin?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
nandorKollar: Makes sense to me, as EnumSet is implemented as a bit vector, the complexity of a lookup should be the same for both. @adutra what do you think?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
nandorKollar: Despite this doesn't seem to be a problem now, it is reasonable to expect that when someone extends `ListenerConfiguration`, then there should be no need to list every event type. How about this:
```
      var supportedTypes =
          listenerConfiguration == null || (listenerConfiguration.enabledEventCategories().isEmpty() && listenerConfiguration.enabledEventTypes().isEmpty())
              ? EnumSet.allOf(PolarisEventType.class)
              : EnumSet.noneOf(PolarisEventType.class);
```

This should make it future-proof though I don't think we can cover this with a test case, as `ListenerConfiguration` has only these two configs now.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
nandorKollar: We can introduce a map to improve lookup speed, if that would help, something like this:
```
  private static final Map<Category, EnumSet<PolarisEventType>> categoryToTypes = new HashMap<>();

  static {
    for (PolarisEventType type : values()) {
      EnumSet<PolarisEventType> types = categoryToTypes.getOrDefault(type.category, EnumSet.noneOf(PolarisEventType.class));
      types.add(type);
      categoryToTypes.put(type.category, types);
    }
  }
```

and rewrite `typeOfCategory`:
```
  public static Set<PolarisEventType> typesOfCategory(Category category) {
    return categoryToTypes.get(category);
  }
```

Did you think of something like this?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
adutra: Yes, good idea.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
dimas-b: @adnanhemani : Do you mean we should ask the listener for the set of event types it needs to process?

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
dimas-b: `typesOfCategory()` is only called on startup.... why bother with a separate pre-computation step?

dimas-b: the overhead of filtering ~200 enum values on startup is negligible, even if done a few times, IMHO.

nandorKollar: That's true, the benefit would be neglibile.

dimas-b:
@nandorKollar : from CI: `ERROR: Configuration reference is out of date. Please run './gradlew :polaris-config-docs-site:copyConfigSectionsToSite' and commit the changes.`

nandorKollar:
> @nandorKollar : from CI: `ERROR: Configuration reference is out of date. Please run './gradlew :polaris-config-docs-site:copyConfigSectionsToSite' and commit the changes.`

The only doc change was this one: https://github.com/apache/polaris/pull/4064/changes/cb31ed9f073b2843a9cd6d5b0089f03fac841c31

`./gradlew :polaris-config-docs-site:copyConfigSectionsToSite` doesn't add any new changes.

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
```diff
@@ -31,23 +31,69 @@
...
 
 @ApplicationScoped
 public class PolarisEventListeners {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PolarisEventListeners.class);
+
   @Inject EventBus eventBus;
   @Inject @Any Instance<PolarisEventListener> eventListeners;
   @Inject PolarisEventListenerConfiguration configuration;
 
+  private final EnumSet<PolarisEventType> eventsByType = EnumSet.allOf(PolarisEventType.class);
```
dimas-b: nit: maybe `enabledTypes`?

dimas-b (dismissed):

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventListeners.java
adnanhemani: I think what @nandorKollar got the logic right on this comment - thanks!

runtime/service/src/main/java/org/apache/polaris/service/events/PolarisEventType.java
adnanhemani: Oh, my bad - I misread this, I thought it was being used in the request path rather than just startup. This is fine with me then.

adnanhemani (dismissed):
Great work, @nandorKollar (and @olsoloviov for getting this work started 🙂)!

nandorKollar:
Oh one more thing: should I also mention this change in the changelog?

dimas-b (dismissed):

dimas-b:
Good point about CHANGELOG. This PR affects how users configure Polaris), but I believe this change is fully backward compatible, right?

nandorKollar:
> Good point about CHANGELOG. This PR affects how users configure Polaris), but I believe this change is fully backward compatible, right?

Yes, it should be. Or well, not sure if we should consider this: https://github.com/apache/polaris/pull/4064#discussion_r3029096621 as an incompatible modification or not.

runtime/service/src/main/java/org/apache/polaris/service/events/jsonEventListener/aws/cloudwatch/AwsCloudWatchEventListener.java
dimas-b: Delegating this check to configuration sounds reasonable to me... it's probably worth mentioning as a (non-breaking) "Change" in CHANGELOG.

nandorKollar:
Added a commit (13335ff0ef1cc3d6d1d83938baa280302433931f) which removes `If a listener configuration exists, events are  disabled by default (unless explicitly listed in either enabled-event-types or  enabled-event-categories).` sentence from the doc, since after 99a6c47de939053b689ac6598acb542fbc337ea6 it is no longer correct. Also, mentioned the new event configs in the changelog. cc @dimas-b @adutra @adnanhemani

adutra (dismissed):

dimas-b (dismissed):

adutra:
@nandorKollar do you mind fixing the conflicts? After that I'll merge immediately. Thanks 🙏

adnanhemani (approved):
Re-approving.

PR #4069, opened 2026-03-27
Core: Add null check and exception handling in ConnectionCredentials

flyrain:
Issue: Unhandled NumberFormatException could crash at runtime

It adds null check before parsing the timestamp and wraps Long.parseLong() in try-catch to handle NumberFormatException, The fix throws IllegalArgumentException with clear error message for invalid values. The fix prevents uncaught runtime exceptions when credentials contain malformed data.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
@@ -70,7 +70,20 @@ public interface ConnectionCredentials {
 
     default Builder put(CatalogAccessProperty key, String value) {
       if (key.isExpirationTimestamp()) {
-        expiresAt(Instant.ofEpochMilli(Long.parseLong(value)));
+        if (value == null) {
+          throw new IllegalArgumentException(
+              "Expiration timestamp value cannot be null for key: " + key.getPropertyName());
+        }
+        try {
+          expiresAt(Instant.ofEpochMilli(Long.parseLong(value)));
+        } catch (NumberFormatException e) {
+          throw new IllegalArgumentException(
+              "Invalid expiration timestamp value for key "
+                  + key.getPropertyName()
+                  + ": "
+                  + value,
+              e);
+        }
       }
 
       if (key.isCredential()) {
--- a/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
@@ -78,10 +78,7 @@ public interface ConnectionCredentials {
           expiresAt(Instant.ofEpochMilli(Long.parseLong(value)));
         } catch (NumberFormatException e) {
           throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -70,7 +70,17 @@ interface Builder {
 
     default Builder put(CatalogAccessProperty key, String value) {
       if (key.isExpirationTimestamp()) {
-        expiresAt(Instant.ofEpochMilli(Long.parseLong(value)));
+        if (value == null) {
```
jbonofre, line 73: nit: Looking at the users of `ConnectionCredentials` (e.g. `OAuthClientCredentialVendor`, `BearerConnectionCredentialVendor`, ...), they all pass `String.valueOf(Long.MAX_VALUE)`.
It means that the value is never `null` or malformed from internal code.
As `Long.parseLong(null)` already throws a `NumberFormatException`, the try-catch alone would handle it.

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -70,7 +70,17 @@ interface Builder {
 
     default Builder put(CatalogAccessProperty key, String value) {
       if (key.isExpirationTimestamp()) {
```
jbonofre, line 72: We do the null check inside the `isExpirationTimestamp()` block.
Shouldn't we do a null check in the `isCredential()` block as well ?

jbonofre (approved):
LGTM overall.

Just two minor comments/questions.

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -70,7 +70,17 @@ interface Builder {
...
-        expiresAt(Instant.ofEpochMilli(Long.parseLong(value)));
+        if (value == null) {
+          throw new IllegalArgumentException(
```
dimas-b, line 74: nit: `Preconditions.checkArgument(...)`

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -70,7 +70,17 @@ interface Builder {
...
+              "Expiration timestamp value cannot be null for key: " + key.getPropertyName());
+        }
+        try {
+          expiresAt(Instant.ofEpochMilli(Long.parseLong(value)));
+        } catch (NumberFormatException e) {
+          throw new IllegalArgumentException(
```
dimas-b, line 80: What difference does it make if this method throws a `NumberFormatException` or an `IllegalArgumentException`?

dimas-b:
From description:

> The fix prevents uncaught runtime exceptions when credentials contain malformed data.

I do not think this PR prevents uncaught runtime exceptions. It only changes the types of runtime exceptions thrown by `ConnectionCredentials`. It does not change any of the callers (who'd have to deal with catching those exceptions).

PR #4071, opened 2026-03-27
NoSQL: Optimize index-impl scratch-buffer reuse

snazy:
The current implementation of reusing scratch `ByteBuffer`s is correct, but can become inefficient over time. It would also be immediately inefficient with virtual threads (which we're not using though).

This change addresses the "pref regression over runtime" and the virtual-thread concern at the same time, while keeping the performance on-par with the current implementation. This has been validated using the existing JMH benchmarks running with JMH's `gc` and `prov` profiles. The remaining differences are rather noise or really marginal performance footprint differences.

### DIFF
--- a/persistence/nosql/persistence/impl/src/test/java/org/apache/polaris/persistence/nosql/impl/indexes/TestIndexImpl.java
+++ b/persistence/nosql/persistence/impl/src/test/java/org/apache/polaris/persistence/nosql/impl/indexes/TestIndexImpl.java
@@ -39,10 +39,8 @@ import static org.apache.polaris.persistence.nosql.impl.indexes.Util.asHex;
 import static org.apache.polaris.persistence.nosql.impl.indexes.Util.randomObjId;
 import static org.assertj.core.groups.Tuple.tuple;
 
-import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.HexFormat;
-import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;


persistence/nosql/persistence/impl/src/test/java/org/apache/polaris/persistence/nosql/impl/indexes/TestIndexImpl.java
```diff
@@ -869,11 +868,125 @@ void multithreaded(Consumer<KeyIndexTestSet<ObjRef>> worker, boolean longTest) t
...
+
+    // The striped pool has a fixed number of slots (POOL_SIZE), each holding at most one buffer.
+    // Releasing from different threads that hash to the same slot just overwrites.
+    soft.assertThat(IndexImpl.POOL_SIZE).isGreaterThan(0);
 
-      CompletableFuture.allOf(futures).get();
+    // After releasing buffers, pool size can never exceed POOL_SIZE
+    var buffers = new ArrayList<java.nio.ByteBuffer>();
+    for (var i = 0; i < IndexImpl.POOL_SIZE + 10; i++) {
+      buffers.add(IndexImpl.acquireScratchKeyBuffer());
```
dimas-b: I'm not sure how this is supposed to work 🤔 no matter how many times we call, the current thread ID (same) will be used to index into the buffer pool.... leading to all buffers being freshly allocated. That is fine, but how is it related to the pool size?

persistence/nosql/persistence/impl/src/test/java/org/apache/polaris/persistence/nosql/impl/indexes/TestIndexImpl.java
```diff
@@ -869,11 +868,125 @@ void multithreaded(Consumer<KeyIndexTestSet<ObjRef>> worker, boolean longTest) t
...
     }
+    buffers.forEach(IndexImpl::releaseScratchKeyBuffer);
```
dimas-b: We're releasing some freshly allocated buffers overwriting previous "releases"? 🤔

persistence/nosql/persistence/impl/src/test/java/org/apache/polaris/persistence/nosql/impl/indexes/TestIndexImpl.java
snazy: Ah, that test was from a discarded approach - it's useless with the current impl, so removed.
But I took the freedom to add some more assertions to another test.

PR #4095, opened 2026-03-31
Update Generated Helm Docs after First RC

adnanhemani:
Mirrors #4093.


.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -167,6 +173,12 @@ jobs:
           Changelog patched successfully
           EOT
 
+      - name: Generate Helm documentation
+        if: env.rc_number == '0'
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+          exec_process make helm-doc-generate
```
adutra: I didn't catch this in #4093 but I'm wondering if we should also generate the schema file as well?

In this case the command to run is `make helm-schema-generate helm-doc-generate `. 

We can also run `make helm` which also lints and runs the unit tests.

adnanhemani: I'm open to doing either (`make helm` or `make helm-schema-generate helm-doc-generate`). Does anyone else have thoughts on this?

adnanhemani: Updated!

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -167,6 +173,12 @@ jobs:
           Changelog patched successfully
           EOT
 
+      - name: Generate Helm documentation
```
adutra:
```suggestion
      - name: Generate Helm schema and documentation
```

.github/workflows/release-2-update-release-candidate.yml
```diff
@@ -167,6 +173,12 @@ jobs:
...
+        if: env.rc_number == '0'
+        run: |
+          source "${LIBS_DIR}/_exec.sh"
+          exec_process make helm-schema-generate helm-doc-generate
```
adutra, line 180:
```suggestion
          exec_process make helm-schema-generate helm-doc-generate
          cat <<EOT >> $GITHUB_STEP_SUMMARY
          ## Helm schema & documentation
          Helm schema and documentation files generated successfully.
          EOT
```

.github/workflows/release-2-update-release-candidate.yml
adutra, line 159: Now that we are generating all derived Helm files here (which I think is the right thing to do), do you mind modifying the `update_helm_version` in `releasey/libs/_version.sh` and remove this line (line 121):

```shell
  exec_process sed -E -i~ "s|/in-dev/unreleased/|/releases/${new_version}/|g" "$HELM_VALUES_SCHEMA_FILE"
```

As this line is not required anymore.

It might also be worth to add a comment there saying something like:

```
It's not necessary to update $HELM_VALUES_SCHEMA_FILE or $HELM_REFERENCE_DOC_FILE here, because these files will be generated from $HELM_VALUES_FILE during the release process, and they will automatically reflect the updated version.
```

adnanhemani: Great catch, done!

.github/workflows/release-2-update-release-candidate.yml
adnanhemani: Done!

.github/workflows/release-2-update-release-candidate.yml
adnanhemani: Done.

adutra (approved):

PR #4103, opened 2026-04-01
Core: Use record for connection credentials

flyrain:
`ConnectionCredentials` is more suitable as a record instead of an interface. 

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
@@ -35,36 +35,25 @@ import org.apache.polaris.core.storage.StorageAccessConfig;
  */
 public record ConnectionCredentials(Map<String, String> credentials, Optional<Instant> expiresAt) {
 
-  public static Builder builder() {
-    return new Builder();
-  }
+  public static final ConnectionCredentials EMPTY =
+      new ConnectionCredentials(Map.of(), Optional.empty());
 
-  public static final class Builder {
-    private final Map<String, String> credentials = new HashMap<>();
-    private Instant expiresAt;
-
-    public Builder putCredential(String key, String value) {
-      credentials.put(key, value);
-      return this;
-    }
-
-    public Builder expiresAt(Instant expiresAt) {
-      this.expiresAt = expiresAt;
-      return this;
-    }
-
-    public Builder put(CatalogAccessProperty key, String value) {
+  /**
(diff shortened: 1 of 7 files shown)


singhpk234:
Makes sense to me, thanks @flyrain !

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -33,52 +32,28 @@
...
-
-  interface Builder {
-    @CanIgnoreReturnValue
-    Builder putCredential(String key, String value);
-
-    @CanIgnoreReturnValue
-    Builder expiresAt(Instant expiresAt);
-
-    default Builder put(CatalogAccessProperty key, String value) {
+  public static ConnectionCredentials of(Map<CatalogAccessProperty, String> properties) {
```
adutra, line 45: Just FYI it is possible to generate a builder for records with the Immutables library. But I agree that here, a factory method is good enough.

runtime/service/src/main/java/org/apache/polaris/service/credentials/connection/BearerConnectionCredentialVendor.java
```diff
@@ -77,9 +78,11 @@ public BearerConnectionCredentialVendor(UserSecretsManager secretsManager) {
     // Bearer tokens don't expire from Polaris's perspective - set expiration to Long.MAX_VALUE
     // to indicate infinite validity. The token itself may have an expiration, but that's managed
     // by the token issuer, not Polaris.
-    return ConnectionCredentials.builder()
-        .put(CatalogAccessProperty.BEARER_TOKEN, bearerToken)
-        .put(CatalogAccessProperty.EXPIRES_AT_MS, String.valueOf(Long.MAX_VALUE))
-        .build();
+    return ConnectionCredentials.of(
```
gh-yzou, line 81: for some reason i kind of the like the original builder usage, makes caller little bit cleaner, but I am fine with that way also

dimas-b: I agree with @gh-yzou that on the caller side the previous builder pattern was more natural and probably less error-prone (non-blocking)

flyrain: Thanks @gh-yzou and @dimas-b, appreciate the feedback! I'm open to either options. Since this is a not a blocker, I will merge it as is. Feel free to improve it further.

PR #4116, opened 2026-04-03
Rename ExternalCatalogFactory to FederatedCatalogFactory

flyrain:
Rename ExternalCatalogFactory to FederatedCatalogFactory because "external catalog" is a broader concept than what this interface represents. An external catalog can be either:

  1. A federated catalog — Polaris connects to a remote catalog service (Iceberg REST, Hive, Hadoop) and delegates operations to it via a ConnectionConfigInfo.
  2. A notification-synced catalog — Polaris manages tables locally that are synced via the notification API, with no remote connection involved.

  This factory interface only applies to the first case: it takes a ConnectionConfigInfoDpo and creates a remote catalog handle. Naming it ExternalCatalogFactory incorrectly implies
   it covers all external catalog types. FederatedCatalogFactory makes the scope clear and is consistent with its existing implementations (HiveFederatedCatalogFactory,
  HadoopFederatedCatalogFactory).

## Checklist
- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [x] 💡 Added comments for complex logic
- [x] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #4117, opened 2026-04-03
Move CallContextCatalogFactory to core module and rename

flyrain:
Polaris currently uses external catalog for two different ideas: the broad concept of catalogs whose source of truth lives outside Polaris, and the narrower runtime path where Polaris builds a remote catalog client from ConnectionConfigInfo and delegates operations to it. That makes the current SPI name misleading, since notification-synced catalogs are also external in the broad sense but do not use this factory. Likewise, CallContextCatalogFactory describes an implementation detail rather than the role of the abstraction.

This PR makes that split explicit by renaming the remote delegation path to `FederatedCatalogFactory` and the Polaris-managed path to `LocalCatalogFactory`. That gives us clearer terminology: external stays the umbrella concept, federated refers to remote-backed execution, and local refers to the in-process Polaris-backed path. No behavioral change is intended.

This PR also moves the interface into `polaris-core`. Keeping it in core should make future SPI enforcement simpler if we treat interfaces in `polaris-core` as the supported SPI surface.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalCatalogFactory.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalCatalogFactory.java
@@ -16,12 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.polaris.service.context.catalog;
+package org.apache.polaris.core.catalog;
 
 import org.apache.iceberg.catalog.Catalog;
 import org.apache.polaris.core.persistence.resolver.PolarisResolutionManifest;
 
-public interface CallContextCatalogFactory {
+public interface LocalCatalogFactory {
 
-  Catalog createCallContextCatalog(PolarisResolutionManifest resolvedManifest);
+  Catalog createCatalog(PolarisResolutionManifest resolvedManifest);
 }
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
@@ -90,6 +90,7 @@ import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
 import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.auth.PolarisAuthorizableOperation;
 import org.apache.polaris.core.catalog.FederatedCatalogFactory;
+import org.apache.polaris.core.catalog.LocalCatalogFactory;
 import org.apache.polaris.core.config.FeatureConfiguration;
 import org.apache.polaris.core.connection.ConnectionConfigInfoDpo;
 import org.apache.polaris.core.connection.ConnectionType;
@@ -122,7 +123,6 @@ import org.apache.polaris.service.catalog.common.CatalogHandler;
 import org.apache.polaris.service.catalog.common.CatalogUtils;
 import org.apache.polaris.service.catalog.io.StorageAccessConfigProvider;
 import org.apache.polaris.service.config.ReservedProperties;
-import org.apache.polaris.service.context.catalog.CallContextCatalogFactory;
 import org.apache.polaris.service.events.EventAttributeMap;
 import org.apache.polaris.service.events.EventAttributes;
 import org.apache.polaris.service.http.IcebergHttpUtil;
@@ -193,7 +193,7 @@ public abstract class IcebergCatalogHandler extends CatalogHandler implements Au
 
   protected abstract ResolverFactory resolverFactory();
 
-  protected abstract CallContextCatalogFactory catalogFactory();
+  protected abstract LocalCatalogFactory localCatalogFactory();
 
   protected abstract ReservedProperties reservedProperties();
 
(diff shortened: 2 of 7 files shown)


polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalCatalogFactory.java
```diff
@@ -16,12 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.polaris.service.context.catalog;
+package org.apache.polaris.core.catalog;
```
dimas-b, line 19: Why move it to core? None of the core classes need to reference this factory 🤔

flyrain: Here are two reasons I moved it:
1. Be consistent with the `FederatedCatalogFactory` in https://github.com/apache/polaris/pull/4116,  `core` owns the abstraction, `service/extensions` own the implementations.
2. It isn’t because `polaris-core` needs to call it directly. It’s because the interface itself is a contract/SPI boundary, and that contract is expressed entirely in core types (`PolarisResolutionManifest`).

jbonofre: `PolarisLocalCatalogFactory` stays in `org.apache.polaris.service.context.catalog` while the interface moved to `org.apache.polaris.core.catalog`. 

This is fine (implementation in service, SPI in core), but the old package `context.catalog` now only has `PolarisLocalCatalogFactory` and `RealmContextHolder` — worth noting if future cleanup consolidates packages.

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalCatalogFactory.java
```diff
@@ -16,12 +16,12 @@
...
 
 import org.apache.iceberg.catalog.Catalog;
 import org.apache.polaris.core.persistence.resolver.PolarisResolutionManifest;
 
-public interface CallContextCatalogFactory {
+public interface LocalCatalogFactory {
```
jbonofre, line 24: nit: I was a bit confused by the PR title and the actual change - The PR title says "Move CallContextCatalogFactory to core module" but it also renames to `LocalCatalogFactory`. The title should reflect both the move and the rename.

flyrain: updated the title

jbonofre: Thanks a lot ! It's all good to me.

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalCatalogFactory.java
dimas-b: > It’s because the interface itself is a contract/SPI boundary, and that contract is expressed entirely in core types (PolarisResolutionManifest).

I agree that this interface is an SPI, however, it is not a "core SPI" in the sense that Polaris Core does not need an implementation of  this SPI to be provided in order to function property.

If the idea behind moving it to `polaris-core` is only to consolidate all SPI classes in `polaris-core`, I'd like to propose postponing the move until we make progress on the general SPI standardization discussion (which we started on `dev` a long time ago, but did not progress, unfortunately).

I imagine different components of Polaris may have different SPI surfaces. In this case `LocalCatalogFactory` is required by the `service` code, so keeping it under `runtime/service` seems more natural to me.

If we were to formalize service SPI, I'd suggest adding a dedicated `runtime/service-spi` module instead of moving to `polaris-core`. That would lead to better isolation and dependency coherence among modules. 

My view of SPI is that it should be strongly associated with the module that calls it in runtime, not with the module that defines its method parameter types.

In any case, as I noted above, I hope for some consensus on the SPI standardization discussion before moving classes across modules.

dimas-b: Furthermore, having `LocalCatalogFactory` in `polaris-core` overloads core with classes that have no meaning within the scope of the `polaris-core` module. `LocalCatalogFactory` has meaning only in `runtime/service`.

dimas-b: Re: `FederatedCatalogFactory`, I believe it also belongs with the "service" SPI, not "core" SPI.

I suppose the only reason it is defined in `polaris-core` is to allow simpler reuse in extension modules (Hadoop, Hive).

IMHO, this only emphasizes the need to accelerate the SPI discussion and cleanup module dependencies with a coherent SPI definition (modularization) strategy.

dimas-b: Yet, I do not insist... this move in current PR does not make the overall situation worse. I still maintain my points (above), but feel free to merge with the move.

dimas-b (approved):

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalCatalogFactory.java
gh-yzou: I think the current change is fine, at least it groups the two catalog related factory interface together. 
I agree with @dimas-b 's point about accelerate the SPI discussion, we should continue push the discussion forward and come up with clear guidance about where to put those interface definitions

HonahX: +1 on current change and the formal SPI discussion. I think it will be good to have a clearer guidance on how and where to put interfaces for factories and other beans.

flyrain: Totally agree that we should move forward with the SPI discussion and provide clear guidance for the interface. Thanks every one for chiming one. cc @flyingImer.

Issue #4107, opened 2026-04-01
Add integration tests for HMS federation


flyrain:
### Is your feature request related to a problem? Please describe.

Per discussion in https://github.com/apache/polaris/pull/3404#pullrequestreview-4046396866, we will need tests to catch any regression for the HMS federation feature.

adutra:
I created a PR for this: #4315.

PR #4127, opened 2026-04-06
Update hadoop dependencies from 3.4.3 to 3.5.0

MonkeyCanCode:
This PR handled the license file check failed for https://github.com/apache/polaris/pull/4110. This is due to changes introduced by https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common/3.5.0/changes

## Checklist


runtime/server/distribution/LICENSE
```diff
@@ -2021,3 +2012,69 @@ Project URL: https://github.com/aeron-io/agrona
 License: Apache License 2.0 - https://www.apache.org/licenses/LICENSE-2.0.txt
 
 --------------------------------------------------------------------------------
+
+This product bundles several Eclipse GlassFish HK2 project artifact.
+
+* Maven group:artifact IDs: org.glassfish.hk2:hk2-api
```
snazy: I'm a bit concerned about adding hk2 (dependency injection framework) to Quarkus. Is this really needed?

runtime/server/distribution/LICENSE
```diff
@@ -2021,3 +2012,69 @@ Project URL: https://github.com/aeron-io/agrona
...
+* Maven group:artifact IDs: org.glassfish.hk2.external:aopalliance-repackaged
+
+Project URL: https://github.com/eclipse-ee4j/glassfish-hk2
+License: Eclipse Public License 2.0 - https://projects.eclipse.org/license/epl-2.0
+
+--------------------------------------------------------------------------------
+
+This product bundles several Jersey project artifacts.
+
+* Maven group:artifact IDs: org.glassfish.jersey.containers:jersey-container-servlet
```
snazy: Similar here, not sure we should add another servlet container to Quarkus.

runtime/server/distribution/LICENSE
```diff
@@ -2021,3 +2012,69 @@ Project URL: https://github.com/aeron-io/agrona
...
+* Maven group:artifact IDs: com.sun.istack:istack-commons-runtime
+
+Project URL: https://github.com/eclipse-ee4j/jaxb-istack-commons
+License: BSD 3-Clause
+
+--------------------------------------------------------------------------------
+
+This product bundles Jakarta Server Pages API.
+
+* Maven group:artifact IDs: jakarta.servlet.jsp:jakarta.servlet.jsp-api
```
snazy: It's just the Jakarta JSP API, but I wonder why JSP is needed at all?

runtime/server/distribution/LICENSE
```diff
@@ -2021,3 +2012,69 @@ Project URL: https://github.com/aeron-io/agrona
...
+
+Project URL: https://projects.eclipse.org/projects/ee4j.jsp
+License: Eclipse Public License 2.0 - https://projects.eclipse.org/license/epl-2.0
+
+--------------------------------------------------------------------------------
+
+This product bundles Angus Activation Registries.
+
+* Maven group:artifact IDs: org.eclipse.angus:angus-activation
```
snazy: Hm, isn't Angus ("activation") also going towards bean instantiation?

snazy (commented):
The change is technically totally fine.

I've got some concerns however about the implications that Hadoop brings into into Quarkus, especially hk2 as another CDI runtime and Jersey as another http server.

MonkeyCanCode:
Thanks for the quick review and feedback @snazy . I haven't check if they are absolute needed but they are getting packaged in the 3.5.0 hadoop-* package. I will take a closer look later today/tomorrow.

gradle/libs.versions.toml
```diff
@@ -19,7 +19,7 @@
 
 [versions]
 checkstyle = "13.4.0"
-hadoop = "3.4.3"
+hadoop = "3.5.0"
```
flyrain, line 22: My concern for this update is lacking of test. We dont' have any test for the module using Hadoop dependency, which is HMS federation.

snazy:
To clarify my earlier comments: Yong's change is technically fine, it's a version bump and the LICENSE updates are thorough. My concern is about the transitive dependencies that Hadoop 3.5 brings, not the PR quality.

@flyrain's point about test coverage is valid though. The federation extensions are opt-in and never ship in the "default" build (the Polaris Docker image(s) for example). But "opt-in and untested" is worse then "opt-in and tested." If we have the code, CI should vet against it.

Every Hadoop version brings new transitive dependency surprises. The exclude lists in the federation build files are telling: for 3.4 we exclude Jersey 1.x, zookeeper, reload4j, shaded protobuf, Jetty, metrics-core. 3.5.0 adds hk2 (CDI runtime, conflicts with Arc) and Jersey 2.x (JAX-RS, conflicts with RESTEasy) on top. The next version will bring something else.

The opt-in/modular approach is a reasonable compromise. Users who want Hadoop/Hive federation actively pull it in, the "default" build stays clean. But I think the project should have a conversation at some point wether the long-term direction for non-REST federation should be an out-of-process adapter (lightweight Iceberg REST bridge in front of Hadoop/Hive) rather then loading these deps into the Polaris JVM.

Not blocking this PR on that, just flagging it for a future dev@ discussion.

MonkeyCanCode:
thanks for the feedback @flyrain and @snazy. Getting a bit busy with other things. Let me review those a bit later this weekend.

MonkeyCanCode:
Sorry for the wait @flyrain  and @snazy. I finally got time later this weekend to revisit and here is what I found. All of these extra deps were added due to https://github.com/apache/hadoop/pull/7019 where Hadoop is updating jersey from 1.19 to 2.x (there are several PRs around this since 2022 but the above link is the last one that went in effect).

Now we were exclude following before this PR:
```
    exclude("com.github.pjfanning", "jersey-json")
    exclude("com.sun.jersey", "jersey-core")
    exclude("com.sun.jersey", "jersey-server")
    exclude("com.sun.jersey", "jersey-servlet")
```

With the updated 2.x version, it is now the following:
```
    exclude("jakarta.servlet.jsp", "jakarta.servlet.jsp-api")
    exclude(group = "org.glassfish.jersey.core")
    exclude(group = "org.glassfish.jersey.containers")
    exclude(group = "org.glassfish.jersey.inject")
    exclude(group = "org.glassfish.hk2")
    exclude(group = "org.glassfish.hk2.external")
```
As we were excluded them before (older version), the above change (newer version) should be safe as well.

Then for the following:
```
    exclude("com.sun.istack", "istack-commons-runtime")
    exclude(group = "org.glassfish.jaxb")
```

These were added into hadoop-common via https://github.com/apache/hadoop/pull/7928 which is related to jetty (already excluded before this change) as "Jetty needs a jaxb-impl to be able to start".

One more thing I did on the updated PR is to matched `integration-tests/build.gradle.kts` to the runtime one for exclusion.

Also, based on my understanding, those are mainly for Hadoop HTTP server.

gradle/libs.versions.toml
MonkeyCanCode: Hmm. I was not aware HMS federation didn't have tests. Now looking at federation, neither Hadoop nor Hive federated catalog have test cases. Here is the tracker for this: https://github.com/apache/polaris/issues/4180

flyrain: Thanks for filing the issue. My point is that we may not update until the tests are there. Otherwise, we might introduce regression with nobody noticed

MonkeyCanCode: Understood. Let me take a stab at it later this week if there is no taker for this.

adutra:
Hi @MonkeyCanCode are you still working on this?

FYI we have many security alerts from the docker image scan coming from Hadoop 3.4.3:

https://github.com/apache/polaris/security/code-scanning?query=is:open+branch:main+tool:Trivy

Upgrading to 3.5.0 should help us reduce the number of alerts.

MonkeyCanCode:
> Hi @MonkeyCanCode are you still working on this?
> 
> 
> 
> FYI we have many security alerts from the docker image scan coming from Hadoop 3.4.3:
> 
> 
> 
> https://github.com/apache/polaris/security/code-scanning?query=is:open+branch:main+tool:Trivy
> 
> 
> 
> Upgrading to 3.5.0 should help us reduce the number of alerts.


There is one concern from @flyrain where there is no tests around HMS. Another community member is working on that. But for this PR itself, I haven't make any more changes to it. This effort is tracked via https://github.com/apache/polaris/issues/4107.

adutra:
FYI, I went ahead and created a (minimalistic) HMS federation integration test: #4315. This will hopefully give us more confidence when upgrading Hadoop.

MonkeyCanCode:
> FYI, I went ahead and created a (minimalistic) HMS federation integration test: #4315. This will hopefully give us more confidence when upgrading Hadoop.

Thanks @adutra for creating https://github.com/apache/polaris/pull/4127. As now that is merged, mind take another look @flyrain @snazy ?

flyrain (approved):

PR #4128, opened 2026-04-06
Credential Vending for Generic Table

gh-yzou:
Add server credential vending support for generic table:
1) Refactor authorizeLoadTable for reuse between genericTable and iceberg Table
2) Add storage credential for generic table createTable and loadTable

## Checklist


polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
```diff
@@ -321,6 +321,17 @@ public static void enforceFeatureEnabledOrThrow(
...
 
+  public static final FeatureConfiguration<Boolean> ENABLE_GENERIC_TABLES_CREDENTIAL_VENDING =
+      PolarisConfiguration.<Boolean>builder()
+          .key("ENABLE_GENERIC_TABLES_CREDENTIAL_VENDING")
+          .catalogConfig("polaris.config.enable.generic-tables.credential-vending")
+          .description(
+              "If set to true, credential vending is enabled for generic tables. "
+                  + "When a generic table has a base location, scoped storage credentials "
+                  + "will be returned in the load and create responses.")
+          .defaultValue(false)
```
dimas-b: nit: why `false`? Credential Vending is enabled over IRC by default 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -263,6 +268,41 @@ protected void authorizeBasicTableLikeOperationsOrThrow(
...
+   *
+   * @param tableIdentifier the table to authorize access for
+   * @param subType the entity subtype (e.g., ICEBERG_TABLE, GENERIC_TABLE)
+   * @param delegationModes the set of access delegation modes requested by the client
+   * @return the set of storage actions granted; empty if credential vending was not requested
+   */
+  protected Set<PolarisStorageActions> authorizeLoadTableLike(
+      TableIdentifier tableIdentifier,
+      PolarisEntitySubType subType,
+      EnumSet<AccessDelegationMode> delegationModes) {
```
dimas-b, line 301: It does not appear to align with #3750 🤔 WDYT?

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
flyrain: Agreed with @dimas-b! I think IRC is always on by default. We don't need a flag here.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
```diff
@@ -110,7 +127,8 @@ public LoadGenericTableResponse createGenericTable(
       String format,
       String baseLocation,
       String doc,
-      Map<String, String> properties) {
+      Map<String, String> properties,
+      EnumSet<AccessDelegationMode> delegationModes) {
     PolarisAuthorizableOperation op = PolarisAuthorizableOperation.CREATE_TABLE_DIRECT;
```
flyrain: We use different privileges per delegation mode in IcebergCatalogHandler. Should we do the same thing here? https://github.com/apache/polaris/blob/c0262b10f200caf91a2b1206b9fb51abf435bae9/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L442

When delegation mode is true, `CREATE_TABLE_DIRECT_WITH_WRITE_DELEGATION` is used.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
```diff
@@ -150,6 +188,87 @@ public LoadGenericTableResponse loadGenericTable(TableIdentifier identifier) {
...
+        CatalogUtils.findResolvedStorageEntity(
+            resolutionManifest, tableIdentifier, PolarisEntitySubType.GENERIC_TABLE);
+    if (resolvedStoragePath == null) {
+      LOGGER.debug(
+          "Unable to find storage configuration information for generic table {}", tableIdentifier);
+      return List.of();
+    }
+
+    Set<String> tableLocations = Set.of(baseLocation);
+    org.apache.polaris.core.storage.StorageAccessConfig storageAccessConfig =
```
flyrain, line 224: nit: -> `var`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
```diff
@@ -150,6 +188,87 @@ public LoadGenericTableResponse loadGenericTable(TableIdentifier identifier) {
...
+      Boolean skipCredIndirection =
+          realmConfig().getConfig(FeatureConfiguration.SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION);
+      Preconditions.checkArgument(
+          !storageAccessConfig.supportsCredentialVending() || skipCredIndirection,
+          "Credential vending was requested for generic table %s, but no credentials are available",
+          tableIdentifier);
+      return List.of();
+    }
+
+    Map<String, String> allConfig = new java.util.HashMap<>(credentials);
```
flyrain: nit: it just feel weird to me that we are using the full path, `java.util.HashMap`. Not a blocker.

flyrain (commented):
Looks like we didn't validate generic table base location with the storage config at the creation time. I think we should do that in  GenericTableCatalogHandler.java, something like this: 

```
 public LoadGenericTableResponse createGenericTable(
      TableIdentifier identifier,
      String format,
      String baseLocation,
      String doc,
      Map<String, String> properties,
      EnumSet<AccessDelegationMode> delegationModes) {
    PolarisAuthorizableOperation op = PolarisAuthorizableOperation.CREATE_TABLE_DIRECT;
    authorizeCreateTableLikeUnderNamespaceOperationOrThrow(op, identifier);
    validateBaseLocation(identifier, baseLocation);
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
```diff
@@ -150,6 +188,87 @@ public LoadGenericTableResponse loadGenericTable(TableIdentifier identifier) {
...
+    allConfig.putAll(storageAccessConfig.extraProperties());
+
+    // Remap Iceberg's "client.region" to the generic table API's "s3.client.region"
+    String clientRegionKey = StorageAccessProperty.CLIENT_REGION.getPropertyName();
+    String region = allConfig.remove(clientRegionKey);
+    if (region != null) {
+      allConfig.put("s3.client.region", region);
+    }
```
flyrain: Can we extract this into a method that converts `org.apache.polaris.core.storage.StorageAccessConfig` to the Generic Table `StorageAccessConfig`, either in the core class or a utility class?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
```diff
@@ -91,6 +92,26 @@ public class StorageUtil {
...
   }
 
+  private static final String GENERIC_TABLE_S3_REGION_KEY = "s3.client.region";
+
+  /**
+   * Converts a core {@link StorageAccessConfig} into a flat config map suitable for the Generic
+   * Table API. Merges credentials and extra properties, and remaps Iceberg's {@code client.region}
+   * to {@code s3.client.region}.
+   */
+  public static Map<String, String> toGenericTableConfig(StorageAccessConfig storageAccessConfig) {
```
dimas-b, line 102: Does this have to be in `polaris-core`? Core classes do not use this method and the javadoc indicates it's only relevant to the Generic Tables API (while core is supposed (conceptually) to be neutral WRT all REST APIs) 🤔

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
```diff
@@ -37,10 +40,33 @@
...
+   * @return the parsed set of delegation modes, empty if the header was {@code null} or blank
+   * @throws IllegalArgumentException if an unsupported delegation mode is requested
+   */
+  public static EnumSet<AccessDelegationMode> parseAccessDelegationModes(
+      String accessDelegationMode) {
+    EnumSet<AccessDelegationMode> delegationModes =
+        AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
+    Preconditions.checkArgument(
+        delegationModes.isEmpty() || delegationModes.contains(VENDED_CREDENTIALS),
+        "Unsupported access delegation mode: %s",
```
dimas-b, line 65: This does not appear to align with #3750 🤔 WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
gh-yzou: think you’re referring to the change we made before, where we refactored the code to pass delegationModesRequested instead of the entire delegationModes.

Previously, this logic was only used by IcebergTable, so it made sense to handle the decision at the caller, since there is only one caller. Now that I’ve pushed it down to support reuse across both GenericTable and IcebergTable, it might be more consistent to also centralize the decision logic for determining delegationModes.

Since the decision itself is fairly straightforward at this moment, I’m also fine keeping it at the caller level. Let me know which approach you’d prefer.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
gh-yzou: I prefer to have an explicit type to be clear

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
gh-yzou: good point, yeas, we should, i pushed down the whole authorizeCreateTableDirect for reuse

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
gh-yzou: updated

polaris-core/src/main/java/org/apache/polaris/core/config/FeatureConfiguration.java
gh-yzou: remove the flag sounds good to me

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
gh-yzou: sorry, forgot to push the reply, it is replied in the above comment https://github.com/apache/polaris/pull/4128#discussion_r3089716717

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
gh-yzou: @flyrain maybe we can work on extracting the functionality into utils later when we have other components also requires such capability?

flyrain: I think it's OK to gather storage related method in StorageUtil. For example, this method in `StorageUtil` is only used by Iceberg tables. I don't have strong option though. If people feel strongly about it, I'm fine to move it to `GenericTableCatalogHandler`, but in that case, we might also move iceberg table related methods (4 in total) to somewhere else. WDYT?

```
  public static @Nonnull Set<String> getLocationsUsedByTable(
      String baseLocation, Map<String, String> properties) {
    Set<String> locations = new HashSet<>();
    locations.add(baseLocation);
    locations.add(properties.get(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_DATA_LOCATION_KEY));
    locations.add(
        properties.get(IcebergTableLikeEntity.USER_SPECIFIED_WRITE_METADATA_LOCATION_KEY));
    locations.remove(null);
    return removeRedundantLocations(locations);
  }
```

gh-yzou: or maybe i can put it under CatalogHandlerUtils for now

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
dimas-b: My interpretation of #3750 is that it changes all processing to use _one_ (resolved) `AccessDelegationMode` (may be optional). Here we have a set.

I believe it would be more coherent to use a `boolean delegationRequested` flag here as in [authorizeLoadTable](https://github.com/apache/polaris/blob/c29e01fbd9b68294f906e00569178c9374e44fe0/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L819-L820).

WDYT?

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
flyrain: hi @dimas-b , what's your take on this?

dimas-b: `CatalogHandlerUtils` SGTM :+1: (or any other utility class under `runtime/service`).

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
gh-yzou: That sounds good to me, i updated it to take delegationRequested instead, we can always extract a common utility later if the logic to decide delegationRequested becomes complicated later

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageUtil.java
gh-yzou: I think it would be good to have our core to stay more concise, and i agree the other table location related utils probably should also go into runtime/service.
 
I moved it to CatalogHandlerUtils for this part. @flyrain i will get a different PR to push the tableLocation related functions to  CatalogHandlerUtils or CatalogUtils.

flyrain: SGTM. Thanks @gh-yzou !

gh-yzou:
@flyrain Agree that we should stay consistent on the catalog semantic, I moved the validationLocationForTableLike logic to CatalogUtils and did the check at GenericTableCatalog level similar as IcebergCatalog.

gh-yzou:
Realize that the location change might require little more change, checked with @flyrain  will put it in a separate PR for better review

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -215,6 +222,20 @@ protected void authorizeCreateTableLikeUnderNamespaceOperationOrThrow(
     initializeCatalog();
   }
 
+  protected void authorizeCreateTableDirect(
+      TableIdentifier identifier, EnumSet<AccessDelegationMode> delegationModes) {
```
dimas-b: Sorry, @gh-yzou , it looks like this still goes against the new delegation mode approach in #3750.

I believe the principle set forth in that PR is to use either one (optional) `AccessDelegationMode` parameter or just a "delegation requested" flags throughout the service code.

Only delegation mode resolvers should deal with a "set" of `AccessDelegationMode`.

dimas-b: Cf. https://github.com/apache/polaris/pull/4128#discussion_r3095142055

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
```diff
@@ -32,10 +36,33 @@
...
+   * <p>This is shared by both Iceberg and Generic Table catalog adapters.
+   *
+   * @param accessDelegationMode the raw header value (comma-separated), may be {@code null}
+   * @return the parsed set of delegation modes, empty if the header was {@code null} or blank
+   * @throws IllegalArgumentException if an unsupported delegation mode is requested
+   */
+  public static EnumSet<AccessDelegationMode> parseAccessDelegationModes(
+      String accessDelegationMode) {
+    EnumSet<AccessDelegationMode> delegationModes =
+        AccessDelegationMode.fromProtocolValuesList(accessDelegationMode);
```
dimas-b, line 55: Why not reuse `AccessDelegationModeResolver` from #3750 ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
gh-yzou: sorry, updated the loadTable check, but forgot the udpateTablecheck.
Today we first do parse to set of AccessDelegationMode, then do authentication check, then do resolve to single AccessDelegationMode.

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
gh-yzou: this part is actually a pure move of utility function from the adapter class https://github.com/apache/polaris/pull/4128/changes#diff-1d906402f6374c004e1ceaafea685c5b65bbba306afe83f15f73c8da2762ee70L251
This function is used to parse the string input to set of AccessDelegationMode.

The AccessDelegationModeResolver is called after authentication. Agree that we should also do a resolve after authentication. I pushed down the whole resolver and did the same thing for GenericTable

gh-yzou:
@dimas-b could you take one more look of the PR ?

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogAdapter.java
```diff
@@ -74,6 +83,8 @@ public Response createGenericTable(
       String polarisGenericTableAccessDelegation,
       RealmContext realmContext,
       SecurityContext securityContext) {
+    EnumSet<AccessDelegationMode> delegationModes =
+        parseAccessDelegationModes(polarisGenericTableAccessDelegation);
```
dimas-b, line 87: > Medium: Invalid Polaris-Generic-Table-Access-Delegation values are silently ignored instead of failing the request. Both generic-table entrypoints parse
     the header in runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogAdapter.java:86, but the shared parser in
     runtime/service/src/main/java/org/apache/polaris/service/catalog/AccessDelegationMode.java:64 drops unknown tokens rather than rejecting them. The generic-
     table spec says that if the header is present, the server must either provide the requested mechanism or fail; today a typo like vendd-credentials degrades
     into a normal 200 response with no storage-access-configs, which violates that contract and makes client misconfiguration hard to detect.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
```diff
@@ -150,6 +189,63 @@ public LoadGenericTableResponse loadGenericTable(TableIdentifier identifier) {
...
+    String baseLocation = entity.getBaseLocation();
+    if (baseLocation == null || baseLocation.isEmpty()) {
+      LOGGER.debug(
+          "No base location set for generic table {}, skipping credential vending",
+          tableIdentifier);
+      return List.of();
+    }
+
+    PolarisResolvedPathWrapper resolvedStoragePath =
+        CatalogUtils.findResolvedStorageEntity(
```
dimas-b, line 222: > High: Federated generic-table credential vending skips the remote-location safety check that Iceberg still enforces. In runtime/service/src/main/java/org/
     apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java:221, the new path resolves storage and immediately calls
     storageAccessConfigProvider().getStorageAccessConfig(...) for baseLocation, but there is no equivalent of Iceberg’s federated
     validateRemoteTableLocations(...) gate in runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java:878. On an
     external/federated generic catalog with credential vending enabled, Polaris can now mint creds for whatever location the remote catalog reports, even if
     that location is outside the catalog’s allowed storage hierarchy.

dimas-b (commented):
passing feedback from artificial helpers:

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
dimas-b: I see... still this function looks like a trivial redirect. Is there much value in it as opposed to calling `AccessDelegationMode.fromProtocolValuesList()`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
```diff
@@ -215,6 +226,42 @@ protected void authorizeCreateTableLikeUnderNamespaceOperationOrThrow(
     initializeCatalog();
   }
 
+  protected void authorizeCreateTableDirect(
+      TableIdentifier identifier, boolean delegationRequested) {
+    PolarisAuthorizableOperation op =
+        !delegationRequested
```
dimas-b, line 232: for a ternary statement with both branches present, using a simple (non-negated) condition would be easier to read.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
```diff
@@ -150,6 +189,63 @@ public LoadGenericTableResponse loadGenericTable(TableIdentifier identifier) {
...
+      Preconditions.checkArgument(
+          !storageAccessConfig.supportsCredentialVending() || skipCredIndirection,
+          "Credential vending was requested for generic table %s, but no credentials are available",
+          tableIdentifier);
+      return List.of();
+    }
+
+    Map<String, String> config = CatalogHandlerUtils.toGenericTableConfig(storageAccessConfig);
+
+    return List.of(StorageAccessConfig.builder().setPrefix(baseLocation).setConfig(config).build());
```
dimas-b, line 249: Why return a `List` when it always has at most one element? Why not `Optional`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
gh-yzou: i don't think anything blocks us to call fromProtocolValuesList directly. Given that i actually don't call this function anymore, i reverted this part of change for now, and i can get a separate pr to handle this small refactoring change

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogAdapter.java
gh-yzou: This is consistent with the behavior with the current iceberg catalog, it seems the previous change https://github.com/apache/polaris/pull/3750/changes#diff-1d906402f6374c004e1ceaafea685c5b65bbba306afe83f15f73c8da2762ee70L251 actually changed this behavior, originally, we will throw an error.

For GenericTable, i think we can be more restricted, i add a check to error out any invalid configuration

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
gh-yzou: this is a valid point, and has been brought up by @flyrain also https://github.com/apache/polaris/pull/4128#pullrequestreview-4110187347, i plan to do the location check in a separate pr since it actually requires quite bunch of changes

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogHandler.java
gh-yzou: updated

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
gh-yzou: the spec today returns a list for extension to handle multiple locations, but for this function, I think we can simply return one value and do the construction at caller, updated

runtime/service/src/main/java/org/apache/polaris/service/catalog/common/CatalogUtils.java
gh-yzou: Yes, it is a trival redirect, and we can remove the redirection. I reverted this change from this PR since i got a separate function for generatic table to do more restricted check. i can get a different pr to handle this simple refactoring

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
gh-yzou: I am doing it in the followup PR here https://github.com/apache/polaris/pull/4237

gh-yzou:
@dimas-b please help take one more look of the PR, Thanks!

dimas-b:
I need some more time to review, please. Let's not rush it :)

gh-yzou:
@dimas-b 
Hi Dmitri,
Thanks! When you have some time, could you please take a look at the PR? If it’s easier, would you be fine doing a post-review after others have reviewed and approved it?
We won’t be marking the feature as ready for use yet, and I’m happy to address any additional concerns in follow-up PRs. I’m aiming to get this resolved soon so I can move forward with other follow-ups, including the location checks.
Appreciate your time—thanks in advance!

dimas-b:
Reviewing :hourglass: @gh-yzou : I appreciate your patience. Let's not rush.

runtime/service/src/main/java/org/apache/polaris/service/catalog/generic/GenericTableCatalogHandler.java
dimas-b: I do not think deferring this is a sound approach.

dimas-b:
I guess #4237 needs to merge first.

snazy:
Heads up: I'm going to delete the branch soon, see [dev-mailing-list discussion](https://lists.apache.org/thread/0pqj4mhpthn0whdzxg84zwh1gb0fp6gf)

dimas-b:
@gh-yzou : Please feel free to re-submit this change from a branch in your fork.

PR #4136, opened 2026-04-08
Refactor LocalCatalogFactory to return IcebergCatalog

dimas-b:
Following up on #4091

* Move LocalCatalogFactory from polaris-core to runtime/service since all consumers are in runtime/service, and change createCatalog() return type from Catalog to IcebergCatalog (all implementations return that type already).

* In IcebergCatalogHandler, replace the single baseCatalog field and isFederated boolean with separate federatedCatalog (Catalog) and localCatalog (IcebergCatalog) fields, eliminating all (IcebergCatalog) baseCatalog casts.

* Update tests to property override FederatedCatalogFactory

## Checklist


runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
```diff
@@ -624,38 +649,44 @@ void testLoadCredentialsFromEntityPropertiesSkipsFileIO() throws Exception {
...
-    Field baseCatalogField = IcebergCatalogHandler.class.getDeclaredField("baseCatalog");
-    baseCatalogField.setAccessible(true);
-    Catalog spiedCatalog = (Catalog) baseCatalogField.get(handler);
-    Mockito.verify(spiedCatalog, Mockito.never()).loadTable(Mockito.any());
+    Assertions.assertThat(spiedCatalog).doesNotHaveNullValue();
+    Mockito.verify(spiedCatalog.get(), Mockito.never()).loadTable(any());
   }
 
   @Test
   void testLoadCredentialsFromEntityPropertiesFallsBackForExternalCatalog() {
```
dimas-b, line 670: TBH, I'm not sure this test fits well with the other tests in this class... WDYT about moving it to a separate test class?

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
```diff
@@ -624,38 +649,44 @@ void testLoadCredentialsFromEntityPropertiesSkipsFileIO() throws Exception {
...
-    assertSuccess(
-        adminService.grantPrivilegeOnCatalogToRole(
```
dimas-b, line 645: note to reviewers: this was not necessary

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/LocalCatalogFactory.java
```diff
@@ -16,12 +16,11 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.polaris.core.catalog;
+package org.apache.polaris.service.catalog.iceberg;
```
flyrain, line 19: Thanks for picking this up, @dimas-b ! 

By moving `LocalCatalogFactory` from `polaris-core` to the runtime module and changing its return type to the concrete `IcebergCatalog`, the PR currently couples the factory interface to a specific implementation.

Instead of moving it here, we could create a new interface/abstract class in the core module for the existing class `IcebergCatalog`. This would allow downstream consumers to reuse the abstraction without being tied to a specific implementation.

An abstract class option:
```
public abstract class PolarisIcebergCatalog extends BaseMetastoreViewCatalog
    implements SupportsNamespaces, SupportsNotifications, Closeable {
...
}
```

An interface option:
```
public interface PolarisIcebergCatalog extends Catalog, ViewCatalog,  SupportsNamespaces, SupportsNotifications, Closeable {
// we could add necessary methods here
...
}
```

I personally like the interface option a bit more. This keeps the core library clean and provides a reusable contract for any customized implementations. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/config/ResolvingFederatedCatalogFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import jakarta.inject.Inject;
+import java.util.Map;
+import org.apache.iceberg.catalog.Catalog;
+import org.apache.polaris.core.catalog.FederatedCatalogFactory;
+import org.apache.polaris.core.catalog.GenericTableCatalog;
+import org.apache.polaris.core.connection.ConnectionConfigInfoDpo;
+import org.apache.polaris.core.connection.ConnectionType;
+import org.apache.polaris.core.credentials.PolarisCredentialManager;
+
+@ApplicationScoped
```
flyrain, line 35: Introducing a new application scoped bean may not be necessary. I think a new method in `IcebergCatalogHandler` to wrap the related logic would be good enough. WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/LocalCatalogFactory.java
dimas-b: Sure. I personally still think this factory belongs in `runtime/service` (rather than `polaris-core`) so the type is not a concern.

However, the interface option also looks reasonable to me. I'll follow that suggestion as I believe you prefer to keep this factory in `core`.

runtime/service/src/main/java/org/apache/polaris/service/config/ResolvingFederatedCatalogFactory.java
dimas-b: It is not strictly necessary, but it offers better separation of concerns. Conceptually, `IcebergCatalogHandler` needs a factory implementation, but it does not need to deal with CDI or with how specific implementations are defined.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/LocalCatalogFactory.java
flyrain: Thanks @dimas-b! Yeah, I'd rather keep it in `core` now instead of introducing a new module like `service-api`. I think it would be simpler and also manageable this way. We could refactor later if needed.

runtime/service/src/main/java/org/apache/polaris/service/config/ResolvingFederatedCatalogFactory.java
flyrain: > Conceptually, IcebergCatalogHandler needs a factory implementation, but it does not need to deal with CDI or with how specific implementations are defined.

+1. I think the `ResolvingFederatedCatalogFactory` is a good idea. Thanks for the explanation.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/LocalCatalogFactory.java
dimas-b: Converted this PR to draft for now - working on this :hourglass_flowing_sand:

runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
dimas-b: Cf. #4540

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/LocalCatalogFactory.java
dimas-b: moved LocalCatalogFactory back to polaris-core

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalIcebergCatalog.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.SupportsNamespaces;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.catalog.ViewCatalog;
+import org.apache.polaris.core.persistence.PolarisMetaStoreManager;
+import org.apache.polaris.core.persistence.pagination.Page;
+import org.apache.polaris.core.persistence.pagination.PageToken;
+
+/** Base interface for local Polaris Catalogs dealing with Iceberg tables. */
+public interface LocalIcebergCatalog extends Catalog, ViewCatalog, SupportsNamespaces {
```
adutra, line 34: Why not extend `SupportsNotifications`?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerFactory.java
```diff
@@ -54,7 +53,7 @@ public class IcebergCatalogHandlerFactory {
   @Inject PolarisAuthorizer authorizer;
   @Inject ReservedProperties reservedProperties;
   @Inject CatalogHandlerUtils catalogHandlerUtils;
-  @Inject @Any Instance<FederatedCatalogFactory> federatedCatalogFactories;
+  @Inject @Any ResolvingFederatedCatalogFactory federatedCatalogFactory;
```
adutra: You probably don't need the `@Any` annotation anymore.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -152,7 +152,7 @@
 
 /** Defines the relationship between PolarisEntities and Iceberg's business logic. */
 public class IcebergCatalog extends BaseMetastoreViewCatalog
```
adutra, line 154: Since we are here, could we consider renaming `IcebergCatalog` to `DefaultLocalIcebergCatalog` or `PolarisLocalIcebergCatalog`? I always found this name confusing because it is not immediately clear that this is a local catalog (as opposed to a federated one), and "Iceberg" tends to indicate that it's not a Polaris-specific component, when in fact it is.

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalIcebergCatalog.java
dimas-b: `SupportsNotifications` is in `runtime/service` :shrug: 

I personally believe my original proposal of moving `LocalCatalogFactory` to `runtime/service` produced better code alignment between `polaris-core` and `runtime/server`... However, @flyrain expressed a different opinion: https://github.com/apache/polaris/pull/4136#discussion_r3066692650

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dimas-b: Sure. I'll do that in a separate PR after this one is resolved :+1:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerFactory.java
dimas-b: Good point - thx!

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalIcebergCatalog.java
adutra: Yeah, I'm not convinced by the idea of having this interface in polaris-core either.

dimas-b:
rebased and fixed conflicts (first commit only)

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalIcebergCatalog.java
dimas-b: @flyrain : What do you think is the best path forward here?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandlerFactory.java
dimas-b: removed

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
dimas-b: #4672

polaris-core/src/main/java/org/apache/polaris/core/catalog/LocalIcebergCatalog.java
flyrain: option 1. Can we position it as a maker interface, so that it doesn't extend any other interfaces?
option 2. Move `SupportsNotifications` to `core`, I do think moving most interfaces to core is a good practice now, given the current convention we have. For example, these 3 related interfaces are in the core: `FederatedCatalogFactory`, `GenericTableCatalog`, and `LocalCatalogFactory`. And a bunch of other interfaces like `PolarisAuthorizer` live in core as well. So that the Polaris core serving as a SPI module for different implementations.

I don't understand why this interface cannot be in the core. If we do have reasons, we could create a new module like service api or service core in the future to move certain interfaces out of `polaris-core`. I don't think it's necessary but I'm open to discussions.

dimas-b: From my POV I do not see a reason for moving classes into `polaris-core` that have not been in core initially.

Core should only contain classes that are used in core.

dimas-b: Marker interfaces are not an ideal pattern here, from my POV, either. This refactoring was originally meant to increase modularity and clarity. Marker interfaces will require a type check later on, which actually reduces clarity.

PR #4146, opened 2026-04-08
feat(labels): Add labels for polaris entities

vignesh-manel:
Add key-value label support to Catalog entities for organizing and filtering catalogs via the management REST API.

- Catalogs can now have optional key-value labels (e.g. env:prod, team:platform) set at creation or updated later
- Labels are returned in catalog responses and omitted entirely when not set
- UpdateCatalogRequest supports labels (replace) and clearLabels (remove all); the two are mutually exclusive
- When labels is omitted on update, existing labels are preserved (non-destructive by default)
- Validation enforces: key ≤ 128 chars, value ≤ 256 chars, max 20 labels per catalog

## Checklist

### DIFF
--- a/client/python/apache_polaris/cli/command/catalogs.py
+++ b/client/python/apache_polaris/cli/command/catalogs.py
@@ -31,6 +31,7 @@ from apache_polaris.cli.constants import (
 from apache_polaris.cli.options.option_tree import Argument
 
 from dataclasses import dataclass
+import click
 from pydantic import StrictStr, SecretStr
 from typing import Dict, List, Optional, Union, Tuple, Callable
 
@@ -116,6 +117,9 @@ class CatalogsCommand(Command):
     catalog_external_id: str
     catalog_signing_region: str
     catalog_signing_name: str
+    label: List[str]
+    label_filter: List[str]
+    clear_labels: bool = False
 
     def validate(self) -> None:
         if self.catalogs_subcommand == Subcommands.CREATE:
@@ -379,10 +383,41 @@ class CatalogsCommand(Command):
             )
         return config
 
+    def _parse_labels(self, label_list: Optional[List[str]]) -> Optional[Dict[str, StrictStr]]:
+        """Parse a list of 'key=value' label strings into a dict."""
+        if not label_list:
+            return None
+        result = {}
+        for entry in label_list:
+            idx = entry.find("=")
+            if idx <= 0:
+                raise click.UsageError(
+                    f"Invalid label format '{entry}': must be 'key=value'"
+                )
+            result[entry[:idx].strip()] = entry[idx + 1 :]
+        return result
+
+    def _resolve_labels_for_update(
+        self,
+    ) -> tuple[Optional[Dict[str, StrictStr]], Optional[bool]]:
+        """Resolve the labels and clearLabels fields for an update request.
+
+        Returns (labels, clear_labels) where:
+          - clear_labels=True → clear all labels (mutually exclusive with labels)
+          - labels is a non-empty dict → replace labels
+          - both None → preserve existing labels (fields omitted from request)
+        """
+        if self.clear_labels and self.label:
+            raise click.UsageError("--clear-labels and --label are mutually exclusive")
+        if self.clear_labels:
+            return None, True
+        return self._parse_labels(self.label), None
+
     def execute(self, api: PolarisDefaultApi) -> None:
         if self.catalogs_subcommand == Subcommands.CREATE:
             storage_config = self._build_storage_config_info()
             connection_config = self._build_connection_config_info()
+            labels = self._parse_labels(self.label)
             if self.catalog_type == CatalogType.EXTERNAL.value:
                 request = CreateCatalogRequest(
                     catalog=ExternalCatalog(
@@ -394,6 +429,7 @@ class CatalogsCommand(Command):
                             additional_properties=self.properties,
                         ),
                         connection_config_info=connection_config,
+                        labels=labels,
                     )
                 )
             else:
@@ -407,6 +443,7 @@ class CatalogsCommand(Command):
                             additional_properties=self.properties,
                         ),
                         connection_config_info=connection_config,
+                        labels=labels,
                     )
                 )
             api.create_catalog(request)
@@ -415,7 +452,8 @@ class CatalogsCommand(Command):
         elif self.catalogs_subcommand == Subcommands.GET:
             print(api.get_catalog(self.catalog_name).to_json())
         elif self.catalogs_subcommand == Subcommands.LIST:
-            for catalog in api.list_catalogs().catalogs:
+            label_filter = self.label_filter or []
+            for catalog in api.list_catalogs(label_filter=label_filter).catalogs:
                 print(catalog.to_json())
         elif self.catalogs_subcommand == Subcommands.UPDATE:
             catalog = api.get_catalog(self.catalog_name)
@@ -479,15 +517,21 @@ class CatalogsCommand(Command):
                         )
                     updated_storage_info.region = self.region
 
+                labels, clear_labels = self._resolve_labels_for_update()
                 request = UpdateCatalogRequest(
                     current_entity_version=catalog.entity_version,
                     properties=catalog.properties.to_dict(),
                     storage_config_info=updated_storage_info,
+                    labels=labels,
+                    clear_labels=clear_labels,
                 )
             else:
+                labels, clear_labels = self._resolve_labels_for_update()
                 request = UpdateCatalogRequest(
                     current_entity_version=catalog.entity_version,
                     properties=catalog.properties.to_dict(),
+                    labels=labels,
+                    clear_labels=clear_labels,
                 )
 
             api.update_catalog(self.catalog_name, request)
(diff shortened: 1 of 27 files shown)


snazy:
Hi @vignesh-manel !

Thanks a lot for this! I took a brief look over the PR and I do not see anything that's technically worrisome. Sure, there are a few things, but those can be sorted out w/o a lot of hassle.

As this is a new feature, I just wonder whether it's better to go on the dev-mailing-list and start a discussion about this new feature to gather the community's opinion and eventually consensus. WDYT?

Robert

vignesh-manel:
> Hi @vignesh-manel !
> 
> Thanks a lot for this! I took a brief look over the PR and I do not see anything that's technically worrisome. Sure, there are a few things, but those can be sorted out w/o a lot of hassle.
> 
> As this is a new feature, I just wonder whether it's better to go on the dev-mailing-list and start a discussion about this new feature to gather the community's opinion and eventually consensus. WDYT?
> 
> Robert

@snazy Thanks for taking a look, I have posted it in dev mail thread: https://lists.apache.org/thread/2ny3gx01dx85b3bzo58zkbddryrjhfr3

client/python/apache_polaris/cli/command/catalogs.py
dimas-b, line 1: @vignesh-manel : If I may suggest moving CLI changes to a separate PR, I think it would simplify review (just different audiences usually review core and CLI changes).

spec/polaris-management-service.yml
```diff
@@ -878,6 +889,12 @@ components:
           description: The version of the catalog object used to determine if the catalog metadata has changed
         storageConfigInfo:
           $ref: "#/components/schemas/StorageConfigInfo"
+        labels:
+          type: object
+          x-extra-annotation: "@com.fasterxml.jackson.annotation.JsonInclude(com.fasterxml.jackson.annotation.JsonInclude.Include.NON_EMPTY)"
```
dimas-b: I believe it is preferable to keep the OpenAPI definition language-neutral.

I believe `null` values are already not included. Could we represent empty labels as `null` here?

Alternatively, it might be a good idea to make a separate change to avoid serializing empty values across all endpoints (cf. #1955, https://lists.apache.org/thread/nfboq6yo01x3lw4t3d5gp78f6lgs5ddz).

WDYT?

dimas-b (commented):
The feature LGTM at high level 👍 ... posting a couple of preliminary comments.

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
```diff
@@ -2484,6 +2502,272 @@ public void testCreateAndUpdatePrincipalWithReservedProperties() {
...
+
+    managementApi.createCatalog(catalog);
+
+    Catalog fetched = managementApi.getCatalog(catalogName);
+    assertThat(fetched.getLabels()).containsEntry("env", "staging");
+
+    // Update with labels omitted (null) — existing labels must be preserved
+    try (Response response =
+        managementApi
+            .request("v1/catalogs/{name}", Map.of("name", catalogName))
```
dimas-b: nit: maybe add a helper method to `ManagementApi`?

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
```diff
@@ -2484,6 +2502,272 @@ public void testCreateAndUpdatePrincipalWithReservedProperties() {
...
+            .setType(Catalog.TypeEnum.INTERNAL)
+            .setName(devCatalogName)
+            .setProperties(new CatalogProperties("s3://filter-bucket/dev"))
+            .setStorageConfigInfo(storageConfig)
+            .setLabels(Map.of("env", "dev"))
+            .build());
+
+    // Filter by env=prod — only prodCatalog should match
+    try (Response response =
+        managementApi.request("v1/catalogs", Map.of(), Map.of("labelFilter", "env=prod")).get()) {
```
dimas-b: nit: new helper method in `ManagementApi`?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
```diff
@@ -31,6 +33,9 @@ public class PolarisBaseEntity extends PolarisEntityCore {
 
   public static final String EMPTY_MAP_STRING = "{}";
 
+  /** Internal-property key under which entity labels are stored as a serialized JSON map. */
+  public static final String LABELS_INTERNAL_KEY = "polaris.entity.labels";
```
dimas-b: `protected` seems sufficient... Let's not expand the public API surface in "core" unless really necessary.

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
```diff
@@ -120,6 +125,33 @@ public Map<String, String> getInternalPropertiesAsMap() {
     return PolarisObjectMapperUtil.deserializeProperties(getInternalProperties());
   }
 
+  /**
+   * Returns the labels attached to this entity. Labels are stored in internal properties under
+   * {@link #LABELS_INTERNAL_KEY} and are never empty — an absent key means no labels.
+   */
+  @JsonIgnore
+  public Map<String, String> getLabels() {
+    String raw = getInternalPropertiesAsMap().get(LABELS_INTERNAL_KEY);
```
dimas-b, line 134: WDYT about modelling each label as a separate internal property? e.g. `polaris.entity.label.env=dev`?

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
```diff
@@ -120,6 +125,33 @@ public Map<String, String> getInternalPropertiesAsMap() {
...
+    }
+    return PolarisObjectMapperUtil.deserializeProperties(raw);
+  }
+
+  /**
+   * Returns {@code true} if this entity matches all entries in {@code labelFilter}. An empty filter
+   * matches every entity.
+   */
+  @JsonIgnore
+  public boolean matchesLabelFilter(@Nonnull Map<String, String> labelFilter) {
```
dimas-b: nit: how about `containsAllLabels`? I think it would represent actual behaviour better.

spec/polaris-management-service.yml
```diff
@@ -41,6 +41,17 @@ paths:
     get:
       operationId: listCatalogs
       description: List all catalogs in this polaris service
+      parameters:
+        - name: labelFilter
```
dimas-b: suggestion: rename to `label` and use `:` as the separator. A sample query could be `curl -X GET http://.../catalogs?label=env:prod&label=team:eng`... It reads nicer from my POV.

runtime/service/src/main/java/org/apache/polaris/service/admin/PolarisServiceImpl.java
```diff
@@ -254,13 +257,41 @@ public Response updateCatalog(
...
+   * Parses a list of {@code "key=value"} label-filter strings into a map. Throws {@link
+   * BadRequestException} for malformed entries.
+   */
+  private static Map<String, String> parseLabelFilter(List<String> labelFilter) {
+    if (labelFilter == null || labelFilter.isEmpty()) {
+      return Collections.emptyMap();
+    }
+    Map<String, String> result = new HashMap<>(labelFilter.size());
+    for (String entry : labelFilter) {
+      int idx = entry.indexOf('=');
```
dimas-b: We never actually process a label's key/value parts individually, right? Do we need to parse it at all?

client/python/apache_polaris/cli/command/catalogs.py
vignesh-manel: Thanks for the review. 
Removed the CLI changes from this PR, will create a different PR for CLI, once this is merged.

spec/polaris-management-service.yml
vignesh-manel: removed this, now the response will have labels as empty map {} , similar to other collection type attributes in the APIs

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
vignesh-manel: added helper method `updateCatalogLabels`

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisManagementServiceIntegrationTest.java
vignesh-manel: added helper method `listCatalogsByLabel`

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
vignesh-manel: updated to `protected`

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisBaseEntity.java
vignesh-manel: renamed to `containsAllLabels`

spec/polaris-management-service.yml
vignesh-manel: thanks for this suggestion, this is better way, updated the code accordingly

flyrain:
Hi @vignesh-manel, thanks for the PR. Left a few questions in the dev mailing list. I think we need to think through the solution before jumping to an implementation.

spec/polaris-management-service.yml
dimas-b: Empty maps in REST responses are probably acceptable.

I wonder if we could tweak OpenAPI code generation to exclude empty values too. For example, if we add a more general purpose flag to this property in the YAML file and later use it to produce `JsonInclude` annotations on the generated code... WDYT?

PR #4159, opened 2026-04-09
feat(authz): support per-realm authorizer selection

### LINKED ISSUE #4112: Introduce Per Realm configuration to `PolarisAuthorizer`

Today, the `PolarisAuthorizer` doesn't support realm configs. We can introduce realm specific configuration overrides, similar to the Authentication config:

https://github.com/apache/polaris/blob/main/runtime/service/src/main/java/org/apache/polaris/service/auth/AuthenticationConfiguration.java

_No response_

_No response_

### Additional context

_No response_

officialasishkumar:
Allow authorization type selection to follow the active realm while keeping `polaris.authorization.type` as the default fallback. This mirrors the authentication configuration structure so different realms can select different authorizer implementations without breaking existing configs.

This change adds a realm-aware authorization config mapping, updates request-scoped authorizer creation to select the configured factory for the current realm, and adds focused tests for both config resolution and producer behavior.

## Testing
- `export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64 && ./gradlew format :polaris-runtime-service:test --tests 'org.apache.polaris.service.config.AuthorizationConfigurationTest' :polaris-runtime-service:intTest --tests 'org.apache.polaris.service.it.ServiceProducersIT$InternalAuthorizationTest' --tests 'org.apache.polaris.service.it.ServiceProducersIT$PerRealmAuthorizationTest'`\n\n## Checklist\n- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)\n- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #4112\n- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)\n- [ ] 💡 Added comments for complex logic\n- [ ] 🧾 Updated `CHANGELOG.md` (if needed)\n- [ ] 📚 Updated `site/content/in-dev/unreleased` documentation (if needed)


runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -152,14 +152,27 @@ public PolarisAuthorizerFactory polarisAuthorizerFactory(
       AuthorizationConfiguration authorizationConfig,
       @Any Instance<PolarisAuthorizerFactory> authorizerFactories) {
     PolarisAuthorizerFactory factory =
-        authorizerFactories.select(Identifier.Literal.of(authorizationConfig.type())).get();
+        authorizerFactories
+            .select(
+                Identifier.Literal.of(
+                    authorizationConfig
+                        .forRealm(AuthorizationConfiguration.DEFAULT_REALM_KEY)
```
dimas-b: Why `DEFAULT_REALM_KEY`? The returned `PolarisAuthorizerFactory` will be used to handle all requests, which may target different realms 🤔

dimas-b (commented):
Thanks for your contribution, @officialasishkumar !

I'm posting one preliminary comment (below) as a start of review on this PR.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
officialasishkumar: Removed the app-scoped factory producer so authorizer selection stays request-scoped and follows the active realm. Force-pushed the branch.

adutra (dismissed):

runtime/service/src/test/java/org/apache/polaris/service/config/AuthorizationConfigurationTest.java
```diff
@@ -0,0 +1,59 @@
...
+@TestProfile(AuthorizationConfigurationTest.Profile.class)
+public class AuthorizationConfigurationTest {
+
+  @Inject AuthorizationConfiguration authorizationConfig;
+
+  @Test
+  void smokeTest() {
+    assertThat(authorizationConfig.realms()).hasSize(3);
+    assertThat(authorizationConfig.realms().get(DEFAULT_REALM_KEY).type()).isEqualTo("internal");
+    assertThat(authorizationConfig.realms().get("realm1").type()).isEqualTo("test-authorizer");
```
dimas-b: Do you find a lot of value in such a test? It looks like it's mostly testing Smallrye / Quarkus than Polaris code... it gives some validation to config annotations in Polaris code, but `@QuarkusTest` is too heavy for that, I guess 🤔 Is there a way to test that in a simple (non-Quarkus) unit test?

runtime/service/src/intTest/java/org/apache/polaris/service/it/ServiceProducersIT.java
```diff
@@ -51,4 +71,108 @@ void testInternalPolarisAuthorizerProduced() {
...
+              authorizerFactories,
+              new NoopRealmConfig());
+      assertThat(realmAuthorizer).isInstanceOf(TestPolarisAuthorizer.class);
+
+      PolarisAuthorizer defaultAuthorizer =
+          serviceProducers.polarisAuthorizer(
+              () -> "other",
+              authorizationConfiguration,
+              authorizerFactories,
+              new NoopRealmConfig());
```
dimas-b: how about `new RealmConfigImpl(RealmConfigurationSource.EMPTY_CONFIG, () -> "other")`

runtime/service/src/test/java/org/apache/polaris/service/config/AuthorizationConfigurationTest.java
dimas-b: `PerRealmAuthorizationTest` probably provides enough coverage... I propose removing this test class completely.

runtime/service/src/intTest/java/org/apache/polaris/service/it/ServiceProducersIT.java
officialasishkumar: Updated the test to use `RealmConfigImpl(RealmConfigurationSource.EMPTY_CONFIG, ...)` instead of the local no-op implementation.

runtime/service/src/test/java/org/apache/polaris/service/config/AuthorizationConfigurationTest.java
officialasishkumar: Removed `AuthorizationConfigurationTest`; `PerRealmAuthorizationTest` now carries the coverage without the extra Quarkus smoke test.

dimas-b (dismissed):
Thanks, @officialasishkumar !

dimas-b:
@officialasishkumar : Please rebase to get recent CI job changes

runtime/defaults/src/main/resources/application.properties
```diff
@@ -304,4 +306,3 @@ quarkus.index-dependency.guava.artifact-id=guava
 
 # force the locale, just in case the system's using another default locale
 quarkus.default-locale=en_US
-
```
flyrain, line 307: nit: unnecessary change

flyrain (changes requested):
The code doesn't compile. 
```
        	at org.jboss.threads.JBossThread.run(JBossThread.java:501)
        Caused by: jakarta.enterprise.inject.UnsatisfiedResolutionException: Unsatisfied dependency for type org.apache.polaris.core.auth.PolarisAuthorizerFactory and qualifiers [@Default]
        	- injection target: parameter 'authorizerFactory' of org.apache.polaris.extension.auth.opa.OpaProductionReadinessChecks#checkOpaAuthorization()
        	- declared on PRODUCER METHOD bean [types=[java.lang.Object, org.apache.polaris.core.config.ProductionReadinessCheck], qualifiers=[@Any, @Default], target=org.apache.polaris.core.config.ProductionReadinessCheck checkOpaAuthorization(org.apache.polaris.core.auth.PolarisAuthorizerFactory authorizerFactory), declaringBean=org.apache.polaris.extension.auth.opa.OpaProductionReadinessChecks]
        	The following beans match by type, but none has matching qualifiers:
        	- CLASS bean [types=[java.lang.Object, org.apache.polaris.core.auth.PolarisAuthorizerFactory, org.apache.polaris.extension.auth.opa.OpaPolarisAuthorizerFactory], qualifiers=[@Any, @Identifier("opa")], target=org.apache.polaris.extension.auth.opa.OpaPolarisAuthorizerFactory]
        	- PRODUCER METHOD bean [types=[java.lang.Object, org.apache.polaris.core.auth.PolarisAuthorizerFactory], qualifiers=[@Any, @Identifier("internal")], target=org.apache.polaris.core.auth.PolarisAuthorizerFactory defaultPolarisAuthorizerFactory(), declaringBean=org.apache.polaris.service.config.ServiceProducers]
```

runtime/defaults/src/main/resources/application.properties
officialasishkumar: Removed while rebasing the branch onto current `main`.

officialasishkumar:
Updated in fa847e0dd. The OPA readiness check no longer injects a single default PolarisAuthorizerFactory; it inspects the configured authorization realm types and selects matching factories, so per-realm authorizer selection can stay request-scoped.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -140,18 +141,22 @@ public RealmConfig realmConfig(
...
   @ApplicationScoped
-  public PolarisAuthorizerFactory polarisAuthorizerFactory(
-      AuthorizationConfiguration authorizationConfig,
-      @Any Instance<PolarisAuthorizerFactory> authorizerFactories) {
-    PolarisAuthorizerFactory factory =
-        authorizerFactories.select(Identifier.Literal.of(authorizationConfig.type())).get();
-    return factory;
+  @Identifier("internal")
+  public PolarisAuthorizerFactory defaultPolarisAuthorizerFactory() {
+    return new DefaultPolarisAuthorizerFactory();
```
flyrain: Why do we need this? After #4267, the default factory is self annotated.

officialasishkumar: Removed the duplicate default authorizer producer and now rely on the self-annotated `DefaultPolarisAuthorizerFactory` bean from main. Commit: 538f9efa1. Local Gradle validation is blocked here because the environment has Java 17 and Polaris requires Java 21.

runtime/service/src/main/java/org/apache/polaris/service/config/AuthorizationConfiguration.java
```diff
@@ -19,11 +19,31 @@
...
+import org.apache.polaris.docs.ConfigDocs;
 
 @ConfigMapping(prefix = "polaris.authorization")
 public interface AuthorizationConfiguration {
 
-  @WithDefault("internal")
-  String type();
+  String DEFAULT_REALM_KEY = "<default>";
+
+  @WithParentName
```
flyingImer: This config shape seems to overload `polaris.authorization` in a way that is hard to reason about. With `@WithParentName` here, every first-level child under `polaris.authorization` starts looking like a realm id, but `polaris.authorization.opa.*` and `polaris.authorization.ranger.*` are already reserved backend config namespaces. So selection becomes per-realm while the backend config remains global, and it is no longer obvious which keys are legal realm ids vs implementation sections. Can we model that boundary explicitly instead of overloading the root prefix?

flyingImer: I think the main issue here is that `polaris.authorization` is now mixing two different namespaces: realm keys via `@WithParentName`, and backend-specific config like `polaris.authorization.opa.*`. That boundary feels too implicit to me.

Would it make sense to make realms explicit under their own subtree, or reserve backend names with validation?

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
```diff
@@ -138,20 +138,17 @@ public RealmConfig realmConfig(
...
   @RequestScoped
   public PolarisAuthorizer polarisAuthorizer(
-      PolarisAuthorizerFactory factory, RealmConfig realmConfig) {
+      RealmContext realmContext,
+      AuthorizationConfiguration authorizationConfig,
+      @Any Instance<PolarisAuthorizerFactory> authorizerFactories,
+      RealmConfig realmConfig) {
+    PolarisAuthorizerFactory factory =
+        authorizerFactories
+            .select(Identifier.Literal.of(authorizationConfig.forRealm(realmContext).type()))
```
flyingImer: I don't think this should fail this late. If a realm references an unknown authorizer type, `.get()` turns that into a request-time CDI failure. Would it make sense to either guard this with `isResolvable()` or validate configured types earlier?

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaProductionReadinessChecks.java
```diff
@@ -51,6 +70,7 @@ public ProductionReadinessCheck checkOpaAuthorization(
 
       return ProductionReadinessCheck.of(errors);
```
flyingImer: IIUC this returns after validating the first OPA realm. If multiple realms use OPA, only the first one gets checked. Is that intentional?

flyingImer (commented):
I think there are two blockers here.

First, `ServiceProducersIT` does not compile on the current head because `RealmConfig` is referenced without an import.

Second, I don't think the config contract is explicit enough yet. `AuthorizationConfiguration` now uses `@WithParentName` to treat `polaris.authorization.<name>` as a realm key space, but `polaris.authorization.opa.*` is already being used as a backend-specific config namespace, and `ranger.*` may follow the same pattern. So `polaris.authorization` is doing two jobs at once, which makes the boundary hard to reason about in a way that `polaris.authentication` does not have.

My bias would be to make that boundary explicit, either by giving realms their own subtree (for example `polaris.authorization.realms.<name>.*`) or by reserving backend names with validation.

Separately, this looks like a reasonable step toward realm-aware authorizer selection, but it is still type routing only. It does not make backend configuration realm-aware. I think it would help to say that explicitly in the PR description so this does not read as full forkless/per-realm authorization config support.

A couple of non-blocking things also stood out: `OpaProductionReadinessChecks` appears to stop after the first OPA realm, and `ServiceProducers.polarisAuthorizer()` currently fails at request time with an opaque CDI error if a realm references an unknown type.

runtime/service/src/intTest/java/org/apache/polaris/service/it/ServiceProducersIT.java
```diff
@@ -51,4 +70,79 @@ void testInternalPolarisAuthorizerProduced() {
...
+              new RealmConfigImpl(RealmConfigurationSource.EMPTY_CONFIG, () -> "other"));
+      assertThat(defaultAuthorizer).isInstanceOf(PolarisAuthorizerImpl.class);
+    }
+  }
+
+  @ApplicationScoped
+  @Identifier("test-authorizer")
+  public static class TestPolarisAuthorizerFactory implements PolarisAuthorizerFactory {
+    @Override
+    public PolarisAuthorizer create(RealmConfig realmConfig) {
```
flyingImer, line 132: This does not compile as-is. `RealmConfig` is used here but not imported.

runtime/service/src/main/java/org/apache/polaris/service/config/AuthorizationConfiguration.java
officialasishkumar: Addressed in 14128de54fdfd3f7ffbb8399e4c710da44f7ae2a by separating realm authorization config from backend-specific config and validating reserved names.

runtime/service/src/main/java/org/apache/polaris/service/config/ServiceProducers.java
officialasishkumar: Addressed in 14128de54fdfd3f7ffbb8399e4c710da44f7ae2a. Unknown authorizer types are now guarded before request-time CDI lookup.

extensions/auth/opa/impl/src/main/java/org/apache/polaris/extension/auth/opa/OpaProductionReadinessChecks.java
officialasishkumar: Addressed in 14128de54fdfd3f7ffbb8399e4c710da44f7ae2a. The OPA readiness check now validates all OPA realms.

runtime/service/src/intTest/java/org/apache/polaris/service/it/ServiceProducersIT.java
officialasishkumar: Fixed in 14128de54fdfd3f7ffbb8399e4c710da44f7ae2a, with the affected service integration tests passing locally.

PR #4160, opened 2026-04-09
fix(iceberg): preserve table-default properties for staged creates

### LINKED ISSUE #4062: table-default.* catalog properties not merged into S3FileIO for staged creates and doCommit
## Bug Report

**What version of Apache Polaris are you using?**
1.3.0-incubating

**What OS and processor architecture?**
Linux x86_64 (OpenShift / Kubernetes pod)

**What did you do?**
Configured Polaris with `SKIP_CREDENTIAL_SUBSCOPING_INDIRECTION=true` and `table-default.s3.endpoint` / `table-default.s3.path-style-access` catalog properties (required for S3-compatible stores like VAST that don't support STS AssumeRole). Then used DuckDB with `stage-create: true` to create an Iceberg table.

**What did you expect to see?**
S3FileIO uses `s3.endpoint` and `s3.path-style-access` from catalog properties when writing table metadata to S3.

**What did you see instead?**
S3FileIO ignores the catalog defaults and falls back to the default AWS S3 endpoint, causing 403 errors when the Polaris S3 credentials are not valid for real AWS.

---

## Root Cause

Three code paths fail to merge `table-default.*` catalog properties into the properties passed to S3FileIO:

### 1. `PolarisIcebergCatalogTableBuilder` constructor (`IcebergCatalog.java`)

`PolarisIcebergCatalogViewBuilder` correctly calls `withProperties(propertiesWithPrefix(..., "table-default."))` in its constructor. The table builder is missing this call.

```java
// Fix: add to PolarisIcebergCatalogTableBuilder constructor
withProperties(
    PropertyUtil.propertiesWithPrefix(IcebergCatalog.this.properties(), "table-default."));
```

### 2. `IcebergCatalogHandler.stageTableCreateHelper()`

Clients like DuckDB send `stage-create: true`, which calls `TableMetadata.newTableMetadata()` directly, bypassing `buildTable()`. The `properties` map never receives catalog defaults.

```java
// Fix: merge catalog defaults before calling newTableMetadata()
Map<String, String> mergedProperties = Maps.newHashMap();
if (baseCatalog instanceof IcebergCatalog polarisCatalog) {
    mergedProperties.putAll(PropertyUtil.propertiesWithPrefix(
        polarisCatalog.properties(), "table-default."));
}
mergedProperties.putAll(properties);
```

### 3. `BasePolarisTableOperations.doCommit()` (`IcebergCatalog.java`)

DuckDB's staged create commits send sparse payloads with no `SetProperties` update, so `metadata.properties()` is empty of catalog defaults. The FileIO for the commit write is initialized only from `metadata.properties()`.

```java
// Fix: use tableDefaultProperties as baseline in doCommit()
Map<String, String> commitFileIOProps = new HashMap<>(tableDefaultProperties);
commitFileIOProps.putAll(metadata.properties());
tableFileIO = loadFileIOForTableLike(..., commitFileIOProps, ...);
```

All three fixes are needed for DuckDB (staged create). Fix 1 alone is sufficient for direct creates (`stage-create: false`).

officialasishkumar:
Ensure catalog-level `table-default.*` properties survive all table creation paths. The table builder now inherits them during direct creates, staged metadata construction adds them before returning table metadata to the client, and sparse staged commits use them as the baseline when FileIO is reinitialized during commit.

The accompanying tests cover staged create metadata plus FileIO initialization in both relational and NoSQL catalog variants.

## Testing
- `export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64 && ./gradlew format`
- `export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64 && ./gradlew :polaris-runtime-service:test --tests 'org.apache.polaris.service.catalog.iceberg.IcebergCatalogRelationalTest.testDirectCreatePassesTableDefaultPropertiesToFileIO' --tests 'org.apache.polaris.service.catalog.iceberg.IcebergCatalogRelationalTest.testSparseCreateCommitPassesTableDefaultPropertiesToFileIO' --tests 'org.apache.polaris.service.catalog.iceberg.IcebergCatalogNoSqlInMemTest.testDirectCreatePassesTableDefaultPropertiesToFileIO' --tests 'org.apache.polaris.service.catalog.iceberg.IcebergCatalogNoSqlInMemTest.testSparseCreateCommitPassesTableDefaultPropertiesToFileIO' --tests 'org.apache.polaris.service.catalog.iceberg.IcebergCatalogHandlerAuthzTest.testCreateTableStagedIncludesTableDefaultProperties' --tests 'org.apache.polaris.service.catalog.iceberg.IcebergCatalogHandlerNoSqlAuthzTest.testCreateTableStagedIncludesTableDefaultProperties'`\n\n## Checklist\n- [x] 🛡️ Don't disclose security issues! (contact security@apache.org)\n- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #4062\n- [x] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)\n- [ ] 💡 Added comments for complex logic\n- [ ] 🧾 Updated `CHANGELOG.md` (if needed)\n- [ ] 📚 Updated `site/content/in-dev/unreleased` documentation (if needed)

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1319,6 +1319,8 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
     public PolarisIcebergCatalogTableBuilder(TableIdentifier identifier, Schema schema) {
       super(identifier, schema);
+      withProperties(
+          PropertyUtil.propertiesWithPrefix(IcebergCatalog.this.properties(), "table-default."));
       this.identifier = identifier;
     }
 
@@ -1531,12 +1533,14 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
               : resolvedTableEntities;
 
       // refresh credentials because we need to read the metadata file to validate its location
+      Map<String, String> commitFileIoProperties = new HashMap<>(tableDefaultProperties);
+      commitFileIoProperties.putAll(metadata.properties());
       tableFileIO =
           loadFileIOForTableLike(
               tableIdentifier,
               StorageUtil.getLocationsUsedByTable(metadata),
               resolvedStorageEntity,
-              new HashMap<>(metadata.properties()),
+              commitFileIoProperties,
               Set.of(
                   PolarisStorageActions.READ,
                   PolarisStorageActions.WRITE,
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1319,6 +1319,7 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
 
     public PolarisIcebergCatalogTableBuilder(TableIdentifier identifier, Schema schema) {
       super(identifier, schema);
+      withProperties(new HashMap<>(tableDefaultProperties));
       this.identifier = identifier;
     }
 
@@ -1535,12 +1536,14 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
               : resolvedTableEntities;
 
       // refresh credentials because we need to read the metadata file to validate its location
+      Map<String, String> commitFileIoProperties = new HashMap<>(tableDefaultProperties);
+      commitFileIoProperties.putAll(metadata.properties());
       tableFileIO =
           loadFileIOForTableLike(
               tableIdentifier,
               StorageUtil.getLocationsUsedByTable(metadata),
               resolvedStorageEntity,
-              new HashMap<>(metadata.properties()),
+              commitFileIoProperties,
               Set.of(
                   PolarisStorageActions.READ,
                   PolarisStorageActions.WRITE,
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -1536,14 +1536,12 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
               : resolvedTableEntities;
 
       // refresh credentials because we need to read the metadata file to validate its location
-      Map<String, String> commitFileIoProperties = new HashMap<>(tableDefaultProperties);
-      commitFileIoProperties.putAll(metadata.properties());
       tableFileIO =
           loadFileIOForTableLike(
               tableIdentifier,
               StorageUtil.getLocationsUsedByTable(metadata),
               resolvedStorageEntity,
-              commitFileIoProperties,
+              new HashMap<>(metadata.properties()),
               Set.of(
                   PolarisStorageActions.READ,
                   PolarisStorageActions.WRITE,
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
@@ -49,6 +49,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.iceberg.BaseMetadataTable;
 import org.apache.iceberg.BaseTable;
+import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.MetadataUpdate;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.SnapshotRef;
@@ -89,6 +90,7 @@ import org.apache.iceberg.rest.responses.ListTablesResponse;
 import org.apache.iceberg.rest.responses.LoadTableResponse;
 import org.apache.iceberg.rest.responses.LoadViewResponse;
 import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
+import org.apache.iceberg.util.PropertyUtil;
 import org.apache.polaris.core.PolarisDiagnostics;
 import org.apache.polaris.core.auth.PolarisAuthorizableOperation;
 import org.apache.polaris.core.catalog.FederatedCatalogFactory;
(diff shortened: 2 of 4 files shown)


runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/AbstractIcebergCatalogHandlerAuthzTest.java
```diff
@@ -501,6 +502,45 @@ Stream<DynamicNode> testCreateTableStagedPrivileges() {
...
+
+    LocalCatalogFactory defaultPropertiesFactory =
+        resolvedManifest -> {
+          Catalog catalog = localCatalogFactory.createCatalog(resolvedManifest);
+          catalog.initialize(
+              CATALOG_NAME,
+              ImmutableMap.of(
+                  CatalogProperties.FILE_IO_IMPL,
+                  "org.apache.iceberg.inmemory.InMemoryFileIO",
+                  CatalogProperties.TABLE_DEFAULT_PREFIX + "default-key1",
```
dimas-b, line 522: Hi @officialasishkumar ! This is an interesting use case, but I wonder whether the changes in this PR have any effect at the Polaris REST API level? How can API users / clients benefit from this change? Thx!

officialasishkumar: Yes. createTableStaged returns a LoadTableResponse, and the staged metadata now carries the catalog table-default entries in tableMetadata(), so REST clients can see those defaults immediately. The same defaults are also used for direct-create and commit-time FileIO initialization.

dimas-b: ... but how can the catalog get those "table-default" entries in a realistic (non-test) env.? :thinking:

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1319,6 +1319,8 @@ private class PolarisIcebergCatalogTableBuilder
 
     public PolarisIcebergCatalogTableBuilder(TableIdentifier identifier, Schema schema) {
       super(identifier, schema);
+      withProperties(
+          PropertyUtil.propertiesWithPrefix(IcebergCatalog.this.properties(), "table-default."));
```
flyrain: Can we use the field `tableDefaultProperties` here directly instead of grabbing them again?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -1531,12 +1533,14 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
               : resolvedTableEntities;
 
       // refresh credentials because we need to read the metadata file to validate its location
+      Map<String, String> commitFileIoProperties = new HashMap<>(tableDefaultProperties);
+      commitFileIoProperties.putAll(metadata.properties());
```
flyrain, line 1540: This is a direct creation path, so do we change the direct table creation behave as well? If yes, this doesn't align with the PR title.

Moreover, this is a behavior change we need to be careful.  Can you share this behavior change in the dev mailing list?

flyrain: Look a bit more, do we even need this change? The direct creation will invoke the `PolarisIcebergCatalogTableBuilder` to get these properties, https://github.com/apache/polaris/blob/c7098e86e9a3c7398668816474026162c4229f91/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L484. Why do we need to inject these properties again?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -517,6 +519,11 @@ private TableMetadata stageTableCreateHelper(Namespace namespace, CreateTableReq
 
     Map<String, String> properties = Maps.newHashMap();
     properties.put("created-at", OffsetDateTime.now(ZoneOffset.UTC).toString());
+    if (baseCatalog instanceof IcebergCatalog polarisCatalog) {
+      properties.putAll(
+          PropertyUtil.propertiesWithPrefix(
+              polarisCatalog.properties(), CatalogProperties.TABLE_DEFAULT_PREFIX));
```
flyrain: Why do we need this given that we use the builder `PolarisIcebergCatalogTableBuilder` a few line down?
```
        baseCatalog
              .buildTable(ident, request.schema())  /// this should get the properties
              .withPartitionSpec(request.spec())
```

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
officialasishkumar: Done in 07d8fbe9e: the builder now initializes from `tableDefaultProperties` directly.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
officialasishkumar: Updated in 07d8fbe9e: the builder owns the catalog defaults, and the handler passes a copied property map so staged metadata keeps the same defaults without mutating shared state.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
officialasishkumar: Removed the direct-commit FileIO property merge and the sparse commit test so the PR stays scoped to the staged/create builder path. Commit: f63dab5d5. I could not run the Gradle test locally because this environment has Java 17 and the Polaris build requires Java 21.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java
```diff
@@ -475,6 +478,11 @@ public LoadTableResponse createTableDirect(
...
```
dimas-b, line 484: I believe this kind of behaviour change deserves a discussed on the `dev` ML... Is there an existing thread for this?

PR #4167, opened 2026-04-10
CLI: preserve optional S3 fields in setup catalog creation

### LINKED ISSUE #4076: Creating catalogs using the cli is buggy
### Describe the bug

I am trying to use the cli command to create catalog using `polaris setup apply test-setup.yaml` command: 
    even though i set the sts_unavailable flag to true it still asks for ARN like so 
```
apache_polaris/sdk/management/exceptions.py", line 161, in from_response
    raise BadRequestException(http_resp=http_resp, body=body, data=data)
apache_polaris.sdk.management.exceptions.BadRequestException: (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'content-length': '90', 'Content-Type': 'application/json', 'X-Request-ID': '819c92cb-c6d7-4fda-8093-82c2a352fdf9_0000000000000000251'})
HTTP response body: {"error":{"message":"ARN must not be empty","type":"IllegalArgumentException","code":400}}
```
when i provide some dummy ARN value then the catalog gets created and also the namespaces but when creating tables i get the following error

```
IllegalArgumentException: Credential vending was requested for table Senec.Stage.Bronze.Ampace_V3LFP.BmsAmpaceV3ModuleMeasurement, but no credentials are available
```

however, using the `polaris catalogs create` commands works as expected.

### To Reproduce

The minimal test-setup.yaml file is as follows:

```
catalogs:
  - name: "senec_catalog_6"
    storage_type: "s3"
    type: internal
    default_base_location: "s3://senec-bucket/senec_catalog_6/"
    allowed_locations:
      - "s3://senec-bucket/senec_catalog_6/"
    region: "us-east-1"
    # Custom S3 endpoint.
    endpoint: "http://localhost:9000/"
    endpoint_internal: "http://rustfs:9000"
    path_style_access: true
    # sts_unavailable: true
    # kms_unavailable: true
    properties:
      catalog: senec_catalog_6
```
my yaml file has a bit more things like setting up the principals, and other entities but creating catalogs is what causing the issue. 

and the exact command that i used is 
`polaris --profile root setup apply ./test-setup.yaml`

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Additional context

I am trying to use Polaris REST catalog along with PyIceberg engine and RustFS for storage. I am still new to this so I might be doing something wrong. I am happy to share more info as required. 

### System information

I am running ubuntu and i have latest polaris version from main.

officialasishkumar:
`polaris setup apply` was rebuilding `CatalogsCommand` instances with empty-string defaults for omitted optional storage and authentication fields. The direct CLI path passes `None` for those fields instead.

For S3 catalogs this caused setup-driven catalog creation to serialize an empty `roleArn`, which the server rejects, and it diverged from the behavior of the working `polaris catalogs create` command. This change aligns the setup path with the direct CLI semantics, updates `CatalogsCommand` annotations to reflect that these inputs are optional, and adds a regression test that compares the direct CLI and setup-apply request payloads for the same S3 catalog definition.

## Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #4076
- [x] 🧾 Updated `CHANGELOG.md` (not needed)
- [x] 📚 Updated documentation in `site/content/in-dev/unreleased` (not needed)

Validation:
- `./.venv/bin/python -m pytest client/python/test -q`
- `make client-lint`

### DIFF
--- a/client/python/apache_polaris/cli/command/catalogs.py
+++ b/client/python/apache_polaris/cli/command/catalogs.py
@@ -74,48 +74,48 @@ class CatalogsCommand(Command):
 
     catalogs_subcommand: str
     catalog_type: str
-    default_base_location: str
-    storage_type: str
-    allowed_locations: List[str]
-    role_arn: str
-    external_id: str
-    user_arn: str
-    region: str
-    tenant_id: str
-    multi_tenant_app_name: str
+    default_base_location: Optional[str]
+    storage_type: Optional[str]
+    allowed_locations: Optional[List[str]]
+    role_arn: Optional[str]
+    external_id: Optional[str]
+    user_arn: Optional[str]
+    region: Optional[str]
+    tenant_id: Optional[str]
+    multi_tenant_app_name: Optional[str]
     hierarchical: bool
-    consent_url: str
-    service_account: str
+    consent_url: Optional[str]
+    service_account: Optional[str]
     catalog_name: str
     properties: Dict[str, StrictStr]
     set_properties: Dict[str, StrictStr]
     remove_properties: List[str]
-    hadoop_warehouse: str
-    hive_warehouse: str
-    iceberg_remote_catalog_name: str
-    endpoint: str
-    endpoint_internal: str
-    sts_endpoint: str
+    hadoop_warehouse: Optional[str]
+    hive_warehouse: Optional[str]
+    iceberg_remote_catalog_name: Optional[str]
+    endpoint: Optional[str]
+    endpoint_internal: Optional[str]
+    sts_endpoint: Optional[str]
     sts_unavailable: bool
     kms_unavailable: bool
     path_style_access: bool
-    current_kms_key: str
-    allowed_kms_keys: List[str]
-    catalog_connection_type: str
-    catalog_authentication_type: str
-    catalog_service_identity_type: str
-    catalog_service_identity_iam_arn: str
-    catalog_uri: str
-    catalog_token_uri: str
-    catalog_client_id: str
-    catalog_client_secret: str
+    current_kms_key: Optional[str]
+    allowed_kms_keys: Optional[List[str]]
+    catalog_connection_type: Optional[str]
+    catalog_authentication_type: Optional[str]
+    catalog_service_identity_type: Optional[str]
+    catalog_service_identity_iam_arn: Optional[str]
+    catalog_uri: Optional[str]
+    catalog_token_uri: Optional[str]
+    catalog_client_id: Optional[str]
+    catalog_client_secret: Optional[str]
     catalog_client_scopes: List[str]
-    catalog_bearer_token: str
-    catalog_role_arn: str
-    catalog_role_session_name: str
-    catalog_external_id: str
-    catalog_signing_region: str
-    catalog_signing_name: str
+    catalog_bearer_token: Optional[str]
+    catalog_role_arn: Optional[str]
+    catalog_role_session_name: Optional[str]
+    catalog_external_id: Optional[str]
+    catalog_signing_region: Optional[str]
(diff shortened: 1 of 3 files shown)


MonkeyCanCode:
thanks for the PR @officialasishkumar . I will take a look a bit later this weekend.

MonkeyCanCode:
Thanks for the PR @officialasishkumar. So the reason on why I put `""|[]|{}` as defaults due to `catalogs.py` (as well as other classes) have all of these fields as non-optional. Thus, I had to list out all fields in the initial PR.

Now back to this problem where an user who tried to create a catalog on s3 without providing `role_arn` (which will result in `""` with this logic), the change in the PR may be in-complete as we are updating the type hints but not provide default values. In python dataclass system, that would still mean they are required fields and can cause TypeError later on. For more completeness fix, we should provide default value for the optional fields (which I think is the right thing to do).

Beside this, we can also update `validate(self)` as well as `_build_storage_config_info` force `""` to become `None` instead (bandaid fixes imo) which will prevent this problem from the first place as well.  

What do you think?

cc @flyrain @HonahX for additional review and suggestion.

officialasishkumar:
Thanks for the detailed read, @MonkeyCanCode.

You're right that switching the annotations to `Optional[...]` without adding default values only fixes the type hint side; the dataclass still treats them as required positional fields. The reason the PR works today is that the only caller (`setup.py`) explicitly passes all of them as kwargs, so the `TypeError` risk you're describing doesnt surface.

On the two options you proposed:

1. **Add defaults (`None` / `[]` / `{}`) on the `CatalogsCommand` dataclass fields.** This is the most correct fix and I'd prefer it. It makes the optionality real at the dataclass level. The one wrinkle is dataclass ordering: every `Optional` field will need to come after the remaining non-default fields, so the declaration order needs a small reshuffle. I don't think that's disruptive.

2. **Coerce `""` → `None` in `validate()` / `_build_storage_config_info`.** Agree this is more bandaid-ish; it hides the problem at one call site but doesn't prevent other callers from passing empty strings.

I'm happy to update this PR with option (1), including reshuffling the field order and adding defaults. Want me to do that in the same PR?

MonkeyCanCode:
> Thanks for the detailed read, @MonkeyCanCode.
> 
> You're right that switching the annotations to `Optional[...]` without adding default values only fixes the type hint side; the dataclass still treats them as required positional fields. The reason the PR works today is that the only caller (`setup.py`) explicitly passes all of them as kwargs, so the `TypeError` risk you're describing doesnt surface.
> 
> On the two options you proposed:
> 
> 1. **Add defaults (`None` / `[]` / `{}`) on the `CatalogsCommand` dataclass fields.** This is the most correct fix and I'd prefer it. It makes the optionality real at the dataclass level. The one wrinkle is dataclass ordering: every `Optional` field will need to come after the remaining non-default fields, so the declaration order needs a small reshuffle. I don't think that's disruptive.
> 2. **Coerce `""` → `None` in `validate()` / `_build_storage_config_info`.** Agree this is more bandaid-ish; it hides the problem at one call site but doesn't prevent other callers from passing empty strings.
> 
> I'm happy to update this PR with option (1), including reshuffling the field order and adding defaults. Want me to do that in the same PR?

That is correct. Again, thanks for the contribution and reporting of this issue. With above fix, there are other potential issues as other data classes which can surface later as they all have the same problem. I had raised https://github.com/apache/polaris/pull/4172 based on discussion above. Let me know if this fixed the problem you are facing.

MonkeyCanCode:
@officialasishkumar https://github.com/apache/polaris/pull/4172 is merged and this should fix the problem for this one.

PR #4172, opened 2026-04-12
Fix setup command default issue

MonkeyCanCode:
The current `setup` command was serializing optional fields such as `roleArn` (reported in https://github.com/apache/polaris/pull/4167) as well as `stsEndpoint` (reported in https://github.com/apache/polaris/pull/4076). Both of them are due to `Command` dataclasses don't support default value which making almost all fields mandatory in the constructor.

While I work on the `setup` command, my assumption was default value such as `""` are safe to provide and server will handle them (miss on my end).

Now to fix this issue, this PR changes all Command dataclasses to use a flexible model by using `Optional` type hints with `None` as defaults. Then I use `__post_init__` to ensure mandatory fields are getting update to proper default values if not been implicitly defined. As they are now optional, I then ensure we have proper tests in `validation()` to cover all required fields as well as using implicitly type casting to ensure mypy is happy with these changes. 

## Checklist

### DIFF
--- a/client/python/apache_polaris/cli/command/catalogs.py
+++ b/client/python/apache_polaris/cli/command/catalogs.py
@@ -30,9 +30,9 @@ from apache_polaris.cli.constants import (
 )
 from apache_polaris.cli.options.option_tree import Argument
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pydantic import StrictStr, SecretStr
-from typing import Dict, List, Optional, Union, Tuple, Callable
+from typing import Dict, List, Optional, Union, Tuple, Callable, cast
 
 from apache_polaris.sdk.management import (
     PolarisDefaultApi,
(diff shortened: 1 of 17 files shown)


MonkeyCanCode:
@officialasishkumar thanks for your PR in https://github.com/apache/polaris/pull/4167 and I believed this PR should addressed all of the issues in other classes as well as correct semantic for using `Optional`. Please take a look if we should proceed with this route.

@Zevrap-81, this PR should fix the problem which you are facing.

client/python/apache_polaris/cli/command/catalogs.py
```diff
@@ -490,24 +520,25 @@ def execute(self, api: PolarisDefaultApi) -> None:
...
-        catalog = api.get_catalog(self.catalog_name)
+        catalog = api.get_catalog(catalog_name)
         print("Metadata")
         print(f"  {'Type:':<30} {catalog.type}")
         storage_info = catalog.storage_config_info
         print(f"  {'Storage Type:':<30} {storage_info.storage_type}")
         print(f"  {'Base Location:':<30} {catalog.properties.default_base_location}")
         print(
-            f"  {'Allowed Locations:':<30} {', '.join(storage_info.allowed_locations or [])}"
+            f"  {'Allowed Locations:':<30} {', '.join(storage_info.allowed_locations)}"
```
jbonofre: Shouldn't we keep the `or []` here ?

If `allowed_locations` is `None` from the API response, this will crash with `TypeError: can only join an iterable`.

jbonofre (dismissed):

client/python/apache_polaris/cli/command/catalogs.py
MonkeyCanCode: Oh yeah, very sharp eyes @jbonofre . That is correct as this is not a must have field and not something users passed-in. Let me fix this shortly.

dimas-b (approved):

flyrain (approved):
LGTM. Thanks @MonkeyCanCode !

PR #4173, opened 2026-04-12
fix: enforce single expiration timestamp per credential bundle

yushesp:
ConnectionCredentials.of() silently overwrites the expiresAt field when multiple expiration timestamp props are present. This produces non-deterministic results since the input is a HashMap (no order).

Instead, this change fails fast by throwing an exception when multiple timestamps are found.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
@@ -41,6 +41,10 @@ public record ConnectionCredentials(Map<String, String> credentials, Optional<In
   /**
    * Creates ConnectionCredentials from a map of catalog access properties. Expiration timestamps
    * are extracted and stored separately; credential properties are stored in the credentials map.
+   *
+   * <p>Only a single shared expiration timestamp is supported for a credentials bundle. If multiple
+   * expiration timestamp properties are present, they must all represent the same instant or an
+   * {@link IllegalArgumentException} is thrown.
    */
   public static ConnectionCredentials of(Map<CatalogAccessProperty, String> properties) {
     Map<String, String> credentials = new HashMap<>();
@@ -48,7 +52,16 @@ public record ConnectionCredentials(Map<String, String> credentials, Optional<In
     for (var entry : properties.entrySet()) {
       CatalogAccessProperty key = entry.getKey();
       if (key.isExpirationTimestamp()) {
-        expiresAt = Instant.ofEpochMilli(Long.parseLong(entry.getValue()));
+        Instant current = Instant.ofEpochMilli(Long.parseLong(entry.getValue()));
+        if (expiresAt == null) {
+          expiresAt = current;
+        } else if (!current.equals(expiresAt)) {
+          throw new IllegalArgumentException(
+              "Multiple distinct expiration timestamps found while building ConnectionCredentials: "
+                  + expiresAt.toEpochMilli()
+                  + " and "
+                  + current.toEpochMilli());
+        }
       }
       if (key.isCredential()) {
         credentials.put(key.getPropertyName(), entry.getValue());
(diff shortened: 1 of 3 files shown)


polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -41,14 +41,27 @@ public record ConnectionCredentials(Map<String, String> credentials, Optional<In
...
     Instant expiresAt = null;
     for (var entry : properties.entrySet()) {
       CatalogAccessProperty key = entry.getKey();
       if (key.isExpirationTimestamp()) {
-        expiresAt = Instant.ofEpochMilli(Long.parseLong(entry.getValue()));
+        Instant current = Instant.ofEpochMilli(Long.parseLong(entry.getValue()));
+        if (expiresAt == null) {
+          expiresAt = current;
+        } else if (!current.equals(expiresAt)) {
+          throw new IllegalArgumentException(
```
dimas-b, line 56: Nice catch, @yushesp !

Side note: we should probably make a similar fix in `StorageAccessConfig` (separate PR)

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -41,14 +41,27 @@ public record ConnectionCredentials(Map<String, String> credentials, Optional<In
...
+              "Multiple distinct expiration timestamps found while building ConnectionCredentials: "
+                  + expiresAt.toEpochMilli()
```
dimas-b: Please include the keys in the error message (for trouble-shooting).

dimas-b (dismissed):

dimas-b:
FYI: @XJDKC

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
yushesp: sounds good, will do!

yushesp:
@dimas-b Thank you for the review! I added in the keys to make it easier to troubleshoot like you suggested. Will follow up with another PR in a bit to fix `StorageAccessConfig`.

dimas-b (dismissed):
Let's keep this PR in review until Apr 15 (usual 2 days review grace time) and merge then.

@yushesp : because this change may affect existing deployments, please add a note to the "breaking changes" section of CHANGELOG.md.

dimas-b (dismissed):

dimas-b:
https://lists.apache.org/thread/9tt2bw1sgjh8tyxmv95r4jbnfy6qy6fo

polaris-core/src/main/java/org/apache/polaris/core/credentials/connection/ConnectionCredentials.java
```diff
@@ -41,14 +41,33 @@ public record ConnectionCredentials(Map<String, String> credentials, Optional<In
...
     Instant expiresAt = null;
+    CatalogAccessProperty expiresAtKey = null;
     for (var entry : properties.entrySet()) {
       CatalogAccessProperty key = entry.getKey();
       if (key.isExpirationTimestamp()) {
-        expiresAt = Instant.ofEpochMilli(Long.parseLong(entry.getValue()));
+        Instant current = Instant.ofEpochMilli(Long.parseLong(entry.getValue()));
+        if (expiresAt == null) {
+          expiresAt = current;
+          expiresAtKey = key;
```
flyrain: Do we want to handle the cases that there are multiple keys for `ExpirationTimestamp`? 
I think each `ConnectionCredentials` will only need one expire-at. cc @XJDKC 

If not, I can we can simplify the logic to throw when there are two `ExpirationTimestamp` presenting.

yushesp: Thanks for the suggestion @flyrain! I agree, each `ConnectionCredentials` bundle should only need one expire-at. I'll follow up with an update to the PR that simplifies this to throw when more than one expiration timestamp property is present.

flyrain (approved):
Thanks @yushesp !

dimas-b (approved):

PR #4190, opened 2026-04-14
Fix entity name overlap handling and tests (#4168)

### LINKED ISSUE #4168: Add entity name overlap tests
Following up on the discussion in #4148 (related to #4108), it seems valuable to add generic entity name collision tests to `polaris-tests`.

Previous work on this from @Subham-KRLX (class `PolarisRestCatalogIntegrationBase`):
```
  @Test
  public void testCreateTableNameThatCollidesWithNamespace() {
    Assumptions.assumeThat(testNamespaceTableNameCollision()).isTrue();

    Namespace parentNamespace = Namespace.of("ns1");
    restCatalog.createNamespace(parentNamespace);
    restCatalog.createNamespace(Namespace.of("ns1", "clash"));

    assertThatThrownBy(
            () ->
                restCatalog
                    .buildTable(TableIdentifier.of(parentNamespace, "clash"), SCHEMA)
                    .create())
        .isInstanceOf(AlreadyExistsException.class)
        .hasMessageContaining("already exists");
  }
```

Each name withing a namespaces should only map to at most one entity type (Table, (nested) Namespace, View, etc.)

Subham-KRLX:
Fixes #4168 by adding shared integration tests for namespace/table/view name collisions and enforcing consistent [AlreadyExistsException](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) behavior across backends. Verified with [RestCatalogFileIT](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [NoSqlCatalogIT](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -648,8 +648,30 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
   }
 
   private boolean namespaceWithSameNameExists(TableIdentifier identifier) {
-    return listNamespaces(identifier.namespace()).stream()
-        .anyMatch(namespace -> namespace.level(namespace.length() - 1).equals(identifier.name()));
+    PolarisResolvedPathWrapper resolvedEntities =
+        resolvedEntityView.getResolvedPath(ResolvedPathKey.ofNamespace(identifier.namespace()));
+    if (resolvedEntities == null) {
+      return false;
+    }
+
+    List<PolarisEntity> catalogPath = resolvedEntities.getRawFullPath();
+    EntityResult lookupResult =
+        getMetaStoreManager()
+            .readEntityByName(
+                getCurrentPolarisContext(),
+                PolarisEntity.toCoreList(catalogPath),
+                PolarisEntityType.NAMESPACE,
+                PolarisEntitySubType.NULL_SUBTYPE,
+                identifier.name());
+    if (lookupResult.isSuccess()) {
+      return true;
+    }
+    if (lookupResult.getReturnStatus() != BaseResult.ReturnStatus.ENTITY_NOT_FOUND) {
+      throw new ServiceFailureException(
+          "Failed to check existing namespace for '%s'. Status: %s ExtraInfo: %s",
+          identifier.name(), lookupResult.getReturnStatus(), lookupResult.getExtraInformation());
+    }
+    return false;
   }
 
   private PolarisResolvedPathWrapper getResolvedParentNamespace(Namespace namespace) {
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
@@ -513,6 +513,24 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
       PolarisResolvedPathWrapper resolvedParent) {
     String baseLocation = resolveNamespaceLocation(namespace, metadata);
 
+    // Prevent namespace creation when a table or view with the same name already exists.
+    Namespace parentNamespace = PolarisCatalogHelpers.getParentNamespace(namespace);
+    String namespaceName = namespace.level(namespace.length() - 1);
+    Optional<PolarisEntitySubType> conflictingSubType =
+        Stream.of(PolarisEntitySubType.ICEBERG_TABLE, PolarisEntitySubType.ICEBERG_VIEW)
+            .filter(
+                subType ->
+                    listTableLike(subType, parentNamespace, PageToken.readEverything())
+                        .items()
+                        .stream()
+                        .anyMatch(identifier -> identifier.name().equals(namespaceName)))
+            .findFirst();
+    if (conflictingSubType.isPresent()) {
+      TableIdentifier tableLikeIdentifier = TableIdentifier.of(parentNamespace, namespaceName);
+      throw alreadyExistsExceptionWithSameNameForTableLikeEntity(
+          tableLikeIdentifier, conflictingSubType.get());
+    }
+
     // Set / suffix
     boolean requireTrailingSlash =
         realmConfig.getConfig(FeatureConfiguration.ADD_TRAILING_SLASH_TO_LOCATION);
@@ -628,6 +646,18 @@ public class IcebergCatalog extends BaseMetastoreViewCatalog
     } else {
       return location;
     }
+  }
+
+  private static Namespace tableLikeIdentifierAsNamespace(TableIdentifier identifier) {
+    String[] namespaceLevels =
+        Arrays.copyOf(identifier.namespace().levels(), identifier.namespace().length() + 1);
+    namespaceLevels[identifier.namespace().length()] = identifier.name();
+    return Namespace.of(namespaceLevels);
+  }
+
+  private boolean namespaceWithSameNameExists(TableIdentifier identifier) {
+    return listNamespaces(identifier.namespace()).stream()
+        .anyMatch(namespace -> namespace.level(namespace.length() - 1).equals(identifier.name()));
   }
 
   private PolarisResolvedPathWrapper getResolvedParentNamespace(Namespace namespace) {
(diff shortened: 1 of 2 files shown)


runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -513,6 +513,24 @@ private void createNamespaceInternal(
...
     String baseLocation = resolveNamespaceLocation(namespace, metadata);
 
+    // Prevent namespace creation when a table or view with the same name already exists.
+    Namespace parentNamespace = PolarisCatalogHelpers.getParentNamespace(namespace);
+    String namespaceName = namespace.level(namespace.length() - 1);
+    Optional<PolarisEntitySubType> conflictingSubType =
+        Stream.of(PolarisEntitySubType.ICEBERG_TABLE, PolarisEntitySubType.ICEBERG_VIEW)
+            .filter(
+                subType ->
+                    listTableLike(subType, parentNamespace, PageToken.readEverything())
```
dimas-b: Can we avoid "list" operations in this case?

Would it be possible to "resolve" the name to be created for any potential previous type and then do the "already exists" check?

dimas-b (commented):
Thanks for working on this, @Subham-KRLX !

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
Subham-KRLX: I replaced the list based check with direct name lookup for table/view under the parent namespace and we still return the same [AlreadyExistsException](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) when a collision is found.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -630,6 +640,18 @@ private static String stripLeadingTrailingSlash(String location) {
...
 
+  private static Namespace tableLikeIdentifierAsNamespace(TableIdentifier identifier) {
+    String[] namespaceLevels =
+        Arrays.copyOf(identifier.namespace().levels(), identifier.namespace().length() + 1);
+    namespaceLevels[identifier.namespace().length()] = identifier.name();
+    return Namespace.of(namespaceLevels);
+  }
+
+  private boolean namespaceWithSameNameExists(TableIdentifier identifier) {
+    return listNamespaces(identifier.namespace()).stream()
```
dimas-b: Actually, can we avoid "list" operations in this case too?

Subham-KRLX: Replaced the list based check with a direct name lookup for namespaces using readEntityByName to avoid the 'list' operation here too Verified the fix with RestCatalogFileIT

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -630,6 +640,40 @@ private static String stripLeadingTrailingSlash(String location) {
...
+    PolarisResolvedPathWrapper resolvedEntities =
+        resolvedEntityView.getResolvedPath(ResolvedPathKey.ofNamespace(identifier.namespace()));
+    if (resolvedEntities == null) {
+      return false;
+    }
+
+    List<PolarisEntity> catalogPath = resolvedEntities.getRawFullPath();
+    EntityResult lookupResult =
+        getMetaStoreManager()
+            .readEntityByName(
```
dimas-b, line 660: Ideally we should do it when we resolve `resolvedEntityView`.

That is, we should add a resolution path with the new entity name for all possible entity types when the `PolarisResolutionManifest` is constructed and do the look up for all of those potential name matches together (via the `Resolver`). WDYT?

Is that feasible?

flyrain:
Thanks @Subham-KRLX for working on it. This a behavior change. It's worth to share it in the dev mailing list for more visibility. Would you mind share it there?

Current behavior in the main branch:
  - Table/View collision prevention, Already enforced on main

What's NEW in this PR (behavior change):
  -  Namespace/Table collision prevention. NEW enforcement
  - Namespace/View collision prevention. NEW enforcement

jbonofre:
I agree with @flyrain here. I think a discussion on the dev@ mailing list is appropriate and welcome.

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
Subham-KRLX: That approach seems very efficient for batching lookups within the Resolver I am planning to refactor the implementation to pre load all potential conflicting entity types (NAMESPACE and TABLE_LIKE) as optional paths in the PolarisResolutionManifest during the CatalogHandler authorization phase. Would this strategy be consistent with what you have in mind? If so then IcebergCatalog can perform collision checks by directly calling resolvedEntityView.getResolvedPath() which eliminates redundant metadata store lookups and ensures all relevant names are resolved together by the Resolver as you suggested.

Subham-KRLX:
> Thanks @Subham-KRLX for working on it. This a behavior change. It's worth to share it in the dev mailing list for more visibility. Would you mind share it there?
> 
> Current behavior in the main branch:
> 
> * Table/View collision prevention, Already enforced on main
> 
> What's NEW in this PR (behavior change):
> 
> * Namespace/Table collision prevention. NEW enforcement
> * Namespace/View collision prevention. NEW enforcement

@flyrain @dimas-b I sent the mailing list proposal regarding Introduce DataSourceResolver for multi-datasource support in JDBC but have not received a response yet should we resume the review here to maintain momentum? Please let me know if there is another communication channel or community meeting where I should present this instead.

dimas-b:
@Subham-KRLX : I believe this change is valuable, but it is not related to your `DataSourceResolver` PR.

My understanding is that other reviewers prefer to have a separate `dev` ML discussion about this PR. I see merit in starting such a discussion because strictly enforcing name uniqueness, while being reasonable from my personal POV, can indeed affect the behaviour of previous deployments (on upgrade).

Would you mind starting this new discussion on `dev`?

Subham-KRLX:
> @Subham-KRLX : I believe this change is valuable, but it is not related to your `DataSourceResolver` PR.
> 
> My understanding is that other reviewers prefer to have a separate `dev` ML discussion about this PR. I see merit in starting such a discussion because strictly enforcing name uniqueness, while being reasonable from my personal POV, can indeed affect the behaviour of previous deployments (on upgrade).
> 
> Would you mind starting this new discussion on `dev`?

@dimas-b would it be okay to start the discussion in the #dev channel on Slack? It might be easier for a quick back-and-forth on the technical approach and 'easier' to maintain momentum once we align there I can post a summary of the outcome to the mailing list for the official record Would that work for you?

dimas-b:
@Subham-KRLX : slack is not "visible" from the ASF process perspective :shrug: Please start an email thread. If a particular topic becomes convoluted we can certainly use slack to resolve it faster, but ultimately all design decisions will have to be carried over to the `dev` email thread.

Alternatively, I can send the initial email if you prefer - please let me know.

Subham-KRLX:
Crisp

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
```diff
@@ -2700,6 +2756,30 @@ private Page<TableIdentifier> listTableLike(
         .map(record -> TableIdentifier.of(parentNamespace, record.getName()));
   }
 
+  private Optional<PolarisEntitySubType> conflictingTableLikeSubtype(
+      List<PolarisEntity> catalogPath, String entityName) {
+    for (PolarisEntitySubType subType :
+        List.of(PolarisEntitySubType.ICEBERG_TABLE, PolarisEntitySubType.ICEBERG_VIEW)) {
```
snazy, line 2762: Do we really need _two_ additional `readEntityByName` invocations against the backend database?

snazy (commented):
I think the PR is on the right track for the behavior we want to enforce.

My concern is with where this invariant is enforced.
The change appears to add additional backend/database reads before creating entities, which follows existing patterns in `IcebergCatalog`.
However, those patterns already have some read-before-write / TOCTOU concerns:
the service checks current state, then relies on a later operation to persist the change.

Should sibling-name uniqueness be enforced as part of the metastore/backend create operation instead?
That would let each backend implement the check atomically and, where possible, more efficiently than adding more preflight reads in `IcebergCatalog`.

WDYT?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java
Subham-KRLX: Backend level enforcement would be ideal for atomicity but the current approach follows the same read before write pattern already used for table ↔ view collision checks in doCommit on main for the two readEntityByName calls can reduce to a single lookup using ANY_SUBTYPE with TABLE_LIKE Let me know if you would prefer that optimization here with backend enforcement as a follow up or the full refactor in this PR.

Subham-KRLX:
Crisp

PR #4198, opened 2026-04-14
feat(helm): add termination and lifecycle options

fivetran-ashokborra:
## Summary

  This PR adds two optional Helm chart settings for Kubernetes termination handling:

  - `terminationGracePeriodSeconds` at the pod spec level
  - `containerLifecycle` for rendering container lifecycle hooks such as `preStop`

  The change is intentionally capability-only and does not change default behavior.

  ## Motivation

  The chart already exposes probe configuration, but it does not currently expose the pod-level termination grace period or container lifecycle hooks.

  These settings are useful for operators who need to coordinate graceful termination with connection draining during rollouts or pod eviction, for example by:

  - allowing more time for in-flight requests to complete
  - delaying container shutdown briefly via `preStop` so endpoint removal can propagate before termination begins

  Without these chart-level knobs, users need to fork or patch the deployment template.

  ## What changed

  - added `terminationGracePeriodSeconds` to `values.yaml`
  - added `containerLifecycle` to `values.yaml`
  - rendered both into `helm/polaris/templates/deployment.yaml`
  - updated `values.schema.json`
  - added Helm unit tests
  - regenerated Helm chart reference docs

  ## Behavior

  - defaults remain unchanged
  - the new settings are optional
  - no runtime behavior changes unless users explicitly configure them

  ## Example

  ```yaml
  terminationGracePeriodSeconds: 330

  containerLifecycle:
    preStop:
      exec:
        command:
          - /bin/sh
          - -c
          - sleep 30
```
  ## Verification

  - make helm-schema-generate
  - make helm-doc-generate
  - make helm-unittest
  - helm lint helm/polaris
  - helm template polaris-release helm/polaris

## Checklist

### DIFF
--- a/helm/polaris/templates/deployment.yaml
+++ b/helm/polaris/templates/deployment.yaml
@@ -72,11 +72,18 @@ spec:
       initContainers:
         {{- tpl (toYaml .Values.extraInitContainers) . | nindent 8 }}
       {{- end }}
+      {{- if .Values.terminationGracePeriodSeconds }}
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           {{- if .Values.containerSecurityContext }}
           securityContext:
             {{- tpl (toYaml .Values.containerSecurityContext) . | nindent 12 }}
+          {{- end }}
+          {{- if .Values.containerLifecycle }}
+          lifecycle:
+            {{- tpl (toYaml .Values.containerLifecycle) . | nindent 12 }}
           {{- end }}
           image: "{{ tpl .Values.image.repository . }}:{{ tpl .Values.image.tag . | default .Chart.Version }}"
           imagePullPolicy: {{ tpl .Values.image.pullPolicy . }}
--- a/helm/polaris/templates/deployment.yaml
+++ b/helm/polaris/templates/deployment.yaml
@@ -72,7 +72,7 @@ spec:
       initContainers:
         {{- tpl (toYaml .Values.extraInitContainers) . | nindent 8 }}
       {{- end }}
-      {{- if .Values.terminationGracePeriodSeconds }}
+      {{- if ne .Values.terminationGracePeriodSeconds nil }}
       terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
       {{- end }}
       containers:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: helm/polaris/templates/deployment.yaml [hunk 1]
```diff
@@ -72,12 +72,19 @@ spec:
       initContainers:
         {{- tpl (toYaml .Values.extraInitContainers) . | nindent 8 }}
       {{- end }}
+      {{- if .Values.terminationGracePeriodSeconds }}
```
adutra: I think we need to account for the case where `terminationGracePeriodSeconds` is set to zero (kill immediately):

```suggestion
      {{- if ne .Values.terminationGracePeriodSeconds nil }}
```

helm/polaris/values.yaml
```diff
@@ -141,6 +141,24 @@ containerSecurityContext:
...
+# Use this to configure a preStop hook for graceful shutdown, e.g.:
+# containerLifecycle:
+#   preStop:
+#     exec:
+#       command: ["/bin/sh", "-c", "sleep 30"]
+# @section -- Pod Configuration
+containerLifecycle: {}
+
+# @schema type: [integer, "null"]
+# @schema minimum: 1
```
```suggestion
# @schema minimum: 0
```

helm/polaris/values.yaml
```diff
@@ -141,6 +141,24 @@ containerSecurityContext:
   seccompProfile:
     type: RuntimeDefault
 
+# @schema type: [object, "null"]
```
adutra: I think just `object` is fine:

```suggestion
# @schema type: object
```

helm/polaris/templates/deployment.yaml
fivetran-ashokborra: That makes sense. Addressed

helm/polaris/values.yaml
```diff
@@ -141,6 +141,24 @@ containerSecurityContext:
...
+# containerLifecycle:
+#   preStop:
+#     exec:
+#       command: ["/bin/sh", "-c", "sleep 30"]
+# @section -- Pod Configuration
+containerLifecycle: {}
+
+# @schema type: [integer, "null"]
+# @schema minimum: 0
+# -- Duration in seconds the pod needs to terminate gracefully. Must be greater than the preStop hook duration.
```
adutra: I just noticed that helm-docs is reporting this field as string. This should fix it:

```suggestion
# -- (int) Duration in seconds the pod needs to terminate gracefully. Must be greater than the preStop hook duration.
```

fivetran-ashokborra: My bad, addressed it

adutra (approved):

dimas-b (approved):

Issue #1545, opened 2025-05-08
Unable to use AWS S3 locations with quotes in path


dimas-b:
### Describe the bug

Polaris fails to accept S3 locations that are valid in AWS, but contain `"` in the path.

For example: `s3://example/pol"3/`

### To Reproduce

1. Start Polaris
2. Try to create catalog:
```
$ ./polaris --client-id *** --client-secret *** catalogs create polaris2 --storage-type S3 --default-base-location 's3://example-bucket-42/pol"2/' --role-arn arn:aws:iam::11111111111:role/test --external-id ***
```

### Actual Behavior

```
Exception when communicating with the Polaris server. IllegalArgumentException: Illegal character in path at index 26: s3://example-bucket-42/pol"2/
```

### Expected Behavior

No exception

### Additional context

The example location is considered valid in AWS UI. It is possible to create it and copy its S3 location in AWS UI.

### System information

Local Polaris Server build as of commit 04c4f9112e450133e1eab2831dcc63f5885e8795

dimas-b:
Cf. https://github.com/projectnessie/nessie/issues/8328

adnanhemani:
This is odd - without adding an external ID, this works as expected. But not when adding the external ID...

```
./polaris --client-id *** --client-secret *** catalogs create polaris2 \
  --storage-type S3 \
  --default-base-location 's3://***/pol"2/' \
  --role-arn arn:aws:iam::***:role/***
```

yields the following logs:

```
2025-05-13 21:17:10,515 INFO  [org.apa.pol.ser.adm.PolarisServiceImpl] [,POLARIS] [,,,] (executor-thread-1) Created new catalog class PolarisCatalog {
    class Catalog {
        type: INTERNAL
        name: polaris2
        properties: class CatalogProperties {
            {default-base-location=s3://***/pol"2/}
            defaultBaseLocation: s3://***/pol"2/
        }
        createTimestamp: 1747196230343
        lastUpdateTimestamp: 1747196230343
        entityVersion: 1
        storageConfigInfo: class AwsStorageConfigInfo {
            class StorageConfigInfo {
                storageType: S3
                allowedLocations: [s3://***/pol"2/]
            }
            roleArn: arn:aws:iam::***:role/***
            externalId: null
            userArn: null
            region: null
        }
    }
}
[...]

Issue #552, opened 2024-12-14
Make Polaris safe against certain unparseable locations


snazy:
### Describe the bug

Using `java.net.URI` for paths/locations generated/used by Apache Iceberg is particularly unsafe.

Characters that are illegal in URI string representations are not escaped by Iceberg and lead to runtime exceptions when parsed for example by `java.net.URI.create(String)`.

Example: A legit column/partition-field name `"_foo_bar_"` will end "as is" in any path generated by Iceberg, leading to ` Caused by: java.lang.IllegalArgumentException: Illegal character in path at index ...`.

The safest approach to prevent this issue entirely is to have a dedicated class that deals with "unsafe encodings", maybe call it "StorageUri".

Mixing "unsafe" and "safe" encodings will cause errors.

flyrain:
Hi @snazy , can you provide more details of which part of Polaris code has this issue?

flyrain:
Related PR: https://github.com/apache/polaris/pull/1586

dimas-b:
More specific issue: #1545 

PR #1586 may not be sufficient for this issue. If it is merged with its current scope more end-to-end testing will still be required with odd base URI to ensure correct operation. Before #1586, end-to-end testing is not possible.

dimas-b:
I think it should be fine to remove this issue from the 1.4.0 milestone. It does not look like users hit these problems often (at least I did not see any user-driven bug reports), but we should keep it open for improving the related code robustness in general.

dimas-b:
@snazy : I removed this issue from the milestones (my rationale is above). Please re-add if you disagree 😅

PR #4210, opened 2026-04-15
Remove double URL-decoding of table identifiers

adutra:
This change fixes several double-decode issues when processing table identifiers in Polaris REST APIs.

The REST framework (Jersey) already URL-decodes path segments. For example, when a REST resource method such as `IcebergRestCatalogApi.loadTable` is invoked, the parameters `namespace` and `table` are already URL-decoded. It is therefore incorrect to further URL-decode table identifiers.

This change does NOT modify namespace decoding: it is actually correct, but inefficient. It will be improved separately.

This change also fixes a few bugs in the test classes `PolicyApi` and `GenericTableApi`: they were incorrectly double-encoding namespace names.

## Checklist

### DIFF
--- a/integration-tests/src/main/java/org/apache/polaris/service/it/env/GenericTableApi.java
+++ b/integration-tests/src/main/java/org/apache/polaris/service/it/env/GenericTableApi.java
@@ -52,7 +52,7 @@ public class GenericTableApi extends PolarisRestApi {
   public List<TableIdentifier> listGenericTables(String catalog, Namespace namespace) {
     String ns = NamespaceUtils.joinNamespace(namespace, NamespaceUtils.DEFAULT_NAMESPACE_SEPARATOR);
     try (Response res =
-        request("polaris/v1/{cat}/namespaces/{ns}/generic-tables", Map.of("cat", catalog, "ns", ns))
+        request("polaris/v1/{cat}/namespaces/" + ns + "/generic-tables", Map.of("cat", catalog))
             .get()) {
       assertThat(res.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
       return res.readEntity(ListGenericTablesResponse.class).getIdentifiers().stream().toList();
@@ -64,8 +64,8 @@ public class GenericTableApi extends PolarisRestApi {
         NamespaceUtils.joinNamespace(id.namespace(), NamespaceUtils.DEFAULT_NAMESPACE_SEPARATOR);
     try (Response res =
         request(
-                "polaris/v1/{cat}/namespaces/{ns}/generic-tables/{table}",
-                Map.of("cat", catalog, "table", id.name(), "ns", ns))
+                "polaris/v1/{cat}/namespaces/" + ns + "/generic-tables/{table}",
+                Map.of("cat", catalog, "table", id.name()))
             .delete()) {
       assertThat(res.getStatus()).isEqualTo(NO_CONTENT.getStatusCode());
     }
@@ -76,8 +76,8 @@ public class GenericTableApi extends PolarisRestApi {
         NamespaceUtils.joinNamespace(id.namespace(), NamespaceUtils.DEFAULT_NAMESPACE_SEPARATOR);
     try (Response res =
         request(
-                "polaris/v1/{cat}/namespaces/{ns}/generic-tables/{table}",
-                Map.of("cat", catalog, "table", id.name(), "ns", ns))
+                "polaris/v1/{cat}/namespaces/" + ns + "/generic-tables/{table}",
+                Map.of("cat", catalog, "table", id.name()))
             .get()) {
       return res.readEntity(LoadGenericTableResponse.class).getTable();
     }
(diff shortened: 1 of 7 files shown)


integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
```diff
@@ -2418,4 +2419,26 @@ public void testNonPaginatedListTablesViewNamespaces() {
     assertThat(nsResponse.namespaces()).hasSize(5);
     assertThat(nsResponse.nextPageToken()).isNull();
   }
+
+  @Test
+  public void testNamespaceAndTableEncodeDecode() {
+    Namespace ns1 = Namespace.of("ns+1%20");
```
dimas-b: Should we also test something like `%q` (an invalid URL escape sequence, but a valid NS name)?

integration-tests/src/main/java/org/apache/polaris/service/it/env/GenericTableApi.java
```diff
@@ -52,7 +52,7 @@ public void purge(String catalog, Namespace ns) {
   public List<TableIdentifier> listGenericTables(String catalog, Namespace namespace) {
     String ns = RESTUtil.encodeNamespace(namespace);
     try (Response res =
-        request("polaris/v1/{cat}/namespaces/{ns}/generic-tables", Map.of("cat", catalog, "ns", ns))
+        request("polaris/v1/{cat}/namespaces/" + ns + "/generic-tables", Map.of("cat", catalog))
```
dimas-b: Does `Namespace.toString()` produce the right value for a URL path segment?

What is the namespace contains a `/`?

adutra: The `ns` variable is a string, not a `Namespace`. The namespace can contain slashes; these will be percent-encoded.

adutra: My other PR #4213 will provide an even better solution for namespace encoding in client/test code.

dimas-b: oops, my mistake 🤦

dimas-b: Still, if #4213 merges first, we won't have to redo the `{ns}` placeholders... right?

adutra: Yes. The two PRs will conflict. It may make sense to merge 4213 first, then remove the test infra changes in this PR.

dimas-b: +1 to #4213 first

integration-tests/src/main/java/org/apache/polaris/service/it/test/PolarisRestCatalogIntegrationBase.java
adutra: That's a very good observation! In fact `%q` in a table or namespace name will not work.

But the reason is not related to a double-decoding, but rather due to the fact that we are computing base locations in many places by simply concatenating strings together without proper escaping.

For example, the namespace base location is computed here:

https://github.com/apache/polaris/blob/e94fdff63852dc41635c9e7eb62b3627ba562b85/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L571

This would create a string containing `%q` unencoded. This string would then be passed to `StorageLocation.of` here:

https://github.com/apache/polaris/blob/e94fdff63852dc41635c9e7eb62b3627ba562b85/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalog.java#L1280

Which would then throw something like this:

```
java.lang.IllegalArgumentException: Malformed escape pair at index 110: file:///var/folders/s4/nm0wnny96t3240rmkw7v1tl00000gn/T/junit-15798120720937239700/path/to/data/ns+1/%20/ns+2/%q/"
```

Unfortunately, fixing this is complex and will take time. It's out of the scope of this PR I'm afraid 😅

dimas-b: Yes... sad, but fair point :shrug:

dimas-b: Cf. https://github.com/apache/iceberg/issues/10279

adutra: And to be fair, I think that many other special characters are being incorrectly mapped to storage locations. 

For example, the namespace `ns+1/%20` generates a storage location ending in `ns+1/%20`, again without encoding – which actually means `ns+1/ ` (with a space)...

Again, fixing this will take time. I wonder if we shouldn't just remove this test for now, as it gives a false sense that things work, when in fact, the API calls succeed, but the underlying storage locations are likely completely messed up.

WDYT?

dimas-b: Right, it does sound like removing this test is probably preferable for now. We can add more coverage under #552 or #1545 perhaps :thinking:

adutra: All tests removed. The PR now doesn't introduce any new test, which is a bit annoying – but the simple fact that the existing tests pass with the fix could be seen as successful test 😄

integration-tests/src/main/java/org/apache/polaris/service/it/env/GenericTableApi.java
adutra: #4213 has been merged and this PR has been rebased (and simplified).

dimas-b (approved):

PR #4213, opened 2026-04-15
Improve namespace decoding

adutra:
This change removes calls to Iceberg `RESTUtil.decodeNamespace()` almost entirely, and replaces those with methods that are more suitable for use in Quarkus/Jersey.

The REST framework (Jersey) already URL-decodes path segments. For example, when a REST resource method such as `IcebergRestCatalogApi.loadTable` is invoked, the parameters `namespace` and `table` are already URL-decoded.

But since `RESTUtil.decodeNamespace()` expects *encoded* inputs, we were forced to re-encode the namespace name before passing it to `decodeNamespace()`.

This change does NOT remove calls to `RESTUtil.decodeNamespace()` in the Polaris entity code – this will be addressed later.

This change also does NOT change calls to `RESTUtil.encodeNamespace()` – these are problematic but will be addressed later as well.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/rest/NamespaceUtils.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/rest/NamespaceUtils.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.polaris.core.rest;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
+import org.apache.iceberg.catalog.Namespace;
+
+/** Utility methods for encoding and decoding {@link Namespace} values. */
+public final class NamespaceUtils {
+
+  /**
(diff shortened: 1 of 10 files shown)


polaris-core/src/main/java/org/apache/polaris/core/rest/NamespaceUtils.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * <p>This is the same separator declared in Iceberg's {@link org.apache.iceberg.rest.RESTUtil}
+   * class, but it's private there and cannot be referenced.
+   */
+  public static final String DEFAULT_NAMESPACE_SEPARATOR = "\u001f";
+
+  private NamespaceUtils() {}
+
+  /** Joins the levels of a namespace into a single string using the given {@code separator}. */
+  public static String joinNamespace(Namespace namespace, String separator) {
```
adutra, line 41: I'm deliberately avoiding to provide an overloaded method without the `separator` argument that would just use the default separator. This is because in Iceberg 1.11 the namespace separator becomes configurable; I'm not sure Polaris would ever allow users to configure the namespace one day, but if this day arrives, it will be the caller's responsibility to figure out which separator to use.

adutra: I'm open to the opposite approach though: if we decide right now that in Polaris, the namespace separator will never change, then we could remove the `separator` parameter.

dimas-b: This makes sense in `polaris-core`, but perhaps we could have a convenience method in test code... just to reduce clutter as we never reconfigure it in tests, IIRC 🤔

dimas-b (approved):

dimas-b:
Nice, fix!.. long overdue :sweat_smile: Thanks, @adutra !

polaris-core/src/main/java/org/apache/polaris/core/rest/NamespaceUtils.java
adutra: Let's defer this enhancement to when we'll have to deal with configurable separators. My gut feeling is that the test infra code will also have to deal with configurable separators by then, in which case, I'm not sure the convenience method would make sense.

adutra:
I'm going to merge this by EOD to unblock further changes to encoding/decoding logic, FYI.

polaris-core/src/main/java/org/apache/polaris/core/rest/NamespaceUtils.java
dimas-b: SGTM :+1:

PR #4214, opened 2026-04-15
Isolate encode/decode logic used in Polaris entities

adutra:
This change replaces calls to `RESTUtil` in Polaris entity code when serializing or deserializing namespaces stored in internal properties.

It replaces those calls with calls to a new utility class, `PolarisEntityUtils`. The logic used there for encoding and decoding namespaces is in fact identical to `RESTUtil` as of today, but having our own logic protects Polaris from future behavioral changes in `RESTUtil` that could potentially cause data corruption in Polaris metastores.

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityUtils.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityUtils.java
@@ -18,16 +18,30 @@
  */
 package org.apache.polaris.core.entity;
 
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.rest.RESTUtil;
 
+/**
+ * Polaris-owned encoding/decoding for namespaces stored in entity internal properties.
+ *
+ * <p>This logic is isolated from {@link RESTUtil} to ensure the storage format remains stable even
+ * if Iceberg's REST utilities change behavior in future versions, preventing potential data
+ * corruption in Polaris metastores.
+ */
 public final class PolarisEntityUtils {
 
   private static final String NAMESPACE_SEPARATOR_ENCODED = "%1F";
+
+  private static final Joiner NAMESPACE_JOINER = Joiner.on(NAMESPACE_SEPARATOR_ENCODED);
+
+  private static final Splitter NAMESPACE_SPLITTER = Splitter.on(NAMESPACE_SEPARATOR_ENCODED);
 
   private PolarisEntityUtils() {}
 
@@ -44,7 +58,7 @@ public final class PolarisEntityUtils {
     for (int i = 0; i < levels.length; i++) {
       encodedLevels[i] = URLEncoder.encode(levels[i], StandardCharsets.UTF_8);
     }
-    return String.join(NAMESPACE_SEPARATOR_ENCODED, encodedLevels);
+    return NAMESPACE_JOINER.join(encodedLevels);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


polaris-core/src/test/java/org/apache/polaris/core/entity/PolarisEntityUtilsTest.java
```diff
@@ -0,0 +1,115 @@
...
+        Arguments.of(Namespace.of("a", "b", "c"), "a%1Fb%1Fc"),
+        Arguments.of(Namespace.of("a/b", "c"), "a%2Fb%1Fc"),
+        // empty level in a multi-level namespace
+        Arguments.of(Namespace.of("a", "", "b"), "a%1F%1Fb"));
+  }
+
+  @ParameterizedTest
+  @MethodSource("encodeNamespaceCases")
+  public void testEncodeNamespace(Namespace ns, String expected) {
+    Assertions.assertThat(PolarisEntityUtils.encodeNamespace(ns)).isEqualTo(expected);
```
dimas-b: Could we add an assertion that the result matches `RESTUtil.encodeNamespace`?

For now it will ensure we're not breaking existing logic. If `RESTUtil` becomes incompatible later, we will remove those assertions.

dimas-b (commented):
+1 to controlling encoding/decoding in Polaris code.

polaris-core/src/test/java/org/apache/polaris/core/entity/PolarisEntityUtilsTest.java
dimas-b: Actually, what I proposed above probably makes more sense for `decodeNamespace` :sweat_smile:

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityUtils.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.apache.polaris.core.entity;
+
+import com.google.common.base.Preconditions;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.rest.RESTUtil;
+
```
flyrain, line 30: Can we add a Java doc, something like 
```
  /**
   * Polaris-owned encoding/decoding for namespaces stored in entity internal properties.
   *
   * <p>This logic is isolated from iceberg.rest.RESTUtil to ensure the storage format
   * remains stable even if Iceberg's REST utilities change behavior in future versions,
   * preventing potential data corruption in Polaris metastores.
   */
```

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityUtils.java
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Returns a Namespace object from a String representation that was encoded using {@link
+   * #encodeNamespace(Namespace)}.
+   *
+   * <p>This method is similar to {@link RESTUtil#decodeNamespace(String)}.
+   */
+  public static Namespace decodeNamespace(String encodedNs) {
+    Preconditions.checkArgument(encodedNs != null, "Invalid namespace: null");
+    String[] levels = encodedNs.split(NAMESPACE_SEPARATOR_ENCODED, -1);
```
flyrain: The RESTUtil used a guava splitter`com.google.common.base.Splitter`. I think we can validate if Java String splitter behaviors a bit different. I mighty be paranoid. There could be some corner cases causing decoding behavior drift.

nandorKollar:
We use `RESTUtil.encodeNamespace` also in other places, like for example in `PolarisResourcePaths`. May I ask why do we replace that call only in the entites?

jbonofre (dismissed):
LGTM.

Maybe only @flyrain's comment about the javadoc link is worth to consider (about Iceberg `RestUtils`).

polaris-core/src/main/java/org/apache/polaris/core/entity/PolarisEntityUtils.java
adutra: That's a good point, I will switch to using `Joiner` and `Splitter`.

adutra:
> We use `RESTUtil.encodeNamespace` also in other places, like for example in `PolarisResourcePaths`. May I ask why do we replace that call only in the entites?

I will tackle this later. It will depend on https://github.com/apache/iceberg/pull/15989.

nandorKollar (approved):

dimas-b (approved):

PR #4097, opened 2026-03-31
Harden GitHub Actions workflows

adutra:
This PR strengthens our CI jobs:

- Prevent template injection by moving `${{ }}` expressions out of run blocks
- Prevent "ArtiPACKED"-like exploits by avoiding persisting credentials
- Fix excessive permissions wherever possible
- Add new job to run Zizmor checks on PRs that modify workflows

Release workflows that push branches/tags (release-1, release-2, release-4) intentionally retain persisted credentials.

## Checklist
- [ ] 🛡️ Don't disclose security issues! (contact security@apache.org)
- [ ] 🔗 Clearly explained why the changes are needed, or linked related issues: Fixes #
- [ ] 🧪 Added/updated tests with good coverage, or manually tested (and explained how)
- [ ] 💡 Added comments for complex logic
- [ ] 🧾 Updated `CHANGELOG.md` (if needed)
- [ ] 📚 Updated documentation in `site/content/in-dev/unreleased` (if needed)

PR #4216, opened 2026-04-15
ci: apply github workflow best practice (zizmor/codeql/asf-allowlist-check)

kevinjqliu:
Follow up to #4097

Based on a few more recent improvements made in the apache/iceberg repo


.github/workflows/ci-asf-allowlist-check.yml
```diff
@@ -43,5 +43,4 @@ jobs:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       with:
         persist-credentials: false
-    # Intentionally unpinned to always use the latest allowlist from the ASF.
-    - uses: apache/infrastructure-actions/allowlist-check@main # zizmor: ignore[unpinned-uses]
+    - uses: apache/infrastructure-actions/allowlist-check@4e9c961f587f72b170874b6f5cd4ac15f7f26eb8  # main
```
kevinjqliu, line 46: made improvements to `apache/infrastructure-actions/allowlist-check` so that we no longer need to pin to `@main`

.github/workflows/ci-codeql.yml
```diff
@@ -46,11 +46,11 @@ jobs:
         persist-credentials: false
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@95e58e9a2cdfd71adc6e0353d5c52f41a045d225 # v4
+      uses: github/codeql-action/init@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
```
kevinjqliu: apparently `v4` tag is rolling and can potential fail CI (see https://github.com/apache/iceberg/pull/15985) 
lets pin to `v4.35.1` which should be immutable

.github/workflows/ci-zizmor.yml
```diff
@@ -20,41 +20,25 @@
...
-      - name: Run Zizmor (SARIF upload)
-        uses: zizmorcore/zizmor-action@71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 # v0.5.2
-        with:
-          min-severity: medium
-          min-confidence: medium
-      - name: Run Zizmor (PR annotations)
+
+      - name: Run zizmor 🌈
         uses: zizmorcore/zizmor-action@71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 # v0.5.2
         with:
```
kevinjqliu: making this change so that zizmor ci check will fail when encountering an issue. otherwise, it will pass and add to the "Security" tab in github

See https://github.com/apache/iceberg/pull/15820

kevinjqliu (commented):
@adutra could you take a look?

flyrain (dismissed):
LGTM. Thanks @kevinjqliu ! We will need a rebase.

kevinjqliu:
rebased, thanks!

.github/workflows/ci-zizmor.yml
adutra: But that's why I included two steps: one with advanced security on, the other with it off. Wasn't that working as expected?

kevinjqliu: oh good point, i missed that. 

it might not be necessary to run both, the one with `advanced-security: false` will fail CI. then there's no need to upload SARIF. wdyt?

.github/workflows/ci-codeql.yml
snazy: This is effectively a _downgrade_ from v4.35.2 to v4.35.1.
What matters is the Git SHA - not the comment.

snazy: BTW: All actions in all versions from `github.com/apache`, `github.com/github`, `github.com/actions` are whitelisted.

.github/workflows/ci-codeql.yml
```diff
@@ -47,11 +47,11 @@ jobs:
...
       with:
         languages: actions
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@95e58e9a2cdfd71adc6e0353d5c52f41a045d225 # v4
+      uses: github/codeql-action/analyze@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
```
snazy: Same here: _downgrade_ from v4.35.2

.github/workflows/ci-asf-allowlist-check.yml
snazy: Do you know how Renovate will handle this one?

.github/workflows/ci-zizmor.yml
```diff
@@ -20,41 +20,25 @@
...
-      - name: Run Zizmor (SARIF upload)
-        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
-        with:
-          min-severity: medium
-          min-confidence: medium
-      - name: Run Zizmor (PR annotations)
-        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 # v0.5.2
```
snazy: Why this downgrade?

.github/workflows/ci-zizmor.yml
adutra: Well, I'd rather have advanced security on, tbh, since that's the recommended way to use the action.

If we are concerned by a PR that could succeed while it contains security warnings, the [documentation](https://github.com/zizmorcore/zizmor-action?tab=readme-ov-file) page says:

> Continue to use zizmor-action with advanced-security: true, but configure a [ruleset](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets#set-code-scanning-merge-protection) to prevent PRs from merging until all code scanning alerts are resolved. This is the recommended approach, but you must configure it manually — zizmor-action cannot do it for you.

I cannot create rulesets though, but it seems to me it's just a matter of having someone create a ruleset like [this one](https://docs.github.com/en/code-security/how-tos/find-and-fix-code-vulnerabilities/manage-your-configuration/set-code-scanning-merge-protection).

Until we have this sorted out, I figured that having two steps (one with advanced security on, the other without) would bring the best of both worlds.

.github/workflows/ci-asf-allowlist-check.yml
kevinjqliu: i dont, i havent used renovate before, but this github action shouldnt change much https://github.com/apache/infrastructure-actions/tree/main/allowlist-check

.github/workflows/ci-codeql.yml
kevinjqliu: oops, this was a merge conflict, the main change is to use the longer form

.github/workflows/ci-codeql.yml
```diff
@@ -47,11 +47,11 @@ jobs:
         persist-credentials: false
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@95e58e9a2cdfd71adc6e0353d5c52f41a045d225 # v4
+      uses: github/codeql-action/init@95e58e9a2cdfd71adc6e0353d5c52f41a045d225 # v4.35.2
```
kevinjqliu, line 50: https://github.com/github/codeql-action/commit/95e58e9a2cdfd71adc6e0353d5c52f41a045d225

.github/workflows/ci-zizmor.yml
kevinjqliu: I reverted the changes to zizmor in this PR. 
In the iceberg repo, the behavior is to block in CI which is the behavior when `advanced-security: false`

The current implementation will both block in CI (the `PR annotations` part) and report to the Security tab (`SARIF upload`). My point was that we dont need to do both. If its already blocked in CI, there's not really a need to also report it to the Security tab

kevinjqliu:
kept the changes minimal, removed the zizmor changes. ptal!

flyrain (approved):

.github/workflows/ci-asf-allowlist-check.yml
snazy: Hm, the question is: who or what cares about updating the action? I suspect, Git tags/releases would be difficult, because that would either tie the apache/infrastructure-actions repo to only one action or all actions happen on the same release schedule?

.github/workflows/ci-codeql.yml
snazy: Not sure this comment-only change is worth it. Renovate _will_ overwrite it anyways.

kevinjqliu: sure, renovate will update the commit hash and the comment accordingly. 

theres a potential bug here where the v4 tag gets remapped to another hash, then CI will fail. This is what happened in the iceberg repo. 

the current hash is map to https://github.com/github/codeql-action/commit/95e58e9a2cdfd71adc6e0353d5c52f41a045d225 for both v4 and v4.35.2

.github/workflows/ci-asf-allowlist-check.yml
kevinjqliu: this prevents any updates from silently failing in github ci 
I wrote up some background in  https://github.com/apache/infrastructure-actions/issues/574

kevinjqliu:
oh looks like this was done in https://github.com/apache/polaris/pull/4224 ...

PR #4220, opened 2026-04-16
Add PyPI Publishing to Automated Release

adnanhemani:
Changes the PyPI nightly build script to use Trusted Publishers, as well as adding publishing scripts when generating a new release candidate and publishing new Polaris versions. 

I've tested this from my fork: https://github.com/adnanhemani/polaris/actions/runs/24500027607/

I've already set up Trusted Publishers for both PyPI and TestPyPI:

Screenshot 2026-04-16 at 1 06 56 AM

Screenshot 2026-04-16 at 1 07 18 AM

Result:

Screenshot 2026-04-16 at 1 27 18 AM

## Checklist
- [X] 🛡️ Don't disclose security issues!

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -99,7 +99,7 @@ jobs:
           make client-nightly-build
 
       - name: Publish Python client to Test PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
         with:
           repository-url: https://test.pypi.org/legacy/
           packages-dir: client/python/dist/
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -77,6 +77,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'apache/polaris'
     permissions:
+      # IMPORTANT: this permission is mandatory for Trusted Publishing to PyPI
       id-token: write
     steps:
       - name: Checkout
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -76,6 +76,8 @@ jobs:
   nightly_build_python_client:
     runs-on: ubuntu-latest
     if: github.repository == 'apache/polaris'
+    permissions:
+      id-token: write
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -93,8 +95,13 @@ jobs:
           distribution: 'temurin'
           java-version: '21'
 
-      - name: Publish Python client to Test PyPI
-        env:
-          UV_PUBLISH_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }} # zizmor: ignore[secrets-outside-env]
+      - name: Build Python client nightly
         run: |
-          make client-nightly-publish
+          make client-nightly-build
+
+      - name: Publish Python client to Test PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+          packages-dir: client/python/dist/
+          skip-existing: true
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -75,7 +75,6 @@ jobs:
 
   nightly_build_python_client:
     runs-on: ubuntu-latest
-    if: github.repository == 'apache/polaris'
     permissions:
       id-token: write
     steps:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -75,6 +75,7 @@ jobs:
 
   nightly_build_python_client:
     runs-on: ubuntu-latest
+    if: github.repository == 'apache/polaris'
     permissions:
       id-token: write
     steps:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -99,7 +99,7 @@ jobs:
           make client-nightly-build
 
       - name: Publish Python client to Test PyPI
-        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
+        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
         with:
           repository-url: https://test.pypi.org/legacy/
           packages-dir: client/python/dist/
--- a/.github/workflows/release-3-build-and-publish-artifacts.yml
+++ b/.github/workflows/release-3-build-and-publish-artifacts.yml
@@ -529,6 +529,7 @@ jobs:
     needs: [prerequisite-checks]
     permissions:
       contents: read
+      # IMPORTANT: this permission is mandatory for Trusted Publishing to PyPI
       id-token: write
     env:
       DRY_RUN: ${{ needs.prerequisite-checks.outputs.dry_run }}
@@ -578,7 +579,12 @@ jobs:
   generate-release-email:
     name: Generate Release Email Body
     runs-on: ubuntu-latest
-    needs: [prerequisite-checks, build-and-publish-artifacts, build-docker, build-and-stage-helm-chart, publish-python-client-rc]
+    needs:
+      - prerequisite-checks
+      - build-and-publish-artifacts
+      - build-docker
+      - build-and-stage-helm-chart
+      - publish-python-client-rc
     permissions:
       contents: read
     env:
--- a/.github/workflows/release-3-build-and-publish-artifacts.yml
+++ b/.github/workflows/release-3-build-and-publish-artifacts.yml
@@ -523,10 +523,55 @@ jobs:
           | Artifact Hub metadata | ✅ Committed to dist dev |
           EOT
 
+  publish-python-client-rc:
+    name: Publish Python Client RC to PyPI
+    runs-on: ubuntu-latest
+    needs: [prerequisite-checks]
+    permissions:
+      contents: read
+    env:
+      DRY_RUN: ${{ needs.prerequisite-checks.outputs.dry_run }}
+      version_without_rc: ${{ needs.prerequisite-checks.outputs.version_without_rc }}
+      rc_number: ${{ needs.prerequisite-checks.outputs.rc_number }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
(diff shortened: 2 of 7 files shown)


.github/workflows/nightly.yml
```diff
@@ -76,6 +76,8 @@ jobs:
   nightly_build_python_client:
     runs-on: ubuntu-latest
     if: github.repository == 'apache/polaris'
+    permissions:
+      id-token: write
```
snazy, line 80: This permission isn't commonly known, please add a comment explaining what it's used for.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -523,10 +523,62 @@ jobs:
           | Artifact Hub metadata | ✅ Committed to dist dev |
           EOT
 
+  publish-python-client-rc:
+    name: Publish Python Client RC to PyPI
+    runs-on: ubuntu-latest
+    needs: [prerequisite-checks]
+    permissions:
+      contents: read
+      id-token: write
```
snazy, line 532: Same here, deserves a comment.

.github/workflows/release-3-build-and-publish-artifacts.yml
```diff
@@ -523,10 +523,62 @@ jobs:
...
+          | --- | --- |
+          | Version | \`${version_without_rc}rc${rc_number}\` |
+          | Registry | Test PyPI |
+          EOT
+
   generate-release-email:
     name: Generate Release Email Body
     runs-on: ubuntu-latest
-    needs: [prerequisite-checks, build-and-publish-artifacts, build-docker, build-and-stage-helm-chart]
+    needs: [prerequisite-checks, build-and-publish-artifacts, build-docker, build-and-stage-helm-chart, publish-python-client-rc]
```
snazy, line 581: Nit: the list is growing, maybe migrate to one line per dependency.

.github/workflows/release-4-publish-release.yml
```diff
@@ -38,6 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      id-token: write
```
snazy, line 41: This one concerns me, because _all_ steps now get access to OIDC/trusted-publishing.

snazy (commented):
Thanks for the effort here!

I cannot judge on the Python/Makefile specific changes.
The changes to the `nightly.yml` + `release-3-...` workflows look fine.

I do have two concerns about `release-4-...`:

The OIDC (`id-token: write`) permission for PyPi trusted publishing is exposed to all other steps that don't need it.

If, for example, everything but the PyPi publishing fails, it cannot be retried, and we have to re-roll a whole new release including all the vote-"ceremony". Not really new, as Docker-Hub publishing can also fail. I think it's better to have separate workflow-jobs for Maven publishing, Docker-Hub publishing and PyPi publising plus an "aggregator" job that depends on those. This gives us the ability to retry only a failed publishing job without failing a whole release.

(Preexisting) The `contents: write` permission is only needed to create the GH release. That permission can then also be limited to that "aggregator-job".

Makefile
```diff
@@ -188,18 +188,31 @@ client-lint: client-setup-env ## Run linting checks for Polaris client
...
-	DATE_SUFFIX=$$(date -u +%Y%m%d%H%M%S) && \
-	NIGHTLY_VERSION="$${CURRENT_VERSION}.dev$${DATE_SUFFIX}" && \
-	echo "Publishing nightly version: $${NIGHTLY_VERSION}" && \
-	uv version "$${NIGHTLY_VERSION}" && \
-	uv build --clear && \
-	uv publish --index testpypi
-	@echo "--- Nightly publish complete ---"
+	echo "Building version: $(BUILD_VERSION)" && \
+	uv version "$(BUILD_VERSION)" && \
+	uv build --clear
```
HonahX, line 197: Instead of making an internal one, could we re-use client-build and extend it to optionally set the version?
Currently, CLI's License & Notice are only valid if we only publish the wheel (it will be a universal wheel): https://github.com/apache/polaris/pull/3891#discussion_r2881190870. So we probably need `FORMAT=wheel` when building artifacts

HonahX (commented):
@adnanhemani Thank you so much for adding this to our release process!

.github/workflows/nightly.yml
kevinjqliu, line 80:
```suggestion
    permissions:
      # IMPORTANT: this permission is mandatory for Trusted Publishing
      id-token: write
```
https://docs.pypi.org/trusted-publishers/using-a-publisher/

.github/workflows/release-4-publish-release.yml
kevinjqliu: +1 we can move this to the step that pushes to pypi

kevinjqliu (commented):
looks great, thanks! 

for pyiceberg, we push RCs as pre-release to pypi. we should double check that its indeed labeled as pre-release. Otherwise, if it accidentally gets label as the normal release, we'd have to yank the entire version (and publish a patch release). very tedious!

.github/workflows/nightly.yml
adnanhemani: Done!

adnanhemani: Done!

.github/workflows/release-3-build-and-publish-artifacts.yml
adnanhemani: Done.

.github/workflows/release-3-build-and-publish-artifacts.yml
adnanhemani: Done.

.github/workflows/release-4-publish-release.yml
adnanhemani: Moved.

Makefile
adnanhemani: Good call, changed.

adnanhemani (commented):
Thank you all for your reviews. At this time, I've only separated out the "publish-python-client" step in the `release-4...` script, as it is the smallest change relevant to this PR. We should continue the refactor as @snazy suggests in a different PR.

adnanhemani:
@kevinjqliu - we are pushing RCs to TestPyPI and only pushing the released versions to PyPI. I believe this should be good enough to not cause errant versions?

.github/workflows/nightly.yml
```diff
@@ -93,8 +96,13 @@ jobs:
           distribution: 'temurin'
           java-version: '21'
 
-      - name: Publish Python client to Test PyPI
-        env:
-          UV_PUBLISH_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }} # zizmor: ignore[secrets-outside-env]
+      - name: Build Python client nightly
         run: |
-          make client-nightly-publish
+          make client-nightly-build
```
kevinjqliu, line 101: i tried running this locally, got an error

```
uv and dependencies installed.
--- Building client distribution ---
error: after parsing `1.4.0.dev20260417172537`, found `.dev20260417172701`, which is not part of a valid version
make[1]: *** [client-build] Error 2
make: *** [client-nightly-build] Error 2
```

Makefile
```diff
@@ -127,18 +127,21 @@ client-install-dependencies: $(VENV_DIR)
...
+		$(ACTIVATE_AND_CD) && uv version "$(VERSION)"; \
+	fi
 	@if [ -n "$(FORMAT)" ]; then \
 		if [ "$(FORMAT)" != "sdist" ] && [ "$(FORMAT)" != "wheel" ]; then \
 			echo "Error: Invalid format '$(FORMAT)'. Supported formats are 'sdist' and 'wheel'." >&2; \
 			exit 1; \
 		fi; \
 		echo "Building with format: $(FORMAT)"; \
-		$(ACTIVATE_AND_CD) && uv build --format $(FORMAT); \
+		$(ACTIVATE_AND_CD) && uv build --clear --format $(FORMAT); \
```
kevinjqliu:
```suggestion
		$(ACTIVATE_AND_CD) && uv build --clear --$(FORMAT); \
```

kevinjqliu (approved):
Minor bug when passing the `format` arg, otherwise LGTM

Applied the fix locally and tested: 
```
make client-nightly-build
RC_VERSION=1.4.0 RC_NUMBER=1 make client-rc-build                                                                    
RELEASE_VERSION=1.4.0 make client-release-build
```

.github/workflows/nightly.yml
adnanhemani: Resolved as per comment below

snazy:
The [recent Polaris 1.4.0 release workflow failed quite early](https://github.com/apache/polaris/actions/runs/24698069238/job/72238085638) during the changes to Apache SVN, before it handles Helm charts, Docker images or Nexus or even create the GitHub release. I think, Helm charts, Docker images and Nexus jobs would have worked fine during that workflow run, if those would have been separate _jobs_ in the workflow.

The unsolvable issue with release-related workflows is that those steps cannot be integration tested. This is why I think splitting the work across multiple jobs reduces the amount of non-executed steps, that require manual work later.

Adding PyPi publishing adds another step that can potentially fail a whole release, but having that as a separate workflow _job_ reduces the risk of affecting other parts of the release automation.

adnanhemani:
@snazy I agree with you - that's why I separated the PyPI uploads into separate Jobs. Did I miss something there?

I also don't disagree with your comments that we should separate all these steps into different jobs - but I don't find that necessary in _this_ PR. I would highly prefer to do that in a follow up PR to reduce the scope of this PR. Is this a blocker for you?

Makefile
```diff
@@ -188,18 +191,22 @@ client-lint: client-setup-env ## Run linting checks for Polaris client
...
+client-nightly-build: client-setup-env ## Build nightly version for publishing to Test PyPI
+	@$(MAKE) client-build \
+		VERSION="$$($(VENV_DIR)/bin/uv --directory $(PYTHON_CLIENT_DIR) version --short).dev$$(date -u +%Y%m%d%H%M%S)" \
+		FORMAT=wheel
+
+.PHONY: client-rc-build
+client-rc-build: ## Build RC version for publishing to Test PyPI (requires RC_VERSION and RC_NUMBER)
+	@if [ -z "$(RC_VERSION)" ]; then echo "ERROR: RC_VERSION is not set"; exit 1; fi
+	@if [ -z "$(RC_NUMBER)" ]; then echo "ERROR: RC_NUMBER is not set"; exit 1; fi
+	@$(MAKE) client-build VERSION="$(RC_VERSION)rc$(RC_NUMBER)"
```
HonahX: Per offline discussion on the LIcense issue, how about let's put `FORMAT=wheel` in both rc-build and release-build for now?

adnanhemani: Done!

HonahX (approved):
LGTM! Only the wheel is currently added to RC and it's with proper version and signature/checksum

adnanhemani:
Leaving this PR open for 24hr more for any final feedback before I merge.

PR #4226, opened 2026-04-16
fix: enforce single expiration timestamp per storage credential bundle

yushesp:
`StorageAccessConfig.Builder.put()` silently overwrites `expiresAt` when multiple expiration timestamp properties are set. Only a single expiration timestamp is valid per credential bundle. Multiple timestamps indicate incorrect behavior that should fail fast rather than silently picking one. 

This mirrors the same fix applied to `ConnectionCredentials.of()` in https://github.com/apache/polaris/pull/4173.

Also fixes `AwsCredentialsStorageIntegration` which was the only production caller setting two expiration timestamps (`EXPIRATION_TIME` and `AWS_SESSION_TOKEN_EXPIRES_AT_MS`).

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessConfig.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessConfig.java
@@ -58,7 +58,7 @@ public interface StorageAccessConfig {
   }
 
   static StorageAccessConfig.Builder builder() {
-    return ImmutableStorageAccessConfig.builder();
+    return new ValidatingBuilder(ImmutableStorageAccessConfig.builder());
   }
 
   interface Builder {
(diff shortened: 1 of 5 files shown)


yushesp:
hey @dimas-b and @flyrain, put together a similar change for `StorageAccessConfig` as suggested in https://github.com/apache/polaris/pull/4173. It was a bit more involved due to the "builder" pattern used here, open to a different approach if needed!

polaris-core/src/main/java/org/apache/polaris/core/storage/StorageAccessConfig.java
```diff
@@ -91,4 +91,76 @@ default Builder put(StorageAccessProperty key, String value) {
...
     StorageAccessConfig build();
   }
+
+  /**
+   * A decorator around {@link ImmutableStorageAccessConfig.Builder} that enforces at most one
+   * expiration timestamp property per {@link StorageAccessConfig}. Without this guard, multiple
+   * calls to {@link Builder#put} with different expiration timestamp properties silently overwrite
+   * {@link StorageAccessConfig#expiresAt()}.
+   */
+  class ValidatingBuilder implements Builder {
```
dimas-b, line 101: Can `@Value.Check` work?

Cf. https://github.com/apache/polaris/blob/b03bc2274ff96aca3cc184cf407aab1385cae185/polaris-core/src/main/java/org/apache/polaris/core/storage/PolarisStorageConfigurationInfo.java#L73

polaris-core/src/test/java/org/apache/polaris/core/storage/StorageAccessConfigTest.java
```diff
@@ -79,15 +80,38 @@ public void testNoCredentialOverride() {
...
-    assertThat(b.build().expiresAt()).hasValue(Instant.ofEpochMilli(222));
-    b.put(EXPIRATION_TIME, "333");
-    assertThat(b.build().expiresAt()).hasValue(Instant.ofEpochMilli(333));
-    b.expiresAt(Instant.ofEpochMilli(444));
-    assertThat(b.build().expiresAt()).hasValue(Instant.ofEpochMilli(444));
+    assertThat(StorageAccessConfig.builder().build().expiresAt()).isEmpty();
+
+    assertThat(
+            StorageAccessConfig.builder()
+                .put(GCS_ACCESS_TOKEN_EXPIRES_AT, "111")
```
dimas-b, line 87: Maybe use `@ParameterizedTest` and also add coverage for `AZURE_SAS_TOKEN_EXPIRES_AT_MS_PREFIX`?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -159,8 +159,6 @@ public StorageAccessConfig getSubscopedCreds(
       Optional.ofNullable(response.credentials().expiration())
           .ifPresent(
               i -> {
-                accessConfig.put(
-                    StorageAccessProperty.EXPIRATION_TIME, String.valueOf(i.toEpochMilli()));
```
dimas-b, line 163: Is the `EXPIRATION_TIME` property used anymore? Let's deprecate it.

dimas-b: Please also mention it in CHANGELOG that we're no longer emitting this property.

AFAIK, it is not used by clients, but let's also discuss this on `dev` to double check.

dimas-b (commented):
Thanks for working on this, @yushesp ! The PR LGTM overall, some comment below, and I believe we also need a `dev` ML discussion because it involves configuration sent to REST API clients.

PR #4228, opened 2026-04-16
fix: reduce AWS STS inline policy size by merging read+write actions

### LINKED ISSUE #3243: Polaris fails to create a Table because of an STS policy size limit

yushesp:
Writable locations previously appeared in two separate statements:
1. Read statement covering every location
2. Write statement covering only writable ones.

This listed the resource ARN twice for every writable location, which pushed deeply nested namespace paths past the 2048-character STS policy limit.

This change emits a single combined statement granting both read and write actions for writable locations, and keep the read-only statement only for locations not already covered.

Also wraps the assumeRole call to rewrap `PackedPolicyTooLargeException` with a descriptive error msg.

Example:

Before (2 statements, ARN duplicated):
```
  {"Action": ["s3:GetObject", "s3:GetObjectVersion"],
   "Resource": ["...bucket/path/*"]}
  {"Action": ["s3:PutObject", "s3:DeleteObject"],
   "Resource": ["...bucket/path/*"]}
```

After (1 statement):
```
  {"Action": ["s3:GetObject", "s3:GetObjectVersion",
              "s3:PutObject", "s3:DeleteObject"],
   "Resource": ["...bucket/path/*"]}
```

Fixes #3243

## Checklist

### DIFF
--- a/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
+++ b/polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
@@ -25,6 +25,7 @@ import jakarta.annotation.Nonnull;
 import java.net.URI;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -51,6 +52,7 @@ import software.amazon.awssdk.policybuilder.iam.IamStatement;
 import software.amazon.awssdk.services.sts.StsClient;
 import software.amazon.awssdk.services.sts.model.AssumeRoleRequest;
 import software.amazon.awssdk.services.sts.model.AssumeRoleResponse;
+import software.amazon.awssdk.services.sts.model.PackedPolicyTooLargeException;
 import software.amazon.awssdk.services.sts.model.Tag;
 
 /** Credential vendor that supports generating */
@@ -151,7 +153,17 @@ public class AwsCredentialsStorageIntegration
       StsClient stsClient =
           stsClientProvider.stsClient(StsDestination.of(storageConfig.getStsEndpointUri(), region));
 
-      AssumeRoleResponse response = stsClient.assumeRole(request.build());
+      AssumeRoleResponse response;
+      try {
+        response = stsClient.assumeRole(request.build());
+      } catch (PackedPolicyTooLargeException e) {
+        throw new RuntimeException(
+            "AWS STS rejected the session policy: it exceeds the 2048-character packed-policy"
+                + " limit. This commonly happens with deeply nested namespace hierarchies or"
+                + " long namespace names. Consider shortening the namespace path or splitting"
+                + " the hierarchy.",
+            e);
+      }
       accessConfig.put(StorageAccessProperty.AWS_KEY_ID, response.credentials().accessKeyId());
       accessConfig.put(
           StorageAccessProperty.AWS_SECRET_KEY, response.credentials().secretAccessKey());
@@ -208,11 +220,16 @@ public class AwsCredentialsStorageIntegration
   }
 
   /**
-   * generate an IamPolicy from the input readLocations and writeLocations, optionally with list
+   * Generate an IamPolicy from the input readLocations and writeLocations, optionally with list
    * support. Credentials will be scoped to exactly the resources provided. If read and write
    * locations are empty, a non-empty policy will be generated that grants GetObject and optionally
    * ListBucket privileges with no resources. This prevents us from sending an empty policy to AWS
    * and just assuming the role with full privileges.
+   *
+   * <p>Write locations are emitted in a single combined statement that grants both read and write
+   * actions, rather than in a separate write-only statement that duplicates the resource ARN
+   * already listed in the read statement. This keeps the packed inline policy within AWS STS's
+   * 2048-character limit for deeply nested namespace hierarchies.
    */
   private IamPolicy policyString(
       AwsStorageConfigurationInfo storageConfigurationInfo,
@@ -221,25 +238,24 @@ public class AwsCredentialsStorageIntegration
       Set<String> writeLocations,
       String region) {
     IamPolicy.Builder policyBuilder = IamPolicy.builder();
-    IamStatement.Builder allowGetObjectStatementBuilder =
-        IamStatement.builder()
-            .effect(IamEffect.ALLOW)
-            .addAction("s3:GetObject")
-            .addAction("s3:GetObjectVersion");
-    Map<String, IamStatement.Builder> bucketListStatementBuilder = new HashMap<>();
-    Map<String, IamStatement.Builder> bucketGetLocationStatementBuilder = new HashMap<>();
-
     String arnPrefix = arnPrefixForPartition(storageConfigurationInfo.getAwsPartition());
     String currentKmsKey = storageConfigurationInfo.getCurrentKmsKey();
     List<String> allowedKmsKeys = storageConfigurationInfo.getAllowedKmsKeys();
+
+    // Locations readable but not writable receive only read actions; everything in writeLocations
+    // receives read and write actions from a single combined statement below. Splitting this way
+    // avoids listing every write-location ARN twice.
+    Set<String> readOnlyLocations = new HashSet<>(readLocations);
+    readOnlyLocations.removeAll(writeLocations);
+    boolean canWrite = !writeLocations.isEmpty();
+
+    Map<String, IamStatement.Builder> bucketListStatementBuilder = new HashMap<>();
+    Map<String, IamStatement.Builder> bucketGetLocationStatementBuilder = new HashMap<>();
     Stream.concat(readLocations.stream(), writeLocations.stream())
         .distinct()
         .forEach(
             location -> {
               URI uri = URI.create(location);
-              allowGetObjectStatementBuilder.addResource(
-                  IamResource.create(
-                      arnPrefix + StorageUtil.concatFilePrefixes(parseS3Path(uri), "*", "/")));
               final var bucket = arnPrefix + StorageUtil.getBucket(uri);
               if (allowList) {
                 bucketListStatementBuilder
@@ -264,21 +280,22 @@ public class AwsCredentialsStorageIntegration
                           .addResource(key));
             });
 
-    boolean canWrite = !writeLocations.isEmpty();
     if (canWrite) {
-      IamStatement.Builder allowPutObjectStatementBuilder =
+      IamStatement.Builder allowReadWriteStatementBuilder =
           IamStatement.builder()
               .effect(IamEffect.ALLOW)
+              .addAction("s3:GetObject")
+              .addAction("s3:GetObjectVersion")
               .addAction("s3:PutObject")
               .addAction("s3:DeleteObject");
       writeLocations.forEach(
           location -> {
             URI uri = URI.create(location);
-            allowPutObjectStatementBuilder.addResource(
+            allowReadWriteStatementBuilder.addResource(
                 IamResource.create(
                     arnPrefix + StorageUtil.concatFilePrefixes(parseS3Path(uri), "*", "/")));
           });
-      policyBuilder.addStatement(allowPutObjectStatementBuilder.build());
+      policyBuilder.addStatement(allowReadWriteStatementBuilder.build());
     }
     if (shouldUseKms(storageConfigurationInfo)) {
       addKmsKeyPolicy(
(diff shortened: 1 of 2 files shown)


polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -151,7 +153,17 @@ public StorageAccessConfig getSubscopedCreds(
       StsClient stsClient =
           stsClientProvider.stsClient(StsDestination.of(storageConfig.getStsEndpointUri(), region));
 
-      AssumeRoleResponse response = stsClient.assumeRole(request.build());
+      AssumeRoleResponse response;
+      try {
+        response = stsClient.assumeRole(request.build());
+      } catch (PackedPolicyTooLargeException e) {
+        throw new RuntimeException(
```
jbonofre, line 160: To be honest, I would expect some more specific than `RuntimeException` here but as the rest of the storage layer is also using `RuntimeException`, it's consistent and make sense 😄

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -151,7 +153,17 @@ public StorageAccessConfig getSubscopedCreds(
...
+            "AWS STS rejected the session policy: it exceeds the 2048-character packed-policy"
```
jbonofre, line 161: I was confused when I saw the corresponding `testDeepNestedNamespaceStaysWithinPackedPolicyLimit` as the assert is on the raw JSON length, not the packed size.

But it's actually correct. I was misleaded by the comment and naming.

jbonofre (approved):
LGTM, thanks!

dimas-b:
I do not think this PR "fixes" #3243 for good, but it certainly improves the situation.

Thanks for this fix, @yushesp !

dimas-b:
@yushesp : please fix conflicts

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
dimas-b, line 161: It does not have to be "AWS"... Polaris supports a few non-AWS S3 implementations too 😅 Would you mind keeping just "STS" in the message?

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
```diff
@@ -302,7 +319,26 @@ private IamPolicy policyString(
     bucketGetLocationStatementBuilder
         .values()
         .forEach(statementBuilder -> policyBuilder.addStatement(statementBuilder.build()));
-    return policyBuilder.addStatement(allowGetObjectStatementBuilder.build()).build();
+
+    // Emit a read-only statement when there are read-only locations, or as a fallback with no
+    // resources when no combined read+write statement was emitted — ensuring the inline policy is
+    // always non-empty.
+    if (!readOnlyLocations.isEmpty() || !canWrite) {
```
dimas-b, line 326: Does `!canWrite` add value here? It is set when write locations are empty, but in that case `readOnlyLocations` cannot be empty, right? 🤔

dimas-b (commented):
LGTM overall 👍 ... with minor comments.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
flyrain: Agreed with JB, PackedPolicyTooLargeException is a RuntimeException, wrapping it with `RuntimeException` seems unnecessary. 
I'm also not sure if we need to add more error message here. Isn't the PackedPolicyTooLargeException descriptive enough?

adnanhemani: `PackedPolicyTooLargeException` is itself a `RuntimeException`, so wrapping it in `new RuntimeException(...)` adds an extra layer without a meaningful type distinction. Consider re-throwing it directly (the AWS message is already descriptive), or at least logging the guidance message and then rethrowing the original. Wrapping also makes the catch site harder to target for callers who might want to handle this specifically.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
adnanhemani: Echoing `dimas-b`'s existing comment: Polaris supports non-AWS S3 implementations, so `"AWS STS rejected..."` is slightly too specific. `"STS rejected..."` is more consistent with the rest of the codebase.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
adnanhemani: The `!canWrite` sub-condition is subtle — it handles the edge case where both `readLocations` and `writeLocations` are empty (preserving the existing behaviour of emitting a no-resource GetObject statement rather than an empty policy). A short inline comment here would save the next reader from having to reason through it:
```java
// !canWrite covers the empty-both case: always emit at least one statement
// to avoid sending a completely empty policy to STS.
```

polaris-core/src/test/java/org/apache/polaris/service/storage/aws/AwsCredentialsStorageIntegrationTest.java
```diff
@@ -93,6 +94,9 @@ class AwsCredentialsStorageIntegrationTest extends BaseStorageIntegrationTest {
   public static final PolarisPrincipal POLARIS_PRINCIPAL =
       PolarisPrincipal.of("test-principal", Map.of(), Set.of());
 
+  /** AWS STS inline session policy size limit (packed), in characters. */
+  private static final int STS_PACKED_POLICY_LIMIT = 2048;
+
```
adnanhemani, line 99: `STS_PACKED_POLICY_LIMIT` implies this maps directly to what AWS measures as "packed" size, but `req.policy().length()` is just the raw JSON character count — AWS's actual enforcement is on the URL-encoded form. The check is conservatively safe (raw JSON < 2048 → encoded form typically ≤ 2048), but the name is misleading. Consider `STS_SESSION_POLICY_RAW_CHAR_LIMIT` or add a comment like `// raw JSON length; conservative proxy for the actual packed-policy limit`.

polaris-core/src/main/java/org/apache/polaris/core/storage/aws/AwsCredentialsStorageIntegration.java
adnanhemani: In theory, both readLocations and writeLocations could've been empty and the current behavior still would've sent a non-empty policy (just with no resources attached to it). Maybe a comment here would help clarify this.

adnanhemani:
Happy to take one last look at this once the conflicts are resolved - but I think this is basically "approved" from my end with what I see so far.

@dimas-b - I'm curious why you say this does not fix #3243? It's obvious that at some point, there will always be a policy that could be constructed with "best-practices" that will still be above the max policy size - so my question is, are there some other fixes you had in mind before we can close out #3243?

dimas-b:
> @dimas-b - I'm curious why you say this does not fix #3243 [...]

My point is that #3243 cannot be "fixed completely" because user inputs are a factor here. I'll post a separate comment on that issue.

This PR does improve the situation - so thanks, @yushesp , again for contributing it (and please resolve conflicts :sweat_smile: )

dimas-b:
@yushesp : Do you have time to push this PR over the finish line (resolve conflicts and address comments)?

PR #4232, opened 2026-04-17
Verify image builds with -PNonRESTCatalogs=HIVE,HADOOP

nandorKollar:
This PR adds `-PNonRESTCatalogs=HIVE,HADOOP` parameters to the `docker-image-build` CI job to verify, that Docker image build succeeds with Hive and Hadoop federated catalog extensions. We can avoid merging PRs, which can break the build with these extensions.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,7 +214,7 @@ jobs:
         uses: ./.github/actions/ci-incr-build-cache-prepare
       - name: Image build
         env: *gradle_env_vars
-        run: ./gradlew :polaris-server:assemble -Dquarkus.container-image.build=true
+        run: ./gradlew :polaris-server:assemble -PNonRESTCatalogs=HIVE,HADOOP -Dquarkus.container-image.build=true
       - name: Save image to tar
         run: |
           docker save apache/polaris:latest | zstd -T0 > polaris-image.tar.zst


.github/workflows/ci.yml
```diff
@@ -214,7 +214,7 @@ jobs:
         uses: ./.github/actions/ci-incr-build-cache-prepare
       - name: Image build
         env: *gradle_env_vars
-        run: ./gradlew :polaris-server:assemble -Dquarkus.container-image.build=true
+        run: ./gradlew :polaris-server:assemble -PNonRESTCatalogs=HIVE,HADOOP -Dquarkus.container-image.build=true
```
jbonofre, line 217: It's a good idea. Let's see if we have stable build/test here.

nandorKollar:
Alternatively, we can add an extra step to `build-checks`, which would build Polaris server with each extension enabled.

jbonofre (approved):

.github/workflows/ci.yml
snazy: Hm, that's then no longer what we publish as our official images.
WDYT of having a separate job instead of changing the existing one?

nandorKollar: We can do that if you think that's better. Or we can even have an extra step in `build-checks` job, which would execute `./gradlew :polaris-server:assemble :polaris-server:quarkusAppPartsBuild -PNonRESTCatalogs=HIVE,HADOOP`.

flyrain: I think we will need a check at build time instead of publishing. If the publish CI reuse this one, we need to separate it out.

PR #4224, opened 2026-04-16
CI: Make security workflows CodeQL + Zizmor a real PR-merge-gate

snazy:
The workflows ci-zizmor.yml, ci-codeql.yml and ci-asf-allowlist-check.yml are currently _not_ part of the CI commit-gate (required action). This change moves those workflows into the main ci.yml workflows or adds those to `.asf.yaml`, letting all 3 act as a real PR-merge-gate.

PR #4233, opened 2026-04-17
NoSQL: fix phantom grants for dropped/non-existent entities

adutra:
The `loadGrants()` method never verified the anchor entity exists, always returning `SUCCESS` with a hardcoded version of 1. This violated the `PolarisGrantManager` contract which requires `ENTITY_NOT_FOUND` when the entity is missing, and allowed stale ACLs to surface as valid grants.

Additionally, the bidirectional ACL model produced self-referencing  entries when `securable == grantee`, causing an entity to appear in its own grant list. These are now filtered out on read.

## Checklist

### DIFF
--- a/persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
+++ b/persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
@@ -1137,13 +1137,26 @@ class NoSqlMetaStore extends NonFunctionalBasePersistence {
   }
 
   LoadGrantsResult loadGrants(long catalogId, long id, int entityTypeCode, boolean onSecurable) {
+    PolarisEntityType entityType = PolarisEntityType.fromCode(entityTypeCode);
     LOGGER.debug(
         "loadGrants on {} for catalog:{}, id:{}, entityType:{}({})",
         onSecurable ? "securable" : "grantee",
         catalogId,
         id,
-        PolarisEntityType.fromCode(entityTypeCode),
+        entityType,
         entityTypeCode);
+
+    var anchorEntity = lookupEntity(catalogId, id, entityTypeCode);
+    if (anchorEntity == null) {
+      LOGGER.trace(
+          "Anchor entity for loadGrants does not exist, returning ENTITY_NOT_FOUND: catalog:{}, id:{}, entityType:{}({})",
+          catalogId,
+          id,
+          entityType,
+          entityTypeCode);
+      return new LoadGrantsResult(ENTITY_NOT_FOUND, null);
+    }
+
     var aclName = new GrantTriplet(true, catalogId, id, entityTypeCode).toRoleName();
 
     var collector = new GrantRecordsCollector();
@@ -1164,6 +1177,20 @@ class NoSqlMetaStore extends NonFunctionalBasePersistence {
               onSecurable
                   ? securableAndGrantee.granteeTypeCode()
                   : securableAndGrantee.securableTypeCode();
+
+          // Skip self-referencing entries where the target is the anchor entity itself.
+          // The bidirectional ACL model produces these when securable == grantee.
+          if (targetId == id && targetTypeCode == entityTypeCode) {
+            if (LOGGER.isTraceEnabled()) {
+              LOGGER.trace(
+                  "    Skipping self-referencing grant entry for catalog:{}, id:{}, entityType:{}({})",
+                  targetCatalogId,
+                  targetId,
+                  PolarisEntityType.fromCode(targetTypeCode),
+                  targetTypeCode);
+            }
+            return;
+          }
 
           var indexedAccess = memoizedIndexedAccess.indexedAccess(targetCatalogId, targetTypeCode);
           var entityOptional =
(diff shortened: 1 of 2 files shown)


persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
```diff
@@ -1164,6 +1177,20 @@ LoadGrantsResult loadGrants(long catalogId, long id, int entityTypeCode, boolean
                   ? securableAndGrantee.granteeTypeCode()
                   : securableAndGrantee.securableTypeCode();
 
+          // Skip self-referencing entries where the target is the anchor entity itself.
+          // The bidirectional ACL model produces these when securable == grantee.
+          if (targetId == id && targetTypeCode == entityTypeCode) {
```
dimas-b, line 1183: It is worth adding a test case?... I'm not sure the new test covers it 🤔

adutra: It's indirectly tested: if you remove this if block, `BasePolarisMetaStoreManagerTest#testDropEntities` will fail in `PolarisTestMetaStoreManager.java:926` because `connectedEntity` is actually the `entityToDrop`, which is wrong.

dimas-b: Interesting... but that test existed before this PR... Why was it not broken :thinking:

dimas-b (dismissed):

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
snazy: Good question, and the answer is a bit complex:
If you remove this check, `testDropEntities` fails. If you also remove the check above, `testDropEntities` succeeds.
The test [failure happens](https://github.com/apache/polaris/blob/1a7b21c54e6464b18ec62c5ef744e424f17bcfff/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java#L926) in [this loop](https://github.com/apache/polaris/blob/1a7b21c54e6464b18ec62c5ef744e424f17bcfff/polaris-core/src/testFixtures/java/org/apache/polaris/core/persistence/PolarisTestMetaStoreManager.java#L908) over `granteeEntities`. Overly simplified: the `testDropEntities` relies on the result of `loadGrantsToGrantee`.

snazy: A matching test case for `TestNoSqlMetaStoreManager` could look like this, and I think the assertions are correct - a grant of an entity to the same entity shouldn't be possible.

The interesting part is: If you add the same test case to `BasePolarisMetaStoreManagerTest`:
* `PolarisTreeMapMetaStoreManagerTest` fails with both `loadGrantsOn*` yielding the entity,
* `AtomicMetastoreManagerWithJdbcBasePersistenceImplTest` tests yielding "empty" for `loadGrantsOnSecurable` but not empty for `loadGrantsToGrantee`.

```java
  @Test
  public void testLoadGrantsSkipsSelfReferencingEntries() {
    PolarisBaseEntity principalRole =
        new PolarisBaseEntity(
            PolarisEntityConstants.getNullId(),
            metaStore.generateNewEntityId(callContext).getId(),
            PolarisEntityType.PRINCIPAL_ROLE,
            PolarisEntitySubType.NULL_SUBTYPE,
            PolarisEntityConstants.getRootEntityId(),
            "selfReferencingPrincipalRole");

    EntityResult createResult = metaStore.createEntityIfNotExists(callContext, null, principalRole);
    assertThat(createResult.isSuccess()).isTrue();
    PolarisBaseEntity createdPrincipalRole = createResult.getEntity();
    assertThat(createdPrincipalRole).isNotNull();

    assertThat(
            metaStore
                .grantPrivilegeOnSecurableToRole(
                    callContext,
                    createdPrincipalRole,
                    null,
                    createdPrincipalRole,
                    PolarisPrivilege.PRINCIPAL_ROLE_USAGE)
                .isSuccess())
        .isTrue();

    assertThat(
            List.of(
                metaStore.loadGrantsOnSecurable(callContext, createdPrincipalRole),
                metaStore.loadGrantsToGrantee(callContext, createdPrincipalRole)))
        .extracting(
            LoadGrantsResult::isSuccess,
            LoadGrantsResult::getGrantRecords,
            LoadGrantsResult::getEntities)
        .containsExactly( //
            tuple(true, List.of(), List.of()), //
            tuple(true, List.of(), List.of()));
  }
```

snazy:
Needs a rebase/merge-main after #4224

persistence/nosql/persistence/metastore/src/main/java/org/apache/polaris/persistence/nosql/metastore/NoSqlMetaStore.java
adutra: I can add this test if you think it's useful, but imho we should fix the write side of things: a call to `grantPrivilegeOnSecurableToRole` with a self-grant situation is meaningless and should be disallowed. (That said, I don't think the REST API allows this situation to happen.)

A privilege on entity R granted to R itself can never be consumed by the authorizer: the authorization traversal walks from principal to principal roles to catalog roles to grants, and a principal role granting itself a privilege on itself fits nowhere in that chain. 

The filtering in the NoSQL implementation is fine, but denying on write is better then filtering on read, imho. The self-grant shouldn't be persisted in the first place.

The transactional implementation is wrong: It exposes self-grants to callers, who have no way of knowing they're meaningless.

snazy: Oh, sure, the write itself should be impossible IMO. Both at the API and persistence layer.

snazy (approved):

PR #4236, opened 2026-04-17
Parse Iceberg client language and version from request headers

adutra:
This change introduces a new `IcebergClientInfo` component to identify the Iceberg client library making each REST request.

This information may be useful not only for logging and tracing, but also for future enhancements such as when deciding:

- How to decode path segments (the decoding algorithm is likely to change)
- How to process a remote signing request (the spec is likely to evolve)

The information is sourced from two HTTP headers:

- `X-Client-Version`: carries the client version string for Java and Python.
- `User-Agent`: carries the version for Python, Go, Rust, and C++.

The client info is propagated to:

- OpenTelemetry spans via `TracingFilter`
- SLF4J MDC via `LoggingMDCFilter`

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfo.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfo.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.http;
+
+import java.util.Optional;
+import org.apache.polaris.immutables.PolarisImmutable;
+import org.immutables.value.Value;
+
+/**
+ * Holds parsed information from the Iceberg client HTTP request headers.
+ *
+ * <p>Iceberg REST clients send a standard {@code X-Client-Version} header on every request: a
+ * client-identification string whose format and semantics vary by language implementation (see
+ * below). In addition, all HTTP clients send a standard {@code User-Agent} header, which some
+ * Iceberg client implementations use to carry their library version.
+ *
+ * <p>The following table summarizes the known formats and the source of {@link #icebergVersion()}
+ * for each language:
+ *
+ * <table border="1">
+ *   <tr><th>Language</th><th>{@code X-Client-Version}</th>
+ *       <th>User-Agent</th><th>{@link #icebergVersion()} source</th></tr>
+ *   <tr><td><a href="https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java">Java</a></td>
+ *       <td>{@code Apache Iceberg <lib version> (commit <hash>)}</td>
+ *       <td>{@code Apache-HttpClient/5.x (Java/...)}</td>
+ *       <td>{@code X-Client-Version}</td></tr>
+ *   <tr><td><a href="https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/rest/__init__.py">Python (PyIceberg)</a></td>
+ *       <td>{@code PyIceberg <lib version>}</td>
+ *       <td>{@code PyIceberg/<lib version>}</td>
+ *       <td>{@code User-Agent}</td></tr>
+ *   <tr><td><a href="https://github.com/apache/iceberg-go/blob/main/catalog/rest/rest.go">Go (iceberg-go)</a></td>
+ *       <td>a stale hardcoded constant, not the client version</td>
+ *       <td>{@code GoIceberg/<lib version>}</td>
+ *       <td>{@code User-Agent}</td></tr>
+ *   <tr><td><a href="https://github.com/apache/iceberg-rust/blob/main/crates/catalog/rest/src/catalog.rs">Rust (iceberg-rust)</a></td>
+ *       <td>a stale hardcoded constant, not the client version</td>
+ *       <td>{@code iceberg-rs/<lib version>}</td>
+ *       <td>{@code User-Agent}</td></tr>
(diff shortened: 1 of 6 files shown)


runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfo.java
```diff
@@ -0,0 +1,80 @@
...
+ *       <td>not sent</td>
+ *       <td>{@code iceberg-cpp/<lib version>}</td>
+ *       <td>{@code User-Agent}</td></tr>
+ * </table>
+ */
+@PolarisImmutable
+public interface IcebergClientInfo {
+
+  /** Programming language of an Iceberg REST client, as determined from request headers. */
+  enum Language {
```
flyrain, line 65: Do we need an enum here? I think a string would be more flexible when clients provides a lang not in this list.

runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfo.java
```diff
@@ -0,0 +1,80 @@
...
+    JAVA,
+    PYTHON,
+    GO,
+    RUST,
+    CPP
+  }
+
+  /** Returns the Iceberg client library version, or empty if it cannot be determined. */
+  @Value.Parameter(order = 1)
+  Optional<String> icebergVersion();
```
flyrain, line 75: Would `client version` a more descriptive name? `icebergVersion` may be confused with table versions. 

Not a blocker, just brain storming: should we add a field called client name? For example, PyIceberg is the client name, 0.9.1 is the client version.

flyrain: Basically, we could have 3 field `Client language`, `Client name`, and `Client version`. WDYT?

flyrain (commented):
Thanks @adutra for adding this. It's a very useful feature to me. Left some minor comments.

runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfo.java
adutra: The language is inferred, not provided. So I think an enum is fine.

runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfo.java
adutra: Not opposed to that, but wouldn't `client name` be a bit redundant with `client language`? I.e. `PyIceberg` vs `PYTHON`.

dimas-b (commented):
Looks like a useful feature to me 👍

runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfoFactory.java
```diff
@@ -0,0 +1,100 @@
...
+        if (end > start
+            && versionHeader.regionMatches(end, " (commit ", 0, " (commit ".length())
+            && versionHeader.endsWith(")")) {
+          icebergVersion = versionHeader.substring(start, end);
+          language = IcebergClientInfo.Language.JAVA;
+        }
+      }
+    }
+
+    // 2. Parse User-Agent for other clients, which (except for Python) send a stale, hardcoded
```
flyrain, line 67: +1 on the logic here. I think `User-Agent` is more reliable, except for Java clients which don't currently include Iceberg version metadata there. This leads to two implications::
1. Given `X-Client-Version` isn't part of IRC spec. It is a convention only working for Java clients, we can change the Iceberg Java clients to send back Iceberg version informations via header `User-Agent`, so that the iceberg clients across language behavior are more consistent. However, we still have to keep this logic in Polaris to be compatible with the old clients until they are irrelevant. 
2. Alternative option: Make `X-Client-Version` part of the IRC, so that other clients need to follow it.

| Language | X-Client-Version | User-Agent | Version Extraction Strategy | Implementation Status |
|----------|------------------|------------|----------------------------|----------------------|
| **Java** | `Apache Iceberg 1.10.1 (commit abc123...)` | `Apache-HttpClient/5.4 (Java/11.0.25)` | Parse X-Client-Version | ⚠️ User-Agent not helpful |
| **Python** | `PyIceberg 0.9.1` | `PyIceberg/0.9.1` | Parse User-Agent (X-Client-Version also works) | ✅ Consistent |
| **Go** | `0.14.1` | `GoIceberg/0.5.0` | Parse User-Agent | ⚠️ X-Client-Version is spec version |
| **Rust** | `0.14.1` | `iceberg-rs/0.3.0` | Parse User-Agent | ⚠️ X-Client-Version is spec version |
| **C++** | *(not sent)* | `iceberg-cpp/0.3.0-SNAPSHOT` | Parse User-Agent | ⚠️ Missing X-Client-Version |

runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfoFactory.java
```diff
@@ -0,0 +1,100 @@
...
+
+      } else if (userAgentHeader.startsWith("GoIceberg/")) {
+        // Go: "GoIceberg/<version>"
+        int start = "GoIceberg/".length();
+        icebergVersion = userAgentHeader.substring(start);
+        language = IcebergClientInfo.Language.GO;
+
+      } else if (userAgentHeader.startsWith("iceberg-rs/")) {
+        // Rust: "iceberg-rs/<version>"
+        int start = "iceberg-rs/".length();
```
flyrain, line 85: I don't think language inferring is reliable. The Iceberg rust clients may change it to `Iceberg-rs` or following the go's convention `RustIceberg`.  I'd suggest to keep the client name as is. I'm also concerned the version inferring, which isn't necessary reliable, what if next rust/go version starts to adopt the Java client information format `Apache Iceberg 1.10.1 (commit abc123...)`.  Do we have strong use case for client versions? If not, I'd suggest to just keep as is in Polaris. Then we leave the parsing logic to the downstream.

singhpk234:
This is really nice change @adutra thank you for adding this !

runtime/service/src/main/java/org/apache/polaris/service/http/IcebergClientInfoFactory.java
adutra: > Do we have strong use case for client versions?

Not now, but maybe one day e.g. in order to spot Java clients using the wrong percent-encoding to encode URL path segments (see https://github.com/apache/iceberg/pull/15989).

> The Iceberg rust clients may change it to Iceberg-rs or following the go's convention RustIceberg

While they *could* do this, I don't think they *would*. Changing the format of a well-established `User-Agent` header is a delicate operation.

adutra: And to complete my answer: a component like this is necessarily brittle, and can only operate on a best-effort basis. If clients change their headers, this component will need to be updated accordingly.

flyrain: I'm hesitant to rely on brittle inference for any feature. If Polaris ever need to depend on some information from client-side infor. to switch logic. I'd prefer explicit ways, like spec changes or headers.

PR #4238, opened 2026-04-18
Refactor: Move ETag computation from adaptor into IcebergCatalogHandler

Wency42601:
## Summary
Push the ETag computation out of `IcebergCatalogAdapter` (a JAX-RS
transport layer) and into `IcebergCatalogHandler`, where the semantics
of "a load-table response's etag" actually live. The adapter now only
copies the handler-provided etag onto the HTTP `ETag` header.
This also addresses the existing TODO in `IcebergCatalogHandler.loadTable`
about creating a canonical interface for associating etags with entities.
## Changes
- New `ETaggedLoadTableResponse` record pairing a `LoadTableResponse`
  with an `Optional<String>` etag.
- `IcebergCatalogHandler`
  - New private `withETag(LoadTableResponse, TableIdentifier)` helper
    that centralizes the "metadata location -> etag" rule and the
    previously-adapter-side warning when `metadataLocation` is null.
  - Adapter-facing overloads now return the wrapper:
    - `createTableDirect(ns, req, delegationModes, refreshEndpoint)`
    - `registerTable(ns, req)`
    - `loadTable(id, snapshots, IfNoneMatch, delegationModes, refreshEndpoint)`
      (returns `Optional<ETaggedLoadTableResponse>`).
  - Simpler test-facing overloads (`loadTable(id, snapshots)`,
    `loadTableIfStale`, `createTableDirect(ns, req)`, etc.) keep their
    existing `LoadTableResponse`-shaped contracts by unwrapping via
    `.response()` / `.map(ETaggedLoadTableResponse::response)`.
- `IcebergCatalogAdapter`
  - Removed `tryInsertETagHeader`.
  - Added `toResponseBuilder(ETaggedLoadTableResponse)`; only sets
    `HttpHeaders.ETAG` when the handler supplied one.
  - Updated `createTable`, `loadTable` (still 304s on `Optional.empty()`),
    and `registerTable` call sites.
  - Removed now-unused imports.
## Test plan
- [x] New unit tests in `IcebergCatalogAdapterETagTest` covering the
      full adapter -> handler path end-to-end (6 tests, all pass):
  - `createTableDirect` sets `ETag` to
    `IcebergHttpUtil.generateETagForMetadataFileLocation(metadataLocation)`.
  - Staged create omits `ETag` (no metadata location yet).
  - `loadTable` returns the same `ETag` as `createTable`.
  - `loadTable` with matching `If-None-Match` -> `304 Not Modified`.
  - `loadTable` with mismatched `If-None-Match` -> `200` plus current
    `ETag`.
  - `registerTable` response carries the expected `ETag`.
- [x] Existing suites still green: `IcebergCatalogHandlerAuthzTest`
      (6005 tests), `IcebergCatalogAdapterTest` (17), `IfNoneMatchTest` (16).
- [x] `:polaris-runtime-service:compileJava` and `compileTestJava`
      succeed with no new warnings.

## Checklist

### DIFF
--- a/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/ETaggedLoadTableResponse.java
+++ b/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/ETaggedLoadTableResponse.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.polaris.service.catalog.iceberg;
+
+import java.util.Optional;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+
+/**
+ * Pairs a {@link LoadTableResponse} with an optional entity tag derived from the response's
+ * metadata location. The handler is responsible for computing the etag (when possible); transport
+ * layers such as the REST adapter translate the etag into an HTTP {@code ETag} header.
+ */
+public record ETaggedLoadTableResponse(LoadTableResponse response, Optional<String> etag) {}
--- a/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapterETagTest.java
+++ b/runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapterETagTest.java
@@ -1,0 +1,316 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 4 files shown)


runtime/service/src/test/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogAdapterETagTest.java
tm0nk, line 1: The Gradle Build Checks indicate there are format violations in this file that can be fixed with `./gradlew spotlessApply`

flyrain:
Hi @Wency42601 , do we still need this PR?

runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/ETaggedLoadTableResponse.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Optional;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+
+/**
+ * Pairs a {@link LoadTableResponse} with an optional entity tag derived from the response's
+ * metadata location. The handler is responsible for computing the etag (when possible); transport
+ * layers such as the REST adapter translate the etag into an HTTP {@code ETag} header.
+ */
+public record ETaggedLoadTableResponse(LoadTableResponse response, Optional<String> etag) {}
```
flyingImer, line 29: Optional<String> etag is the narrow form. I fell a canonical one with an open extension slot (LoadTableResult<E>, empty default) can benefit broader. This provides iceberg catalog impl provider an official operation metadata channel to pass down info like Etag cross boundary (runtime/ and core/) easily without hacking around. And create, register, and later metadata reuse one carrier instead of a wrapper per field if needed.

flyingImer (commented):
Good change, worth finishing. ETaggedLoadTableResponse puts the version identity where its meaning lives, and since the handler already owned the freshness decision it also drops the etag computation the adapter duplicated. 

The one push: the etag is the first field of an operation's result metadata, and this carrier is where that generalizes or hardens into a one-off. Worth shaping as a general result-with-metadata, not an etag-specific wrapper. Direction, not a blocker.

PR #4240, opened 2026-04-18
Migrate polaris-core to JSpecify null annotations

viditochani:
Migrates `polaris-core` from Jakarta `@Nullable`/`@Nonnull` to JSpecify `@Nullable`/`@NonNull` annotations. Same motivation as #4130: Jakarta's null annotations lack `@Target(ElementType.TYPE_USE)`, causing IDE and expressiveness issues.

Part of #4124

### Changes

Mechanical import swap across polaris-core. A few places needed manual fixes where JSpecify's TYPE_USE targeting doesn't allow annotations on scoping constructs. For example, `@NonNull BaseResult.ReturnStatus` became `@NonNull ReturnStatus`, and a fully-qualified type reference was updated to use a short import.

No new dependencies because `polaris-core` already has JSpecify available transitively.

### Testing

Verified with `./gradlew :polaris-core:build` (compilation, tests, checkstyle, and formatting all pass).

### Checklist
- [x] 🔗 Clearly explained why the changes are needed, or linked related issues: #4124


snazy:
Looks like JSpecify's `@Nullable` causes issues with Immutables, causing a bunch of new compilation warnings.
I think, those should be fixed first, until we change more things:

Repro:
```
./gradlew :polaris-persistence-nosql-api:compileTestFixturesJava --rerun
```

```
> Task :polaris-persistence-nosql-api:compileTestFixturesJava
./persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/AnotherTestObj.java:33: warning: (immutables) Type parsing problem in FromSupertypesModel: java.lang.IllegalArgumentException: Cannot parse type from input string 'byte @org.jspecify.annotations.Nullable []'. unexpected trailing terms '@org.jspecify.annotations.Nullable[]'
public interface AnotherTestObj extends Obj {
       ^
        at org.immutables.value.processor.encode.Type$Parser.doParse(Type.java:940)
./persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/AnotherTestObj.java:33: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface AnotherTestObj extends Obj {
       ^
./persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/CommitTestObj.java:33: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface CommitTestObj extends BaseCommitObj {
       ^
./persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/SimpleTestObj.java:33: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface SimpleTestObj extends Obj {
       ^
./persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/VersionedTestObj.java:30: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface VersionedTestObj extends Obj {
       ^
5 warnings
```

flyrain (dismissed):
LGTM. cc @HonahX for more visibility.

dimas-b:
I second @snazy 's concern (above). Let's address those warning before we merge.

viditochani:
@snazy This looks like an upstream Immutables issue with TYPE_USE `@Nullable` on concrete array types. `FromSupertypesModel` doesn't handle `byte @Nullable []` syntax. I filed immutables/immutables#1655 to confirm.

Looking at the generated code, fields are nullable and builders accept null. The `.from()` supertype overload drops the attribute, while the concrete overload works (and is what our tests use).

Since this only affects test fixtures in `persistence/nosql`, should we wait for upstream confirmation or suppress the warning for now? (Changing to `@Nullable byte[]` would silence it but changes Immutables behavior from optional to required.)

snazy:
@viditochani My concern is not about that it doesn't work at the moment, but that it produces a bunch of new warnings. Adding more warnings adds more noise, which may hide other, really important warnings.

Not sure whether suppressing `immutable`globally  is feasible, as it suppresses more warnings related to Immtable, so more than just the concrete one. If that works against the affected fields, that might work.

BTW: `Changing to @Nullable byte[] ` doesn't work, because that's the wrong target for JSpecify's `@Nullable`.

I'd just like to have some clarity that the change to JSpecify doesn't introduce other "silent" issues or adds more noise to the compiler output.

viditochani:
@snazy This PR introduces no new warnings. I verified with `--rerun` across `compileJava`, `compileTestJava`, and `compileTestFixturesJava`. The warnings are only in `persistence/nosql` (#4130), in 4 test fixture files that use `byte @Nullable []`. No other files in the migration use that pattern either.

But I agree that we don't want to introduce any "silent" issues by ignoring these warnings and/or add noise to compiler output. What I was suggesting is a targeted suppress on those 4 test fixtures (there is also precedent for this in the codebase).

viditochani:
@snazy Any other concerns, or are we good to move forward with this PR? I can create a separate PR to add a targeted suppress on those 4 test fixtures in `persistence/nosql`.

viditochani:
@flyrain Would you mind re-approving? Your earlier approval was dismissed by the rebase.

flyrain (dismissed):

viditochani:
@flyrain Rebased to resolve conflicts, need another re-approval :) This PR touches a lot of files. Is it possible to merge it quickly to avoid another rebase and re-approval?

flyrain:
Hi @viditochani, thanks for working on it. There is another conflict. Would you mind rebasing it again? 

Hi @dimas-b and @snazy, please share your concern if you are not OK with the merge. Thanks.

flyrain (dismissed):

viditochani:
@flyrain Thanks for the re-approval. Could you merge this before we get the next batch of conflicts? :)

dimas-b:
I'd appreciate @snazy reviewing because of the earlier concern:

https://github.com/apache/polaris/pull/4240#issuecomment-4287215836

snazy:
The jspecify migration makes a lot of sense to me and I'd love to see this eventually land across the code base.

What I'm concerned about is the immutable-warnings saying that not all properties are correctly handled, but the build passes. Omitting a property during a `.from()` can cause, as the warning message says, appears worrisome and, assuming the warning message is correct there, could lead to issues causing undetected failure scenarios.

The mentioned issue only affects some test-fixtures in a currently harmless way. That's okay for me, as the tests still pass.

I'd feel more comfortable to wait until https://github.com/immutables/immutables/issues/1655 has been fixed and the warnings disappears.

flyrain:
> The mentioned issue only affects some test-fixtures in a currently harmless way. That's okay for me, as the tests still pass.

Hi @snazy , thanks for the thoughtful review. Is your comment (https://github.com/apache/polaris/pull/4240#issuecomment-4433762247 ) a blocker? Are you OK to merge this PR?

dimas-b:
Given the above discussion and #4124, it looks like the choice is between one kind of developer inconvenience vs. another kind of developer inconvenience. 

Actual runtime behaviour is not affected, as far as I can tell.

From my POV we should wait for https://github.com/immutables/immutables/issues/1655 and then have a well-rounded PR.

viditochani:
@dimas-b @snazy Thanks for the comments. Just want to make sure I understand - should we hold this PR even though it doesn't introduce any new warnings? The Immutables warnings are from the already merged PR. I've offered to help fix the upstream issue (immutables/immutables#1655).

flyrain:
Thanks for the confirmation and voluntarily working in the upstream, @viditochani !

Given that this PR does not introduce any new warnings, what would be the reason to block it while waiting for the upstream fix? @dimas-b @snazy

dimas-b:
@flyrain : My understanding is that this PR does introduce new warnings: https://github.com/apache/polaris/pull/4240#issuecomment-4433762247

I'd prefer not to expand the area of application of JSpecify annotations until we get a new Immutables library.

https://github.com/immutables/immutables/issues/1655 is already fixed :)

flyrain:
> My understanding is that this PR does introduce new warnings

What are they?

If you run `./gradlew :polaris-persistence-nosql-api:compileTestFixturesJava --rerun` against the current main, the warnings are there already.

```
> Task :polaris-persistence-nosql-api:compileTestFixturesJava
/Users/ygu/asf/polaris/persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/SimpleTestObj.java:33: warning: (immutables) Type parsing problem in FromSupertypesModel: java.lang.IllegalArgumentException: Cannot parse type from input string 'byte @org.jspecify.annotations.Nullable []'. unexpected trailing terms '@org.jspecify.annotations.Nullable[]'
public interface SimpleTestObj extends Obj {
       ^
        at org.immutables.value.processor.encode.Type$Parser.doParse(Type.java:940)
/Users/ygu/asf/polaris/persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/SimpleTestObj.java:33: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface SimpleTestObj extends Obj {
       ^
/Users/ygu/asf/polaris/persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/AnotherTestObj.java:33: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface AnotherTestObj extends Obj {
       ^
/Users/ygu/asf/polaris/persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/VersionedTestObj.java:30: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface VersionedTestObj extends Obj {
       ^
/Users/ygu/asf/polaris/persistence/nosql/persistence/api/src/testFixtures/java/org/apache/polaris/persistence/nosql/api/obj/CommitTestObj.java:33: warning: (immutables:from) Generated builder '.from' method will not copy from attribute 'binary' because it has some not-yet-generated type, or different return type in supertype  (we cannot handle generic specialization or covariant overrides yet). Sometimes it is possible to avoid this by providing abstract override method in this value object
public interface CommitTestObj extends BaseCommitObj {
       ^
5 warnings
```

viditochani:
The Immutables fix has been released (2.12.2) and merged into Polaris. The warnings are gone. Can we merge this PR now? :) @snazy @dimas-b

dimas-b:
@viditochani : Could you rebase to include #4469 into this PR (for local testing)?

polaris-core/src/main/java/org/apache/polaris/core/auth/AuthorizationRequest.java
```diff
@@ -76,7 +73,7 @@ default List<PolarisSecurable> getTargets() {
    *
    * <p>Compatibility accessor derived from {@link #getTargetBindings()}.
    */
-  @Nonnull
+  @NonNull
```
dimas-b, line 76: Changes LGTM, but we should probably have an explicit (compile-time) dependency on JSpecify... WDYT?

dimas-b: related: does `polaris-core` still need `jakarta.annotation`?

viditochani: > Changes LGTM, but we should probably have an explicit (compile-time) dependency on JSpecify... WDYT?

Yea, that is a good idea. Let me do that.

viditochani: > related: does `polaris-core` still need `jakarta.annotation`?

Got added again in rebase. Removing it.

dimas-b (dismissed):
Thanks, @viditochani !

snazy (dismissed):

dimas-b:
@flyrain : any concerns with merging from your side?

flyrain:
All good from my side. I have approved it long time ago. With that, there is a new conflict. Would you mind rebasing it, @viditochani ?

flyrain (approved):

dimas-b (approved):

flyrain:
Thanks a lot for the fix and your patience, @viditochani ! Thanks @dimas-b @snazy for the review.

viditochani:
Thank you all. I will send the next one later today or tomorrow :)

